PR #748, opened 2015-09-30
PCHR-423: Modal css fixes

AkA84:
#### CSS-only custom select dropdown
Instead of relying on the [jQuery plugin](https://github.com/adamcoulombe/jquery.customSelect), I wrote a css-only responsive select dropdown, `.chr_custom-select`:

##### Standard
custom-select
custom-select-responsive

```html
<div class="chr_custom-select">
  <select>
    <option value="">- select -</option>
    <!-- other options... -->
  </select>
</div>
```
Just wrap the select in a div with the `.chr_custom-select` class applied to it

##### Full width
custom-select-full
```html
<div class="chr_custom-select chr_custom-select--full">
  <select>
    <option value="">- select -</option>
    <!-- other options... -->
  </select>
</div>
```
For when the select must occupy the full width of its container

##### Transparent
custom-select-transparent
```html
<div class="chr_custom-select chr_custom-select--transparent">
  <select>
    <option value="">- select -</option>
    <!-- other options... -->
  </select>
</div>
```
For when the select must have the same color of its container

##### IE9 fallback
select-ie9-fallback
When the body has a `ie9` class, then the custom arrow is removed, given that in IE9 it's not possible to hide it

#### Custom outline buttons
Inspired by the upcoming Bootstrap 4 [outline butons](http://v4-alpha.getbootstrap.com/components/buttons/#outline-buttons):
outline-primary
outline-success
outline-warning
outline-danger
```html
<button class="btn btn-primary-outline">Outline primary</button>
<button class="btn btn-primary-success">Outline success</button>
<button class="btn btn-primary-warning">Outline warning</button>
<button class="btn btn-primary-danger">Outline danger</button>
```

#### Added `.input-group-unstyled` variation
A variation of Bootstrap's [`.input-group`](http://getbootstrap.com/components/#input-groups), where the icon next to the field doesn't have any background or borders
calendar-button-after

```html
<div class="input-group input-group-unstyled">
  <input class="form-control" type="text">
  <span class="input-group-btn">
    <i class="fa fa-calendar"></i>
  </span>
</div>
```

##### Before
asterisks-before

##### After
asterisks-after


colemanw:
Hang on a minute... why is this needed? CiviCRM core already ships with [select2](http://select2.github.io/select2/) and uses it basically everywhere. This seems to be reinventing that wheel.

gregmeszaros:
Hi Coleman,

These changes are only for our custom civihr angular extension pages. That select2 what is in civi already doesn't seems to fit to our designs and it's not easy to change the current implementation without changing it everywhere in civi.

colemanw:
What did you try in terms of getting select2 integrated with your designs? It's a very versatile plugin and all settings for it can be overridden on a per-instance basis.

gregmeszaros:
but this is civi 4.5.x version, i thought the angular implementation for select2 is only in the later versions?

colemanw:
No it was added in 4.5 - the case configuration UI was done in Angular for 4.5 and that uses select2.

PR #759, opened 2015-11-09
PCHR-400: Cost Center field is not searchable and can save any random value as input

kkalamarski:


kkalamarski:
Added required changes

PR #801, opened 2015-12-04
PCHR-575: Outer layout (AngularJS)

AkA84:
This PR plugs in an AngularJS apps to the appraisals page. The app is structured as described in #796.

In this PR also the main routes (`/dashboard`, `/import`, `/profile`) of the app are defined, and we're using `ui.router` instead of the default `ngRouter` module as the route manager.

For now the focus is on the dashboard page, so the other two pages contain only placeholder templates.

routes

A new gulp task has also been added, `requirejs-bundle`. This tasks simply runs the requireJS optimizer, but coupled with a watcher on any js file in `js/src/` removes the annoyance of having to run `r.js` on every change


kkalamarski:
Approved

PR #802, opened 2015-12-07
PCHR-576: Dashboard (AngularJS)

AkA84:
This PR plugs in AngularJS components (services, directives, controllers) to the Appraisals dashboard page

Things done:
* Implemented `ui-bootstrap` directives like `modal`, `dropdown` and `collapse`
* Wrote `AppraisalsDashboardCtrl` and `AppraisalCycleModalCtrl` controllers
* Wrote the `AppraisalCycle` model
* Wrote the basic `api` service, and its `api.appraisals` extension
* Wrote the `crm-grades-chart` and `crm-show-more` directives
* Customized the style of the bootstrap datepicker, to match the design
* Wrote unit tests for controllers, directives and services

What the PR **does not** contain is any actual implementation of the api communication with the CiviCRM/HR backend endpoints (that will be the focus of the next task). All the responses from the api are mocked, resolved with static data. The structure of the model data and the model api are consequently just an approximation, and will be certainly amended once the real api is plugged in 

#### common/bundles
In the common dependencies extension (#796) a `modules/` folder was used to store a `vendor.js` file that bundled all the vendor files together. The file, along with a new bundle file for the apis (see below) is now moved in the more appropriately named `common/bundles` folder. `common/modules/` is used to store actual angular modules instead
```js
// build.js
// ...
include: [
    'common/bundles/vendors',
    'common/bundles/apis'
]
```
#### `api` service
The communication between the CiviCRM/HR api endpoints and the angular apps will be done via the common `api` service, registered on the `common.apis` angular module. The service's responsibility is simply to correctly form and send requests to the backend and return a `Promise` object, and as such its interface consists of simple methods like `sendGET()` and `sendPOST()`.

`api` can (and should) be extended by other services via the `extend` method, which allows the creation of more specialised api services.
```js
angular('common.apis').factory('api.specializedApi', ['api', function (api) {
    return api.extend({
        specializedMethod1: function() { //... },
        specializedMethod2: function(param1, param2) { //... }
        // ....
    });
```
For example, this PR introduces the `api.appraisals` methods, which exposes methods like `all()`, `create()`, `find()`, etc, specifically for appraisals. Under the hood all it does is to delegate to the the `api` methods `sendGET()` and `sendPOST()` and return their `Promise` objects.
```js
// common/services/api/appraisals.js

//...
find: function (id) {
   // ...
   return this.sendGET(params);
}
```

This mechanism should promote the creation of distinct api services like `api.appraisals`, `api.contacts`, `api.tasks`, etc, which should be far easier to use and maintain rather than a single monolithic api service exposing methods for every CiviCRM/HR entity

#### Model
A Model is mainly the middle man between a controller/directive and the api, so that the communication with the api doesn't occur directly. Other than allowing a better abstraction, it also makes possible for the Model to normalise the data returned from the api, so that it can be used in the app

Example of the `AppraisalCycle` model in action:
```js
// appraisals/controllers/modals/appraisal-cycle-modal-ctrl.js
function createCycle() {
    AppraisalCycle.create(vm.cycle).then(function (cycle) {
        $rootScope.$emit('AppraisalCycle::new', cycle);
        $modalInstance.close();
    });
};
```
```js
// appraisals/models/appraisal-cycle.js
create: function (attributes) {
    // ...
    return appraisalsAPI.create(attributes).then(function (newCycle) {
        // ...
        return newCycle;
    });
},
```

#### `crm-grades-chart` directive
grades-chart
```html
<div crm-grades-chart chart-data="ctrl.chartData"></div>
```
It takes care of displaying the d3 chart on the page, using the data passed by the controller via the `chart-data` attribute

#### `crm-show-more` directive
show-more
```html
<div crm-show-more callback="ctrl.loadItems()" done="ctrl.loadingDone">
    <article ng-repeat="item in ctrl.items">
        <!-- ... -->
    </article>
</div>
```
Wrap this directive over a list of items to display a "show more" button that triggers the next page of items. It has two mandatory parameters, the function to call in order to load more items, and a flag that notifies it when all the items are loaded (the page/page size/total count logic is kept in the controller owning the directive)

#### Test run automation
The gulp tasks suite has been extended with some tests-related functionality:
* `gulp test` makes a single run of all the tests
* The `gulp watch` task can now automatically trigger a test run. There are three possible scenarios:
  * A test file (in `js/test`) has been changed. Only the tests in the file are run
  * A src file with an equivalent test file has been changed. Only the tests in the test file are run. Examples of "equivalent test files": 
    * `src/appraisals/controllers/my-controller.js` -> `test/controllers/my-controller_test.js`
    * `src/appraisals/model/my-model.js` -> `test/model/my-model_test.js`
    * `src/appraisals/controllers/modal/my-modal.js` -> `test/controllers/modal/my-modal_test.js`
  * A src without an equivalent test file has been changed. All the tests are run


kkalamarski:
approved

PR #807, opened 2015-12-15
PCHR-577: Dashboard (CiviHR api)

AkA84:
#### Show all available statuses
statuses

#### Show all available cycle types
types

#### Total cycles count
total-cycle-count

#### Create cycle
create-cycle

#### Update cycle
update-cycle

#### List cycles
cycle-list

#### Paginate cycles
paginate-list

#### Filters cycle
filter-list

### Features NOT plugged-in yet to CiviHR API
These are features still missing, mostly because we don't have yet either the model structure for them, or we don't have custom endpoints for them:
* **Active cycles count** - There is no "active" boolean flag on the model
* **Appraisals count in all (active) cycles** - There is no "active" boolean flag on the model
* **Filters for active state** - There is no "active" boolean flag on the model
* **Status Overview** - There is no custom endpoint
* **Appraisals count in each step** - There is no custom endpoint
* **Next due field** - The way the steps due date should be defined is not clear yet
* **Appraisals count** - To be done
* **Completion percentage** - To be done
* **View cycle** - The single appraisal page is not yet done

#### `api.optionGroup`
This is another api-related service, specific for the `OptionGroup` CiviCRM entity. Its interface offers a way to quickly get all the `OptionValue` entities of a group, via its name
```js
// get all the appraisal status values
optionGroupAPI.valuesOf('appraisal_status');
```

#### Model instance
The idea here is to have models, like `AppraisalCycle`, returning from their methods "instances" of themselves, each instance offering an interface to perform operations on the specific record that the instance represents. Examples:
```js
// update a cycle
var cycle = AppraisalCycle.find('1234'); // returns an instance representing the cycle with id = '1234'

cycle.name; // "Old name"

cycle.name = 'New name';
cycle.update();

AppraisalCycle.find('1234').name; // "New name"
```

```js
// destroy a group of cycles (not yet implemented, just an example!)
var cycles = AppraisalCycle.all({ 'cycle_start_date': { '>': '2015-07-01' } }); // returns an array of instances

$q.all(cycles.map(function (cycle) {
    return cycle.destroy();
}))
.then(function () {
    console.log('Cycles destroyed successfully');
});
```

For now there is only one instance (`AppraisalCycleInstance`), but its interface it's made so that most of its methods are going to be placed in a more generic instance that other instances will be able to delegate to. The "common interface" is as follows:
* `init(attributes, fromAPI)` - Initializes the instances and normalizes the data if it comes directly from the CiviCRM api
* `attributes()` - Returns a plain no-prototype Object only the instance's data, without any of its methods
* `fromAPI()` - This one must be overridden in each specific instance (in the generic instance it will just be empty), and it will contain the logic that will normalize the data coming from the API to a format more suitable for the app to use
* `toAPI()` - The opposite of `fromAPI()`, it will normalize the current instance attributes values to a format that the CiviCRM API can use


uk.co.compucorp.civicrm.appraisals/js/test/mocks/models/appraisal-cycle.js
```diff
@@ -37,204 +39,207 @@ define([
...
-                        { name: 'Awaiting HR Approval', count: 11 },
-                        { name: 'Complete', count: 2 }
+            /**
+             * Mocked cycles
+             */
+            mockedCycles: function () {
+                return {
+                    total: 10,
+                    list: [
+                        {
```
kkalamarski: You're using other appraisal cycle model to mock appraisal cycles in /services/api/appraisal.js
Why it's needed to duplicate this code, and why isn't it consistent?

kkalamarski:
@AkA84 just this one issue, if it's explained/corrected it's approved

uk.co.compucorp.civicrm.appraisals/js/test/mocks/models/appraisal-cycle.js
AkA84: Right, forgot to clean things up!

The mocked cycles in `api.appraisals` were used before we could get the cycles list directly from the API. Once I implemented the real `api.appraisals.all()` method, i forgot to remove them

For `api.appraisals.activeCycles()` I still need to use a fake list since we don't have the `active` flag on the model, but definitely I don't need anymore all that detailed list

PR #810, opened 2015-12-17
PCHR-786 Angular-Date Module

kkalamarski:
# Angular-Date Module

Contains set of Directives, Filters, Services and Controllers that provide additional (or adjust) functionality of BootstrapUI module.

Short description:

### CustomDateInput
Formats dates in textfields to chosen format.
If date is invalid or provided value is 'Unspecified' it returns empty string.
directive
Usage:
```html
<input custom-date-input />
```

### DateValidationService
Provides set of methods that check whether date is valid. 
Depends on CustomDateFilter. 
And returns errors (if any) as callback.
Errors

### DateFactory
Converts Javascript Date Object/String/Timestamp or whatever to momentJs object.
Contains one method:
```javascript
createDate: function createDate(dateString, format, strict){
    return this.moment.apply(null, arguments);
}
```

### CustomDateFilter
Extends functionality of angular DateFilter. Uses momentJs for parsing.

Usage:
```javascript
$filter('CustomDate')(value);
```
```html
{{ value | CustomDate }}
```
List of acceptable formats:
```javascript
'DD-MM-YYYY',
'DD-MM-YYYY HH:mm:ss',
'YYYY-MM-DD',
'YYYY-MM-DD HH:mm:ss',
'DD/MM/YYYY',
'x' // timestamp

```
If date is invalid displays string: 'Unspecified'.

### DatePickerController
Extends BootstrapUI's default controller for directive 'datepicker'

### DatepickerDirectiveDecorator
Decorates 'datepicker' directive.
Changes order of displayed days.
days

### DatepickerPopupDirectiveDecorator
Decorates 'datepicker-popup' directive.
Sets default date format for all datepicker popups. No longer we need to specify it on input.

Usage before:
```html
<input datepicker-popup="dd/mm/yyy" />
```
Usage now:
```html
<input datepicker-popup />
```

### DatepickerPopupWrapDirectiveDecorator
Overrides default directive's template.

### DaypickerDirectiveDecorator
Overrides default directive's template.

## Templates

The src/templates directory contains custom templates used for i.e to override directives or provide additional features to directives or controllers.

### templates.js
Angular module which contains all templates from the template directory. It is a dependency for angular-date.

### day.html
Provides new look for daypicker directive.
daypicker

### datepickerPopup.html
Removes redundant parts of popup template.
daypicker


org.civicrm.reqangular/src/common/angular-date/src/services/DateValidationService.js
```diff
@@ -56,7 +56,7 @@ function DateValidationService($filter) {
 
     me.validate = function validate(start_date, end_date) {
```
AkA84: I haven't seen this file before, so now I can't comment on it except for this few lines that you changed from the last PR. I'm gonna bundle every note here then

----
```js
me.start_parts = [];
me.end_parts = [];
```
*(wherever else applies)*
You are mixing `snake_case` and `camelCase` for variable names, please stick to `camelCase`

---
```js
me._error = function (error_msg, fields) {
    throw new Error(error_msg);
};
```
*(wherever else applies)*
No matter if we use CommonJS or AMD, if a function is private (I believe that's why you prefixed it with `_`), you can just put it outside of the object/function you are returning from the module. This way the module api is comprised of actual "public methods" only:
```js
function error (error_msg, fields) {
    throw new Error(error_msg);
};

function DateValidationService($filter) {
    // ...
}

module.exports = DateValidationService;
```

---
```js
me._error = function (error_msg, fields) {
    throw new Error(error_msg);
};
```
It doesn't seem you are using `fields` for anything?

---

```js
if (!!(error && error.constructor && error.call && error.apply)) {
    me._error = error;
} else {
    throw new TypeError('Error callback must be a function.');
}
```
Why not simply `typeof error === 'function'`, or since we have [Lodash](https://lodash.com/) as a common dependency, you could inject that and use the [`_.isFunction`](https://lodash.com/docs#isFunction) utility

---

```js
if(!!start_date)
```
*(wherever else applies)*
JS already coerces the value to a boolean inside an `if()`, it's not necessary to do it explicitly

---
```js
if(!me.start_date || !!start_date) me.setDates(start_date, end_date);
```
*(wherever else applies)*
Please don't use this inline form, for consistency rewrite as:
```js
if (!me.start_date || !!start_date) {
    me.setDates(start_date, end_date);
}
```
----
```js
if(me.end) {
```
*(wherever else applies)*
Make sure the code style is consistent, put a space between `if` and `(` like everywhere else
```js
if (me.end) {
```

----
```js
me.setOptions = function setOptions(newOptions) {
    angular.extend(me.options, newOptions);
};
```
You are using `angular` as a global object instead of injecting it into the module

---
You think it would be better to use [Moment.js](http://momentjs.com/) in this service? It has already built in functionality for parsing dates, checking if a date is before/after an another, checking if the date is valid and so on

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
```diff
@@ -0,0 +1,20 @@
+var moment = require('../../../vendor/moment.min.js');
+
+function DateFactory(){
```
AkA84: I'm not sure I understand this. If `DateFactory` is just a wrapper around moment.js that does nothing else other than passing the given params to the library (but maybe i'm reading the code wrong), why not use moment.js directly then?

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -1,33 +1,25 @@
...
+            'x'
+        ], true);
 
-            if(date < 0 || datetime.length < 10){
-                return 'Unspecified';
-            } else {
-                return $filter('date')(date, 'dd/MM/yyyy');
-            }
+        var beginningOfEra = DateFactory.createDate('01/01/1970');
+        var isHighEnough = !Date.isSame(beginningOfEra);
```
AkA84: What's the logic behind this? You force the date to have a positive unix timestamp?

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -1,33 +1,25 @@
...
-            if(datetime.getTime){
-                return $filter('date')(datetime.getTime(), 'dd/MM/yyyy');
-            }
-        } else if(typeof datetime === 'number'){
-            return $filter('date')(datetime, 'dd/MM/yyyy');
-        }
+        if(Date.isValid() && isHighEnough) return Date.format('DD/MM/YYYY');
 
-        return null;
+        return 'Unspecified';
```
AkA84: Why a specific string instead of `null`?

org.civicrm.reqangular/src/common/angular-date/src/controllers/DatePickerController.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * @extends DatepickerController
+ */
+function DatePickerController($scope, $controller, DateFactory, $log){
+    $scope.implements = 'DatepickerController';
```
AkA84: What's this for?

org.civicrm.reqangular/src/common/angular-date/src/controllers/DatePickerController.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * @extends DatepickerController
```
AkA84: Please provide a description about what's the purpose of the extension

org.civicrm.reqangular/src/common/angular-date/src/controllers/DatePickerController.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+function DatePickerController($scope, $controller, DateFactory, $log){
+    $scope.implements = 'DatepickerController';
+
+    var me = this,
+        ngModelCtrl = { $setViewValue: angular.noop };
+
+    angular.extend(this, $controller('DatepickerController', {
```
AkA84: since you are already referring `this` through `me`, use `me` here

org.civicrm.reqangular/src/common/angular-date/src/controllers/DatePickerController.js
```diff
@@ -0,0 +1,69 @@
...
+        $scope: $scope,
+        $attrs: {}
+    }));
+
+
+    $scope.isActive = function(dateObject) {
```
AkA84: *(wherever else applies)*
Please document the function/method
```js
/**
 * Description
 *
 * @param {type} paramName - description
 * @param {type} paramName - description
 * @return {type} description
 */
```

org.civicrm.reqangular/src/common/angular-date/src/controllers/DatePickerController.js
```diff
@@ -0,0 +1,69 @@
...
+        var newDate = DateFactory.moment(formatted, 'DD/MM/YYYY');
+        return newDate.toDate();
+    };
+
+    /* Overriding Methods */
+    me.render = function() {
+        if ( ngModelCtrl.$modelValue ) {
+
+            var date = me.parseDate(ngModelCtrl.$modelValue),
+                isValid = !isNaN(date);
```
AkA84: Need consistency on how new `var`s are declared + defined, either
```js
var foo = 'foo';
var bar = 'bar';
```
or
```js
var foo = 'foo', bar = 'bar';
```
personally i'm more inclined for the former (i use the latter only when i declare them), but anyway the syntax must be consistent

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerDirectiveDecorator.js
```diff
@@ -0,0 +1,24 @@
+function DatepickerDirectiveDecorator($delegate) {
```
AkA84: *(wherever else applies)*
Please provide a description about what this decorator does

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerDirectiveDecorator.js
```diff
@@ -0,0 +1,24 @@
...
+    var directive = $delegate[0];
+
+    var old_link = directive.link;
```
AkA84: Group `var` declarations/definitions together, makes for a cleaner code
```js
var directive = $delegate[0];
var old_link = directive.link;
```

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerDirectiveDecorator.js
```diff
@@ -0,0 +1,24 @@
...
+
+    delete directive.link;
```
AkA84: Is this necessary? I'm not really familiar with decorators to be honest, but i was reading this [article here](http://angular-tips.com/blog/2013/09/experiment-decorating-directives/) and it does seem there is a necessity to delete the link function

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerDirectiveDecorator.js
```diff
@@ -0,0 +1,24 @@
...
+    directive.compile = function(){
+        return function(scope, element, attrs, ctrls){
+
+            /**
+             * @override
+             * @type {number}
+             */
+            ctrls[0].startingDay = 1;
+
+            old_link(scope, element, attrs, ctrls);
```
AkA84:
```js
old_link.apply(this, arguments);
```

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerPopupWrapDirectiveDecorator.js
```diff
@@ -0,0 +1,9 @@
+function DatepickerPopupWrapDirectiveDecorator($delegate) {
+    var directive = $delegate[0];
+
+    directive.templateUrl = 'templates/datepickerPopup.html';
+
+    return $delegate;
```
AkA84: *(wherever else applies)*
Can't you just write
```js
function DatepickerPopupWrapDirectiveDecorator($delegate) {
    $delegate[0].templateUrl = 'templates/datepickerPopup.html';

   return $delegate;
}
```

AkA84:
About files and folder structure

1. The file name convention should be consistent, please use dasherized names: `DatepickerDirectiveDecorator.js -> datepicker-directive-decorator.js`, `CustomDateInput.js -> custom-date-input.js`, etc

2. All the files for `angular-date` should be placed in their correspondent folder under `src/common` to keep the structure consistent. If you want, you can "namespace" them with a `angular-date/` subfolder, so you could have things like:
  * `src/common/directives/angular-date/custom-date-input.js`
  * `src/common/decorators/angular-date/datepicker-popup-wrap-directive-decorator.js`
  * `src/common/modules/angular-date.js`
  * `src/common/modules/angular-date/templates-main.js`
  * etc.

3. Finally you could create a bundle under `src/common/bundles`, like `src/common/bundles/angular-date`, where you require everything related to `angular-date`, and you would include it in `src/build.js`
  ```js
  include: [
      'common/bundles/vendors',
      'common/bundles/apis',
      'common/bundles/angular-date'
  ]
  ```
  doing so makes it unnecessary to create a separate production-ready file for the module

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
kkalamarski: It's a filter to format date for displaying. We don't want to show null to customers, so it saves one condition + keeps consistency.
```javascript
{{ invalidDate | CustomDate }}
```
will display 'Unspecified'.

kkalamarski:
Changes in this PR address folowing:
https://compucorp.atlassian.net/browse/PCHR-607
https://compucorp.atlassian.net/browse/PCHR-555
https://compucorp.atlassian.net/browse/PCHR-505

And issues that came out in the meantime, like timezone problems.
Using momentJs rather than JS Date Object.

org.civicrm.reqangular/src/common/angular-date/src/decorators/DatepickerDirectiveDecorator.js
kkalamarski: fixed

org.civicrm.reqangular/src/common/angular-date/src/services/DateValidationService.js
kkalamarski:
```javascript
me._error = function (error_msg, fields) {
    throw new Error(error_msg);
};
```
'fields' param, although isn't used within default error function, provides information on which date field error was thrown. i.e:
```javascript
me._error('Date is invalid! Valid date format is: dd/mm/yyyy.', ['end_date']);
```
Developer can override this function, and use fields parameter. Basically this param is here, because error function takes 2 parameters always, even if there is no purpose in using one of them. (for example we are validating one datefield - we don't need information about field).

It gives freedom do provide custom error functions as done in HRJobRoles:
```javascript
 $scope.validateDates(data.start_date.$viewValue, data.end_date.$viewValue, function(error, field){
                    errors++;
                    if(field.indexOf('start_date') > -1) {
                        data.start_date.$error.custom.push(error);
                    }
                    if(field.indexOf('end_date') > -1){
                        data.end_date.$error.custom.push(error);
                    }
                });
```

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
kkalamarski: In AMD module, it would be, indeed, better to use moment directly.
In CommonJS it prevents including moment multiple times.

Moreover, this class is supposed to do more things in the future.

kkalamarski:
Addressed all issues pointed in this PR, except for these caused by choice of CommonJS module type.

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
kkalamarski: 0 timestamp appears when date is invalid or unset. To simplify we assume that all dates with 0 timestamp are invalid.

AkA84:
Test

AkA84:
Approved

PR #813, opened 2016-01-04
PCHR-614: Hidden leave entitlement for contracts other than current

Agares4:
The problem solved by this PR was that absence entitlement was shown for contracts that are not current. 
Before:
before
After:
after


Agares4:
Closing the PR, as the task has been cancelled.

PR #814, opened 2016-01-04
PCHR-621: Changed 'Current Year' to real absence period name

Agares4:
(Contact Summary, "Sickness" section)
Before:
screenshot 23

After:
screenshot 24


contactsummary/js/src/contact-summary/controllers/sickness.js
```diff
@@ -20,6 +20,7 @@ define([
         this.takenPreviously = 0;
         this.staffAverage = 0;
         this.ready = false;
+        this.currentPeriod = 'Unknown';
```
AkA84: Unless stated otherwise, please put a `-` (dash), so that we don't need to translate this string in the future

contactsummary/js/src/contact-summary/services/leave.js
```diff
@@ -261,6 +261,7 @@ define([
                     return period;
                 });
         }
+        factory.getCurrentPeriod = getCurrentPeriod;
```
AkA84: `getCurrentPeriod` is not a private method anymore since you are referencing it in the `factory` object, so just move the implementation directly in `factory.getCurrentPeriod` and reference it in the private methods

AkA84:
Approved

PR #816, opened 2016-01-04
PCHR-612: Fixed 'active roles or projects' count for contacts without…

Agares4:
… contract

The number of active roles or projects (under "Key dates section") was wrong for contacts without a job contract.


contactsummary/js/src/contact-summary/services/jobRole.js
```diff
@@ -80,6 +80,10 @@ define([
                         contractIds.push(contract.id);
                     });
 
+                    if(contractIds.length === 0) {
```
AkA84: Please keep code style consistent
```js
if (contractIds.length === 0) {
  //...
}

AkA84:
Minor comment, otherwise approved

PR #817, opened 2016-01-11
PCHR-764: Cacelled leave shows 0 now

nishant-bhorodia:
**Issue:**
When a leave is Rejected/Cancelled, the count is displayed as -1 on list tab under Absences tab of a contact in CiviCRM admin.
absence-tab-before

**Solution:**
Check if the status is "Cancelled" (numerical value for cancelled leave is 3) and return 0.
absence-tab-after


gregmeszaros:
Approved!

AkA84:
Merging on behalf of @nishant-bhorodia

PR #818, opened 2016-01-11
UY-29: Correct input date formats

nishant-bhorodia:
**Issue:**
Dates were not formatted right in date input fields on absence edit page in CiviCRM because the format was statically defined in php code and it was not respecting the format chosen by admin in CiviCRM localization settings.
uy-29-before

**Solution:**
Removed static date format and added a logic to change format in php specific date formats.
uy-29-after


gregmeszaros:
Approved!

AkA84:
Merging on behalf of @nishant-bhorodia

PR #820, opened 2016-01-12
Pchr 784 contracts import

Agares4:
Fixing issues with import of job contracts and roles:
1. Non-existants values for pseudocontant fields cause errors (now they're simply ignored, just like in Civi's importers)
2. Date/time format setting is ignored
3. Mandatory fields aren't correctly marked as mandatory


coldrunKacper:
PR Approved.

PR #822, opened 2016-01-13
PCHR-787: Dashboard (Layout fixes)

AkA84:
#### Problem
The main section of the `.crh_page` component doesn't have any style in place that prevents it getting stretched to cover the full available space on the user's screen, resulting on its content getting distorted

before

#### Solution
We're using Bootstrap grid system, but in order to achieve, on medium/large viewports, the layout shown [in the design](https://projects.invisionapp.com/share/M34QCER6H#/screens), we need to override/change things a bit:

* The `.chr_page__main` element must be centered and have a specific `max-width` (`1004px`)
* The first column must have a specific `max-width` (`183px`)
* The second column, on a custom breakpoint (see next point), must spread to the full width and drop the `float`, so that `.chr_page__main` can correctly be centered on the available space
* Unfortunately the layout broke in a small interval between Bootstrap's large breakpoint (`1200px`) and a seemingly arbitrary width (around `1240px`). I didn't see any specific math behind it that I could trace back to the widths defined above, it must be because of layout change with a fixed + fluid column. To address the issue then, a custom breakpoint `$crm-page-breakpoint-min` was added

The result:
after

Example from large to small viewport is below (ignore the inner elements breaking, they're not responsive yet). A `padding-left` was also added for small viewports in order to center `.chr_page__main`.

shrink-test

---

#### Problem
The dashboard looks completely broken in IE (sorry, no screenshot)

#### Solution
Turns out that IE, not wanting to disappoint, doesn't support the `initial` value which was used to revert some styles applied by the CiviCRM css. Using specific values which match the initial values solves the problem


org.civicrm.bootstrap/scss/overrides/reset/_civicrm.scss
```diff
@@ -3,20 +3,20 @@
     #{$bootstrap-namespace} {
 
         h3 {
-            background-color: initial;
```
Agares4: A comment here saying that `initial` is not supported by IE and these are the initial values would be helpful, I think.

org.civicrm.bootstrapcivicrm/scss/components/_page.scss
```diff
@@ -3,23 +3,45 @@
...
+                    max-width: $crm-page-sidebar-max-width + $grid-gutter-width;
+                }
             }
         }
 
+        // On medium/large viewports the main block must
+        // be centered and should not stretch after a certain point
+        .crm_page__main {
+            margin: 0 auto;
+            max-width: $crm-page-main-max-width;
```
Agares4: Why not use a media query here?

org.civicrm.bootstrap/scss/overrides/reset/_civicrm.scss
AkA84: Right, will do

org.civicrm.bootstrapcivicrm/scss/components/_page.scss
AkA84: Not sure what you mean, this rule is inside already in a media query
```scss
@media (min-width: $screen-md-min) {
```

Agares4: Yup, but if it could be in another media query, like: 

    @media (min-width: $crm-page-main-max-width) {
         width: $crm-page-main-max-width;
         margin: 0 auto;
     }

I think this way it would better show what you're intending to do (and therefore the comment wouldn't be needed).

AkA84: I think that media query is a bit incorrect :smile:

Even if applied, it would mean that those rules would be applied to a viewport only `180px` wide, which is obviously not what we want to do here.

Instead what we want to do is to make the first column wide max `180px` but only for medium to large viewports, which is what `@media (min-width: $screen-md-min) {` accomplishes

Agares4: Okay, so it appears that I misunderstood your code. Nevermind :)

Agares4:
Approved

PR #824, opened 2016-01-14
PCHR-262: Apply reports style to all reports pages

AkA84:
#### Problem
The sliders in the new *Leave and Absences* report page are not styled

before

#### Solution
It was because the reports page style were applied to an hardcoded list of `<body>` classes, so the new page wasn't included in the list. To fix the problem a more generic class selector was used

after


kkalamarski:
Is it intentional that slider in solution has so big padding applied?

AkA84:
Yes is to make sure that there is space between the chart and the slider:

example

Nevermind that in the solution screenshot there is no chart, I didn't bother setting up one

kkalamarski:
Sure, then Approved

PR #825, opened 2016-01-15
PCHR-621: Fixed absence period name

Agares4:
Problem: In "Contact Summary" the name of absence period was mangled (eg. "this_year" instead of "this year").
Solution: Used "title" property of the period, instead of "name" (which is machine friendly version of "title").


AkA84:
Approved

PR #828, opened 2016-01-22
PCHR-505: Start-End Dates Fix

kkalamarski:
# 1. EndDate
### Problem:
User wasn't able to save empty end date while creating new job role.
### Solution: 
Added a condition in validator that checks if end_date is null. If true it doesn't check for it's validity.

# 2. Parsing Output
### Problem:
Dates were saved either as a Date object or a 'dd/mm/yyyy' string. This inconsistency caused some dates to be saved incorrectly.
### Solution:
Method $scope.parseDate parses all dates into 'yyyy-mm-dd' format. These are read by the backend and saved to database correctly.

# 3. Directives
### Problem:
Two inputs were lacking directive which resets $parsers of ui-datepicker - it caused a chain of useless transformations of date format affecting performance. Also few directives are no longer needed (i.e show-weeks).
### Solution:
Add missing directives, and remove redundant ones.


com.civicrm.hrjobroles/js/controllers/HRJobRolesController.js
```diff
@@ -330,10 +328,36 @@ define(['controllers/controllers'], function(controllers){
...
+                || $scope.edit_data['new_role_id']['job_contract_id'] == '');
+            };
+
+            /**
+             * @description Parse dates so they can be correctly read by server.
+             * Use od DateFactory.CreateDate() takes care of timezone issue.
+             * @param date {string|timestamp|Date}
+             * @returns {Date}
+             */
+            $scope.parseDate = function (date) {
```
AkA84: Can't you do the calculations in this function via moment.js?

com.civicrm.hrjobroles/js/controllers/HRJobRolesController.js
```diff
@@ -330,10 +328,36 @@ define(['controllers/controllers'], function(controllers){
...
-                    || typeof $scope.edit_data['new_role_id']['job_contract_id'] === 'undefined'
-                    || $scope.edit_data['new_role_id']['job_contract_id'] == '');
+                || typeof $scope.edit_data['new_role_id']['title'] === 'undefined'
+                || $scope.edit_data['new_role_id']['title'] == ''
+                || typeof $scope.edit_data['new_role_id']['job_contract_id'] === 'undefined'
+                || $scope.edit_data['new_role_id']['job_contract_id'] == '');
+            };
+
+            /**
+             * @description Parse dates so they can be correctly read by server.
```
AkA84: You can omit the `@description` tag here, also make sure that you have a new line between the description and the rest of the tags
```js
/**
 * Lorem ipsum dolor sit amet
 *
 * @param {type} param1
 * @return {type}
 */

com.civicrm.hrjobroles/js/controllers/HRJobRolesController.js
kkalamarski: DateFactory uses moment.js to do that. 
All I wanted to do was to get offset before converting Date object to moment.js object.
In the process of convertion this information is somehow lost, so methods utcOffset(), zone() return 0 (or -0) instead of correct value. Not sure why is that, that's why i decided that getting offset before convertion is more secure.

kkalamarski:
Actually all our problems with timezones were solved by changing column data type and $scope.parseDate method.

AkA84:
Approved

PR #829, opened 2016-01-22
PCHR-796: Date Format

kkalamarski:
> This PR should not be merged until #850 is done and approved.

## Problem:
Date format should be loaded from (somewhere) in bootstrap phrase across all angular apps, and accessible by simple service.

### Create compuNgRoute and compu.ui.router modules extending ngRoute and ui.router respectively

Usage for compuNgRoute
```javascript
angular.module('testNgRoute', ['compuNgRoute'])
    .config(['$routeProvider', function ($routeProvider) {
        $routeProvider.resolveForAll({
            test: function () {
                return 5;
            }
        }).when('/', {
            template: 'Test',
            resolve: {
                a: 'b',
                c: 'd',
                e: function () {
                    return 'f';
                }
             }
          });    
    }]);

```

Usage for compu.ui.router
```javascript
angular.module('test.ui.router', ['compu.ui.router'])
    .config(['$stateProvider', function ($stateProvider) {
        $stateProvider.resolveForAll({
            test: function () {
                return 5;
            }
        }).state('main', {
            url: '/',
            resolve: {
                a: 'b',
                c: 'd',
                e: function () {
                    return 'f';
                }
             }
          });    
    }]);

```

Now controller is resolved with: 
```javascript
{
    a: 'b',
    c: 'd',
    e: 'f',
    test: '5'
}
````

#### DateFormatService
Simple service containing one field and one method.
The purpose of this service is to deliver date format from external source and cache it in itself.

Example:
```javascript
DateFormatService.getDateFormat().then(function(result){
    console.log(result); // date format {string}
});
```
Once finished we can also access date format synchronously:
```javascript
console.log(DateFormatService.dateFormat); // date format
console.log(HR_settings.DATE_FORMAT); // date format
```


org.civicrm.reqangular/src/common/angular-date/main.js
```diff
@@ -32,4 +32,8 @@ Module.config(function($provide) {
     $provide.decorator('daypickerDirective', require('./src/decorators/DaypickerDirectiveDecorator'));
 
     $provide.decorator('datepickerPopupWrapDirective', require('./src/decorators/DatepickerPopupWrapDirectiveDecorator'));
-});
\ No newline at end of file
+});
+
+Module.run(function(DateFactory){
+    DateFactory.fetchDateFormatFromSettings();
```
AkA84: Is `DateFactory` is the right place to put methods concerned with getting the current date format? Shouldn't the concern of a factory be only to create new object, rather than additionally provide information about the current state of an application?

If we have already this `angular-date` module which is a collection of controllers/services/directives related to the date, why not make a specific service whose job is only to retrieve the current date format?

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -1,5 +1,6 @@
 module.exports = function ($filter, DateFactory) {
-    return function (datetime) {
+
+    var filter = function (datetime) {
         var Date;
 
         if(typeof datetime == 'object'){
```
AkA84: Here you have
```js
if(typeof datetime == 'object'){
    datetime = $filter('date')(datetime, 'dd/MM/yyyy');
}
```
but shouldn't it be
```js
if(typeof datetime == 'object'){
    datetime = $filter('date')(datetime, DateFactory.getDateFormat());
}
```
?

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -18,8 +19,12 @@ module.exports = function ($filter, DateFactory) {
...
 
-        if(Date.isValid() && notEmpty) return Date.format('DD/MM/YYYY');
+        if(Date.isValid() && notEmpty) return Date.format(DateFactory.getDateFormat());
 
         return 'Unspecified';
     };
-};
\ No newline at end of file
+
+    filter.$stateful = true;
```
AkA84: Is it necessary to make the filter stateful? It impacts the app performance

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
```diff
@@ -20,13 +23,33 @@ function DateFactory() {
...
+         */
+        getDateFormat: function () {
+            return this.dateFormat;
+        },
+
+        /**
+         * @description Fetches date format setting from
+         * @returns {Promise}
+         */
+        fetchDateFormatFromSettings: function(){
```
AkA84: Why having two methods for the same thing (`getDateFormat` and `fetchDateFormatFromSettings`)? Can't you have the same method that checks if the format has already been stored in a property and, if not, then calls the service?
```js
getDateFormat: function () {
    if (!this.dateFormat) {
        // get the date format
    }
```

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
```diff
@@ -20,13 +23,33 @@ function DateFactory() {
...
+            var me = this;
```
AkA84: No need for this, you can use `.bind` on the function
```js
.then(function (result) {

}.bind(this));
```

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
```diff
@@ -20,13 +23,33 @@ function DateFactory() {
...
+            return $q.when('DD/MM/YYYY').then(function(result){
+                me.dateFormat = result;
+                return result;
```
AkA84: Why returning the result? It's not used anywhere

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactoryTest.js
```diff
@@ -11,9 +11,4 @@ describe('Unit: DateFactory', function() {
...
         expect(Factory).toBeDefined();
     });
-
-    it('Moment.js should be included', function() {
-        expect(Factory.moment).toBeDefined();
-    });
-
-});
\ No newline at end of file
+});
```
AkA84: We need tests for the new functionality

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
```diff
@@ -1,8 +1,12 @@
 var moment = require('../../../vendor/moment.min.js');
 
-function DateFactory() {
+function DateFactory($q) {
     return {
-        moment: moment,
+        /**
+         * Default Format
+         */
+        dateFormat: 'DD-MM-YYYY',
```
AkA84: I think dateFormat should be `null` at the beginning (see below), while a default/fallback value should be assigned to it only if the asynchronous call fails

AkA84:
Can you write an example of how this would be integrated on a real app?

AkA84:
There is an issue with this (obligatorily) async call to fetch the request data: the app bootstrap phase will not wait for it and you will end up with an app which is supposed to use a date format which is not fetched yet.

I googled a bit around for things like "async .run" and similar, and I ended up in this [issue](https://github.com/angular/angular.js/issues/4003), which contains several solutions to the problem, please take a look

kkalamarski:
.run() does not support the async calls. 
Therefore I decided to use stateful filter to format dates asynchronously (most of cases just to display date in template).
If needed you can always use async function to do get format in more complicated functions.

I checked how use of stateful filters influences application's performance - it's almost unnoticeable (tried up to 10), I know it's not the best idea though.

But still I think that's better than some dirty hacks.

I read this before I even started working on that. Actually this solution isn't that bad:
> The way Jasmine does this is by setting a flag, that will let the function run and wait for the callback until the flag is set to true, or until a timeout is fired.
>
>Basically something like this:
>
```javascript
var flag = false;
$http(...).success(flag = true);
$timeout(function () {
    flag = true;
    // Or add some error handling stuff here
}, 5000);
while (!flag) {};
```
> It's not pretty, but should do the job. The code above isn't tested, so you may need to modify it a bit. It's getting late here. Hopefully something to get you by until something a bit prettier comes along.

But still it's fighting against Angular.
Let me know what you think of this, and In the meanwhile I'll fix other issues.

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
kkalamarski: Actually this method should be synchronous. That's why I decided to delegate async call to another method. Also, I'd need to return string or a promise object depending on this condition - that wouldn't be good idea. 
It's better that function return one exact type.

Also these functions do completely different things - one fetches data from server, another delivers it to the user.

kkalamarski:
### Usage in real app

Set angular-date as a dependency in any module.
Use either CustomDateFilter or get format from DateFactory and pass it to DateFilter to format date.

that's it.

org.civicrm.reqangular/src/common/angular-date/src/services/DateFactory.js
AkA84: > Also, I'd need to return string or a promise object depending on this condition 

That's not necessary: the method will still be async no matter if it actually needs to fetch data or not, because the data will eventually be wrapped in a Promise, which would make the method's behaviour and consumption consistent (either it waits for the data from the server to arrive, or just the next tick to immediately return the data already available).

This way the consumer of the method would not have to worry about its actual implementation, she would just need to know that it returns a Promise:
```js
DateFactory.getDateFormat().then(function (format) {
    format; // DD/MM/YYYY
});
```
The method's implementation would be something like:
```js
getDateFormat: function () {
    if (this.dateFormat) {
        // wraps the data in a Promise and resolves it
        return $q.when(this.dateFormat);
    } else {
        // makes an actual async request
        return $http({ /* fetchData */ }).then(function (format) {
            this.dateFormat = format;
            return this.dateFormat;
        });
    }
}
```

AkA84:
> Set angular-date as a dependency in any module.
Use either CustomDateFilter or get format from DateFactory and pass it to DateFilter to format date.

Sorry I meant: can you write code/pseudocode based on any of the current app to see how this would be used?

AkA84:
So what you are doing here is having a filter directly using a service, whose job consists of returning an async response
```js
// CustomDateFilter
if(Date.isValid() && notEmpty) return Date.format(DateFactory.getDateFormat());
```
but I'm wondering if this is the correct way to go, because now you're tightly coupling a filter to a specific service (async, even), which is something that also forces you to make the filter stateful. And also this approach means that every app component needs to inject this specific service in case the date format is needed (thus increasing the instances of tight coupling)

The initial intention was to fetch the date format on the app run phase and store it in a way that any part of the app (service, filter, controller) could retrieve it and use it if needed (this is pseudo code which ignores the issue with async in run, it's just to illustrate the idea):
```js
// main app file
app.run(['settings', 'fetchDate', function (settings, fetchDateFormat) {
    fetchDateFormat().then(function (format) {
        settings.constant('DATE_FORMAT', format);
    });
}]);

// a controller
app.module('app.controllers').controller('MyController', ['settings', function (settings) {
    this.manipulateSomeDate(date) {
        moment(date, settings.DATE_FORMAT);
    }
}]);

// a filter
app.module('app.filters').filter('MyFilter', ['settings', function (settings) {
    return function (value) {
        // ...
        if (Date.isValid() && notEmpty) {
           return Date.format(settings.DATE_FORMAT);
        }
    }
});
``` 
This way we wouldn't have to have filters/directives/etc use the service directly, as they shouldn't really care where the date format comes from, they would only need to know that it's available as a app-wide setting. And of course the value would be already available, without the need to do async fetch operations.

Of course the main issue here is that the app must get bootstraped **after** the date format value is fetched and, given that in the near future the value will have to come from the API, the question are: can we do that? If yes, does it works as expected or creates unexpected issues?

kkalamarski:
Actually this makes sense. I can use while() to wait until promise is resolved, although it's not an elegant solution.

AkA84:
Feel free to experiment with different solutions (we have after all different apps which bootstrap in different ways), better to spend a little more time now to get it right the first time

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
kkalamarski: Yes, since we get date format from outside of the filter it has to be stateful.

kkalamarski:
I was not able to find a way to make bootstraping wait for async call in .run(), but I think resolve does the job.

## Usage

### module
```javascript
app.config(function($routeProvider){
    $routeProvider.
        when('/', {
            templateUrl: '[path_to_template]',
            resolve: {
                dateFormat: function(DateFormatFactory){
                    return DateFormatFactory.getDateFormat();
                }
            }
        });
    }
);
```

### controller
```javascript
app.controller('MyController', ['dateFormat', function (dateFormat) {
    this.formatDate(date) {
        moment(date, dateFormat);
    }
}]);
```

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -1,8 +1,10 @@
-module.exports = function ($filter, DateFactory) {
-    return function (datetime) {
+module.exports = function ($filter, DateFactory, DateFormatFactory) {
+    var dateFormat = null;
+    var filter = function (datetime, format) {
```
AkA84: So how did the usage of this filter change? You can now pass the format to the filter itself?

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
```diff
@@ -18,8 +20,18 @@ module.exports = function ($filter, DateFactory) {
         var beginningOfEra = DateFactory.createDate(0);
         var notEmpty = !Date.isSame(beginningOfEra);
 
-        if(Date.isValid() && notEmpty) return Date.format('DD/MM/YYYY');
+        if (!dateFormat) {
+            DateFormatFactory.getDateFormat().then(function (result) {
```
AkA84: Again, you're coupling the filter to the service, which in turns forces you to make the filter stateful because the service is async. What has changed since the last discussion about what approach to take? Why not store the format in a constant and inject it into the filter?

org.civicrm.reqangular/src/common/angular-date/src/services/DateFormatFactory.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * This factory provides access to date format from CRM settings
+ * @param $q
+ * @returns {{dateFormat: null, getDateFormat: getDateFormat}}
+ * @constructor
+ */
+function DateFormatFactory($q) {
```
AkA84: Why is this service called a `Factory`? This doesn't seem to match the definition of a factory, it just returns a string

org.civicrm.reqangular/src/common/angular-date/src/services/DateFormatFactory.js
```diff
@@ -0,0 +1,36 @@
...
+            } else {
+                return $q.when('DD/MM/YYYY')
+                    .then(function (result) {
+                        this.dateFormat = result;
+                        return result;
+                    }.bind(this))
+                    .catch(function (error) {
+                        // Fallback to default format
+                        this.dateFormat = 'YYYY-MM-DD';
+                        return error;
```
AkA84: How do you want to handle errors? You are still assigning a default fallback to the `dateFormat` property, but instead of returning the default value, you are returning the error. But doesn't it go against the value of providing a default value?

org.civicrm.reqangular/src/common/angular-date/src/services/DateFormatFactory.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param $q
+ * @returns {{dateFormat: null, getDateFormat: getDateFormat}}
+ * @constructor
+ */
+function DateFormatFactory($q) {
+    return {
+        /**
+         * keeps information about date format
+         */
+        dateFormat: null,
```
AkA84: Maybe you can put the default date format directly here. If the async call will succeed, then the value of `this.dateFormat` will get replaced, otherwise it will be already defined here instead of getting assigned in the `catch` block later on

org.civicrm.reqangular/src/common/angular-date/src/services/DateFormatFactory.js
```diff
@@ -0,0 +1,36 @@
...
+        /**
+         * @description Fetches date format setting from TODO where?
```
AkA84: the `@description` tag is not necessary if the description is the first thing in the comment section

org.civicrm.reqangular/src/common/angular-date/src/services/DateFormatFactory.js
kkalamarski: > I think dateFormat should be null at the beginning (see below), while a default/fallback value should be assigned to it only if the asynchronous call fails

That's what you said when I put it directly last time :D

AkA84: :facepalm:

right and I was right back then cause otherwise you cannot check if you have already cached the value. Apologies

org.civicrm.reqangular/src/common/angular-date/src/filters/CustomDateFilter.js
kkalamarski: How can I tell if the constant has already been initialised and has correct value, since there is no way of stoping the bootstrap of application until it's done?

org.civicrm.reqangular/gulpfile.js
```diff
@@ -23,8 +32,88 @@ gulp.task('requirejs-bundle', function (done) {
...
+         * Looks for a test file (*_test.js) in `test/`, using the same path
+         * of the source file in `src/appraisals/`
+         *   i.e. src/appraisals/models/model.js -> test/models/model_test.js
+         *
+         * @throw {Error}
+         */
+        for: function (srcFile) {
+            var srcFileNoExt = path.basename(srcFile, path.extname(srcFile));
+            var testFile = srcFile
+                .replace('src/appraisals/', 'test/')
```
AkA84: Referring Appraisals, both here and in the function description

org.civicrm.reqangular/src/common/bundles/angularDate.js
```diff
@@ -0,0 +1,5 @@
+define([
+    'common/services/angularDate/date-format'
```
AkA84: Will this bundle be filled in the future with other components of angular-date?

AkA84: Also, for file names (and therefore module names): please use dasherized style (`angular-date`)

org.civicrm.reqangular/src/common/services/settings/hr-settings.js
```diff
@@ -0,0 +1,12 @@
+define([
+    'common/angular',
+    'common/modules/settings'
+], function (angular, module) {
+    'use strict';
+
+    module.factory('HR_settings', function(){
```
AkA84: Aren't services in angular singletons? Doesn't it mean that in our case all the apps are going to share the same `HR_settings` object?

org.civicrm.reqangular/src/common/bundles/settings.js
```diff
@@ -0,0 +1,5 @@
+define([
+    'common/services/settings/hr-settings'
```
AkA84: Why a bundle? You expect to have different kind of settings?

org.civicrm.reqangular/src/common/bundles/routes.js
```diff
@@ -0,0 +1,5 @@
+define([
+    'common/services/compuRoute/route-decorator'
```
AkA84: Why a bundle? (I also don't understand the naming here)

org.civicrm.reqangular/src/common/modules/angularDate.js
```diff
@@ -0,0 +1,8 @@
+define([
+    'common/angular',
+    'common/modules/settings'
```
AkA84: Why coupling this module to the settings? Only the actual app should be concerned with the settings, what's the benefit of creating a direct dependency here?

org.civicrm.reqangular/src/common/modules/compuRoute.js
```diff
@@ -0,0 +1,7 @@
+define([
+    'common/angular'
+], function (angular) {
+    'use strict';
+
+    return angular.module('common.compuRoute', []);
```
AkA84: Unless you see the possibility of future components all linked together by a "compuRoute" context (like we have with `common.apis` and `common.dialog`), I think we can simply use `common.services` as the module for the decorator

org.civicrm.reqangular/src/common/modules/settings.js
```diff
@@ -0,0 +1,7 @@
+define([
+    'common/angular'
+], function (angular) {
+    'use strict';
+
+    return angular.module('common.settings', []);
```
AkA84: (Same thing for the bundle) Why a specific module? Do you see the possibility of a collection of various components all related to a `settings` theme?

org.civicrm.reqangular/src/common/services/compuRoute/route-decorator.js
```diff
@@ -0,0 +1,56 @@
...
+    'use strict';
+
+    module.factory('RouteDecorator', function () {
+        return {
+            /**
+             * Decorator function that provides new functionality either to $stateProvider or $routeProvider
+             * enabling them to declare common resolve across all states/routes using .resolveForAll() method.
+             * @param provider
+             */
+            enableCommonResolve: function (provider) {
```
AkA84: I like the idea of augmenting both `ngRoute` and `ui.router`, but wouldn't it be better to have, instead of a decorator that needs to handle both and that you need to explicitly use, two separate modules that overrides them?

Right now you need to do this:
```js
define([
    // ...
    'common/services/compuRoute/route-decorator'
], function (angular) {
    angular.module('myApp', [
        'ngRoute', // module to augment
        // ...
        'common.compuRoute' // additional module to reference
    ]).config([ /* ... */, function ($routeProvider, decorator,  /* ... */) {
        decorator.enableCommonResolve($routeProvider);  // explicit opt-in
        
        $routeProvider.resolveForAll({ /* ... */ }).when({ /* ... */ })
    }]);
});
```

But if you were to create two modules, say `compuNgRoute` and `compu.ui.route` which would extend respectively the two modules, everything would happen behind the scenes:

```js
define([
    // ...
    'common/module/compu-ng-route'
], function (angular) {
    angular.module('myApp', [
        'compuNgRoute', // takes care of requiring ngRoute, as its a dependency
    ]).config([ /* ... */, function ($routeProvider, decorator,  /* ... */) {
        // resolveForAll is already available on the provider        
        $routeProvider.resolveForAll({ /* ... */ }).when({ /* ... */ })
    }]);
});
```

org.civicrm.reqangular/src/common/filters/angularDate/format-date.js
```diff
@@ -0,0 +1,34 @@
+define([
+    'common/moment',
+    'common/modules/angularDate',
+    'common/services/settings/hr-settings'
+], function (moment, module) {
+    'use strict';
+
+    module.filter('formatDate', ['$filter', 'HR_settings', function ($filter, HR_settings) {
+        return function (datetime, format) {
+            var Date;
```
AkA84: Aren't you overwriting the default `Date` object here?

org.civicrm.reqangular/src/common/services/angularDate/date-format.js
```diff
@@ -0,0 +1,38 @@
+define([
+    'common/angular',
+    'common/modules/angularDate'
+], function (angular, module) {
+    'use strict';
+
+    /**
+     * This service provides access to date format from CRM settings
+     */
+    module.factory('DateFormatService', ['$q', function ($q) {
```
AkA84: I think the naming convention for services is not to put `Service` as a prefix for the name

org.civicrm.reqangular/src/tests/filters/angularDate/format-date_test.js
```diff
@@ -0,0 +1,91 @@
...
+    'common/angular',
+    'common/moment',
+    'common/angularMocks',
+    'common/filters/angularDate/format-date',
+    'common/services/angularDate/date-format',
+    'common/services/settings/hr-settings'
+], function (angular, moment) {
+    'use strict';
+
+    describe('angularDate: FormatDateFilter Unit Test', function () {
```
AkA84: Just put the name of what you are testing, it's already obvious this is a Unit Test
```js
describe('FormatDateFilter', function () {

});

**Comment by kkalamarski** [comment]
Actually I'm wondering if use of HR_settings is necessary, since we have date format saved in DateFormat.dateFormat anyway.

File: org.civicrm.reqangular/src/common/services/hr-settings.js [hunk 27]
```diff
@@ -0,0 +1,12 @@
+define([
+    'common/angular',
+    'common/modules/services'
+], function (angular, module) {
+    'use strict';
+
+    module.factory('HR_settings', function(){
```
AkA84: `module.constant` not `module.factory`
```js
module.constant('CIVIHR_SETTINGS', function () {
});
```

org.civicrm.reqangular/src/common/filters/angular-date/format-date.js
```diff
@@ -0,0 +1,34 @@
+define([
+    'common/moment',
+    'common/modules/angular-date',
+    'common/services/hr-settings'
+], function (moment, module) {
+    'use strict';
+
+    module.filter('formatDate', ['$filter', 'HR_settings', function ($filter, HR_settings) {
+        return function (datetime, format) {
+            var Date;
```
AkA84: I think you're overriding the built-in `Date` object here

org.civicrm.reqangular/src/common/modules/routers/compu-ui-router.js
```diff
@@ -0,0 +1,28 @@
...
+         * Share resolve across all states
+         * @param {object} _commonResolve
+         * @returns Provider
+         */
+        $stateProvider.resolveForAll = function (_commonResolve) {
+            $stateProvider.commonResolve = _commonResolve;
+            return $stateProvider;
+        };
+
+        // check if it's a ngRoute
```
AkA84: old comment

org.civicrm.reqangular/src/common/modules/routers/compu-ng-route.js
```diff
@@ -0,0 +1,28 @@
...
+         * Share resolve across all states
+         * @param {object} _commonResolve
+         * @returns Provider
+         */
+        $routeProvider.resolveForAll = function (_commonResolve) {
+            $routeProvider.commonResolve = _commonResolve;
+            return $routeProvider;
+        };
+
+        // check if it's a ngRoute
```
AkA84: old comment

org.civicrm.reqangular/src/common/services/hr-settings.js
kkalamarski: This cannot be a constant since we're assigning it's value later.

AkA84: Sorry, I remembered you could alter constants, but apparently the overriding value is applied only in the component where the override happens

AkA84:
Approved

PR #830, opened 2016-01-22
PCHR-800 - Matching job contract and job role at import 

Agares4:
Job role couldn't be imported before.
1. Changed type of job role's `start_date` and `end_date` from `TIMESTAMP` to `DATETIME`
2. Import now uses HrJobRole instead of the (not used anymore) HrJobContractRole
3. All fields appear on the list of fields to import (before, if there were fields with the same name in different entities, only one of them would appear)


coldrunKacper:
Approved.

PR #832, opened 2016-01-26
PCHR-791: Fixed length of service for end_date in the future

Agares4:
Fixed length of service in case of end_date of primary contract that is in the future.
Also fixed display issues (now the duration is rounded properly).


contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -42,11 +42,38 @@ define([
     // Private Members //
     /////////////////////
 
+    function formatDuration(duration) {
```
AkA84: Comment the function

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -42,11 +42,38 @@ define([
...
+        var texts = [],
+            days = duration.days(),
+            months = duration.months(),
+            years = duration.years();
+
+        if (years > 0) {
+            texts.push(years + " years");
```
AkA84: Aren't these `years`, `months`, `days` more appropriate for the view layer? Shouldn't you use a `ng-if` directive on the template?

Agares4: You mean returning: {years: 12, months: 34, days: 56} from here, and then using ng-if in the view?

AkA84: Yes, it makes more sense to put the labels in the view

Agares4: Ok, will do

AkA84:
Approved

PR #833, opened 2016-01-27
PCHR-805: Fixed import of health and insurance providers

Agares4:
Fixed import of insurance providers. They can now be referred to in import file either by ID or by display name.


hrjobcontract/js/controllers/contract.js
```diff
@@ -105,9 +105,17 @@ define(['controllers/controllers',
                 hour: ContractHourService.getOne({ jobcontract_id: contractId}),
                 pay: ContractPayService.getOne({ jobcontract_id: contractId}),
                 leave: ContractLeaveService.get({ jobcontract_id: contractId}),
-                health: ContractHealthService.getOne({ jobcontract_id: contractId}),
                 pension: ContractPensionService.getOne({ jobcontract_id: contractId})
-            }).then(function(results){
+            })
+            .then(function(results){
+                return ContractHealthService.getOne({ jobcontract_revision_id: results.details.jobcontract_revision_id })
+                    .then(function(health) {
```
AkA84: I think it could be rewritten for better clearity as:
```js
.then(function (results) {
    return ContractHealthService.getOne(//...);
})
.then(function (health) {
    results.health = health;

    return results;
)}
.then(function (results) {

});

```

Agares4: Results would be out of scope in the second "then" then

AkA84: Right, my bad. So just mind the code style
```js
.then(function (results) {
    return ContracHealthService.getOne({
        jobcontract_revision_id: results.details.jobcontract_revision_id
    })
    .then(function (health) {
    });
})
.then(function (results) {
});
```

coldrunKacper:
Approved

PR #834, opened 2016-01-27
PCHR-805: Fixed normalisation of Life Insurance Providers

Agares4:
A fix for a bug introduced in my fix for PCHR-805. Life Insurance Provider ID was not normalised correctly.


coldrunKacper:
Approved

PR #837, opened 2016-02-02
PCHR-823: differentiated fields from different entities in the UI

Agares4:
Added headers in the selects in "MapField" step of job contract import.


coldrunKacper:
Approved.

PR #839, opened 2016-02-03
PCHR-517: Fixed handling of required fields

Agares4:
Now the required fields are checked correctly, and if they're missing an error is reported.


coldrunKacper:
Approved.

PR #840, opened 2016-02-03
PCHR-817: cancelled leaves on calendar

nishant-bhorodia:
This PR removes the "cancelled" leaves from Calendars.


coldrunKacper:
Approved.

PR #841, opened 2016-02-03
PCHR-838: Fix Job Contracts tab is empty on Firefox

AkA84:
JS `strict mode` doesn't allow function declarations inside, for example, `if`s blocks. This PR move function declarations so they become strict mode compliant


kkalamarski:
Approved

PR #844, opened 2016-02-04
PCHR-839: Setup test environment (Appraisals)

AkA84:
The test environments are being set up in the other angular apps, this PR standardizes the test environment of the Appraisals app, so that it matches how things are done in the other apps


kkalamarski:
Approve

PR #845, opened 2016-02-04
PCHR-839: Setup test environment (Job Roles)

AkA84:
This PR sets up a test environment for the Job Roles app

#### Things done
* The dependencies list and the bootstrap code were in `job-roles.js` instead of `app.js`. The role of `job-roles.js` should be just to require the app and start its initialization, while `app.js` should contain everything related to the app (as the name suggests). This setup was also preventing the tests to run, given that the entry point for them is `app.js` and not `job-roles.js`, and as result they couldn't find all the  dependencies
* Added `test-main.js`, the entry file for the tests (bootstraps the whole test suite via requireJS)
* Added a `global.js` files, which is used to define the global variables (like `CRM`) that the tests should have access to
* Added an example test


kkalamarski:
Approved

PR #847, opened 2016-02-05
PCHR-839: Setup test environment (Contact Summary)

AkA84:
This PR fixes the test environment for the Contact Summary, which wasn't really working up to this point

#### Things done
* Added the npm modules used in all the other angular apps
* Added the gulp file used in all the other angular apps
* Standardized `karma.conf.js` and `test-main.js` based on how the test environment has been set up in the other angular apps
* Added a global.js files, which is used to define the global variables (like CRM) that the tests should have access to
* Moved the mocked components under `test/mocks`
* Moved the app initialization in `contact-summary.js`


kkalamarski:
Approved

PR #848, opened 2016-02-05
PCHR-840: Upgrade to AngularJS v1.4.9

AkA84:
This PR upgrades AngularJS and its modules to version `1.4.9` (from `1.2.27`). The older angular+modules have been moved to `vendor/angular/1.2.27`

No fixes were required except for the modal form of the Job Contract app, where the newer version complained about directives sharing the same DOM element


kkalamarski:
Approved

PR #849, opened 2016-02-05
PCHR-804: Fixed PHP warnings in hrjobcontract import

Agares4:
Fixed some warnings occurring on import of a hrjobcontract. 
Some of them were caused by invalid handling of date fields (too much fields were considered to be date/time, and therefore they were converted to the start of Unix epoch).
Others were caused by invalid array lookups.


coldrunKacper:
Approved.

PR #851, opened 2016-02-08
PCHR-567: Appraisal Cycle - Summary (static)

AkA84:
This PR adds a static version of the Individual Cycle page, with static versions of its modals

cycle

##### Add Contacts modal
add-contacts-modal

##### Access Settings modal
access-settings-modal

##### Edit dates modal
edit-dates-modal

#### `.nav-tabs` in `.panel-primary`
Placing a `.nav-tabs` element in a `.panel-primary`'s `.panel-heading` element will result in a custom-styled tabs element

nav-tabs

#### `.panel-body-small`
Simple variation of `.panel-body` which has less padding than the normal version (basically the same padding as `.panel-heading`)

panel-body-small

#### `.panel-heading-w-buttons`
A variation of `.panel-heading`, which takes care of vertically aligning buttons placed within it

panel-heading-w-buttons

#### Misc
* Resolved a float/clear issue in `.crm-page`
* Added reset styles for tables (they were getting some custom style from the CiviCRM css, which was interfering with Bootstrap's style)
* Customized the color of `.text-danger` so that it matches the `$brand-danger` color
* Removed the forced color on the `<strong>` elements


kkalamarski:
Approved

PR #855, opened 2016-02-11
PCHR-806: Properly handle primary contracts on import

Agares4:
Marks contract that is current as primary


coldrunKacper:
I think 'setPrimary()' method's name may be misleading a bit (as it always sets the 'primary' value to '0').

coldrunKacper:
Approved.

PR #857, opened 2016-02-12
PCHR-816: Replace location values with Location Labels in search results

nishant-bhorodia:
Replace location values with Location Labels in search results


coldrunKacper:
Approved.

PR #858, opened 2016-02-12
PCHR-578: Appraisal Cycle - Summary (AngularJS)

AkA84:
This PR contains several improvements and new components for the `org.civicrm.reqangular` (common AngularJS components) and the `uk.co.compucorp.civicrm.appraisals` (Appraisals app) extensions

#### Things done
* The data displayed on the Appraisal Cycle (Summary tab) page now comes from the model -> api -> db flow (except for 1) grades - those are still just static until we build the functionality 2) overdue appraisals, those are mocked as the api is not connected to the backend yet)
* Implemented the Edit Dates modal
  edit-date-modal
* Implemented the on-the-fly change of the cycle type and active settings, with confirmation dialog which, if rejected, rolls back the changes
  edit-data

#### Things not done
This PR is already massive enough, so the remaining things left to do will be done in a different PR
* Update overdue appraisals table to the latest design
* Add contacts modal
* Access settings modal
* Fetch date format via the `DateFormat` service
* Tests for the `api`s services
* Various refactoring and improvements

#### Fixed folder structure
The folder structure wasn't matching the structure of the other angular apps, specifically concerning where the tests and the configuration files were located.

##### Before
```
.
└── org.civicrm.reqangular/
    ├── dist/
    ├── src/
    │   ├── common/
    │   │   ├── angular-date/
    │   │   ├── ...
    │   │   └── vendor/
    │   └── tests/
    │       ├── directives/
    │       ├── ...
    │       ├── karma.conf.js
    │       └── test-main.js
    ├── build.js
    └── gulpfile.js
```
The `tests/` folder was placed inside `src/`, and `karma.conf.js` was inside `test/` instead of being right in the root folder.

##### After
```
.
└── org.civicrm.reqangular/
    ├── dist/
    ├── src/
    │   └── common/
    │       ├── angular-date/
    │       ├── ...
    │       └── vendor/
    ├── test/
    │   ├── directives/
    │   ├── mocks/
    │   ├── ...
    │   └── test-main.js
    ├── build.js
    ├── build.mocks.js
    ├── karma.conf.js
    └── gulpfile.js
```
`test/` (renamed from `tests/`) is now directly under the root folder at the same level of `src/`, and `karma.conf.js` is now placed where the other configuration files (`build.js`, `gulpfile.js`) are.

#### Fixed automated test run based on file change
`gulp` was supposed to be set up in a way that whenever a component file or its relative test file were changed, its tests were run automatically. This wasn't happening due to wrong configuration settings in `gulpfile.js`.

The issue is now fixed and the automated test run now works just like for the other angular apps

#### Set up support for mocked common components
For testing purposes we need to be able to define mocked versions of the common components, bundle them in a file and having that included by the karma server so that they can be injected in the tests.

To achieve this, a new folder `test/mocks/` has been added alongside a separate build file (`build.mocks.js`), which outputs a `dist/reqangular.mocks.min.js` file.

##### `build.mocks.js`
The `build.mocks.js` file is pretty simple:
```js
({
    baseUrl : 'test',
    out: 'dist/reqangular.mocks.min.js',
    uglify: {
        no_mangle: true,
        max_line_length: 1000
    },
    paths: {
        'common': 'empty:',
        'common/mocks': 'mocks/'
    },
    include: [
        'common/mocks/services/api/appraisal-mock',
        'common/mocks/services/api/appraisal-cycle-mock'
    ]
})
```
It defines the `common/mocks` alias namespace (without the alias the namespace would simply be `mocks/`) and then it bundles all the mocks together

#### Extended `api` service
The `api` service has now a `getAll()` method that the specific api services (`api.appraisal`, `api.appraisal-cycle`, etc) can delegate to, given that all apis should return the same response in terms of structure:
```js
.then(function (results) {
    return {
        list: results[0],
        total: results[1].count,
        allIds: results[1].values.map(function (cycle) {
            return cycle.id;
        }).join(',')
    };
});
```

#### Added mocked versions of `api`s
We have a this point two api services: `api.appraisal-cycle`, `api.appraisal`. For both of them there are two equivalent mocked versions: `api.appraisal-cycle.mock`, `api.appraisal.mock`.

The mocked versions basically contains some fake data (that is exposed through a public method) and simulate the logic of the backend code when asking for a collection (`all()`) or for a specific item (`find()`).

#### Added `Model` 
This is the "abstract" model that the specific models (`Appraisal`, `AppraisalCycle`, etc) will delegate to, and what it does is to provide common methods that should be available in all models.
```js
models.factory('MyModel', ['Model', function (Model) {
    return Model.extend({
        foo: function foo () {},
        bar: function bar() {}        
    });
}]);
```
Currently it provide one common method, `processFilters()` which processes query filters so that the backend endpoints can use them
```js
// MyModel.js
all: function (filters, pagination) {
    return myModelAPI.all(this.processFilters(filters));
},
```

#### Added `ModelInstance` 
This is the "abstract" model instance that the specific instances (`AppraisalInstance`, `AppraisalCycleInstance`, etc) will delegate to, and what it does is to provide common methods that should be available in all model instances
```js
models.factory('MyModelInstance', ['ModelInstance', function (ModelInstance) {
    return ModelInstance.extend({
        foo: function foo () {},
        bar: function bar() {}        
    });
}]);
```
Every instance can perform transformations on its attributes when it fetches data from the API, or when it's about to send it back to it. The transformations are defined in two methods: `fromAPIFilter`, `toAPIFilter`.
```js
// ModelInstance.js
fromAPI: function (attributes) {
    return _.transform(
        attributes,
        this.fromAPIFilter.bind(attributes), // transformations
        Object.create(null)
    );
},

toAPI: function () {
    var attributes = this.attributes();

    return _.transform(
        attributes,
        this.toAPIFilter.bind(attributes), // transformations
        Object.create(null)
    );
},
```
These methods are implemented in `ModelInstance` already but they just return the attributes as they are
```js
fromAPIFilter: function (result, __, key) {
    result[key] = this[key];
},
    result[key] = this[key];
}
```
and thus each specific instance can override them to perform specific transformations
```js
// AppraisalInstance.js
fromAPIFilter: function (result, __, key) {
    if (_.endsWith(key, '_date') || _.endsWith(key, '_due')) {
        result[key] = moment(this[key], 'YYYY-MM-DD').format('DD/MM/YYYY');
    } else if (_.includes(boolFields, key)) {
        result[key] = !!(+this[key]);
    } else {
        result[key] = this[key];
    }
},
```

#### Added a `appraisals` parent, abstract state
All the application states are now children of the `appraisals` state (`appraisals.dashboard`, `appraisals.appraisal-cycle`, etc). By doing this the `statuses` and `types` list are accessible in every status' controller

#### Added `Appraisal` model and `AppraisalInstance` model instance
`Appraisal` model offers the standard `all` and `find` methods, plus a `overdue` method that returns the overdue appraisals.
`AppraisalInstance` doesn't have custom methods for now, it just defined the `fromAPI` and `toAPI` transformations.

#### Augmented `AppraisalCycleInstance`
Every cycle instance can now get the appraisals of the cycle it represents via the new `loadAppraisals()` method
```js
var cycle = AppraisalCycleInstance.init(attributes);

cycle.loadAppraisals(function () {
    // the appraisals are now loaded in `cycle.appraisals`
});
```

#### Temporarily included mocked apis in the app
At this point of development, the `api.appraisal` service is not yet (fully) plugged in to the real backend api (this will be done in a future task). In order to simulate the data coming from the backend `api.appraisal.mock` (contained in `reqangular.mocks.min.js`) is temporarily used in its place
```php
// uk.co.compucorp.civicrm.appraisals/appraisals.php
CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.reqangular', 'dist/reqangular.mocks.min.js', 1010);
```


kkalamarski:
Approved

PR #859, opened 2016-02-16
PCHR-567: Fixes

AkA84:
This PR contains several fixes for the static version of the "Cycle Summary" page

##### Before
button-before
##### After
button-after

##### Before
table-border-before
##### After
table-border-after

##### Before
flowchart-before
##### After
flowchart-after

#### `.crm_custom-select`
This is an improved version of `.chr_custom-select` introduced first in PR #748. The markup is slightly more complex (the arrow had to but put in a separate element instead of being an `:after` pseudo-selector), but it allows full support for states like `disabled` or for Bootstrap's feedback classes

custom-select

```html
<div class="crm_custom-select crm_custom-select--full">
    <select id="cycle-type" class="form-control">
        <!-- ... -->
    </select>
    <span class="crm_custom-select__arrow"></span>
</div>
```

custom-select-disabled
```html
<div class="crm_custom-select crm_custom-select--full">
    <select disabled id="cycle-type" class="form-control">
        <!-- ... -->
    </select>
    <span class="crm_custom-select__arrow"></span>
</div>
```

custom-select-feedback-error
```html
<div class="form-group has-feedback has-error">
    <div class="crm_custom-select crm_custom-select--full">
        <select id="cycle-type" class="form-control">
            <!-- ... -->
        </select>
        <span class="crm_custom-select__arrow"></span>
    </div>
    <span class="glyphicon glyphicon-remove form-control-feedback"></span>
</div>
```


kkalamarski:
Approved

PR #862, opened 2016-02-18
PCHR-505 Job Roles dates bugs

kkalamarski:
# Bugfixes
## Bug:
**While creating a JobRole using Contract with only start date, end date gets set to current date (todays date).** 
Expected - End Date should be empty.
## Solution:
Not using `new Date()`. When param was in wrong format `new Date()` pointed to today.
empty_end_date

## Bug: 
**Job role dates do not change on modifying job contract dates.**
Expected: Job role dates should get updated on changing job contract dates
## Solution
On JobRoles load check if current dates are other than ones in contract and if they're the same as dates from recent revision. If yes, update JobRole with new contract dates.
update contract

## Bug:
**Edit existing job role -> modify the start/end dates -> and save: dates get reset**
Expected - Should be able to save successfully
## Solution
Using `moment()` instead of `new Date()`. Fixed formatting in `$scope.parseDate()` method.
edit

## Bug:
**Edit existing job role -> modify the contract to use another one: The dates do not get updated with the new contract.**
Expected: Dates should get updated with new contract dates and be able to save successfully.
## Solution
Fix onContractEdited() method.
edit_contract

# Refactor
Moved `DateValidationService` service from angular-date to HRJobRoles, as it's extension specific validator and renamed it into `DateValidation`.
Now we're using `moment()` rather than `new Date();`
Created promise chain instead of randomly creating async calls to API.


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -15,6 +17,7 @@ define([
             $scope.present_job_roles = [];
             $scope.past_job_roles = [];
 
+
```
AkA84: unnecessary new line

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -1,11 +1,13 @@
...
+    'common/moment',
     'common/filters/angular-date/format-date'
-], function (controllers) {
+], function (angular, controllers, moment) {
     'use strict';
 
-    controllers.controller('HRJobRolesController', ['$scope', 'HR_settings', '$log', '$routeParams', 'HRJobRolesService', '$route', '$timeout', '$filter', 'DateValidationService', 'DateFactory',
-        function ($scope, HR_settings, $log, $routeParams, HRJobRolesService, $route, $timeout, $filter, DateValidationService, DateFactory) {
+    controllers.controller('HRJobRolesController', ['$scope', 'HR_settings', '$log', '$routeParams', 'HRJobRolesService', '$route', '$timeout', '$filter', 'DateValidation', '$q',
+        function ($scope, HR_settings, $log, $routeParams, HRJobRolesService, $route, $timeout, $filter, DateValidation, $q) {
```
AkA84: Just to have some kind of consistent approach, let's have the angular built-in components listed first, and then our custom ones:
```js
controllers.controller('HRJobRolesController', ['$filter', '$log', '$q', '$route', '$routeParams', '$scope', '$timeout', 'DateValidation', 'HR_settings', 'HRJobRolesService',
    function ($filter, $log, $q, $route, $routeParams, $scope, $timeout, DateValidation, HR_settings, HRJobRolesService) {
        // ...
    }
]);
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -312,6 +310,40 @@ define([
                         }
                     }
                 }
+
+                if(form_id == 'end_date' && !$scope.edit_data[role_id].end_date){
```
AkA84:
```js
if (form_id === 'end_date' && !$scope.edit_data[role_id].end_date) {
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -312,6 +310,40 @@ define([
...
+                if ($scope.edit_data[role_id].job_contract_id
+                    && $scope.edit_data[role_id].start_date
+                    && typeof $scope.edit_data[role_id].end_date != 'undefined'
+                    && (form_id == 'start_date' || form_id == 'job_contract_id' || form_id == 'end_date')) {
+
+                    $scope.updateRolesWithContractData(role_id);
+                }
+            };
+
+            $scope.updateRolesWithContractData = function (role_id) {
```
AkA84: It doesn't seem you use this function anywhere on the view, so it probably shouldn't be published in the `$scope`

AkA84: Please document every function you wrote

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -759,51 +798,53 @@ define([
...
+                        // Assign data
+                        job_roles.present_job_roles = [];
+                        job_roles.past_job_roles = [];
 
-                                    if (data.count == 0) {
-                                        job_roles.empty = 'No Job Roles found!';
-                                    } else {
-                                        job_roles.empty = null;
-                                    }
+                        angular.forEach(data.values, function (object_data) {
```
AkA84: i don't think you need to use `angular` for this
```js
data.values.forEach(function (objectData) {
});
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -312,6 +310,40 @@ define([
...
                 }
+
+                if(form_id == 'end_date' && !$scope.edit_data[role_id].end_date){
+                    $scope.edit_data[role_id].end_date = null;
+                }
+
+                if ($scope.edit_data[role_id].job_contract_id
+                    && $scope.edit_data[role_id].start_date
+                    && typeof $scope.edit_data[role_id].end_date != 'undefined'
+                    && (form_id == 'start_date' || form_id == 'job_contract_id' || form_id == 'end_date')) {
```
AkA84: Please use `===` instead of `==`

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -312,6 +310,40 @@ define([
...
+
+                    $scope.updateRolesWithContractData(role_id);
+                }
+            };
+
+            $scope.updateRolesWithContractData = function (role_id) {
+                var contract_id = $scope.edit_data[role_id].job_contract_id;
```
AkA84: Please use camelCase (`contractId`) instead of snake_case (`contract_id`) when possible. Some of the property names come directly from the backend so they're stuck with the snake_case, but other than those the standard is the camelCase

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -312,6 +310,40 @@ define([
...
+
+                if ($scope.checkIfDatesAreCustom($scope.edit_data[role_id]['start_date'], $scope.edit_data[role_id]['end_date'])) {
+                    var contract = me.contractsData[contract_id];
+
+                    // search for revision containing these dates
+                    var revision = contract.revisions.filter(function (rev) {
```
AkA84: Since you are checking for `revision.length` (without then actually using `revision`) afterwards, you could use ES5 `some`
```js
var foo = contract.revisions.some(function (revision) {
   return //...
});

if (foo) {
    // ...
}

File: com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js [hunk 9]
```diff
@@ -312,6 +310,40 @@ define([
...
+                        return rev.period_start_date == $filter('formatDate')($scope.edit_data[role_id]['start_date'])
+                            && rev.period_end_date == $filter('formatDate')($scope.edit_data[role_id]['end_date']);
+                    });
+
+                    // check if dates match with revision
+                    if (revision.length) {
```
AkA84: Not sure I understand the logic here: any revision is fine in order to override the dates? It doesn't have to be the latest?

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -398,6 +432,8 @@ define([
 
                         // Hide the empty message if visible
                         $scope.empty = false;
+
```
AkA84: Shouldn't you keep the `// Get job roles based on the passed Contact ID (refresh part of the page)` comment?

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -386,10 +424,6 @@ define([
 
                     // Create the new job role
                     createJobRole($scope.edit_data.new_role_id).then(function () {
-
-                        // Get job roles based on the passed Contact ID (refresh part of the page)
```
AkA84: Maybe this comment could be useful in the function description

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -759,51 +798,53 @@ define([
...
+                    });
+                } else {
+                    contractsPromise = $q.when(job_roles.job_contract_ids);
+                }
 
-                                if (data.is_error == 1) {
-                                    job_roles.error = 'Data load failure';
-                                } else {
+                return contractsPromise.then(function (result) {
+                    return HRJobRolesService.getAllJobRoles(result).then(function (data) {
```
AkA84: Couldn't you write it as
```js
return contractsPromise
    .then(function (result) {
        return HRJobRolesService.getAllJobRoles(result);
    })
    .then(function (data) {
        // ...
    });
```
?

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -826,28 +867,22 @@ define([
 
             // Implements the "createJobRole" service
             function createJobRole(job_roles_data) {
+                return HRJobRolesService.createJobRole(job_roles_data).then(function (data) {
+                    if (data.is_error == 1) {
```
AkA84: Again, please make sure to use `===` instead of `==`

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
...
+    'common/moment'
+], function (module, moment) {
+    /**
+     * Service responsible for validating dates in HRJobRoles
+     * @constructor
+     */
+    module.service('DateValidation', ['HR_settings', function (HR_settings) {
+        var me = this;
+
+        me._error = function (error_msg, fields) {
```
AkA84: Are you using this `_error` property as a public method anywhere?

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
+define([
+    'job-roles/services/services',
+    'common/moment'
+], function (module, moment) {
+    /**
+     * Service responsible for validating dates in HRJobRoles
+     * @constructor
+     */
+    module.service('DateValidation', ['HR_settings', function (HR_settings) {
```
AkA84: Please use `.factory`
```js
module.factory('DateValidation', ['HR_settings', function (HR_settings) {

    function checkIfValuesAreValid(date, fields) {}
    function checkIfStartDateIsLower(start, end) {

    // ...

    return {
        dateFormats: [],
        setErrorCallback: function () {},
        // ...
        validate: function () {}
    };
});

File: com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js [hunk 16]
```diff
@@ -0,0 +1,78 @@
...
+        var me = this;
+
+        me._error = function (error_msg, fields) {
+            throw new Error(error_msg, fields[0]);
+        };
+
+        me.defaultFormat = HR_settings.DATE_FORMAT;
```
AkA84: Is this used anywhere other than inside the service itself?

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
...
+        };
+
+        me.dateFormats = [
+            'x',
+            'YYYY-MM-DD'
+        ];
+
+        me.defaultFormat && me.dateFormats.push(me.defaultFormat.toUpperCase());
+
+        me.validate = function validate(start, end) {
```
AkA84: Please document these functions. For example here it seems `start` and `end` can be of different type, so we need to document it

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
...
+            if (end instanceof Date) {
+                end = end.getTime();
+            }
+
+            var start_date = moment(start, me.dateFormats, true);
+            var end_date = moment(end, me.dateFormats, true);
+
+            checkIfValuesAreValid(start_date, ['start_date']);
+
+            if (end !== '' && typeof end !== 'undefined' && end !== null) {
```
AkA84:
```js
if (end === 0 || end) {
}
```

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
...
+                checkIfValuesAreValid(end_date, ['end_date']);
+            }
+
+            if (start && end_date.isValid()) {
```
AkA84: What's the logic in this `if`? You are checking if `start` (a timestamp) exists and that `end_date` (a moment object) is valid, why the inconsistency between the start and end date?

Also, you just checked in `checkIfValuesAreValid` if `end_date.isValid()`, why redoing it right after it?

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,78 @@
...
+        };
+
+        function checkIfValuesAreValid(date, fields) {
+            if (!date.isValid()) {
+                me._error('Date is not valid!!', fields);
+            }
+        }
+
+        function checkIfStartDateIsLower(start, end) {
+            if (start.isAfter(end)) {
```
AkA84: Why not
```js
function checkIfStartDateIsLower (start, end) {
    if (start.isSameOrAfter(end)) {
        me._error('Start Date cannot be the same as or after the End Date.', ['start_date', 'end_date']);
    }
}

File: com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js [hunk 21]
```diff
@@ -17,12 +16,64 @@ define([
                     "contact_id": contact_id,
                     "deleted": 0,
                     "return": "title,period_end_date,period_start_date"
-                }).done(function(result) {
+                }).done(function (result) {
+                    //get revisions
+                    var revisions = [];
+                    result.values.forEach(function (contract) {
```
AkA84:
```js
var revisions = result.values.map(function (contract) {
    return CRM.api3('HRJobContractRevision', 'get', {
        // ...
    }).then(function (response) {
        // ...
    });
});

File: com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js [hunk 22]
```diff
@@ -17,12 +16,64 @@ define([
...
+                            });
+                        }));
+                    });
 
-                    // Passing data to deferred's resolve function on successful completion
-                    deferred.resolve(result);
 
-                }).error(function(result) {
+                    $q.all(revisions).then(function (response) {
+                        return [].concat.apply([], response);
```
AkA84: Now i'm not actually sure, but doesn't this equal to 
```js
[].concat(response);
```
?

com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
```diff
@@ -17,12 +16,64 @@ define([
...
+                            return CRM.api3('HRJobDetails', 'get', {
+                                "sequential": 1,
+                                "jobcontract_revision_id": item.id
+                            }).then(function (result) {
+                                result.job_contract_id = item.contract_id;
+                                return result;
+                            });
+                        });
+
+                        return $q.all(response);
```
AkA84: More terse
```js
return $q.all(response.map(function (item) {

}));
```

com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
```diff
@@ -17,12 +16,64 @@ define([
...
+                    }).then(function (response) {
+
+                        result.values.map(function (contract) {
```
AkA84: You are not assigning the resulting array to any variable, how is this working?

com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
```diff
@@ -17,12 +16,64 @@ define([
...
+                                }
+                            });
+                        }));
+                    });
 
-                    // Passing data to deferred's resolve function on successful completion
-                    deferred.resolve(result);
 
-                }).error(function(result) {
+                    $q.all(revisions).then(function (response) {
```
AkA84: This here is a very long chain, it would be best explaining what each step (`.then()`) does with a comment

com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/fetching-dates-from-contract_test.js
```diff
@@ -0,0 +1,171 @@
+define([
```
AkA84: These are tests for the `HRJobRolesController` controller, why aren't them in the `hr-job-roles-controller_test.js` file then?

com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
kkalamarski: Not really. I want to flatten this array, so just concat won't work.
Example:
````javascript
[].concat([[1, 2, 3], [4, 5, 6]]); // [[1, 2, 3], [4, 5, 6]]
[].concat.apply([], [[1, 2, 3], [4, 5, 6]]); // [1, 2, 3, 4, 5, 6]
````

File: com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/fetching-dates-from-contract_test.js [hunk 27]
```diff
@@ -0,0 +1,171 @@
...
+            scope = $rootScope.$new();
+            ctrl = $controller('HRJobRolesController', { $scope: scope, format: 'DD/MM/YYYY' });
+
+            ctrl.contractsData = angular.copy(Mock.contracts_data);
+        }));
+
+        describe('Checking if dates entered in job role are th same as those in contracts', function () {
+
+            it('should check if entered dates are custom', function () {
+                expect(scope.checkIfDatesAreCustom('2005-01-01', null)).toBe(true);
```
AkA84: Shouldn't this be `ctrl.checkIfDatesAreCustom`?

com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/fetching-dates-from-contract_test.js
```diff
@@ -0,0 +1,171 @@
...
+        beforeEach(inject(function ($controller, $rootScope) {
+            scope = $rootScope.$new();
+            ctrl = $controller('HRJobRolesController', { $scope: scope, format: 'DD/MM/YYYY' });
+
+            ctrl.contractsData = angular.copy(Mock.contracts_data);
+        }));
+
+        describe('Checking if dates entered in job role are th same as those in contracts', function () {
+
+            it('should check if entered dates are custom', function () {
```
AkA84: The `it` descriptions should be in this format `it('checks if entered dates are custom')` not in the "should" format

com.civicrm.hrjobroles/js/src/job-roles/services/date-validation.js
```diff
@@ -0,0 +1,91 @@
...
+                }
+            },
+
+            /**
+             * Validates Dates
+             * @param {Date|string|int} start
+             * @param {Date|string|int} end
+             */
+            validate: function validate(start, end) {
+                HR_settings.DATE_FORMAT && this.dateFormats.push(HR_settings.DATE_FORMAT.toUpperCase());
```
AkA84: Why are you doing this every time you call `validate`? In the previous iteration you were doing once while cosructing the service

AkA84:
Approved

PR #864, opened 2016-02-19
PCHR-814: Min and Max date range removed from HRJobRoles

kkalamarski:
## Problem
User was unable to select dates prior to year 2000, and after year 2020 through datepicker.
## Solution
Remove code responsible for minDate and maxDate

no_min_max


AkA84:
Approved

PR #868, opened 2016-02-25
PCHR-858: Appraisal Cycle - Summary (AngularJS) (2/2)

AkA84:
This PR is the continuation of #858 , adding more custom and third-party components to `org.civicrm.reqangular` (common AngularJS components), implementing new functionalities to `uk.co.compucorp.civicrm.appraisals` (Appraisals app), and fixing more general issues

#### Things done
* The build process in the `gulpfile`s of the angular apps didn't output anything if an error occured (for example in case of a missing module). The process has now been improved, and will display the `r.js` output in case of an error

 ```js
  gulp.task('requirejs-bundle', function (done) {
      exec('r.js -o build.js', function (err, stdout, stderr) {
          err && err.code && console.log(stdout);
          done();
      });
  });
 ```
* Amended the "Overdue Appraisals" table, so that it matches in structure the appraisals table in the "Appraisals in Cycle" table
  ##### Before
  overdue-table-before

  ##### After
  overdue-table-after

* Implemented functionality of the "Add Contacts" modal
  add-contact-modal
* The "Cycle Summary" and "Appraisals in Cycle" tabs are now sub-states (`appraisals.appraisal-cycle.cycle-summary` and `appraisals.appraisal-cycle.appraisals-in-cycle`) instead of simple Bootstrap tabs. This allows us to programmatically switch to one or the other after a certain event occurs
* Removed the "Access Settings" link, as we don't have that functionality finalized yet
  ##### Before
  access-settings-before
  
  ##### After
  access-settings-after
* Added the angular edit-in-place module ([`angular-xeditable`](https://vitalets.github.io/angular-xeditable/)) as a common dependency. Not implemented yet as we need to figure out some things re the layout
* Implemented the angular WYSIWYG module ([`text-angular`](https://github.com/fraywing/textAngular)), added as a common dependency
  wysiwyg
* Implemented the lookup select module ([`ui-select`](https://github.com/angular-ui/ui-select)), added as a common dependency
  lookup

#### Things not done
This PR is already massive enough, so the remaining things left to do will be done in a different PR
* Implementation of the edit-in-place module
* Tests for the `api`s services
* Various refactoring and improvements

#### Changed file structure of custom modules
We have custom modules, like `angular-date` or `dialog`, that are a collection of different components (directives, controllers, services, etc).

##### Before
The module definition and the registration of its components would be spread across multiple files in multiple directories:
```js
// src/common/modules/dialog.js
define([ /* ... */ ], function (angular) {
  'use strict';
  return angular.module('common.dialog', ['ui.bootstrap', 'common.templates']);
});
```
```js
// src/common/services/dialog.js
define([ /* ... */ ], function (dialog) {
  'use strict';
  dialog.factory('dialog', [ /* .. */, function (/* .. */) {
    return { /* .. */ };
  }]);
});
```
```js
// src/common/controllers/dialog.js
define([ /* ... */ ], function (dialog) {
  'use strict';
  dialog.controller('DialogCtrl', [ /* .. */,  function (/* .. */) {
    // ...
  }]);
});
```
As a result it wasn't possible for any given app to include just one "main" file in order to be able to use the module, as there was no such thing (the module file just registered the module, and nothing else):
```js
// src/my-app/app.js
define([
  // ...
  'common/services/dialog',
  'common/controllers/dialog'
], function (angular) {
  angular.module('myApp', ['dialog']);
});
```

##### After
The module file now contains both the module **and** the components registration:
```js
// src/common/modules/dialog.js
define([
  'common/angular',
  'common/services/dialog/dialog',
  'common/controllers/dialog/dialog-ctrl',
  // ...
], function (angular, dialog, DialogCtrl) {
  'use strict';

  return angular
    .module('common.dialog', ['ui.bootstrap', 'common.templates'])
    .factory('dialog', dialog)
    .controller('DialogCtrl', DialogCtrl);
});
```
which allows for an easier insertion in any app that requires the module as a dependency:
```js
// src/my-app/app.js
define([
  // ...
  'common/module/dialog',
], function (angular) {
  angular.module('myApp', ['dialog']);
});
```
The individual components' files contain only the implementation, which is then injected in the module file and used for the component's registration. This is not optimal, as the name of the component is decoupled from the implementation, but it's the best solution I could find (which I think was also suggested once by @kkalamarski !). Example:
```js
// src/common/services/dialog/dialog.js
define([], function () {
  'use strict';

  return [ '$modal', '$rootElement', '$templateCache', function (/* .. */) {
    return { /* .. */ };
  }];
});
```

#### `text-angular`
Adding the latest release of this module has been quite problematic, as both the module itself and one of its dependencies have outstanding issues with RequireJS. Eventually I had to amend two of the files to make it work.

I left a comment in both files to serve both as an explanation and as a reminder, in case we will have to upgrade the module:
```js
// src/common/vendor/rangy-core.js
/*
    !CUSTOMIZED VENDOR FILE!

    Rangy is written to correctly work as an AMD module, but textAngular.js
    (which is a shimmed, not-AMD "module") expects window.rangy to exist.

    The workaround consists of making sure a global variable is created before
    defining the module


    More details on dependencies for shimmed modules here:
    http://requirejs.org/docs/api.html#config-shim

    > Only use other "shim" modules as dependencies for shimmed scripts,
    or AMD libraries that have no dependencies and call define() after they
    also create a global
*/
```
```js
// src/common/vendor/textAngular.js
/*
	!CUSTOMIZED VENDOR FILE!

	Contains workaround for these issues:
	  * https://github.com/fraywing/textAngular/issues/892
	  * https://github.com/fraywing/textAngular/issues/1016

	Removed the require() calls in the .run() block
*/
```

#### CSS for 3rd-party modules
The css of 3rd-party modules like `text-angular` or `ui-select` are placed in the `org.civicrm.bootstrapcivicrm` extension, not in the `org.civicrm.reqangular` one. This is to have all styles under the same extension, and to be able to include them in the `style.scss` main file, which allows us to have their CSS namespaced, and thus avoiding specificity issues

```scss
/* ... */

#{$bootstrap-namespace} {
    @import "vendor/text-angular/text-angular";
    @import "vendor/ui-select/ui-select";

    @import "overrides/style/*";
    @import "components/*";
}

@import "vendor/angular-xeditable/angular-xeditable";
@import "vendor/font-awesome/font-awesome";
```
#### Amended `Appraisal` API service and model with `.create()`
It is now possible to create a new Appraisal via the `Appraisal.create()` method, which in turn will call a method with the same name on the API. The method has both `contact_id` and `appraisal_cycle_id` as mandatory parameter
```js
Appraisal.create({ appraisal_cycle_id: '3', contact_id: '4' })
  .then(function (appraisal) {
    appraisal; // A `AppraisalInstance` object
  });
```
#### Amended `OptionGroup` API service
The API service for the `OptionGroup` entity (`api.optionGroup`) has now a `valuesOf` method that will also accept multiple group names and will return a list of `OptionValue`s grouped by the group name:
```js
optionGroupAPI.valuesOf(['hrjc_department', 'hrjc_region', 'hrjc_location'])
  .then(function (values) {
    values['hrjc_department']; // Array of values for hrjc_department
    values['hrjc_region']; // Array of values for hrjc_region
  });
```
The method still accepts a simple string with the name of one group, returning directly the array of values of the group:
```js
optionGroupAPI.valuesOf('hrjc_location')
  .then(function (values) {
    values; // Array of values for hrjc_location
  });
```
#### Added `OptionGroup` model 
The model offers a `valuesOf` method that simply wraps the API method of the same name

#### Added `Group` API, model and model instance
*(For now the API is only mocked)*

The API service (`api.group`) offers basic `all()` and `find()` methods.
The `Group` model additionally offers a `contactIdsOf()` method, which will return the list of contact ids of the group with the given id:
```js
Group.contactIdsOf('4').then(function (contactIds) {
  contactIds; // Array of contact ids
});
```
`GroupInstance` doesn't have custom methods for now, it just defines the `fromAPI` transformation.

#### Added `JobRole` API, model and model instance
*(For now the API is only mocked)*

The API service (`api.job-role`) offers basic `all()` and `find()` methods. 
The `JobRole` model offers the same method of the API service, `all()` and `find()`
`JobRoleInstance` doesn't have custom methods for now, it just defines the `fromAPI` and `toAPI` transformations.

#### Added `Contact` API, model and model instance
*(For now the API is only mocked)*

The API service (`api.contact`) offers basic `all()` and `find()` methods. 

The `Contact` model offers the same method of the API service, `all()` and `find()`.
The `all()` method though is peculiar as it allows the "masking" of queries that should be in fact directed to models of other entities, in particular (at this moment) `JobRole` and `Group`. What it means is that it is possible to use the `Contact` model to, for example, ask all the contacts born after 01/01/1980, with their `JobRole` in department X and location Y, while also belonging to `Group` Z:
```js
Contact.all({
  'birth_date': { from: '01/01/1980' },
  'department': 'X',
  'location': 'Y',
  'group_id': 'Z'
})
.then(function (contacts) {
  contacts.list; // List of all the contacts matching the criteria
});
```
Querying by location, or department, or group, etc is not possible on the `Contact` entity itself as those are not fields present in its table, so what is happening under the hood here is that the `Contact` model recognizes those "foreign" parameters and passes them to the entity that they belong to. The entity models will execute their queries and pass back to the `Contact` model the list of contact ids that matched the criteria, and at that point `Contact` can use those ids to finalize the query.
The advantage of this solution is that it allows to write more terse and logical code when performing `Contact`-related queries, while hiding the implementation details of those queries behind the model.

`ContactInstance` doesn't have custom methods or transformations for now, it just extends the basic `ModelInstance`


kkalamarski:
Approved

PR #869, opened 2016-02-26
PCHR-478: Datepicker must fit into window

kkalamarski:
## Problem:
Sometimes datepicker wasn't fully visible.
before1
## Solution:
When datepicker is opened calculate distance from window edge. If needed change position of datepicker.
after1


org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
```diff
@@ -16,6 +16,23 @@ define([], function () {
                  */
                 attrs.datepickerPopup = HR_settings.DATE_FORMAT;
 
+                scope.$watch('isOpen', function (value) {
```
AkA84: Please move the logic to a separate function

org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
```diff
@@ -16,6 +16,23 @@ define([], function () {
...
+                    if (value) {
+                        scope.popupPosition = $position.position(element);
+                        scope.popupPosition.top += scope.popupPosition.height;
+
+                        var popupWidth = 316;
```
AkA84: Two things:
1) This seems like a magic constant. Do we need this exact number, or can we extrapolate it in some way? If we cannot, please move it to a "constant" with a comment as to why we need this number
2) Please always declare `var`s at the beginning of the function

org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
```diff
@@ -16,6 +16,23 @@ define([], function () {
...
+
+                        var popupPosition = element.offset().left + element.width();
+                        popupPosition += popupWidth;
+                        
+                        if( popupPosition > $window.innerWidth ){
```
AkA84:
```js
if (popupPosition > $window.innerWidth) {
```

org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
```diff
@@ -16,22 +16,30 @@ define([], function () {
                  */
                 attrs.datepickerPopup = HR_settings.DATE_FORMAT;
 
-                scope.$watch('isOpen', function (value) {
+                /**
+                 * Calculates position of the datepicker popup.
+                 * @param value
+                 */
+                var adjustPopupPosition = function (value) {
```
AkA84: You could move this function definition outside the `compile` function, to make the latter cleaner and more readable
```js
var original_link = $delegate[0].link;

function adjustPopupPosition(value) {
    // ...
}

$delegate[0].compile = function ()  {
    return function (scope, element, attrs, ngModel) {
        attrs.datepickerPopup = HR_settings.DATE_FORMAT;
        
        scope.$watch('isOpen', adjustPopupPosition);
        original_link.apply(this, arguments);
    }
};

// ...
```

AkA84: Also, please put the entire module in strict mode

org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
```diff
@@ -16,22 +16,30 @@ define([], function () {
                  */
                 attrs.datepickerPopup = HR_settings.DATE_FORMAT;
 
-                scope.$watch('isOpen', function (value) {
+                /**
+                 * Calculates position of the datepicker popup.
+                 * @param value
```
AkA84: Define the type of `value` and give a brief description of what it is

```js
/**
 * Calculates position of the datepicker popup.
 * @param {int} value - Lorem ipsum
 */

File: org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js [hunk 6]
```diff
@@ -16,22 +16,30 @@ define([], function () {
...
+                 */
+                var adjustPopupPosition = function (value) {
                     if (value) {
+                        var popupPosition;
+                        var popupWidth = element.parent().find('.datepicker-popup').outerWidth();
+
+
```
AkA84: Unnecessary new line

org.civicrm.reqangular/src/common/decorators/angular-date/datepicker-popup.js
kkalamarski: I'm using `scope` and `element` vars inside this function. If I move this function outside compile closure, these variables will become undefined.

AkA84: You're right, my bad!

AkA84:
Approved

PR #870, opened 2016-03-21
PCHR-860: Fix bug in my absences to use title instead of name

omar-compucorp:
## Problem :

When you add new absence type with spaces , and then you visit "my absences" or any other contact "absences tab" . the name ( which contains underscores ) of the absence type appears instead of the title  . 

before

## Solution :

I've changed the logic that retrieve absence types list to use "title" field instead of "name".

selection_001

selection_002


AkA84:
Approved

PR #874, opened 2016-03-30
PCHR-868: Fix a bug that prevent importing any job role

omar-compucorp:
### Problem 

When trying to import  job roles the following error appear for all the records :

selection_001
 
selection_002


coldrunKacper:
Approved.

PR #875, opened 2016-03-31
PCHR-871-creating-additional-BAO-appraisal-validation-method

coldrunKacper:
I've created a new 'validateNewAppraisal()' static method which is executed every time we try to create new Appraisal. The method check for some requirements plus there is a new one which checks if there is already existing Appraisal for given Contact and Cycle ID.
We assume that Contact can have only single Appraisal per Cycle so if there is already current Appraisal existing for given Contact and Cycle then an exception is thrown.

### DIFF
--- a/uk.co.compucorp.civicrm.appraisals/CRM/Appraisals/BAO/Appraisal.php
+++ b/uk.co.compucorp.civicrm.appraisals/CRM/Appraisals/BAO/Appraisal.php
@@ -17,18 +17,12 @@ class CRM_Appraisals_BAO_Appraisal extends CRM_Appraisals_DAO_Appraisal
         $now = CRM_Utils_Date::currentDBDate();
         
         if ($hook === 'create') {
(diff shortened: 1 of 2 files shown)


davialexandre:
@coldrunKacper Approved
I believe this will work only for new appraisals, right? Are you going to change anything in the SSP dashboard code to fix the error where Appraisal Cycles already have more than one Appraisal?

coldrunKacper:
Yeah, I was thinking about it but don't see any solution to handle currently existing multiple appraisals per cycle / contact. @AkA84 do you think we should have some kind of Upgrader function deleting all duplicated appraisals excepting the newest (or oldest)?

AkA84:
@coldrunKacper yes that would work, I believe it's better to keep the newest appraisal

coldrunKacper:
Added the Upgrader function removing multiple Appraisals per Contact per Appraisal Cycle.

davialexandre:
@coldrunKacper approved

PR #876, opened 2016-04-01
PCHR-505 Empty dates save fix

kkalamarski:
## Problem
When creating Job Role without `end_date`, it was set to current date.

## Solution
Sending `null` to the API instead of `0`.
When API receives `null` as an end date, it saves it to database as `0000-00-00 00:00:00`.
Thus, needed to change condition to distinguish past job roles from present job roles.


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -55,8 +65,8 @@ define([
              * @returns {boolean}
              */
             $scope.checkIfDatesAreCustom = function (start, end) {
-                if (start === '0000-00-00 00:00:00') start = null;
-                if (end === '0000-00-00 00:00:00') end = null;
+                if (start === isDateEmpty(start)) start = null;
```
AkA84: I don't understand this comparison, shouldn't it be
```js
if (isDateEmpty(start)) {
}
```

AkA84:
Can you write tests for this fix?

PR #879, opened 2016-04-01
PCHR-851: Removes rejected leaves from the Calendar in Absences tab of the contact

davialexandre:
Now, besides the cancelled leaves, the calendar will also not show the rejected ones.

I've updated the Activity.getAbsences API action to handle multiples IDs passed to the status_id parameter. It now supports the != and the "NOT IN" operators.

Bellow are some screenshots showing the results:

Given 3 leave requests (one rejected, one cancelled and one waiting approval)
captura de tela 2016-04-01 as 15 29 25

Before the change, the calendar was showing the rejected request:
captura de tela 2016-04-01 as 15 30 46

After the change, the rejected request is longer displayed:
captura de tela 2016-04-01 as 15 29 37

### DIFF
--- a/hrabsence/tests/phpunit/api/v3/ActivityGetAbsencesTest.php
+++ b/hrabsence/tests/phpunit/api/v3/ActivityGetAbsencesTest.php
@@ -239,4 +239,49 @@ class api_v3_ActivityGetAbsencesTest extends CiviUnitTestCase {
       }
     }
   }
+
+  public function testStatusIdParameter() {
+    $allActivities = $this->callAPISuccess('Activity', self::ACTION, array('sequential' => 1));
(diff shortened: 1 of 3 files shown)


coldrunKacper:
Approved.

hrabsence/tests/phpunit/api/v3/ActivityGetAbsencesTest.php
```diff
@@ -239,4 +239,49 @@ public function testOptionAbsenceRange() {
       }
     }
   }
+
+  public function testStatusIdParameter() {
```
AkA84: Could you split these test into 3 different tests (with meaningful method names), one for each assertion blocks? It improves readability and structure of the test suite

coldrunKacper:
Approved.

PR #880, opened 2016-04-01
PCHR-880: Fix the wrong data inserted in contracts after importing them

omar-compucorp:
#### The Problem 

When importing contracts from "import job contracts" page , the inserted contracts are different from the imported ones.

**sample file**
selection_004

**when viewing them**
selection_005

The source of problem was when importing new contracts , the revision IDs  inserted in the database are wrong and refer to previously inserted contracts .


coldrunKacper:
Does this change not have any side-effects as you removed $contractRevison->save(); ? Do revisions still get saved during the import?

omar-compucorp:
@coldrunKacper I checked the saved contracts and it does not appear that it does have any side effects and the revisions saved correctly . I've also edited the imported contracts from UI and there where no problem.

coldrunKacper:
Approved.

PR #882, opened 2016-04-04
PCHR-853: Fix to ui.select w/ AngularJS v1.5

AkA84:
This is a fix related to https://github.com/compucorp/civihr-tasks-assignments/pull/31, where the class-based fix is implemented on one of our custom directive


davialexandre:
@AkA84 Approved

PR #883, opened 2016-04-04
PCHR-774: Removes Initial/Final contract dates fields from Job Contract Summary

davialexandre:
The definition XML for both custom fields was removed, so they won't be included in a new installation. For existing installations, the fields can be manually deleted through the UI.

I couldn't figure out a way to test if this is working. I followed Jamie's comment on this (https://compucorp.atlassian.net/browse/PCHR-774?focusedCommentId=25640&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25640) and I believe it's working, but I would appreciate it if anyone could tell me  a way to test this.


coldrunKacper:
Approved.

PR #884, opened 2016-04-05
PCHR-799: import contracts revision

omar-compucorp:
Allowing users to be able to import job contracts revision by introducing a new "import mode" checkbox at "contracts import" page.

selection_002

screenshot from 2016-04-05 09 59 16


coldrunKacper:
Approved.

PR #885, opened 2016-04-05
PCHR-865 - Implementing 'effective_end_date' revision field

coldrunKacper:
- Adding a new field called 'effective_end_date' to Job Contract Revision entity which keep an end of revision effective date if there is any newer revision.
- Creating Upgrader script which populates 'effective_end_date' values for every existing Job Contract revisions.
- Value of the field is automatically generated while saving Job Contract's revision.

The new field is designed mostly to handle Reports requirements but it can also be used with any other cases in the future.


omar-compucorp:
Approved

PR #887, opened 2016-04-06
PCHR-890: Job roles are always fetched even if contact has no contract

AkA84:
## Problem
As a result of #862 , the Promise chains that would fetch the job contracts of a given contact was altered, and we ended up fetching job roles that didn't belong to the current user.

Before #862 the chain would be as follow:
1. Get the user's contracts starting from the user id
2a. If contracts are returned, fetch the job roles using the contracts ids
2b. I no contracts are returned, stop there and display message

After #862 the chain was changes as such
1. Gets th user's contracts starting from the user id
2. no matter if the user has job contracts assigned to him, fetch the job roles

## Solution
The chain logic has been restored, and tests for it have been added


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -810,60 +810,60 @@ define([
...
-                            job_roles.contractsData = contractsData;
+            function contractIdsFromContact(contactId) {
+                return HRJobRolesService.getContracts(contactId).then(function (data) {
+                    var jobContractIds = [];
+                    var contractsData = {};
+
+                    // If we have job contracts, try to get the job roles for the contract
+                    if (data.count !== 0) {
+                        for (var i = 0; i < data.count; i++) {
+                            // Job contract IDs which will be passed to the "getAllJobRoles" service
```
davialexandre: I think this comment is too specific about getAllJobRoles. Can't this method be used by other services?

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -810,60 +810,60 @@ define([
...
+                            contract.label = contract.title + ' (' + $filter('formatDate')(contract.start_date) + ' - ' + optionalEndDate + ')';
+
+                            contractsData[data.values[i]['id']] = contract;
+                        }
 
-                            job_roles.job_contract_ids = job_contract_ids;
+                        // Store the ContractsData what we can reuse later
+                        job_roles.contractsData = contractsData;
+                        job_roles.job_contract_ids = jobContractIds;
+                    } else {
```
davialexandre: As a matter of style, what do you think about returning early instead of a if/else? Something like:
```
if (data.count === 0) {
   job_roles.empty = "...";
   return jobContractIds;
}
for (var i = 0; i < data.count; i++) {
...
}
```
This looks simpler to me and you get fewer levels of indentation.

davialexandre:
@AkA84 Approved. Left a couple of comments about coding style

kkalamarski:
Approved

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
AkA84: (Given this controller is a behemoth of almost 1000 lines and it is in dire need of some refactoring) It doesn't seem so, this method is called only by the original `getJobRolesList`, and it's the only place where `HRJobRolesService.getContracts` is called so there doesn't seem to be a duplication happening here

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
AkA84: Always felt split between the two different styles, because while it's true you reduce the levels of indentation, it's also true that you end up with two different returns, while with the other style you have only one place (the end of the function) where any value is returned and thus maybe the code flow is easier to follow

PR #889, opened 2016-04-07
PCHR-889: Add Job Role fields in Advanced Search

davialexandre:
A new panel was added to the Avanced Search, containing
all of the Job Role fields.

This is what the panel looks like:

captura de tela 2016-04-07 as 16 04 31


coldrunKacper:
Approved.

PR #890, opened 2016-04-08
PCHR-901: Fix rangy-selectionsaverestore RequireJS issue

AkA84:
This is the same fix applied in #868 to `rangy-core.js`, only applied to `rangy-selectionsaverestore.js`.
Without the fix you would have a *"rangy-selectionsaverestore.js is required for textAngular to work correctly"* error, caused by the `rangy` property in the global object not being handled correctly


kkalamarski:
Approved

PR #891, opened 2016-04-08
PCHR-892: Cleaning contracts import validation

omar-compucorp:
When importing contracts , many of  validation rules are incorrect because they are relying on field type attribute from DAO file. also i fixed  other issues in contract import such as the error page that appear if the health or life insurance provider field is empty.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobHealth.php
+++ b/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobHealth.php
@@ -56,6 +56,35 @@ class CRM_Hrjobcontract_BAO_HRJobHealth extends CRM_Hrjobcontract_DAO_HRJobHealt
 
 
   /**
+   * Check insurance provider if exist   .
+   *
+   * @param String $searchValue
+   * @param String $providerType
+   * @return Integer ( Provider ID or 0 if not exist)
+   */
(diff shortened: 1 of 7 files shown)


hrjobcontract/CRM/Hrjobcontract/BAO/HRJobHealth.php
```diff
@@ -56,6 +56,35 @@ public static function create($params) {
...
+   */
+  public static function checkProvider($searchValue, $providerType) {
+    if (is_numeric ($searchValue))  {
+      $searchField = 'id';
+    }
+    else {
+      $searchField = 'display_name';
+    }
+    $queryParam = array(1 => array($searchValue, 'String'));
+    $query = "SELECT id,contact_sub_type FROM civicrm_contact WHERE ".$searchField." = %1 " .
```
nishant-bhorodia: I think we should use `"SELECT id,contact_sub_type FROM civicrm_contact WHERE {$searchField}"` instead of `"SELECT id,contact_sub_type FROM civicrm_contact WHERE ".$searchField."`. Nothing programatically wrong with this but it just decreased the number of double quotes :)

nishant-bhorodia:
Thanks @omar-compucorp . Approved !!

PR #893, opened 2016-04-11
PCHR-618: Set contact summary tab to be the default one

omarabuhussein:
## Problem
Contact "Personal Details" tab is the first and default tab in contact summary page but "Contact Summary" tab is needed to be the first instead.

selection_004

## Solution 
Civicrm Core files are altered and the patch submitted to Civi-Core both master and 4.5 versions (https://issues.civicrm.org/jira/browse/CRM-18390)  Also currently **jobcontract** extension override the summary page template file . So this file is used  **hrjobcontract/templates/CRM/Contact/Page/View/Summary.tpl**  as a template file instead of this **CRM/Contact/Page/View/Summary.tpl** which means that i have to remove the hardcoded summary tab from their too.

selection_006

## Related Links
- CiviCRM Jira : https://issues.civicrm.org/jira/browse/CRM-18390
- CiviCore patch (Master) : https://github.com/civicrm/civicrm-core/pull/8110https://github.com/civicrm/civicrm-core/pull/8110
- CiviCore patch (4.5) : https://github.com/civicrm/civicrm-core/pull/8109

### DIFF
--- a/contactsummary/js/src/contact-summary.js
+++ b/contactsummary/js/src/contact-summary.js
@@ -6,7 +6,5 @@ require.config({
 });
 
 require(['contact-summary/app'], function () {
-    document.addEventListener('contactsummaryLoad', function () {
-        angular.bootstrap(document.getElementById('contactsummary'), ['contactsummary']);
-    });
+    angular.bootstrap(document.getElementById('contactsummary'), ['contactsummary']);
 });
--- a/contactsummary/contactsummary.php
+++ b/contactsummary/contactsummary.php
@@ -127,9 +127,12 @@ function contactsummary_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
  */
 function contactsummary_civicrm_pageRun($page) {
     if ($page instanceof CRM_Contact_Page_View_Summary) {
-        CRM_Core_Resources::singleton()->addVars('contactsummary', array(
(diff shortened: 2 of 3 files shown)


contactsummary/js/src/contact-summary.js
```diff
@@ -6,7 +6,5 @@ require.config({
 });
 
 require(['contact-summary/app'], function () {
-    document.addEventListener('contactsummaryLoad', function () {
```
AkA84: Why did you remove the event listener?

AkA84:
Sorry could you explain this in more details?
> Also currently jobcontract extension override the summary page template file so necessary changes done there too.

contactsummary/js/src/contact-summary.js
omarabuhussein: after set the "contact summary" tab to be the first and the default tab instead of "personal details" tab the tab content don't get loaded and that's because the line below do not get called :

` angular.bootstrap(document.getElementById('contactsummary'), ['contactsummary']);`

omarabuhussein:
@AkA84 description updated

contactsummary/js/src/contact-summary.js
omarabuhussein: @AkA84 also if you have any idea in how to fix that without removing the listener that would be better .

AkA84:
Approved

PR #894, opened 2016-04-15
PCHR-876: Fix job roles import validation problems

omarabuhussein:
## Problem 

The following errors are reported by QA while trying to import job roles :

1- Region not updated when value is valid, error not thrown on value mismatch.
2- Cost center => Type and amount not displayed on front end.
3- Cost center => if percent and absolute amount where both exist in the import file No error shown saying either one should be removed.
4- Funder => Type and amount not displayed on front end.
5- Funder => if percent and absolute about where both exist in the import file No error shown saying either one should be removed.

Also here some typos corrected :

1- start_date in import match select box should be => "Start Date"
2- end_date in import match select box  should be => "End Date"
3- job id in import match select box  should be=> "Job ID"
4- import page title  should be  "Import Job Roles" instead of "import job".

## Solution 

The logic in summary function is partially modified to fix these import problems. and the typos get fixed in the related files.


davialexandre:
@omar-compucorp approved

PR #895, opened 2016-04-15
PCHR-891: Complete Job Contract and Role Export Feature

davialexandre:
### Problem

The Contract ID and the Job Role fields weren't available to be included in the file exported from the Advanced Search

### Solution

The fields were configured to be available to the exported file. The Job Role fields were prefixed with 'Job Role' so they could be easily found on the list:

captura de tela 2016-04-15 as 16 18 13

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
@@ -37,9 +37,18 @@ class CRM_Hrjobroles_BAO_Query extends CRM_Contact_BAO_Query_Interface {
         '1' => '%'
     );
 
+    /**
+     * @var array A list of fields available while importing/exporting
+     */
+    private $hrjobRoleFields = array();
+
+
     public function &getFields() {
-        $arr = array();
-        return $arr;
+        if (empty($this->hrjobRoleFields)) {
+            $this->hrjobRoleFields = CRM_Hrjobroles_DAO_HrJobRoles::export();
+        }
+
+        return $this->hrjobRoleFields;
     }
 
     /**
@@ -53,7 +62,7 @@ class CRM_Hrjobroles_BAO_Query extends CRM_Contact_BAO_Query_Interface {
         $from = '';
         switch ($name) {
             case 'civicrm_contact':
-                $from .= " $side JOIN civicrm_hrjobroles hrjobroles ON hrjobcontract.id = hrjobroles.job_contract_id ";
+                $from .= " $side JOIN civicrm_hrjobroles ON hrjobcontract.id = civicrm_hrjobroles.job_contract_id ";
                 break;
         }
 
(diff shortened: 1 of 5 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
```diff
@@ -37,9 +37,18 @@ class CRM_Hrjobroles_BAO_Query extends CRM_Contact_BAO_Query_Interface {
         '1' => '%'
     );
 
+    /**
+     * @var array A list of fields available while importing/exporting
+     */
+    private $hrjobRoleFields;
```
omarabuhussein: i think it's better to give this an initial empty value , something like :

private $hrjobRoleFields = array();

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
```diff
@@ -37,9 +37,18 @@ class CRM_Hrjobroles_BAO_Query extends CRM_Contact_BAO_Query_Interface {
...
+
+
     public function &getFields() {
-        $arr = array();
-        return $arr;
+        if (!$this->hrjobRoleFields) {
```
omarabuhussein: and here instead of   if (!$this->hrjobRoleFields) use  if (empty($this->hrjobRoleFields))

com.civicrm.hrjobroles/hrjobroles.php
```diff
@@ -188,3 +188,84 @@ function hrjobroles_civicrm_queryObjects(&$queryObjects, $type) {
...
+ * @param $exportTempTable
+ * @param $headerRows
+ * @param $sqlColumns
+ * @param $exportMode
+ */
+function hrjobroles_civicrm_export( $exportTempTable, $headerRows, $sqlColumns, $exportMode ) {
+
+    $splitItemsQuery = function($items) {
+        return "SELECT SUBSTRING_INDEX(SUBSTRING_INDEX($items, '|', n.n), '|', -1) value
+FROM (
```
omarabuhussein: I think the returned value may require a better code styling and indentation .. for example instead of one big string you can separate lines into strings and concat them  something like this :

return "Select * From ...... " .
           "Where ....... " .
           "And ......";

com.civicrm.hrjobroles/hrjobroles.php
```diff
@@ -188,3 +188,84 @@ function hrjobroles_civicrm_queryObjects(&$queryObjects, $type) {
...
+    // not letting any leading or trailing commas
+    $fieldsToSplitValues = array(
+        'hrjc_role_amount_pay_cost_center',
+        'hrjc_role_amount_pay_funder',
+        'hrjc_role_percent_pay_cost_center',
+        'hrjc_role_percent_pay_funder'
+    );
+    foreach($fieldsToSplitValues as $i => $field) {
+        if(isset($sqlColumns[$field])) {
+            $optionsUpdates[] = "et.$field = (SELECT REPLACE(REVERSE(SUBSTRING(REVERSE(SUBSTRING(et.$field, 2)), 2)), '|', ', '))";
```
omarabuhussein: That's a crime against MySQL  :P .... but i know it's not your fault :D

com.civicrm.hrjobroles/hrjobroles.php
davialexandre: I'm really not sure about this. I think that one big string, though it looks ugly, makes things easier if I need to copy the query and run it somewhere else.

omarabuhussein:
I left some notes it's up to you if you find them make sense to correct them .. otherwise everything is fine .

com.civicrm.hrjobroles/hrjobroles.php
omarabuhussein: yeah sounds  like a good idea , but at least correct the indentation if that possible

PR #896, opened 2016-04-18
PCHR-772: Improvements to lookup directive

AkA84:
Switching the caret icon for the font awesome search icon. The markup of the directive template has been changed to make use of Bootstrap's `form-control-feedback` element

#### Before
before

#### After
after


kkalamarski:
Approved

PR #897, opened 2016-04-18
PCHR-943: Create Leave and Absences extension

davialexandre:
This creates only the extension skeleton with basic hooks implementations
to define the extension's menu items.

None of the links/pages the menu items point to are yet implemented.

These are the new menus:

captura_de_tela_2016-04-18_as_13_49_09
captura_de_tela_2016-04-18_as_13_48_56

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
@@ -1,0 +1,284 @@
+<?php
+
+require_once 'hrleaveandabsences.civix.php';
+
+/**
+ * Implements hook_civicrm_config().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_config
+ */
+function hrleaveandabsences_civicrm_config(&$config) {
+  _hrleaveandabsences_civix_civicrm_config($config);
+}
+
+/**
+ * Implements hook_civicrm_xmlMenu().
+ *
+ * @param array $files
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_xmlMenu
+ */
+function hrleaveandabsences_civicrm_xmlMenu(&$files) {
+  _hrleaveandabsences_civix_civicrm_xmlMenu($files);
+}
+
+/**
+ * Implements hook_civicrm_install().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_install
+ */
+function hrleaveandabsences_civicrm_install() {
+  _hrleavesandabsences_create_main_menu();
+  _hrleaveandabsences_create_administer_menu();
+
+  _hrleaveandabsences_civix_civicrm_install();
+}
+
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -0,0 +1,238 @@
...
+function hrleaveandabsences_civicrm_xmlMenu(&$files) {
+  _hrleaveandabsences_civix_civicrm_xmlMenu($files);
+}
+
+/**
+ * Implements hook_civicrm_install().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_install
+ */
+function hrleaveandabsences_civicrm_install() {
```
AkA84: This function is a bit long, could you split it into smaller functions? It would improve the readability and maintainability of the function as a whole

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -0,0 +1,238 @@
...
+  $reportWeight = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'Reports', 'weight', 'name');
```
AkA84: Why Reports in particular?

davialexandre: I was using the hrabsences extension as a reference to implement this and it uses Reports.  It made sense to me to use it here too, since it is the first non-civihr extension on the menu (not counting search and contacts) its weight - 1 works as a way to "add this menu item after the last civihr item"

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
davialexandre: I wanted to do this, but I was not sure if it was OK to add non-hook functions to this file. Is that ok? If so I would be really happy to do it. I hate those long methods/functions :(

coldrunKacper: Yup, sure you can add non-hook functions (see _hrleaveandabsences_civix_civicrm_xmlMenu() helper function for example).

coldrunKacper: OR you can move the navigation install part into Upgrader class but at this point there is no Upgrader.php as I can see.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
AkA84: Is there a way to have the "last civihr item" without referencing any extension in particular? (just wondering)

coldrunKacper:
Not much to say except the things mentioned above. Regarding the "Reports" CiviCRM top navigation - it may not be too safe to relay on it as "Reports" are just CiviCRM extension which may not be always installed.

AkA84:
Approved

PR #898, opened 2016-04-19
PCHR-938: Add Leave & Absences CiviHR permissions

davialexandre:
The following permissions were added:
- Access Leave and Absences CiviHR (Dashboard)
- Administer Leave and Absences

Here you can see the permissions available on Drupal:
captura de tela 2016-04-19 as 08 38 03


omarabuhussein:
Approved

PR #899, opened 2016-04-20
PCHR-978 - creating Upgrader function which installs scheduled job and updating enable/disable/uninstall extension functions.

coldrunKacper:
- Creating new Upgrader function which installs CiviCRM scheduled daily Job upgrading Length of Service values for each Contact.
- Including the new scheduled job into extension's enable/disable/uninstall actions.

After executing the upgrade the Scheduled Job should be visible and active on index.php?q=civicrm/admin/job&reset=1 page.


omarabuhussein:
Approved

PR #900, opened 2016-04-20
PCHR-905: Merge contracts and roles when merging contacts

omarabuhussein:
## Problem

When merging duplicate contacts , Job contracts checkbox was appear under "move related" and if the checkbox is checked then all contracts and job roles will move to the main contact. The problem is there is a role to have only one active contract at a time (PCHR-869) and merging all contracts will violate this rule .
selection_002

## Solution 

A list of all job contracts and roles now appear  allow the user select which one to keep after the merge    and which to remove , also in case there is a conflicting contracts between the two contacts then an error at the top of the page will be shown describing which contracts conflict with main contact contracts along with contracts IDs before the merge . also if a role selected but the assosiatve contract is not selected for the merge the merge will fail and an error message will appear telling the user the source of the problem.

selection_003

selection_004

selection_005

Technically speaking , this was achieved by overriding some core files by jobcontract extension since there is no way to achieve this using hooks , Here is the list of overridden files :

- CRM/Dedupe/Merger.php
- CRM/Contact/Form/Merge.php
- Templates/CRM/Contact/Form/Merge.tpl
- Templates/CRM/Contact/Form/Merge.hlp

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
@@ -23,6 +23,30 @@ class CRM_Hrjobroles_BAO_HrJobRoles extends CRM_Hrjobroles_DAO_HrJobRoles {
   }
 
   /**
+   * Return an associative array with Roles for specific contact.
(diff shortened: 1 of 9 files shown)


coldrunKacper:
Approved.

PR #901, opened 2016-04-20
PCHR-927: Create AbsenceType entity

davialexandre:
This is the basic structure of the AbsenceType entity. At this point,
it only has the create method, which can be used to create and
update Absence types. The method runs a large number of
validations to make sure only types folowing all the rules are
stored in the database.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -0,0 +1,206 @@
...
+    CRM_Core_DAO::executeQuery($query);
+  }
+
+  private static function validateParams($params) {
+    if(!empty($params['add_public_holiday_to_entitlement'])) {
+      self::validateAddPublicHolidayToEntitlement();
+    }
+
+    if (!empty('allow_request_cancelation') &&
+        !array_key_exists($params['allow_request_cancelation'], self::getRequestCancelationOptions())
```
omarabuhussein: if (!empty('allow_request_cancelation') should be if (!empty($params['allow_request_cancelation'])

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -0,0 +1,206 @@
...
+    $query = "UPDATE $tableName SET is_default = 0 WHERE is_default = 1";
+    CRM_Core_DAO::executeQuery($query);
+  }
+
+  private static function validateParams($params) {
+    if(!empty($params['add_public_holiday_to_entitlement'])) {
+      self::validateAddPublicHolidayToEntitlement();
+    }
+
+    if (!empty('allow_request_cancelation') &&
```
omarabuhussein: why if (!empty('allow_request_cancelation') ??

omarabuhussein:
i just left one simple note ... other than that , you are good to go.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -0,0 +1,37 @@
...
+     `allow_accruals_request` tinyint   DEFAULT 0 ,
+     `max_leave_accrual` int unsigned    COMMENT 'Value is the number of days that can be accrued. Null means unlimited',
+     `allow_accrue_in_the_past` tinyint   DEFAULT 0 ,
+     `accrual_expiration_duration` int unsigned    COMMENT 'An amount of accrual_expiration_unit',
+     `accrual_expiration_unit` int unsigned    COMMENT 'The unit (year, month, etc) of accrual_expiration_duration of this type default expiry',
+     `allow_carry_forward` tinyint   DEFAULT 0 ,
+     `max_number_of_days_to_carry_forward` int unsigned    ,
+     `carry_forward_expiration_duration` int unsigned    COMMENT 'An amount of carry_forward_expiration_unit',
+     `carry_forward_expiration_unit` int unsigned    COMMENT 'The unit (year, month, etc) of carry_forward_expiration_duration of this type default expiry',
+     `carry_forward_expiration_date` varchar(5)    COMMENT 'If expiration_unit + expiration_duration is not informed, this should be. Its a date in the format dd-mm',
```
coldrunKacper: I think that storing date in 'dd-mm' is not a good idea according to further sorting, etc. Maybe just use MySQL DATE type as it looks like it's still more reasonable than VARCHAR even if we don't need year value. See:
http://stackoverflow.com/questions/4998470/storing-day-and-month-without-year
and
http://stackoverflow.com/questions/9134497/mysql-datatype-to-store-month-and-year-only
But I'm not sure how we need to use this field so I'm just guessing.

davialexandre: You're right. It really doesn't look like a good idea for me too, but the other options don't look too good too. I would have to add a year to it and make sure to remember every time that it should be ignored and has no meaning at all. Not very explicit. So, since we won't need any sorting or filtering on this, I'll keep the date as a string.

PR #902, opened 2016-04-21
PCHR-982: Rename job contract import page title to 'Import Job Contracts'

omarabuhussein:
Changing Import job contracts page title from "JOB import" to "Import Job Contracts"

### Before
selection_006

### After

selection_007


davialexandre:
Approved

PR #903, opened 2016-04-21
PCHR-853: Increase z-index of ui-select dropdown list

AkA84:
*(Related to https://github.com/compucorp/civihr-tasks-assignments/pull/51*)

The `ui-select` directive can be used as a [angular-xeditable](https://vitalets.github.io/angular-xeditable/) field, meaning that the `z-index` of all the different elements ("follow link" links, action buttons, dropdown list, etc) can be in conflict with one another.

before

The dropdown list however should always be on top of all the other elements, so its `z-index` has been increased

after


kkalamarski:
Approved

PR #904, opened 2016-04-22
PCHR-869 - fixing HRJobDetails::validateDates() BAO method issue when…

coldrunKacper:
… checking for conflicting Job Contracts with no effective_end_date specified.

- Modifying SQL query for getConflictingContracts() method in case when $periodEndDate parameter is not specified.
- Adding 'effective_end_date' field to DAO class.


omarabuhussein:
Approved

coldrunKacper:
Unit Test added. It can be run by:
civix test CRM_Hrjobcontract_ValidateDatesTest
from extension main directory.

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -15,7 +15,9 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevision extends CRM_Hrjobcontract_DAO_
   public static function create($params) {
     global $user;
 
-    $params['editor_uid'] = $user->uid;
+    if (!empty($user->uid)) {
```
omarabuhussein: is this check really needed , the user cannot reach here if he is not logged-in and have the permissions needed .. and if he is logged-in then for sure he have an ID so i feel it is unnecessary.

coldrunKacper: It wasn't necessary until I've added Unit Test as when running Unit Test we don't have any "signed in" user by default.

omarabuhussein:
Approved

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/ValidateDatesTest.php
```diff
@@ -0,0 +1,147 @@
...
+    $this->assertEquals(0, $jobContract2Result['is_error']);
+    $jobDetails2Result = civicrm_api3('HRJobDetails', 'create', array(
+      'sequential' => 1,
+      'jobcontract_id' => $jobContract2Result['id'],
+      'period_start_date' => '2016-02-01',
+    ));
+    $this->assertEquals(0, $jobDetails2Result['is_error']);
+    return TRUE;
+  }
+}
```
davialexandre: I know this is already merged, but I believe these tests would be better placed at **tests/phpunit/api/v3/HRJobDetailsTest.php** since this is only testing HRJobDetails API

PR #905, opened 2016-04-22
PCHR-995 : Remove -select- option from regions and locations select boxes

omarabuhussein:
## Problem
in contacts manage access popup , the -select- option was appearing in both region and location select boxes and you can even choose it.

selection_011

selection_012

## Solution
the -select- option was removed from options list array.

selection_013


nishant-bhorodia:
Approved

PR #906, opened 2016-04-22
PCHR-903: Create Leaves/Absences list view

davialexandre:
The Leave/Absences types list is accessible only by the manager and it lists all the available types. Each type may have a different set of actions available, according to if it is reserved, default and/or enabled.

captura de tela 2016-04-22 as 12 09 26


coldrunKacper:
Approved.

PR #907, opened 2016-04-22
PCHR-904: Create the default leave types

davialexandre:
These are the default leave types that will be available right after the
installation of the extension


omarabuhussein:
Approved

PR #908, opened 2016-04-22
PCHR-934: Create the NotificationReceiver entity

davialexandre:
This entity will be used to store the contacts that will
be notified when an employee that doesn't have a
default approver requests a leave of the related type.


coldrunKacper:
Approved.

PR #909, opened 2016-04-22
PCHR-857-Job role screen ui improvements

danilo-compucorp:
Now the "action" column is hidden when the values are being edited, and the "Cost Centre" are ordered according to the order chosen in the options screen.

Hiding column:
hiding-action-column

Ordering the values:
ordering

### DIFF
--- a/com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/hr-job-roles-controller_test.js
+++ b/com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/hr-job-roles-controller_test.js
@@ -27,6 +27,92 @@ define([
             DateValidation.dateFormats.push('DD/MM/YYYY');
         }));
 
+        describe('getOptionValues', function() {
+          beforeEach(function () {
+            spyOn(HRJobRolesService, 'getOptionValues').and.returnValue($q.resolve({
+               "count":5,
+               "values":[
+                  {
+                     "id":"845",
+                     "option_group_id":"111",
+                     "label":"Senior Manager",
+                     "value":"Senior Manager",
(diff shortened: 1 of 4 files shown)


com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/hr-job-roles-controller_test.js
```diff
@@ -27,6 +27,91 @@ define([
             DateValidation.dateFormats.push('DD/MM/YYYY');
         }));
 
+        describe('getOptionValues', function() {
+
+          it('should build the "CostCentreList" array of objects containing the "weight" property', function(){
```
AkA84: Small thing: the `it` statements should be in a declarative form, so instead of `it('should do X')`, it should be `it('does X')`

com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/hr-job-roles-controller_test.js
```diff
@@ -27,6 +27,91 @@ define([
...
+            spyOn(HRJobRolesService, 'getOptionValues').and.returnValue($q.resolve({
```
AkA84: The `spyOn`, `initController()`, and `$rootScope.$digest()` are better placed in a `beforeEach()` block rather than in the `it` block where, as much as possible, only the `expect()` statements should be placed

AkA84:
Small thing: please use the `PCHR-XYZ: Subject of the commit` format for the commits subject line

AkA84:
Left a couple of comments, once the issue addressed are fixed, the PR is approved

PR #911, opened 2016-04-26
PCHR-997: new permission - manage roles and Teams

omarabuhussein:
## The Problem

On contact summary page , anybody can see "**manage roles and teams**" link in action menu 

selection_004

and if it clicked the following modal will appear to allow the user to manage that contact roles and teams :

selection_005

or even by using the link directly : <civicrm URL>/admin/contact-access-rights/manage?&cid=CONTACT_ID a similer page will opened. 

But this is not correct behavior  , only specific users should be allowed to see the screens above .

## The Solution

A new permission created named "administer roles and teams" and only users with that permission are allowed to see the action bar menu or even access the page directly from the browser. Also this permission is set to be enabled by default for "civihr_admin" and "administer" Drupal roles.

selection_003

## Related PRs

https://github.com/compucorp/civihr-employee-portal/pull/115

### DIFF
--- a/contactaccessrights/contactaccessrights.php
+++ b/contactaccessrights/contactaccessrights.php
@@ -59,6 +59,16 @@ function contactaccessrights_civicrm_disable() {
 }
 
 /**
+ * Implements hook_civicrm_permission().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_permission
+ */
+function contactaccessrights_civicrm_permission(&$permissions) {
+  $prefix = ts('CiviHRContactAccessRights') . ': ';
+  $permissions['administer roles and teams'] = $prefix . ts('Administer roles and teams');
+}
+
+/**
  * Implements hook_civicrm_upgrade().
  *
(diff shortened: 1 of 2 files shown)


contactaccessrights/contactaccessrights.php
```diff
@@ -189,11 +199,13 @@ function contactaccessrights_civicrm_aclWhereClause($type, &$tables, &$whereTabl
...
+  if (CRM_Core_Permission::check('administer roles and teams')) {
+    $actions['casework'] = array(
+      'title'  => 'Manage roles and teams',
+      'weight' => 999,
+      'ref'    => 'manage-contact-access-rights',
+      'key'    => 'manage-contact-access-rights',
+      'href'   => "/civicrm/admin/contact-access-rights/manage?"
+    );
+  }
 }
```
nishant-bhorodia: I think we need to add a new line at the end of file maybe ?

AkA84: Yes, please make sure to configure your IDE to do it automatically for you

PR #912, opened 2016-04-26
PCHR-1002: clean up local vendors

brunoqueiros:
Clean up local vendors to avoid have duplicate vendors.

e.g: `job-contract/vendor/angular-select` was using an older version than `common/ui-select`.


AkA84:
Two things:
1) You are missing the minified js file `job-contract.min.js`. Make sure to run `gulp` while you work on js or css file so the minified files are automatically updated
2) The description is lacking. Please explain exactly what you changed, what local vendors you removed, what files you have added/removed, etc. Examples are fine for ticket descriptions, not for PR descriptions

brunoqueiros:
It was removed the ui-select vendor which existed on `common/ui-select` and changed the path in the files that were using it. Also was removed two vendors `moment.min.js` `require.js` that weren't been using for any file.

AkA84:
Approved

PR #913, opened 2016-04-27
PCHR-889: Fix search for Job Role start and end dates

davialexandre:
**Problem**: The where clause for those fields was being built with the
'hrjobroles' table alias, which isn't used anymore.

**Solution**: Replace the table alias with the actual table name


nishant-bhorodia:
Approved!

PR #914, opened 2016-04-27
PCHR-857: job role screen ui improvements

danilo-compucorp:
On both "Cost Centres" and "Funding" tabs, made the "action column" visible only when editing the values, and aligned the "action button" to the left, directly under the column label.
Updated the minified js file (job-roles.min.js) so the changes will be present when the debug mode is disabled.

costcentres

funding


AkA84:
Approved

PR #916, opened 2016-04-29
PCHR-857: UI adjustments for existing data

danilo-compucorp:
For contacts with existing data, on both "Cost Centres" and "Funding" tabs, aligned the "action button" to the left directly under the column label, and ordered the "Cost Centres" options according to the configured ordering.

pchr-857-existing-data


AkA84:
Approved

PR #917, opened 2016-04-29
PCHR-1009 - Removing old Job_Summary data migration when installing the extension.

coldrunKacper:
There was still 'Job_Summary' data migrating from old HRJob extension when installing new HRJobContract. It was causing an error during buildkit installation when using current 'develop' branch. We don't use Job Summary data anymore so the data migration is now removed from the installer.


nishant-bhorodia:
Approved!

PR #918, opened 2016-05-02
PCHR-841 - Moving input field under Contract end date field instead of Contract Duration.

coldrunKacper:
"Contract end reason" filed is now moved below "Contract End Date" instead of "Contract Duration" field.


brunoqueiros:
Approved.

PR #920, opened 2016-05-03
PCHR-874: Update to newest ui-boostrap version

AkA84:
## Problem
We are currently using a very old (actually the very first) version of [ui-boostrap](https://angular-ui.github.io/bootstrap/), v.`0.12.0`, which obviously needs to be updated

## Solution
Updating it to v.`1.3.3`, latest version available to date. Of course this introduced a variety of backward compatibility breaking changes, that needed to be addressed

#### Prefixes
All the directive and services have now a `uib` (for the former) and a `$uib` (for the latter) prefixes, that needed to be added following this [migration guide](https://github.com/angular-ui/bootstrap/wiki/Migration-guide-for-prefixes)

#### `$uibModal` interface
The service had initially a `targetDomEl` property to indicate the DOM element to which the modal should be attached. The newer version has changed it to an `appendTo` property

#### CSS
The library comes with custom css that gets injected directly via javascript to the `<head>` element:
```js
angular.module('ui.bootstrap.carousel').run(function() {
    !angular.$$csp().noInlineStyle &&
    !angular.$$uibCarouselCss &&
    angular.element(document).find('head').prepend('<style type="text/css"></style>');    
    angular.$$uibCarouselCss = true;
});
```
Given that our custom style is namespaced, the aforementioned css would not get applied to our theme given it would lay outside of the namespace (the namespace is not usually applied to the `<html>` tag in our cases). To solve this problem, the css has been copied over a proper scss partials who would then be included in the theme
```sass
#{$bootstrap-namespace} {
    /* ... */
    @import "vendor/ui-bootstrap/ui-bootstrap";
    /* ... */
}
```
Additionally, our theme had already the datepicker style customized in an override partial, and the selectors of the overriding rules had to be updated to match the new classes used by the library

Unfortunately some of the extensions still don't use the common theme provided by the `org.civicrm.(bootstrap|bootstrapcivicrm|bootstrapcivihr)` extensions, so the CSS had to be, temporarily, copied in their local scss folders

#### `ngAnimate`
The library now relies on the `ngAnimate` module to animate the components, and as such it has been added to the extensions' applications

#### `prevent-animations` directive
The addition of `ngAnimate` would create a display issue with the `textAngular` directive in modals, flashing for a moment the `<textarea>` underneath.

A `prevent-animations` directive have been added as a `common/` dependency to avoid the issue, as described [here](https://github.com/fraywing/textAngular/issues/280):
```html
<div text-angular
    prevent-animations
    ta-toolbar="[['bold','italics','underline','strikeThrough','ul','ol','undo','redo','clear']]"
    ng-model="task.details">
</div>
```
The directive has also been added to the `loading` directive, given that `ngAnimate` would create a conflict with that as well (showing the spinner for longer than necessary)
```html
<div>
    <div prevent-animations class="crm_spinner" ng-show="!show">
        <span class="crm_spinner__img"></span>
    </div>
    <div ng-transclude ng-show="show"></div>
</div>
```
#### `formatDate` filter
The `formatDate` filter has now been augmented to be able to convert a string to a `Date` object. A `Date` can be passed as the `format` argument:
```js
$filter('formatDate')('22-01-2016', Date); // Fri Jan 22 2016 00:00:00 GMT+0100 (CET)
```

#### Datepicker
The Datepicker popup was the directive that proved most challenging in the migration, mainly because it now accepts **only** `Date` objects as the value of the `ng-model` property, while in the earliest version it would accept also basic date strings

##### Custom decorators
We had in place some custom decorators to augment the `datepicker-popup` (now `uib-datepicker-popup`) directive. All of them have been removed except one, the `uibDatepickerPopupDirective` decorator.

The role of this decorator, before this PR, was to enforce the `HR_settings.DATE_FORMAT` format under the hood, ignoring the format passed to the directive itself. Additionally, it would have a custom fix to prevent bad positioning of the popup when closer to the window border:
```js
attrs.datepickerPopup = HR_settings.DATE_FORMAT;

/**
 * Calculates position of the datepicker popup.
 * @param {bool} isDatepickerOpen
 */
var adjustPopupPosition = function (isDatepickerOpen) {
```
The amended version removes the `adjustPopupPosition` workaround (it's taken care by the library itself now), and provides a better handling of the date format, which can be overridden by passing the format to the directive as by original design
```js
attrs.uibDatepickerPopup = attrs.uibDatepickerPopup || HR_settings.DATE_FORMAT || 'yyyy-MM-dd';
```

##### Custom `date-input` directive
The `date-input` directive was applied to all `datepicker-popup` directive. It has now been removed, as it was conflicting with how the directive now works

##### Converting all passed datepicker's `ng-model` values to `Date` objects
All the `ng-model` values must now be `Date` objects when passed to the datepicker. This meant converting dates of tasks, assignments, job contracts, job roles, and appraisals from date strings to `Date` objects (to display them in the datepicker) and back (to save the selected date in the DB)

#### `dateFilter` decorator
In our datepicker we want to display the name of the days just by their first letter (M, T, W, etc), but the only two formats available, which are parsed via the native `dateFilter`, are either the `EEE` three letters (Mon, Tue, Wed, etc) or the `EEEE` full name (Monday, Tuesday, Wednesday, etc) formats

A custom decorator, included in the `angularDate` module, has been attached to the filter, so that it now support an additional `E` format, which returns the first letter of the day name
```js
$filter('date')('2016-05-05', 'E'); // T
```

#### `uibDatepickerPopupDirective` default values
The decorator now sets other default settings like the starting day of the week, the format of the day in the header, etc
```js
attrs.showButtonBar = attrs.showButtonBar || false;

scope.datepickerOptions = _.assign({
    formatDayHeader: 'E',
    showWeeks: false,
    startingDay: 1
}, scope.datepickerOptions);
```

#### Custom templates
Some of the datepicker templates required customization (using font-awesome icons, adding borders, etc). Instead of providing a branch new template, the `angularDate` module directly customizes the build-in templates provided by `ui-bootstrap`
```js
return angular.module("common.angularDate", /* .. */)
    .config(/* .. */)
    .run(['$templateCache', function (_$templateCache_) {
        /* .. */
        ['day', 'month', 'year'].forEach(customizeDatepickerTpl);
    }]);
```

#### Fixed gulp watch on scss files
This is not related to ui-bootstrap, but it was a quick fix to include in this PR.

In the `gulpfile` of those extensions that still rely on their own local copy of bootstrap (job roles, job contract, tasks and assignments) we are using compass to compile sass, and gulp was wrongly using `compass watch` in its `sass` task
```node
gulp.task('sass', function (done) {
    exec('compass watch', function (_, stdout, stderr) {
```
which basically meant the the sass files weren't recompiled when they changed. What we needed to use was instead `compass compile`
```node
gulp.task('sass', function (done) {
    exec('compass compile', function (_, stdout, stderr) {
```

#### Implement new prefixes in contact access rights extension
The new prefixes weren't implemented in the `uk.co.compucorp.contactaccessrights` extension

#### Use `datepicker-options` for min and max date settings
In earlier versions of the library you would use the `min-date` and `max-date` html attributes to set what is the min and max date that a given datepicker allows to choose. Those attributes are now deprecated, and those settings must come from the `datepicker-options` object

#### Update to v1.3.3
A new version, `1.3.3` has been release in the meantime

#### Use `datepicker-options` object in scope in `hrjobcontract`
Instead of an inline object passed directly to the html attribute
```html
datepicker-options="{ maxDate: dpDateStartMax }"
````
now a proper object is placed in the controller's scope
```js
$scope.datepickerOptions = {
  start: { maxDate: null },
  end: { minDate: null }
};
```

#### Changing how the datepicker popup is open
Before, the datepicker popup would be opened only when the user would click on the calendar icon, not on the input field itself.

before

To make things more consistent with how datepicker fields work in CiviCRM, now the datepicker automatically opens when the user clicks on the input field, while the calendar icon is just a visual indicator that the field uses a datepicker.

after

Additionally, the style of the datepicker field has been standardized across the extensions


hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.js
```diff
@@ -53,6 +54,9 @@ define([
             angular.copy(entity,$scope.entity);
             angular.copy(files,$scope.files);
 
+            $scope.entity.details.period_start_date = convertToDateObject($scope.entity.details.period_start_date);
```
kkalamarski: Probably changes to 'filterDate' filter were introduced later, but it may be worth changing.
```javascript
$scope.entity.details.period_start_date = $filter('formatDate')($scope.entity.details.period_start_date, Date);
```
Or is there any other reason you don't use this filter here?

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -998,6 +997,18 @@ define([
...
               return me.contractsData[contractId];
             }
+
+            /**
+             * Parse dates so they can be correctly read by server.
+             *
+             * @param {string|Date} date
+             * @returns {string|null}
+             */
+            function convertDateToServerFormat(date) {
```
kkalamarski: Do you think this code may be useful in different CiviHR angular apps? 
Maybe it'd be good to have it in common dependency...

kkalamarski:
Looks good. If above is changed or reason is given why shouldn't we change this, PR is approved.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.js
AkA84: Not sure I understand the comment correctly, but the `$filter` is used in the `convertToDateObject` function:
```js
function convertToDateObject(dateString) {
    var dateObj = $filter('formatDate')(dateString, Date);

    return dateObj !== 'Unspecified' ? dateObj : dateString;
}
```
the reason for this was because the filter still returns the very specific `Unspecified` string (instead of, say, `undefined`). I didn't want to change the return value of the filter itself and risking to introduce yet more breaking changes, so I'm using the `convertToDateObject` to make sure that the invalid (usually `null`) date is returned in case the filter returns `Unspecified`. The same logic is applied in some other components in other extensions

kkalamarski: Yeah, I see now. I somehow missed it (maybe hadn't enough coffee...) :(
Approved :)

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
AkA84: Tbh I wanted `$filter('formatDate')` to return `null` or `undefined` when the date wasn't valid or set, but apparently some applications rely on the specific `Unspecified` string. This function was a simple way to make things still work without changing the response from the filter (and then introducing some more breaking changes), but if there was some work to be put to get a more common logic, it wouldn't be to reuse this function, but to amend the filter to *not* return `Unspecified`, and then refactor the places where we rely on that string. Given the scope of this PR I had to give up on that idea (for now!)

kkalamarski:
Newest changes Approved

kkalamarski:
approved

kkalamarski:
Just one idea:
could you organise description with dates? It would be easier to get glimpse of what has been done in certain period of time. For instance:

### Updated on [date] ### 
``` description for changes you made. ```

or any other markup to divide description into reasonable blocks.

This PR is being worked on for quite long time, and I think it will be easier to track changes looking on description instead of diving into code.

AkA84:
Yep, will do, makes sense!

kkalamarski:
Approved

PR #922, opened 2016-05-05
PCHR-892: fix job contract import issues

omarabuhussein:
Fix job contracts import validation problems reported by QA  which are as well as changing in some names in import match fields select list which are :

- Health Insurance fields marked incorrectly as Healthcare
- removing Estimated Auto Pay option in import list
- Id should be ID everywhere
- Unable to import min fields ( when using contact email )
- Fields other than mandatory fields were left blank. However, this generated error and import did not go through
- Reason for end of contract is not imported
- Unable to import job leaves and pay deduction/benefits

Also I've added unit tests and refactored part of  code.


coldrunKacper:
Approved.

PR #924, opened 2016-05-05
PCHR-921: Remove warnings on Contact page

AkA84:
#### Problem
We have this error appearing on the civihr pages:
```
Notice: Undefined offset: 1 in CRM_Core_BAO_Navigation::createNavigation() (line 656 of /var/www/civicrm-buildkit/build/hr16/sites/all/modules/civicrm/CRM/Core/BAO/Navigation.php).
```
This started to happen after the work PCHR-343 was merged, which changes the CiviHR homepage from `civicrm/dashboard?reset=1` to `civicrm/tasksassignments/dashboard#/tasks` (see [this file](https://github.com/compucorp/civihr-tasks-assignments/blob/develop/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/DashboardSwitcher.php)), basically making T&A the CiviHR dashboard

The hrnavigation code was still tring to split the url using the `?` querystring delimiter, but given that the T&A url doesn't have one, the offset notice was thrown

#### Solution
Simply checking if the url has a querystring before splitting it by `?` solves the issue


omarabuhussein:
approved

PR #925, opened 2016-05-06
PCHR-769: design for permissions pop up and button

danilo-compucorp:
Replacing the partial implementation of the Permission modal, using AngularJS. 
The menu option for that:
pchr-769-button-before

got replaced by the new button:
pchr-769-button2

The old modal:
pchr-769-modal-before

got replaced by the new one:
pchr-769-modal

Also in this PR, I've implemented and tested a refactored version of the API services, using angular-resource to map the calls to the server.


contactaccessrights/contactaccessrights.php
```diff
@@ -180,32 +180,20 @@ function contactaccessrights_civicrm_aclWhereClause($type, &$tables, &$whereTabl
...
-      'key'    => 'manage-contact-access-rights',
-      'href'   => "/civicrm/admin/contact-access-rights/manage?"
-    );
+function contactaccessrights_civicrm_summary( $contactID, &$content, &$contentPlacement ) {
+  CRM_Core_Resources::singleton()->addVars('summaryPage', array(
+    'contactID' => $contactID
+  ));
+}
+
+define('CONTACT_ACCESS_RIGHTS', 'uk.co.compucorp.contactaccessrights');
```
AkA84: Is it necessary to have it defined as a (global?) constant? If you are using it just inside `contactaccessrights_civicrm_pageRun`, why not a simple variable instead?

```php
function contactaccessrights_civicrm_pageRun($page) {
    $extName = 'uk.co.compucorp.contactaccessrights';
    // ...
}
```

contactaccessrights/contactaccessrights.php
```diff
@@ -180,32 +180,20 @@ function contactaccessrights_civicrm_aclWhereClause($type, &$tables, &$whereTabl
...
-    $actions['casework'] = array(
-      'title'  => 'Manage roles and teams',
-      'weight' => 999,
-      'ref'    => 'manage-contact-access-rights',
-      'key'    => 'manage-contact-access-rights',
-      'href'   => "/civicrm/admin/contact-access-rights/manage?"
-    );
+function contactaccessrights_civicrm_summary( $contactID, &$content, &$contentPlacement ) {
+  CRM_Core_Resources::singleton()->addVars('summaryPage', array(
+    'contactID' => $contactID
```
AkA84: So you are providing the contact id server-side? Was there any other way to do it "properly", by having the angular app fetching the contact id via some service or something like that? (in the contact summary app the contact id is [fetched](https://github.com/civicrm/civihr/blob/develop/contactsummary/js/src/contact-summary/modules/settings.js#L8) via the url, but that's pretty suboptimatl)

contactaccessrights/js/src/access-rights.js
```diff
@@ -0,0 +1,22 @@
...
+  var extPath = CRM.vars.contactAccessRights.baseURL + '/js/src/access-rights';
+  require.config({
+    urlArgs: 'bust=' + (new Date()).getTime(),
+    paths: {
+      'access-rights': extPath
+    }
+  });
+
+  require(['access-rights/app'], function (angular) {
+    var actions = angular.element('.crm-actions-ribbon ul#actions');
```
AkA84: Please encapsulate all this DOM manipulation in a function (or even better, a separate RequireJS module) where you document it and explain what and why you do this

contactaccessrights/js/src/access-rights.js
```diff
@@ -0,0 +1,22 @@
...
+    var app = angular.element('<li id="access-rights"><div id="bootstrap-theme"> \
+        <div ng-controller="AccessRightsCtrl as $ctrl"> \
```
AkA84: Couldn't you put the bootstrap namespace on this element and ditch the wrapper, to keep the markup at a minimum?

contactaccessrights/js/src/access-rights.js
```diff
@@ -0,0 +1,22 @@
...
+          <a href class="btn btn-secondary-outline text-uppercase pull-right" ng-click="$ctrl.openModal()"> \
```
AkA84: Just curious, is the button supposed to look different than the other (old-style) buttons in the `.crm-actions-ribbon`?

danilo-compucorp: Well that's a good question. I didn't see any examples regarding the button, aside from the wireframe on which the "Manage roles and teams" button looks just like the modal buttons. I'll ask Natsuki about it

contactaccessrights/js/src/access-rights/app.js
```diff
@@ -0,0 +1,30 @@
...
+  'access-rights/controllers/access-rights-ctrl',
+  'access-rights/controllers/access-rights-modal-ctrl',
+  'access-rights/models/region',
+  'access-rights/models/location',
+  'access-rights/models/right'
+], function (angular) {
+  angular.module('access-rights', [
+      'access-rights.controllers',
+      'access-rights.models',
+      'ui.select',
```
AkA84: Just nitpicking here, but for consistency's sake list the vendor/common modules first
```js
'ui.select',
'xeditable-civi',
'access-rights.controllers',
// etc
```

contactaccessrights/js/src/access-rights/app.js
```diff
@@ -0,0 +1,30 @@
...
+      'xeditable-civi'
+    ])
+    .run(['$log', 'editableOptions', 'editableThemes',
+      function ($log, editableOptions, editableThemes) {
+        $log.debug('app.run');
+        editableOptions.theme = 'bs3';
+      }
+    ]).config(['$locationProvider', function ($locationProvider) {
```
AkA84: nitpicking again, be consistent with the code style
```js
])
.run([], function () {

})
.config([], function () {

});
```

contactaccessrights/js/src/access-rights/controllers/access-rights-ctrl.js
```diff
@@ -0,0 +1,22 @@
+define(['access-rights/modules/controllers'], function (controllers) {
```
AkA84: Still for consistency, define the modules like this (even if it's just with one dependency)
```js
define([
    'dependency 1',
    'dependency 2',
    // ...
    'dependency 3'
], function (1, 2, 3) {

});
```

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
+define(['access-rights/modules/controllers'], function (controllers) {
+  'use strict';
+
+  controllers.controller('AccessRightsModalCtrl', ['Region', 'Location', 'Right', '$q', '$modalInstance',
```
AkA84:
```js
controllers.controller('AccessRightsModalCtrl', [
    'dependency 1', 'dependency 2', 'dependency 3',
    function (1, 2, 3) {

    }
]);
```
(it's future-friendly if we'll need to break line when we have many injected dependencies)

AkA84: nitpicking n.3: please list first the built-in, then the vendor, then the common, finally the local dependencies
```js
['$q', '$modalInstance', 'Location', 'Region']
```

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
...
+          });
+        var promises = [];
+        if (newEntityIds.length > 0)
+          promises.push(fnSave(newEntityIds));
+        if (removedRightIds.length > 0)
+          promises.push(Right.deleteByIds(removedRightIds));
+        return $q.all(promises);
+      }
+
+      Region.getAll().then(function (regions) {
```
AkA84: Wrap all this initialization code in a `init()` function

danilo-compucorp: Curious about that. This is a controller constructor, so the initialization code is expected to be here. Besides, why define a function that nobody else is ever going to call, and immediately call it after the definition?

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
...
+        vm.availableData.regions = regions;
+        return Right.getRegions();
+      }).then(function (regionRights) {
+        vm.originalData.regions = regionRights;
+        vm.selectedData.regions = regionRights.map(function (regionRight) {
```
AkA84: Why the difference in data structure between `originalData` and `selectedData`? One is the collection returned by the `Right` service, the other is an array of ids of said collection. Cannot we have it standardized, one way or the other?

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
AkA84: There is no benefit per se , other than better code organization. In my experience, non-wrapped initialization code tends to get lost amidst the rest of scope and private methods, especially when the controller grows in size over time.

Wrapping the init code in a function makes the controller easier to read and more maintainable ([example](https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.appraisals/js/src/appraisals/controllers/appraisals-dashboard-ctrl.js))

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
...
+          return regionRight.entity_id;
+        });
+      });
+
+      Location.getAll().then(function (locations) {
```
AkA84: Can the user begin to use the modal while the locations and regions are being fetched from the server?

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
danilo-compucorp: To be able to delete the Rights, we need to use the "id" property. For checking the existing ones, and to save new ones, we need the "entity_id" property. So for the "originalData" I need to keep the whole objects.

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      /**
+       * Saves the new locations, and deletes the removed ones
+       *
+       * @return {Promise}
+       */
+      function persistLocations() {
+        return persistValues(vm.originalData.locations, vm.selectedData.locations,
+          Right.saveLocations.bind(Right));
```
AkA84: Given that the only difference in `persistValues` is whether you are dealing with regions or locations, why not something like:

```js
/**
 * Saves the new values, and deletes the removed ones
 *
 * @param  {string} type - either 'regions' or 'locations'
 * @return {Promise}               The result of all promises
 */
function persistValues(type) {
  var originalData = vm.originalData[type];
  var selectedData = vm.selectedData[type];

  // ....

  if (newEntityIds.length > 0)
    promises.push(Right['save' + _.upperFirst(type)](newEntityIds));

  if (removedRightIds.length > 0)
    promises.push(Right.deleteByIds(removedRightIds));

  return $q.all(promises);
}


// Usage
/**
 * Saves data and closes the modal
 */
vm.submit = function () {
  // persistRegions and persistLocations are not needed anymore
  $q.all([persistValues('regions'), persistValues('locations')])
  // ...
};
```
this way the function's signature is simplified, you can scrap two private functions and you can avoid passing a binded method as a callback

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+        var newEntityIds = _.difference(selectedData, originalEntityIds);
+        var removedRightIds = _.difference(originalEntityIds, selectedData)
+          .map(function (entityId) {
+            return _.find(originalData, function (i) {
+              return i.entity_id === entityId;
+            }).id;
+          });
+        var promises = [];
+        if (newEntityIds.length > 0)
```
AkA84: Always use brackets
```js
if (newEntityIds.length > 0) {
    promises.push(fnSave(newEntityIds));
}

if (removedRightIds.length > 0) {
    promises.push(Right.deleteByIds(removedRightIds));
}

File: contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js [hunk 15]
```diff
@@ -0,0 +1,113 @@
...
+          promises.push(fnSave(newEntityIds));
+        if (removedRightIds.length > 0)
+          promises.push(Right.deleteByIds(removedRightIds));
+        return $q.all(promises);
```
AkA84: Let the code breathe a little, it's easier to read if you leave some blank lines between, say, the initialization part, the actual logic, and the return value
```js
  // ...
  var promises = [];

  if (newEntityIds.length > 0)
    promises.push(fnSave(newEntityIds));
  if (removedRightIds.length > 0)
    promises.push(Right.deleteByIds(removedRightIds));

  return $q.all(promises);
}
``

File: contactaccessrights/js/src/access-rights/models/location.js [hunk 16]
```diff
@@ -0,0 +1,15 @@
+define([
+  'access-rights/modules/models',
+  'common/services/api/api-builder'
+], function (models) {
+  'use strict';
```
AkA84: _(Applies everywhere)_

Leave an empty line between `'use strict'` and the actual code

contactaccessrights/js/src/access-rights/modules/controllers.js
```diff
@@ -0,0 +1,6 @@
+define([
+  'common/angular',
+  'common/angularBootstrap'
+], function (angular) {
+  return angular.module('access-rights.controllers', ['ui.bootstrap']);
```
AkA84: `'use strict'`

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
danilo-compucorp: I think we need a styleguide wiki page ;-)

contactaccessrights/js/test/controllers/access-rights-ctrl_test.js
```diff
@@ -0,0 +1,33 @@
...
+  'access-rights/controllers/access-rights-ctrl'
+], function () {
+  'use strict';
+
+  describe('AccessRightsCtrl', function () {
+    var ctrl, modalSpy;
+
+    beforeEach(module('access-rights.controllers'));
+    beforeEach(inject(function (_$controller_, _$rootScope_) {
+      CRM = {
```
AkA84: You could use the same solution used in [appraisals](https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.appraisals/js/test/globals.js) for these global vars

contactaccessrights/js/test/controllers/access-rights-ctrl_test.js
```diff
@@ -0,0 +1,33 @@
...
+      modalSpy = jasmine.createSpyObj('modalSpy', ['open']);
+      ctrl = _$controller_('AccessRightsCtrl', {
+        $scope: _$rootScope_.$new(),
+        $modal: modalSpy
+      });
+    }));
+
+    describe('openModal', function () {
+      it('opens the modal', function () {
+        ctrl.openModal();
```
AkA84: if possible keep all the "set up" code in the `beforeEach` function, so that the `it` block only has the `expect` statement(s)

```js
describe('openModal', function ( ) {
  beforeEach(function () {
    ctrl.openModal();
  });

  it('opens the modal', function ( ) {
    expect(modalSpy.open).toHaveBeenCalled();
  });
});
```

contactaccessrights/js/test/controllers/access-rights-modal-ctrl_test.js
```diff
@@ -0,0 +1,179 @@
...
+        vars: {
+          summaryPage: {
+            summaryPage: '1'
+          }
+        }
+      };
+      $scope = _$rootScope_.$new();
+      $q = _$q_;
+
+      function getEntityMock(sequential, isRegion) {
```
AkA84: This `beforeEach` block is too long (and there is a function definition in the middle of the implementation that kinda disrupts the flow), i think you could refactor to isolate this spies initialization logic

```js
beforeEach(function () {
  $scope = _$rootScope_.$new();
  $q = _$q_;

  initSpies();

  ctrl = _$controller_('AccessRightsModalCtrl', {});
  $scope.$digest();
}));

/**
 *
 *
 */
function initSpies($q) {
  modalInstanceSpy = jasmine.createSpyObj('modalInstanceSpy', ['dismiss']);
  regionSpy = jasmine.createSpyObj('regionSpy', ['getAll']);
  locationSpy = jasmine.createSpyObj('locationSpy', ['getAll']);
  rightSpy = jasmine.createSpyObj('rightSpy', ['getRegions', 'getLocations',
    'saveRegions', 'saveLocations', 'deleteByIds'
  ]);

  [regionSpy.getAll, locationSpy.getAll].forEach(function (fn, idx) {});
  [rightSpy.getRegions, rightSpy.getLocations].forEach(function (fn, idx) {});
  [rightSpy.saveRegions, rightSpy.saveLocations].forEach(function (fn) {});
}

/**
 *
 *
 */
function getEntityMock(sequential, isRegion) {
  var difference = isRegion ? 0 : 10;
  return {
    id: sequential + difference,
    entity_id: sequential * 100 + difference
  };
}
```

contactaccessrights/js/test/controllers/access-rights-modal-ctrl_test.js
```diff
@@ -0,0 +1,179 @@
...
+    describe('cancel', function () {
+      it('closes the modal instance', function () {
+        ctrl.cancel();
+        expect(modalInstanceSpy.dismiss).toHaveBeenCalled();
+      });
+    });
+
+    describe('submit', function () {
+
+      describe('new Locations', function () {
```
AkA84: _(Applies everywhere)_

Let's write more english-like tests, which also allows a better break up of the single tests into logic blocks where you can use more effectively the `beforeEach` blocks. For example:
```js
describe('new Locations', function () {
  // ...

  describe('when there are no errors', function () {
    beforeEach(function () {
      ctrl.submit();
    });

    it('saves the new Locations', function () {
      expect(/* .. */);
    });
  })

  describe('when there are errors', function () {
    beforeEach(function () {
      rightSpy.saveLocations.and.returnValue($q.reject());
      ctrl.submit();
      $scope.$digest();
    });

    it('sets the error message', function () {
      expect(/* .. */);
    });
  });
});
```

contactaccessrights/scss/access-rights.scss
```diff
@@ -0,0 +1,10 @@
+li#access-rights {
```
AkA84: Please document why the need for these `!important`s

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
danilo-compucorp: I think that when the controller gets too big, it's a signal that it should be refactored and its logic should be moved into services, making it as small as possible.

AkA84: I agree, but that doesn't diminish the benefit of encapsulating the initialization code

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
```diff
@@ -0,0 +1,116 @@
...
+          promises.push(Right.deleteByIds(removedRightIds));
+        }
+
+        return $q.all(promises);
+      }
+
+      /**
+       * Load the API data
+       */
+      $q.all([Region.getAll()
```
AkA84: There is a lot of repetition here, you are basically doing the same thing with both promises' results.
What about something like (not sure it's the optimal way, but you get the idea)
```js
$q.all([
  Region.getAll(),
  Location.getAll()
])
.then(function (values) {
  return { regions: values[0], locations: values[1] };
})
.then(function (values) {
  return _.map(values, function (value, key) {
    vm.availableData[key] = value.values;
    return Right['get' + _.capitalize(key)];
  });
})
.then(function (values) {
  return { regions: values[0], locations: values[1] };
})
.then(function (values) {
  _.forEach(function (value, key) {
    vm.originalData[key] = value.values;
    vm.selectedData[key] = value.values.map(function (entity) {
      return entity.entity_id;
    });
  });
});
```

contactaccessrights/js/src/access-rights/models/location.js
```diff
@@ -0,0 +1,14 @@
+define([
+  'access-rights/modules/models',
+  'access-rights/services/location',
+  'common/services/api',
```
AkA84: Always first define the `common/` dependencies, then the local ones

contactaccessrights/js/src/access-rights/models/location.js
```diff
@@ -0,0 +1,14 @@
...
+  'common/models/model'
+], function (models) {
+  'use strict';
+
+  models.factory('Location', ['Model', 'api.location', function (Model, api) {
+    return Model.extend({
+      getAll: api.query.bind(api)
```
AkA84: `api.location.query` gets the `OptionValues` of a given group, couldn't have you used [`OptionGroup.valuesOf`](https://github.com/civicrm/civihr/blob/develop/org.civicrm.reqangular/src/common/models/option-group.js)?
```js
getAll: function() {
  return OptionGroup.valuesOf('hrjc_location');
}
```
or it's not that useful in your case?
_(same goes for the `Region` model)_

contactaccessrights/js/src/access-rights/services/location.js
```diff
@@ -0,0 +1,19 @@
...
+], function (models) {
+  'use strict';
+
+  models.factory('api.location', ['api', function (api) {
+    return api.extend({
+      query: function () {
+        return this.sendGET('OptionValue', 'get', {
+          'option_group_name': 'hrjc_location',
+          'json': {
+            sequential: 1
```
AkA84: `sequential: 1` is already defined in [`api`](https://github.com/civicrm/civihr/blob/develop/org.civicrm.reqangular/src/common/services/api.js#L116), I don't think you need to pass `json` at all

contactaccessrights/js/src/access-rights/services/location.js
```diff
@@ -0,0 +1,19 @@
...
+          }
+        }, false);
```
AkA84: Why the need for not caching the results?

contactaccessrights/js/src/access-rights/services/location.js
```diff
@@ -0,0 +1,19 @@
+define([
```
AkA84: Put the apis in `access-rights/services/api` (or maybe `accesss-rights/apis`? not sure :/)

contactaccessrights/js/src/access-rights/services/right.js
```diff
@@ -0,0 +1,46 @@
...
+], function (models) {
+  'use strict';
+
+  models.factory('api.right', ['api', '$q', '$location', function (api, $q, $location) {
+    var entityName = 'Rights';
+    var additionalParams = {
+      'contact_id': $location.search().cid
+    };
+
+    var methods = {
```
AkA84: Matter of stylistic preferences, but what about a more terse
```js
return api.extend({
    getLocations:
    getRegions:
    // etc...
});
```

contactaccessrights/js/src/access-rights/controllers/access-rights-modal-ctrl.js
AkA84: And wrap all of this in a `init()` function! 😄

AkA84:
What's left to do:
* Review/address the new comments
* Remove the components and tests for the proposed API solution (as suggested please store them in a branch)
* Updates the tests
* Document methods/functions

contactaccessrights/js/src/access-rights/models/location.js
danilo-compucorp: As we'd talked about keeping each API on its own extension, I preferred to do it like that.

contactaccessrights/js/src/access-rights/models/location.js
danilo-compucorp: Why not first define the ones we're going to use a reference to, then the ones we don't need a reference to?

AkA84:
Approved 👍

PR #926, opened 2016-05-06
PCHR-891: Fix Job Contracts fields configuration for export

davialexandre:
The CRM_Hrjobcontract_BAO_Query class relies on the 'where' property
of the fields to build the field list of the SELECT clause of the query used
to export contacts. Some of the Contracts fields didn't have this property
and when one of it was selected to be included in the exported
file, the built SQL query ended up incomplete giving a DB Syntax error.


nishant-bhorodia:
Approved!

PR #927, opened 2016-05-06
PCHR-891: Remove contracts legacy fields

davialexandre:
These fields are legacy and will be completely removed from hrjobcontract on a
future PR. For now, we just don't allow they to be searched or included in a export.


omarabuhussein:
Looks good.

PR #928, opened 2016-05-09
PCHR-985: calculate and set contract auto-populated fields during import

omarabuhussein:
When creating a new contract from the front end , the following fields are auto populated :

Hours - Actual Hours (Unit)
Hours - FTE Denominator Equivalence
Hours - FTE Numerator Equivalence
Hours - FTE Equivalence
Pay - Annual pay estimate before benefits and deductions
Pay - Gross Pay per Cycle
Pay - Net pay per cycle

This PR implement the same logic when importing contracts through the importer as well as removing these fields from the "**match fields**" page select boxes.

Here is how these fields are populated :

===========================================
**Hours Tab :**
1- Actual Hours (Unit) : this value is taken from **Location/Standard hours**.
2- FTE Equivalence : Hours Amount / Actual Hours Amount .  where **Hours Amount** is the one taken from the imported file and **Actual Hours Amount** is taken from **Location/Standard hours**.
3- FTE Numerator Equivalence : the Numerator of the simplest form for (Hours Amount / Actual Hours Amount )
4- FTE Denominator Equivalence : the Denominator of the simplest form for (Hours Amount / Actual Hours Amount ).

Note : I've used a special function that calculate the simplest fractions from decimal value of (**FTE Equivalence**).

===========================================

**Pay Tab :**
1- Annual pay estimate before benefits and deductions : (pay amount  * multiplication factor) where **pay amount** is taken from the imported file and **multiplication factor** value is based on imported Pay Unit as the following :
a- Hour = 2080
b- Day = 260
c- Week = 52
d- Month = 12
e- Year = 1
f- not specified (default) = 0

2- Gross Pay per Cycle : (Annual pay estimate before benefits and deductions / division factor ) where 
**Annual pay estimate before benefits and deductions** is the value calculated above and division factor is based on imported pay cycle value as the following :
a- weekly = 52
b- monthly = 12
c- not specified  (default) = 1

3- Net pay per cycle : (Gross Pay per Cycle + (total benefits - total deductions) ). Where **Gross Pay per Cycle** is the value calculated above.

- total benefits are calculated as following : sum all the benefit amounts and then divide the sum on (division factor) which is the same as the one above . if the amount is percent then convert it to absolute using :( (percent amount/100) * Annual pay estimate before benefits and deductions).

===========================================

**Note** : All values are rounded to 2 decimal precision (that's how fields are populated when creating new contract from front end side)..


hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php
```diff
@@ -755,4 +750,134 @@ private function validateBenefitsAndDeductions($value_type, $value)  {
     }
     return array('value'=>$outputArray, 'error_message'=>$errorMessage);
   }
+
+  private function setAutoPopulatedFields()  {
```
coldrunKacper: Maybe split these IFs logic into separate private/protected methods which would set $this->_params? The method would be shorter and easier to read.

coldrunKacper:
Missing functions comments.

coldrunKacper:
Approved.

PR #929, opened 2016-05-10
PCHR-945: Create the default Work Pattern

davialexandre:
The pattern is created during the extension's installation.

The small change in WorkDayTest.php is just a removal of
a small debug code, which didn't deserve a commit for itself.


omarabuhussein:
:panda_face: :+1:

PR #931, opened 2016-05-11
PCHR-769: Fixing production version issues

danilo-compucorp:
This PR intents mainly to solve the issues that appear when the debug options in CiviCRM are turned off.
Apparently, when the debug options are off, the ajax requests have to contain a "HTTP_X_REQUESTED_WITH" header equal to "XMLHttpRequest", otherwise it will be seen as "CSRF suspected".
Also, this PR inserts the default "loader icon" when the modal data is loading.


contactaccessrights/js/src/access-rights/app.js
```diff
@@ -1,7 +1,7 @@
 define([
   'common/angular',
   'common/ui-select',
-  'common/directives/loading',
```
AkA84: You are not using this anymore?

AkA84:
PR approved, just make sure to update the PR description before merging

PR #933, opened 2016-05-11
PCHR-1016: Resolve conflict between hrcase and T and A

nishant-bhorodia:
1. Detection of activity being created by tasks and assignments or not.
2. Passing component_id in api call so it returns single value only.


coldrunKacper:
Approved.

PR #936, opened 2016-05-11
PCHR-1036:Removing the "Follow Link" button everywhere

danilo-compucorp:
This PR removes the "Follow Link" button feature from the "editable-ui" directives.
The old behavior was:
pchr-1036-before

And the new behavior is:
pchr-1036-after


AkA84:
Approved

PR #937, opened 2016-05-12
PCHR-1043 - missing CRM_Case_BAO_Case::addCaseToContact() method

coldrunKacper:
Using CRM_Case_BAO_CaseContact class if CRM_Case_BAO_Case::addCaseToContact method is not available.

CRM_Case_BAO_CaseContact is a new class which comes in CiviCRM 4.7 while CRM_Case_BAO_Case::addCaseToContact is CiviCRM 4.5 / 4.6 method which we were originally using when developing the extensions.


omarabuhussein:
approved

PR #939, opened 2016-05-17
PCHR-1068: Fix import job role issues 4.7 migration

omarabuhussein:
## Problem
After migrating to civicrm 4.7 , The job roles page is no longer opening and through an HTTP 500 error page . The problem was in this file :

**com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Form/DataSource.php**

Specifically in this line :
 
**CRM_Core_Config_Defaults::formatUnitSize(ini_get('post_max_size'), TRUE);** 

But since **CRM_Core_Config_Defaults** class is no longer exist in 4.7 the code was failing and the error page appear. 

## Solution

I modified  the DataSource.php Form to extend the core class CRM_Import_Form_DataSource instead of CRM_Core_Form , Where the implementation is a little bit diffrent and using :

 **CRM_Utils_Number::formatUnitSize(ini_get('post_max_size'), TRUE);** 

Instead of **CRM_Core_Config_Defaults::formatUnitSize(ini_get('post_max_size'), TRUE);**  .

I also done some refactoring so the code adhere to 4.7 changes.

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Controller.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Controller.php
@@ -1,9 +1,9 @@
 <?php
 /*
  +--------------------------------------------------------------------+
- | CiviCRM version 4.5                                                |
+ | CiviCRM version 4.7                                                |
  +--------------------------------------------------------------------+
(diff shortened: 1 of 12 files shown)


coldrunKacper:
Approved.

PR #940, opened 2016-05-18
PCHR-1067: Fix contracts import issues after 4.7 migration

omarabuhussein:
## Problem
After migrating to civicrm 4.7 , The job contacts import page is no longer opening and   this exception is thrown 

`Cannot read unrecognized property CRM_Core_Config::$maxImportFileSize.`

The problem was in this file :

**hrjobontract/CRM/Hrjobcontract/Import/Form/DataSourceBaseClass.php**

Specifically in these lines :
 
`    if ($config->maxImportFileSize >= 8388608) {
      $uploadFileSize = 8388608;
    }
    else {
      $uploadFileSize = $config->maxImportFileSize;
    }`

But since **CRM_Core_Config** is no longer have this property `maxImportFileSize` , The error is thrown.

## Solution

I modified  the DataSourceBaseClass.php class to extend the core class CRM_Import_Form_DataSource instead of CRM_Core_Form , Where the implementation is a little bit different and using this to calculate $uploadFileSize :

    `$uploadFileSize = CRM_Utils_Number::formatUnitSize($config->maxFileSize . 'm', TRUE);
    $uploadSize = round(($uploadFileSize / (1024 * 1024)), 2);`

I also done some refactoring so the code adhere to 4.7 changes (as much as possible). And Also combined  DataSourceBaseClass.php  into DataSource.php  since I don't see a point from having it.


hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/DataSource.tpl
```diff
@@ -30,10 +30,11 @@
  {* WizardHeader.tpl provides visual display of steps thru the wizard as well as title for current step *}
  {include file="CRM/common/WizardHeader.tpl"}
 
- <div id="help">
-    {ts}The API Import Wizard allows you to easily upload data against any API create method from other applications into CiviCRM.{/ts}
-    {ts}Files to be imported must be in the 'comma-separated-values' format (CSV) and must contain data needed to match the data to an existing record in your CiviCRM database.{/ts} {help id='upload'}
- </div>
+  <div class="help">
+    <p>
+      {ts}The Job Contracts Import Wizard allows you to easily upload Job Contracts from other applications into CiviHR. Contacts must already exist in your CiviHR database prior to importing Job Contracts.{/ts}
```
davialexandre: Why the paragraph about the file format was removed? Was it removed in 4.7?

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/MapField.tpl
```diff
@@ -28,13 +28,13 @@
...
 {* WizardHeader.tpl provides visual display of steps thru the wizard as well as title for current step *}
 {include file="CRM/common/WizardHeader.tpl"}
-<div id="help">
-    <p>{ts}Review the values shown below from the first 2 rows of your import file and select the matching CiviCRM database fields from the drop-down lists in the right-hand column. Select '- do not import -' for any columns in the import file that you want ignored.{/ts}</p>
+  <div class="help">
+    <p>{ts}Review the values shown below from the first 2 rows of your import file and select the matching CiviHR database fields from the drop-down lists in the right-hand column. Select '- do not import -' for any columns in the import file that you want ignored.{/ts}</p>
     {if $savedMapping}
     <p>{ts}Click 'Load Saved Field Mapping' if data has been previously imported from the same source. You can then select the saved import mapping setup and load it automatically.{/ts}<p>
-    {/if}
+      {/if}
```
davialexandre: Please remove the spaces to make indentation of the closing {if} match the indentation of the opening {if}

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/MapTable.tpl
```diff
@@ -85,13 +85,12 @@
...
-            {else}
-              $('#saveDetails').hide();
-            {/if}
-          });
+          {if $mappingDetailsError }
+          cj('#saveDetails').show();
+          {else}
+          cj('#saveDetails').hide();
+          {/if}
+
```
davialexandre: Fix the indentation here too, please

hrjobcontract/CRM/Hrjobcontract/Import/Form/DataSource.php
```diff
@@ -36,24 +36,74 @@
...
   /**
-  * Include duplicate options
-  */
+   * Include duplicate options
+   */
   protected $isDuplicateOptions = FALSE;
 
-    /**
+
+  protected $_entity = array('HRJobContract', 'HRJobContractRevision', 'HRJobDetails', 'HRJobPay', 'HRJobHealth', 'HRJobPension', 'HRJobHour', 'HRJobLeave');
```
davialexandre: Where are $_enableContactOptions, $isDuplicateOptions and $_entity being used? I checked both CRM_Import_Form_DataSource and CRM_Core_Form (from civicrm-core  4.7.7 branch) and it seems they are not inherited from them. I also checked if this class is extended by another and didn't find any. So, since it seems they are declared here and not used anywhere else, maybe we don't need them? Am I missing something?

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/Preview.tpl
```diff
@@ -91,7 +91,7 @@
 
 
  {* Table for mapping preview *}
- {include file="CRM/Contact/Import/Form/MapTable.tpl}
+ {include file="CRM/Hrjobcontract/Import/Form/MapTable.tpl"}
```
davialexandre: Why you needed to change this?

omarabuhussein: This is how it should be ... This file should be taken from the extension template file . (but anyway I  don't really think it does matter  because they are probably similar )

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/DataSource.tpl
omarabuhussein: It is just replaced with this :

**<div class="help">
 <p>
 {ts}The Job Contracts Import ....... etc**

To have more consistent information paragraph across import pages.

davialexandre: That is what I thought, but with the change the text doesn't include the information stating the required file format. I think it's important to have this information here.

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/Preview.tpl
davialexandre: So, basically, it means this was wrong/"not working" before?

omarabuhussein: It was working ... but it wasn't  the right way ...

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/DataSource.tpl
omarabuhussein: This information is already mentioned under the "choose file" button :

`File format must be comma-separated-values (CSV).`

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/MapTable.tpl
```diff
@@ -85,13 +85,12 @@
...
-            {if $mappingDetailsError }
-              $('#saveDetails').show();
-            {else}
-              $('#saveDetails').hide();
-            {/if}
-          });
+          {if $mappingDetailsError }
+            cj('#saveDetails').show();
+          {else}
+            cj('#saveDetails').hide();
```
davialexandre: Just noticed this now while reviewing your updates. Why changing _$_ to _cj_? Is this something new in 4.7? It looks odd to me for a few reasons:
- Anyone looking at the code would know that $ is a jQuery object. On the other hand, **cj** is kind of meaningless. Are they the same thing?
- A $ variable is still being passed as an argument to the callback, so why not use it?

omarabuhussein: I actually copied this from the core Contact MapTable.tpl file .... the difference between cj and $ is that there is two copies of Jquery included , one shipped with drupal and the other with civicrm so to avoid the conflict the decided to create an alias for civicrm juqery copy so $ become cj.

davialexandre: I see.. so cj stands for CiviCRM Jquery. Well, in this case we could still use the $ variable, because it enclosed to the CRM.$ callback and would point to the jQuery instance loaded by CiviCRM.
Anyway, since you just copied it from the ContactMapTable.tpl, I think it is fine to keep it as this.

davialexandre:
Just reviewed your last changes and everything looks good. Approved!

hrjobcontract/templates/CRM/Hrjobcontract/Import/Form/MapTable.tpl
omarabuhussein: When enclosing that code inside CRM.$ the the import page don't work properly and in the console the following error appear ==> (CRM is not defined ) ... So instead of wasting time on investigating  I just checked the core template files and the CRM.$ was removed from there (in MapTable.tpl ) so I did the same.

PR #941, opened 2016-05-19
PCHR-1069: Fix import appraisals issues after 4.7 migration

omarabuhussein:
## Problem
After migrating to civicrm 4.7 , The  appraisals import page is no longer opening and   this exception is thrown 

`Cannot read unrecognized property CRM_Core_Config::$maxImportFileSize.`

The problem was in this file :

**uk.co.compucorp.appraisals/CRM/Import/Form/DataSourceBaseClass.php**

Specifically in these lines :
 
`    if ($config->maxImportFileSize >= 8388608) {
      $uploadFileSize = 8388608;
    }
    else {
      $uploadFileSize = $config->maxImportFileSize;
    }`

But since **CRM_Core_Config** is no longer have this property `maxImportFileSize` , The error is thrown.

## Solution

I modified  the DataSourceBaseClass.php class to to use this formula instead ( which is the same used in the core import files ) :

    `$uploadFileSize = CRM_Utils_Number::formatUnitSize($config->maxFileSize . 'm', TRUE);
    $uploadSize = round(($uploadFileSize / (1024 * 1024)), 2);`

I also done some minor changes in template files.


coldrunKacper:
Approved.

PR #943, opened 2016-05-19
PCHR-959: Add extensions to enable by default

AkA84:
Adding more of the extensions that need to be enabled by default when installing CiviHR. Also at the end of the install script we make sure any update that the new extensions need (not sure if needed though?)


omarabuhussein:
> Also at the end of the install script we make sure any update that the new extensions need (not sure if needed though?)

I am not sure if that needed either .. but let's keep it for now ..

Aprroved

AkA84:
Removed the `extension.upgrade` bit, kinda feels redundant

PR #946, opened 2016-05-25
PCHR-1119: Fix L&A errors on CiviCRM 4.7

davialexandre:
After moving the development to CiviCRM 4.7 some errors were found on this extension:

### Adding/Editing Absence Types
**Problem**: When adding or editing an Absence Type, the following happened: Exception: "Cannot determine default entity. CRM_HRLeaveAndAbsences_Form_AbsenceType should implement getDefaultEntity()."
**Solution**: Added the getDefaultEntity() to the CRM_HRLeaveAndAbsences_Form_AbsenceType class

### Editing existing Work Patterns
**Problems**:
1. When editing an existing Work Pattern, a black page was displayed and the server response had the status code of 500. The webserver error log had the message: PHP Fatal error: Class name must be a valid object or a string in /vagrant/hr16/sites/all/modules/civicrm/Civi/API/SelectQuery.php on line 99.
2. Trying to disable a Work Pattern or set it as default, by clicking on the actions of the list page, resulted in the error: Class "" does not map to an entity
3. Trying to delete a Work Pattern, by clicking on the Delete action of the list page, resulted in the following error: No delete method found

**Solution**: The solution was the same for all the problems. All of theses were caused by failures on API calls. I had to add the WorkPattern entities to the hook_civicrm_entityType hook implementation, so the WorkPattern APIs would work.

### Running tests
**Problem**: Due to some changes on how to create and run tests on 4.7, it wasn't possible to run the extension's tests.
**Solution**:  I updated the tests to remove some unnecessary code and added the new HeadlessInterface and the setupHeadless method to install the extension on the test database before running the tests.

### DIFF
--- a/doc/DEVELOP.md
+++ b/doc/DEVELOP.md
@@ -51,20 +51,36 @@ schema, be sure to:
 ## Test
 
 To run the unit-tests, one must configure CiviCRM to run unit-tests, install
-civix, and link civix to CiviCRM. To validate that the link is setup, run:
+cv, and populate the cv vars file. To do that, run:
 
 ```bash
-user@host:/path/to/civihr/hrjob$ civix civicrm:ping
-Ping successful
+$ cv vars:fill
+Site: /path/to/your/installation/sites/default/civicrm.settings.php
+These fields were missing. Setting defaults:
+{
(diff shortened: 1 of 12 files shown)


omarabuhussein:
Approved

PR #947, opened 2016-05-26
PCHR-1080: changes to export contract type labels instead of values

nishant-bhorodia:
**Issue:**
Contract type is an option group which has multiple options. Every options has a value and a label, eg:
```
| Value      	| Label      
|------------	|------------	
| Contractor 	| Contractor 
| 1          	| Fixed Term
```

When exporting "Contract type" of contacts, The "value" was getting exported instead of "labels".

**Solution:**
Before export, CiviCRM creates a $exportTempTable. Replace the values with respective labels in $exportTempTable using SQL UPDATE query.


omarabuhussein:
Approved

PR #948, opened 2016-05-26
PCHR-889: Exclude deleted contracts from the query of the advanced search

davialexandre:
The JOIN clause that joined the contact table with the contract table would match every
contract for the given contact, even the deleted ones. Since the job roles are related to the
contract, this would also affect job roles searchs. A search for it could result in matches for
roles related to deleted contracts.


coldrunKacper:
Approved.

PR #950, opened 2016-05-26
PCHR-952: Create the AbsencePeriod Entity

davialexandre:
The AbsencePeriod is just a simple class to represent a period of time.
It has the start_date and end_date fields, which both are mandatory
and the start_date can't be greater or equal to the end_date.

The weight is used to determine which is the previous/next period. If
someone tries to create a period without the date, it will automatically be
the maximum existing weight + 1. If someone tries to insert or update a
period with an weight already in use, all the period with the weight equal
to or greater than it will be increased.


uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/AbsencePeriod.entityType.php
```diff
@@ -0,0 +1,11 @@
...
+// This file declares a new entity type. For more details, see "hook_civicrm_entityTypes" at:
+// http://wiki.civicrm.org/confluence/display/CRMDOC/Hook+Reference
+return array (
+  0 => 
+  array (
+    'name' => 'AbsencePeriod',
+    'class' => 'CRM_HRLeaveAndAbsences_DAO_AbsencePeriod',
+    'table' => 'civicrm_absenceperiod',
+  ),
+);
```
nishant-bhorodia: Please add a new line here!

davialexandre: This file was automatically generated wi GenCode.php and it will be overwritten whenever I generate or update any of the xml schema files.

nishant-bhorodia:
Approved!

PR #951, opened 2016-05-27
PCHR-953: Create Absence Periods list page

davialexandre:
This is how the list looks like:

captura de tela 2016-05-27 as 10 01 36

The list contains only four fields:
- Title
- Start Date
- End Date
- Order

When there's more than one period, their order can be changed,
by clicking on the arrow on the Order column, to move it up or down.

The Start and End date columns use the crmDate smarty filter to
make sure the dates follow the default date format selected in CiviCRM
settings.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsencePeriod.php
@@ -1,0 +1,131 @@
+<?php
+
+/**
+ * Class CRM_HRLeaveAndAbsences_Page_AbsencePeriod
+ *
+ * This is the class that handles requests to the Absence Period list and form
+ * pages.
+ */
+class CRM_HRLeaveAndAbsences_Page_AbsencePeriod extends CRM_Core_Page_Basic {
+
+  /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsencePeriod.php
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function browse() {
+    $object = new CRM_HRLeaveAndAbsences_BAO_AbsencePeriod();
+    $object->orderBy('weight');
+    $object->find();
+    $rows = [];
```
omarabuhussein: either to use array() notation here as other places , or convert any array() usage to short array syntax [] such as in `private $links = array(); ` && `$rows[$object->id] = array();`

omarabuhussein:
Just one minor note... Other than that it looks good .

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsencePeriod.php
davialexandre: Nice catch. Short array syntax FTW!!!

PR #952, opened 2016-05-27
PCHR 1034 - funder and cost centre data

brunoqueiros:
**Before**
pchr-1034-before

**After**
pchr-1034-after

# Problem
When was added a new funder or a cost centre without specify any data and clicked in _Cancel_ the row was saved with 'empty' as Funder and '0' as % value. No rows should get saved if Cancel is clicked.

# Solution
It was created filters to check if the row is right, the filters are:
## issetFunder.js
This filter check if they have a value for `funder_id` and if it has an amount value. This filter receives an array, if isn't an array it'll return the parameter.

## issetCostCentre.js
This filter check if they have a value for `cost_centre_id` and if it has an amount value. This filter receives an array, if isn't an array it'll return the parameter.

## isNotUndefined.js
This filter check if there's a `undefined` value in array, it's used on `initData` method to avoid an error when it's populating the role data.


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -238,7 +239,7 @@ define([
                     $scope.edit_data[role_id]['funders'] = [];
 
                     // Split data from the stored funder contact IDs
-                    var funder_contact_ids = data['funder'].split("|");
+                    var funder_contact_ids = $filter('isNotUndefined')(data['funder'].split("|"));
```
danilo-compucorp: Why introducing new $filters instead of creating simple filtering functions in the controller, using either "vanilla javascript" or lodash to filter the arrays ?

com.civicrm.hrjobroles/js/test/filters/isNotUndefined_test.js
```diff
@@ -0,0 +1,28 @@
...
+
+    beforeEach(module('hrjobroles'));
+    beforeEach(inject(function (isNotUndefinedFilter) {
+      isNotUndefined = isNotUndefinedFilter;
+    }));
+
+    it('should remove undefined values', function () {
+      var array = [1, 2, 'undefined', 'test', undefined];
+
+      expect(isNotUndefined(array).length).toBe(3);
```
danilo-compucorp: To be safer, I'd like to check the array contents, like: `expect(isNotUndefined(array)).toEqual(1, 2, 'test');` for example.

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
brunoqueiros: It was just to keep the code separated and to give the possibility to use on view if it's necessary.

danilo-compucorp: Why do you think it would be necessary to use it in the view?

brunoqueiros: In this case will not use but if there's a reason in the future the filter will be there. The main reason was keep the code modular.

danilo-compucorp: If you want to separate it from the controller, please move those methods into a service. When (if)  we need to use those in a view, we can introduce $filters that can reuse the service methods.

com.civicrm.hrjobroles/js/test/services/hr-job-roles-service-filter_test.js
```diff
@@ -0,0 +1,64 @@
...
+        expect(HRJobRolesServiceFilters.isNotUndefined(true)).toBe(true);
+        expect(HRJobRolesServiceFilters.isNotUndefined(undefined)).toBe(undefined);
+      });
+    });
+
+    describe('issetFunder', function () {
+      it('should remove the entries which are without cost_centre_id', function () {
+        var cost_centers = angular.copy(Mock.roles_data[3]['cost_centers']);
+
+        expect(HRJobRolesServiceFilters.issetCostCentre(cost_centers).length).toBe(2);
```
danilo-compucorp: Test the array contents here as well (same as line 22)

com.civicrm.hrjobroles/js/test/services/hr-job-roles-service-filter_test.js
```diff
@@ -0,0 +1,64 @@
...
+        expect(HRJobRolesServiceFilters.issetCostCentre(true)).toBe(true);
+        expect(HRJobRolesServiceFilters.issetCostCentre(undefined)).toBe(undefined);
+      });
+    });
+
+    describe('issetFunder', function () {
+      it('should remove the entries which are without funder_id', function () {
+        var funders = angular.copy(Mock.roles_data[3]['funders']);
+
+        expect(HRJobRolesServiceFilters.issetFunder(funders).length).toBe(2);
```
danilo-compucorp: Test the array contents here as well (same as line 22)

danilo-compucorp:
Approved

PR #953, opened 2016-05-31
PCHR-1150: installing org.civicrm.hrstaffdir throwing an error

omarabuhussein:
## Problem 

When trying to install org.civicrm.hrstaffdir extension the following error is thrown :

**A fatal error was triggered: Could not find custom field for custom.civicrm_value_jobcontract_summary_10.final_termination_date_57**

The issue is when trying to install the extension , Its run this file  (**auto_install.xml**) which is still use 
**Final Termination Date** which is removed in this ticket  PCHR-774 .

## Solution 

I replaced **final termination date** with **job contract end date** .


nishant-bhorodia:
Approved!

PR #954, opened 2016-05-31
PCHR-954: Create the Absence Period form

davialexandre:
This is new Absence Period form for the Leaves and Absences extension. It's a quite simple form, with just 4 fields. All of them are mandatory. When adding a new Absence Period, the start date is automatically filled with the most recent date available, which can be:
- The current date, if there's no other Absence Period
- The maximum end date among all the existing Absence Periods + 1 day

The order number is also automatically filled following these rules:
- It's value is 1 if there's no other Absence Period
- It's value is the maximum weight among all the existing Absence Period + 1

This is how the form looks like:
captura de tela 2016-05-31 as 07 34 50

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -196,7 +196,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
    *
    * @return int the maximum weight
    */
-  private static function getMaxWeight() {
+  public static function getMaxWeight() {
     $tableName = self::getTableName();
     $query = "SELECT MAX(weight) as max_weight FROM {$tableName}";
     $dao = CRM_Core_DAO::executeQuery($query);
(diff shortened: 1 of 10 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * Validates if the Start Date is less than the End Date
+   *
+   * @param array $values An array containing all the form's fields values
+   * @param array $errors A reference to the errors array where errors will be
+   *                      added if dates are invalid
+   */
+  private function validatePeriodDates($values, &$errors) {
+    $startDate = strtotime($values['start_date']);
```
omarabuhussein: you should check the dates if they are valid dates and in valid format too  , and not just if the start date is less than end date ... since there is no guarantee on the data coming from the user ...

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -208,19 +208,15 @@ public static function getMaxWeight() {
...
+   * @param string $date The date to be validated
    *
    * @return bool
    */
   private static function isValidDate($date)
   {
-    $parsed = date_parse($date);
-
-    return $parsed['warning_count'] == 0 && $parsed['error_count'] == 0;
+    return CRM_HRLeaveAndAbsences_Validator_Date::isValid($date);
```
omarabuhussein: Why not moving the logic from **CRM_HRLeaveAndAbsences_Validator_Date::isValid()** method into this one and removing that class ... I mean it just simple thing so there is no need to over complicate stuff ... What do you think ?

omarabuhussein: Or you know ... Just keep it as is ... keeping them separated is better.

omarabuhussein:
Approved

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
davialexandre: You comment actually made me realize we don't need the isValidDate method anymore. I could just use CRM_HRLeaveAndAbsences_Validator_Date::isValid instead

omarabuhussein: Yeah CRM_HRLeaveAndAbsences_Validator_Date::isValid  could do the job alone

PR #955, opened 2016-05-31
PCHR-889: Fix the search description for Job Roles regions on the Adv. Search

davialexandre:
When searching for Job Role Regions, the search values where not displayed on the
search description. For example, when searching for the "South America" region, the user would see this:

captura de tela 2016-05-31 as 15 08 05

instead, they should see this:

captura de tela 2016-05-31 as 15 06 17

This was happening because instead of the Region Options, we were passing the
list of Department Options to the method that builds the description making impossible
to get the correct name of the selected region.

It was also made an small change to the CRM_Hrjobroles_BAO_Query class to remove a dependency on the CRM_Hrjobcontracts_BAO_HrJobRole BAO to load the options to the location, department, region and level type fields.


nishant-bhorodia:
Approved!

PR #956, opened 2016-06-01
PCHR-1109: Add the getNumberOfWorkingDays method to the AbsencePeriod

davialexandre:
The number of Working Days is given by:
Number of days in a Period - Number of weekend days - public holidays

If a public holiday falls on a weekend, it's not counted twice.

In order to get the number of public holidays between two dates, I added the
getNumberOfPublicHolidaysForPeriod method to the PublicHoliday BAO.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -246,4 +246,42 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
 
     return date('Y-m-d');
   }
+
+  /**
+   * Returns the number of working days for this Absence Period.
+   *
+   * The number is given by "number of ways in period" - "weekends" - "public holidays".
+   *
+   * If a public holiday falls on a weekend, we only count one day.
+   *
+   * @return int
+   */
+  public function getNumberOfWorkingDays()
+  {
+    $startDate = new DateTime($this->start_date);
+    $endDate = new DateTime($this->end_date);
+    $oneDayInterval = new DateInterval('P1D');
+
+    // DatePeriod doesn't include the end date,
+    // so we add one more day for it to be included
+    $endDate->add($oneDayInterval);
+
+    $numberOfWorkingDays = 0;
+    $period = new DatePeriod($startDate, $oneDayInterval, $endDate);
+    foreach($period as $date) {
+      $dayOfTheWeek = $date->format('N');
+      $dayIsWorkingDay = $dayOfTheWeek > 0 && $dayOfTheWeek < 6;
+      if($dayIsWorkingDay) {
+        $numberOfWorkingDays++;
+      }
+    }
+
+    $numberOfPublicHolidays = CRM_HRLeaveAndAbsences_BAO_PublicHoliday::getNumberOfPublicHolidaysForPeriod(
+      $this->start_date,
+      $this->end_date,
+      true
+    );
+
+    return $numberOfWorkingDays - $numberOfPublicHolidays;
+  }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
@@ -68,4 +68,39 @@ class CRM_HRLeaveAndAbsences_BAO_PublicHoliday extends CRM_HRLeaveAndAbsences_DA
     }
   }
 
+  /**
+   * Returns the number of active Public Holidays between the given
+   * start and end dates (inclusive)
+   *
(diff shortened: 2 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -246,4 +246,42 @@ public static function getMostRecentStartDateAvailable()
...
+
+    // DatePeriod doesn't include the end date,
+    // so we add one more day for it to be included
+    $endDate->add($oneDayInterval);
+
+    $numberOfWorkingDays = 0;
+    $period = new DatePeriod($startDate, $oneDayInterval, $endDate);
+    foreach($period as $date) {
+      $dayOfTheWeek = $date->format('N');
+      $dayIsWorkingDay = $dayOfTheWeek > 0 && $dayOfTheWeek < 6;
```
omarabuhussein: this assuming weekend is (Saturday + Sunday )  , but I as far as I remember it should be configurable from L&A extension so I think this code have to take that into consideration .

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
```diff
@@ -68,4 +68,39 @@ private static function validateDate($params)
...
+
+    $tableName = self::getTableName();
+    $query = "
+      SELECT COUNT(*) as public_holidays
+      FROM {$tableName}
+      WHERE date >= %1 AND date <= %2 AND is_active = 1
+    ";
+
+    if($excludeWeekends) {
+      $query .= ' AND DAYOFWEEK(date) BETWEEN 2 AND 6';
```
omarabuhussein: same here 

> this assuming weekend is (Saturday + Sunday ) , but I as far as I remember it should be configurable from L&A extension so I think this code have to take that into consideration .

omarabuhussein: same here 

> also please in case my comment above is not what its really needed by specification and the code will not change ... then replace these magic numbers ( 2 , 6 ) to constants .. I know what they are but someone else may not ) .. it always a good practice not to leave unexplained numbers in the code .

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
davialexandre: About the flexible weekends, we don't need to consider them for this calculation.
Now, about the constants, I don't think they will improve the code on this specific situation. Here's why:
- I would need to add the constants to the class even though they have nothing to do with an AbsencePeriod. Something like AbsencePeriod::SATURDAY doesn't make any sense.
- I would only use the constants in a single method. That is a code smell indicating it should be extracted.
- Class constants in PHP are always public (this is change in PHP 7.1), so It means we are exposing some not very good code to be used elsewhere.

I thought about all this while developing and that's why I decided to include the $daysIsWorkingDay variable. Here my reasoning:
A working day is pretty much the same everywhere in the world. Except for a very specific case (that I have no knowledge of) working days are always monday to friday. So, when I say "a day is a working day if its day of the week is greater than 0 and less than 6" it's kind of implicit that each days of the week is a number starting with from Sunday (0) and ending in Saturday (6).

I know, it's not perfect but I think it's better than the constants.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
davialexandre: See my comment above. Also, this is part of a SQL query, which doesn't support constants. I would have to add new params to the query or make more concatenations to the query string.

omarabuhussein: I was meaning by concatenations.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
omarabuhussein: To be honest I don't feel  leaving it as 0 and 6 even if you used $dayIsWorkingDay still good and maybe it could be solved by creating another class that map days to constants since PHP don't support enumeration yet ...  but anyway I don't want to complicate things  I would say at least document that 0 = Sunday and Saturday = 6 with a link to php documentation http://php.net/manual/en/function.date.php   ...

davialexandre: I don't think we need the comment either. The $dayOfTheWeek is set right above the statement, so it's pretty clear you just need to check DateTime::format documention to see that the number mean.

omarabuhussein:
Approved

PR #957, opened 2016-06-01
PCHR-1045 - creating PublicHoliday schema and entity files; Updating …

coldrunKacper:
- Created PublicHoliday xml schema
- Built PublicHoliday entity DAO
- Created PublicHoliday BAO class and API entry point
- Updated install / uninstall SQL script
- Updated currently existing DAO classes with build-dao.sh script


uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/PublicHoliday.xml
```diff
@@ -0,0 +1,47 @@
...
+    <length>127</length>
+    <required>true</required>
+    <comment>The Public Holiday title</comment>
+  </field>
+  <field>
+    <name>date</name>
+    <title>Date</title>
+    <type>date</type>
+    <required>true</required>
+    <unique>true</unique>
```
davialexandre: Was this enough for it to generate the correct UNIQUE index to the generated SQL in civicrm.mysql? Whenever I tried this it didn't work. I had to manually add the index to the schema like this: https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/WorkDay.xml#L98-L104

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/PublicHoliday.entityType.php
```diff
@@ -0,0 +1,11 @@
...
+// This file declares a new entity type. For more details, see "hook_civicrm_entityTypes" at:
+// http://wiki.civicrm.org/confluence/display/CRMDOC/Hook+Reference
+return array (
+  0 => 
+  array (
+    'name' => 'PublicHoliday',
+    'class' => 'CRM_HRLeaveAndAbsences_DAO_PublicHoliday',
+    'table' => 'civicrm_hrleaveandabsences_public_holiday',
+  ),
+);
```
davialexandre: You also need to add the entity to the **hrleaveandabsences_civicrm_entityTypes** hook in **hrleaveandabsences.php**, otherwise you will get some errors trying to use the API.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
```diff
@@ -0,0 +1,27 @@
...
+    $instance->copyValues($params);
+    $transaction = new CRM_Core_Transaction();
+    $instance->save();
+    $transaction->commit();
+    CRM_Utils_Hook::post($hook, $entityName, $instance->id, $instance);
+
+    return $instance;
+  }
+
+}
```
davialexandre: I think we should validate the date here. Take a look at **CRM_HRLeaveAndAbsences_BAO_AbsencePeriod** to get an idea.

Also, please add tests to the BAO. This class is really simple, but we can test if the unique clause is really working (there's a test on AbsenceTypeTest.php that does this) and test the date validation.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/PublicHoliday.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="iso-8859-1" ?>
+
+<table>
+  <base>CRM/HRLeaveAndAbsences</base>
+  <class>PublicHoliday</class>
+  <name>civicrm_hrleaveandabsences_public_holiday</name>
+  <comment>FIXME</comment>
```
davialexandre: I also always forget to fix this FIXME! I think we can left it blank and add some description

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/PublicHoliday.xml
coldrunKacper: Nope, I had to add UNIQUE index manually so I guess I need to include the index in xml schema too.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php
```diff
@@ -22,7 +24,10 @@ class CRM_HRLeaveAndAbsences_Validator_Date {
   public static function isValid($date, $format = 'Y-m-d')
   {
     $dateTime = DateTime::createFromFormat($format, $date);
-
+    if (!$dateTime && $format != 'YmdHis') {
+      $format = 'YmdHis';
+      $dateTime = DateTime::createFromFormat($format, $date);
+    }
```
davialexandre: What if, instead of this new if, we change the default value for format to 'YmdHis'? I think that the dates will already be in this format most of the time when we get to the validation

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/PublicHolidayTest.php
```diff
@@ -0,0 +1,43 @@
...
+      'title' => 'Public holiday 1',
+      'date' => CRM_Utils_Date::processDate('2016-06-01'),
+    ]);
+    CRM_HRLeaveAndAbsences_BAO_PublicHoliday::create([
+      'title' => 'Public holiday 2',
+      'date' => CRM_Utils_Date::processDate('2016-06-01'),
+    ]);
+  }
+
+}
```
davialexandre: Please add tests to check the date validation

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php
coldrunKacper: Agree, if we are going to use CRM_Utils_Date::processDate for each date passed to BAO then it can check only for YmdHis format when validating. Ok, I'll change that.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/PublicHolidayTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+use Civi\Test\HeadlessInterface;
+
+/**
+ * Class CRM_HRLeaveAndAbsences_BAO_PublicHolidayTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_BAO_PublicHolidayTest extends CiviUnitTestCase implements HeadlessInterface {
```
davialexandre: Since you're not using anything from CiviUnitTestCase, you can extend PHPUnit_Framework_TestCase instead and your tests will run MUCH faster.

If you do that, you won't need the $_tablesToTruncate property, neither the setUp nor the tearDown methods. Also, you'll need to implement the TransactionalInterface to make sure the date created by a test will be deleted at its end.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php
davialexandre: Another thing. This change will certainly break things on the AbsencePeriod BAO validation and the AbsencePeriod form. It might be good to run their tests too and check what will break

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php
```diff
@@ -21,13 +19,10 @@ class CRM_HRLeaveAndAbsences_Validator_Date {
    *
    * @return bool
    */
-  public static function isValid($date, $format = 'Y-m-d')
+  public static function isValid($date, $format = 'YmdHis')
```
davialexandre: Do the tests in CRM_HRLeaveAndAbsences_BAO_DateTest still pass after this change?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/WebTest/AbsencePeriod/FormTest.php
```diff
@@ -117,6 +118,7 @@ private function addAbsencePeriod($submit = true) {
     $title = 'Title ' . microtime();
     $this->type('title', $title);
 
+    $this->type('start_date', $this->getValue('start_date'));
```
davialexandre: Why do you needed this? You are filling the field with its own value

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/WebTest/AbsencePeriod/FormTest.php
```diff
@@ -44,6 +44,7 @@ public function testCanAddPeriodWithMinimumRequiredFields() {
   public function testStartDateShouldBeLessThanEndDate() {
     $this->loginAsAdmin();
     $this->openAddForm();
+    $this->type('title', 'Test Absence Period');
```
davialexandre: Is there any specific reason for this? It doesn't affect the test, since we're just checking the dates, I'm  just curious

coldrunKacper: Sure it isn't. Previously I didn't notice that the fields are already filled.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -80,8 +80,8 @@ private static function validateDates($params)
       );
     }
 
-    $startDateIsValid = CRM_HRLeaveAndAbsences_Validator_Date::isValid($params['start_date']);
-    $endDateIsValid = CRM_HRLeaveAndAbsences_Validator_Date::isValid($params['end_date']);
+    $startDateIsValid = CRM_HRLeaveAndAbsences_Validator_Date::isValid($params['start_date'], 'Y-m-d');
+    $endDateIsValid = CRM_HRLeaveAndAbsences_Validator_Date::isValid($params['end_date'], 'Y-m-d');
     if(!$startDateIsValid || !$endDateIsValid) {
```
davialexandre: This won't solve the API problem. If we try to create an AbsencePeriod using the API, the validation will fail. I thought the plan was to change the BAO to always expect dates in Ymd format. For that, we would need to:

- Remove the CRM_Utils_Date::processDate call from the create method
- Change the AbsencePeriod form to convert de date to Ymd before passing the form data to the BAO

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
```diff
@@ -293,6 +297,7 @@ public function startAndEndDatesDataProvider()
   public function startAndEndInvalidDatesDataProvider()
   {
     return [
+      ['2010-01-01', '2015-10-10'],
       ['2015-01-01', 'fdafdasfdsafdsafdsa'],
       ['232131232111', '2015-01-01'],
       ['2015-01-01', 12321321321],
```
davialexandre: A lot of these tests are failing now probably because of the changes on AbsencePeriod BAO:
captura de tela 2016-06-02 as 15 26 39

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Validator/DateTest.php
```diff
@@ -34,7 +34,11 @@ public function testCanValidateDateInDifferentFormats($date, $format)
   public function datesDataProvider()
   {
     return [
-      [date('Y-m-d'), true],
+      [CRM_Utils_Date::processDate(date('Y-m-d')), true],
+      [CRM_Utils_Date::processDate(date('Y-m')), true],
+      [CRM_Utils_Date::processDate(date('Y')), true],
+      [CRM_Utils_Date::processDate(date('m')), true],
+      [date('Y-m-d'), false],
```
davialexandre: Wouldn't it be better to just use date('Ymd') instead of process date? Also, I'm not sure if all those dates should be valid

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php
davialexandre: Sorry I didn't notice this before, but this actually is changing the class/method behavior. This class is just supposed to validate dates but with this its now validating datetimes. Maybe we should rename the class? Although I think having two separated classes would be better?

coldrunKacper:
Latest results:
phpunit4 tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
OK (29 tests, 124 assertions)

phpunit4 tests/phpunit/WebTest/AbsencePeriod/FormTest.php
OK (5 tests, 28 assertions)

phpunit4 tests/phpunit/CRM/HRLeaveAndAbsences/BAO/PublicHolidayTest.php
OK (4 tests, 8 assertions)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Validator/DateTest.php
coldrunKacper: CRM_Utils_Date::processDate() always returns YmdHis format so even if we pass only year it converts it to YmdHis (as we assume to get this format in BAO currently). Anyway I've added also additional date() values to check and only full date('YmdHis') is valid in our case.

davialexandre:
Everything looking good now. PR Approved! 👍

PR #963, opened 2016-06-02
PCHR-1037: Create the Entitlement entity

davialexandre:
The Entitlement entity holds the information about the calculated entitlement
for one specific contract, for an absence type for an an absence period.

It also holds some information that is generated during the entitlement calculation,
like the pro_rata and the number of days brought from the previous absence period.


coldrunKacper:
Approved.

PR #966, opened 2016-06-03
PCHR-1110: Add the getNumberOfWorkingDaysToWork to the AbsencePeriod class

davialexandre:
This method returns the number of days to work on an AbsencePeriod between the
given start and end dates.

This method doesn't count days outside the AbsencePeriod. Meaning that, If the given
start date is less than the AbsencePeriod start date, then the AbsencePeriod's start
date will be used. If the given end date is greater than AbsencePeriod end date,
then the AbsencePeriod's end date will be used.


nishant-bhorodia:
Approved!

PR #967, opened 2016-06-06
PCHR-869 - Validating against date period within job contract date.

coldrunKacper:
Including validation against given date period which is within Job Contract's date periods.


nishant-bhorodia:
Approved !

PR #972, opened 2016-06-09
PCHR-1111: Add the add_public_holidays field to the HRJobLeave

davialexandre:
The future changes to the add/edit contract form will allow the admin to
control if  the number of public holidays should be added to a leave
entitlement. This information will be stored in this new field.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
@@ -106,6 +106,12 @@ class CRM_Hrjobcontract_DAO_HRJobLeave extends CRM_Hrjobcontract_DAO_Base
    * @var int unsigned
    */
   public $leave_amount;
+  /**
+   * If the number of public holidays should be added to the leave_amount
+   *
+   * @var boolean
+   */
+  public $add_public_holidays;
 
   /**
    * class constructor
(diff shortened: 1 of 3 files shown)


hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
```diff
@@ -197,6 +211,7 @@ static function &fieldKeys()
                 'id' => 'id',
                 'leave_type' => 'hrjobcontract_leave_leave_type',
                 'leave_amount' => 'hrjobcontract_leave_leave_amount',
+                'add_public_holidays' => 'hrjobcontract_leave_add_public_holidays',
```
coldrunKacper: Shouldn't it be 'hrjobcontract_leave_leave_add_public_holidays' (twice 'leave' in the string)?

davialexandre: I don't think so. The prefix is just **hrjobcontract_leave** and the field name if **add_public_holidays**, hence hrjobcontract_leave_add_public_holidays. For a coincidence the other field names start with **leave** and that is why we have it twice.

coldrunKacper: Seems legit, but fieldKeys() method description says: "Returns an array containing, for each field, the arary key used for that field in self::$_fields" so it would mean that the array values should be equal to the self::$_fields keys.

davialexandre: Oh, now I see your point! In this case, I need to change the field key in self::$_fields. 
Nice catch!

davialexandre: @coldrunKacper I have updated the field key in self::$_fields. Can you check if it's ok now?

coldrunKacper:
Approved.

PR #973, opened 2016-06-13
PCHR-1146 : Commenting out the line causing "Default location type" to reset. 

vidhatanand:
The issue was causing civicrm to reset the "Default location type". The line  of code responsible for this has been commented, but I am still not sure why was this done at first place.

PR #974, opened 2016-06-13
PCH-1146 bug default location type

vidhatanand:


davialexandre:
Is this PR blocked/waiting for something or was this fixed by https://github.com/civicrm/civihr/pull/973? If the latter, I think it should be closed.

omarabuhussein:
@davialexandre yes it should be closed ... no need for it now ... thanks

PR #975, opened 2016-06-15
PCHR-1139: Change contact tabs order

omarabuhussein:
Related PR https://github.com/compucorp/civihr-tasks-assignments/pull/63
-------------------------------------------------------------------------------

Changing contact tabs order to make more sense to the user , They are order according to this list :

* Summary
* Job contract
* Job role
* Personal details
* Absences
* Appraisals
* Assignments
* Tasks
* Documents
* Identification
* Immigration
* Emergency contacts
* Relationships ( the title changed to Managers )
* Bank details
* Career History
* Medical and Disability
* Qualifications
* Notes
* Groups
* Change log

Also Relationships tab title changed to "Managers" .

The change of order performed by playing with weight property for each tab and between each tab  and the next in order the weight is increased by 10 to give more space in the future to put other tabs in between without missing with other tabs weights. Also since some of the tabs are from the core and some are created by the extension upgrader instead of hook_civicrm_tab  , the only way was to take advantage of &tabs variable which is passed by reference to hook_civicrm_tab and allow you to change the weight for these tabs their ( these changes done in hrui extension ) .  

### Before

selection_027

### After

selection_028

### DIFF
--- a/hrabsence/hrabsence.php
+++ b/hrabsence/hrabsence.php
@@ -454,7 +454,7 @@ function hrabsence_civicrm_tabs(&$tabs, $contactID) {
     )),
     'count' => $absenceDuration/(8*60),
     'title' => ts('Absences'),
-    'weight' => 300
+    'weight' => 10
   );
 }
 
--- a/hrui/hrui.php
+++ b/hrui/hrui.php
@@ -59,7 +59,7 @@ function hrui_civicrm_buildForm($formName, &$form) {
     //HR-358 - Set default values
     //set default value to phone location and type
     $locationId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_LocationType', 'Main', 'id', 'name');
-    // PCHR-1146 : Commenting line ahead to fix the issue, but figuring why it was done at first place coul be useful. 
+    // PCHR-1146 : Commenting line ahead to fix the issue, but figuring why it was done at first place coul be useful.
     //$result = civicrm_api3('LocationType', 'create', array('id'=>$locationId, 'is_default'=> 1, 'is_active'=>1));
     if (($form->elementExists('phone[2][phone_type_id]')) && ($form->elementExists('phone[2][phone_type_id]'))) {
       $phoneType = $form->getElement('phone[2][phone_type_id]');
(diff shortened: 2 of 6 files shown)


davialexandre:
As this is an UI change, can you please add before/after screenshots to the PR?

hrabsence/hrabsence.php
```diff
@@ -454,7 +454,7 @@ function hrabsence_civicrm_tabs(&$tabs, $contactID) {
     )),
     'count' => $absenceDuration/(8*60),
     'title' => ts('Absences'),
-    'weight' => 300
+    'weight' => 10
```
davialexandre: Why the weight starts with negative numbers (-200, -190, -180) and them jumps to 10?

hrui/hrui.php
```diff
@@ -418,28 +418,47 @@ function hrui_civicrm_upgrade($op, CRM_Queue_Queue $queue = NULL) {
...
-      $newTabs[$i] = $tab['title'];
-    }
-    else {
-      $changeLogTabID = $i;
+    switch($tab['title'])  {
+      case 'Assignments':
+        $tabs[$i]['weight'] = 20;
+        break;
+      case 'Identification':
+        $tabs[$i]['weight'] = 50;
```
davialexandre: Maybe we should add a comment here explaining why the weight jumps from 20 to 50 (because of tasks and assignments). Or even better, document the method explaining the tab order and that it depends on other extensions.

uk.co.compucorp.civicrm.appraisals/appraisals.php
```diff
@@ -149,12 +149,12 @@ function appraisals_civicrm_entityTypes(&$entityTypes) {
...
 function appraisals_civicrm_tabs(&$tabs) {
     CRM_Appraisals_Page_Appraisals::registerScripts();
-    
+
     $tabs[] = Array(
         'id'        => 'appraisals',
         'url'       => CRM_Utils_System::url('civicrm/contact/view/appraisals'),
         'title'     => ts('Appraisals'),
-        'weight'    => 1,
+        'weight'    => 20,
```
davialexandre: Absences has this same weight. Is this correct?

AkA84:
Why is this done on the 4.7 branch? It shouldn't be applied for the 4.5-compatible civihr?

hrabsence/hrabsence.php
omarabuhussein: the default summary page in civicrm (Personal details tab) is  to 0 in the core files so I sorted the tabs before it to negative and the tabs after it to positive ... i started from -200 to give a huge space for other tabs to be before  Personal details tab  ... But I now think about it and I think ( Personal details tab ) could be added to the switch statement in hrui extension and i could give better numbering for weights of these tabs

uk.co.compucorp.civicrm.appraisals/appraisals.php
omarabuhussein: oh .. but it not absences but Assignments tab .. thanks

omarabuhussein:
@AkA84  I don't see why it should applied to 4.5 only ... maybe to both branches !! ?

AkA84:
i didn't say 4.5 only in fact! But the way this PR is set up, the changes will be applied only for 4.7. If that should not be the case (as in, they should be applied to both 4.5 and 4.7), then the branch should merge back into `develop`, and later the 4.7 branch should be synced. Makes sense?

omarabuhussein:
@AkA84 Yes Indeed ... will close this then and reopen new one in develop branch.

PR #976, opened 2016-06-15
PCHR-1139: Change contact tabs order

omarabuhussein:
Related PR https://github.com/compucorp/civihr-tasks-assignments/pull/63
-------------------------------------------------------------------------------

Changing contact tabs order to make more sense to the user , They are order according to this list :

* Summary
* Job contract
* Job role
* Personal details
* Absences
* Appraisals
* Assignments
* Tasks
* Documents
* Identification
* Immigration
* Emergency contacts
* Relationships ( the title changed to Managers )
* Bank details
* Career History
* Medical and Disability
* Qualifications
* Notes
* Groups
* Change log

Also Relationships tab title changed to "Managers" .

The change of order performed by playing with weight property for each tab and between each tab  and the next in order the weight is increased by 10 to give more space in the future to put other tabs in between without missing with other tabs weights. Also since some of the tabs are from the core and some are created by the extension upgrader instead of hook_civicrm_tab  , the only way was to take advantage of &tabs variable which is passed by reference to hook_civicrm_tab and allow you to change the weight for these tabs their ( these changes done in hrui extension ) .  

### Before

selection_027

### After

selection_028

### DIFF
--- a/com.civicrm.hrjobroles/hrjobroles.php
+++ b/com.civicrm.hrjobroles/hrjobroles.php
@@ -135,6 +135,11 @@ function hrjobroles_civicrm_navigationMenu( &$params ) {
  * @param $tabs
  * @param $contactID
  * Create a custom tab for civicrm contact which will implement custom drupal callback function
+ * this tab should appear after job contracts tab directly
+ * and since contact summary tab weight is
+ * -190 we chose this to be -180
+ * to give some room for other extensions to place
+ * their tabs between these two.
  */
 function hrjobroles_civicrm_tabs(&$tabs, $contactID) {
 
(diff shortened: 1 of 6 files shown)


com.civicrm.hrjobroles/hrjobroles.php
```diff
@@ -135,14 +135,19 @@ function hrjobroles_civicrm_navigationMenu( &$params ) {
  * @param $tabs
  * @param $contactID
  * Create a custom tab for civicrm contact which will implement custom drupal callback function
+ * this tab should appear after job contracts tab directly
+ * and since contact summary tab weight is
+ * -190 we choosed this to be -180
```
davialexandre: Everything look much better now with the comments! Just one minor problem: Choose is an irregular verb, so the past tense here would be chose

davialexandre:
Approved. Please just take a look at the comment I left about a minor problem on the function documentation.

PR #977, opened 2016-06-16
PCHR-1210: Remove the carry forward expiration date from Absence Type

davialexandre:
To make things simpler on the entitlement calculation, it was decided to remove
these fields from the AbsenceType. Now, the expiration of carry forward days will
only be calculated in terms of duration.

This is how the Absence Type form was before the change:
captura de tela 2016-06-16 as 14 33 18

And this is how it looks like after the change:
captura de tela 2016-06-16 as 14 30 20


omarabuhussein:
Approved. :+1:

PR #979, opened 2016-06-20
PCHR-1039: Entitlement Calculation page

davialexandre:
This adds the new Entitlement Calculation page. At this point we only have the basic structure for the page, which is a simple list displaying all the calculated entitlements for active contracts on a specific AbsencePeriod.

The page is available at: **civicrm/admin/leaveandabsences/periods/manage_entitlements** and it has a required parameter named **id**, which is the ID of the AbsencePeriod to calculated the entitlements for. It also accepts a **cid** parameter containing Contracts IDs. When this is present, the list will only display calculations for Contracts with the given IDs.

This is how the page looks like:

captura de tela 2016-06-20 as 16 45 58

Next to each proposed entitlement, there is a small set of controls that allow the user to override that value (for now it only shows/hides the text field, nothing is saved yet). This is how this works:

untitled5

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
+++ b/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
@@ -1,7 +1,7 @@
 <?php
 
 class CRM_Hrjobcontract_BAO_HRJobContract extends CRM_Hrjobcontract_DAO_HRJobContract {
-    
+
     static $_importableFields = array();
 
   /**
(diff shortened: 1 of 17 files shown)


hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -439,4 +439,87 @@ static function &importableFields($contactType = 'Individual',
...
+    $query = "
+      SELECT c.*
+      FROM civicrm_hrjobcontract c
+        INNER JOIN civicrm_hrjobcontract_revision r
+          ON r.id = (SELECT id
+                     FROM civicrm_hrjobcontract_revision r2
+                     WHERE
+                      r2.jobcontract_id = c.id AND
+                      (
+                        r2.effective_date <= '$startDate'
```
nishant-bhorodia: Please enclose variables in {}.
So for example, $startDate will be {$startDate}

nishant-bhorodia:
Approved!

PR #980, opened 2016-06-21
PCHR-1086: Add a popup to display the details about the entitlement calculation

davialexandre:
When the user clicks on a row of the entitlement calculation list, a popup should be
displayed containing the description of how the proposed entitlement was calculated:

calculation_details_popup

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/hrleaveandabsences.form.manage_entitlements.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/hrleaveandabsences.form.manage_entitlements.js
@@ -16,6 +16,7 @@ CRM.HRLeaveAndAbsencesApp.Form.ManageEntitlements = (function($) {
   function ManageEntitlements() {
     this._listElement = $('.entitlement-calculation-list');
     this._instantiateProposedEntitlements();
+    this._addEventListeners();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/EntitlementCalculationTest.php
```diff
@@ -515,6 +515,63 @@ public function testGetContractShouldReturnTheContractUsedToCreateTheCalculation
...
+    //Contractual Entitlement = 20
+    //Number of Public Holidays = 2
+    //Number of days to work = 64
+    //Number of working days = 259
+    //Pro rata = 5.5
+    //Brought forward = 5
+    //Proposed Entitlement = 10.5
+    $expected = '((20 + 2) * (64 / 259)) = (5.5) + (5) = 10.5 days';
+    $calculationDetails = sprintf('%s', $calculation);
+    var_dump($calculationDetails);
```
coldrunKacper: Should this var_dump be there?

davialexandre: ops... 😅

coldrunKacper:
Approved.

uk.co.compucorp.civicrm.hrleaveandabsences/js/hrleaveandabsences.form.manage_entitlements.js
```diff
@@ -29,6 +30,46 @@ CRM.HRLeaveAndAbsencesApp.Form.ManageEntitlements = (function($) {
...
+  /**
+   * This is the event handler for when the user clicks on a row of the calculations
+   * list.
+   *
+   * It shows the user a popup with details of the selected calculation. Even if the
+   * proposed entitlement was overridden, we display the original calculation.
+   *
+   * @param event
+   * @private
+   */
```
kkalamarski: Actually this method is still public - you can access it from outside.

As long as this is closure you can make this method fully private by using normal function statement:
````javascript
function addEventListeners() {}
````

This way it won't be accessible outside class.

davialexandre: But with the function I would have **this** pointing to the object, right (unless I called it with bind or something like that)?

Anyway, most of this front-end code will be rewritten in the future, when applying the new layout. I just wanted to do something a little better than a bunch of code inside a $(document).ready() but not too complicated.

PR #984, opened 2016-06-23
PCHR-874: Fix dates issue in Job Contract

AkA84:
#### Problem
When editing a Job Contract , the min/max rules would not get applied to the datepickers, even thought the job contract had obviously a start and end date. What it then meant was that one could choose an end date that could come before the start date.

From a technical level, there are two `$watch`s on each period dates that do get triggered also when the form is first loaded (so, not only when the period dates get changed), but for some reason the value does not get stored in the `$scope`, resulting in min/max dates to always be `null`
```js
$scope.$watch('entity.details.period_start_date', function(){
  // this assignment doesn't get into the $scope when the controller is initialized
  $scope.datepickerOptions.end.minDate = getLimitDate(entityDetails.period_start_date, 'min');
});
```
#### Solution
The min/max dates are "manually" initialized (that is, not relying on `$watch`) with the period dates in case the form is editing an already existing job contract


deb1990:
Instead of == and != can we use === and !==?
Applicable for all files

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
```diff
@@ -3,99 +3,129 @@ define([
...
+      /**
+       * Calculates the duration of the period between the given start and end dates
+       *
+       * @param  {Date} dateStart
+       * @param  {Date} dateEnd
+       * @return {string}
+       */
+      function duration(dateStart, dateEnd){
+        if (!dateStart || !dateEnd) {
+          return 'Unspecified';
```
deb1990: Can we keep a constants object to define and use all string values like this?

hrjobcontract/js/test/controllers/form/form-general_test.js
```diff
@@ -0,0 +1,89 @@
...
+      }]);
+    });
+
+    beforeEach(inject(function (_$controller_, _$rootScope_) {
+      $controller = _$controller_;
+      $rootScope = _$rootScope_;
+    }));
+
+    describe('Min and max allowed dates', function () {
+      describe('when there are no period dates already set', function () {
```
deb1990: If there is only a single test case, it is better to remove the describe block and the "it" directly

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
```diff
@@ -3,99 +3,129 @@ define([
...
-            $scope.$watch('entity.details.notice_unit', function(newVal, oldVal){
-                if (newVal != oldVal && entityDetails.notice_unit_employee == oldVal) {
-                    entityDetails.notice_unit_employee = newVal;
-                }
-            });
-
-        }]);
+  'use strict';
+
+  controllers.controller('FormGeneralCtrl',['$scope','$log', 'HR_settings',
```
deb1990: Can we keep similar spacing for all? Maybe we can use some tool to enforce this?

hrjobcontract/js/test/controllers/form/form-general_test.js
AkA84: The `describe` block gives a lexical scope to a set of tests, whether it's one or multiple. Notice that it allows me to have control over the "before" condition for each set of tests, something that I could not do without it.

Also it gives better structure in terms of readability of the test suite, making it more english-like

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
AkA84: I've moved this string in the template (the view), where it should've belonged from the start

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
```diff
@@ -3,99 +3,129 @@ define([
...
+        }
+      });
+
+      $scope.$watch('entity.details.notice_amount', function(newVal, oldVal){
+        if (+newVal && !entityDetails.notice_unit) {
+          $scope.contractForm.detailsNoticeUnit.$setValidity('required', false);
+          $scope.contractForm.detailsNoticeUnit.$dirty = true;
+        }
+
+        if (newVal != oldVal && entityDetails.notice_amount_employee === oldVal) {
```
deb1990: You missed this one ;)

deb1990:
Approved

PR #986, opened 2016-06-24
PCHR-1089: Add the "Copy to All" and "Add 1 Day" buttons

davialexandre:
These buttons are inside the "New Proposed Entitlement" header, and work as follow:

**Add 1 Day button**: Adds an extra day to every entitlement on the list and marks the
entitlement as overridden.

add_1_day

**Copy to All**: Copy the value of the first proposed entitlement on the list to all others and
marks the entitlements as overriden.

copy_to_all


danilo-compucorp:
Looks good. Approved.

PR #987, opened 2016-06-27
PCHR-875: Fix contract's start date validation

brunoqueiros:
When wasn't informed the contract's end date it wasn't validated if the start date is lower than the contract's start date, the solution was remove the `checkIfStartIsLowerThanContractStart` method from the end's statement.

**Before**
pchr-875-before

**After**
pchr-875-after


danilo-compucorp:
Looks ok. Approved.

PR #988, opened 2016-06-27
PCHR-1090: Add a dialog box to allow the user to add/edit an entitlement comment

davialexandre:
There's a new action available on the "Comment" column of the entitlement
calculations list. By clicking it, the user is presented with a dialog box where they
can add/edit a comment to describe the the how that entitlement calculation was made.

Here is how it works:
comment_dialog


brunoqueiros:
Looks good, approved!

PR #989, opened 2016-06-28
PCHR-1105: Fix $compile "non-assignable expression" errors for calendars

AkA84:
### Problem
The is-open attribute on the calendar directives of the "edit role" panel had a value set to an expression (`is-open="!!(CalendarShow.start_date)"`), that caused a [non-assignable expression](https://docs.angularjs.org/error/$compile/nonassign?p0=isOpen(%27start_date%27)&p1=isOpen&p2=uibDatepickerPopup) error.

### Solution
Simply assigning the value of the `CalendarShow.{x}` property  instead of the "force to bool" expression solved the problem


danilo-compucorp:
Looks ok, approved.

PR #992, opened 2016-06-30
PCHR-1108: Add the "Save new entitlements" button

davialexandre:
The button is at the bottom right of the entitlement calculation list and,
when clicked, submits all the calculations, overridden proposed 
entitlements and comments to be stored as entitlements. Any existing 
entitlement for the given period will be deleted.

In this screenshot you can see where the button is positioned:
captura_de_tela_2016-06-30_as_07_29_32


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
```diff
@@ -129,22 +129,33 @@ private function getContractsForCalculation(AbsencePeriod $absencePeriod) {
    */
   private function getActiveContractsForPeriod(AbsencePeriod $absencePeriod, $filter = []) {
     try {
-      $result = civicrm_api3('HRJobContract', 'getactivecontracts', [
-        'start_date' => $absencePeriod->start_date,
-        'end_date' => $absencePeriod->end_date,
-        'api.Contact.getvalue' => ['return' => 'display_name']
+      $contracts = CRM_Hrjobcontract_BAO_HRJobContract::getActiveContracts(
```
omarabuhussein: regarding getActiveContracts() method ... are you aware that now only one active contract should be allowed at one time and you can get it by this method which I created before :
CRM_Hrjobcontract_BAO_HRJobContract::getCurrentContract($contactID)

davialexandre: Yep, but using that method the performance would be even poor, as I would have to call it for each contact. With getActiveContracts() I can get all the active contracts with a single query

omarabuhussein: no you don't get my idea ... The contact cannot have more than one active contract .. so your method will always return on contract only ... check this PCHR-869

omarabuhussein: forget about this ... they are different ... getCurrentContract retrive the current user contract in the absence period ... while yours get all the active contracts in a specific period.

omarabuhussein:
Approved

PR #993, opened 2016-06-30
PCHR-1114: Add the "Manage entitlements" action to the Absence Periods list

davialexandre:
From the Absence Periods list is now possible to go to the Entitlement Calculation page to manage the staff entitlements. To do this, there's a new action available to every period on the list, called "Manage Entitlements". By clicking on it, the user will get a confirmation message and, if they confirm, they will be redirect to the calculations page.

This is the action and the confirmation message:

manage_entitlements


coldrunKacper:
Approved.

PR #995, opened 2016-07-01
PCHR-1225: fix contact search results page warnings

omarabuhussein:
After searching for contacts , a list of (undefined index) warnings notices appear in the search page :

find contacts civihr 1 6 demo

the source of the problem was in hrjobcontract_civicrm_searchColumns Hook implementation in job contract extension where the search result rows are not checked if they contain the location field  (hrjobcontract_details_location) which came from the contact current job contract , so in case some contacts in the search results don't have active job contract the notice will appear for everyone who doesn't have a contract.

I fixed the problem by adding  checks on $row['hrjobcontract_details_location'] so it doesn't through a notice if it is not exist.

selection_053


vidhatanand:
Approved.

PR #996, opened 2016-07-01
PCHR-1107: Add the getCountForCurrentPeriod action to the PublicHoliday API

davialexandre:
This new API action basically returns the value of the (also new)
PublicHoliday.getNumberOfPublicHolidaysForCurrentPeriod method. The API can
accept a single paramenter, named "exclude_weekends" that, when true, will not
count Public Holidays that falls on a weekend.

#### Usage examples for the new API action
Returning the number of public holidays including those that fall on a weekend
```php
$result = civicrm_api3('PublicHoliday', 'getcountforcurrentperiod');
```

Returning the number of public holidays excluding those that fall on a weekend:
```php
$result = civicrm_api3('PublicHoliday', 'getcountforcurrentperiod', [
    'exclude_weekends' => 1
]);
```

As the default **getcount** action of the CiviCRM API, this action returns an integer.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -258,6 +258,25 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
   }
 
   /**
+   * Returns the current AbsencePeriod. That is, the period that contains the
+   * current date. If no such period is found, null will be returned.
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod|null
+   */
+  public static function getCurrentPeriod()
+  {
+    $period = new self();
+    $period->whereAdd("start_date <= CURDATE()");
+    $period->whereAdd("end_date >= CURDATE()");
+    $period->limit(1);
+    if($period->find(true)) {
+      return $period;
+    }
+
+    return null;
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -258,6 +258,25 @@ public static function getMostRecentStartDateAvailable()
...
   /**
+   * Returns the current AbsencePeriod. That is, the period that contains the
+   * current date. If no such period is found, null will be returned.
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod|null
+   */
+  public static function getCurrentPeriod()
+  {
+    $period = new self();
+    $period->whereAdd("start_date <= NOW()");
```
omarabuhussein: correct me if I am wrong but as far as I remmber the absence period use date not datetime .. so its probably better to use CURDATE() here

davialexandre: That is better indeed. Fixed.

omarabuhussein: shouldn't the unit tests should be fixed too ?

davialexandre: No need. In this situation NOW() and CURDATE() will work pretty much the same way. The tests are still passing.

omarabuhussein:
Approved

PR #998, opened 2016-07-01
PCHR-1082: Changing css file name to avoid conflicts

danilo-compucorp:
I've been talking with @brunoqueiros about some styling issues in 4.7.1 and we figured out that changing the css file name (civicrm.css)  or loading it from an external source resolves the issue.
Apparently it just loads one "civicrm.css" file. If we provide one, it skips the default one.


AkA84:
Approved

PR #999, opened 2016-07-04
PCHR-1070: Fix contact merge on civicrm 4.7.7

omarabuhussein:
## Problem

When trying to merge contacts on civihr 1.6 ( with civicrm 4.7.7 ) this error appear :
screen shot 2016-06-24 at 11 25 20

This is a civicrm 4.7.7 problem and not related to civihr that it doesn't create "**Contact Deleted by Merge**" activity type automatically when installed via the buildkit which is used in the core files to create an new activity after the merge get complete .

## Solution

I added a new code to hrjobcontract_civicrm_install hook to check for existence of "**Contact Deleted by Merge** activity type and in case it doesn't exists it will be created. and for existing installations I've added an upgrader method to do the same thing. 

Note : this problem is solved in civicrm 4.7.8+ and "**Contact Deleted by Merge**" is created automatically.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1238,6 +1238,36 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


nishant-bhorodia:
Approved!

PR #1003, opened 2016-07-05
PCHR-1062: Improvements to the Add/Edit Job Contract modal

davialexandre:
This modal was updated to reflect the changes introduced by the new
Leave and Absences extension:

- The list of Leave types is now populated with data from the new
AbsenceType entity
- When adding a new Contract, the form shows the default entitlement
and if the AbsenceType requires public holidays to be added to the
entitlement
- It's now possible to set in the modal if the entitlement for a specific
contract will include public holidays
- Only one leave entitlement should have "Add public holidays" set

This is the screen before the changes:
captura de tela 2016-07-05 as 18 05 07

This is the screen after the changes:
captura de tela 2016-07-05 as 18 06 35

Here you can see how only one type can have "Add Public Holidays" set and how the "Estimated Leave Entitlement"  is updated to include/remove the number of public holidays accordingly:
job_contract


hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
+define([
+    'job-contract/vendor/fraction',
```
AkA84: It doesn't seem you're using this vendor at all. If that's not the case, please remove it from this module and maybe check also if it's used at all in the entire extension, and if not just delete it altogether

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+    'job-contract/controllers/controllers'
+], function (Fraction, controllers) {
+    'use strict';
+
+    controllers.controller('FormLeaveCtrl',['$scope', '$rootScope', '$filter', '$log', '$q', 'UtilsService',
```
AkA84:
```js
controllers.controller('FormLeaveCtrl', ['$scope', // ...
```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+      function ($scope, $rootScope, $filter, $log, $q, UtilsService) {
+        $log.debug('Controller: FormLeaveCtrl');
+
+        var promiseNumberOfPublicHolidays = UtilsService.getNumberOfPublicHolidaysInCurrentPeriod();
```
AkA84: Just chain it, no use in having this additional variable
```js
UtilsService.getNumberOfPublicHolidaysInCurrentPeriod()
  .then(function (result) {

  });
```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+
+        promiseNumberOfPublicHolidays.then(function(result){
+          var number = parseFloat(result);
```
AkA84: I think that this entire logic here
``js
var number = parseFloat(result);

if (isNaN(number)) {
  number = 0;
}
```
should be in the `getNumberOfPublicHolidaysInCurrentPeriod` method, as it's its responsibility to make sure that a valid value is returned

File: hrjobcontract/views/modalForm.html [hunk 5]
```diff
@@ -560,33 +560,55 @@ <h4 class="modal-title">{{copy.title}}</h4>
                 </div>
             </uib-tab>
             <uib-tab heading="Leave">
+              <div ng-controller="FormLeaveCtrl">
```
AkA84: Please use the [`controller as`](https://toddmotto.com/digging-into-angulars-controller-as-syntax/) syntax:

```html
<div ng-controller="FormLeaveCtrl as form">
</div>
```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+          if(isNaN(number)) {
+            number = 0;
+          }
+          $scope.numberOfPublicHolidays = number;
```
AkA84: Since you're gonna use the `controller as` syntax (see comment [here](https://github.com/civicrm/civihr/pull/1003/files#diff-4b270671b8363b7cd64d3be53cb39614R563)), you can put every `$scope` property/method in the controller object itself, see for example [here](https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.appraisals/js/src/appraisals/controllers/appraisal-cycle-ctrl.js#L14)

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+        });
+
+        $scope.$watch('entity.leave', function(newValue, oldValue) {
+          var newTypeWithPublicHolidays = null;
```
AkA84: Put all of this logic in its separate, documented function
```js
$scope.$watch('entity.leave', fooBarBaz);

/**
 *
 *
 */
function fooBarBaz(newValue, oldValue) {

}
```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+          newValue.forEach(function(value, i) {
+            if(value.add_public_holidays === '1' && oldValue[i].add_public_holidays === '0') {
```
AkA84: Why are you not simply using `true` and `false` here? If it's because the DB is storing this as a 1/0 value, I'm not 100% sure but i think you can still pass a bool value to the API and it will be correctly converted to the expected value anyway. This could solve a lot of this ugliness with `'1'` and `'0'` that i see going on here

AkA84: If the above doesn't work, can't you at least do some bool casting to simplify this line
```js
if(!!value.add_public_holidays && !oldValue[i].add_public_holidays) {
```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+          var number = parseFloat(result);
+          if(isNaN(number)) {
+            number = 0;
+          }
+          $scope.numberOfPublicHolidays = number;
+        });
+
+        $scope.$watch('entity.leave', function(newValue, oldValue) {
+          var newTypeWithPublicHolidays = null;
+          newValue.forEach(function(value, i) {
```
AkA84: _Apply for all instances_
Please leave whitespaces like this
```js
function (value, i) {}
for (i = 0; i < 10; i++) {}
if (condition) {}

Comment (AkA84) [inline hrjobcontract/js/src/job-contract/controllers/form/form-leave.js]: What about this
```js
// you must load the dependency: common/lodash as `_`
var newTypeWithPublicHolidays = _.find(newValue, function (value, i) {
  return !!value.add_public_holidays && !oldValue[i].add_public_holidays; 
});

```

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
```diff
@@ -0,0 +1,38 @@
...
+            }
+          });
+          if(newTypeWithPublicHolidays) {
+            newValue.forEach(function(value) {
+              if(value.leave_type != newTypeWithPublicHolidays) {
+                value.add_public_holidays = '0';
+              }
+            });
+          }
+        }, true);
```
AkA84: Is `objectEquality` needed?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.js
```diff
@@ -49,6 +49,14 @@ define([
                 is_primary: 0
             };
 
+            // Since we are adding a new Contract, we set the values for each leave type with the
+            // AbsenceTypes values
+            $scope.entity.leave.forEach(function(leave, index) {
+              var absenceTypeID = $scope.entity.leave[index].leave_type;
```
AkA84: Let the code breathe a little, leave a new line between `var` declaration/definition and the rest of the code
```js
var absenceTypeID = $scope.entity.leave[index].leave_type;

$scope.entity.leave[index].leave_amount = utils.absenceTypes[absenceTypeID].default_entitlement;
$scope.entity.leave[index].add_public_holidays = utils.absenceTypes[absenceTypeID].add_public_holiday_to_entitlement;
```

hrjobcontract/js/src/job-contract/services/contract-leave.js
```diff
@@ -155,6 +155,25 @@ define([
...
+                     * If this leave type has "add_public_holidays" set, the estimated amount will be
+                     * the leave amount + the given number of Public Holidays. Otherwise, the estimated
+                     * amount is the same as the leave amount.
+                     *
+                     * @param {number} numberOfPublicHolidays
+                     * @returns {number}
+                     */
+                    modelEntry.getEstimatedLeaveAmount = function(numberOfPublicHolidays) {
+                      if(this.add_public_holidays == '1') {
+                        return parseFloat(this.leave_amount) + parseFloat(numberOfPublicHolidays);
```
AkA84: Out of curiosity, what values are you expecting to come from the api? (given the usage of `parseFloat')

hrjobcontract/js/src/job-contract/services/utils.js
```diff
@@ -127,6 +138,26 @@ define([
...
+
+              API.resource('PublicHoliday', 'getcountforcurrentperiod', {
+                sequential: 1
+              }).get(function (data) {
+                deffered.resolve(data.result);
+              }, function () {
+                deffered.reject('Unable to fetch the number of public holidays in current period');
+              });
+
+
```
AkA84: Empty line

hrjobcontract/views/modalForm.html
```diff
@@ -560,33 +560,55 @@ <h4 class="modal-title">{{copy.title}}</h4>
...
                 <div class="row">
-                    <div class="col-xs-12">
-                        <h4>Leave</h4>
-                        <hr>
-                    </div>
+                  <div class="col-xs-12">
+                      <h4>Leave</h4>
```
AkA84: It seems you are using a mix of 4 and 2 spaces for indentation in this file. Please be consistent with 2

hrjobcontract/views/modalForm.html
```diff
@@ -560,33 +560,55 @@ <h4 class="modal-title">{{copy.title}}</h4>
...
+                        </div>
+                        <div class="col-sm-3 control-label">
+                          Add Public Holidays
+                        </div>
+                        <div class="col-sm-3 control-label">
+                          Estimated Annual Contractual Leave Entitlement
+                        </div>
+                    </div>
+                    <div class="form-group" ng-repeat="leaveEntry in entity.leave">
+                      <label for="{{prefix}}leave-{{leaveEntry.leave_type}}" class="col-sm-3 control-label">{{options.leave.leave_type[leaveEntry.leave_type]}}</label>
```
AkA84: Let's not have unnecessary long lines:
```html
<label for="{{prefix}}leave-{{leaveEntry.leave_type}}" class="col-sm-3 control-label">
  {{options.leave.leave_type[leaveEntry.leave_type]}}
</label>
```

AkA84:
What about when editing an already existing contract? Is that scenario covered in this PR?

AkA84:
What about tests for the services, controllers, methods you wrote?

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
davialexandre: That was indeed really ugly. The reason for this is part because of how MySQL stores the data and part because of how the API returns the data (as a string). I fixed this by updating the service methods that fetch the data to covert them to booleans.

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
davialexandre: About the whitespaces, both reviewers and contributors shouldn't have to check this manually. Do you think we can go ahead with that idea of having Travis CI running a linter for PRs?

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
davialexandre: entity.leave is an array of objects and we need to listen to when the properties of these objects change. $watchCollection wasn't able to do this.

hrjobcontract/js/src/job-contract/services/contract-leave.js
davialexandre: This was an error  😕
The number of public holidays should always be an int.

hrjobcontract/views/modalForm.html
davialexandre: This was because I was basing myself on the existing code, which doesn't use the **controller as** syntax

davialexandre:
Editing is also covered by the PR. Since the edit screens are exactly the same as the add, I didn't add them to the PR description

hrjobcontract/js/src/job-contract/services/contract-leave.js
```diff
@@ -14,6 +14,18 @@ define([
...
 
+        /**
+         * The API returns values as strings, so we convert them to booleans to
+         * make it easy to use them inside conditions
+         *
+         * @param {Object} values - The values object as returned by the API
+         */
+        function adjustAddPublicHolidaysValue(values) {
+          angular.forEach(values, function (value) {
+            value.add_public_holidays = !!parseInt(value.add_public_holidays);
```
AkA84: is `parseInt` necessary in addition to the bool coercion?

hrjobcontract/js/src/job-contract/services/utils.js
```diff
@@ -87,19 +87,36 @@ define([
...
+             *
+             * @returns {Promise}
+             */
+            getAbsenceTypes: function(){
                 var deffered = $q.defer();
 
-                API.resource('HRAbsenceType','get', {
-                    "return": "id,name,title"
-                }).get(function(data){
+                API.resource('AbsenceType','get', {
```
AkA84: 4 space indentation instead of 2 (while you're at it, just fix the whole method). Shouldn't the .editorconfig file take over here?

hrjobcontract/js/src/job-contract/services/utils.js
```diff
@@ -87,19 +87,36 @@ define([
...
+                    "return": "id,title,default_entitlement,add_public_holiday_to_entitlement"
+                }).get(function (data) {
+                    angular.forEach(data.values, function (value) {
+                      value.add_public_holiday_to_entitlement = !!parseInt(value.add_public_holiday_to_entitlement);
```
AkA84: Same as before, do you need `parseInt`?

hrjobcontract/js/src/job-contract/services/utils.js
```diff
@@ -127,6 +144,33 @@ define([
...
+              var deffered = $q.defer();
+
+              API.resource('PublicHoliday', 'getcountforcurrentperiod', {
+                sequential: 1
+              }).get(function (data) {
+                var number = parseInt(data.result);
+
+                if(isNaN(number)) {
+                  number = 0;
+                }
```
AkA84: Just realized you can rewrite this whole bit as such:
```js
var number = parseInt(data.result) || 0;

deffered.resolve(number);
```

Or, if you like one-liners:
```js
deferred.resolve(parseInt(data.result) || 0);
```

hrjobcontract/js/src/job-contract/controllers/contract-list.js
```diff
@@ -29,7 +29,8 @@ define([
                 },
                 promiseUtils = {
                     hoursLocation: UtilsService.getHoursLocation(),
-                    payScaleGrade: UtilsService.getPayScaleGrade()
+                    payScaleGrade: UtilsService.getPayScaleGrade(),
+                    absenceTypes: UtilsService.getAbsenceTypes()
```
AkA84: Fix the indentation in this bit (since we have to do it incrementally for already existing files anyway)

hrjobcontract/js/src/job-contract/services/contract-leave.js
```diff
@@ -155,6 +171,25 @@ define([
...
+                     * leave type.
+                     *
+                     * If this leave type has "add_public_holidays" set, the estimated amount will be
+                     * the leave amount + the given number of Public Holidays. Otherwise, the estimated
+                     * amount is the same as the leave amount.
+                     *
+                     * @param {number} numberOfPublicHolidays
+                     * @returns {number}
+                     */
+                    modelEntry.getEstimatedLeaveAmount = function(numberOfPublicHolidays) {
```
AkA84: whitespace

hrjobcontract/js/src/job-contract/controllers/form/form-leave.js
AkA84: Unfortunately the talk about a CI server is on hold for now, can't do much about it

hrjobcontract/js/src/job-contract/services/contract-leave.js
davialexandre: It is. The API returns the number as a string (`'1'` or `'0'`) and the coercion alone would return true for both:
captura de tela 2016-07-07 as 07 18 10

hrjobcontract/js/src/job-contract/services/utils.js
davialexandre: I'm not a big fan of one-liners, so I went with the first option :D

hrjobcontract/js/src/job-contract/services/utils.js
davialexandre: I think it should. Don't know what happened there, but it's fixed now

davialexandre:
This should be merged to the 1.7 branch. Closing in favor of https://github.com/civicrm/civihr/pull/1007

PR #1004, opened 2016-07-07
PCHR-1277: Fix AngularJS tests in CiviCRM 4.7

AkA84:
The paths of the global dependencies for karma tests have changed, and some dependencies have been added, in CiviCRM 4.7

Also as a minor fix: fixed the import path in the scss file of the appraisals extensions, has the old path isn't valid anymore


brunoqueiros:
Looks good, approved!

PR #1006, opened 2016-07-07
PCHR-1264: Fix vertical tabs overlap

brunoqueiros:
This PR fix the vertical tabs overlap, it was broken because the property `position: absolute;` to solve this was a big padding/margin bottom, the solution is the same from index.php?q=civicrm/appraisals/dashboard#/dashboard.

***Before***
image

***After***
anim


AkA84:
Approved

PR #1007, opened 2016-07-07
PCHR-1062: Improvements to the Add/Edit Job Contract modal

davialexandre:
This modal was updated to reflect the changes introduced by the new
Leave and Absences extension:

- The list of Leave types is now populated with data from the new
AbsenceType entity
- When adding a new Contract, the form shows the default entitlement
and if the AbsenceType requires public holidays to be added to the
entitlement
- It's now possible to set in the modal if the entitlement for a specific
contract will include public holidays
- Only one leave entitlement should have "Add public holidays" set

This is the screen before the changes:
captura de tela 2016-07-05 as 18 05 07

This is the screen after the changes:
captura de tela 2016-07-05 as 18 06 35

Here you can see how only one type can have "Add Public Holidays" set and how the "Estimated Leave Entitlement"  is updated to include/remove the number of public holidays accordingly:
job_contract


AkA84:
Approved

PR #1008, opened 2016-07-07
PCHR-1280: Small fixes on contact page before presentation

danilo-compucorp:
This PR does some small visual changes.

- Remove margins on sides
- Hide breadcrumbs from contact page
- Increase pixels above names
- Equal width within the page

Before:
before

After:

Without photo:
no-photo

With photo:
with-photo


AkA84:
After fixes for `container-fluid` and padding for all "new" extensions' tabs, PR approved

PR #1009, opened 2016-07-08
PCHR-1178: Fixing the missing "date format"

danilo-compucorp:
This PR adds the missing "date format" in the "HR_settings" object, which was causing the modals in the Contact page / Task tab to not be displayed.


AkA84:
This is not correct for a couple of reasons:

1. If makes the automated test fails
2. You're changing the default value of a common service that is used across various extensions, instead of fixing the problem specifically on T&A

Now `HR_settings`, and the architecture of modules revolving dates that is using it, is far from perfect in the current implementation and the way it works is too much obfuscated in the implementation of some of those modules, so I understand this was a bit tricky to figure out.

It doesn't help that the current architecture is T&A is not much clear as well. What was missing was the fetching of the date format for all the states registered in the `civitasks.appTasks` module:
```js
.resolveForAll({
  format: ['DateFormat', function (DateFormat) {
    return DateFormat.getDateFormat();
  }]
});
```
That method assigns the value to `HR_settings.DATE_FORMAT` (again, not the clearest logic here), which solves the issue.

You might want to check if the problem is also in the "Documents" tab

PR #1010, opened 2016-07-12
PCHR-1060: Fix Brought Forward calculation

davialexandre:
The new BroughtForward entity was added to improve the way expired brought forward
is handled. This entity doesn't have any methods and it's basically just a representation of
a table in the database, where the brought forward balance for one specific entitlement will
be store. It has the following fields: entitlement_id, expiration_date and balance.

Before this, each entitlement had an field to store its number of brought forward days and
another one to store the expiration date of these days. Getting the brought forward balance
was difficult, as one would have to check the number of leaves takes before the
brought forward expiration date and subtract it from the number of days brought forward.

With this change, we can store a positive balance  to representing the number of days brought
forward on a given entitlement; and a negative balance, representing the number of days
expired. This way, one can easily get the balance by just summing the brought forward balances.
Besides making the calculation part simpler, this also make things easier on the reporting part,
as the balance can be easily retrieved with a SELECT SUM() query.

The idea is to have a scheduled job that will run daily, that will check for expired brought forward
and will create the respective record with a negative balance.


davialexandre:
Closing as this will require some more work and a different solution

PR #1011, opened 2016-07-13
PCHR-1064: Direct to calculation page after saving contract

davialexandre:
When adding or editing a contract, a confirmation message will be displayed
saying that, if the user proceeds, the leave entitlements will be updated.
This update happens on the Manage Entitlements page, meaning that,
after saving the contract, the user will be redirected to it. The ID of the
contract's contact is added to the URL as the cid parameter in order to
only update that employee entitlements.

After the user saves the employee leave entitlements, they should be 
redirected to the contact page.

To make it easier the redirect to the Manage Entitlements page, the
mandatory paramenter id is now optional. When not passed in the URL,
it will use the current period by default.

Here is how the confirmation is displayed when adding a new contract:
add

And here is the same confirmation when editing an existing contract:
change


brunoqueiros:
Looks good, approved!

PR #1013, opened 2016-07-18
PCHR-1286: Moving styles from hrui extension to custom civicrm

danilo-compucorp:
The _hrui_ extension defines some style customizations that depend on the _custom-civicrm.css_ from the _org.civicrm.bootstrapcivicrm_ extension.
The problem is that, the way it is right now, those _hrui_ customizations are always applied even if  _custom-civicrm.css_ is not enabled.
This PR moves the styles from _hrui_ into _custom-civicrm.css_, in the _org.civicrm.bootstrapcivicrm extension_.

Regarding the code changes in the _org.civicrm.bootstrapcivicrm_ extension:

- The common style overrides were moved to the _common_ directory;
- The _hrui_ styles overrides were moved to the _contact_ directory

Result:

With custom theme:
with-custom-styles

Without custom theme:
without-custom-styles


AkA84:
@danilo-compucorp this PR seems straightforward enough, but it's clear that you changed the scss partials structure of the civicrm part of the extension, could you elaborate more in the PR description? What did you change? What did you add? Where did the hrui styles ended up, in the theme extension?

AkA84:
Approved

PR #1015, opened 2016-07-18
PCHR-1065: Add the "Manage leave entitlement" task to the Advanced Search

davialexandre:
This commit adds a new "Manage leave entitlement" to the Advanced Search. When
the user selects this action, they're redirected to a form containing a dropdown
listing all Absence Period, so they can select to which one they want to manage
the entitlements to. After selecting the period, we should redirect to the Manage
Entitlements page (with the selected Period ID and the ID of the contacts on the
search form).

Here you can see the new Task in action:
manage_leave_entitlements


omarabuhussein:
Looks good , Approved

PR #1016, opened 2016-07-19
PCHR-459: Update standard timeline and activity types for built-in case types

omarabuhussein:
I've updated the standard timeline and activity types for build-in case types according to this wiki page :

https://compucorp.atlassian.net/wiki/pages/viewpage.action?pageId=75857955

### DIFF
--- a/hrcase/xml/case/Appraisal.xml
+++ b/hrcase/xml/case/Appraisal.xml
@@ -23,7 +23,6 @@
     <ActivityType>
       <name>Link Cases</name>
     </ActivityType>
-
     <ActivityType>
       <name>Prepare formats</name>
     </ActivityType>
@@ -51,6 +50,12 @@
     <ActivityType>
       <name>References Check</name>
     </ActivityType>
+    <ActivityType>
+      <name>Collect completed Appraisals</name>
+    </ActivityType>
+    <ActivityType>
+      <name>Complete contract revisions</name>
+    </ActivityType>
   </ActivityTypes>
 
   <ActivitySets>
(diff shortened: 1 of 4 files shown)


hrcase/xml/case/Appraisal.xml
```diff
@@ -60,36 +65,23 @@
...
+          <name>Complete contract revisions</name>
           <reference_offset>0</reference_offset>
         </ActivityType>
-	<ActivityType>
-          <name>Follow up on progress</name>
-	  <reference_offset>30</reference_offset>
-          <reference_select>newest</reference_select>
-        </ActivityType>	
-	<ActivityType>
-          <name>Collection of Appraisal forms</name>
```
davialexandre: Shouldn't these be also removed from the ActivityTypes node above?

hrcase/xml/case/Exiting.xml
```diff
@@ -59,20 +58,22 @@
         </ActivityType>
```
davialexandre: The document you linked on the description says the "Schedule Exit Interview" offset should be -10, but you didn't change it (it's 0 now)

davialexandre: I just noticed that, different from the wiki page, the ticket description says it should be 0. So I don't know what value it should actually be

davialexandre:
The wiki page linked on the descriptions says Appraisals and Probation don't have any Roles, but there are roles in the XML files. Is that correct?

hrcase/xml/case/Appraisal.xml
omarabuhussein: No I've checked with mike and he told me to leave them for now.

hrcase/xml/case/Exiting.xml
omarabuhussein: indeed , thanks for the catch

omarabuhussein:
> The wiki page linked on the descriptions says Appraisals and Probation don't have any Roles, but there are roles in the XML files. Is that correct?

Yes ... I've check this with mike too and we will keep them for now.

davialexandre:
Approved after the offset fix for the "Schedule Exit Interview" activity type

PR #1017, opened 2016-07-19
PCHR-911: Add a label to the Actions column of the absence type list

davialexandre:
**Before**:
captura de tela 2016-07-19 as 09 20 53

**After**:
captura_de_tela_2016-07-19_as_09_21_16


coldrunKacper:
Approved.

PR #1018, opened 2016-07-19
PCHR-1066: Allow AbsenceType with add_public_holiday_to_entitlement to be updated

davialexandre:
#### Problem
Trying to update an existing Absence Type where add_public_holiday_to_entitlement
was checked, was throwing the error: There is already one Absence Type where "Must
staff take public holiday as leave" is selected.

#### Solution
That happened because the validation rule that checks for existing Absence Types
with add_public_holiday_to_entitlement == 1 wasn't taking into account if the existing
type was the one being updated. I fixed this, by adding a "id <> $this->id" condition  
to the validation query, so it won't consider the type being updated as the existing one.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -197,7 +197,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    */
   private static function validateParams($params) {
     if(!empty($params['add_public_holiday_to_entitlement'])) {
-      self::validateAddPublicHolidayToEntitlement();
+      self::validateAddPublicHolidayToEntitlement($params);
     }
 
     if (!empty($params['allow_request_cancelation']) &&
@@ -219,15 +219,20 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    * method checks if one such type already exists and throws an error if that
    * is the case.
    *
+   * @param array $params The params array received by the create method
+   *
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
-   * @throws \CiviCRM_API3_Exception
    */
-  private static function validateAddPublicHolidayToEntitlement() {
-    $result = civicrm_api3('AbsenceType', 'getcount', array(
-        'sequential' => 1,
-        'add_public_holiday_to_entitlement' => 1,
-    ));
-    if(!isset($result['result']) || $result['result'] > 0) {
+  private static function validateAddPublicHolidayToEntitlement($params) {
+    $dao = new self();
+    $dao->add_public_holiday_to_entitlement = 1;
+
+    $id = empty($params['id']) ? null : intval($params['id']);
+    if($id) {
+      $dao->whereAdd("id <> {$id}");
+    }
+
+    if($dao->count() > 0) {
       throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
           'There is already one Absence Type where "Must staff take public holiday as leave" is selected'
       );
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
@@ -22,6 +22,13 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceTypeTest extends PHPUnit_Framework_TestC
     return \Civi\Test::headless()->installMe(__DIR__)->apply();
   }
 
+  public function setUp() {
+    // We delete everything two avoid problems with the default absence types
+    // created during the extension installation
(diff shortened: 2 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -197,7 +197,7 @@ private static function unsetDefaultTypes() {
    */
   private static function validateParams($params) {
     if(!empty($params['add_public_holiday_to_entitlement'])) {
-      self::validateAddPublicHolidayToEntitlement();
+      self::validateAddPublicHolidayToEntitlement($params);
```
omarabuhussein: since you are only using the ID parameter, there is no need to pass all the parameters to  validateAddPublicHolidayToEntitlement method and you can only pass the ID , another way is to create a class property called $_ID which store the ID for the leave type you are currently editing

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -219,15 +219,18 @@ private static function validateParams($params) {
...
-    $result = civicrm_api3('AbsenceType', 'getcount', array(
-        'sequential' => 1,
-        'add_public_holiday_to_entitlement' => 1,
-    ));
-    if(!isset($result['result']) || $result['result'] > 0) {
+  private static function validateAddPublicHolidayToEntitlement($params) {
+    $dao = new self();
+    $dao->add_public_holiday_to_entitlement = 1;
+    if(!empty($params['id'])) {
+      $dao->whereAdd('id <> '. $params['id']);
```
omarabuhussein: since It 09:30 PM here I may miss something , but I think you are passing  $params['id'] without escaping it.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: I'm passing the $params instead of only the ID to keep it consistent with the other validation methods that also expect the array. There's also the little advantage that, if we need to use more fields in the validation, they'll already be there.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: whops... Nice catch! My bad

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -91,6 +98,14 @@ public function testThereShouldBeOnlyOneTypeWithAddPublicHolidayToEntitlementOnU
     $this->updateBasicType($basicEntity2->id, ['add_public_holiday_to_entitlement' => true]);
   }
 
+  public function testUpdatingATypeWithAddPublicHolidayToEntitlementShouldNotTriggerErrorAboutHavingAnotherTypeWithItSelected() {
```
omarabuhussein: shouldn't you add 

```
/**
   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
*/
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
omarabuhussein: okay no problem then

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -22,6 +22,13 @@ public function setUpHeadless() {
     return \Civi\Test::headless()->installMe(__DIR__)->apply();
   }
 
+  public function setUp() {
+    // We delete everything two avoid problems with the default absence types
+    // created during the extension installation
+    $tableName = CRM_HRLeaveAndAbsences_BAO_AbsenceType::getTableName();
```
omarabuhussein: I see there is no need to do this ,so in this test method :

testUpdatingATypeWithAddPublicHolidayToEntitlementShouldNotTriggerErrorAboutHavingAnotherTypeWithItSelected

You can retrieve the default absence type holiday/vacation and get its ID and then perform
  $this->updateBasicType($entity1->id, ['add_public_holiday_to_entitlement' => true]); on it.

davialexandre: This is a bad practice for tests and can make the test to fail if we decide to remove the default types in the future. In doing unit tests, we have to make sure to isolate each test in itself. That is, a test cannot depend on something other than what it or the setup method creates.

omarabuhussein: agree

omarabuhussein:
Approved

PR #1020, opened 2016-07-20
PCHR-1263: ensure hrcase dependency on T&A

omarabuhussein:
Linked PR : https://github.com/compucorp/civihr-tasks-assignments/pull/79

## Problem
hrcase extension specifies default set of case type, activity types etc. This extension should not be installed without Tasks and Assignments extension.

## Solution

I've checked if T&A is installed and enabled inside both  **hrcase_civicrm_install** and **hrcase_civicrm_enable** so if it is not installed and enabled the user will be redirected to extensions manager page and an error notification will appear informing the user to installed/enable T&A extension first .

also a readme file was added so the user know about the dependency .


coldrunKacper:
Approved.

PR #1023, opened 2016-07-21
PCHR-1146: Set the default phone location type when creating/editing a contact

davialexandre:
When creating a new contact, the location type of the first phone number 
is automatically set to the default location type, but this doesn't happen to
the second number (Mobile), which end up with another location type by default.

We use the buildForm hook to alter the form in order to change the default
value of the location type field to the default location type. This is done only
if the phone number empty, so to avoid overwrite any previously selected
value.

Having "Work" set as the default Location type, this is how the field value was set before this change:

captura_de_tela_2016-07-21_as_07_53_10

Now, this is the location type field showing the default location type for both phones after this change:

captura_de_tela_2016-07-21_as_07_53_36

It still uses the default location type as long as the phone is empty:

captura de tela 2016-07-21 as 08 01 39

If the phone is not empty, the saved location type is displayed:

captura de tela 2016-07-21 as 08 03 01

### DIFF
--- a/hrui/hrui.php
+++ b/hrui/hrui.php
@@ -57,18 +57,11 @@ function hrui_civicrm_buildForm($formName, &$form) {
   if ($form instanceof CRM_Contact_Form_Contact) {
     CRM_Core_Resources::singleton()
       ->addSetting(array('formName' => 'contactForm'));
-    //HR-358 - Set default values
-    //set default value to phone location and type
-    $locationId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_LocationType', 'Main', 'id', 'name');
-    // PCHR-1146 : Commenting line ahead to fix the issue, but figuring why it was done at first place coul be useful.
-    //$result = civicrm_api3('LocationType', 'create', array('id'=>$locationId, 'is_default'=> 1, 'is_active'=>1));
-    if (($form->elementExists('phone[2][phone_type_id]')) && ($form->elementExists('phone[2][phone_type_id]'))) {
-      $phoneType = $form->getElement('phone[2][phone_type_id]');
-      $phoneValue = CRM_Core_OptionGroup::values('phone_type');
-      $phoneKey = CRM_Utils_Array::key('Mobile', $phoneValue);
-      $phoneType->setSelected($phoneKey);
-      $phoneLocation = $form->getElement('phone[2][location_type_id]');
-      $phoneLocation->setSelected($locationId);
+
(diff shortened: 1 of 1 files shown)


hrui/hrui.php
```diff
@@ -57,18 +57,11 @@ function hrui_civicrm_buildForm($formName, &$form) {
...
-    if (($form->elementExists('phone[2][phone_type_id]')) && ($form->elementExists('phone[2][phone_type_id]'))) {
-      $phoneType = $form->getElement('phone[2][phone_type_id]');
-      $phoneValue = CRM_Core_OptionGroup::values('phone_type');
-      $phoneKey = CRM_Utils_Array::key('Mobile', $phoneValue);
-      $phoneType->setSelected($phoneKey);
-      $phoneLocation = $form->getElement('phone[2][location_type_id]');
-      $phoneLocation->setSelected($locationId);
+
+    $phoneIndex = 2;
+    if (_hrui_phone_is_empty($phoneIndex, $form)) {
```
omarabuhussein: While it much cleaner like that , I am not really convinced by separating the logic here into functions , I don't believe these functions _hrui_phone_is_empty() , _hrui_set_phone_type_as_mobile() and _hrui_set_phone_type_as_mobile() will be used in somewhere else and you can just move there implementation here and remove them .

davialexandre: Well... even you believe this is cleaner. This alone is a good enough reason for having them. Besides, even after the refactoring, the hrui_civicrm_buildForm function still has almost 40 lines, which is A LOT.

omarabuhussein:
Approved.

hrui/hrui.php
davialexandre: Ok, so instead of function size, let's consider its responsibility. One function should have a single responsibility and that's was clearly not the case here. If we respect this, functions we'll automatically get smaller functions in most of the cases

omarabuhussein: I think single responsibility principle applies to OOP not to functional Programming

davialexandre: Sorry, but I'll have to say you're wrong there. The reasoning for single responsibility applies to procedural programming as well (functional programming is another beast). Single responsibility is important for a number of reasons like: improve code readability, improved reusability, improved maintainability, better cohesion and so on. 
If you think about the very basic structure of OOP, it boils to functions operating over data structure. In other words, procedural programming (of course, there's a lot more to it OOP, but it is still true)! Making this another reason why the same principle applies here.

PR #1024, opened 2016-07-21
PHCR-1306: Styleguide extension (WIP)

AkA84:
This PR adds a new extension _org.civicrm.styleguide_ which adds a styleguide page at _civicrm/styleguide_. The styleguide is still a WIP but eventually will show how the civicrm/civihr elements will look like with the currently enabled theme

#### Notes
*  Given that the styleguide itself has specific css, mixins, etc, a gulpfile had been added to watch and compile the scss partials
* The theme itself (_org.civicrm.bootstrapcivicrm_) had been amended while the styleguide was being built:
  * the colors had been tweaked so that they match the [toolkit](https://invis.io/MU4RTI1S5#/111453672_Page_1_-_Common_Elements)
  * custom variables are now all prefixed with `$crm-` to better distinguish them from the original variables that come with Bootstrap
  * form elements, buttons and navs style had been tweaked
  * an `initial` mixin had been added, as a quick way to reset values for given elements (IE does not support the `initial` value for properties)

styleguide


danilo-compucorp:
Some colors seem to be wrong now:

2_0_body_0_desktop

To be more specific, the "pink areas" are different:
colors-errors

danilo-compucorp:
Approved

AkA84:
@danilo-compucorp thanks a lot for your commit, saved a lot of time!

PR #1025, opened 2016-07-22
PCHR-1327: Disable CiviReport

davialexandre:
CiviReport was removed from the list of components enabled
during hrui installation. With this, the component will be disabled
right after a new installtion with civibuild.

Since CiviReport won't be available anymore, I also removed a
piece of code that was deleting some of the reports created by
this component.


coldrunKacper:
Approved.

PR #1027, opened 2016-07-25
PCHR-459: Fix a missing semicolon that cause buildkit setup to crash

omarabuhussein:
Fix a missing semicolon that cause setting up a new civihr instance via the buildkit  to crash


AkA84:
Approved

PR #1029, opened 2016-07-25
PCHR-914: Show no. of hours as "Various" when work pattern has more than 1 week

davialexandre:
**Before**:
captura_de_tela_2016-07-25_as_09_52_21

**After**:
captura_de_tela_2016-07-25_as_09_49_01


nishant-bhorodia:
Looks good to me! Approved.

PR #1030, opened 2016-07-25
PCHR-914: If WorkPattern if the last active one, don't allow it to be disabled

davialexandre:
When disabling a WorkPattern, we now throw a validation error it's it the last one
enabled.

This is how the error message is displayed to the user:
disable_work_pattern

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
@@ -41,8 +41,12 @@ class CRM_HRLeaveAndAbsences_BAO_WorkPattern extends CRM_HRLeaveAndAbsences_DAO_
       self::unsetDefaultWorkPatterns();
     }
 
-    if(empty($params['id'])) {
+    if($hook == 'create') {
       $params['weight'] = self::getMaxWeight() + 1;
+    }
+
+    if($hook == 'edit' && isset($params['is_active']) && !$params['is_active']) {
+      self::checkIfPatternCanBeDisabled($params['id']);
     }
 
     CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
@@ -79,6 +83,38 @@ class CRM_HRLeaveAndAbsences_BAO_WorkPattern extends CRM_HRLeaveAndAbsences_DAO_
     $workPattern->id = $id;
     $workPattern->find(true);
     $workPattern->delete();
+  }
+
+  /**
+   * This method checks if a Work Pattern can be disabled.
+   *
+   * A Work Pattern can be disabled only if it's not the last enabled pattern.
+   *
+   * @param int $id
+   *  The ID of the WorkPattern to be checked
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException
+   */
+  private static function checkIfPatternCanBeDisabled($id) {
+    $tableName = self::getTableName();
+    $id = (int)$id;
+
+    $query = "
+        SELECT COUNT(*) as total
+        FROM {$tableName}
+        WHERE is_active = 1 AND id <> $id
+    ";
+
+    $dao = CRM_Core_DAO::executeQuery($query);
+    $dao->fetch();
+
+    $total = (int)$dao->total;
+
+    if($total == 0) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException(
+        "You cannot disable a Work Pattern if it's the last one"
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -45,6 +45,10 @@ public static function create($params) {
       $params['weight'] = self::getMaxWeight() + 1;
     }
 
+    if(!empty($params['id']) && isset($params['is_active']) && !$params['is_active']) {
```
omarabuhussein: you called :

`empty($params['id'])`

multiple times in create method , you can group all of them is a single if statement like this :

```php
  $hook = 'edit';
    if(empty($params['id'])) {
    $hook = 'create';
      $params['weight'] = self::getMaxWeight() + 1;
 if (isset($params['is_active']) && !$params['is_active'])  {
self::checkIfPatternCanBeDisabled($params['id']);
}
    }
```

omarabuhussein: also in this case , I think you can just do this :

if (!empty($params['is_active']))

instead of : 

`if (isset($params['is_active']) && !$params['is_active'] )`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -82,6 +86,38 @@ public static function del($id)
...
+   * This method checks if a Work Pattern can be disabled.
+   *
+   * A Work Pattern can be disabled only if it's not the last enabled pattern.
+   *
+   * @param int $id
+   *  The ID of the WorkPattern to be checked
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException
+   */
+  private static function checkIfPatternCanBeDisabled($id) {
```
omarabuhussein: since we are inside WorkPattern BAO file I think you can rename this method to **canBeDisabled** instead of this long name.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -82,6 +86,38 @@ public static function del($id)
...
+    $tableName = self::getTableName();
+    $id = (int)$id;
+
+    $query = "
```
omarabuhussein: you can just check the total count without AND id <> $id .. and below if($total == 0)  to 

 if($total == 1)  ,, and you can remve the id from the method parameters.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkPatternTest.php
```diff
@@ -204,6 +204,25 @@ public function testGetValuesArrayShouldReturnEmptyArrayWhenWorkPatternDoesntExi
         $this->assertEmpty($values);
     }
 
+    /**
+     * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException
+     * @expectedExceptionMessage You cannot disable a Work Pattern if it's the last one
+     */
+    public function testCannotDisabledPatternIfItIsTheLastWorkPatternEnabled() {
```
omarabuhussein: change testCannotDisabledPatternIfItIsTheLastWorkPatternEnabled

to 

testCannotDisablePatternIfItIsTheLastWorkPatternEnabled ... just remove the d from disabled :D ... the same in the other test method

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
davialexandre: We are inside the BAO but in a static context, so that will not make much sense.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
davialexandre: Yes. Actually, that was my first implementation of this method, but then I decided to pass the ID because I think it makes more sense to check if a specific Work Pattern can be disabled.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
omarabuhussein: agree

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
omarabuhussein: mmm...  yes maybe ... okay no problem you can keep it like that.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
davialexandre: That will not work here. The is_active param might not be present all the time and I want to call checkIfPatternCanBeDisabled only if it is. If I use `!empty($params['is_active'])` will return true even it is not present.

omarabuhussein:
approved

PR #1031, opened 2016-07-25
PCHR-898:  Remove "Years" from toil and carry forward expiry dropdown

davialexandre:
- Tests that used the "Year" were updated to use "Months" or "Days".
- The expiry date for the "Vacation" absence type was change to 12 months instead of 1 year

| Before | After |
|--------|-------|
|captura de tela 2016-07-25 as 15 32 57|captura de tela 2016-07-25 as 15 32 30|


coldrunKacper:
Approved.

PR #1032, opened 2016-07-26
PCHR-1340: Creating/Editing/Importing role throw error.

omarabuhussein:
## Problem

While trying to import job roles and in case there is a location, department, region cost center or level is exist in the import file  , the import will not work and the following errors appear in the error report excel file :

`Error message: '1118' is not a valid option for field location `

and if location field is set to ( don't import ) a similar error will appear but for another available field.

The same error will also appear in the browser console when trying to create or edit job roles from UI .

The source of the problem is this PR https://github.com/civicrm/civihr/pull/955 , the DAO file for job roles is changed to use psudoconstant which make the BAO methods and API calls to accept only option value values and nothing else , but since hrjobroles and its importer too was programmed to use the option value IDs instead,  the problem appeared.

## Solution

I've changed any place in the hrjob roles that still use the ID of option values to use the value instead . I also wrote unit tests for hrjobrole extension BAO methods and APIs.


com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -47,21 +47,21 @@ public static function getContactRoles($cuid)  {
...
               LEFT JOIN civicrm_option_group cpg on cpv.option_group_id = cpg.id
               WHERE cpg.name = %1";
     $options = array();
     $result = CRM_Core_DAO::executeQuery($query, $queryParam);
     while ($result->fetch()) {
-      $options[] =  array( 'id'=>$result->id, 'label'=>strtolower($result->label) );
+      $options[] =  array( 'value'=>$result->value, 'label'=>strtolower($result->label) );
     }
     return $options;
   }
```
davialexandre: Since now the DAO have pseudoconstants, I think you can use the buildOptions method (inherited from CRM_Core_DAO) instead of this buildDbOptions and won't have to depend on the internal structure of the option_groups/option_values table anymore. The only difference is that it has a different output format (but returns the same information).

This is the output format of buildDbOptions:
```
array (size=2)
  0 => 
    array (size=2)
      'value' => string 'Headquarters' (length=12)
      'label' => string 'headquarters' (length=12)
  1 => 
    array (size=2)
      'value' => string 'Home' (length=4)
      'label' => string 'home or home-office' (length=19)
```

And this is the output format of buildOptions:
```
array (size=2)
  'Headquarters' => string 'Headquarters' (length=12)
  'Home' => string 'Home or Home-Office' (length=19)
```

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -418,26 +418,20 @@ function fini() {}
...
-       return 0;
-     }
+  private function getOptionKey($option, $value)  {
+    $index = array_search(strtolower($value), array_column($this->_optionsList[$option], 'label'));
+    if ($index !== FALSE)  {
+      return $this->_optionsList[$option][$index]['value'];
+    }
+    else {
+      return NULL;
+    }
```
davialexandre: The output format of buildOptions can make this piece of code a little bit simpler. 
Civi also has the CRM_Utils_Array::key method that can be used to avoid the if/else. It would be something like this:

```php
private function getOptionKey($option, $value) {
    return CRM_Utils_Array::key($value, $this->_optionsList[$option]);
}
```

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -418,26 +418,20 @@ function fini() {}
 
 
   /**
-   * get the option database ID given its label or ID
+   * get the Option Value database value given its label
```
davialexandre: Please make the first letter uppercase and add a space between the method description and the @params block

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
...
+    $contactID = $this->individualCreate($contactParams);
+    // create contract
+    $contract = $this->createJobContract($contactID, date('Y-m-d', strtotime('-14 days')));
+    // create role
+    $roleParams = array(
+      'job_contract_id' => $contract->id,
+      'title' => 'test role'
+    );
+    $jobRole = $this->createJobRole($roleParams);
+    $this->assertNotEquals(NULL, $jobRole);
```
davialexandre: hmmmm.... I don't think this is a good assertion. If createJobRole return anything that is not null (a Case, for example) the test will still pass. This would be better here:
- Assert that createJobRole return a class with an ID
- Load the job role with the returned ID
- Assert that the loaded job roles has the same title as the one you created

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * @expectedException PEAR_Exception
```
davialexandre: I think this is a bit too generic. Maybe it would be better if you also add the @expectedExceptionMessage. Without it, any PEAR_Exception will make the test pass. Even if it's not the one you're expecting.

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
...
+   */
+  function testCreateJobRoleWithoutContractID() {
+    $roleParams = array(
+      'title'
+    );
```
davialexandre: The format of this params array is a bit weird. Shouldn't it be something like $roleParams = [ 'title' => '' ]? With this format, I believe the expected error should be caused by the invalid data and not because of a missing contract id

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
...
+      'title' => 'test role',
+      'location' => "amman",
+      'region' => "south amman",
+      'department' => "amman devs",
+      'level_type' => "guru",
+      'cost_center' => "abdali"
+    );
+    // create role
+    $jobRole = $this->createJobRole($roleParams);
+    $this->assertNotEquals(NULL, $jobRole);
```
davialexandre: See my comment above about loading the job role and asserting its values

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
...
+    );
+    $this->createJobRole($roleParams1);
+    $this->createJobRole($roleParams2);
+    $this->createJobRole($roleParams3);
+
+    $this->assertCount(3, CRM_Hrjobroles_BAO_HrJobRoles::getContactRoles($contactID));
+  }
+
+  function testBuildDbOptions() {
+    $params = ['name' => 'test_group', 'is_active' => 1];
```
davialexandre: Nice to see you using the short array syntax here, but you used the array() syntax on the other methods. Can you pick just one or the other?

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
```diff
@@ -0,0 +1,103 @@
...
+    return CRM_Hrjobroles_BAO_HrJobRoles::create($params);
+  }
+
+  /**
+   * Creates sample option group and values to be used in tests
+   *
+   */
+  public function createSampleOptionGroupsAndValues()  {
+    // Create required Option Groups
+    CRM_Core_DAO::executeQuery('UPDATE civicrm_option_group SET is_active = 1');
```
davialexandre: Do you really need this? It will set every option group as active, not only those you need in your test. Also, since this is not wrapped in any kind of transaction and it's not rolled back in a tearDown method, it can cause problems of test isolation (the data created for one test will still exists during another test)

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
```diff
@@ -0,0 +1,134 @@
...
+  function setUp() {
+    parent::setUp();
+  }
+
+  function tearDown() {
+    parent::tearDown();
+  }
+
+  /**
+   * check with empty array
```
davialexandre: Start comment with upper case

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
```diff
@@ -0,0 +1,134 @@
...
+   */
+  function testJobRoleCreateEmpty() {
+    $this->callAPIFailure('HrJobRoles', 'create', array());
+  }
+
+  /**
+   * check if required fields are not passed
```
davialexandre: Upper case here as well

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+require_once 'HrJobRolesTestBase.php';
+
+class api_v3_HrJobRolesTest extends HrJobRolesTestBase {
```
davialexandre: I'm not sure if we really need these tests and if they're adding any value. The HrJobRoles API is just using the BAO methods, so the BAO is the one that needs to be tested

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+require_once 'HrJobRolesTestBase.php';
+
+class CRM_Hrjobroles_BAO_HrJobRolesTest extends HrJobRolesTestBase {
```
davialexandre: this is valid for all the code: try to separate the code in logical blocks and add spaces between them to make it easier to read.

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
```diff
@@ -0,0 +1,134 @@
...
+    $this->assertEquals($result['values'][0]['cost_center'], 'abdali');
+  }
+
+  /**
+   *  Test creating new role with Invalid
+   * (location, region, department, level, cost center ) fields
+   *
+   */
+  function testJobRoleCreateWithInvalidOptionValues() {
+    // create contact
```
davialexandre: If you decided to keep these tests, You should add this test to the BAO too

davialexandre:
How do you plan to port these changes to 1.6 (and 1.7)?

In 1.6 the tests were changed to work with CiviCRM 4.7, so you'll probably need to work on that. Also, I tried to run this in 4.7.7 and got an error, because CRM_Hrjobroles_Import_Form_DataSource is using CRM_Core_Config_Defaults which was removed in 4.7 (https://github.com/civicrm/civicrm-core/commit/d7c217ae866ed4977ad1c89229ff925647695943)

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
omarabuhussein: yes I am going to keep them , since BAO methods unlike API methods don't respect pusdoconstant fields and accept any value.

davialexandre: I'm not sure I understand you. Can you elaborate a bit more on this?

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/BAO/HrJobRolesTest.php
omarabuhussein: oh , It was a mistake

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
omarabuhussein: for example doing this :

```php
$params = [
 'job_contract_id' => $contract->id,
 'title' => 'test role2',
 'location' => 'soutsh ammansasdad',
     ];
$result = civicrm_api3('HrJobRoles', 'create', $params);
```
where location value  'soutsh ammansasdad' is not a valid value will through and error telling you that.

but doing this 
```php
CRM_Hrjobroles_BAO_HrJobRoles::create($params);
```
with the same parameters used in the api method will not throw any error and will pass and store wrong data in the database .


I really didn't tracked why this happen , but it appear there are some additional validation happen in the core API layer that do that.

davialexandre: So what you're actually doing here is testing code from the core, not ours. One more reason to remove these tests, since this should already be covered by the tests in core.

omarabuhussein:
@davialexandre  well , I think I need another PR to port these tests to 1.6 ...

com.civicrm.hrjobroles/tests/phpunit/api/v3/HrJobRolesTest.php
omarabuhussein: The thing is and for example when you try to create new job role from the UI , the code that process the form is using the API to store the job role and not the BAO method . So let assume someone again changed DAO/XML files for the job roles and removed the pusdoconstant from the fields without changing other parts of the code then and if these API tests are removed  all the tests will pass and the system will break due to this change without the developer realize that , but if I kept them then this test specially testJobRoleCreateWithInvalidOptionValues will fail so the developer know that removing the psudoconstant is wrong without changing the other part of the code. so I still think these tests are still important . 

I hope you understood my example above :laughing:

davialexandre: If someone removes a pseudoconstant from the schema/DAO, any value should be accepted, so, if that happens, this test will need to be removed as well. I can't see why the code will break because of that, unless you are expecting a specific value for that field (but then the test would be different and not depend on the API).

In fact, as you said yourself, if you use the create() method you do can create a job role with an invalid option value. If that is really a problem (I don't think it is) we should not allow that to happen and fix the create() method!

What concerns me here is that this is not testing a single line of code from civihr. That is, the test is more likely to  break because of a change in how this validation thing in the api works (I hate this civicrm API!), than in a change in our code. You didn't added any specific code to the job roles API, so there's nothing to test there.

Now, if you really want to make sure that the pseudoconstants exists (I don't think that this is necessary), a better way would be to call the fields() method in the BAO and assert that the pseudoconstant is in the field definition. And also change the test name to clearly state what you're trying to test.

I think that you're afraid the the problem introduced by adding the pseudoconstants will happen again if we don't keep this test, right? This type of test won't help you with that. If the problem was on the import feature, that is what needs to be tested. We also had a problem creating the job roles using the UI, right? So we should also check if the javascript tests need to be fixed to cover any missing scenario.

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
omarabuhussein: if I want to use buildOptions method it will require me to add sample option values in the tests via this method  CRM_Core_BAO_OptionValue::add due to way civicrm cache option values so inserting option values via sql query will never work probably with buildOptions ... but unfortunately using CRM_Core_BAO_OptionValue::add to add option values was always cause civix to hit the time limit so I dropped using it and stayed with my method.

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -47,6 +47,20 @@ public static function getContactRoles($cuid)  {
...
 
   /**
+   * Check Contact if exist   .
+   *
+   * @param String $searchValue
+   * @param String $searchField
+   * @return Integer ( Contact ID or 0 if not exist)
+   */
+  public static function checkContact($searchValue, $searchField) {
```
davialexandre: Why did you move this method up? I'm just curious
Also, since you touched it, what about renaming the method clearly describe what it really does? I think that something like checkIfContactExists of just contactExists (then you can have something like `if(contactExists())`) would be better than this generic "checkContact" that can means anything

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -58,28 +72,14 @@ public static function buildDbOptions($fieldName) {
     $query = "SELECT cpv.value, cpv.label from civicrm_option_value cpv
               LEFT JOIN civicrm_option_group cpg on cpv.option_group_id = cpg.id
               WHERE cpg.name = %1";
-    $options = array();
+    $options = [];
     $result = CRM_Core_DAO::executeQuery($query, $queryParam);
     while ($result->fetch()) {
-      $options[] =  array( 'value'=>$result->value, 'label'=>strtolower($result->label) );
+      $options[$result->value] = strtolower($result->label);
     }
```
davialexandre: So you've changed the output of buildDbOptions to be exactly like the output of buildOptions (except for the use of strlower on the label). Why don't you remove this method and use buildOptions instead?

omarabuhussein:
> How do you plan to port these changes to 1.6 (and 1.7)?
In 1.6 the tests were changed to work with CiviCRM 4.7, so you'll probably need to work on that. Also, I tried to run this in 4.7.7 and got an error, because CRM_Hrjobroles_Import_Form_DataSource is using CRM_Core_Config_Defaults which was removed in 4.7 (civicrm/civicrm-core@d7c217a)

@davialexandre for 1.6 , currently it will be hard to write and be able to run the tests without the new modifications  to the code ,  so it is better to wait until we sync 1.6 with develop .

@AkA84 ^

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
davialexandre: Looking at CRM_Core_BAO_OptionValue::add() I can see it calls CRM_Core_PseudoConstant::flush() to clear the cache, but it clears ALL the cache, and that might be why it is taking so long. I also see that the flush method accepts a pseudconstant name as parameter and, in that case, it will only clear this pseudoconstant values. Maybe that would be faster?

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
omarabuhussein: the output of buildOptions  is different from buildDbOptions  ... buildOptions get the result from the cache where buildDbOptions get it from the database ... so if you have added some option values via  sql query  then buildOptions will not be able to output the new inserted option value even if you cleared the cache , the only way to retrieve this newly inserted option value is to add it via this method CRM_Core_BAO_OptionValue::add so in this case it will be added to the cache and you can use buildOptions to retreive all the option values including it ... but unfortunately using CRM_Core_BAO_OptionValue::add in the tests to add option values was always cause civix to hit the time limit .

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
omarabuhussein: don't know , but if you can look again on CRM_Core_BAO_OptionValue::add() implementation  then this line will always be called CRM_Core_PseudoConstant::flush(); no matter what you pass , so I don't see how what are you talking about will solve the issue

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
omarabuhussein: it may happen while i refactor and doing changes to code so moving this up was unintentional  ... for the name , yes I agree with you and I will change it .

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Form/DataSource.php
```diff
@@ -60,8 +58,12 @@ public function preProcess() {
...
     //Setting Upload File Size
     $config = CRM_Core_Config::singleton();
-
-    $uploadFileSize = CRM_Core_Config_Defaults::formatUnitSize($config->maxFileSize.'m');
+    if ($config->maxImportFileSize >= 8388608) {
+      $uploadFileSize = 8388608;
+    }
+    else {
+      $uploadFileSize = $config->maxImportFileSize;
+    }
```
davialexandre: I don't understand this change. Where did this 8388608 come from? Why we override the value in $config->maxImportFileSize?
Also, the commit says "PR Review Suggestions" (which is pretty generic and will not be very helpful in the future if someone is looking at the commit history) but this was not a suggestion during the PR review

omarabuhussein: I did this change due to your comment on this PR :

>I tried to run this in 4.7.7 and got an error, because CRM_Hrjobroles_Import_Form_DataSource is using CRM_Core_Config_Defaults which was removed in 4.7 (civicrm/civicrm-core@d7c217a)

so this remove the dependency on CRM_Core_Config_Defaults  which is part of you suggestions . Also I am going to squash  ( "PR Review Suggestions") commit before merging.

for 8388608 to be honest  I saw it implemented in https://github.com/civicrm/civihr/blob/07576b915e22cc64a1b5211cc7916bef36f01667/hrjob/CRM/HRJob/Import/Form/DataSourceBaseClass.php and https://github.com/civicrm/civihr/blob/07576b915e22cc64a1b5211cc7916bef36f01667/uk.co.compucorp.civicrm.appraisals/CRM/Appraisals/Import/Form/DataSourceBaseClass.php 

so I copied it to insure consistency among importers code , personally I don't see a big value for it rather than limiting the size of imported file to a certain value regarding the configs which maybe be part of some old requirements . do you think I should keep or remove it ?

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
```diff
@@ -98,6 +108,27 @@ public function createSampleOptionGroupsAndValues()  {
...
+
+  /**
+   * Find and retrieve job role by any of its properties
+   *
+   * @param int|string $value
+   * @param string $identifier
+   *
+   * @return \CRM_Hrjobroles_BAO_HrJobRoles|NULL
+   */
+  public function findRoleByProperty($value, $identifier = 'id')  {
```
davialexandre: Do we really need this method? I see that you made it possible use other properties instead of the ID, but  in every call you're always using the ID. You have the findByID method (inherited from CRM_Core_DAO) that does exactly the same thing.
To retrieve records matching other properties, you can use the commonRetrieve method, also inherited from CRM_Core_DAO

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -418,20 +427,19 @@ function fini() {}
...
-    $index = array_search(strtolower($value), array_column($this->_optionsList[$option], 'label'));
-    if ($index !== FALSE)  {
-      return $this->_optionsList[$option][$index]['value'];
-    }
-    else {
-      return NULL;
+    if (CRM_Utils_Array::value($value, $this->_optionsList[$option])){
+      return $value;
     }
+    return CRM_Utils_Array::key(strtolower($value), $this->_optionsList[$option]);
```
davialexandre: I got a little lost here. Why does this method sometimes returns the value and sometimes returns the label? This seems different from the past version, which returned the value of null.

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -398,6 +406,7 @@ function import($onDuplicate, &$values) {
     $newJobRole = civicrm_api('HrJobRoles', 'create', $params);
 
       if (!empty($newJobRole['is_error'])) {
+        var_dump($newJobRole['error_message']);exit;
```
davialexandre: ops... this can be removed :D

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -148,7 +148,11 @@ function summary(&$values) {
     $session = CRM_Core_Session::singleton();
     $dateType = $session->get('dateTypes');
 
-    $contractDetails = CRM_Hrjobcontract_BAO_HRJobContract::checkContract($params['job_contract_id']);
+    $contractDetails = NULL;
```
davialexandre: All the changes in this file seem OK, but they were not suggested during the PR review. Maybe they should be placed in another commit?

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/Import/Parser/HrJobRolesTest.php
```diff
@@ -0,0 +1,125 @@
...
+    $roleEntity = $this->findRoleByProperty($importParams['title'], 'title');
+    $this->assertEquals($importParams['title'], $roleEntity->title);
+  }
+
+  private function runImport($params, $code = 'assert_success')  {
+    $fields = array_keys($params);
+    $values = array_values($params);
+    $importObject = new CRM_Hrjobroles_Import_Parser_HrJobRoles($fields);
+    $importObject->init();
+    ($code == 'assert_success') ? $code = CRM_Import_Parser::VALID : $code = CRM_Import_Parser::ERROR;
```
davialexandre: I think this $code parameter and how you're asserting things here looks weird. What if you return the value from $importObject->import and do the assertion in the test methods? This will clearly separate the responsibilities:
The test method will do the assertions and the runImport will only, as the name says, run the Import

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/Import/Parser/HrJobRolesTest.php
```diff
@@ -0,0 +1,125 @@
...
+    // run importer
+    $importParams = [
+      'job_contract_id' => $contract->id,
+      'title' => 'test import role2',
+      'location' => 'amman',
+      'hrjc_region' => 'southhggh ammandshhghg',
+      'hrjc_role_department' => 'amman devs',
+      'hrjc_level_type' => 'guru'
+    ];
+    $this->runImport($importParams, 'assert_fail');
```
davialexandre: This test is a lot nicer than the old one, using the API. This way we know that the import feature should never allow invalid option values (even if they change this in the API). Thanks for this improvement!

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Form/DataSource.php
davialexandre: Oh, that's true. I thought you would just fix this in a future commit/PR for 1.6. Good idea using the same fix for 1.5 and 1.6.

About the value, there are two things that bother me:
- It comes from nowhere. Why 8mb instead of 8.5, 7 or even 9? The way it's is it's just a random value. It could be anything
- It overrides the maxImportFileSize in $config. This is quite confusing. Imagine you've configure 10mb as the max import file, then you try to upload a 10mb file and will get an error. It'll probably take some time until you figure out this piece of code overriding the configured value.

If we can't find out what's this 8388608 I think we should always use the value in $config-maxImportFileSize

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
davialexandre: I was thinking that maybe you could do something like this on the tests:
```php
CRM_Core_BAO_OptionValue::create(); //Or maybe insert directly into the database
CRM_Core_PseudoConstant::flush('my_pseudo_constant');
```

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
omarabuhussein: > but in every call you're always using the ID

no I have used the title in some places. but I wasn't aware of commonRetrieve method ... mmmm honestly I want to avoid as much as possible to re run the tests again (they take a lot of time) , so if I didn't made another changes that may require me to run the tests again I will keep it like that for now.

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
omarabuhussein: thanks for the catch

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
omarabuhussein: as I said in another replay I will squash "PR review suggestions" commit , but these changes were required to be able to run the tests .

com.civicrm.hrjobroles/tests/phpunit/CRM/Hrjobroles/Import/Parser/HrJobRolesTest.php
omarabuhussein: good point

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
omarabuhussein: won't work ... to explain it to you let's assume that when you call :

`$optionValues = DAO_NAME::buidOptions('my_pseudo_constant');`

$optionValues will contain the following options :
`[ 'test1' => 'test 1' , 'test2' => 'test 2' ]` 

if you then via database insert you added a new option value to 'my_pseudo_constant' let us call it 'test3' and then you called 

`$optionValues = DAO_NAME::buidOptions('my_pseudo_constant');`  again 

... in this case $optionValues will still contain 

`[ 'test1' => 'test 1' , 'test2' => 'test 2' ]` 

and test 3 will not be added to the list ... the only for test3 to be added to the list is to add it via 
CRM_Core_BAO_OptionValue::create();  or CRM_Core_BAO_OptionValue::add() ... but CRM_Core_BAO_OptionValue::create() is a wrapper for CRM_Core_BAO_OptionValue::add()  with some extra bussnies logic ... so at the end both of them will call CRM_Core_PseudoConstant::flush();

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Form/DataSource.php
omarabuhussein: this don't answer the question but 8388608 is the default upload size value in php.

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
omarabuhussein: /quote  You have the findByID method (inherited from CRM_Core_DAO) that does exactly the same thing.

BTW there is not method called findByID in CRM_Core_DAO

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Form/DataSource.php
davialexandre: Well... if the default size in PHP, it should be set in php.ini not here. One more reason to get rid of this

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
davialexandre: Oh, it's must have been introduced recently https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/DAO.php#L988
But you can still use commonRetrieve and pass the id in the $params array.

davialexandre:
Looks good now. Approved

omarabuhussein:
I run the tests again and everything is fine.

PR #1033, opened 2016-07-27
PCHR-1315: Fixing contactsummary initialization

danilo-compucorp:
This PR fixes the contact summary initialization.

**Issues**: 

1. There was a race condition issue between the tab template, and the "contactsummary" angular app
2. Missing momentjs file

**Solutions**:

1. Making the angular app wait for the DOM document to be loaded and then, either checking a global variable (window.contactsummaryLoad) or the "contactsummaryLoad" event, both defined by the tab template
2. Fixing the path for the momentjs file


AkA84:
approved

PR #1034, opened 2016-07-27
PCHR-977: Job role date treated wrong

brunoqueiros:
This PR fix the validation which check if the job end's date is after then today's date, also it was adjust `hr-job-roles-controller.js` code indentation.

***Before***
image

***After***
image


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -913,8 +913,9 @@ define([
 
                     data.values.forEach(function (object_data) {
                         var end_date = isDateEmpty(object_data.end_date) ? null : object_data.end_date;
```
danilo-compucorp: As we're using `.startOf('day')` for `todaysDate`, shouldn't we use it for the `end_date` as well ?

danilo-compucorp:
Approved

PR #1037, opened 2016-07-28
PCHR-1343: Leave Balance Change Type option list

davialexandre:
This PR adds a new option list that contains the valid types for a Leave Balance Change (still not implemented). All the values are reserved. The available values are:
- Leave
- Brought Forward
- Public Holiday
- Credit
- Debit

I also did some refactoring to the Upgrader class to make it possible to have each option group in a separate xml definition file.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader.php
```diff
@@ -4,6 +4,33 @@
  * Collection of upgrade steps.
  */
```
omarabuhussein: I like this change... it really better organized like this ...

omarabuhussein:
approved

PR #1040, opened 2016-07-28
PCHR-1346: Add the LeavePeriodEntitlement entity

davialexandre:
This entity is part of the new data structure for Period Entitlements and Leave
Requests. It will replace the current Entitlement entity once all the development on
PCHR-1342 is complete.

A LeavePeriodEntitlement is used to group a lot of information about how many days 
a staff member is entitled to during an Absence Period. Related to a LeavePeriodEntitlement, 
there will be: LeaveBalanceChange and LeaveRequests


coldrunKacper:
Approved.

PR #1041, opened 2016-07-28
PCHR-1347: Add the LeaveBalanceChange entity

davialexandre:
This entity is used to represent balance changes to a LeavePeriodEntitlement.
That is, days added or removed from the entitlement. An added day, will be
represented by a LeaveBalanceChange with a positive amount value. Deducted
days are represented as a negative amount. The amount can be fractioned, like
0.5 (half a day).

Some balances change can expire. Those have an expire date and, in a future
PR, a scheduled job will be added to check this date and record a new, negative
change to what has expired.

Some balances can also have a source (so we can track down what caused the
balance change). The idea is to have multiple sources, that's why source_id is not
a real FK, althought in this first version the only existing source will be a leave request
date (still not implemented). In the future, when more sources exist, we'll also need
a source_type field to identify the source.


uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+  <class>LeaveBalanceChange</class>
+  <name>civicrm_hrleaveandabsences_leave_balance_change</name>
+  <comment>Store balance changes to a Leave Period Entitlement</comment>
+  <add>4.4</add>
+  <log>true</log>
+
+  <field>
+    <name>id</name>
+    <type>int unsigned</type>
+    <required>true</required>
```
omarabuhussein: why there is required attribute on the ID field  ?

davialexandre: Because it is required. The ID cannot be null

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+    <comment>One of the values of the Leave Balance Type option group</comment>
+    <add>4.4</add>
+    <pseudoconstant>
+      <optionGroupName>hrleaveandabsences_leave_balance_change_type</optionGroupName>
+    </pseudoconstant>
+  </field>
+
+  <field>
+    <name>amount</name>
+    <type>decimal</type>
```
omarabuhussein: on the ticket description :

`amount: type=decimal(5,2)`

but here in the xml file you don't define the the precision (5) nor the scale ( 2) , I don't know if that even possible to do in the xml file .

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+    <required>true</required>
+    <default>0</default>
+    <comment>The amount of days this change in balance represents to the entitlement</comment>
+    <add>4.4</add>
+  </field>
+
+  <field>
+    <name>expiry_date</name>
```
omarabuhussein: in the ticket description the default value for expiry_date field should be NULL , but this is not defined here.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+    <type>date</type>
+    <comment>Some balance changes can expire. This is the date it will expire.</comment>
+    <add>4.4</add>
+  </field>
+
+  <field>
+    <name>expired_balance_id</name>
```
omarabuhussein: also for expired_balance_id the default is NULL

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+    <pseudoconstant>
+      <optionGroupName>hrleaveandabsences_leave_balance_change_type</optionGroupName>
+    </pseudoconstant>
+  </field>
+
+  <field>
+    <name>amount</name>
+    <type>decimal</type>
+    <required>true</required>
+    <default>0</default>
```
omarabuhussein: please update the ticket description for this PR to include that the default value for amount field is 0

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+    <type>int unsigned</type>
+    <comment>FK to LeaveBalanceChange. This is only used for a balance change that represents expired days, and it will be related to the balance change that has expired.</comment>
+    <add>4.4</add>
+  </field>
+  <foreignKey>
+    <name>expired_balance_id</name>
+    <table>civicrm_hrleaveandabsences_leave_balance_change</table>
+    <key>id</key>
+    <add>4.4</add>
+    <onDelete>CASCADE</onDelete>
```
omarabuhussein: are you sure ( on delete cascade ) is the required behavior for this field ?

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
```diff
@@ -0,0 +1,99 @@
...
+  </foreignKey>
+
+  <field>
+    <name>source_id</name>
```
omarabuhussein: the default is null for this field too

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
omarabuhussein: since it  is declared as primary key , It cannot be null .

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
davialexandre: If we don't add a <required>true</required> to the definition, the field will be null if no value is specified in the INSERT statement.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
davialexandre: Yes. There's no point to keep the child record if we delete the parent.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
omarabuhussein: I also may be mistaken but I interrupt the required tag as a DAO rule and since BAO class extends DAO class , calling the create method without id parameter will lead it to fail since the required ID field is not present .

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
davialexandre: I could not figure out a way to specify the precision in the XML, but CiviCRM uses 20,2 which is equivalent (you can check that in the auto_install.sql file)

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.xml
davialexandre: Yes, I primary key cannot be null, but in MySQL (don't know about other databases), this statement is valid:
```SQL
CREATE TABLE `table` (
  `id` INT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`));
```
With the <required>true</true> a NOT NULL will be added to the statement. In the end both will work the same way (MySQL automatically adds the not null the the primary key, if it's not present), but I prefer to keep the required.
Also, you're almost right about the BAO and the create method. The required info won't have any effect in that method, but if you're in a form it will be used as a validation rule.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.entityType.php
```diff
@@ -0,0 +1,11 @@
+<?php
+// This file declares a new entity type. For more details, see "hook_civicrm_entityTypes" at:
+// http://wiki.civicrm.org/confluence/display/CRMDOC/Hook+Reference
+return array (
+  0 =>
```
omarabuhussein: you already declared this entity type in hrleaveandabsences_civicrm_entityTypes() , why do you do this here again ? or is this how civi work ?

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/files.xml
```diff
@@ -10,4 +10,5 @@
     <xi:include href="PublicHoliday.xml" parse="xml" />
```
omarabuhussein: just for me to understand .. what is this file about ?

davialexandre: You mean the PublicHoliday.xml or the whole files.xml? If mean the whole file, it's kind of index file listing all the schemas for this extension. This is then added to the CiviCRM own index and it's what GenCode.php reads to generate the DAOs

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveBalanceChange.entityType.php
davialexandre: I honestly don't know what this file is for. It automatically generated when you add a new entity with `civix generate:entity`, so I just let it there and forget about it.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/files.xml
omarabuhussein: yes I meant the whole file , thanks for the info .

omarabuhussein:
Approved

PR #1042, opened 2016-07-29
PCHR-1278: Do not inject civihr angular in angular-based core pages

AkA84:
In the `org.civicrm.reqangular` extension we always inject the _reqangular.min.js_ script, even in pages that are not really for CiviHR. Given that we're using (one of) the latest angular version, this seems to cause a conflict with the core angular pages in CiviCRM 4.5

A quick patch for this is to avoid injecting the file altogether for the angular-based core pages, while we figure out how to avoid relying on our own library version for the CiviHR pages

### DIFF
--- a/org.civicrm.reqangular/reqangular.php
+++ b/org.civicrm.reqangular/reqangular.php
@@ -111,19 +111,38 @@ function reqangular_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
  * Implementation of hook_civicrm_pageRun
  */
 function reqangular_civicrm_pageRun($page) {
+  /**
+   * Avoids injecting the common dependencies (which include angular) in core pages
+   * that are already using angular themselves. A quick patch while we
+   * figure out a better solution to avoid having our own copy of angular in CiviHR
+   */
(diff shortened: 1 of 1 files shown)


org.civicrm.reqangular/reqangular.php
```diff
@@ -111,19 +111,38 @@ function reqangular_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
...
+
+/**
+ * Checks if the given page is an angular-based page from core
+ *
+ * Note: CRM_Core_Page_Angular is CiviCrm v1.5, Civi\Angular\Page\Main is CiviCrm v4.7
+ *
+ * @param  [object]  $page
+ * @return boolean
+ */
+function _isAngularCorePage($page) {
```
omarabuhussein: can you change the function name to _reqangular__isAngularCorePage since unfortuntly any function defined here become global and may cause duplication error if some one used the same function name in some where else

AkA84: Oh sorry, forgot what scope these functions have here, will fix

omarabuhussein:
left one comment , otherwise approved

PR #1044, opened 2016-07-29
PCHR-1348: Add the LeaveRequest and LeaveRequestDate entities

davialexandre:
The LeaveRequest entity will represent a Leave Request in the system.
For now, we only need the basic structure to store the requests and it's dates.
This entity will be augmented in the future to fully support the whole leave requests
functionality.

Each LeaveRequest has one or more related LeaveRequestDate, that represents
each individual date of the request. Example: for a leave request from 2016-01-01
to 2016-01-03 we will have 3 LeaveRequestDates (one for 2016-01-01, another
for 2016-01-02 and another for 2016-01-03)


coldrunKacper:
Approved.

PR #1045, opened 2016-08-01
PCHR-1080 - Contract Type exports as Number or Blank issue

coldrunKacper:
When editing a Job Contract start date to a past one, its effective date didn't get changed to the past one and therefore the Job Contract data was not seen on a Summary screen.
The fix contains updating Contract revision effective date when editing Job Contract with past start date or creating a revision with past effective date.


nishant-bhorodia:
Approved!

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -147,5 +118,54 @@ static function importableFields($contactType = 'HRJobContractRevision',
...
+    $effectiveEndDate = $currentEffectiveDate;
+    if ($previousEffectiveDate === $currentEffectiveDate) {
+      // If 'effective_date' is equal to previously iterated revision's
+      // effective date we mark this revision as overrided by the one newly
+      // created.
+      $overrided = 1;
+    } else {
+      // Else, we set current's revision effective day to a day before the
+      // previously iterated (newer) one.
+      $effectiveEndDate = date('Y-m-d', strtotime($previousEffectiveDate) - 3600 * 24);
```
davialexandre: instead of this seconds * hours math, what about this?:
`$effectiveEndDate = (new Datetime($previousEffectiveDate))->modify('+1 day')`

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/RevisionEffectiveDateTest.php
```diff
@@ -0,0 +1,145 @@
+<?php
+
+require_once 'CiviTest/CiviUnitTestCase.php';
+
+/**
+ * Class CRM_Hrjobcontract_RevisionEffectiveDateTest
+ */
+class CRM_Hrjobcontract_RevisionEffectiveDateTest extends CiviUnitTestCase {
```
davialexandre: Usually unit tests test a single class, so we follow the standard of having the test class with the same name as the class being tested suffixed by 'Test'. This also makes it easier to find the tests for a specific class.
So, since you're testing changes in HRJobContractRevision, I think this class should be renamed to CRM_Hrjobcontract_BAO_HRJobContractRevisionTest

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/RevisionEffectiveDateTest.php
```diff
@@ -0,0 +1,145 @@
...
+    $this->_time = time();
+    $this->_createTestContact();
+    $this->_createTestJobContract($this->_testContactId);
+  }
+
+  function tearDown() {
+    parent::tearDown();
+  }
+
+  function testRevisionEffectiveDates() {
```
davialexandre: What about splitting this test method into smaller ones with more descriptive names? The idea is that each test method would test only a single scenario. Just one example:
```php
public function testIfNewEffectiveDateIsEqualsToPreviousEffectiveDateThenItShouldBeOverridden() {
...
}
```

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -147,5 +118,54 @@ static function importableFields($contactType = 'HRJobContractRevision',
...
+   * Helper function used in updateEffectiveEndDates() loop.
+   * It updates revision entry with calculated 'effective_end_date'
+   * and 'overrided' values basing on specified date of previously iterated
+   * revision and date of currenly iterated revision.
+   * 
+   * @param int $revisionId
+   * @param string $previousEffectiveDate
+   * @param string $currentEffectiveDate
+   */
+  protected static function _updateRevisionByEffectiveDates($revisionId, $previousEffectiveDate, $currentEffectiveDate) {
```
davialexandre: Since PHP allows properties and methods visibility to be controlled with public/protected/private, there's no need to use an underscore prefix for private and protected methods.
This also goes against the conventions of the Drupal code standard: https://www.drupal.org/node/608152#naming

> 8 . Protected or private properties and methods should not use an underscore prefix.

davialexandre:
Can you please update the PR description mentioning what causing the problem and how it was fixed?

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/RevisionEffectiveDateTest.php
```diff
@@ -0,0 +1,145 @@
...
+    civicrm_api3('HRJobDetails', 'create', array(
+      'sequential' => 1,
+      'id' => $details['id'],
+      'jobcontract_id' => $this->_testJobContractId,
+      'jobcontract_revision_id' => $details['jobcontract_revision_id'],
+      'title' => "Job Contract started two weeks ago",
+      'period_start_date' => date('Y-m-d', strtotime('-2 week', $this->_time)),
+    ));
+    $details = $this->_getJobContractDetails($this->_testJobContractId);
+    $this->assertEquals("Job Contract started two weeks ago", $details['title']);
```
davialexandre: Let's see if I understood this correctly: These first three assertions are testing that you can change the Contract period without changing the Revision effective date. Is that correct?

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/RevisionEffectiveDateTest.php
```diff
@@ -0,0 +1,145 @@
...
+      'title' => "Job Contract started one week ago",
+      'period_start_date' => date('Y-m-d', strtotime('-1 week', $this->_time)),
+      'jobcontract_id' => $this->_testJobContractId,
+    ));
+    civicrm_api3('HRJobContractRevision', 'create', array(
+      'sequential' => 1,
+      'id' => $detailsRevision['values'][0]['jobcontract_revision_id'],
+      'effective_date' => date('Y-m-d', strtotime('-1 week', $this->_time)),
+    ));
+    $details = $this->_getJobContractDetails($this->_testJobContractId);
```
davialexandre: Why did you have to manually create a new Revision here? I thought the HRJobDetails.create call would create one automatically.

davialexandre:
The tests are much easier to understand now with more comments. Thanks for that! 
I noticed you removed the underscore prefix from the test private methods, but the methods you added to HRJobContractRevision still have it. Once this is fixed, the PR is approved

davialexandre:
Approved

PR #1050, opened 2016-08-03
PCHR-1060: Update entitlement calculation to use new data structure

davialexandre:
This PR is one more step towards fully using the new data structure for period entitlements and leave requests. This is what have changed:
- Usages of the Entitlement class were replaced by LeavePeriodEntitlement
- The getBroughtForward() method was changed to work based on the balance of a LeavePeriodEntitlement. Basically, the amount of days brought forward is the balance from the previous period, adjusted accordingly to the absence type rules
- The getNumberOfDaysTakenOnThePreviousPeriod() now returns the real number of days instead of a hard coded 0 (the method was named getNumberOfLeavesTakenOnThePreviousPeriod() before, but was renamed to better reflect the actual data returned). New tests were added to cover this.


coldrunKacper:
Approved.

PR #1052, opened 2016-08-05
PCHR-1340 : Fix adding and importing job roles errors

omarabuhussein:
## Problem 

when trying to import job roles , an HTTP 500 error page appears .

image

another problem is when trying to save or edit a job role :
civihr roles

here is a quote from the QA team :

> *The problem is: *
I add a Title, choose the contract that has start and end dates as well, try to save it, but then the Title and end date disappears
I try to edit it, enter a title and save it but the title still not getting saved and both dates get unspecified

## Solution

for the importer problem , I forget to change the method name in  CRM_Hrjobroles_BAO_HrJobRoles to contactExists but I was using this new name in the importer so I fixed this .

for the form problem , cost centers can be multiple and they are inserted in the database separated by vertical bar ( | ) so pusdocontant rule on cost center field doesn't accept that and through an error so i removed it.


davialexandre:
Approved

PR #1063, opened 2016-08-08
PCHR-799: convert (change reason) options to export label instead of value

omarabuhussein:
when exporting a job contract , the (change reason) should be exported as a label instead of as a value.


coldrunKacper:
Approved.

PR #1065, opened 2016-08-08
UY-51 - Fixing of fetching Job Contract entities (pay and health).

coldrunKacper:
The issue was caused by these two commits in the past:
https://github.com/civicrm/civihr/commit/f5da231fc6486558f97e98e4af6a15fb3937bfe7
and
https://github.com/civicrm/civihr/commit/4210a556cd231ffe5ee0dc9f511cb2492a0bae0d
which seem to move Pay and Health entity fetching below the rest of the entities. The issue was that for both entities there was used Details revision ID instead of particular entities revision.
I've moved Pay and Health fetching into the $q.all() method to get current valid revisions.


danilo-compucorp:
Approved

PR #1071, opened 2016-08-08
PCHR-1080 - Including future Job Contract to the advanced search results; Adding validation to Job Contract advanced search to skip deleted and overrided revisions.

coldrunKacper:
Extending SQL query for Contacts Advanced Search to include also Job Contracts with revision's future effective date (if Contacts don't have any currently active Job Contract).


omarabuhussein:
approved

PR #1072, opened 2016-08-09
PCHR-1080 - SQL query changing for Revision join.

coldrunKacper:
Changing two checks for past and future revisions to a single one using end of revision's effective date.


jamienovick:
Is this the change I suggested @coldrunKacper?

omarabuhussein:
@coldrunKacper  I can't really tell about the result of this change  , but the query looks ok .

coldrunKacper:
@jamienovick no, it is using effective_end_date for simlify the query. This field is also useful to be able to use date filters in Reports. Imagine that you have effective_date (start date) only and then you specify a date in People date filter. How can you be sure to get only one valid revision for a Job Contract (without using LIMIT 1 in Drupal filter)? I mean - it's possible but makes things a bit more complicated.

coldrunKacper:
Ok so I'll merge the PR as it doesn't affect any logical changes and will send an email to @jamienovick and @AkA84 regarding 'effective_end_date' purpose.

PR #1075, opened 2016-08-09
PCHR 1410: Update info.xml files

AkA84:
Updating the info.xml file of the extensions to update `<releaseDate>` and `<version>` that currently make up CiviHR. Also updated the `<develStage>` and `<ver>` fields where it made sense


omarabuhussein:
approved

PR #1078, opened 2016-08-09
PCHR-1351: Create a LeavePeriodEntitlement from an EntitlementCalculation

davialexandre:
This is the final step towards using the new data structure for leave period entitlements and leave requests.

The LeavePeriodEntitlement was updated with a new method, named saveFromCalculation(), that can be used to save period entitlements (including all the balance changes and leave requests required) from an EntitlementCalculation instance. This new method is used mainly by the Manage Entitlements page, when the user clicks on "Save new entitlements".

Also included in this PR is a small fix to the Manage Entitlements page. Now, If a contract has a previously calculated entitlement for the given period, which was overridden, the overridden value will be displayed on the list, instead of the newly calculated value. It will also be marked as overridden.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/Entitlement.php
```diff
@@ -1,249 +0,0 @@
-<?php
```
omarabuhussein: I am not sure what's going on here .. why is this entity get removed ? is this because the new data structure ?

davialexandre: Yes. It was replaced by the LeavePeriodEntitlement entity: https://github.com/civicrm/civihr/pull/1078/files/475ccfbec867d6df85cb9135720362d10e91b5a4#diff-c9b255fb4e8e91bee16a7b2970371439R70

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -118,6 +122,222 @@ public static function getPeriodEntitlementForContract($contractId, $periodId, $
...
+    global $user;
+    $absenceTypeID   = $calculation->getAbsenceType()->id;
+    $contractID      = $calculation->getContract()['id'];
+    $absencePeriodID = $calculation->getAbsencePeriod()->id;
+
+    $params = [
+      'type_id'     => $absenceTypeID,
+      'contract_id' => $contractID,
+      'period_id'   => $absencePeriodID,
+      'overridden'  => FALSE,
```
omarabuhussein: I would use ternary operator here

**'overridden'  => $params['overridden'] ? TRUE : FALSE ,**

davialexandre: there's no $params parameter for this method

omarabuhussein: sorry I meant :

**'overridden'  => $overriddenEntitlement? TRUE : FALSE ,**

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
```diff
@@ -174,4 +174,53 @@ public static function getNumberOfPublicHolidaysForCurrentPeriod($excludeWeekend
...
+
+    $tableName = self::getTableName();
+
+    $where = 'date >= %1 AND date <= %2 AND is_active = 1';
+
+    // Weekends are Saturday and Sunday
+    // So, to exclude them we return only the public holidays
+    // between monday (2) and friday (6)
+    if($excludeWeekends) {
+      $where .= ' AND DAYOFWEEK(date) BETWEEN 2 AND 6';
```
omarabuhussein: I know me , you and others used mysql date related functions in queries before , but what if mysql timezone is different than server timezone .. I think we should stop relaying on mysql date related functions in our code and pass whatever we want from php ... I think it is even faster from performance perspective due how mysql internally work and cache stuff .

davialexandre: I understand your point, but I don't think it applies here. The output of DAYOFWEEK is not affected by the timezone, since it just returns a number representing the day of the week for a given date.
I'm not sure about performance, I would be interesting to have some benchmarks. But I'm afraid MySQL can be faster sometimes since query results can be cached

omarabuhussein: yup it doesn't matter in this case ... my mistake

omarabuhussein:
approved

PR #1079, opened 2016-08-10
PCHR-1308: Update (current contract) retrieval for contact page header

omarabuhussein:
I've Update the query to retrieve the current contract for a contact to take **overrided** field into consideration ( this field is used to identify the current revision in case there are two revision have the same effective_date ) , I also updated the query to use a date passed from PHP instead of using the mysql CURDATE() function since in case the MySQL time configuration is different than php time configurations then it better to relay on php one . Finally I've added an **getcurrentcontract** API method which accepts a contact_id which is a wrapper for **getCurrentContract** BAO method to allow retrieving the current contract for a contact via the API if needed.


hrjobcontract/js/src/job-contract/services/contract.js
```diff
@@ -58,6 +58,29 @@ define([
 
                 return deffered.promise;
             },
+            getCurrentContract: function(contactId) {
```
brunoqueiros: Add a comment to these methods.

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -524,32 +524,42 @@ public static function getActiveContracts($startDate = null, $endDate = null)
...
+      ON hrjr.details_revision_id = hrjd.jobcontract_revision_id
+      WHERE hrjc.contact_id = %1
+      AND hrjr.effective_date <= '{$currentDate}'
+      AND hrjc.deleted = 0
+      AND hrjr.deleted = 0
+      ORDER BY hrjr.effective_date DESC , created_date DESC
+      LIMIT 1";
+    $response = CRM_Core_DAO::executeQuery($query, $queryParam);
+    if ($response->fetch())  {
+      if (empty($response->period_end_date) || $response->period_end_date >= $currentDate )  {
```
coldrunKacper: I'm just curious why don't you check period_end_date in SQL query instead of here?

hrjobcontract/api/v3/HRJobContract.php
```diff
@@ -179,3 +179,22 @@ function _civicrm_api3_h_r_job_contract_getlist_output($result, $request) {
...
 }
+
+/**
+ * HRJobContract.getcurrentcontract API
+ *
+ * @param array $params The accepted params are: contact_id
+ * @return array API result descriptor
+ */
+function civicrm_api3_h_r_job_contract_getcurrentcontract($params) {
+  $contactID = !empty($params['contact_id']) ? (int) $params['contact_id'] : 0;
```
coldrunKacper: Just a question - do we allow go get current contract of user with ID = 0? Maybe just throw an exception saying that contact_id can't be empty?

coldrunKacper:
A backend note: we currently have 'getcurrentrevision' method for HRJobContractRevision entity which may be not useful for you (regarding the task requirements) but you can compare its logic to the one applied by your getCurrentContract() BAO method to make sure that both ways of getting current contract / revision are using the same logic.

omarabuhussein:
> A backend note: we currently have 'getcurrentrevision' API call for HRJobContractRevision entity which may be not useful for you (regarding the task requirements) but you can compare its logic to the one applied by your getCurrentContract() BAO method to make sure that both ways of getting current contract / revision are using the same logic.

apparently they are very similar but there are some stuff confusing with getcurrentrevision that i have to clarify with you ... but the problem is getcurrentrevision expect contract ID where getcurrentcontract expect contact id as a parameter so I cannot use it here . I will leave it until we finish discussing all the points about job contracts and agree on how to move forward .

@coldrunKacper

hrjobcontract/js/src/job-contract/services/contract.js
omarabuhussein: I've added that in this new PR https://github.com/civicrm/civihr/pull/1084

hrjobcontract/api/v3/HRJobContract.php
omarabuhussein: yes throwing an exception is a better idea , I've updated this in the new PR : https://github.com/civicrm/civihr/pull/1084

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
omarabuhussein: that won't work , to achieve the same behavior with SQL only I have to write a nested select statement ... Imagine the following scenario to understand why what you are suggesting won't work : 
1- suppose the current day date is 2016-08-13
2 - add a contract with start date = 2016-08-01 and end date = 2016-08-31 . (the effective date will automatically be 2016-08-01)
3- create a new revision with start date = 2016-08-01 and end date = 2016-08-11 (less than today date which 2016-08-13) and set the effective date to (2016-08-04).

now since the new revision effective date = 2016-08-04 so it is the most current revision but since the end date for the contract is 2016-08-11 which is in the past ( remmber that we supposed that today date is 2016-08-13 ) then this user will not have any active contract since the most recent revision end date is in the past .

if I want to update the query according to you suggestion to look like this :

```php
SELECT hrjc.id as contract_id , hrjd.*
      FROM civicrm_hrjobcontract hrjc
      LEFT JOIN civicrm_hrjobcontract_revision hrjr
      ON hrjr.jobcontract_id = hrjc.id
      LEFT JOIN civicrm_hrjobcontract_details hrjd
      ON hrjr.details_revision_id = hrjd.jobcontract_revision_id
      WHERE hrjc.contact_id = %1
      AND hrjr.effective_date <= '{$current_date}'
      AND (hrjd.period_end_date  >= '{$current_date}'OR hrjd.period_end_date  IS NULL)
      AND hrjc.deleted = 0
      AND hrjr.deleted = 0
      ORDER BY hrjr.effective_date DESC , created_date DESC
      LIMIT 1 
 ```

then the output will be the first revision since it meet the criteria ( effective date <= current date .. and end date >= current date ) ... which is not we want , we want it to return null since this contract is not an active contract . to solve this problem I can rewrite it and use nested SQL statements but the current method is just easier and have better performance .

PR #1080, opened 2016-08-10
PCHR-1409: Future jobcontract details export

coldrunKacper:
I've fixed an order of results returned by JOIN contact with job contract to get a proper current or nearest job contract and it's revision. Previously the correct revision was picked for a first job contract found by JOIN query.


hrjobcontract/CRM/Hrjobcontract/BAO/Query.php
```diff
@@ -314,13 +314,21 @@ function whereClauseSingle(&$values, &$query) {
...
+              . " SELECT hjc.* FROM civicrm_hrjobcontract hjc "
+              . " LEFT JOIN civicrm_hrjobcontract_revision hjr ON hjr.jobcontract_id = hjc.id AND hjr.deleted = 0 "
+              . " LEFT JOIN civicrm_hrjobcontract_details hjd ON hjd.jobcontract_revision_id = hjr.details_revision_id "
+              . " WHERE hjd.period_end_date >= '{$todayDate}' OR hjd.period_end_date IS NULL "
+              . " ORDER BY hjd.period_start_date ASC, hjr.effective_date ASC "
+              . " ) AS hrjobcontract ON hrjobcontract.contact_id = contact_a.id AND hrjobcontract.deleted = 0 "
+              . " $side JOIN civicrm_hrjobcontract_revision rev ON rev.jobcontract_id = hrjobcontract.id "
+              . " AND ("
+              . "   rev.id = (SELECT id FROM civicrm_hrjobcontract_revision WHERE jobcontract_id = hrjobcontract.id "
+              . "   AND (effective_end_date >= '{$todayDate}' OR effective_end_date IS NULL) AND deleted = 0 AND overrided = 0 ORDER BY effective_date ASC LIMIT 1) "
```
omarabuhussein: I am not really aware of the full context for this query so i may be wrong  but are you sure of this :

**ORDER BY effective_date ASC**
 because if you trying to get the most current revision it should be "DESC" instead "ASC" also i think since you are checking effective_end_date it is not need at all

omarabuhussein:
also please fix the PR title

hrjobcontract/CRM/Hrjobcontract/BAO/Query.php
coldrunKacper: ORDER BY effective_date ASC is needed to get a first revision which meets the other requirements. Whithout that we could end up getting a future revision even if there is a currently valid revision present.

omarabuhussein:
complex queries  like this is hard to review properly without investigating the full context and do some testing .. but if there is anything it will effect the result of the search only and QA should detect any problem in case there is ... so I think it is fine to be merged .

PR #1084, opened 2016-08-13
PCHR-1308: Update contact page tab header according to changes in contracts list

omarabuhussein:
## Problem :

When a contract get added/updated/corrected/deleted the contact page header which color doesn't change as well as the contract info and the user have to refresh the page to see the changes . Also the query to fetch the current contract for the contact/staff was not accurate.

## Solution :

I've updated the current query to fetch the current contract which is defined by : 

> **the contract with current revision start date is on or before the current date and end date
>  (is more than or equal the current date) or (null/empty).**

also I'v updated the angular scripts for job contract to fetch the current contract via ajax request in case the contract get added/updated/corrected/deleted and change the header color and current contract info based on the result .


coldrunKacper:
Approved (backend part).

brunoqueiros:
Approved!

PR #1087, opened 2016-08-15
PCHR-1239: Cleaning and fix Hrjobcontract Extenstion tests

omarabuhussein:
## Problem
 
Some jobcontracts unit tests were not working.

## Solution 

I walked through the extension tests and here is the changes I've made :

1- removing the following test classes :

- CRM/Hrjobcontract/AddRevisionTest.php
- CRM/Hrjobcontract/EditRevisionTest.php
- CRM/Hrjobcontract/CreateRevisionTest.php

which I found useless.

2- refactoring CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php to extend PHPUnit_Framework_TestCase and implement the headless interface instead of extending the slow CiviUnitTestCase class , also I still not convinced with the test cases inside it but I will keep it for now until Hrjobcontract data structure and implementation being revisited and everything is clear about it.

3- removing HRJobContractTestBase.php which is not needed anymore.

4- replacing CRM/Hrjobcontract/ValidateDatesTest.php with CRM/Hrjobcontract/BAO/HRJobDetailsTest.php and use BAO methods inside it instead of API calls.

5- removing testGetActiveContractsReturnsActiveContracts() method from  api/v3/HRJobContractTest.php since CRM/Hrjobcontract/BAO/HRJobContractTest.php tests the same thing.


hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -1,32 +1,26 @@
...
-      'civicrm_hrjobcontract_role',
-      'civicrm_hrjobcontract_hour',
-      'civicrm_hrjobcontract_pay',
-      'civicrm_hrjobcontract_leave',
-      'civicrm_hrjobcontract_pension',
-      'civicrm_hrjobcontract_revision',
-      'civicrm_hrjobcontract',
-    ));
     $upgrader = CRM_Hrjobcontract_Upgrader::instance();
     $upgrader->install();
```
davialexandre: Since this is part of the extension installation, you can move those two lines to the setupHeadless method. It will avoid running the upgrader after every test method and will probably save you a few precious seconds

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -0,0 +1,78 @@
...
+  use HRJobContractTestTrait;
+
+  public function setUpHeadless() {
+    return \Civi\Test::headless()->installMe(__DIR__)->apply();
+  }
+
+  function setUp() {
+    $upgrader = CRM_Hrjobcontract_Upgrader::instance();
+    $upgrader->install();
+  }
```
davialexandre: You can also move this to setupHeadless

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+  function testValidateDatesWithDifferentDateSets() {
+    // Create test Contact.
+    $contactParams = array("first_name" => "chrollo", "last_name" => "lucilfer");
```
davialexandre: Since the rest of the code is using the short array syntax, please use it here as well

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -0,0 +1,78 @@
...
+    $contactID =  $this->createContact($contactParams);
+
+    // Create Two test Job Contracts.
+    $this->createJobContract($contactID, '2016-01-01', '2016-01-10');
+    $this->createJobContract($contactID, '2016-02-01');
+
+    // Set of tests against invalid dates period (there are Job Contracts
+    // already with period dates overlapping.
+
```
davialexandre: I think it would be better to use a [Data Provider](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) instead of putting the array inside the test method. It would also make thins clearer

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $params['period_end_date'] = $datePair['period_end_date'];
+      }
+
+      $result = CRM_Hrjobcontract_BAO_HRJobDetails::validateDates($params);
+
+      $this->assertFalse($result['success']);
+    }
+
+    // Now we test against valid dates so we expect the result of
+    // 'validateDates' call to be TRUE.
```
davialexandre: It would be better to split this into two (or even more) tests. One checking what is the expected output for invalid dates and another one checking the output for valid dates

davialexandre:
Would you care to explain why do you think AddRevisionTest.php, EditRevisionTest.php and CreateRevisionTest.php are useless? They're are far from what I would call good tests, but, if they're are correct (in other words, they will fail if the code is wrong), it's better to keep them than removing. Maybe it would be better to rewrite them in a proper way? This would be a good chance to have good examples of how this revisions logic is supposed to work and add coverage to such a complicate piece of code.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
omarabuhussein: I will remove them ...  when installing an extension the upgrader install() method will be automatically called  ... but not adding these two lines was causing my tests not to work and a "table not found" error will be thrown ... I just recreated my test database and removed these lines from all the tests and everything working fine now .. so apparently  something was wrong with my test database.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
omarabuhussein: same as my reply above

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
omarabuhussein: oh .. I copied this from another test and forget to change it

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
omarabuhussein: well I wasn't really certain what do in this case , since I initialy thought to create a seperate test case for each start,end date pair  but that would be too much and choosing names for test cases will be a bit tricky .. but separating them into two test methods sounds like a good option.

omarabuhussein:
> Would you care to explain why do you think AddRevisionTest.php, EditRevisionTest.php and CreateRevisionTest.php are useless? They're are far from what I would call good tests

these are really test nothing , look for example on the at the expected array :

```
$expected = array(		
 "id" => "8",		
"jobcontract_id" => "1",		
"details_revision_id" => "1",		
 "health_revision_id" => "3",		
"role_revision_id" => "4",		
 "hour_revision_id" => "5",		
 "pay_revision_id" => "6",		
  "leave_revision_id" => "7",		
 "pension_revision_id" => "8",		
  );
```
which is testing auto-generated database IDs and expect them to be the same for each test which doesn't make sense and will usually fail .


I thought about writing them in a proper way , but the problem that we are currently in the process of revisiting job contract data structure and implementation .. I expect many changes may happen so investing time on writing tests for this may be a waste of energy ... 

an example of that , the current way of creating job contracts should be inside a transaction and in order :
- creating a record in hrjobcontract table.
- creating a record in hrjobcontract_revision table
- creating a recored in hrjobcontract_details table
- creating a recored in hrjobcontract_pay table
- creating a recored in hrjobcontract_hours table
... etc ( hence that even me not still sure about the order above ) .

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -18,61 +18,77 @@ public function setUpHeadless() {
...
+  public function invalidDatesProvider()
+  {
+    return [
+      ['2016-01-01', '2016-10-10'],
+      ['2016-01-03', '2016-01-05'],
+      ['2016-01-01', '2016-01-02'],
+      ['2016-01-31', '2016-02-01'],
+      ['2016-02-10', '2016-02-20'],
+      ['2015-01-01'],
+      ['2016-03-01'],
```
davialexandre: The testValidateDatesWithInvalidDates is failing because of the missing values for end_date in the last two itens of the data provider.  You can fix it by passing `null` as the end_date value:
captura de tela 2016-08-15 as 15 46 43

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -18,61 +18,77 @@ public function setUpHeadless() {
...
+      $params['period_end_date'] = $end_date;
     }
+
+    $result = CRM_Hrjobcontract_BAO_HRJobDetails::validateDates($params);
+
+    $this->assertTrue($result['success']);
+  }
+
+  public function invalidDatesProvider()
+  {
```
davialexandre: The Drupal standard is to have the opening brace at the same line as the function

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
```diff
@@ -18,61 +18,77 @@ public function setUpHeadless() {
...
+  public function testValidateDatesWithValidDates($start_date, $end_date) {
 
-    foreach ($sampleCorrectDates as $datePair)  {
-      $params = ['contact_id' => $contactID, 'period_start_date' => $datePair['period_start_date']];
-      if (!empty($datePair['period_end_date']))  {
-        $params['period_end_date'] = $datePair['period_end_date'];
-      }
+    // Create test Contact.
+    $contactParams = array("first_name" => "chrollo", "last_name" => "lucilfer");
+    $contactID =  $this->createContact($contactParams);
```
davialexandre: With the method split, these two lines are now repeated. What about moving them to a setUp method?

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobDetailsTest.php
omarabuhussein: I run them before committing and they worked without problem ... Let me try again and fix it.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
davialexandre: I'm not sure about the upgrader running during the extension installation. For one of the tasks in L&A I needed to add a new field to hrjobleave and, in order to have it available in my tests, I had to run the upgrader manually in a setupHeadless method. Maybe you didn't have the same problem because the tests didn't need any of changes made by the upgrader?

omarabuhussein: can you show me the test you are talking about in L&A ?

davialexandre:
Expecting the IDs to be same in every test DOES make sense if you have a way to reset the auto_incremet value, which is exactly what TRUNCATE, which was called after every test, does. So, even if the tests were bad, they're supposed to work.

Anyway, I didn't know hrjobcontract was about to be refactored. In that case I think it makes sense to not spend time on this.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
davialexandre: Here https://github.com/civicrm/civihr/blob/1.7-wip/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/EntitlementCalculationTest.php#L21-L27

omarabuhussein: can you please show the field you added and the upgrader method that add it

omarabuhussein: because if you checked these files respectively , you can see than install method run at extension installation :

https://github.com/civicrm/civihr/blob/1.7-wip/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.civix.php#L54

https://github.com/civicrm/civihr/blob/1.7-wip/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Base.php#L243

https://github.com/civicrm/civihr/blob/1.7-wip/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Base.php#L257

davialexandre: The field: https://github.com/civicrm/civihr/blob/1.7-develop/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php#L186-L192
The upgrader method: https://github.com/civicrm/civihr/blob/1.7-develop/hrjobcontract/CRM/Hrjobcontract/Upgrader.php#L1234-L1239

davialexandre: You're right here. The upgrade actually runs during the extension installation. I did a quick test here by removing the manual upgrader call on EntitlementCalculationTest and the tests still pass. I probably solved the problem with something else and thought it was fixed by the manual call :(

omarabuhussein: the problem is actually something else :) ... can you take a look at this https://compucorp.atlassian.net/wiki/pages/viewpage.action?pageId=80478227

omarabuhussein: I noticed that we the  upgrader  in a wrong way ... generally , adding stuff via the upgrader should be done only for exsiting installations ... for new installations you should add it to the install.sql script

davialexandre: So, if I add a new field to an existing extension, I would need to:
- Add the field to the schema XML 
- Regenerate the DAO to get the updated field
- Add the field to the _install.sql file
- Add an upgrader method to update existing installations

Is that correct?

Also, if the above is correct, since the upgrader runs during the installation, wouldn't the upgrade method conflict with the field which was already added by the _install.sql file?

omarabuhussein: > The upgrade actually runs during the extension installation. I did a quick test here by removing the manual upgrader call on EntitlementCalculationTest and the tests still pass. I probably solved the problem with something else and thought it was fixed by the manual cal

the thing is the new field is already created ... if you now re-created your test database from scratch (you can follow what's written here https://wiki.civicrm.org/confluence/display/CRM/Setting+up+your+personal+testing+sandbox+HOWTO#SettingupyourpersonaltestingsandboxHOWTO-Createatestdatabase )

and then run your tests without executing the upgrader in the headless or setUp method they will fail ... if you then added the lines to execute the upgrader and run the tests again they will work ... if you then removed upgrader execution lines and run the tests again they will work .. because the field is created in the second tests run ... transaction interface which you implemented in your tests doesn't work very well when altering the schema so the field will stay even after tests execution finish running.

davialexandre:
Looks good now. Approved

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
omarabuhussein: just adding a part of what we discussed at hipchat :

> So, if I add a new field to an existing extension, I would need to:
Add the field to the schema XML
Regenerate the DAO to get the updated field
Add the field to the _install.sql file
Add an upgrader method to update existing installations
Is that correct? 

Yes 

> Also, if the above is correct, since the upgrader runs during the installation, wouldn't the upgrade method conflict with the field which was already added by the _install.sql file?

when installing and extension , civicrm will automatically set the version for that extension in civicrm_settings table to the highest upgrader verrsion(number) .. so the upgrader will not run for new installtions and only _install.sql file will get executed so on conflict should happen.

PR #1088, opened 2016-08-16
PCHR-1241: Fix hrcaseutils unit tests

omarabuhussein:
## Problem 

hrcaseutils unit tests don't pass testing.

## Solution

- First I changed the test class to extend PHPUnit_Framework_TestCase instead of the slow   CiviUnitTestCase  and any use to it inside the test class.

- I removed the creation of case types and activity types inside the tests since they automatically get created by civicrm via hook_civicrm_caseTypes extenstion  when the extension get installed . I also removed /tests/phpunit/api/v3/HRCaseUtils/Hrdata.xml which is no longer needed after this change , and removed the underscore from "Background Check" to be the same in all the extenstion ( another option I had is to keep it and change the test class to check for Background_Check instead but it does not really matter if you want to keep it or not ).

- I cleaned some stuff such as the usage of deprecated activity_type API and replaced it with option_value API.


coldrunKacper:
Approved.

coldrunKacper:
Approved.

PR #1089, opened 2016-08-16
PCHR-902: Fix the visibility of the "Expire after certain duration" dropdown

davialexandre:
**Problem**
When opening or editing an existing Absence Type which allows carry forward, the "Never expire" option was always selected, even if the Absence Type had an expiry duration set.

**Solution**
 1. Fix the template code which adds a "checked" attribute to the "Never expire" radio ONLY if the Absence Type doesn't have an expiry unit and an expiry duration. It was checking the wrong variables, making the attribute to always be added to the radio.
 2. Fix the template code to add the "checked" attribute to the "Expire after certain duration" radio only if the Absence Type have an expiry unit or expiry duration.
 3. Fix the piece of javascript code which runs after the page load and shows the expiry unit and expiry duration fields if "Never expire" is not checked.

So, when opening/editing an existing Absence Type, the visibility of the carry forward expiry fields should follow these rules:
- If the Absence Type doesn't have an expiry duration
 - "Never expire" should be checked
 - "Expire after certain duration" should not be checked. The respective text and dropdown fields should not be visible

- If the Absence Type have an expiry duration
 - "Never expire" should not be checked
 - "Expire after certain duration" should be checked. The respective text and dropdown fields should be visible

New tests were added to cover these rules.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
@@ -126,12 +126,12 @@
                     <div class="label">{ts}Carry forward leave expiry{/ts}</div>
                     <div class="content">
                         {assign var="carry_forward_never_expire" value=false }
-                        {if not $carry_forward_expire_after_duration and not $carry_forward_expire_after_date }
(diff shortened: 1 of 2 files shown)


coldrunKacper:
Approved.

davialexandre:
This fix should be based on the PCHR-898 branch and be merged on 1.7-develop. Closing in favor of https://github.com/civicrm/civihr/pull/1093

PR #1092, opened 2016-08-17
PCHR-1477: Remove Cases tab if T&A is enabled

AkA84:
**Before**
before

**After**
after

Also refactored the code a bit as well

### DIFF
--- a/hrui/hrui.php
+++ b/hrui/hrui.php
@@ -38,10 +38,7 @@ function hrui_civicrm_pageRun($page) {
     CRM_Core_Resources::singleton()
       ->addScriptFile('org.civicrm.hrui', 'js/contact.js')
       ->addScriptFile('org.civicrm.hrui', 'js/hrui.js')
-      ->addSetting(array('pageName' => 'viewSummary'))
-      ->addSetting(array('extensionsEnabled' => array(
-        'taskAssignments' => !!CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Extension', 'uk.co.compucorp.civicrm.tasksassignments', 'is_active', 'full_name')
-      )));
+      ->addSetting(array('pageName' => 'viewSummary'));
 
     //set government field value for individual page
     $contactType = CRM_Contact_BAO_Contact::getContactType(CRM_Utils_Request::retrieve('cid', 'Integer'));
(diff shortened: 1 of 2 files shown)


hrui/hrui.php
```diff
@@ -33,13 +33,19 @@ function hrui_civicrm_pageRun($page) {
...
+
   if ($page instanceof CRM_Contact_Page_View_Summary) {
     CRM_Core_Resources::singleton()
       ->addScriptFile('org.civicrm.hrui', 'js/contact.js')
       ->addScriptFile('org.civicrm.hrui', 'js/hrui.js')
-      ->addSetting(array('pageName' => 'viewSummary'));
+      ->addSetting(array('pageName' => 'viewSummary'))
+      ->addSetting(array('extensionsEnabled' => array(
+        'taskAssignments' => !!CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Extension', 'uk.co.compucorp.civicrm.tasksassignments', 'is_active', 'full_name')
+      )));
```
omarabuhussein: you can wrap checking for T&A if enabled or not in a seperate method , something like what i did here for hrcase https://github.com/civicrm/civihr/blob/ab6a7b72984a481717f2535bbeaf90f0dce98ea1/hrcase/hrcase.php#L506

also it's better not to do my mistake and make the method receive any extension name ( for more usability in the feature )

omarabuhussein:
I believe there is a better way to do this rather than using javascript to achieve that ... there is this hook which you can use to do that , there is even an example to remove the contribition tab https://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_tabset

hrui/hrui.php
```diff
@@ -399,24 +396,44 @@ function hrui_civicrm_upgrade($op, CRM_Queue_Queue $queue = NULL) {
...
 
 /**
- * Implementation of hook_civicrm_tabs
+ * Check if the extension with the given key is enabled
+ *
+ * @param string $extensionKey
+ * @return boolean
+ */
+function _hrui_check_extension($extensionKey)  {
+  return !!CRM_Core_DAO::getFieldValue(
```
omarabuhussein: you can cast the value to boolean :

return (boolean) CRM_Core_DAO::getFieldValue ... etc 

it's more clear like that.

omarabuhussein:
just one minor note which is up to you to do or not , otherwise approved

jamienovick:
We should definitely do this via hook not via javascript please. Can we make sure that we are using the hooks as much as possible? J

AkA84:
@jamienovick it was indeed done via hook after @omarabuhussein 's comment

PR #1094, opened 2016-08-18
PCHR-1433: New vacancy style fix

brunoqueiros:
This PR update some style adjustments pointed at the ticket.

- Style some button with the primary color (aqua blue)
- Adjust some spacings
- Style the block's title (Stages, Application Form, Evaluation Form, Permission)
- Add a calendar icon on the datepicker's input
- Style the trash & spanner icons.
- Adjust inputs & dropdowns width

***Before***
image

***After***
image


org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
```diff
@@ -318,3 +318,44 @@ $button-border-radius: 2px;
     }
   }
 }
+
+%btn-primary {
```
AkA84: Why are you redefining `btn-primary`? Isn't it already available, given that this partial is importing it from the theme extension?
```scss
@import '../../../org.civicrm.bootstrap/scss/vendor/bootstrap/buttons';
 @import '../bootstrap/overrides/style/buttons';
```
(same goes with `%btn-default` below)

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
```diff
@@ -318,3 +318,44 @@ $button-border-radius: 2px;
...
+
+  &:hover {
+    background: darken($brand-primary, 10%) !important;
+
+    > input.crm-form-submit {
+      border: 1px solid darken($brand-primary, 10%) !important;
+    }
+  }
+
+  > input.crm-form-submit {
```
AkA84: So you are implicitly expecting that whatever is extending `%btn-primary` will contain a `.crm-form-submit`? Seems very specific. Could you give here an example of what element you are using this on?

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
brunoqueiros: Because this new `%btn-primary` uses the `!important` rule, without it the style don't work.

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
brunoqueiros:
```html
<span class="crm-button crm-button-type-next crm-button_qf_HRVacancy_next crm-i-button">
  <i class="crm-i fa-check"></i>
  <input class="crm-form-submit default validate" accesskey="S" crm-icon="fa-check" name="_qf_HRVacancy_next" value="Save" type="submit" id="_qf_HRVacancy_next">
</span>
```

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -152,4 +152,13 @@
       }
     }
   }
+
+  .collapsible-title {
```
AkA84: This seems a very generic style. Are you sure it should be relegated to the vacancies page only, and not be move in the `common/` folder instead?

hrui/js/hrui.js
```diff
@@ -12,7 +12,23 @@
       }
     });
   }
+
+  /**
+   * Update label 'for' attr to works with the datepicker
+   *
+   * @param  {jQuery Obejct} line [tr parent element]
+   */
+  function updateLabelFor(line) {
```
AkA84: This function name is extremely generic given the very specific job that it does (updating the `[for]` attr to have the same value as the `[id]` attr of the datepicker field)

Maybe something like:

```js

/**
 * ....
 *
 * @param {jQuery object} input the input field the label should linked to
 */
function linkLabelToDatepickerInput($input) {

}

```
and also, why passing the entire line to the function? If you know already what element you're looking for, why not passing it directly to the function (like above)?

hrui/js/hrui.js
```diff
@@ -12,7 +12,23 @@
...
+    line.find('label').attr('for', line.find('.crm-form-date').attr('id'));
+  }
+
   $(document).on('crmLoad', function(e) {
+    if ($('.CRM_HRRecruitment_Form_HRVacancy').length === 1) {
+      var lines = $('.CRM_HRRecruitment_Form_HRVacancy tbody:first-child tr');
+      updateLabelFor($(lines.get(7)));
```
AkA84: I know this is all temporary, but targeting an element just based on the parent's position doesn't seem very reliable. Isn't there a better way to target?

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
brunoqueiros: This style is under `.page-civicrm-vacancy` namespace, I don't know if I move to `common/` folder it'll broke another page, it's better stay under `.page-civicrm-vacancy` and move after if is necessary.

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -114,4 +114,42 @@
     margin: 0 !important;
     min-height: 35px !important;
   }
+
+  .CRM_HRRecruitment_Form_HRVacancy {
+    tbody:first-child {
+      tr:nth-child(8),
```
AkA84: As for javascript part, you're targeting a specific type of element only by using the parent's position. Isn't there something in the markup to identify fields for which you need to add the icon?

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -1,31 +1,169 @@
...
+
+  .crm-submit-buttons {
+    padding: 20px !important;
+    margin: 0 !important;
+    min-height: 35px !important;
+  }
+
+  .CRM_HRRecruitment_Form_HRVacancy {
+    tbody:first-child {
+      tr:nth-child(8),
```
AkA84: As for javascript part, you're targeting a specific type of element only by using the parent's position. Isn't there something in the markup to identify fields for which you need to add the icon?

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
AkA84: Might be worth spending 10 minutes checking now, because usually this "will do later" improvements will never be done

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -1,31 +1,169 @@
...
+        top: -3px;
+        position: relative;
+        font-family: 'FontAwesome';
+        font-weight: normal;
+        color: $brand-danger !important;
+      }
+    }
+  }
+
+  .crm-option-edit-link.crm-hover-button {
```
AkA84: Another style that seems could belong to `common/`

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
AkA84: What if instead you use a more specific selector it? Would that be enough to ensure your styles get applied on top?

AkA84:
Approved

PR #1096, opened 2016-08-19
PCHR-1496: Add the "add_public_holiday" field during the extension in…

davialexandre:
**Problem**

The piece of code that adds the field was in an upgrade method, which is not called during the extension installation, causing the field to not be available right after a new install.

**Solution**

The code that adds the field was also added to the extension's hook_civicrm_install implementation.

### DIFF
--- a/hrjobcontract/hrjobcontract.php
+++ b/hrjobcontract/hrjobcontract.php
@@ -113,6 +113,9 @@ function hrjobcontract_civicrm_install() {
     ));
   }
 
+  // Adding 'add_public_holidays' field to 'civicrm_hrjobcontract_leave' table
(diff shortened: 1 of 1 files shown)


omarabuhussein:
to be honest I just checked the code for job contract and their is no need for that ... hrjobcontract run the upgrader methods on installation by calling upgradeBundle() method https://github.com/civicrm/civihr/blob/develop/hrjobcontract/CRM/Hrjobcontract/Upgrader.php#L13 ... and inside upgradeBundle you already call the upgrader method create this field which is upgrade_1005 https://github.com/civicrm/civihr/blob/develop/hrjobcontract/CRM/Hrjobcontract/Upgrader.php#L939

PR #1101, opened 2016-08-22
PCHR-1186: Clean old hrjob extenstion usage

omarabuhussein:
I investigated any usage in any of civihr extensions for the old hrjob extension and removed it as well as some correction for some job contract extension doc blocks.

For now I left one usages for hrjob tables inside hrreport extension tests which is going to be removed so no need to clean it .


coldrunKacper:
Approved.

coldrunKacper:
Approved.

PR #1104, opened 2016-08-23
PCHR-1443: Create the work pattern get leave days for date method

davialexandre:
This method will return the leave_days amount for any given date, in a period given by a start and an end date. The main use case for it is to calculate the amount of days to be deducted if a staff member using this work pattern takes a leave on that date.

It will rotate through the pattern's weeks to get the leave_days value. That is, starting from the Start Date, if the date falls on the first week, we get the leave_days amount from the first week of the pattern; if it falls on second week, we get it from the pattern's second week; if it's on the third week, we rotate and get the value from the pattern's first week again and so on.

While doing this, I noticed some of the WorkPattern tests were failing. This PR also includes fixes for them. I also did some overall refactoring to the extension tests to make it easier/faster to run them all at once.


omarabuhussein:
just the notes I left , otherwise approved.

PR #1107, opened 2016-08-24
PCHR-965: Job roles dont fetch contacts

deb1990:
# Problem: 
1. When Trying add funder to job role , only 25 contacts are shown.
2. typeahead widget is used which is not consistent with newer ui-select. 

# Fix:
1. Using ajax to fetch contacts based on what user types.
2. Implemented ui-select for the dropdown.

# Other Fixes:
1. Code optimized for Javascript to fetch data.
2. _config.scss was missing in jobroles and jobcontracts and because of that css was not getting generated from scss. Reverted the _config.scss file.

# Result
pchr-965


com.civicrm.hrjobroles/css/hrjobroles.css
```diff
@@ -4711,7 +4707,7 @@ fieldset[disabled]
 }
 @media (min-width: 768px) {
   #hrjobroles .form-horizontal .form-group-lg .control-label {
-    padding-top: 11px;
+    padding-top: 14.333333px;
```
danilo-compucorp: Why the fractional pixels?

danilo-compucorp:
Approved

PR #1109, opened 2016-08-24
Pchr 1084 implement new design for export contact

danilo-compucorp:
This PR improves the layout for the "export contact" page. It was also used to do some code refactorings.
Please note that it also implements the "custom radio buttons and checkboxes" as per the invision layout. In order to customize these elements, the standard inputs are hidden, and a font-awesome icon will replace it. 
This is possible since the sibling _label_ (present immediately after the _input_) contains the _for_ attribute for the matching radio/checkbox id, and as this condition is not always present, a javascript file ( checkbox.js ) was added to try to add the missing _for_ attribute, or to make the standard radio/checkbox visible again when there's no sibling _label_.
In other words, when we have the following:
`<input type="radio" id="radio-id">
<label for="radio-id">Test</label>`
The radio button would be hidden ( `display: none` ), and a font-awesome icon would be added before the label, replacing the radio button.

**Example**:
radio

In order to avoid issues with markup inconsistencies, if the _input_ element lacks the _id_ attribute, the script would add a random id to it. 
For example: 
from this 
`<input type="radio">` 
to this 
`<input type="radio" id="1472062278795">`

If the _label_ element lacks the _for_ attribute, the script would add it, setting its value to the matching _input_ id.
For example: 
from this 
`<input type="radio" id="radio-id">
<label>Test</label>`
to this 
`<input type="radio" id="radio-id">
<label for="radio-id">Test</label>`

If a _label_ is not found immediately after the _input_, the script will make it visible, so we don't lose any feature. 

**Before**:
before

**After**:
after2


AkA84:
> Please note that it also implements the "custom radio buttons and checkboxes" as per the invision layout. In order to customize these elements, the standard inputs are hidden, and a font-awesome icon will replace it. This is possible since the sibling label contains the for attribute for the matching radio/checkbox id, and as this condition is not always present, a javascript file ( checkbox.js ) was added to try to add the missing for attribute, or to make the standard radio/checkbox visible again when there's no sibling label.

please provide code snippets showing what you are talking about, it makes easier to follow the description

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -128,3 +128,10 @@ function bootstrapcivicrm_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
 function bootstrapcivicrm_civicrm_pageRun($page) {
   CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css');
 }
+
+/**
+ * This hook is invoked after all the content of a CiviCRM form or page is generated. It allows for direct manipulation of the generated content.
+ */
+function bootstrapcivicrm_civicrm_alterContent( &$content, $context, $tplName, &$object ) {
+  $content = '<script src="'.CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.bootstrapcivicrm').'/js/radio-checkbox.js"></script>'. $content;
```
AkA84: Why not adding this script in the `_pageRun` hook and just wrap the script in a `$(function () {})` to make sure that it's run only when the document is ready? (i don't know if `CRM.$` is equivalent to that)

org.civicrm.bootstrapcivicrm/js/radio-checkbox.js
```diff
@@ -0,0 +1,24 @@
...
+   * This is a workaround to add any missing "for" attributes to the markup.
+   * @see: org.civicrm.bootstrapcivicrm/scss/civicrm/common/_radio-checkbox.scss
+   */
+  CRM.$('.crm-container input[type=checkbox], .crm-container input[type=radio]').each(function() {
+    var $this = CRM.$(this);
+    if (!$this.attr('id')) {
+      // There's no id. Add a unique random value as the id
+      $this.attr('id', new Date().valueOf());
+    }
+    var label = $this.next('label');
```
AkA84: Let's move all `var` definitions at the top
```js
var $this = CRM.$(this);
var label = $this.next('label');

if (!$this.attr('id')) {
// ...
```

org.civicrm.bootstrapcivicrm/js/radio-checkbox.js
```diff
@@ -0,0 +1,24 @@
...
+    if (label.length === 0) {
```
AkA84:
```js
if (!label.length) {
```
?

org.civicrm.bootstrapcivicrm/js/radio-checkbox.js
```diff
@@ -0,0 +1,24 @@
...
+      // There's no sibling label. Let's just show the checkbox / radio button, in order to avoid feature loss
+      $this.show();
+    } else if (!label.attr('for')) {
+      label.attr('for', $this.attr('id'));
```
AkA84: No comment/explanation for this one?

org.civicrm.bootstrapcivicrm/scss/bootstrap/mixins/_fa-icon.scss
```diff
@@ -1,17 +1,18 @@
-@mixin fa-icon($font-size: null, $content: null, $color: null) {
+@mixin fa-icon($args...) {
```
AkA84: I know what this change does, but I'm curious if this solve anything or is a plain code improvement?

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_common.scss
```diff
@@ -7,3 +7,11 @@
 @import 'alerts';
 @import 'modals';
 @import 'datepickers';
+@import 'radio-checkbox';
+
+.crm-container {
```
AkA84: Any partial you could put this in? It was nice to have the `_common` file just being a collection of partials

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_radio-checkbox.scss
```diff
@@ -0,0 +1,36 @@
+.crm-container {
+  input[type=checkbox],
+  input[type=radio] {
+    display: none;
+  }
+  input[type=checkbox] + label[for],
```
AkA84: Put a whiteline between a rule and another

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
```diff
@@ -36,47 +36,33 @@ $button-border-radius: 2px;
   .crm-form-block {
     &.crm-search-form-block > .crm-submit-buttons,
     &.crm-form-block > .crm-submit-buttons {
-      padding-top: $toolbar-top-space;
-      background: $crm-accordion-header-color !important;
-
-      %btn {
```
AkA84: I think @brunoqueiros has been working on this partials as well for his work on PCHR-1429, you might want to check his branch (and talk with him) to see if you're gonna have any conflicts down the line and just to make sure that the work you are both doing on this is consistent

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
danilo-compucorp: The `_pageRun` hook doesn't work in this case, that's why I'm using `_alterContent`. And yes, `CRM.$` is equivalent, and as it's wrapping the function, it waits for the document to be ready, and actually doesn't depend on where it's inserted in the page.

org.civicrm.bootstrapcivicrm/js/radio-checkbox.js
danilo-compucorp: Could also be

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
AkA84: > The _pageRun hook doesn't work in this case
Doesn't work as in...? It doesn't inject the script? It injects but nothing happens? It injects it too soon?

org.civicrm.bootstrapcivicrm/js/radio-checkbox.js
danilo-compucorp: I thought it was straightforward enough ;-) but I'll add it.

org.civicrm.bootstrapcivicrm/scss/bootstrap/mixins/_fa-icon.scss
danilo-compucorp: Yes, this is a code improvement. The more I used this mixin, the more I felt it was not flexible enough, so I've refactored it. And yes, this way you're not forced to follow a particular order for the parameter, and just have to provide the ones you need.

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_common.scss
danilo-compucorp: Yeah I agree. Will add this to `_forms.scss` then.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
danilo-compucorp: It doesn't get called at all.

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
danilo-compucorp: I'm asking @brunoqueiros about it.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
AkA84: Ok I see why:

> CiviCRM's pages are classified as either 'Forms' or 'Pages', and this only runs on pages classified as 'Pages'.

_(from the [_pageRun wiki page](https://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_pageRun))_

So the "export contact" page is treated as a form, and for that there is the [_buildForm hook](https://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_buildForm)

You could do something like

```php
function bootstrapcivicrm_civicrm_buildForm($formName) {
  // if you want to limit the script only to specific forms..
  if ($formName == 'CRM_Contact_Form_Search_Advanced') {  
    // inject the script
  }
}
```

danilo-compucorp: Thanks for pointing this out (I guess I'll use this for my other PR)! 
But in this case, I need to run the script always (for all pages, forms, etc).

AkA84: Ok then you can simply inject the script both in _pageRun and in _buildForm, to have some consistency with the type of hook used

AkA84:
Approved once you fix how you inject the script, and once you make sure your work doesn't conflict with @brunoqueiros 's one

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
danilo-compucorp: Ok, I'll change it then. Thanks

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
brunoqueiros: @AkA84 @danilo-compucorp It'll be some conflicts but I'll fix then latter.

brunoqueiros:
Approved

danilo-compucorp:
The changes from this PR will be merged on https://github.com/civicrm/civihr/pull/1135

PR #1110, opened 2016-08-24
PCHR-1437: New Vacancy Template page redesign

brunoqueiros:
This PR fix some issues on New Vacancy Template page, such the calendar icons and stage's buttons which weren't aligned.

***Before***
image

***After***
image


danilo-compucorp:
Approved

PR #1112, opened 2016-08-25
PCHR-1509: Show departments in contact page header

AkA84:
## Problem
The department did not appear anymore in the "enhanced" contact summary header

before

This was due to a change in how we store the department in the `civicrm_hrjobroles` table. Initially the department was stored as the `id` of the option value (each department is an option value in the _hrjc_department_ option group), but it was later changed, in [this PR](https://github.com/civicrm/civihr/pull/955) to be stored as the `value` of the option value, by leveraging the [Pseudoconstant](https://wiki.civicrm.org/confluence/display/CRMDOC/Pseudoconstant+(option+list)+Reference) mechanism that CiviCRM offers.

So basically we changed how we store the reference to the department, but forgot to also amend how we fetch it, since we were still using the `id` to do that
```php
public static function getDepartmentsList($contractID)  {
  // ...
  $query = "SELECT cov.id, cov.label
          FROM civicrm_hrjobroles hrjr
          LEFT JOIN civicrm_option_value cov
          ON hrjr.department = cov.id
          WHERE hrjr.job_contract_id = %1";
```

Additionally  the `getDepartmentsList` method would return the departments of *all* the job roles, including the ones that have expired

## Solution
after

By changing the query in `getDepartmentsList`, the department is now fetched correctly, by using its `value`. Since different option values can have the same value, the query limits the search in the _hrjc_department_ option group
```php
$query = "SELECT cov.id, cov.label
          FROM civicrm_hrjobroles hrjr
          INNER JOIN civicrm_option_value cov
            ON hrjr.department = cov.value
          INNER JOIN civicrm_option_group cog
            ON cov.option_group_id = cog.id
          WHERE hrjr.job_contract_id = %1
            AND cog.name = 'hrjc_department'";
```

Also now only the departments of the current job roles are returned

### Technical notes
I've also refactored the `_hrui_updateContactSummaryUI` function by breaking it down in smaller functions

## Problem
The "Departments" section of the contact page header didn't automatically update when a job role would get added/updated/deleted

before

## Solution
Now on create/update/edit the contact header gets updated on the fly to reflect the correct current list of departments

after

#### `HrJobRoles.getcurrentdepartments` API endpoint
In order to update the list of departments on the fly, the angular app needed a way to access the `CRM_Hrjobroles_BAO_HrJobRoles::getCurrentDepartmentsList` method, and for this reason a new api endpoint was added

#### `DOMEventTrigger` service
Initially this update-contact-header features only happened in the job contracts angular app (whenever a job contract was added/updated/deleted the header would get updated as well), and as a temporary solution all the jQuery DOM modification required for the update were directly in a service
```js
// In `ContractService`
changeHeaderInfo: function() {
  this.getCurrentContract(settings.contactId).then(function(currentContract){
    if(currentContract)  {
      cj('.crm-summary-contactname-block').removeClass('crm-summary-contactname-block-without-contract');
      // ...
    } else {
      cj('.crm-summary-contactname-block').addClass('crm-summary-contactname-block-without-contract');
      // ...
    }
  })
```
now that the same thing was required to happen also in the job roles extension, this temporary solution wasn't good anymore

So the best (although *still temporary*) solution is to have the `hrui` extension take care of the DOM manipulation (that's what it is for after all), which would be triggered by the two angular apps. Basically you need two angular apps emitting an event that a jQuery script would intercept and act upon

For this reason the `DOMEventTrigger` was written, which is a very simple service that triggers a given event (with given custom data attached) to the `document`, which can then be intercepted by any scripts that needs to listen to it. The reason it is a service it's so that in case the implementation can be improved, the angular apps will not be affected as the implementation is hidden within the service:
```js
// In HRJobRolesController
HRJobRolesService.getCurrentDepartments(jobRole.job_contract_id).then(function (departments) {
  DOMEventTrigger('updateContactHeader', {
    roles: { departments: departments }
  });
});
```
```js
// In contact.js in the hrui extension
$(document)
  .on('updateContactHeader', function (e, data) {
    if (typeof data.contract !== 'undefined')  {
      updateContactHeaderContractDetails(data.contract);
    }

    if (typeof data.roles !== 'undefined')  {
      updateContactHeaderRolesDetails(data.roles);
    }
  });
```
Is is likely that this kind of communication could happen in other areas in CiviHR, given that the our angular apps cover only a portion of the overall application, so having this kind of service (and approach) could be, although not perfect, a good step in the right direction


com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -68,18 +68,24 @@ public static function contactExists($searchValue, $searchField) {
    * @return array
    */
   public static function getDepartmentsList($contractID)  {
+    $result = array();
+
     $queryParam = array(1 => array($contractID, 'Integer'));
     $query = "SELECT cov.id, cov.label
-            FROM civicrm_hrjobroles hrjr
-            LEFT JOIN civicrm_option_value cov
```
AkA84: @omarabuhussein to be honest i'm not sure why there was a `LEFT JOIN` here

omarabuhussein: @AkA84 to get the label of the department from the option value since it was stored in civicrm_hrjobroles table by it's ID .

omarabuhussein:
approved

omarabuhussein:
@AkA84 

there are two things that just flashed to my mind ...

first the one you pointed out before that the current query don't exclude past roles departments. 

the other thing , assume someone has two or three roles that all belong to the same department (e.g development) , in this case in the UI it will look like ( development, development, development) , adding a **group by cov.label** should solve the issue but you need to test it .

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -78,6 +79,11 @@ public static function getDepartmentsList($contractID)  {
               INNER JOIN civicrm_option_group cog
                 ON cov.option_group_id = cog.id
               WHERE hrjr.job_contract_id = %1
+                AND hrjr.start_date <= '$today'
```
omarabuhussein: it is a common thing we do here  is to put variables inside strings inside curely braces , so it should be something like :

AND hrjr.start_date <= '{$today}'

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -78,6 +79,11 @@ public static function getDepartmentsList($contractID)  {
...
+                AND (
+                  hrjr.end_date = 0
```
omarabuhussein: should be hrjr.end_date IS NULL

AkA84: i tried that, it doesn't work. the field in the table has the setting NULL = no, the default value is a 0 date '0000-00-00 00:00:00'

omarabuhussein: ohh .. really .. that's bad ...

omarabuhussein: anyway don't worry , I believe job roles need data structure review ( though it is minor review since it is not complex like job contract extension)

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/HrJobRoles.php
```diff
@@ -61,14 +61,15 @@ public static function contactExists($searchValue, $searchField) {
...
 
   /**
-   * Returns an array containing the list of departments
+   * Returns an array containing the list of departments of the current job roles
    * for specific contact based on the current contract ID
    *
    * @param int $contractID contract ID
    * @return array
    */
   public static function getDepartmentsList($contractID)  {
```
omarabuhussein: maybe this should be renamed to getCurrentDepartments

omarabuhussein:
The backend part is approved ... regarding this test https://github.com/civicrm/civihr/pull/1112/commits/a11afc780b2ccabc798ce13e393998164da23f4d#diff-81d0bb720502c2baeb6e32507a8092deR23 ,

Well in getCurrentDepratments BAO method is already has tests and the API end point in this case  is just a wrapper for it and doesn't contain any extra logic so testing it is really pointless . I will remove it before merging this PR once the the JS part is reviewed and approved by @deb1990  and it pass QA testing .

hrui/js/contact.js
```diff
@@ -1,15 +1,68 @@
...
+      updateContactHeaderRolesDetails(null);
+    }
+  }
+
+  /**
+   * Updates the contact header with the given roles details
+   *
+   * @param  {object} contract
+   */
+  function updateContactHeaderRolesDetails(roles) {
```
deb1990: Function can be optimized,

``` 
if (roles && roles.departments && roles.departments.length > 0) {
   $('.crm-contact-detail-departments').html('<strong>Department:</strong> ' + roles.departments.join(', '));
} else {
   $('.crm-contact-detail-departments').html('');
} 
```

com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
```diff
@@ -417,15 +417,29 @@ define([
...
                 });
-            }
+            },
 
+            /**
+             * [getCurrentDepartments description]
+             * @param  {[type]} contractId [description]
+             * @return {[type]}            [description]
+             */
+            getCurrentDepartments: function (contractId) {
```
deb1990: Its not clear to me that why we are using 2 then statement? 
1. Here
2. Where it is called i.e. hr-job-roles-controller.js

deb1990:
Approve

PR #1113, opened 2016-08-25
PCHR-1081: Staff directory fixes

davialexandre:
The Staff Directory (accessible on index.php?q=civicrm/profile/table&reset=1&gid=19&force=1) were presenting a number of problems:
- Trying to access the page resulted in a DB error
- Clicking on name does not take you back to Contact page
- The Title was coming from a past contract
- Department was not coming through
- The manager was not coming though

This required a number of approaches to fix the problems:
- The DB error was caused by a misconstructed query. A method on CRM_HRProfile_Page_HRProfile was trying to add the Final_Termination_Date and Initial_Join_Date fields to the query, but those have been removed a long time ago, on PCHR-774. To fix the error, I removed the piece of code that was doing this.
- To fix the link on the contact name, I updated the hrstaffdir_civicrm_searchColumns hook which was the one who was setting the link pointing to the profile page instead of the contact one
- To show the correct contract Title, I updated the JOIN on CRM_Hrjobcontract_BAO_Query::from to properly join with the latest contract revision
- To be able to show the department, I needed to:
  - Fix the JOIN clause from the job role to the contract, to get the roles
    for the active contract
  - Add the job roles to the list of fields of the select statement. This
    was achieved by implementing the select() method on the CRM_Hrjobroles_BAO_Query
    class
  - Update the Profile configuration to use the job roles field instead of the
    old/deprecated jobcontract_role
- To show the manager, I created a new Query Object to the hrprofile extension that injects all the necessary joins and fields to fetch the managers for a contract. Since we have multiple managers from a single contact and we need to return this information in a single field, I used the GROUP_CONCAT function to group all the managers into a single field, separated by commas.

### DIFF
--- a/hrprofile/CRM/HRProfile/BAO/Query.php
+++ b/hrprofile/CRM/HRProfile/BAO/Query.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * Class CRM_HRProfile_BAO_Query
+ *
+ * This class is a little bit different from the regular Query classes, as it's
+ * not related to a BAO.
+ *
+ * The query that loads the data for the Staff Directory is the same one that is
+ * used for the AdvancedSearch (with the exception that is uses the Directory
+ * Profile to get the fields to be displayed). So this class injects the JOINS
+ * and fields required to load the information for the "Managers" field of this
+ * profile.
+ */
+class CRM_HRProfile_BAO_Query extends CRM_Contact_BAO_Query_Interface {
+
+  private $fields = [];
+
+  /**
+   * The managers is not a real BAO/DAO field, so we manually add it to the
+   * fields list.
+   *
+   * @return array
+   */
+  public function &getFields() {
+    if(empty($this->fields)) {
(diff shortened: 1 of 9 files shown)


hrprofile/CRM/HRProfile/BAO/Query.php
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @return string
+   */
+  public function from($name, $mode, $side) {
+    $from = '';
+    switch ($name) {
+      case 'civicrm_contact':
+        $from .= "$side JOIN civicrm_relationship cr
+                    ON cr.contact_id_a = contact_a.id AND
+                       (cr.end_date IS NULL OR cr.end_date >= CURDATE()) AND
```
omarabuhussein: pass CURDATE() via php so the query can be cached by mysql.

omarabuhussein:
i left one note , the rest looks good.

davialexandre:
Closing since the ticket as also closed

PR #1114, opened 2016-08-26
PCHR-1516: modifications to menu items

omarabuhussein:
## Requirements

A set of changes to the navigation menu was required which is :

Remove the following menu items : 
- Search >> Full-text search
- Search >> Search builder
- Search >> Find assignments (Remove when T&A is enabled)
- Search >> Find activities (Remove when T&A is enabled)
- Search >> Custom searches
- Contacts >> Contact reports
- Contacts >> New activity (Remove when T&A is enabled)
- Contacts >> Import activities (Remove when T&A is enabled)
- Contacts >> New tag 
- Contacts >> Manage tags
- Appraisals (and its child menus)
- Reports (and its child menus)
- Administer >> People management >> Redaction rules
- Administer >> CiviReport (and its child menus)
- Job contracts (and move its child menus under "Contacts" below "Import contacts")

Also another menu items need to be created for "Import custom fields" and should be placed under "Contacts" blow "Import contacts".

## Implementation

Inside hrui I removed the core menu items that are required to be removed such as "new tag" , I leveraged  hrui_civicrm_navigation hook which receive menu items by reference in a variable called $params , so I unset all of them their .

 For Appraisals menu items and its child menus , I did nothing since appraisals menu item is created by appraisals extension which is disabled by default when installing a new buildkit instance.

For reports I did nothing too since reports component is already disabled when installing a new buildkit instance .

For "Job contract" menu item , I removed it and moved its sub-menus by adding a new upgrader method ( upgrader  methods are set to run automatically when installing job contract extension ) that remove and then re-create the menu items .

For "Import custom fields" it was already created but with different name "Import Job" via job contract extension and it already had a set of sub-menus work as a shortcut to reach specific custom fields importer pages , 
So I  removed it from job contract and added a code to code to manage it in hrui extension since it does not belong to hrjobcontract.

I also did some general refactoring and removing for unused and commented code .

anim

### DIFF
--- a/bin/drush-install.sh
+++ b/bin/drush-install.sh
@@ -28,9 +28,7 @@ APP_EXTS=\
 org.civicrm.hrreport,\
 org.civicrm.hrui,\
 org.civicrm.hrcase,\
-org.civicrm.hrstaffdir,\
 org.civicrm.hrim,\
-org.civicrm.hrprofile,\
 org.civicrm.hrrecruitment,\
 org.civicrm.reqangular,\
 org.civicrm.contactsummary,\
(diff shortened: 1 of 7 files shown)


coldrunKacper:
Approved.

PR #1115, opened 2016-08-26
PCHR-1526: New applicant redesign

brunoqueiros:
This PR overrides new applicant page styles to look as close to CiviHR design.

***Before***
image

***After***
image


danilo-compucorp:
Approved

PR #1116, opened 2016-08-26
PCHR-1437: Fix New vacancy template issues

brunoqueiros:
This PR fix some design issues on New Vacancy page.

## Problems
- Adjust alignment and spacing from Stages buttons
- Adjust spacing and remove the screen icon from Preview buttons
- Adjust permissions select width to keep the trash icon inline

***Before***
image

***After***
image


org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -2,14 +2,39 @@
...
+        td:nth-child(2) {
+          vertical-align: bottom;
+          padding-bottom: 4px;
+        }
       }
     }
 
     .crm-block.crm-form-block {
       .form-layout-compressed .crm-profile-selector-preview {
         @extend .btn-primary;
```
danilo-compucorp: could we reuse any of the overrides already done for the buttons?

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
```diff
@@ -45,6 +70,37 @@
         }
       }
     }
+
+    @media (max-width: 1340px) {
```
danilo-compucorp: Are you sure those media-queries will work?
I know they should, but since the markup lacks the proper viewport setup, I've ended up using `max-device-width` instead.
Please, try those in chrome to be sure.

brunoqueiros: Yes they work, I tried with `max-device-width` and didn't work. The tests were made on Chrome and Firefox.

org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss
brunoqueiros: I think isn't the case here, because it's a specific align/spacing change.

danilo-compucorp: If this only happens in this page, ok.

danilo-compucorp:
Approved

PR #1117, opened 2016-08-26
PCHR-1517: Fix non-compatible declaration of overridden methods in hrreport

omarabuhussein:
## Problem 

The following warning was appearing :

`Strict warning: Declaration of CRM_HRReport_Form_Contact_HRSummary::customDataFrom() should be compatible with CRM_Report_Form::customDataFrom($joinsForFiltersOnly = false) in require_once() (line 37 of civicrm/tools/extensions/civihr/hrreport/CRM/HRReport/Form/Contact/HRSummary.php).`

## Solution

In civicrm 4.7 the signature for **CRM_Report_Form::customDataFrom** method is slightly changed so I've updated the extensions in civihr that override this method to match its signature . I also explicitly delcared all methods in (HRReport/Form/Contact/HRDetail.php) & ( HRReport/Form/Contact/HRSummary.php ) as public.


nishant-bhorodia:
Approved!

PR #1119, opened 2016-08-29
PCHR-1229: Fix hrrecruitment extenstion momentjs script location

omarabuhussein:
## Problem

- Open Vacancies/Dashboard
- Then try to click on Absences, Mailing, Search, etc.
- The drop down menus don't appear, and nothing happens when clicking on these menus.

before

The reason behinde this is that  momentjs is used in hrrecruitment extension but it  points to wrong location ( momentjs was shipped with civicrm previously but not anymore in 4.7 and that's where it was pointing ).

## Solution

since momentjs library is available in org.civicrm.reqangular extension i just changed hrrecruitment to use it.

after


nishant-bhorodia:
Approved!

PR #1122, opened 2016-08-31
PCHR-1435: Find activies redesign

brunoqueiros:
This PR overrides Find activies page styles to look as close to CiviHR design.

***Before***
image

image

***After***
image

image


danilo-compucorp:
Approved

PR #1126, opened 2016-09-01
PCHR-1516: Remove (Job Contract Report) menu item

omarabuhussein:
I created a new upgrader method to remove ( Job contract Report ) menu item if it exist , and also updated the existing upgrader ( upgrade_1017) so it doesn't get created at all for new installations .


davialexandre:
Approved

PR #1127, opened 2016-09-01
PCHR-1485: Fix displaying roles with no end date

davialexandre:
This PR fixes two problems related to the job role end date.

**Problem 1**
Job roles with no end date were displayed as a past role. This was caused by an error on the condition which checked if the end_date was empty.  It was checking for an empty end_date using !end_date, but, since the end_date was converted to a moment object, this condition would always evaluate to false and the role would be considered to be in the past

**Solution 1**
This has been fixed by only converting the end_date to an moment object if it's not empty.

----------------------------------------------------------------------------------------

**Problem 2**
When updating a job role with an empty end_date, its value was automatically updated to the current date. This was happening because the type of the end_date field on the database was set to be a TIMESTAMP, which sets the field value as the current date when it's NULL. 

**Solution 2**
The field type was changed to DATETIME, which allows NULL values. The same change was applied to the start_date field, for consistency.


brunoqueiros:
Front-end part it's ok!

omarabuhussein:
Approved

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader.php
```diff
@@ -6,4 +6,8 @@
 class CRM_HRLeaveAndAbsences_Upgrader extends CRM_HRLeaveAndAbsences_Upgrader_Base {
   // By convention, functions that look like "function upgrade_NNNN()" are
   // upgrade tasks. They are executed in order (like Drupal's hook_update_N).
+
+  public function upgrade_5000() {
```
coldrunKacper: This function is creepy.

com.civicrm.hrjobroles/js/test/mocks/job-roles.js
```diff
@@ -69,6 +66,7 @@ define(function () {
                 "title": "Test",
                 "job_contract_id": "1",
                 "start_date": "2015-12-30 00:00:00",
+                "end_date": '0000-00-00 00:00:00',
```
omarabuhussein: shouldn't you update the tests here ?

com.civicrm.hrjobroles/js/test/mocks/job-roles.js
```diff
@@ -92,6 +90,7 @@ define(function () {
               "title": "Test filter",
               "job_contract_id": "1",
               "start_date": "2015-12-30 00:00:00",
+              "end_date": '0000-00-00 00:00:00',
```
omarabuhussein: here too

com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
```diff
@@ -508,25 +508,27 @@ static function &fields()
         ) ,
         'hrjc_role_start_date' => array(
           'name' => 'start_date',
-          'type' => CRM_Utils_Type::T_DATE | CRM_Utils_Type::T_TIME,
+          'type' => CRM_Utils_Type::T_DATE + CRM_Utils_Type::T_TIME,
+          'default' => 'NULL',
```
omarabuhussein: as I said on hipchat , the start date should not be null ..

com.civicrm.hrjobroles/js/test/mocks/job-roles.js
davialexandre: They were updated. This is the old version

com.civicrm.hrjobroles/js/test/mocks/job-roles.js
```diff
@@ -66,7 +66,7 @@ define(function () {
                 "title": "Test",
                 "job_contract_id": "1",
                 "start_date": "2015-12-30 00:00:00",
-                "end_date": '0000-00-00 00:00:00',
+                "end_date": null,
```
omarabuhussein: oh I see , you updated the tests in the commit

com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
davialexandre: It had the default 0 before, that's why I changed it do default NULL, to avoid breaking something else. I agree with you that it should be not null, I'm just not sure it should be fixed on this ticket/PR

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -36,9 +36,11 @@ public function upgrade_1002(){
     return TRUE;
   }
```
omarabuhussein: can you remove upgrade_1001 and upgrade_1002 , there are not needed

omarabuhussein: also can you modefiy upgrade_1004 or add another upgrader ( 1005 ) to update current job roles table (end_date) field and in case it is = 0000-00-00 or 0 then set it to null .

`UPDATE civicrm_jobroles SET end_date = NULL where (end_date = '0000-00-00 00:00:00' OR end_date = '0' ) `

davialexandre: You're 100% sure about that? I mean, I'm ok about removing 1002, since the same thing is ran during the installation, but what about 1001? How can we know that all civihr installs have already been upgraded?

davialexandre: Any specific reason to split 1004 into two methods? I like having all that into a single method because it's easy to spot that we're updating the end_date because the type has changed

com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
omarabuhussein: well, since the default was 0 then it fine for now to change it to null , it could be solved  later in another ticket ..

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
omarabuhussein: regarding 1001, it's been there since  Nov 3, 2015 and I don't think there is any existing client or install not get and update since then , but let's keep it to be in the safe side , but please remove 1002 then .

omarabuhussein: > Any specific reason to split 1004 into two methods? I like having all that into a single method because it's easy to spot that we're updating the end_date because the type has changed

no , it doesn't really matter .. whatever you like :smile:

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -36,9 +36,11 @@ public function upgrade_1002(){
...
 
-  public function upgrade_1003() {
-    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrjobroles` MODIFY COLUMN `start_date` DATETIME DEFAULT 0');
-    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrjobroles` MODIFY COLUMN `end_date` DATETIME DEFAULT 0');
+  public function upgrade_1004() {
+    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrjobroles` MODIFY COLUMN `start_date` DATETIME DEFAULT NULL');
+    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrjobroles` MODIFY COLUMN `end_date` DATETIME DEFAULT NULL');
+    //Update any end_date using "0000-00-00 00:00:00" as an empty value to NULL
+    CRM_Core_DAO::executeQuery("UPDATE `civicrm_hrjobroles` SET `end_date` = NULL WHERE end_date = '0000-00-00 00:00:00'");
```
omarabuhussein: tbh I just realized you already have an sql query to update existing records

omarabuhussein:
okay all good , just please remove upgrade_1002 .. otherwise approved

PR #1130, opened 2016-09-02
PCHR-1322: Adjusting the ui-select for the tasks and assigments

danilo-compucorp:
This PR removes the 'set status' option from document status dropdown and uses the "_ui-select_" dropdown, with the "_allow-clear_" option.

Before:
before

After:
pchr-1322


brunoqueiros:
Approved

PR #1131, opened 2016-09-02
PCHR-1435: Find activities design issues

brunoqueiros:
This PR fix some design issues on find activities page.

## Problems
- Edit Modal
- View Modal
- Pager

***Before***
image

image

image

***After***
image

image

image

image


JonScreat:
**Results page:**
Could this section be in white with the same shadow as other elements. Also can the rounded corners be removed?
5ccc7007-41ad-42e8-8d1a-1584f90de66a

**Popups:**
Can the accordion section have 20px padding around it? I realise this is contrary to what I requested earlier, but I didn’t notice at the time they would bleed into the background. If you could fix that it would be great.
4bf15cb1-0594-4f6a-ad7f-524837218962

danilo-compucorp:
Approved

PR #1136, opened 2016-09-06
PCHR-1236: Blank start and end dates  input boxes when editing vacancies 

omarabuhussein:
## Problem

When Trying to edit a vacancy the start and end dates input boxes are shown empty 

before

The issue is inside quick form **setDefaultValues** method , the start and end dates are converted to a format not recognized by the used JavaScript date picker plugin.

another problem is after submitting the form , the following notices will appear in vacancy listing page :

selection_121

which is because inside quick form **formRule** method , **start_date_display** and **end_date_display** where used instead of **start_date** and **end_date**.

## Solution

Actually the date picker plugin appear to accept MySQL DateTime format so I just removed the date converting inside **setDefaultValues** method .

after

for the notices , i corrected the field names inside **formRule** method .

### DIFF
--- a/hrrecruitment/CRM/HRRecruitment/Form/HRVacancy.php
+++ b/hrrecruitment/CRM/HRRecruitment/Form/HRVacancy.php
@@ -97,13 +97,6 @@ class CRM_HRRecruitment_Form_HRVacancy extends CRM_Core_Form {
 
     if (!empty($params['id'])) {
       CRM_HRRecruitment_BAO_HRVacancy::retrieve($params, $defaults);
-      //format vacancy start/end date
-      if (!empty($defaults['start_date'])) {
-        list($defaults['start_date'], $defaults['start_date_time']) = CRM_Utils_Date::setDateDefaults($defaults['start_date'], 'activityDateTime');
-      }
-      if (!empty($defaults['end_date'])) {
-        list($defaults['end_date'], $defaults['end_date_time']) = CRM_Utils_Date::setDateDefaults($defaults['end_date'], 'activityDateTime');
-      }
 
       //show that only number of permission row(s) which have defaults if any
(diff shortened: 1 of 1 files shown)


hrrecruitment/CRM/HRRecruitment/Form/HRVacancy.php
```diff
@@ -140,8 +133,8 @@ function buildQuickForm() {
     $this->add('wysiwyg', 'benefits', ts('Benefits'), array('rows' => 2, 'cols' => 40));
     $this->add('wysiwyg', 'requirements', ts('Requirements'), array('rows' => 2, 'cols' => 40));
 
-    $this->add('datepicker', 'start_date', ts('Start Date'), FALSE, array('formatType' => 'activityDateTime'));
-    $this->add('datepicker', 'end_date', ts('End Date'), FALSE, array('formatType' => 'activityDateTime'));
+    $this->add('datepicker', 'start_date', ts('Start Date'), NULL, TRUE, array('minDate' => time()));
```
coldrunKacper: It would be good to add such validation (for future vacancy dates) in PHP too.

coldrunKacper:
I'm not sure about $fields['start_date_display'] and $fields['end_date_display'] date format in formRule() but if they are in JS format (m/d/Y) then the IF which compares the start and end date may not be correct currently.

omarabuhussein:
> I'm not sure about $fields['start_date_display'] and $fields['end_date_display'] date format in formRule() but if they are in JS format (m/d/Y) then the IF which compares the start and end date may not be correct currently.

The JS date picker plugin convert the date and time to Y-m-d H:i:s format , so it works fine :) ..

coldrunKacper:
Approved.

PR #1141, opened 2016-09-07
PCHR-1535: Fix malformed return URL for the Absence Record form

davialexandre:
After making changes to an Absence Request, the user is redirect to an URL. This URL has a fragment, which is used by the backbone app to select which tab should be displayed. This fragment was added to the end of the URL by passing it to the CRM_Utils_System::url() method as part of the $query paramenter. This commit on the civicrm-core https://github.com/civicrm/civicrm-core/commit/c80e2dbf changed the behavior of how the method handles the $query value if it has a fragment (a string starting with #), turning part of the param value as the path for the URL, which led to a malformed URL.

The CRM_Utils_System::url() have a specific param for URL fragments, so, in order to fix the malformed URLs,  I updated every call to this method to pass the fragment using this param.


nishant-bhorodia:
Approved!

PR #1143, opened 2016-09-07
PCHR-1296: Hide default messages in tooltips on contact details in se…

danilo-compucorp:
This PR hides the default messages in tooltips on contact details in search.

Before:
before

After:
after


brunoqueiros:
Approved

PR #1149, opened 2016-09-09
PCHR-1490: Integrate hrnavigation with hrui extension

omarabuhussein:
Yes another extension we get rid off here ( let's have a more cleaner git repo :smile: ) , Basically **uk.co.compucorp.civicrm.hrnavigation** get removed which is only override CRM/Core/BAO/Navigation.php in it and do a minor modification to replace civicrm **hide menu** menu item to **Self Service Portal** menu item since the hide menu is hard-coded and not accessible via core_civicrm_navigation hook .

**Without hrnavigation :** 
selection_142

**With hrnavigation :** 
selection_141

This PR only move the ovveridden file to hrui extension instead of having a separate extension to do such thing .


coldrunKacper:
Approved.

PR #1151, opened 2016-09-09
PCHR-1558: Find and merge duplicate contacts redesign

brunoqueiros:
This PR overrides Find and merge duplicate contacts page and sub-pages style to look as close to CiviHR design.

It was created a placeholder with the table style to avoid adding the new selector on `table.scss` and when you need to style a specific table you just need to extend the new placeholder `@extend civihr-table;`

***Before***
image

image

image

***After***
image

image

image


danilo-compucorp:
Approved

PR #1152, opened 2016-09-09
PCHR-1329: Contact Summary wrong age calculation

omarabuhussein:
## Problem 

Contact summary extension does not calculate the age correctly , For example if date of birth is (1957-10-02), the age should be 58 but it show as 59

image

the reason of that is how momentJS **fromNow()** method work   , it round the result up instead of rounding it down .

## Solution 

I wrote a new method **calculateAge** in **contactDetails** service to calculate the age properly .


contactsummary/js/src/contact-summary/services/contactDetails.js
```diff
@@ -86,6 +86,28 @@ define([
             return deferred.promise;
         }
 
+      /**
```
AkA84: The indentation is wrong, and you need to specify the type of the param, and the types are always written as {types} in js docs
```js
/**
 * Calculate age from birth date
 *
 * @param {object} dateOfBirth Date of birth in a YYYY-MM-DD format
 * @return {string}
 */
function

File: contactsummary/js/src/contact-summary/services/contactDetails.js [hunk 2]
```diff
@@ -86,6 +86,28 @@ define([
...
+       * Calculate age from birth date
+       *
+       * @param dateOfBirth Date of birth as momentJS acceptable format
+       * @returns string
+         */
+        function calculateAge(dateOfBirth)
+        {
```
AkA84: Brackets always inline
```js
function calculateAge(dateOfBirth) {

}

File: contactsummary/js/src/contact-summary/services/contactDetails.js [hunk 3]
```diff
@@ -86,6 +86,28 @@ define([
...
+          dateOfBirth = moment(dateOfBirth, 'YYYY-MM-DD').format('YYYYMMDD');
```
AkA84: This is extremely over engineered , you can achieve the same thing by simply doing

```js
function calculateAge(dateOfBirth) {
  return moment().diff(moment(dateOfBirth, 'YYYY-MM-DD'), 'years');
}

AkA84:
The implementation could be simplified by a lot, please see my comments

AkA84:
Approved

PR #1154, opened 2016-09-12
PCHR-1562 update styling of advanced search results 

danilo-compucorp:
This PR improved the style of the advanced search results.

- Fix background, paddings, font styles, buttons table styles to match civihr styles
- Makes checkboxes in the first column visible.
- Fix preview tooltip (on hover on icon)
- On selection contact row should be highlighted. 
- Replace icons

### Page

**before**
before

**after**
after

### "more" 

**before**
more-before

**after**
more-after

### Tooltip

**before**
tooltip-before

**after**
tooltip-after

### Row highlight

row-highlight


brunoqueiros:
Approved

JonScreat:
The only comments I have are the styling for the contact tool tip and the 'more' dropdown. A styling reference for these can be seen here: https://invis.io/DB6WA0JGR#/150009229_2_Search_Results

JonScreat:
image

I've added a couple of comments to this attachment - apart from that it's looking good to me

jamienovick:
Small point - the before image (https://cloud.githubusercontent.com/assets/18520391/18450299/bd28403c-7932-11e6-8226-f6e2d63fa124.png) of the results looks nicer without the table border around each row. Thats a design pattern that we are using so would be good to use that. Jon - please note that behaviour for tables.

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_tables.scss
```diff
@@ -24,36 +25,30 @@
...
-      &:first-of-type {
-        border-top-left-radius: $button-border-radius !important;
-        border-bottom-left-radius: $button-border-radius !important;
-        border-top-right-radius: 0 !important;
-        border-bottom-right-radius: 0 !important;
-      }
+    border: 0;
+    background-image: none !important;
+
+    &, * {
```
brunoqueiros: One selector per line, the same on line 43 and 49

brunoqueiros:
Approved

brunoqueiros:
Approved

PR #1156, opened 2016-09-13
PCHR-1287 - Modifying Contact Summary block to read Length of Service from API entry point instead of latest Job Contract dates. Creating BAO and API changes required by front-end part.

coldrunKacper:
A "Length of Service" value on "Contact Summary" tab was displaying incorrect value as it was calculated on the front-end side depending on latest Job Contract dates.
Things which got changed with this PR:
- Changed Length of Service front-end calculation method so currently it's retrieved by API call and matches "Personal Details" tab displaying overall length of service.
- Modified length of service calculation to include both start and end date of service period. Previously for Job Contract starting on 2016-09-01 and ending 2016-09-02 it was returning 1 day which was not correct (should be 2 days). It's fixed now.
- Added new Upgrader function to recalculate Length of Service fields.
- Added unit tests.

## Javascript:
**Problem**: 
It was required to send messages between different angular apps(ContactSummary, JobContracts etc) to notify the other about the changes in one app. So that the other can update itself.

**Solution**
A new service "pubSubService" is added, **using which communication between different angular apps can be achieved**. 

Here is how to use it-
**Dependency Injection**
1. Include the file in the define section
2. Inject the service 
```javascript
define([
  // ...
  'common/services/pub-sub-service'  // inject the service
], function () {
  'use strict';
  function SomeController(/* ... */, pubSubService) {
  }
});
```

**Subscribing to Events**
```javascript
pubSubService.subscribe("event_a",  function(){
//Callback
});
```

**Publish Events**
```javascript
pubSubService.publish("event_a");
```

Using the same service, Contact Summary is getting updated whenever there is a change in Job Contract Tab.
**Inject**
```
define([
  'common/moment',
  'contact-summary/modules/controllers',
  'contact-summary/services/contactDetails',
  'contact-summary/services/contract',
  'common/services/pub-sub',
], function (moment, controllers) {
  'use strict';
  function KeyDetailsCtrl($log, ContactDetails, Contract, pubSub) {

```

**Subscribing**
`pubSub.subscribe('contract-refresh',  resetKeyDates);`

**Publishing**
`pubSub.publish('contract-refresh');`

**Caching has been disabled for API calls in contactsummary because of which contracts were not getting updated**


contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -180,6 +180,40 @@ define([
                 });
         };
 
+        /**
+         * @ngdoc method
+         * @name getLengthOfServiceYmd
+         * @methodOf ContractService
+         * @returns {*}
+         */
+        factory.getLengthOfServiceYmd = function () {
```
brunoqueiros: Create a test for this method

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -31,8 +31,13 @@ define([
                 }
 
                 this.primaryContract = response;
-                this.primaryContract.lengthOfService = getLengthOfService(response.start_date, response.end_date);
-            }.bind(this))
+            }.bind(this)).then(function (response) {
```
AkA84: put this on a new line, like the other `then`

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -31,8 +31,13 @@ define([
...
+                return Contract.getLengthOfServiceYmd();
```
AkA84: we're following a 2 space indentation convention now. Use it for all of your new code and when possible (and when reasonable) amend the old code

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -180,6 +180,40 @@ define([
                 });
         };
 
+        /**
+         * @ngdoc method
```
AkA84: Please describe the method

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -180,6 +180,40 @@ define([
...
+         * @name getLengthOfServiceYmd
+         * @methodOf ContractService
+         * @returns {*}
+         */
+        factory.getLengthOfServiceYmd = function () {
+            var deferred = $q.defer();
+            var val = null;
```
AkA84: What's the point of this variable? Couldn't you simply do

```js
.then(function (response) {
  if (!response.is_error) {
    deferred.resolve(response.values);
  }
}
```

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -180,6 +180,40 @@ define([
...
+                    {
+                        sequential: 0,
+                        contact_id: response.id
+                    },
+                    'getlengthofserviceymd'
+                )
+            })
+            .then(function(response) {
+                if (response.is_error) {
+                    return;
```
AkA84: Shouldn't you `deferred.reject` at this point? How do other angular serviced in civihr handle it?

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -180,6 +180,40 @@ define([
                 });
         };
 
+        /**
+         * @ngdoc method
+         * @name getLengthOfServiceYmd
```
omarabuhussein: naming the method as getLengthOfServiceYmd is a bit akward ( to me ) .. I think getLengthOfService is better .

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -193,21 +194,26 @@ protected static function getContractDates($contracts) {
     // If there are two (or more) Contracts starting on the same day
     // then we pick only the one with latest End date.
     foreach ($contracts['values'] as $contract) {
-      $details = civicrm_api3('HRJobDetails', 'getsingle', array(
+      $details = civicrm_api3('HRJobDetails', 'get', array(
```
omarabuhussein: since we are inside job contract extension it is better to use BAO method to get the contract details ( it should be much faster too )

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -193,21 +194,26 @@ protected static function getContractDates($contracts) {
...
+      }
+      $detailsValues = CRM_Utils_Array::first($details['values']);
+      if (empty($detailsValues['period_end_date'])) {
+        $dates[$detailsValues['period_start_date']] = null;
+        continue;
       }
-      if (!empty($dates[$details['period_start_date']])) {
-        if ($details['period_end_date'] > $dates[$details['period_end_date']]) {
-          $dates[$details['period_start_date']] = $details['period_end_date'];
+      if (!empty($dates[$detailsValues['period_start_date']])) {
```
omarabuhussein: this can be removed since currently we cannot have two contracts with the same "start_date" :
```php
      if (!empty($dates[$detailsValues['period_start_date']])) {
        if ($detailsValues['period_end_date'] > $dates[$detailsValues['period_start_date']]) {
          $dates[$detailsValues['period_start_date']] = $detailsValues['period_end_date'];
        }
        continue;
      }
```

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -193,21 +194,26 @@ protected static function getContractDates($contracts) {
     // If there are two (or more) Contracts starting on the same day
```
omarabuhussein: also this comment should be removed too

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -193,21 +194,26 @@ protected static function getContractDates($contracts) {
...
     // then we pick only the one with latest End date.
     foreach ($contracts['values'] as $contract) {
-      $details = civicrm_api3('HRJobDetails', 'getsingle', array(
+      $details = civicrm_api3('HRJobDetails', 'get', array(
         'sequential' => 1,
```
omarabuhussein: also I was thinking maybe you can remove this API call from **calculateLengthOfService()** :

```php
    // Getting all Job Contracts for given Contact ID.
    $contracts = civicrm_api3('HRJobContract', 'get', array(
      'sequential' => 1,
      'contact_id' => (int)$contactId,
      'deleted' => 0,
      'options' => array('limit' => 0),
    ));
```

and here inside **getContractDates()** you can call **getContactContracts()** and then loop through the results here  ... it may be also a good chance to improve **getContactContracts()** and make it depend on BAO methods instead of API since.

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
AkA84: Problem is if you don't have a BAO method for something as basic as a `get`, it means that you need to manually implement it (and write the sql), just to get rid of the performance issues of the api...is that correct?

hrjobcontract/api/v3/HRJobContract.php
```diff
@@ -110,6 +119,35 @@ function civicrm_api3_h_r_job_contract_getlengthofservice($params) {
...
+ * 
+ * Return an object containing years, months and days of Length of Service
+ * for specific 'contact_id'.
+ * Useful for front-end part of Contact Summary block.
+ * 
+ * @param array $params
+ * @return array API result descriptor
+ * @throws API_Exception
+ */
+function civicrm_api3_h_r_job_contract_getlengthofserviceymd($params) {
```
omarabuhussein: what is the problem of using **civicrm_api3_h_r_job_contract_getlengthofservice** and do the converting inside the angular part as it was before ?

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
omarabuhussein: no need to write anything .. any BAO class extends a DAO class which also extends CRM_Core_DAO which contain some handy methods such as **commonRetrive** which can be used to get the data , something like this :

https://github.com/civicrm/civihr/blob/1c5b932a7aef97690fd6cc95c7d4f974e3a3ad3d/com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php#L123

coldrunKacper: Agree, but as confirmed with Omar, currently we stick to API method as it provides current Details revision of requested Job Contract. We want to change it later but it would be more global change probably for every API method /  entity which use any additional logic.

hrjobcontract/api/v3/HRJobContract.php
coldrunKacper: I though it may be a good thing to have one entry point for getting these days/months/years in case we want to use it somewhere else (maybe in SSP My Details or so) and not recalculating it in each place using different angular/JS/template methods.

omarabuhussein: aha I see , cool then.

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
coldrunKacper: After spotting some weird behavior of CRM_Core_DAO::commonRetrieve() method (which is almost perfect to getting a set of entity items) we decided to keep this part as it is :D

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
```diff
@@ -157,6 +157,7 @@ public static function getContactContracts($cuid)  {
     $contracts = civicrm_api3('HRJobContract', 'get', array(
       'sequential' => 1,
       'contact_id' => $cuid,
+      'options' => array('limit' => 0),
```
omarabuhussein: you forget to add 'deleted' => 0

omarabuhussein:
Looks good now , Approved

deb1990:
A new service "pubSubService" is added, using which communication between different angular apps can be achieved. Using the same Contact Summary is getting updated whenever there is a change in Job Contract Tab.

contactsummary/js/src/contact-summary/controllers/contactSummary.js
```diff
@@ -1,2 +1,38 @@
...
        //     this.contact = response;
        //     this.ready = true;
        // }.bind(this));
-    }

    controllers.controller('ContactSummaryCtrl', ['$log', 'ContactService', 'settings', ContactSummaryCtrl]);
});
\ No newline at end of file
+define([
+    'contact-summary/modules/controllers',
```
AkA84: Why is this controller shown as an (almost) entire diff? This doesn't look right

contactsummary/js/src/contact-summary/controllers/keyDates.js
```diff
@@ -1,84 +1,97 @@
...
-     * @constructor
-     */
-    function KeyDatesCtrl($log, Contract, JobRole) {
-        $log.debug('Controller: KeyDatesCtrl');
+  /**
+   * @ngdoc controller
+   * @name KeyDatesCtrl
+   * @param $log
+   * @param {ContractService} Contract
+   * @param {JobRoleService} JobRole
```
AkA84: Didn't update the documentation

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -14,37 +15,52 @@ define([
...
-      .then(function (response) {
-        this.contactDetails = response;
-        return Contract.getPrimary();
-      }.bind(this))
-      .then(function (response) {
-        if (_.isEmpty(response)) {
-          return;
+
+
+    var getContacts = function(){
```
AkA84: At this point this, being a "private" function, you could move it outside `KeyDetailsCtrl` completely

AkA84: Also, and I understand this is a matter of standards, why
```js
var getContacts = function () {
```
instead of
```js
function getContacts () {
```
If you don't have any reason for preferring the first over the second, then please use the second so you can exploit hoisting

AkA84: Can you document the function?

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -14,37 +15,52 @@ define([
...
-      .finally(function () {
-        this.ready = true;
-      }.bind(this));
+          .bind(this))
+        .finally(function () {
+          this.ready = true;
+        }.bind(this));
+    }.bind(this);
+
+    var resetKeyDetails = function() {
```
AkA84: All that applies for the previous function

contactsummary/js/src/contact-summary/controllers/keyDetails.js
```diff
@@ -14,37 +15,52 @@ define([
...
+      Contract.resetContracts();
+      ContactDetails.data.item = {};
+      getContacts();
+    };
+
+    getContacts();
+
+    pubSubService.subscribe("contract-refresh",  resetKeyDetails);
```
AkA84: I'd say let's use single quotes by default

contactsummary/js/src/contact-summary/services/api.js
```diff
@@ -45,7 +45,6 @@ define([
          */
         function sendRequest(method, data, config) {
             config = angular.extend({
-                cache: true,
```
AkA84: Can you document this change in the PR description? Seems important to me in terms of reasons and consequences

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -89,13 +89,31 @@ define([
 
     /**
      * @ngdoc method
-     * @name getContracts
```
AkA84: In the documentation you need to write also what the method does

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -89,13 +89,31 @@ define([
...
+     * @name resetContracts
      * @methodOf ContractService
      * @returns {*}
      */
+    factory.resetContracts = function () {
+      contracts=[];
```
AkA84:
```js
contracts = [];
```

contactsummary/js/src/contact-summary/services/contract.js
```diff
@@ -89,13 +89,31 @@ define([
...
+      factory.collection = {
```
AkA84: Where does the structure of this object comes from (`insertItem`, `getItem`, etc)

hrjobcontract/js/src/job-contract/controllers/contract-list.js
```diff
@@ -196,6 +197,7 @@ define([
                       if (!result.is_error) {
                         ContractService.updateHeaderInfo();
                         removeContractById($scope.contractCurrent, contractId) || removeContractById($scope.contractPast, contractId);
+                        pubSubService.publish("contract-refresh");
```
AkA84: Single quotes. Also, in order to highlight this, it would make this more readable if you put a newline before it
```js
ContractService.updateHeaderInfo();
removeContractById($scope.contractCurrent, contractId) || removeContractById($scope.contractPast, contractId);

pubSubService.publish('contract-refresh');
```

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.js
```diff
@@ -213,6 +212,7 @@ define([
                             }).then(function(){
                                 $scope.$broadcast('hrjc-loader-hide');
                                 $modalInstance.close(contract);
+                                pubSubService.publish("contract-refresh");
```
AkA84: Add a new line for highlight and readability purpose

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
+define([
+  'common/modules/services'
+], function (module) {
```
AkA84: Let's keep this consistent with how we name things in other places
```js
function (services)
```

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+  'use strict';
+
+  var topics = Object.create(null);
+  module.factory("pubSubService", ['$rootScope', function ($rootScope) {
```
AkA84: Add a new line before this, let's make sure the code is as readable as we can

AkA84: What about documenting this service? We don't do it yet, it would be good to start

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+    return {
+      subscribe: function (topic, listener) {
```
AkA84: Document the service's methods (same for below)

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        if (!topics[topic]) {
```
AkA84: Was thinking of a one-line. Not sure about it, but what do you think of
```js
topics[topic] = topics[topic] || [];
```

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+          topics[topic] = [];
+        }
+        listener.$rootScope = $rootScope;
```
AkA84: What are you doing here?

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        var index = topics[topic].push(listener) - 1;
```
AkA84: Doesn't look nice to have a `var` declaration in the middle of the code implementation. Split things up
```js
function (topic, listener) {
  var index;
  // ....

  index = topics[topic].push(listener) - 1;
```

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        return {
```
AkA84: Add a new line before a `return` statement, make the code breathe a little

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+          remove: function () {
+            delete topics[topic][index];
+          }
+        };
+      },
+      publish: function (topic, info) {
+        if (!topics[topic]) return;
```
AkA84: Don't inline `if` statements
```js
  if (!topics[topic]) {
    return;
  }
```

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        topics[topic].forEach(function (item) {
+          item(info != undefined ? info : {});
+          if (!item.$rootScope.$$phase)
```
AkA84: *always* add brackets on `if` statements

AkA84: What are you checking for here?

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        }
+        listener.$rootScope = $rootScope;
+        var index = topics[topic].push(listener) - 1;
+        return {
+          remove: function () {
+            delete topics[topic][index];
+          }
+        };
+      },
+      publish: function (topic, info) {
```
AkA84: Wouldn't the `info` parameter more understandable if it was called `data`?

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        if (!topics[topic]) return;
+        topics[topic].forEach(function (item) {
+          item(info != undefined ? info : {});
```
AkA84: This normalization you do for the `info` (or, as I was suggesting before, `data`), you could do it once before the `forEach`, maybe in a private method.

AkA84: Why are you defaulting to an empty object? Why not `null`?

org.civicrm.reqangular/src/common/services/pub-sub-service.js
AkA84: Please remove the `Service` from the name (and the file name)

contactsummary/js/src/contact-summary/controllers/keyDetails.js
deb1990: If we do that it wont have access to `$log, ContactDetails, Contract, pubSubService` variables.

deb1990: I used first one so that bind can be used. Otherwise another line would be needed to bind
``` 
function getContacts() {}
getContacts.bind(this);
```

contactsummary/js/src/contact-summary/controllers/keyDetails.js
deb1990: kept this the same to maintain consistency

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Generally always an object is sent and in the subscribe methods we check `data.property`, so incase of not sending any value, those cases will generate error. This is to avoid that.

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Checking if digest cycle is i progress. Then only apply otherwise not.

deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

hrjobcontract/js/src/job-contract/controllers/contract-list.js
deb1990: Done

contactsummary/js/src/contact-summary/services/contract.js
deb1990: Structure was getting initialized on top, I had copied the same.Code changed now to remove duplicasy

contactsummary/js/src/contact-summary/services/contract.js
deb1990: Done

contactsummary/js/src/contact-summary/services/contract.js
deb1990: Done

contactsummary/js/src/contact-summary/controllers/keyDetails.js
deb1990: Done

contactsummary/js/src/contact-summary/controllers/keyDetails.js
deb1990: Done

contactsummary/js/src/contact-summary/controllers/keyDates.js
deb1990: Done

contactsummary/js/src/contact-summary/services/api.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub-service.js
```diff
@@ -0,0 +1,32 @@
...
+        var index = topics[topic].push(listener) - 1;
+        return {
+          remove: function () {
+            delete topics[topic][index];
+          }
+        };
+      },
+      publish: function (topic, info) {
+        if (!topics[topic]) return;
+        topics[topic].forEach(function (item) {
```
AkA84: I think you should call the callback parameter `listener` to keep things consistent and more understandable

deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub.js
```diff
@@ -0,0 +1,59 @@
...
+  'use strict';
+
+  var topics = Object.create(null);
+
+  /**
+   * This service is used for communication between different angular apps using pub-sub pattern
+   * @ngdoc service
+   * @name pubSub
+   * @requires $rootScope
+   **/
```
AkA84: Sorry, still find syntax and styling problems with your documentation, let's be on the same page about it:
```js
/**
 * This service is used for communication between different angular apps using pub-sub pattern
 *
 * @ngdoc service
 * @name pubSub
 * @requires $rootScope
 */
services.factory('pubSub', /* .. */)
```

org.civicrm.reqangular/src/common/services/pub-sub.js
```diff
@@ -0,0 +1,59 @@
...
+
+  services.factory("pubSub", ['$rootScope', function ($rootScope) {
+    return {
+      /**
+       * Subscribe method for events
```
AkA84: Again, problem with the documentation. I don't care much about the angular's ngdoc param (we don't use them consistently), but there are three things any method documentation should get right:
1. Description. Should be a good middle ground between being descriptive enough and not being overly detailed
2. Params. All function parameters should be listed with their {type}, name, and if needed description
3. Return value. Should be marked as `@return` and declare the {type} and if necessary the description

With the above said:

```js
/**
 * Subscribe a given listener callback to a specific topic
 *
 * @param {string} topic
 * @param {Function} listener
 * @return {Object} an object that provides the ability to cancel the subscription
 */
subscribe: function (topic, listener) {

}
```

org.civicrm.reqangular/src/common/services/pub-sub.js
```diff
@@ -0,0 +1,59 @@
...
+        index = topics[topic].push(listener) - 1;
+
+        return {
+          remove: function () {
+            delete topics[topic][index];
+          }
+        };
+      },
+
+      /**
```
AkA84: Same thing as above

org.civicrm.reqangular/src/common/services/pub-sub.js
```diff
@@ -0,0 +1,59 @@
...
+      publish: function (topic, data) {
+        if (!topics[topic]) {
+          return;
+        }
+        data = data != undefined ? data : {};
+
+        topics[topic].forEach(function (listener) {
+          listener(data);
+          listener.$rootScope.$applyAsync();
+
```
AkA84: Empty line

org.civicrm.reqangular/src/common/services/pub-sub.js
```diff
@@ -0,0 +1,59 @@
...
+       */
+      publish: function (topic, data) {
+        if (!topics[topic]) {
+          return;
+        }
+        data = data != undefined ? data : {};
+
+        topics[topic].forEach(function (listener) {
+          listener(data);
+          listener.$rootScope.$applyAsync();
```
AkA84: Can you use `$applyAsync` the standard way?
```js
listener.$rootScope.$applyAsync(function () {
  listener(data);
});
```
both `$apply` and `$applyAsync` are expecting an expression to evaluate passed as a param

deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub.js
deb1990: Done

org.civicrm.reqangular/src/common/services/pub-sub.js
deb1990: Done

PR #1157, opened 2016-09-13
PCHR-1572: Show radiobuttons

deb1990:
**Problem**
Radio buttons missing in Job contract Pay page

**Fix**
1. For the custom radiobuttons to work "for" has to be mentioned for the label of the input tag(CSS written that way), which was missing. Adding the same fixes the visibility of radio buttons.
2. Markup modified to show the radio buttons properly.

**Before**
image

**After**
image


AkA84:
Approved

PR #1160, opened 2016-09-14
PCHR-1567: adding missing script

danilo-compucorp:
This PR makes the checkboxes to finish tasks visible again, on tasks and assignments dashboard.

after


brunoqueiros:
Approved

PR #1164, opened 2016-09-15
PCHR-1238: Fix application details not being stored

omarabuhussein:
## Problem

When someone apply for a vacancy , the application details for that applicant are not stored :

selection_153
## Solution

The applicantion form fields are actually a custom fields and storing them was handled by calling **CRM_Core_BAO_CustomValueTable::postprocess()** method but in this case it was called with wrong parameters which prevent custom fields from being stored and displayed . so I just fixed that .

selection_152

---

To be more detailed : this method arguments has been changed in civicrm 4.7 https://github.com/civicrm/civicrm-core/commit/5fc3ea244918837ef10e58b172ab84a97f190b86#diff-927e742818deb9c89a00d2528e360bf0R361

( Related PR : https://github.com/civicrm/civihr/pull/1164)

### DIFF
--- a/hrrecruitment/CRM/HRRecruitment/Form/Application.php
+++ b/hrrecruitment/CRM/HRRecruitment/Form/Application.php
@@ -238,7 +238,7 @@ class CRM_HRRecruitment_Form_Application extends CRM_Core_Form {
       }
 
       //process Custom data
-      CRM_Core_BAO_CustomValueTable::postprocess($params,CRM_Core_DAO::$_nullArray, 'civicrm_case', $caseObj->id, 'Case');
+      CRM_Core_BAO_CustomValueTable::postProcess($params, 'civicrm_case', $caseObj->id, 'Case');
 
       //Process case to vacancy one-to-one mapping in custom table 'application_case'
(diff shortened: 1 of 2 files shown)


hrrecruitment/CRM/HRRecruitment/Form/Application.php
```diff
@@ -238,7 +238,7 @@ public function postProcess() {
       }
 
       //process Custom data
-      CRM_Core_BAO_CustomValueTable::postprocess($params,CRM_Core_DAO::$_nullArray, 'civicrm_case', $caseObj->id, 'Case');
+      CRM_Core_BAO_CustomValueTable::postprocess($params, 'civicrm_case', $caseObj->id, 'Case');
```
coldrunKacper: Would be good to replace 'postprocess' with 'postProcess'. PHP method names are case-insensitive, but as they say (here: http://fr.php.net/manual/en/functions.user-defined.php) :
"Note: Function names are case-insensitive, though it is usually good form to call functions as they appear in their declaration."

omarabuhussein: fixed

coldrunKacper:
Approved.

PR #1165, opened 2016-09-16
PCHR-1542: New individual fix dropdown color and text size

brunoqueiros:
This PR change the selector `.select2-chosen`which were applying the style to all pages and now it just aplly to new individual/organization page. It was updated the dropdown color and text-size.

***Before***
image

***After***
image


deb1990:
Approved

PR #1167, opened 2016-09-16
PCHR-1472: Prevent having more than one contract revision with the same effective date

omarabuhussein:
## Problem 

There cannot be more than one contract revision with the same effective date. 
So a validation  need to be added when creating a new revision to prevent that .

## Solution 

I've created a new API end-point for **HRJobContractRevision** entity called **validateeffectivedate** which is accept a contact_id and effective_date as parameters , the API end-point is an actually a wraper for a new static method **CRM_Hrjobcontract_BAO_HRJobContractRevision::validateEffectiveDate** which handle the validation logic. The API end-point is called via the job contract angular app after click on "Save Revision" and throw an error with the title of the contract that has conflicts with inputted effective date in case there is any :

anim


hrjobcontract/js/src/job-contract/services/contract-revision.js
```diff
@@ -0,0 +1,66 @@
...
+             * @returns {*}
+             */
+            validateEffectiveDate: function(params) {
+                params.effective_date = convertToDateString(params.effective_date);
+                params.sequential = 0;
+                params.debug = settings.debug;
+
+                var deffered = $q.defer(), val;
+
+                ContractRevision.save({
```
danilo-compucorp: You don't need to use `$q` here.

Just return the `$promise` from the `ContractRevision` resource.

coldrunKacper:
Approved.

hrjobcontract/js/src/job-contract/services/contract-revision.js
omarabuhussein: I tired this :

```javascript
                return ContractRevision.save({ action: 'validateeffectivedate', json: params }).$promise
                    .then(function(result){
                        return result.values;
                    });
```
but in the case the Ajax request will not be triggered at all .

danilo-compucorp: Please have a look at http://jsbin.com/qaxamas/edit?js,output

danilo-compucorp:
Approved

PR #1170, opened 2016-09-20
PCHR-1576: Change style of CiviCRM menu

deb1990:
# Requirement
1. Fix menu height, make menu item hover state and delimiter take full height of the menu, so red area on the screenshot, instead of how it is right now.
image-2016-09-16-10-13-03-633

# After
image

### DIFF
--- a/hrui/css/hrui.css
+++ b/hrui/css/hrui.css
@@ -15,3 +15,8 @@
   height: 26px !important;
   line-height: 26px;
 }
+
+#civicrm-menu #crm-qsearch {
+  line-height: 28px;
+  height: 30px !important;
+}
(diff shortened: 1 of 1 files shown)


AkA84:
Are we sure we want this? the hover state of the menu item is supposed to be like that (see [here](http://dmaster.demo.civicrm.org/civicrm/dashboard)

To be honest this solution doesn't look right at all, now the nav items go all the way down but the text now is not vertically centered anymore and the search bar also looks decentered

JonScreat:
I agree with Alessandro here. It looks odd to not have the text centrally aligned.

mi-kon:
Tbh, I don't like how civicrm menu looks either, it looks odd to me as well.  But I also agree that the alignment is off now and that's not good. Let's hold off merging this fix i guess...

deb1990:
Should I center the text and searchbar vertically?

PR #1171, opened 2016-09-20
PCHR-1442: Hide inactive pay scales in job contract modal

omarabuhussein:
## Problem 

If you have a disabled pay scale 

selection_171

then it will appear when creating/editing job contracts but it should not :

selection_172

## Solution 

I changed getPayScaleGrade method in job-contract utils service to pass is_active as paramter when requesting pay scales list .

Also I added a new upgrader to create civicrm_hrpay_scale and its default data if it is not exists since it had been removed in another commit mistakenly and I forget to run Importer tests at that time  so it end up undiscovered until now. 

I also removed some unused methods in job contract upgrader.

selection_169


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -920,6 +857,41 @@ function upgrade_1018() {
...
+        ('US', 'Senior', 'USD', 38000, 'Year', 1),
+        ('US', 'Junior', 'USD', 24000, 'Year', 1),
+        ('UK', 'Senior', 'GBP', 35000, 'Year', 1),
+        ('UK', 'Junior', 'GBP', 22000, 'Year', 1),
+        ('Not Applicable', NULL, NULL, NULL, NULL, 1)
+    ");
+      return true;
+    }
+
+    return false;
```
davialexandre: Is this return false ok? The upgrader method will return false if there pay scales, right? Meaning that the upgrade will stop in that case. Is that the expected behavior? What if we have other upgrader methods to be executed after this one?

omarabuhussein: oops .. thanks for spotting this mistake

hrjobcontract/js/src/job-contract/services/utils.js
```diff
@@ -118,7 +118,8 @@ define([
                 var deffered = $q.defer();
 
                 API.resource('HRPayScale','get',{
-                    sequential: 1
+                    sequential: 1,
+                    is_active: 1
```
davialexandre: Job contracts doesn't have much test coverage, so I think it would be great if you could add a test to cover this change.

omarabuhussein: mmmm I am not really sure about this .. this method only make an ajax call to get some data so I Am not sure how it could be tested ... at the end , the ajax call will operate on the database you are using and not on test database .

AkA84: The thing you can test is simply: does `getPayScaleGrade` call the `API.resource` service with the following parameter: `'HrPayScale', 'get', ' { sequential: 1, is_active: 1}`?

You will mock the 'API' service to do that, so no ajax request will be made:
```js
spyOn(API, 'resource');
// ...
expect(API.resource).toHaveBeenCalledWith(/* list of param */);
```

davialexandre: hmmmm.... taking a second look at this I agree it's not easy to figure out what to test here. Maybe you can mock the API object and check if the gets called with the is_active parameter?
 Let's see what Alessandro thinks about this

omarabuhussein: :smile:

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,47 @@
+define([
+    'common/angular',
+    'common/angularMocks',
+    'job-contract/app'
+], function (angular, Mock) {
+    'use strict';
+
+    describe('UtilsService', function () {
+        var UtilsService;
```
AkA84:
```js
var UtilsService, API;
```

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,47 @@
...
+
+        beforeEach(module('hrjc'));
+        beforeEach(inject(function (_UtilsService_, _API_) {
+            UtilsService = _UtilsService_;
+            API = _API_;
+        }));
+
+        beforeEach(function () {
+            spyOn(API, 'resource').and.callFake(function () { return { get: function () {} } })
+            UtilsService.getAbsenceType();
```
AkA84: I think the only thing that should be in this common `beforeEach` should be the `spyOn` bit, all the others are specific to the `describe`s for each method

```js
describe('getAbsenceType', function () {
  beforeEach(function () {
    UtilsService.getAbsenceType();
  });
});

// etc..
```

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,47 @@
...
+            UtilsService.getHoursLocation();
+            UtilsService.getPayScaleGrade();
+        });
+
+        describe('getAbsenceType', function () {
+            it('Test getAbsenceType', function () {
```
AkA84: this is not phpunit where there is no way to decently give a description to the test :D give a meaningful one

(remember to be imperative, as in "does this...". instead of "should do this...")

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,47 @@
...
+                var params = { return: 'id,name,title' };
+                expect(API.resource).toHaveBeenCalledWith('HRAbsenceType', 'get', params);
```
AkA84: why not simply
```js
expect(API.resource).toHaveBeenCalledWith('HRAbsenceType', 'get', { return: 'id,name,title' });
```
_(same for the other tests)_

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,47 @@
...
+            });
+        });
+
+        describe('getPayScaleGrade', function () {
+            it('Test getPayScaleGrade', function () {
+                var params = { sequential: 1, is_active: 1 };
+                expect(API.resource).toHaveBeenCalledWith('HRPayScale', 'get', params);
+            });
+        });
+
```
AkA84: empty line

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,51 @@
+define([
+    'common/angular',
+    'common/angularMocks',
+    'job-contract/app'
+], function (angular, Mock) {
+    'use strict';
```
AkA84: Just realized you are using a 4 space indentation here, please use 2 (aren't we using and editorconfig file for this?)

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,51 @@
...
+        beforeEach(function () {
+            spyOn(API, 'resource').and.callFake(function () { return { get: function () {} } })
+        });
+
+        describe('getAbsenceType', function () {
+            beforeEach(function () {
+                UtilsService.getAbsenceType();
+            });
+
+            it("API parameters must be called with 'HRAbsenceType', 'get' and { return: 'id,name,title' }", function () {
```
AkA84:
```js
it('returns the id, name, and title of the absence types', function () { })
```

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,51 @@
...
+                expect(API.resource).toHaveBeenCalledWith('HRAbsenceType', 'get', { return: 'id,name,title' });
+            });
+        });
+
+        describe('getHoursLocation', function () {
+            beforeEach(function () {
+                UtilsService.getHoursLocation();
+            });
+
+            it("API parameters must be called with 'HRHoursLocation', 'get' and { sequential: 1, is_active: 1 }", function () {
```
AkA84:
```js
it('returns only the active hours/location entries', function () {})
```

hrjobcontract/js/test/services/utils_test.js
```diff
@@ -0,0 +1,51 @@
...
+                expect(API.resource).toHaveBeenCalledWith('HRHoursLocation', 'get',  { sequential: 1, is_active: 1 });
+            });
+        });
+
+        describe('getPayScaleGrade', function () {
+            beforeEach(function () {
+                UtilsService.getPayScaleGrade();
+            });
+
+            it("API parameters must be called with 'HRPayScale', 'get' and { sequential: 1, is_active: 1 }", function () {
```
AkA84:
```js
it('returns only the active pay scale entries', function () {})
```

hrjobcontract/js/test/services/utils_test.js
omarabuhussein: yes , but in another PR ( I guess ) you told me that the indentation is wrong ( where it was 2 spaces ) so I thought you have different configuration for js files and you use 4 spaces instead of 2 as for php files  but that configuration is not specified in the editoconfig file .

AkA84: i don't remember doing that :| if i did it i apologies, but the agreement [see here](https://compucorp.atlassian.net/browse/PCHR-1022) was to have 2 spaces for everything. does the .editorconfig discriminates based on the file type though? You're saying it works for php files but not for js files?

omarabuhussein: > You're saying it works for php files but not for js files?

nope I just thought maybe you front end guys use a different config file that use 4 indentation for s files.

PR #1172, opened 2016-09-20
PCHR-1582: Small optimizations

AkA84:
The PR contains small optimizations that could be bundled together

### Minify `hrui` scripts
In the `hrui` extensions, the scripts are neither minified nor bundled together, meaning that we are doing unnecessary additional requests to the server.

I've used the usual `gulp` to minify and concat the files, which are then served by the extension

### Strip multiline comments in minified `org.civicrm.bootstrapcivicrm's` css file
The file still contained the multiline comments because of the vendors css included in it. This is because the `gulp-sass` module is set *not* to remove those even at the max compression setting. Using `gulp-strip-css-comments` as an additional step solves the issue

### Do not load the style guide css in every page
The `org.civicrm.styleguide` css was mistakenly loaded in every page instead of being loaded just in _/styleguide_

### Do not re-add extensions' resources in ajax snippets
_(This is not really a performance optimization per se given that the browser cache mechanism would prevent additional requests for the same file, but it's just bad that it happens)_

Every resource file (js, css) added via the `_pageRun` hook in any of the extensions is added over and over every time a user clicks on a tab in the contact summary page (or when a widget is loaded in the dashboard). It seems that rendering the tabs/widget content triggers the hook, and as a result we end up adding multiple times, for example, _reqangular.min.js_ (almost 900kb) and _bootstrap.css_ (almost 350kb).

This was prevented by simply checking if $_GET['snippet'] exists and if it's equal to `'json'`. If it does and it is, then we don't add the resource

----

Additionally I fixed this notice that would popup in the admin:
> Notice: Undefined index: operator in CRM_Core_BAO_Navigation::getMenuName()

This was due to the "Import Job Roles" menu item not having the `operator` property set (see [here](https://github.com/civicrm/civihr/blob/18519e12b4de5ac22e2efbbe587e04baf74aeae5/com.civicrm.hrjobroles/hrjobroles.php#L118)]. Adding the operator solves the issue


hrui/hrui.php
```diff
@@ -48,18 +45,19 @@ function hrui_civicrm_pageRun($page) {
...
           'cid' => CRM_Utils_Request::retrieve('cid', 'Integer'),
-          'hideGId' => $hideGId));
+          'hideGId' => $hideGId)
+        );
     }
   }
 
-  CRM_Core_Resources::singleton()
-    ->addScriptFile('org.civicrm.hrui', 'js/perfect-scrollbar/perfect-scrollbar.min.js');
+  CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/dist/hrui.min.js');
```
davialexandre: This mean we'll get the minified version even if debug is activate, right? Do you think it's possible/necessary that we get the minified file only if debug is disabled?

coldrunKacper:
Latest two commits approved.

PR #1173, opened 2016-09-21
PCHR-1566: Administer Subitems Redesign

kkalamarski:
# What has been changed

The goal of this PR is to "update civicrm styles to match civihr styles". Therefore I was asked not to change any layout structure (that's why this PR consists only of CSS). The biggest problem was that the page structure is very different in most of the cases. It was impossible to write only one - global - css rule, that's why there's many page-specific stylesheets. To keep the code clean and "DRY" i used mixins for common parts. 

There were made changes to almost all screens under `Administer` menu - it's not possible to include all screenshots here, that's why I'll leave screenshots for specific types of components:
### before

civicrm system status - before
### after

civicrm system status
### before

custom data fields - before
### after

custom data fields
### before

profiles - before
### after

profiles
### before

display settings - before
### after

display settings
### before

extensions - before
### after

extensions
# Update 19.10.2016
### before

resp_tables_before
### after (horizontal scroll possible)

resp_tables_after
## Popup is not truncated anymore

This is a regression issue caused by setting table wrapper overflow to scroll.
This is fixed by adding extra `padding-bottom` to container, and cutting it using `margin-bottom` which equals negative value of `padding-bottom`.

``` css
overflow: scroll;
padding-bottom: 150px;
margin-bottom: -150px;
```
### before

popup_before
### after

popup_after
# Update 24.10.2016

Fixing issues pointed by @JonScreat after design testing. 
Commits include following changes:
- Fixed few overflowing tables to be scrollable.
- To avoid overflowing tables in vertical axis by popup, decided to change the look pf popup itself, rather than changing it's parent table. Result is seen below (compare with above screenshots).
### 'More' Popup - after:

more-popup

@JonScreat _justifying choice of horizontal popup_ - I was facing the problem, tables would either overflow or would truncate the popup. The problem was caused by CSS not allowing me to scroll content horizontally, while vertically content would still be visible. Tried many solutions, but all of them had some side effects (i.e. scrollbar overflowing over footer). The only solution that does not break anything is setting popup into horizontal layout. 
# Update 25.10.2016

Fixed popup to be vertical again + applied styles to match civihr design. The solution I chose was to write simple JS plugin displaying popup in right place with `position: absolute`.
Result seen below:

new popup

popup-gif
## Attrchange Plugin

`$.fn.attrchange` This plugin creates `MutationObserver` which listens to all changes to DOM Node. If change has been done to it's attributes callback will be called.

``` javascript

$(<selector>).attrchange(function(e) {
  // attrubite changed
});
```
### attrchange with civihr popup

Plugin is bound to `$('.btn-slide')` - container and button triggering popup. When button is clicked a class `btn-slide-active` is added to it, which triggers `attrchange`.

`attrchange` checks if class `btn-slide-active` is present and if so it calculates button offset, moves popup under `body` element and sets it's position to absolute. Later applies properties `top` and `left` according to button offset.

When popup close `btn-slide-active` class is removed from button. This triggers `attrchange` again. This time this class is no longer present and popup is moved back under button and it's position is set to static.


JonScreat:
On the whole this looks good to me. The only thing is that there shouldn't be banding/alternate coloured rows anymore for the tables. Other than that I think it's good.

org.civicrm.bootstrapcivicrm/gulpfile.js
```diff
@@ -2,31 +2,29 @@ var gulp = require('gulp');
 var bulk = require('gulp-sass-bulk-import');
 var sass = require('gulp-sass');
 var postcss = require('gulp-postcss');
-var stripCssComments = require('gulp-strip-css-comments');
```
AkA84: Why did you remove this? It was added in one of the latest PRs

org.civicrm.bootstrapcivicrm/gulpfile.js
```diff
@@ -2,31 +2,29 @@ var gulp = require('gulp');
...
+gulp.task('sass-civicrm', function (next) {
   gulp.src('scss/civicrm/custom-civicrm.scss')
     .pipe(bulk())
     .pipe(sass({ outputStyle: 'compressed' }).on('error', sass.logError))
     .pipe(postcss([postcssPrefix({
       prefix: '.crm-container ',
       exclude: [/page-civicrm/, /crm-container/]
     }), postcssDiscardDuplicates]))
-    .pipe(gulp.dest('css/'));
+    .pipe(gulp.dest('css/')).on('end', next);
```
AkA84: What's the purpose of this?

kkalamarski:
@JonScreat you mean odd/even colouring or green rows in Extension table?

JonScreat:
I mean tables in custom data fields and profiles screens. They should just have white backgrounds. thanks

kkalamarski:
Ok, I see will change this

org.civicrm.bootstrapcivicrm/gulpfile.js
kkalamarski: uh, sorry, messed up with resolving conflicts - will fix

org.civicrm.bootstrapcivicrm/gulpfile.js
kkalamarski: after running `gulp sass` command task was finished immediately, but output file was not yet generated. 
By passing this callback  `gulp sass` waits until tasks `sass-civicrm` and `sass-bootstrap`  finish, and later display to console message indicating that all tasks are done.

`sass-civicrm` takes ~ 11 sec

kkalamarski:
@AkA84 Actually i did use some, but will search for variables I duplicated, delete them and use ones from theme.

JonScreat:
Sorry to say that I'm not a fan of the new layout for the 'More' pop up - I think it looks a little confusing to have them all in a row.

Also, the styling of the pop up should be as seen on this page if at all possible: https://invis.io/WX71LBMPJ#/172206676_5-2b_Admin_Leave_Calendar_Dropdown

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -130,6 +130,7 @@ function bootstrapcivicrm_civicrm_pageRun($page) {
     CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/radio-checkbox.js');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/add-missing-date-addons.js');
+    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/civihr-popup.js');
```
AkA84: I'm not sure this scripts belong to the bootstrap extension, i think it makes more sense to have it in `hrui`

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
+/*
+Prevent Popups to overflow tables.
+*/
+CRM.$(function($) {
+  'use strict';
+
+  (function($) {
```
AkA84: You are putting in the same file both the plugin definition (line 7-28) and its implementation (line 30-59). I think you should separate the two

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+    var MutationObserver = window.MutationObserver || window.WebKitMutationObserver || window.MozMutationObserver;
+
+    $.fn.attrchange = function(callback) {
```
AkA84: _(Applies for all instances of this)_

If the function is anonymous, leave a whitespace between `function` and the brackets
```js
function (callback) {
```

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+      if (MutationObserver) {
+        var options = {
```
AkA84: You can save space and put everything in a line, there are only two options
```js
var options = { subtree: false, attributes: true };
```

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
AkA84: Document what this plugin does

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+        return this.each(function() {
+          observer.observe(this, options);
+        });
+      }
+    }
+  })($);
+
+  // .btn-slide is a continer of popup and a button triggering it
+  $('.btn-slide').attrchange(function(e) {
+    var $button = $(this),
```
AkA84: Put everything in one line, i don't think it will go over the 80 columns mark

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+        $popup,
+        button_offset;
+
+    if($button.hasClass('btn-slide-active')) {
```
AkA84:
```js
if ($button.hasClass('btn-slide-active')) {
```

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+      $('body').addClass('civihr-popup-open');
+    } else {
+      // popup closed
+
+      $('body > ul.panel')
+        .appendTo($button)
+        .css('position', 'static')
+        .css('display', 'none');
+
+      $('body').removeClass('civihr-popup-open');
```
AkA84: You should cache `$('body')` since you are reusing it multiple times
```js
var $body = $('body');
```

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+    if($button.hasClass('btn-slide-active')) {
+      // check if popup is open
+      $popup = $button.children('ul.panel');
+
+      $popup.appendTo('body')
+        .addClass('civihr-popup');
+
+      button_offset = $button.offset();
+
+      $popup.css('position', 'absolute')
```
AkA84: Be consistent
```js
$popup
  .css('position', 'absolute')
  .css(/* .. */)
  .css(/* .. */)
```

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+  })($);
+
+  // .btn-slide is a continer of popup and a button triggering it
+  $('.btn-slide').attrchange(function(e) {
+    var $button = $(this),
+        $popup,
+        button_offset;
+
+    if($button.hasClass('btn-slide-active')) {
+      // check if popup is open
```
AkA84: This comment should go above the `if`, otherwise it's confusing, it seems like you are checking something *inside* the `if`, but you're not

JonScreat:
looks good to me

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+      $popup = $button.children('ul.panel');
```
AkA84: You are only expecting to have one `ul.panel` in a given button right? You could make it more explicit
```js
$popup = $button.find('ul.panel').first();
```
or something like that

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+
+      $popup.appendTo('body')
+        .addClass('civihr-popup');
```
AkA84: Why do you need to add an additional class to the popup? Can't you style with what you already have?

AkA84: Why do you need to add an additional class to the popup? Can't you style with what you already have?

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+
+      button_offset = $button.offset();
```
AkA84: it's `buttonOffset`, in js you use camelCase

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+
+      $popup.css('position', 'absolute')
+        .css('top', parseInt(button_offset.top, 10) + 24) // 24 is the fontsize + paddings
+        .css('left', parseInt(button_offset.left, 10) - ($popup.width() - 57)); // 57 is the width of more button
```
AkA84: `24` and `57` seem magic numbers, isn't there a way to get the more programmatically, or avoid using them altogether?

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
AkA84: It seems that this could be handled by css better.

You could either write a css selector like
```css
body > {popup class} { position: absolute; }
```
or use the `.civihr-custom` class not for styling it (see my previous comment), but just for applying the `position` property

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
...
+
+      $('body').addClass('civihr-popup-open');
+    } else {
+      // popup closed
+
+      $('body > ul.panel')
+        .appendTo($button)
+        .css('position', 'static')
+        .css('display', 'none');
```
AkA84: Again, these two classes added via JS seem a bad idea. Aren't those properties, with those values, the default for the popup anyway? (`static` is the default value for any element, and having `display: none` seem like the default state of a popup anyway)

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -0,0 +1,62 @@
+/*
+Prevent Popups to overflow tables.
```
AkA84: How does this logic deals with multiple popups? What if i click on a popup and then immediately click on another one without first closing the other?

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
kkalamarski: Yeah, it makes sense

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
kkalamarski: actually I'm afraid it may cause conflicts. There could be many `ul.panel` elements, not necessarily serving as popups, also this name is not descriptive. I think that adding this class may save us from some troubles in the future.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
kkalamarski: @AkA84 I moved plugin to `hrui`. Should I move the rest too? (meaning 
```javascript
$(...).attrchange(() => { 
  // this too?
});
```
)

AkA84: @kkalamarski yes please

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -0,0 +1,22 @@
+(function($) {
+  var MutationObserver = window.MutationObserver || window.WebKitMutationObserver || window.MozMutationObserver;
+
+  $.fn.attrchange = function(callback) {
```
AkA84: Document the plugin

org.civicrm.bootstrapcivicrm/js/civihr-popup.js
```diff
@@ -6,33 +6,31 @@ CRM.$(function($) {
 
   // .btn-slide is a continer of popup and a button triggering it
   $('.btn-slide').attrchange(function(e) {
-    var $button = $(this),
-        $popup,
-        button_offset;
+    var $button = $(this), $body = $('body'), $popup, buttonOffset;
 
+    // check if popup is open
     if($button.hasClass('btn-slide-active')) {
```
AkA84:
```js
if ($button.hasClass('btn-slide-active')) {
```

PR #1175, opened 2016-09-22
PCHR-1584: Improve the loading of tabs' resources

AkA84:
## Problem
The resources (css/jss) for angular apps contained in the Contact Summary tabs are loaded immediately on page load, instead of being loaded only when their tab is first displayed. This result in unnecessary requests to the server

before

## Solution
Instead of loading the files using the `_pageRun` hook and the `CRM_Contact_Page_View_Summary` object
```php
function hrjobroles_civicrm_pageRun($page) {
    // load resources ...
  }
}
```
the resources are now loaded via the individual page's `run` method, for example:
```php
class CRM_Hrjobroles_Page_JobRoles extends CRM_Core_Page {

  function run() {
    CRM_Utils_System::setTitle(ts('JobRoles'));

    self::registerScripts();
    parent::run();
  }

  private static function registerScripts() {
    // load files
  }
}
```
The end result is far less requests and data transferred on initial page load
after

This was done also for the `contactsummary` extension even if it's the tab to be displayed by default, just to keep the approach consistent.

## Problem
The way the tab's angular apps were bootstrapped was a bit convoluted, with the assumption (i believe) that it wasn't known which would load first, the template file or the javascript file. So both ended up registering and emetting the "init"/"ready" signals to cater for both scenarios
```html
<div id="{$module}" hrjc-loader hrjc-loader-show="true">
  <div class="container" ng-view>
  </div>
</div>
<script type="text/javascript">
  (function(){
    function hrjcInit(){
      document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('hrjcInit') : (function(){
        var e = document.createEvent('Event');
        e.initEvent('hrjcInit', true, true);
        return e;
      })());
    };
    hrjcInit();

    document.addEventListener('hrjcReady', function(){
      hrjcInit();
    });
  })();
</script>
```
```js
require([
  'job-contract/app'
], function () {
  'use strict';

  document.addEventListener('hrjcInit', function(){
    angular.bootstrap(document.getElementById('hrjob-contract'), ['hrjc']);
  });

  document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('hrjcReady') : (function(){
    var e = document.createEvent('Event');
    e.initEvent('hrjcReady', true, true);
    return e;
  })());
});
```

## Solution
Given that now the individual app's script file is always appended after the template, the bootstrap logic can be simplified, with the template listening for the "ready" signal that the js file will emit once loaded
```html
<div id="{$module}" hrjc-loader hrjc-loader-show="true">
  <div class="container" ng-view>
  </div>
</div>
<script type="text/javascript">
  document.addEventListener('hrjcReady', function () {
    angular.bootstrap(document.getElementById('hrjob-contract'), ['hrjc']);
  });
</script>
```
```js
require([
  'job-contract/app'
], function () {
  'use strict';

  document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('hrjcReady') : (function(){
    var e = document.createEvent('Event');
    e.initEvent('hrjcReady', true, true);
    return e;
  })());
});
```


hrjobcontract/hrjobcontract.php
```diff
@@ -278,10 +278,10 @@ function hrjobcontract_civicrm_buildForm($formName, &$form) {
  * to give some room for other extensions to place
  * their tabs between these two.
  */
-function hrjobcontract_civicrm_tabs(&$tabs) {
+function hrjobcontract_civicrm_tabs(&$tabs, $contactId) {
```
davialexandre: Just a note: The _tabs hook has been deprecated in civicrm 4.7 (https://wiki.civicrm.org/confluence/pages/viewpage.action?pageId=86213406) and _tabSet should be used instead. Since we use _tabs in a lot of other extensions, I think we need a separate ticket to update them all at once.

PR #1183, opened 2016-09-26
PCHR-1600: Remove link from LeavePeriodEntitlement to LeaveBalanceChange

davialexandre:
Instead of using the entitlement_id FK, the LeavePeriodEntitlement will now be linked to LeaveBalanceChange via a polymorphic relationship using the source_id and source_type fields. Every LeaveBalanceChange record linked to a LeavePeriodEntitlement will have the value 'entitlement' on the source_type field and the related period entitlement id on source_id.

Since we had to create the new source_type field on LeaveBalanceChange, we also needed new value for this field for when the LeaveBalanceChange is linked to a LeaveRequestDay. In that case, the source_type is 'leave_request_day'

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -104,7 +104,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
       FROM {$balanceChangeTable}
       WHERE source_id = {$entitlementID} AND
             source_type = '" . self::SOURCE_ENTITLEMENT . "' AND 
-            expired_balance_id IS NULL
+            expired_balance_change_id IS NULL
       ORDER BY id
     ";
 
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -59,15 +62,15 @@ public static function getBalanceForEntitlement($entitlementID, $leaveRequestSta
       FROM {$balanceChangeTable} leave_balance_change
       LEFT JOIN {$leaveRequestDateTable} leave_request_date 
              ON leave_balance_change.source_id = leave_request_date.id AND 
-                leave_balance_change.source_type = 'leave_request_day'
+                leave_balance_change.source_type = '". self::SOURCE_LEAVE_REQUEST_DAY ."'
```
omarabuhussein: it is better to stick with curly braces style since you used it in the rest of the code {self::SOURCE_LEAVE_REQUEST_DAY} or maybe it it doesn't work with non variables ?

omarabuhussein: just checked and it do not work :

> Using single curly braces ({}) will not work for accessing the return values of functions or methods or the values of class constants or static class variables.

so don't worry about it

PR #1185, opened 2016-09-27
PCHR-1586: Improve api calls of Job Roles extension

AkA84:
## Problem
The following API calls are made by the angular apps, but could be grouped together to decrease the numbers of requests to the server
```
OptionGroup, get, { name: { IN : [ "hrjc_department", "hrjc_region", "hrjc_location", "hrjc_level_type", "cost_centres" ] } }
OptionValue, get, { name: { IN : [ “1”, “2”, “3”, “4”, “5” ] } }
```
```
HRJobContract, get, { contact_id: ‘1’, deleted: 0, return: "title, period_end_date, period_start_date" }
HRJobContractRevision, get, { jobcontract_id: ‘1’ }
HRJobDetails, get, { jobcontract_revision_id: ‘1’ }
```

before

## Solution
_(Note: the api calls were refactored but not how the returned data was rearranged. This was done to avoid a snowball effect that would have bloated this PR outside its scope)_

### Option values
To get the option values directly, the new ability to join entities via the api was used.
```js
  'sequential': 1,
  'option_group_id.name': { 'IN': groupNames },  // join on OptionGroup
  'return': [ 'id', 'label', 'weight', 'value', 'option_group_id', 'option_group_id.name' ],
  'options': {
    'limit': 1000,
    'sort': 'id'
  }
})
```

### Revision details
To get the revision details directly, the join functionality could have been used but would have been highly impractical, due to the fact that every field of every entity would have had to be specified (it is not possible to use a "\*" wildcard, like `jobcontract_id.* `, in the `return` option of an api call). This would have made the solution less maintainable, since that if any of fields of the entities involved were changed, the api call details would have to be changed as well (not to mention possible introduction of regression issue if a field were to be missed in the `return` list)

For this reason a less "advanced" but more safe approach was chosen, by chaining the API calls
```js
CRM.api3('HRJobContract', 'get', {
  'sequential': 1,
  'contact_id': contactId,
  'deleted': 0,
  'return': 'title,period_end_date,period_start_date',
  'api.HRJobContractRevision.get': {
    'jobcontract_id': '$value.id',
    'api.HRJobDetails.getsingle': {
      'jobcontract_revision_id': '$value.id'
    }
  }
})
```
after


com.civicrm.hrjobroles/js/src/job-roles/services/hr-job-roles-service.js
```diff
@@ -1,103 +1,69 @@
...
+         * properties, then removes the current one and format the dates
+         *
+         * @param  {Object} contract
+         */
+        function processContractRevisions(contract) {
+          var contractRevisions = contract['api.HRJobContractRevision.get'].values;
+          delete(contract['api.HRJobContractRevision.get']);
+
+          contract.revisions = contractRevisions
+            .map(function (revision) {
```
deb1990: Instead of looping 3 times, we can loop once using for loop and make all the adjustments

PR #1188, opened 2016-09-28
PCHR-1585: Improve api calls of Job Contract extension

AkA84:
## Problem
_(all the examples are taken from a contact with 1 contract with 3 revisions)_

The Job Contract api call makes, on initial load, a large amount of api calls: 2 for fetching options + 7 for each contract + 3 for each revision:

before

The are re different aspects of how we load the contract data that contribute to the overall problem:

### Revisions data loaded from the start
Despite being hidden initially in a different tab, the revisions' details are immediately loaded.

### Revisions data loaded across different calls
Each revision's data (details, hour, pay) is fetched across different api calls instead of being fetched all at once using aggregation features (chain, join) offered by the CiviCRM api.
```js
promiseRevisionList.push($q.all({
  revisionEntityIdObj: revision,
  details: ContractDetailsService.getOne({
    jobcontract_revision_id: revision.details_revision_id,
    return: 'position, location'
  }),
  hour: ContractHourService.getOne({
    jobcontract_revision_id: revision.hour_revision_id,
    return: 'hours_type'
  }),
  pay: ContractPayService.getOne({
    jobcontract_revision_id: revision.pay_revision_id,
    return: 'pay_scale, pay_annualized_est, pay_currency'
  }),
  files: $q.all({
    details: ContractFilesService.get(revision.details_revision_id,'civicrm_hrjobcontract_details')
  })
}));
```

### Contract details loaded across different api calls
Same goes for the contract's details (details, hour, leave, pay, health, pension)
```js
$q.all({
  details: ContractDetailsService.getOne({ jobcontract_revision_id: revisionEntityIdObj.details_revision_id }),
  hour: ContractHourService.getOne({ jobcontract_revision_id: revisionEntityIdObj.hour_revision_id }),
  pay: ContractPayService.getOne({ jobcontract_revision_id: revisionEntityIdObj.pay_revision_id }),
  leave: ContractLeaveService.get({ jobcontract_revision_id: revisionEntityIdObj.leave_revision_id }),
  health: ContractHealthService.getOne({ jobcontract_revision_id: revisionEntityIdObj.health_revision_id }),
  pension: ContractPensionService.getOne({ jobcontract_revision_id: revisionEntityIdObj.pension_revision_id })
})
.then(function (results) {
});
```

## Solution
Addressing all the problems listed above (see below for the solution breakdown) considerably reduced the number of calls made on initial load

after

### Revisions are loaded only on tab selection
The revision data is now loaded only when the user decides to take a look at the "Full History" tab

full-history-tab

### Revision details are loaded at once
Leveraging the join functionality of the CiviCRM api, the revision details can now be aggregated in one single api call
```js
$q.all({
  files: $q.all({
    details: ContractFilesService.get(revision.details_revision_id,'civicrm_hrjobcontract_details')
  }),
  aggregated: ContractRevisionService.get({
    action: 'getsingle',
    json: {
      sequential: 1,
      id: revision.id,
      return: [
        'details_revision_id.position',
        'details_revision_id.location',
        'hour_revision_id.hours_type',
        'pay_revision_id.pay_scale',
        'pay_revision_id.pay_annualized_est',
        'pay_revision_id.pay_currency'
      ]
    }
  })
});
```

### Contract details are loaded at once via a new custom api endpoint
A new custom api endpoint, `HRJobContract.getfulldetails`, had been added so that all the contract details can be called at once
```js
ContractService
  .fullDetails(contractId)
  .then(function (results) {
  });
```

---

## Problem
The `ContractRevisionService` service was being duplicated within the `Contract` service file for no reason
```js
services.factory('ContractRevision', [ /* ... */, function (/* ... */) {
  //.. 
  return $resource(settings.pathRest,{
    action: 'get',
    entity: 'HRJobContractRevision',
    json: {}
  })

  services.factory('ContractService', [ '$log', '$q', 'Contract', 'ContractRevision', /* ... */
```

## Solution
The duplicate had been removed and the original had been changed so that it could be used by the `Contract` service

--

## Problem
In `RevisionListCtrl` the `$scope.$parent` property was used an arbitrary number of times in order to access parent scope's property

```js
revisionDataListLocal = $scope.revisionDataList = $scope.$parent.$parent.$parent.$parent.revisionDataList;
```

## Solution
There is no need for doing that, every parent scope property is inherited by the current scope
```js
revisionDataListLocal = $scope.revisionDataList;
```

--

## Problem
The `CRM_Hrjobcontract_Page_JobContractTab:: getRevisionChangeReasons` and `CRM_Hrjobcontract_Page_JobContractTab:: getContractEndReasons` were not used anymore

## Solution
The two methods had been removed


hrjobcontract/api/v3/HRJobContract.php
```diff
@@ -204,3 +204,56 @@ function civicrm_api3_h_r_job_contract_getcurrentcontract($params) {
...
+    'health'  => 'getsingle',
+    'pension' => 'getsingle',
+  ];
+
+  // get the current revision, holds a pointer to the revision each
+  // detail "category" points to (ie: details_revision_id, leave_revision_id)
+  $currentRevision = _civicrm_hrjobcontract_api3_get_current_revision($params)['values'];
+
+  foreach ($entitiesToChain as $entity => $action) {
+    $callParams["api.HRJob" . ucfirst($entity) . ".$action"] = [
```
davialexandre: It's not a big deal, but since you're using this chain key logic twice here, what about extracting it to another function?

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -210,4 +236,84 @@ protected static function clearEffectiveEndDate($revisionId) {
...
+        $query['select'][] = "{$entityAlias($entity)}.{$field} as {$entity}__{$field}";
+      }
+
+      $query['from'][] = "civicrm_hrjobcontract_{$entity} {$entityAlias($entity)}";
+      $query['where'][] = "{$entityAlias($entity)}.jobcontract_revision_id = " . $revision["{$entity}_revision_id"];
+    }
+
+    return sprintf(
+      "SELECT %s FROM %s WHERE %s ",
+      join(', ', $query['select']),
```
omarabuhussein: join is just an alias for implode but implode is more popular

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -210,4 +236,84 @@ protected static function clearEffectiveEndDate($revisionId) {
...
+  protected static function buildFullDetailsQuery($entities, $revision) {
+    $query = ['select' => [], 'from' => [], 'where' => []];
+
+    /**
+     * Creates an alias for the given entity
+     *
+     * @param string $entity
+     * @return  string
+     */
+    $entityAlias = function ($entity) {
```
davialexandre: I'm ok with that. I've done it myself a few time on another project. I would just ask to remove the trailing semicolon after the closing bracket

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -53,6 +53,32 @@ public static function create($params) {
   }
 
   /**
+   * Builds and executes a query that return all the details of a given revision
+   *
+   * Note: as there could be multiple leave records, the leave data must be
+   * fetched and processed separately
+   *
+   * @param  Array $revision
```
davialexandre: please use array instead of Array for the @param type (applies for all the other methods)

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
davialexandre: I think you don't need this entityAlias function at all since you can just use the entity as the alias. What do you think?
Besides, the way it's was implemented, it can cause problems in the future. What if we need to add a new entity which starts with the same 3 letters as one of those entities?

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -210,4 +236,84 @@ protected static function clearEffectiveEndDate($revisionId) {
...
+   * The field's entity is inferred by the prefix ({entity}__) in the field's name
+   *
+   * @param  Array $result
+   * @return Array
+   */
+  protected static function normalizeFullDetailsResult($result) {
+    $normalized = [];
+
+    foreach ($result as $key => $value) {
+      if (strpos($key, '_') == 0) { continue; } // ignores "internal" fields
```
davialexandre: Instead of strpos, you can do something like this:

```php
if($key[0] == '_')
```

It's faster and more secure, since you won't have to mind about strpos which can return 0 and false

omarabuhussein:
okay generally it is fine to ( I only left a note about implode and join ) ... it also lack tests ( I should not expect a PHP unit tests from a front end developer anyway or should I ? :smile:  ) .. so otherwise these points it is approved from my side

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -210,4 +236,84 @@ protected static function clearEffectiveEndDate($revisionId) {
...
+      $normalized[$entity][$field] = self::isJSON($value) ? json_decode($value) : $value;
+    }
+
+    return $normalized;
+  }
+
+  /**
+   * Determines if the given string is a JSON or not
+   *
+   * @param  String  $string
```
davialexandre: Please using string instead of String as the param type

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
davialexandre: can you please add a description to the $revision param saying what is the expected format for the array? I had to read all the code to figure that out

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -53,6 +53,32 @@ public static function create($params) {
...
   /**
+   * Builds and executes a query that return all the details of a given revision
+   *
+   * Note: as there could be multiple leave records, the leave data must be
+   * fetched and processed separately
+   *
+   * @param  Array $revision
+   * @return Array
+   */
+  public static function fullDetails($revision) {
```
davialexandre: can you please add some tests to this new method?

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -19,43 +28,18 @@ public function setUpHeadless() {
   }
 
   function testFullDetailsResponse() {
```
omarabuhussein: Since this is testing an API method , I think it should be moved to another folder (**tests/phpunit/api/v3/HRJobContractTest.php** )

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -18,6 +18,56 @@ public function setUpHeadless() {
...
+
+    $pension = civicrm_api3('HRJobPension', 'create', array(
+      'jobcontract_id' => $contract->id
+    ));
+
+    $fullDetails = civicrm_api3('HRJobContract', 'getfulldetails', array(
+      'jobcontract_id' => $contract->id
+    ));
+
+    $this->assertArrayHasKey('details', $fullDetails);
```
omarabuhussein: you can just do 

```php
$entities = ['health', 'hour', ...etc];
foreach ($entities as $entitiy ) {
  this->assertArrayHasKey($entitiy, $fullDetails);
}
```

hrjobcontract/tests/phpunit/fabricators/HRJobContractFabricator.php
```diff
@@ -0,0 +1,30 @@
...
+    ];
+
+    public static function fabricate($params, $detailsParams) {
+      if (!isset($params['contact_id'])) {
+        throw new Exception('Specify contact_id value');
+      }
+
+      $contract = civicrm_api3('HRJobContract', 'create', array_merge(self::$default, $params))['values'][0];
+
+      if (isset($detailsParams)) {
```
coldrunKacper: I'm curious why Details are included into this Fabricator instead of having its own? It looks inconsistent comparing to other entities.

AkA84: Yeah initially it was like that, but then i took a look at https://github.com/civicrm/civihr/blob/staging/hrjobcontract/tests/phpunit/HRJobContractTestTrait.php#L29 and from that it seems that HRJobDetails is an integral part of a HRJobContract, unlike things like pension, or health that are "additional"

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -113,7 +114,7 @@ public static function updateEffectiveEndDates($jobcontractId) {
    * @param int $contactID Contact ID to validate against
    * @param string $effectiveDate Date in Y-m-d format
    *
-   * @return array Array in the following format :
+   * @return array array in the following format :
```
davialexandre: It's not duplicated. The first array is the type of the param and the second one is part of the description: "Array in the following format"

omarabuhussein: I know but there is no need for it I think ..

PR #1189, opened 2016-09-28
PCHR-1579 : Remove civicrm default case/activity/relationship types.

omarabuhussein:
## requirements .
1. Remove default activity types
2. Remove default relationship types
## Solution

1- the following non-reserved civicrm default activity types were removed : 

``` php
     [
      'Medical evaluation',
      'Mental health evaluation',
      'Secure temporary housing',
      'Income and benefits stabilization',
      'Long-term housing plan',
      'ADC referral',
    ];
```

2- the following non-reserved civicrm default relationship types were removed : 

``` php
[
      'Benefits Specialist',
      'Case Coordinator',
      'Parent of',
      'Health Services Coordinator',
      'Homeless Services Coordinator',
      'Partner of',
      'Senior Services Coordinator',
      'Sibling of',
      'Spouse of',
      'Supervisor',
      'Volunteer is',
    ];
```

3- the following civicrm default case types were  removed (these were previously were marked as disabled but actually there is no need for them at all ) : 

``` php
    [
      'adult_day_care_referral',
      'housing_support'
    ];
```

All the above is done in a new upgrader (upgrade_1401) so these will be applied to existing installations , but for new installations I had to change the code to run upgraders methods during extension installation as we do in other civihr extensions.

but since was some duplicate logic inside civicrm_hook_install and some upgrader methods , I moved the logic inside civicrm_hook_install to an old upgrader so it does not run again on current installation but only for new ones . 

I also refactored and cleaned the code in hrcase.php file as well as upgrader.php class and moved most logic to upgrader class . Also the hrcase legacy business logic is not needed anymore since we no one should use activities or cases directly anymore and they are only accessible through T&A extension context .

So that make the hrcase extension purpose is to create  case types, activity types and relationship types for T&A extension use and also to remove the default ones created by civicrm in case they are not reserved . So in other words preparing the environment for T&A for HR related work.


hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -148,85 +62,214 @@ public function setComponentStatuses($components) {
...
+    foreach(self::relationshipsTypesList() as $relationship) {
+      $toInsert .= "('{$relationship['name_a_b']}','{$relationship['name_a_b']}','{$relationship['name_b_a']}','{$relationship['name_b_a']}','{$relationship['name_b_a']}','Individual','Individual',NULL,NULL,0,1),";
     }
-    if (!empty($values)) {
-      $manageSql = "INSERT INTO civicrm_managed (module, name, entity_type, entity_id, cleanup) VALUES " . implode(',', $values) ;
-      CRM_Core_DAO::executeQuery($manageSql);
+    $toInsert = rtrim($toInsert, ',');
+    $sql = "INSERT INTO `civicrm_relationship_type`
+            (`name_a_b`, `label_a_b`, `name_b_a`, `label_b_a`, `description`, `contact_type_a`, `contact_type_b`, `contact_sub_type_a`, `contact_sub_type_b`, `is_reserved`, `is_active`)
+            VALUES {$toInsert}";
```
davialexandre: why don't you use the values here directly instead of assigning them to a variable first? Also, the code can get more readable if you split the params into multiple lines instead of these long lines you have

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -148,85 +62,214 @@ public function setComponentStatuses($components) {
...
+  }
+
+  /**
+   * Remove the case types which created by CiviCRM.
+   *
+   */
+  public static function removeCivicrmCaseTypes() {
+    $toDelete = self::civicrmCaseTypesList();
+    $toDelete = "'" . implode("','", $toDelete) . "'";
+    $sql = "DELETE FROM `civicrm_case_type` WHERE name IN  ({$toDelete})";
```
davialexandre: I think this is a little bit better/clearer:
```php
$toDelete = implode("','", $toDelete);
$sql = "DELETE FROM `civicrm_case_type` WHERE name IN  ('{$toDelete}')";
```

What do you think? It applies to all the other methods on this class

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -148,85 +62,214 @@ public function setComponentStatuses($components) {
...
-    $values = array();
-    while ($caseFetch->fetch()) {
-      $values[] = "('org.civicrm.hrcase', '{$caseFetch->name}', 'CaseType', {$caseFetch->id}, 'never')";
+  public static function createRelationshipTypes() {
+    $toInsert = '';
+    foreach(self::relationshipsTypesList() as $relationship) {
+      $toInsert .= "('{$relationship['name_a_b']}','{$relationship['name_a_b']}','{$relationship['name_b_a']}','{$relationship['name_b_a']}','{$relationship['name_b_a']}','Individual','Individual',NULL,NULL,0,1),";
     }
-    if (!empty($values)) {
-      $manageSql = "INSERT INTO civicrm_managed (module, name, entity_type, entity_id, cleanup) VALUES " . implode(',', $values) ;
```
coldrunKacper: If this INSERT is going to be removed then shouldn't already inserted values be removed too? In other words - does this PR also clean the db from previously created data, relationships, etc? (to not leave any unused db entries)

omarabuhussein:
@davialexandre 
> To be honest, it was a bit hard to review all this changes in a single commit

I know and sorry for that , but I didn't imagined I will do all of this in this ticket when I started working on it and ended up doing all of that without proper commits log.

hrcase/CRM/HRCase/Upgrader.php
omarabuhussein: > why don't you use the values here directly instead of assigning them to a variable first?

I don't think  I got your point , can you explain a bit here . Do you mean I should hardcode the values ( relationship types in this case) into the query itself ?

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -148,85 +62,214 @@ public function setComponentStatuses($components) {
...
+            VALUES {$toInsert}";
+    CRM_Core_DAO::executeQuery($sql);
+  }
+
+  /**
+   * Remove a defined list of relationship types
+   *
+   */
+  public static function removeRelationshipTypes() {
+    $toDelete = array_map(function ($ar) {return $ar['name_b_a'];}, self::relationshipsTypesList());
```
omarabuhussein: I remember it is 5.4 that's why I didn't used array_column .. but thanks I didn't knew about collect method so I will use it (though I hate also to use common civicrm utility methods because who knows if one of them get removed without us noticing that :smile: )

hrcase/CRM/HRCase/Upgrader.php
omarabuhussein: this insert is inside manageCaseTypes() method which is only used in this line :+1: 

```php
    $existingCaseType = array('Appraisal');
    $this->manageCaseTypes($existingCaseType);
```

so as you see it is only add (appriasals) case type to managed entities , but that actually is not needed because the cases which defined in xml files in xml/case directory are already added to this table via civicrm_hook_caseTypes so I actually don't know what is the purpose of this method at all .

hrcase/CRM/HRCase/Upgrader.php
davialexandre: Sorry, I was talking about the $toInsert variable, but now I just realized you are inserting multiple record with a single SQL statement. In this case, I think this is ok

omarabuhussein:
this is moved to hrcore extension in this PR https://github.com/civicrm/civihr/pull/1330

PR #1191, opened 2016-09-29
PCHR-1601: Change LeaveRequest to no depend on LeavePeriodEntitlement

davialexandre:
With this change, the LeaveRequest won't be linked to the LeavePeriodEntitlement anymore. Now, in order to know which LeaveRequest was taken during a period, we just fetch the LeaveRequest between the LeavePeriodEntitlement dates (which are based on the AbsencePeriod and Contract dates). This gives us some flexibility to change the contract or period dates and "move" the LeaveRequest to another period.

Two new fields where added to the LeaveRequest entity: **type_id**, which is a FK to an AbsenceType; and **contact_id**, which is a FK to Contact and represents the use who will be taking the leave. All the methods (and related tests) used to fetch the the leave balance during an leave period were updated to use these new fields. So, basically, to get the number of leaves taken during a period we now do something like: get all leave balance changes from leave requests where the type_id is equal to the leave period entitlement type_id and contact_id is equals to the leave period entitlement contract contact_id and which the dates are in between the leave period entitlement dates.

Note: On LeaveBalanceChangeTest, I commented out a few tests about expiring records. These tests started failing after this change, but there's another ticket to change their logic, so there's no point to spend time fixing something that will change soon.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -1,5 +1,6 @@
 <?php
 
+use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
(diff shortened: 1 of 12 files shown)


coldrunKacper:
Approved.

PR #1193, opened 2016-10-04
PCHR-1602: Change LeavePeriodEntitlement to support multiple contracts for a single period

davialexandre:
An employee can have more than one contract during one Absence Period, meaning that the entitlement for a period is based on the contractual entitlement from each contract. For that reason, it doesn't make sense to have the LeavePeriodEntitlement linked to one specific contract. To fix this, the entity is now linked to the Contact, using the contact_id field, and then we can get the contracts and information related to it using both the contact_id and contract dates.

The main changes caused by this are:

- On LeavePeriodEntitlement, the getStartAndEndDates was updates to return an array with the start and end dates for all of the employee's contracts during that period.

- The getLeaveRequestBalanceForEntitlement and getBalanceForEntitlement methods (on LeaveBalanceChange) where updated to work with the new format of the return value from getStartAndEndDates, which now returns an array of days intead of a single pair of  start and end date.

- The getPeriodEntitlementForContract method was updated and renamed to fetch the same information using the Contact ID instead of the Contract ID.

- The getContactIDFromContract method was removed

- Lots of tests updated to follow the changes on the class

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
+++ b/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
@@ -473,14 +473,14 @@ class CRM_Hrjobcontract_BAO_HRJobContract extends CRM_Hrjobcontract_DAO_HRJobCon
   }
 
   /**
-   * Returns a list of active contracts.
+   * Returns a list of active contracts, together with their details.
    *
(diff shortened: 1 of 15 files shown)


nishant-bhorodia:
Approved @davialexandre !

PR #1199, opened 2016-10-06
PCHR-1560: Hotfix positioning elements in quick edit form

kkalamarski:
# Description
This issue was caused by setting one of two children of container to 100% width, which caused wrapping content, and displaying second children (form) in second line.

This was solved by using float to fit both children in one line. 
Image takes space it needs, and rest is used by the form, without wrapping.

Also added some fixes to paddings/margins to make it look good in this composition.

# Screenshots
## before:
before

### open
quick-edit-open

### closed
quick-edit-closed

### opening animation
animation


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -65,11 +65,28 @@
...
+    height: 60px;
     max-height: 60px;
+
+    .crm-summary-display_name {
+      padding-top: 22px;
+    }
+  }
+
+  .crm-summary-contactname-block {
+    display: flex;
```
AkA84: Flexbox is not fully supported by IE11 (http://caniuse.com/#feat=flexbox), so either you check that this solution works fine in it, or you change the solution so that it doesn't rely on flexbox

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -65,11 +65,28 @@
 
   .crm-summary-block-wrap {
     width: 100%;
+    margin-top: 0;
   }
 
   #crm-contactname-content {
     border: none !important;
+    height: 60px;
```
AkA84: Magic number (as in, a very specific number that seems ad-hox). What's the reason for this? Doesn't forcing the height of this block create problems with the dynamic content?

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -65,11 +65,28 @@
...
     max-height: 60px;
+
+    .crm-summary-display_name {
+      padding-top: 22px;
```
AkA84: Magic number, please justify it

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -65,11 +65,28 @@
...
+    }
+  }
+
+  .crm-summary-contactname-block {
+    display: flex;
+    flex-direction: row;
+    max-height: 120px;
```
AkA84: Magic number, please justify it. Also you are forcing a constraint on the height of the block, are we sure it's not gonna create a problem with the dynamic content?

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
kkalamarski: removed this line

PR #1203, opened 2016-10-12
PCHR-1323: Store CiviHR version

coldrunKacper:
- Creating new CiviHR extension which provides CRM_CiviHR class.
- Creating CRM_CiviHR::getVersion() static method which returns a version value from extension's info.xml file (with caching the value so xml file is fetched only once during CiviCRM session).
- Creating 'getversion' API method which returns CRM_CiviHR::getVersion() value.
- Modifying hrui extension (adding custom directory with smarty plugins).
- Creating smarty_function_civihrVersion() which gets CiviHR data by API call.
- Override core footer.tpl template to add current CiviHR version info.

A footer image after adding CiviHR version info:
civihr_version


uk.co.compucorp.civicrm.civihr/info.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<extension key="uk.co.compucorp.civicrm.civihr" type="module">
+  <file>civihr</file>
+  <name>CiviHR</name>
+  <description>Provides current CiviHR release details</description>
```
davialexandre: I think this is a bit too specific. My idea was to have a core/base civihr extension where we could put all types of generic code that might be reused by other extensions, not only release details. What do you think?

omarabuhussein: yes  I was writing the same thing

uk.co.compucorp.civicrm.civihr/info.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<extension key="uk.co.compucorp.civicrm.civihr" type="module">
```
omarabuhussein: also the extension key , I think it would be better to make it :

uk.co.compucorp.civicrm.civihrcore

and updating the extension directory name to be the same

uk.co.compucorp.civicrm.civihr/api/v3/CiviHR.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * CiviHR.getversion API specification.
+ *
+ * @param array $spec description of fields supported by this API call
+ * @return void
+ * @see http://wiki.civicrm.org/confluence/display/CRMDOC/API+Architecture+Standards
+ */
+function _civicrm_api3_civi_h_r_getversion_spec(&$spec) {
```
omarabuhussein: no need to define this spec function since I don't think we are gonna use it for this API end point.

hrui/templates/CRM/common/footer.tpl
```diff
@@ -0,0 +1,48 @@
...
+    {ts 1=$version}Powered by CiviCRM %1.{/ts}
+    {ts}CiviHR version{/ts}
+    {civihrVersion assign=civihrversion}.
+    {if !empty($footer_status_severity)}
+      <span class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
+      <a href="{crmURL p='civicrm/a/#/status'}">{$footer_status_message}</a>
+    </span>
+    {/if}
+    {ts 1='http://www.gnu.org/licenses/agpl-3.0.html'}CiviCRM is openly available under the <a href='%1'>GNU AGPL License</a>.{/ts}<br/>
+    <a href="https://civicrm.org/download">{ts}Download CiviCRM.{/ts}</a> &nbsp; &nbsp;
```
omarabuhussein: I know this out of this ticket scope , but since it is trivial , I think it is good chance to changes these urls to point to civihr.org site :smile:

hrui/CRM/Smarty/plugins/function.civihrVersion.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * {crmVersion redact=auto assign=ver}The version is {$ver}.
+ * @endcode
+ *
+ * @param $params
+ * @param $smarty
+ *
+ * @return string
+ */
+function smarty_function_civihrVersion($params, &$smarty) {
```
omarabuhussein: well .. defining this as a smarty is really nice idea  , but now I think it could fit to be moved to civihr core extension ( in case that is possible ) .. and just keep the template override in hrui .

uk.co.compucorp.civicrm.civihr/README.md
```diff
@@ -0,0 +1,20 @@
+CiviHR extension
```
omarabuhussein: so as davi stated in another place , this readme should be  more generic

uk.co.compucorp.civicrm.civihr/README.md
```diff
@@ -0,0 +1,20 @@
...
+Also it provides CiviHR entity API entry points listed below:
+* _getversion_ - returning current CiviHR version defined in info.xml file
+
+Installation
+------
+
+Go to 
+- Administer -> System Settings -> Manage Extensions
+
+and install Activity Report (uk.co.compucorp.civicrm.civihr) extension.
```
omarabuhussein: you mistakenly wrote `and install Activity Report`  but this extension is called currently "CiviHR"

uk.co.compucorp.civicrm.civihr/api/v3/CiviHR.php
omarabuhussein: I think we agreed to create it in case it is needed for specific API endpoint .. but I think it won't hurt to keep it .

hrui/CRM/Smarty/plugins/function.civihrVersion.php
coldrunKacper: I'm not sure about this one. I can move smarty function to the CiviHR Core extension but it seems to be more "ui" functionality. I really don't know, @davialexandre what do you think?

omarabuhussein: for me , it feel like a more generic  method that it should belong to a generic place like civihr core extensions . for example if other extensions wanted to use it then it make more sense to haveit defined in civihr core rather than hrui .

uk.co.compucorp.civicrm.civihrcore/CRM/Smarty/plugins/function.civihrVersion.php
```diff
@@ -48,7 +48,7 @@
  * @return string
  */
 function smarty_function_civihrVersion($params, &$smarty) {
-  $result = civicrm_api3('CiviHR', 'getversion');
+  $result = civicrm_api3('CiviHRCore', 'getversion');
```
omarabuhussein: since you are now inside the same extension , it would probably make more sense ( and faster) to call the method directly rather than using the APi

hrui/CRM/Smarty/plugins/function.civihrVersion.php
davialexandre: To be honest both options are good enough for me, but I'm slightly in favor of keeping it on hrui. A smarty function will only be used by smarty templates, so it's totally a UI thing. And, since hrui is another one of the "required" civihr extensions, I don't think it's a problem if other extensions use a function defined by it.

bin/drush-install.sh
```diff
@@ -9,6 +9,7 @@
 ##################################
 ## List of extensions defining basic entity types
 ENTITY_EXTS=\
+uk.co.compucorp.civicrm.civihrcore,\
```
AkA84: This extension doesn't really define an entity, so it shouldn't be here (not that I think this differentiation is important, but still..)

You could add another group called `CORE_EXTS` which for now will include only this new extension, then you can do

```
drush "$@" cvapi extension.install keys=$CORE_EXTS,$ENTITY_EXTS,$APP_EXTS
```

hrui/templates/CRM/common/footer.tpl
```diff
@@ -0,0 +1,48 @@
...
+ |                                                                    |
+ | You should have received a copy of the GNU Affero General Public   |
+ | License and the CiviCRM Licensing Exception along                  |
+ | with this program; if not, contact CiviCRM LLC                     |
+ | at info[AT]civicrm[DOT]org. If you have questions about the        |
+ | GNU Affero General Public License or the licensing of CiviCRM,     |
+ | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
+ +--------------------------------------------------------------------+
+*}
+{if call_user_func(array('CRM_Core_Permission','check'), 'access CiviCRM')}
```
AkA84: Since this is an override of a core template, shouldn't we add a comment block explaining what we are overriding? @omarabuhussein @davialexandre what do you think?

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
```diff
@@ -0,0 +1,21 @@
...
+
+class CRM_CiviHRCore {
+  /**
+   * Get version value from info.xml. This method uses caching so it reads
+   * xml file only once during CiviCRM session.
+   *
+   * @return string
+   */
+  public static function getVersion() {
+
```
AkA84: Since noone bothered you about this:
```php
public static function getVersion() {
  $version = CRM_Core_BAO_Cache::getItem('CiviHRCore', 'version');

  if (empty($version)) {
    // ...
  }

  return $version;
}
```
let the code breathe a little

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+class CRM_CiviHRCore {
+  /**
+   * Get version value from info.xml. This method uses caching so it reads
+   * xml file only once during CiviCRM session.
+   *
+   * @return string
+   */
+  public static function getVersion() {
```
AkA84: @davialexandre didn't you say we should avoid static methods if possible? Although this really do seem something that just needs to be static, but worth bringing this up

Is it possible to write a test for this?

uk.co.compucorp.civicrm.civihrcore/CRM/Smarty/plugins/function.civihrVersion.php
```diff
@@ -0,0 +1,56 @@
...
+ * {crmVersion redact=auto assign=ver}The version is {$ver}.
+ * @endcode
+ *
+ * @param $params
+ * @param $smarty
+ *
+ * @return string
+ */
+function smarty_function_civihrVersion($params, &$smarty) {
+  $result = CRM_CiviHRCore::getVersion();
```
AkA84: Was wondering for some time but never figured it out fully: is there a way to do this in PHP
```php
return (CRM_CiviHRCore::getVersion() || t('unknown'));
```

uk.co.compucorp.civicrm.civihrcore/civihrcore.php
```diff
@@ -0,0 +1,157 @@
...
+ * Implements hook_civicrm_alterSettingsFolders().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_alterSettingsFolders
+ */
+function civihrcore_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
+  _civihrcore_civix_civicrm_alterSettingsFolders($metaDataFolders);
+}
+
+/**
+ * Functions below this ship commented out. Uncomment as required.
```
AkA84: Can we remove this placeholder functions?

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+class CRM_CiviHRCore {
```
AkA84: I'm curious as to why this is a simple class and not a BAO class?

hrui/templates/CRM/common/footer.tpl
omarabuhussein: I agree .. also this need to be documented here https://compucorp.atlassian.net/wiki/display/PCHR/Upgrade+to+a+newer+CiviCRM+version in (Things to keep in mind when upgrading ) section

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
omarabuhussein: usually BAO classes are created if there is an entity supporting it , so for example in job contract  extension you have HRJobDetails entity which have a DAO (which mean this is an entity which have a table in the DB) and BAO  classes . so I think no need.

uk.co.compucorp.civicrm.civihrcore/CRM/Smarty/plugins/function.civihrVersion.php
davialexandre: You can use the ternary operator for that:

```
return CRM_CiviHRCore::getVersion() ?: t('unknown');
```

Note that this doesn't work very well for arrays because you would get an error if the index doesn't exist (different from Javascript where you would only get undefined). A better option is the beautiful null coalesce operator, which works both for methods and arrays, but it's only available in PHP 7: http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
davialexandre: Ideally yes, it would be better for this to be an instance method, but since it uses CRM_Core_BAO_Cache and CRM_Extension_Info static methods, you would still have problems to isolate the method and unit test it.

MAYBE we can use vfsStream (https://phpunit.de/manual/current/en/test-doubles.html#test-doubles.mocking-the-filesystem) to mock the filesystem so the method will always access a stub info.xml file and then we can assert that is returns the stub's version information.

hrui/templates/CRM/common/footer.tpl
coldrunKacper: Right, https://compucorp.atlassian.net/wiki/display/PCHR/Upgrade+to+a+newer+CiviCRM+version updated.

uk.co.compucorp.civicrm.civihrcore/CRM/CiviHRCore.php
omarabuhussein: maybe **CRM_CiviHRCore_Core** ... isn't this is the common pattern we use in other extensions

coldrunKacper: Okay so I assume that we could leave CRM/CiviHRCore namespace as it is, but move the static method into Info class (or something) - CRM_CiviHRCore_Info. Would it be fine?

PR #1205, opened 2016-10-13
FAR-24: Fix normalization of annual benefit/deduction data

AkA84:
The `annual_benefit` and `annual_deduction` fields do not have a fixed default value (something that should be discussed in the bigger scope of data structure discussion for the extension), since, if the contract is created manually, the value is `[]`, but if the contract is created via the importer, the value is `NULL`

The `FormPay` controller is in charge of always normalizing the value to an empty array (if empty), but there was a check on the `length` property on both fields that was happening before the normalization had taken place, causing the app to errors out if the default value was `null` (given that the length property would be non-existing)

Doing the normalization before anything else solves the issue


hrjobcontract/js/src/job-contract/controllers/form/form-pay.js
```diff
@@ -1 +1,167 @@
...
+                    Week: 52,
+                    Fortnight: 26,
+                    Day: 260,
+                    Hour: 2080
+                };
+
+            entityPay.is_paid = entityPay.is_paid || 0;
+            entityPay.pay_is_auto_est = '0';
+            entityPay.annual_benefits = entityPay.annual_benefits || [];
+            entityPay.annual_deductions = entityPay.annual_deductions || [];
```
AkA84: Unfortunately the whole file is diff'd since the line ending had been changed from MacOS to UNIX. Anyway the change consists in lines 27-30 having been moved from below line 35

PR #1207, opened 2016-10-17
PCHR-501: Update sample data extension

omarabuhussein:
## Requirements

1. When extension is installed - display a warning message "If you proceed all existing data will be wiped out and replaced with test data". 
2. When warning message is confirmed:
truncate all the data tables that should include test data
replace all existing data with test data
3. When extension is uninstalled - display a warning message "If you proceed all existing data will be wiped out". 
4. When warning message is confirmed:
truncate all the data tables that include test data
leave a clean installation.

## Solution

1- The old sample data extension code get removed and replaced with a new code .
2- extension get renamed to reflect our new standard naming convention for extensions.
3- ability to install the extension when creating a new buildkit instance via environment variables is added.
4- installing or uninstalling the extension will clean all the following tables first : 

```
civicrm_activity;
civicrm_activity_contact;
civicrm_address;
civicrm_case;
civicrm_case_activity;
civicrm_case_contact;
civicrm_contactaccessrights_rights;
civicrm_group_contact;
civicrm_group_organization;
civicrm_hrabsence_entitlement;
civicrm_hrjobcontract;
civicrm_hrjobcontract_details;
civicrm_hrjobcontract_health;
civicrm_hrjobcontract_hour;
civicrm_hrjobcontract_leave;
civicrm_hrjobcontract_pay;
civicrm_hrjobcontract_pension;
civicrm_hrjobcontract_revision;
civicrm_hrjobcontract_role;
civicrm_hrjobroles;
civicrm_hrvacancy;
civicrm_hrvacancy_permission;
civicrm_hrvacancy_stage;
civicrm_im;
civicrm_loc_block;
civicrm_log;
civicrm_note;
civicrm_phone;
civicrm_website;
civicrm_relationship;
```
also all custom value table will be truncated too .

for civicrm_contact and civicrm_email all the contacts and their emails will get truncated expect the "default organization" and any contact who have a corresponding drupal user.

For the following table :

```
civicrm_hrabsence_type
civicrm_hrhours_location
civicrm_hrpay_scale
civicrm_location_type
civicrm_option_value
```
only the records created by sample data extension will get deleted.

All the csv data files are placed inside *resources/csv* directory , and all photo files are placed inside *resource/photos* directory where the photos will get copied to the public upload directory while installing the extension .

Also to make the code easier to read and extend a base abstract class **CRM_CiviHRSampleData_CSVHandler** is created where each entity or a group of related entities ( such as job contracts entities [details,leave, pay. ..etc] ) should define a class inside CRM/CiviHRSampleData/Importers directory that extends this base class and should implement the abstract method operate() ).  

Also the drush-install script get modified to to accept WITH_HR_SAMPLE environment variable .. so now you can install this extension via buildkit by setting the value of WITH_HR_SAMPLE to 1 :
`export WITH_HR_SAMPLE=1`
and then run 
`civibuild create ....etc` #


uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/VacancyStage.php
```diff
@@ -0,0 +1,34 @@
...
+ * Class CRM_CiviHRSampleData_Importers_VacancyStage
+ *
+ */
+class CRM_CiviHRSampleData_Importers_VacancyStage extends CRM_CiviHRSampleData_DataImporter
+{
+
+  /**
+   * @var array To store case statuses names/values
+   */
+  private $caseStatuses =[];
```
davialexandre: Drupal standard and missing space after =

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/VacancyValue.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+class CRM_CiviHRSampleData_Importers_VacancyValue extends CRM_CiviHRSampleData_Importers_CustomFields
+{
+
+  public function __construct() {
+    parent::__construct('application_case');
+  }
+
+  protected function insertRecord(array $row) {
```
davialexandre: missing doc block

uk.co.compucorp.civicrm.civihrsampledata/README.md
```diff
@@ -0,0 +1,25 @@
+## Intro
+CiviHR Sample Data is an extensions that creates a sample data for CiviHR which
```
davialexandre: CiviHR Sample Data is an extension**~~s~~** that creates **~~a~~** sample data for CiviHR

uk.co.compucorp.civicrm.civihrsampledata/README.md
```diff
@@ -0,0 +1,25 @@
...
+gives you a better way to experience CiviHR easily.
+
+Sample data include:
+ * Contacts data with photos
+ * Vacancies
+ * Job Contracts
```
davialexandre: **The** sample data include**s**

uk.co.compucorp.civicrm.civihrsampledata/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+To install the extension when creating a new buildkit instance then add *--with-hrsample-data* option.
+
+*Example :*
+
+```bash
+civibuild create hr16 --civi-ver 4.7.9 --url http://localhost:8090 --with-hrsample-data
+```
+
+you can also enable it from the extension manager page , but please consider changing your php timeout
```
davialexandre: Start the sentence with an uppercase letter. Don't add spaces before commas

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/DataImporter.php
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Convert civicrm API response to key=>value format
+   * by specifying the key field and value field.
+   *
+   * @param array $values The API response array
+   * @param string $key The returned field name to be used as a key
+   * @param string $value The returned field name to be used as a value
+   * @return array In [`key`=>`value`, ..] format
+   */
+  private function dataToKeyValue($values, $key, $value) {
```
davialexandre: I read all the code and I still don't get what this method does. Can you please add an example of the conversion to the method description?

uk.co.compucorp.civicrm.civihrsampledata/civihrsampledata.php
```diff
@@ -0,0 +1,116 @@
+<?php
```
omarabuhussein: well we discussed that when you created leave and absence extension and we agreed to use :

**uk.co.compucorp.civicrm.ExtensionName** 

for extension names ( so I think it is fair to assume it is our new standard naming convention ) .

 Also this extension is  never used before so it is like a new extension so It is better to rename it to follow the other new ones.

also for the last part .. I think it reflect the extension name and it does not matter if it start with hr or not .. for example we have uk.co.compucorp.civicrm.appraisals which do not have the hr part ... I think the important here is to use uk.co.compucorp.civicrm. prefix and the last part can depend on the extension name .

bin/drush-install.sh
```diff
@@ -68,11 +68,11 @@ function set_resource_urls() {
 ##################################
 ## Main
 
-# Get CiviCRm Path and shift to the next option
+# Get CiviCRM Path and shift to the next option
 CIVI_PATH=$1
 shift
 
-if [ "$1" == "--with-sample-data" ]; then
+if [ "$1" == "--with-hrsample-data" ]; then
```
omarabuhussein: well .. currently when you create a new buildkit instance there is no option to not allow it to create the sample data that is shipped with civicrm core .. so I am thinking to add another option to the buildkit called (--with-sample-data) to either allow civicrm to create its sample data or not .. and --with-hrsample-data will be reserved for civihr sample data.

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/DataImporter.php
```diff
@@ -0,0 +1,134 @@
...
+  protected function setDataMapping($mappingKey, $oldValue, $newValue) {
+    self::$data[$mappingKey][$oldValue] = $newValue;
+  }
+
+  /**
+   * Get stored mapping for specific mapping key and value.
+   *
+   * @param string $mappingKey
+   * @param int|string $oldValue
+   * @return array
```
omarabuhussein: oh right .. thanks for spotting this

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/DataImporter.php
omarabuhussein: suppose you have this api response 

```php
'values' => 
0=> ['id' => 1, 'name' => 'omar' , 'country' => 'jordan', 'gender' => 'male' ],
1=> ['id' => 2, 'name' => 'davi' , 'country' => 'brazil', 'gender' => 'male' ],
```
and you just want to get the response to be like 
```php
['omar' => 1, 'davi' => 2 ]  
```
or maybe
```php
['jordan' => 'omar' , 'brazil' => 'davi']
```
then you can pass the API response to this method along with the field you want to use as a key and the field you want to use as a value .

so  in example 1 you can pass dataToKeyValue($response, 'name', 'id') and  in example 2 you can pass dataToKeyValue($response, 'country', 'name') to get the required response .. this is important for mapping so for example in the csv files you have absence types or relationship types stored by their names or labels .. and you want to convert them to IDs before using them in the API so you need to return the relationship types or absence types in ['name' => 'id', ...] format to be able to convert these types to their IDs quickly .

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/AbsenceEntitlement.php
```diff
@@ -0,0 +1,80 @@
...
+    $row['type_id'] = $this->absenceTypes[$row['type_id']];
+
+    // convert absence period name to ID
+    $row['period_id'] = $this->absencePeriods[$row['period_id']];
+
+    civicrm_api3('HRAbsenceEntitlement', 'create', $row);
+  }
+
+  /**
+   * Fetch absence types and cache them in ['name' => 'id' ...] format.
```
omarabuhussein: :smile:

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/AbsenceEntitlement.php
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * @throws CiviCRM_API3_Exception
+   */
+  private function fetchAbsenceTypes() {
+    $absTypes = civicrm_api3('HRAbsenceType', 'get', array(
```
omarabuhussein: no it does not :smile:  I will fix that .

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/AbsencePeriod.php
```diff
@@ -0,0 +1,31 @@
...
+   */
+  protected function insertRecord(array $row) {
+    $this->callAPI('HRAbsencePeriod', 'create', $row);
+  }
+
+  /**
+   * Remove any existing absence period.
+   */
+  private function removeAllPeriods() {
+    $query = "DELETE FROM civicrm_hrabsence_period";
```
omarabuhussein: the API for delete only work for single record but I need to remove any defined absence period .. also I think it is more efficient  than fetching all absence periods via the API and then loop through the results and delete them one by one  . ( but anyway there is probably only single absence period on fresh installations so I think that might work )

uk.co.compucorp.civicrm.civihrsampledata/civihrsampledata.php
davialexandre: Yeah, we discussed about the uk.co.compucorp.civicrm prefix, but I don't remember anything about the extension name. 
I know there are extensions not following any pattern, but, as I mentioned in a comment on https://github.com/civicrm/civihr/pull/1203, most of our extensions names start with hr (including L&A which is uk.co.compucorp.civicrm.hrleaveandabsences). Even the core extension which just got merged follow this (uk.co.compucorp.civicrm.hrcore). So, unless we decide to use one specific convention, I believe the best thing to do is to follow what we've been doing for most of the extensions.

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/ActivityTest.php
```diff
@@ -0,0 +1,95 @@
...
+    ];
+
+    $mapping = [
+      ['activity_mapping', $sourceRecordID],
+      ['contact_mapping', $this->testContact['id']],
+    ];
+
+    $this->runImporter('CRM_CiviHRSampleData_Importer_Activity', $this->rows, $mapping);
+
+    $this->assertNotEmpty($this->apiQuickGet('Activity','source_record_id', $sourceRecordID));
```
davialexandre: This is a very, very, very poor test. If apiQuickGet simply returns` [true]` it will make the test pass but the code will be completely broken.

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/JobContractTest.php
```diff
@@ -0,0 +1,168 @@
...
+      ['contact_mapping', $this->testContact['id']],
+    ];
+
+    $this->runImporter('CRM_CiviHRSampleData_Importer_JobContract', $this->rows, $mapping);
+
+
+    $this->assertNotEmpty($contract = $this->apiQuickGet('HRJobContract',null, null, ['contact_id' => $this->testContact['id']]));
+
+    $contractID = $contract['id'];
+    $this->assertNotEmpty($revision = $this->apiQuickGet('HRJobContractRevision',null ,null, ['jobcontract_id' => $contractID]));
```
davialexandre: Please don't do this. Do the assignment first, outside the assertion, and then assert on the assigned value.

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/JobRolesTest.php
```diff
@@ -0,0 +1,99 @@
...
+      ->install('com.civicrm.hrjobroles')
+      ->apply();
+  }
+
+  public function setUp() {
+    $this->rows = [];
+    $this->rows[] = $this->importHeadersFixture();
+
+    $this->testContact = ContactFabricator::fabricate(['first_name' => 'chrollo', 'last_name' => 'lucilfer']);
+    $this->testJobContract = JobContractFabricator::fabricate(['contact_id' => $this->testContact['id']], ['period_start_date' => '2015-01-01']);
```
davialexandre: can you please split these into multiple lines?

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/LocationTypeTest.php
```diff
@@ -0,0 +1,48 @@
...
+      'Correspondence',
+      'Correspondence',
+      'Postal address',
+      0,
+      0,
+      0
+    ];
+
+    $this->runImporter('CRM_CiviHRSampleData_Importer_LocationType', $this->rows);
+
```
davialexandre: why so many blank lines?

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/VacancyStageTest.php
```diff
@@ -0,0 +1,56 @@
...
+      ->install('org.civicrm.hrjobcontract')
+      ->install('org.civicrm.hrrecruitment')
+      ->apply();
+  }
+
+  public function setUp() {
+    $this->rows = [];
+    $this->rows[] = $this->importHeadersFixture();
+
+    $vacancy = civicrm_api3('HRVacancy', 'create', [
```
davialexandre: Please create a fabricator for this

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Case.php
```diff
@@ -0,0 +1,58 @@
...
+    $result = civicrm_api3(
+      'Case',
+      'create',
+      $params
+    );
+
+    return array_shift($result['values']);
+  }
+
+  public static function fabricateCaseType($params = []) {
```
davialexandre: Since there are two separate entities, I think it would be better to have two separate fabricators. Is there any specific reason for having both at a single fabricator?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/OptionValue.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+class CRM_HRCore_Test_Fabricator_OptionValue {
+
+  private static $defaultParams = [
+    'name' => 'test option',
+    'sequential' => 1
+  ];
+
+  public static function fabricate($groupName, $params = []) {
```
davialexandre: It would be better if to make this method consistent with the other fabricators which only receive the $params array (I know HRJobContract has two params, but both are arrays to be passed to the api)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Relationship.php
```diff
@@ -0,0 +1,24 @@
...
+    'name_b_a' => "test BA",
+    'contact_type_a' => "Individual",
+    'contact_type_b' => "Individual",
+  ];
+
+  public static function fabricateRelationshipType($params = []) {
+    $params = array_merge(self::$defaultParams, $params);
+
+    $result = civicrm_api3(
+      'Activity',
```
davialexandre: Is this correct? If it's not, it looks like your tests didn't catch this and we definitely need more tests (specially negative tests) to the importers

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Relationship.php
```diff
@@ -0,0 +1,24 @@
...
+
+  private static $defaultParams = [
+    'sequential' => 1,
+    'name_a_b' => "test AB",
+    'name_b_a' => "test BA",
+    'contact_type_a' => "Individual",
+    'contact_type_b' => "Individual",
+  ];
+
+  public static function fabricateRelationshipType($params = []) {
```
davialexandre: Relationship and RelationshipType are two different entities, so I think it's better to have two separate fabricators, unless you have a good reason to have this fabricateRelationshipType method in this fabricator

omarabuhussein:
> Overall the changes look good. The tests need some more work though, as they're won't catch a lot of possible errors

I know but I had 21 class to write tests for , not to mention this ticket took to long and I just wanted to finish it that's why only basic tests are written .

davialexandre:
> I know but I had 21 class to write tests for , not to mention this ticket took to long and I just wanted to finish it that's why only basic tests are written.

In that case, please confirm with Mike what is the priority here and if ok to finish this later. Unfortunately, from a technical point of view, I cannot consider this a complete and the standard is that we must have tests for every PR

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/CustomFields.php
```diff
@@ -1,24 +1,31 @@
...
 
   public function __construct($customGroupName) {
     $this->columns = $this->getFixData('CustomField', 'name', 'id', [
       'custom_group_id' => $customGroupName,
     ]);
   }
 
+  /**
+   * {@inheritdoc}
+   * Generic method to handle custom fields import
```
omarabuhussein: but {@inheritdoc} just inherit the parent description

uk.co.compucorp.civicrm.civihrsampledata/civihrsampledata.php
```diff
@@ -130,7 +130,9 @@ function _civihrsampledata_importSampleData($extensionDirectory) {
     $fileToImport = new SplFileObject("{$csvDir}/{$file}.csv");
 
     $importerClassName = "CRM_CiviHRSampleData_Importer_{$class}";
-    $importer = new $importerClassName($fileToImport);
+    $importer = new $importerClassName();
+    $importer->setSplFileObject($fileToImport);
```
omarabuhussein: passing the dependency via the constructor is what I tried to do at first , but doing this will require repeating passing the dependency for all child classes that implements its own constructor which I think is not Ideal .. so doing it via setter was a cleaner solution to this in my opinion .

AkA84:
@davialexandre @omarabuhussein what about something in the middle: if the basic tests are written (meaning: tests that make sure that the basic functioning is done), then the PR can be approved. But if the ticket will be reopened then more tests will have to be added to cover the issues that will have been reported.

What do you think?

davialexandre:
@AkA84 @omarabuhussein sounds good to me

omarabuhussein:
@AkA84 @davialexandre 

It usually better to try to cover as much as you can from the beginning , but  in case of big tickets like this where a lot of stuff need to be tested I agree with that .

uk.co.compucorp.civicrm.civihrsampledata/tests/phpunit/CRM/CiviHRSampleData/Importer/AbsencePeriodTest.php
```diff
@@ -0,0 +1,47 @@
...
+    $this->rows[] = [
+      '2016',
+      '2016 (Jan 1 to Dec 31)',
+      '2016-01-01 02:00:00',
+      '2017-01-01 01:59:59',
+    ];
+
+    $this->runImporter('CRM_CiviHRSampleData_Importer_AbsencePeriod', $this->rows);
+
+    $this->assertEquals('2016', $this->apiQuickGet('HRAbsencePeriod','name', '2016'));
```
omarabuhussein: for 21 class , and suppose at least I am going to write at least 5 extra test cases =>  5*21 = at least 105 test case .. hell this will take the eternity :smile: ... also wouldn't you think this is not testing what the code actually do .. at the end there is a BAO methods that have this kind of tests  ( eg. wrong dates parsing)... my code is just in someway a wrapper for the API ,, I know it is not an exact wrapper and there are some mapping logic happen but at the end no insertion happen to the db without the API so I think it is safe to assume that BAO tests are sufficient for this kind of stuff , what do you think ?

omarabuhussein: but I agree to (what about negative tests?) & ( What if the row doesn't have the right number of columns? )  ... my comment above apply to your questions about (what if the data is invalid? Also, you should at least assert the one fields. Without it, how do know, for example, if the dates were parsed correctly?)

davialexandre: Sorry, there is a typo on my first comment. It should say "Also, you should at least assert the other fields".

For example, in that test you're asserting only that you got the expected name. But you should, test that you have the expected values for all the other fields. This will give you some assurance that, at least for this minimal set of data, the parsing/validation is right.

Here is an example of a breaking change that this test will not catch: Currently, you're using the API to import the data. Your test will pass because the API automatically handles the date format conversion and will, in the end, pass it to the BAO in YmdHis format. Now, imagine that, for some reason, you have to stop using the API and call the BAO directly. It will not receive the date in the correct format and maybe it might truncate it and save the wrong information on the database base, but your test will still pass because it's only checking the name.

Yeah, I know it's quite a lot of tests but it's part of the job. We already found a few bugs just by reviewing the code. It wouldn't be a surprise to find a few more when we actually start using this. Tests can help us find these faster and also fix them faster when they appear (yeah, we'll have bugs even if we do a lot of tests). So time should no be an excuse to not write tests. Also, when something it's hard to test it might be a sign that the code itself can be improved to be more flexible and easier to test (can't say if this is the case here though)

Anyway, I agreed we should meet at the middle for this specific case. So I think it will be good enough if you add assertion to all the properties instead of checking only for one of them.

uk.co.compucorp.civicrm.civihrsampledata/civihrsampledata.php
davialexandre: I see. To be honest I don't a see the problem with calling `parent::__construct($splFileObject)`, but the setter is also good enough as long as you don't forget to update the import() method to check that the $fileHandler is not null

omarabuhussein:
@davialexandre please check my last two commits , I did some improvements to some tests .

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/AbsencePeriod.php
```diff
@@ -1,31 +1,40 @@
...
+      'sequential' => 1,
+      'return' => ["id"],
+      'options' => ['limit' => 0],
+    ]);
+
+    foreach($absencePeriods['values'] as $absencePeriod) {
+      $this->callAPI('HRAbsencePeriod', 'delete', [
+        'id' => $absencePeriod['id'],
+      ]);
+    }
```
omarabuhussein: yup because there will not be a lot of periods ( don't think more than one ) , so  I decided it is better to stick with the API.

uk.co.compucorp.civicrm.civihrsampledata/CRM/CiviHRSampleData/Importers/CustomFields.php
omarabuhussein: ops , I should start pay more attention when reading documents :smile:  .

uk.co.compucorp.civicrm.civihrsampledata/civihrsampledata.php
omarabuhussein: I end up adding a check on the $fileHandler , I just don'k like the idea of  repeat passing the dependencies for all these classes via constructor .

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Relationship.php
omarabuhussein: nope , it is just a tests that I forget to run since I was running them one by one ... thanks for your note about installing all extensions in the base class headless method I now can just run all the tests at once so no chance to forget anything.

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/OptionValueTest.php
```diff
@@ -32,10 +30,15 @@ public function testImport() {
...
-    $this->assertEquals(
-      'Compassionate_Leave',
-      $this->apiGet('OptionValue','name', 'Compassionate_Leave', ['option_group_id' => 'activity_type'])
+    $optionValue = $this->apiGet(
+      'OptionValue',
+      ['option_group_id' => 'activity_type', 'name' => 'Compassionate_Leave']
     );
+
+    foreach($this->rows[0] as $index => $fieldName) {
+      if (!in_array($fieldName, ['option_group_id', 'component_id']))
```
davialexandre: always use curly braces, even if the if/else has only one line. Also, the indentation seems wrong here

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/RelationshipsTest.php
```diff
@@ -47,7 +45,12 @@ public function testImportWithoutCase() {
 
     $this->runImporter('CRM_HRSampleData_Importer_Relationships', $this->rows, $mapping);
 
-    $this->assertEquals($this->testContactA['id'], $this->apiGet('Relationship', 'contact_id_a', $this->testContactA['id']));
+    $relationship = $this->apiGet('Relationship', ['contact_id_a' => $this->testContactA['id']]);
+
+    foreach($this->rows[0] as $index => $fieldName) {
+      if (!in_array($fieldName, ['relationship_type_id', 'case_id']))
```
davialexandre: same thing here: curly braces and indentation

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/RelationshipsTest.php
```diff
@@ -71,7 +74,12 @@ public function testImportWithCase() {
 
     $this->runImporter('CRM_HRSampleData_Importer_Relationships', $this->rows, $mapping);
 
-    $this->assertEquals($this->testContactA['id'], $this->apiGet('Relationship', 'contact_id_a', $this->testContactA['id']));
+    $relationship = $this->apiGet('Relationship', ['contact_id_a' => $this->testContactA['id']]);
+
+    foreach($this->rows[0] as $index => $fieldName) {
+      if (!in_array($fieldName, ['relationship_type_id', 'start_date']))
```
davialexandre: and here too

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/OptionValueTest.php
omarabuhussein: I actually like using curly places .. I just forget to do that .

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/DataImporter.php
```diff
@@ -56,13 +40,17 @@ public function getDataMapping($mappingKey, $oldValue) {
   /**
    * Imports csv file into the database.
    *
+   * @param SplFileObject $fileHandler
    */
-  public function import() {
+  public function import(SplFileObject $fileHandler = null) {
```
davialexandre: Why the default null? Is there a case where you should not pass a fileHandler?

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/DataImporter.php
```diff
@@ -56,13 +40,17 @@ public function getDataMapping($mappingKey, $oldValue) {
...
     $header = null;
 
-    if (!empty($this->fileHandler)) {
-      while (!$this->fileHandler->eof()) {
-        $row = $this->fileHandler->fgetcsv();
+    if (empty($fileHandler)) {
```
davialexandre: if you remove the default null value you'll also be able to remove this if, because the type declaration on the method signature will not allow null values to be passed

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/DataImporter.php
omarabuhussein: true .. thanks for the note

davialexandre:
Oh, one thing I only noticed now. You forgot to put the ticket number on 3b11710 commit message. Please fix that too

hrabsence/hrabsence.php
```diff
@@ -587,3 +589,16 @@ function hrabsence_civicrm_pageRun($page) {
       ->addStyleFile('org.civicrm.hrabsence', 'css/hrabsence.css');
   }
 }
+
+/**
+ * Remove any defined relationship type by this extension
+ *
+ */
+function _hrabsence_removeRelationshipTypes() {
```
davialexandre: Wouldn't it be better to use managed entities for this? Is it even possible possible?

davialexandre: It's fine the way you did it. I was just curious about if it was possible

davialexandre:
Approved

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/CSVHandler.php
```diff
@@ -86,6 +91,24 @@ protected function callAPI($entity, $action, $params) {
   }
 
   /**
+   * Deletes entity record from the database based on
+   * search criteria
+   *
+   * @param string $entity
+   * @param array $searchParams
+   */
+  protected function deleteRecord($entity, $searchParams) {
```
davialexandre: I liked the idea of trying to make this a more generic thing and renaming the class to CSVHandler and the import to iterate, but things got a little confusing when, for example, the CSVHandler has methods like `deleteRecord()` (which is specific for cleaners) and `callAPI()`.

I think the [Visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) would be a perfect fit for this as it solves exactly the problem you tackling here. I created a small gist to show how it could be implemented here: https://gist.github.com/davialexandre/6df1ac26450563cf59ac294a9bc5987e

And this is a diagram of how the classes would work together:
captura de tela 2016-10-31 as 11 31 29

I just don't know how things are in terms of time for this ticket. If it's something really urgent, then it's ok to leave it as it is now, otherwise It would definitely be worth to improve this and separate the responsibilities in a better way

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Data such as option values, pay scales .. etc
+   * which are created by this extension are cleaned here.
+   *
+   */
+  private function cleanSampleDataValues() {
+    $filesToClean = [
+      'AbsenceType' => 'civicrm_hrabsence_type',
```
omarabuhussein: yes we don't want to remove default values .. only the ones created by the sample data extension

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Cleaner/HRHoursLocationTest.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+require_once EXTENSION_ROOT_DIR . 'CRM/HRSampleData/Importer/HRHoursLocation.php';
```
omarabuhussein: yes ..  I wonder why the test didn't fail , I will check that again.

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/CSVHandler.php
```diff
@@ -86,6 +91,24 @@ protected function callAPI($entity, $action, $params) {
...
+    $searchParams['options'] = ['limit' => 0];
+
+    $entityResult = $this->callAPI($entity, 'get', $searchParams);
+
+    if (!empty($entityResult['id'])) {
```
omarabuhussein: nope actually 'get' action will return the ID if only one result retrieved

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/CSVHandler.php
```diff
@@ -86,6 +91,24 @@ protected function callAPI($entity, $action, $params) {
...
 
   /**
+   * Deletes entity record from the database based on
+   * search criteria
+   *
+   * @param string $entity
+   * @param array $searchParams
+   */
+  protected function deleteRecord($entity, $searchParams) {
+    $searchParams['options'] = ['limit' => 0];
```
omarabuhussein: from my quick testing I found that you cannot delete more than one record with the API at time , so I guess it make sense to keep the limit = 0

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/CSVHandler.php
omarabuhussein: thanks so much for the detailed explanation  , I thought the template pattern the I've used is the perfect fit for this kind or problems but apparently it appear that the visitor pattern play more nicely here , let me see how could I play it out.

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Cleaner/HRHoursLocationTest.php
omarabuhussein: mmmm strange .. I couldn't find any clue  .. but apparently I can remove all require_once and all tests works fine ( which is really really weird ) .

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/JobContractTest.php
```diff
@@ -44,15 +49,15 @@ public function testIterate() {
       'Subject Head - Literacy',
       'Subject Head - Literacy',
       'Donor funded position',
-      'test option',
+      $this->optionVales['hrjc_contract_type']['name'],
```
davialexandre: Typo here. s/optionVales/optionValues/g

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Importer/JobRolesTest.php
```diff
@@ -17,6 +15,8 @@ class CRM_HRSampleData_Importer_JobRolesTest extends CRM_HRSampleData_BaseImport
 
   private $testJobContract;
 
+  private $optionVales = [];
```
davialexandre: typo here too

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Cleaner/PhotoFilesCleaner.php
```diff
@@ -0,0 +1,33 @@
...
+  protected function operate(array $row) {
+    if (!empty($row['image_URL'])) {
+      $this->deletePhoto($row['image_URL']);
+    }
+  }
+
+  /**
+   * Deletes contact photo
+   *
+   * @param $photoName
```
davialexandre: Missing param type

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Importer/AbsencePeriod.php
```diff
@@ -13,7 +13,14 @@ public function __construct() {
   /**
    * {@inheritdoc}
    */
-  protected function operate(array $row) {
+  public function visit(array $row) {
+    $this->importRecord($row);
```
davialexandre: I see that for all the Importers, the implementation of the visit() method is exactly the same. In that case, I think it would be better to move it CRM_HRSampleData_CSVImporterVisitor and avoid the duplication.

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Cleaner/OptionValue.php
```diff
@@ -13,10 +13,31 @@ class CRM_HRSampleData_Cleaner_OptionValue extends CRM_HRSampleData_CSVCleanerVi
...
     $this->deleteRecord(
       'OptionValue',
       ['name' => $row['name'], 'option_group_id' => $row['option_group_id']]
     );
   }
 
+  /**
+   * Gets option group name by its title
+   *
+   * @param $groupTitle
```
davialexandre: missing data type

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Cleaner/OptionValue.php
```diff
@@ -13,10 +13,31 @@ class CRM_HRSampleData_Cleaner_OptionValue extends CRM_HRSampleData_CSVCleanerVi
...
+   */
+  private function getOptionGroupNameByTitle($groupTitle) {
+    $optionGroup = $this->callAPI( 'OptionGroup', 'get', [
+      'sequential' => 1,
+      'return' => ["name"],
```
davialexandre: Please use single quotes for name, just to keep it consistent with the other strings

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Importer/OptionValue.php
```diff
@@ -23,4 +27,20 @@ protected function importRecord(array $row) {
     }
   }
 
+  /**
+   * Gets option group name by its title
+   *
+   * @param $groupTitle
+   */
+  private function getOptionGroupNameByTitle($groupTitle) {
```
davialexandre: My comments on the cleaner apply here as well

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+    }
+  }
+
+  /**
+   * Updates UFMatch attached contact given the record
+   * ID and the new contact email.
+   *
+   * @param $ufmatchID
+   * @param $newContactEmail
+   *
```
davialexandre: empty line

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+      }
+
+    }
+  }
+
+  /**
+   * Updates UFMatch attached contact given the record
+   * ID and the new contact email.
+   *
+   * @param $ufmatchID
```
davialexandre: missing data type on both params

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+   * @param $newContactEmail
+   *
+   */
+  private function updateUFMatchContact($ufmatchID, $newContactEmail) {
+    $newContact = civicrm_api3('Email', 'get', [
+      'sequential' => 1,
+      'return' => ["email", "contact_id"],
```
davialexandre: please use single quote for the array items

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+        'domain_id' => $currentDomainID,
+      ]);
+    }
+  }
+
+  /**
+   * Updates user email on the CMS side.
+   * (Currently Supports Drupal 7 CMS Only)
+   *
+   * @param $cmsID
```
davialexandre: missing data types

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+    $existingUser = user_load($cmsID);
+
+    // save new user email
+    user_save($existingUser, ['mail' => $newEmail]);
+  }
+
+  /**
+   * Deletes contact by ID
+   *
+   * @param $id
```
davialexandre: data type

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -129,4 +128,98 @@ private function copyContactPhotos() {
...
+    ]);
+  }
+
+  /**
+   * Gets current domain ID
+   */
+  private function getCurrentDomainID() {
+    $domain = civicrm_api3('Domain', 'get', [
+      'sequential' => 1,
+      'return' => ["id"],
```
davialexandre: single quotes

uk.co.compucorp.civicrm.hrsampledata/resources/csv/civicrm_activity.csv
```diff
@@ -1,46 +1,6 @@
 id,source_record_id,activity_type_id,subject,activity_date_time,duration,details,status_id,priority_id,source_contact_id,assignee_id,target_id
-1,,Public Holiday,New Year's Day,2016-01-01 00:00:00,,,Scheduled,Normal,229,,
```
davialexandre: Why are you removing all this?

Actually, it would be nice to add more details to the commit message explaining what and why are you adding these changes

omarabuhussein: Some sample contacts are required to be removed according to the QA .. so I removed them and their related records

PR #1208, opened 2016-10-18
PCHR-1603: Make EntitlementCalculation deal with multiple contracts for a single period

davialexandre:
It is possible for an employee to have multiple contracts during a single Absence Period. For example:
Absence Period from 2016-01-01 to 2016-12-31
Contract 1 from 2016-01-01 to 2016-03-31
Contract 2 from 2016-05-01 to 2016-10-01
Contract 3 from 2016-10-02 to null (when there is no end date, we use the period end date)

The EntitlementCalculation was updated to be able to deal with such a scenario. Heres is what changed:

### 1. Instantiate an EntitlementCalculation with a Contact instead of a Contract
Instead of passing a single contract to the class, we now pass a contact and then all the calculation logic will be based on all active contracts of that contact during the absence period. Lots of tests had to be updated to reflect this.

### 2. Remove getContractualEntitlement
This method was removed because there's no such information in a multiple contract scenario. You can't just sum up the contracts entitlements to get that, because you might end up with a huge number that doesn't reflect an actual number. On the Entitlement Calculation page, the column which displayed this information was also removed.

### 3. Change how Pro Rata is calculated
For multiple contracts, the Pro Rata is basically the sum of pro ratas for each of the contracts. The rounding happens only at the end, after summing up the numbers for each contract.
Another change is that Public Holidays are not pro rated anymore. Even if the contract says public holidays should be added to the entitlement, we only pro rate the contractual entitlement and then the number of public holidays will be added to the proposed entitlement.

### 4. Add Public Holidays to the proposed entitlement
This is a continuation of the change made in item 3. Since now the public holidays are not pro rated anymore, we added them on the proposed entitlement. So, instead of Pro Rata + Brought Forward, it's now Pro Rata + Brought Forward + Public Holidays.

### 5. Removal of the __toString()
With this multiple contract scenario, it's not simple to represent the calculation as a string anymore. There are a lot more details involved now. For that reason, I changed the method to just return an empty string and removed the related tests. This will be fixed in another PR where a new way to display de calculation description will be introduced.


omarabuhussein:
> he calculation logic will be based on all active contracts

but any contact cannot have more than one active contract .. I think you mean "will be based on contact contracts during the absence period"

omarabuhussein:
> 2. Remove getContractualEntitlement
This method was removed because there's no such information in a multiple contract scenario. You can't just sum up the contracts entitlements to get that, because you might end up with a huge number that doesn't reflect an actual number. On the Entitlement Calculation page, the column which displayed this information was also removed.

I don't have enough context to discuss this so ignore it in case what I am saying does not make sense ... but getContractualEntitlement could be used to show the current Contractual Entitlement for the **current** contract. which as we know will be only one contract since no user is allowed to have more than one active contract.

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -16,11 +16,11 @@ define([
         '$scope', '$route', '$filter', '$uibModal', '$rootElement', '$q', '$window', 'settings', 'API',
```
davialexandre: nice catch! I always forget that

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractTest.php
```diff
@@ -170,6 +170,225 @@ public function testGetActiveContractsWithDetailsCanReturnOnlyContractsOfASpecif
     $this->assertNull($contact2ActiveContracts[0]['period_end_date']);
   }
 
+  public function testGetActiveContractWithDetailsShouldReturnTheDetailsFromTheLatestDetailsRevision() {
+    $this->createContacts(1);
```
davialexandre: Yes, but this branch doesn't have the recent updates on staging yet. Once I update it (not on this pr) I switch things to use Fabricators

omarabuhussein: cool not proplem

hrjobcontract/js/src/job-contract/controllers/contract.js
omarabuhussein: me too :(

davialexandre:
> I don't have enough context to discuss this so ignore it in case what I am saying does not make sense ... but getContractualEntitlement could be used to show the current Contractual Entitlement for the current contract. which as we know will be only one contract since no user is allowed to have more than one active contract.

This doesn't work. The contractual entitlement was displayed on the entitlement calculation screen and it was supposed to represent the entitlement for the whole period, so it cannot be based only on the current contract

davialexandre:
@omarabuhussein I've implemented your suggestions. Please take a second look

omarabuhussein:
@davialexandre you forget to updated the minifed js  files

davialexandre:
Nope, I didn't: https://github.com/civicrm/civihr/pull/1208/commits/8cf9e1af6cecda2c1a3ad77085285f2cae944843

omarabuhussein:
Approved @davialexandre

PR #1210, opened 2016-10-19
PCHR-1353: Update styleguide

AkA84:


totten:
I'm using `org.civicrm.bootstrapcivicrm` from branch `PCHR-1353-update-styleguide`, and there's an extra icon in `crm-ui-datepicker` widgets (screenshot attached). 

screen shot 2016-12-13 at 3 01 51 pm

Not sure if the issue is particular to this branch, but it affects more than Mosaico -- i.e. it also happens with plain CiviMail (which hasn't opted into Bootstrap CSS). It only happens when `bootstrapcivicrm` is active. (Noted here after discussion on Mattermost.)

colemanw:
@AkA84 I'm depending on the changes in this PR for CiviCase work. Will you please merge it soon?

totten:
We're in the same situation with Mosaico. At the end of the call last week, Alessandro and I negotiated a bit on how to make this  easier to install. We leaned toward "Option B", which I'm supposed to make an attempt at this week (while Aka84 is on break). Overall:

 * Update this branch to include other changes from `civihr`'s `master` branch. Resolve merge-conflicts.
 * Combine `org.civicrm.bootstrapcivicrm` and `org.civicrm.bootstrap` into one extension. Change paths to remove `..` business. Publish this extension at a new repo under `https://github.com/civicrm/`. 

This arrangement as a few consequences:

 * It will be compatible with release processes supported by `civicrm.org`'s extension directory (make a tag to mark alphas and stables).
 * Less confusion about `org.civicrm.bootstrap` vs `org.civicrm.bootstrapcivicrm`.
 * CiviHR would then need a different PR to pull in the new repo, but that would be handled by someone on their team, and the timing doesn't need to be coordinated down-to-the-minute. (More like "within a week or so.") Additionally, if there's been concurrent updates to `civihr/master` during a transitional period, they would need to be ported over.
 * If someone wants to make a separate extension that builds on the SCSS from here, they can still do that, e.g.
    * Make a new extension which depends on this extension.
    * In the `gulpfile`, use [`cvRes.getPath`](https://github.com/civicrm/cv-nodejs) and [`includePaths`](http://stackoverflow.com/a/36022951/4195300) to locate any assets that you need.

guanhuan:
Can we prioritise this @AkA84 ?

totten:
Replaced by #1375.

PR #1211, opened 2016-10-19
PCHR-1409 - Export Job Contract end_reason field.

coldrunKacper:
The 'end_reason' field of Details entity was being exported as a number instead of label.

I've created CRM_Hrjobcontract_Export_Export class which keeps writeCSVFromTable() method basing on original CRM_Export_BAO_Export one but adding our Export/Import converter mechanism inside.

Then I've used hook_civicrm_export to use that custom method instead of original one. It allows to change raw values with our Export/Import converter and export them with proper format.
Also I've removed "Replace contract type option values with respective labels" change as it's no longer needed because the same thing is now done with Export/Import converter.


hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+class CRM_Hrjobcontract_Export_Export {
```
omarabuhussein: yeah probably you could change it to CRM_Hrjobcontract_Export_Converter or maybe CRM_Hrjobcontract_Export_Main

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
...
+   * @param $sqlColumns
+   * @param $exportMode
+   * @param null $saveFile
+   * @param string $batchItems
+   */
+  public static function writeCSVFromTable($exportTempTable, $headerRows, $sqlColumns, $exportMode, $saveFile = NULL, $batchItems = '') {
+    $writeHeader = TRUE;
+    $offset = 0;
+    $limit = CRM_Export_BAO_Export::EXPORT_ROW_COUNT;
+    $ei = CRM_Hrjobcontract_ExportImportValuesConverter::singleton();
```
omarabuhussein: can you please change the variable name to something make more sense .

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
...
+
+    $query = "SELECT * FROM $exportTempTable";
+
+    while (1) {
```
omarabuhussein: I know this is based on civicrm class , but civicrm implementation looks a bit weird to me and I think we are not compelled to follow what their class do  .. so  what if for some reason the break; statement never reached ? well , we will have a nice infinite loop ... I think it would be better to query the database to get the full rows count with :
$totalCount = executeQuery(SELECT COUNT(*) FROM $exportTempTable");

and then Then :

$loopCount = $totalCount/limit; 

and finally you can use :

while($loopCount) {
// logic here 
}

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
...
+      $limitQuery = $query . " LIMIT $offset, $limit ";
+      $dao = CRM_Core_DAO::executeQuery($limitQuery);
+
+      if ($dao->N <= 0) {
```
omarabuhussein: so you could remove this if what above is implemented

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
...
+        break;
+      }
+
+      $componentDetails = array();
+      while ($dao->fetch()) {
+        $row = array();
+
+        foreach ($sqlColumns as $column => $dontCare) {
+          $row[$column] = $dao->$column;
+          // Convert Job Contract entity values using to their proper export
```
omarabuhussein: type :  using to their proper  should be :  using their proper

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
```diff
@@ -0,0 +1,70 @@
...
+          // format defined in CRM_Hrjobcontract_ExportImportValuesConverter class.
+          if (substr($column, 0, 13) === 'hrjobcontract') {
+            list(, $entity, $field) = explode('_', $column, 3);
+            $row[$column] = $ei->export($entity, $field, $row[$column]);
+          }
+        }
+
+        $componentDetails[] = $row;
+      }
+      if ($exportMode == 'financial') {
```
omarabuhussein: I am not sure about that but I think no need to do this check here .. we only use this for job contracts export

coldrunKacper: But the hook in hrjobcontract extension gets executed not only for Job Contract export but probably on each export so maybe better to leave this IF as it is.

hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
```diff
@@ -0,0 +1,54 @@
...
+        // format defined in CRM_Hrjobcontract_ExportImportValuesConverter class.
+        if (substr($column, 0, 13) === 'hrjobcontract') {
+          list(, $entity, $field) = explode('_', $column, 3);
+          $row[$column] = $converter->export($entity, $field, $row[$column]);
+        }
+      }
+
+      $componentDetails[] = $row;
+    }
+    if ($exportMode == 'financial') {
```
omarabuhussein: I think you forget to  remove this extra if ... no need for it inside job contract exporter ( I think ) .. you can just call :

$getExportFileName = CRM_Export_BAO_Export::getExportFileName('csv', $exportMode);

directly

hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
```diff
@@ -0,0 +1,54 @@
...
+   * @param $sqlColumns
+   * @param $exportMode
+   * @param null $saveFile
+   * @param string $batchItems
+   */
+  public static function writeCSVFromTable($exportTempTable, $headerRows, $sqlColumns, $exportMode, $saveFile = NULL, $batchItems = '') {
+    $converter = CRM_Hrjobcontract_ExportImportValuesConverter::singleton();
+    $componentDetails = array();
+
+    $dao = CRM_Core_DAO::executeQuery("SELECT * FROM {$exportTempTable}");
```
omarabuhussein: this is the first solution came to my mind when I saw how civicrm implemented it .. but then I followed why they where doing the LIMIT thing and it was to keep memory usages limited and not to grow so much specially when the table contain a lot of rows  .. so I think implementing it  in the way I suggested may be a better approach .. what do you think ?

hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
omarabuhussein: oh sorry just realized you already replied their .

hrjobcontract/CRM/Hrjobcontract/Export/Export.php
omarabuhussein: I believe their should be  check inside the hook to ensure it only get executed for job contract export .. and any check inside this class should be removed .

coldrunKacper: The issue is that there may not be just 'Job Contract' export or non-Job Contract export. You can do an export of Contacts with let's say two Job Contract fields and a lot of non-Job Contract fields.

omarabuhussein: ah I see ...

hrjobcontract/CRM/Export/BAO/Export.php
```diff
@@ -0,0 +1,2208 @@
...
+   * @param array $returnProperties
+   *   Return Properties
+   * @param object $query
+   *   CRM_Contact_BAO_Query
+   *
+   * @return string $groupBy
+   *   Group By Clause
+   */
+  public static function getGroupBy($exportMode, $queryMode, $returnProperties, $query) {
+    // HRJobContract extension override by PCHR-1409.
```
omarabuhussein: there are couples things are needed :

1- overriding the file should be in a commit and the changes happen to this file is in another commit so it will be easier to track down the changes ,

2- add  @custom before the comment to point that this part of the code is a custom code 

```// @custom HRJobContract extension override by PCHR-1409.```

3- document the changes in this page https://compucorp.atlassian.net/wiki/display/PCHR/Overridden+CiviCRM+files

PR #1218, opened 2016-10-25
PCHR-1665: fix civihr core extension key in drush-install

omarabuhussein:
when creating a new site via the buildkit you wil get the following :
```
Array
(
is_error => 1
error_message => Unknown extension: uk.co.compucorp.civicrm.civihrcore
)
```
that's because the extension name is changed to uk.co.compucorp.civicrm.hrcore but it was not changed in in drush-install.sh file which was causing the issue.


coldrunKacper:
Approved.

PR #1219, opened 2016-10-27
PCHR-1353 - Support for multiple style guides

totten:
This updates the branch `PCHR-1353-update-styleguide` to include support for
multiple style-guides.  To see this, navigate to "Support => Developer =>
Style Guide".  Each subitem ( "crm-*", "Bootstrap", "Bootstrap-CiviCRM",
"Bootstrap-CiviHR") loads a different style-guide.

To create a new style-guide directly in this extension, one would:
 * Add a folder (eg `org.civicrm.styleguide/guides/mynewguide`). It should have subfolders for `docs`, `markup`, etal.
 * Update `CRM/Styleguide/Manager.php` to declare it.

To create a new style-guide in another extension, one would:
 * Add a folder (e.g. `org.civicrm.foozball/guides/foozball`). It should have subfolders for `docs`, `markup`, etal.
 * Implement  `hook_civicrm_styleGuides(&$styleGuides)`

This current set of patches has a couple rough-spots/TODOs, although I don't
think that should block merging into `PCHR-1353-update-styleguide`:

 * The Bootstrap style guide still lives under `partials/`. For consistency, we should move it to `guides/bootstrap`.
   (I didn't make the move because it sounded like there would be more development in there - and I didn't want to cause a merge-conflict on that.)
 * It should have a basic, high-level test-case which renders a couple style-guides. (I haven't written this because
   I'm on the fence about which tool to use; may be good to sleep-on-it and/or discuss.)
 * The prose within a style guide cannot be localized because the partials are expressed as `*.html`. (I don't think this is particularly important, though.)


org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
...
+          'path' => "{$extPath}/guides/bootstrapcivicrm",
+        ),
+        // FIXME: Consider moving declaration to another extension.
+        'bootstrapcivihr' => array(
+          'label' => ts('Bootstrap-CiviHR'),
+          'path' => "{$extPath}/guides/bootstrapcivihr",
+        ),
+      );
+
+      CRM_Utils_Hook::singleton()->invoke(1, $this->all,
```
davialexandre: I think it would be better to not expose the property like this (it's is private after all, isn't?). What do you think of this?:
- Add a public `addStyleGuide()` method to CRM_Styleguide_Manager
- Pass the CRM_Styleguide_Manager instance to the hooks, so people can use the `addStyleGuide()` instead of inserting everything in to the array

This will also avoid people from removing style guides from array. If that's is not desired, then you can all add a `removeStyleGuide()` method.

org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
...
+        CRM_Utils_Hook::$_nullObject,
+        CRM_Utils_Hook::$_nullObject,
+        'civicrm_styleGuides'
+      );
+
+      foreach ($this->all as $name => &$value) {
+        $value['name'] = $name;
+      }
+    }
+    return $this->all;
```
davialexandre: I think this whole method can be refactored a bit to improve readability:
- The default style guides can be defined in another method
- The hook invocation can also be in a separate method
- Finally, the foreach should also be in a separate method, with a proper description of why you need the 'name' index for

If you think that's too much, at the least the default style guide definition should be in another method. That's what is making this method soo long

org.civicrm.styleguide/styleguide.php
```diff
@@ -123,3 +123,41 @@ function styleguide_civicrm_angularModules(&$angularModules) {
...
+ */
+function styleguide_civicrm_navigationMenu(&$menu) {
+  _styleguide_civix_insert_navigation_menu($menu, 'Support/Developer', array(
+    'label' => ts('Style Guide', array('domain' => 'org.civicrm.styleguide')),
+    'name' => 'developer_styleguide',
+    'permission' => 'access CiviCRM',
+    'operator' => 'OR',
+    'separator' => 0,
+  ));
+  foreach (Civi::service('styleguides')->getAll() as $name => $defn) {
```
davialexandre: I don't think that saving 6 letters is worth the reduced readability. `$definition` would be better here. Also, you're not using `$name` anywhere inside the loop, so no need for it.

org.civicrm.styleguide/styleguide.php
```diff
@@ -123,3 +123,41 @@ function styleguide_civicrm_angularModules(&$angularModules) {
...
+    _styleguide_civix_insert_navigation_menu($menu, 'Support/Developer/developer_styleguide', array(
+      'label' => $defn['label'],
+      'name' => 'developer_styleguide_' . $defn['name'],
+      'url' => 'civicrm/styleguide/' . $defn['name'],
```
davialexandre: oh, I see the name is used to build the url. So maybe it's good to add some parsing on the CRM_Styleguide_Manager in order to avoid invalid URL chars

org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+class CRM_Styleguide_Manager {
```
davialexandre: Why not CRM_StyleGuide_Manager?

org.civicrm.styleguide/styleguide.php
```diff
@@ -123,3 +123,41 @@ function styleguide_civicrm_angularModules(&$angularModules) {
...
+  }
+}
+
+/**
+ * Implements hook_civicrm_container().
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_container
+ */
+function styleguide_civicrm_container(\Symfony\Component\DependencyInjection\ContainerBuilder $container) {
+  $container->setDefinition('styleguides', new \Symfony\Component\DependencyInjection\Definition(
```
davialexandre: Wouldn't it be better to name the service as something like `style_guides`, `style-guides` or `styleGuides`? I'm in favor of the first option, as it seems to be the standard in Symfony and its bundles

org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * @param string $name
+   * @return array|NULL
+   *   If the style-guide exists, it is an array with properties:
+   *     - name: string
+   *     - label: string
+   */
+  public function get($name) {
+    $all = $this->getAll();
+    return @$all[$name];
```
davialexandre: Please change the error suppression to a if or something like:
```php
return empty($all[$name]) ? null : $all[$name];
```

Anything is better than the error suppression operator :D

org.civicrm.styleguide/templates/CRM/Styleguide/Page/StyleGuide.tpl
```diff
@@ -1,5 +1,5 @@
 {php}
-  $htmlBuilder = new CRM_Styleguide_HtmlBuilder();
+  $htmlBuilder = new CRM_Styleguide_HtmlBuilder($this->get_template_vars('styleguide'));
```
davialexandre: Wouldn't it be better to instantiate the HtmlBuilder inside the Page controller and then assign it to the template?

davialexandre:
@totten @AkA84 about the tests, it does indeed need some. Since this is a totally visual thing, a tool like https://github.com/garris/BackstopJS sounds appropriate

AkA84:
@davialexandre yes implementing Backstop was already in the pipeline, but I think it's outside the scope of this PR

@totten I'll got more into the details of this PR later, just one quick thing: you're missing the ticket number in the last commit title

org.civicrm.styleguide/CRM/Styleguide/Manager.php
totten: In its defense, I don't think this line is why `@` gets a bad rep (e.g. `@include "file"` would be a real problem), but thanks for calling it out since it's better to have a culture where `@` just doesn't exist. ;) I'll update it.

org.civicrm.styleguide/CRM/Styleguide/Manager.php
totten: I'd also lean toward `CRM_StyleGuide_Manager`, but the namespace was already setup when I added the class.

Guess: The extension is using the default namespace after running `civix generate:module org.civicrm.styleguide`. 

It could be changed, but that seemed outside scope of my patch. Steps would be:
 * Rename the folder
 * Rename all the classes in the folder
 * Update `info.xml` (`<civix>...<namespace>...`)
 * Make sure that any dev machines respect the new name. (Changing file capitalization is wonky across platforms.)

org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Get a list of all style guides.
+   *
+   * @return array
+   */
+  public function getAll() {
+    if ($this->all === NULL) {
+      $extPath = CRM_Core_Resources::singleton()->getPath('org.civicrm.styleguide');
+
+      $this->all = array(
```
totten: The `[...]` notation is nicer, but I consider the "style guide" functionality on a track for official adoption (i.e. aiming to bundle with core). As of today, that still requires PHP 5.3 compatibility -- thus `array(...)`.

org.civicrm.styleguide/CRM/Styleguide/Manager.php
totten: This is a good idea. Makes it easier to adjust the implementation without changing the contract.

org.civicrm.styleguide/CRM/Styleguide/Manager.php
```diff
@@ -0,0 +1,70 @@
...
+
+class CRM_Styleguide_Manager {
+
+  /**
+   * @var array
+   */
+  private $all = NULL;
+
+  /**
+   * Get the definition of specific style-guide by name.
```
totten: Fair enough. 

When the phrase appears by itself (like in a page-title), it doesn't matter much. For prose in a sentence, I find "style guide" harder to read than "style-guide". The issue is that both words are multipurpose (e.g. "style" can be noun or verb; "guide" can be noun or verb), and you have to absorb more of the sentence to be sure of the meaning. With the "-", you can rest assured that it's intended as the compound noun.

org.civicrm.styleguide/styleguide.php
totten: Agree `style_guides` is closer to Symfony convention.

org.civicrm.styleguide/templates/CRM/Styleguide/Page/StyleGuide.tpl
totten: Stylistically, agree. I played with changing that and decided to pick my battles.

org.civicrm.styleguide/CRM/Styleguide/Manager.php
totten: It's a little nicer now that:
 * `add(...)` handles the cleanup for the `name` property.
 * `getAll()` has been split apart (`getAll()` and `init()`).

totten:
I've incorporated most of Davi's feedback and prepended that commit message with the issue number.

I did the rename of `CRM_Styleguide` to `CRM_StyleGuide` last because this has the most potential to be problematic. I don't mind much if we drop those commits. You might need to take special measures on your local (esp. if you have a case-insensitive filesystem). However, since none of this stuff has been really published yet, it's probably the best time to change it.

totten:
Regarding testing: BackstopJS sounds like a pretty cool way to test the visuals and CSS. Wouldn't that test be aimed at a particular theme -- i.e. the "theme" actually defines the look-and-feel. Given two different themes, each one would have different benchmarks for what looks "right".

For this patch, all I imagined was testing that the page loads different+hookable style-guides.

davialexandre:
@totten it's approved now!

PR #1221, opened 2016-10-28
PCHR-1645: Ensure revision is created for each job contract entity via the importer

omarabuhussein:
## Problem

When importing contracts, if the data for specific entity is not exists such as leave entitlements or health data then no revision will get created for that entity which cause loading the imported contracts to fail, The problem does not happen when creating job contract from the UI because a revision for each entity will be created even if no data inserted for it.
## Solution

Changing the entity handler importer code for each entity to ensure the revision will get created , and changing how the contracts get imported via the API rather than calling the BAO methods directly for consistency since the API methods for some entities such as leave entity contain some additional logic not exits in BAO methods.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobHealth.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobHealth.php
@@ -8,15 +8,11 @@ class CRM_Hrjobcontract_Import_EntityHandler_HRJobHealth extends CRM_Hrjobcontra
   public function handle(array $params, CRM_Hrjobcontract_DAO_HRJobContractRevision $contractRevision, array &$previousRevision) {
     $entityParams = $this->extractFields($params);
 
-    if(count($entityParams) === 0) {
-      return array();
-    }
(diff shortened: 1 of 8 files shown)


hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobHealth.php
```diff
@@ -8,15 +8,13 @@ public function __construct() {
...
-
-    $entityParams['import'] = 1;
+    $entityParams['sequential'] = 1;
     $entityParams['jobcontract_id'] = $contractRevision->jobcontract_id;
     $entityParams['jobcontract_revision_id'] = $contractRevision->id;
 
-    return array(CRM_Hrjobcontract_BAO_HRJobHealth::create($entityParams));
+    $result = civicrm_api3('HRJobHealth', 'create', $entityParams)['values'][0];
+
+    return $result;
```
coldrunKacper: Why not return civicrm_api3(...) (one line)?

coldrunKacper:
Approved.

PR #1225, opened 2016-11-01
PCHR-1692: Update entitlements when a new absence period is added

davialexandre:
**Problem**
When a new absence period is added, the entitlement for each contact is set to 0 instead of fetching the value from the contractual entitlement (the value set on the Leave tab of the contract dialog)

**Solution**
Whenever a new absence period is saved, a new task, containing the ID of the just saved period, is added to a "Entitlement Recalculation Queue", which will be processed by a new scheduled job that will update the entitlements for that period for all contacts.

This was created as a scheduled job because organizations can have a lot of contacts and doing this calculation for all of them might take some time.

-------------------------------------------------------------------------

**Refactoring**
The HRJobLeave BAO had a lot of methods which accessed directly tables from other extensions. I moved these the extension/classes they belong to.

**Why some few tests?**
The hrabsence extension will soon be replaced by L&A, so there's no point on spending too much time on something that won't last long. I just added a few to check that the positive path is working.

### DIFF
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
@@ -83,4 +83,53 @@ class CRM_HRAbsence_BAO_HRAbsencePeriod extends CRM_HRAbsence_DAO_HRAbsencePerio
     $absencePeriod->find(TRUE);
     $absencePeriod->delete();
   }
+
+  /**
+   * Returns an array of Absence Periods dates overlapping the given start and
+   * end dates
+   *
+   * @param string|null $startDate
+   * @param string|null $endDate
+   *
+   * @return array
+   *  An array of Absence Periods dates as:
+   *  [
+   *   ['start' => '2016-01-01', 'end' => '2016-12-31'],
+   *   ['start' => '2017-01-01', 'end' => '2017-12-31'],
+   *   ...
+   *  ]
+   */
+  public static function getAbsencePeriods($startDate = NULL, $endDate = NULL) {
+    $data       = [];
+    $query      = "SELECT * FROM civicrm_hrabsence_period ";
+    $where      = [];
+    $params     = [];
+    $whereQuery = '';
+
+    if ($startDate) {
+      $startDate = date('Y-m-d H:i:s', strtotime($startDate));
+      $where[]   = " end_date >= %1 ";
+      $params[1] = [$startDate, 'String'];
+    }
+
+    if ($endDate) {
+      $endDate   = date('Y-m-d H:i:s', strtotime($endDate));
+      $where[]   = " start_date < %2 ";
+      $params[2] = [$endDate, 'String'];
+    }
+
+    if (!empty($where)) {
+      $whereQuery = ' WHERE ' . implode(' AND ', $where);
+    }
+
+    $periods = CRM_Core_DAO::executeQuery($query . $whereQuery, $params);
+    while ($periods->fetch()) {
+      $data[$periods->id] = array(
+        'start' => $periods->start_date,
+        'end'   => $periods->end_date,
+      );
+    }
+
+    return $data;
+  }
 }
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
@@ -25,6 +25,8 @@
 +--------------------------------------------------------------------+
 */
 
+use CRM_HRAbsence_Queue_EntitlementRecalculation as EntitlementRecalculationQueue;
+
 class CRM_HRAbsence_BAO_HRAbsencePeriod extends CRM_HRAbsence_DAO_HRAbsencePeriod {
 
   public static function create($params) {
@@ -47,7 +49,7 @@ class CRM_HRAbsence_BAO_HRAbsencePeriod extends CRM_HRAbsence_DAO_HRAbsencePerio
 
     CRM_Utils_Hook::post($hook, $entityName, $instance->id, $instance);
 
-    CRM_HRAbsence_BAO_HRAbsenceEntitlement::recalculateAbsenceEntitlementsForPeriod($instance->id);
+    self::enqueueEntitlementRecalculation($instance);
 
     return $instance;
   }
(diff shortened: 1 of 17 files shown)


hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsencePeriod.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+class CRM_HRAbsence_Test_Fabricator_HRAbsencePeriod {
+
+  public static function fabricate($params = []) {
+    if(empty($params['title'])) {
```
omarabuhussein: I guess it would be better to set a default data to the current year start and end data if the parameters are not set.

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsencePeriodTest.php
```diff
@@ -0,0 +1,72 @@
...
+    $absencePeriodTable = AbsencePeriod::getTableName();
+    // We need to disable fk checks in order to avoid constraint
+    // errors when truncating the absence period table
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
+    CRM_Core_DAO::executeQuery("TRUNCATE $absencePeriodTable");
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 1;");
+
+    $this->absencePeriods[] = AbsencePeriodFabricator::fabricate([
+      'name' => 'Period 1',
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
```
omarabuhussein: I think it would make more sense to update absence period create method to process the data .

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsencePeriodTest.php
```diff
@@ -0,0 +1,72 @@
...
+    $this->absencePeriods[] = AbsencePeriodFabricator::fabricate([
+      'name' => 'Period 3',
+      'start_date' => CRM_Utils_Date::processDate('2018-01-01'),
+      'end_date'   => CRM_Utils_Date::processDate('2018-12-31'),
+    ]);
+  }
+
+  public function testGetAbsencePeriodWithoutDatesReturnsAllTheAbsencePeriods() {
+    $periods = AbsencePeriod::getAbsencePeriods();
+    // Assert only that the number of items is the expected one,
```
omarabuhussein: if you changed the create method for absence period to process the data instead ... then I guess it will be easier to assert the other fields such as start and end data

hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
```diff
@@ -134,4 +136,21 @@ public static function getAbsencePeriods($startDate = NULL, $endDate = NULL) {
 
     return $data;
   }
+
+  /**
+   * Add a new task to the Entitlement Recalculation queue to recalculate it for
+   * the given absence period
+   *
+   * @param \CRM_HRAbsence_BAO_HRAbsencePeriod $period
```
omarabuhussein: you only need the period id here so why not just pass the ID instead of the whole object instea

hrabsence/hrabsence.php
```diff
@@ -235,6 +235,8 @@ function hrabsence_civicrm_install() {
   );
   civicrm_api3('message_template', 'create', $msg_params);
 
+  _hrabsence_add_scheduled_jobs();
```
omarabuhussein: I think it would be a better Idea to move anything inside this hook to the upgrader install() method , as well as any function you defined here could be moved to the upgrader class .. it is better than polluting civicrm with more global functions and make this file easier to understand ... I already did the same in same data extension and thinking to to do the same for anything I work on .

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsencePeriodTest.php
davialexandre: It's not that simple. I would have process the date before inserting it into the database and after, meaning I would need to add logic to the fabricator, which I would prefer to avoid as much as possible (it's used only to create data).

Besides, as I mentioned on the PR description, this is for hrabsence extension which will soon be replaced by L&A. I don't think it's worth spend time making it perfect. When we migrated from 4.7 this extensions tests were not even updated for the same reason. So I think that the test is good enough just the way it is

hrabsence/CRM/HRAbsence/BAO/HRAbsencePeriod.php
davialexandre: I prefer to pass objects as much as possible (we're doing OOP after all, aren't we?). This has a couple of advantages:
- If we need to change the method to work based on more absence period fields, then we won't need to update the method signature
- We can use IDE autocomplete feature

Also, MAYBE, there is a bit of performance improvement as objects are passed as a reference, whereas the id would be passed by value, resulting a new copy of the variable (maybe the copy-on-write feature helps with that, but I don't know tbh)

hrabsence/hrabsence.php
davialexandre: Yeah, I totally agree with you here. To be honest I even tried to do that, but I noticed some of the code here is already duplicated on the extension's upgrader methods. So I was afraid to break something in doing this. Since the extension will soon be removed, I think the global function is good enough and it avoids the duplication between the hook_install and the upgrader

davialexandre:
@omarabuhussein I am 99.99999999999999% sure that the scripts are not used anywhere. Nobody really knows what are they for or even knew about them. Kacper thinks they were created to run some tests on the command line, but he is not sure.

I decided to keep them for now and then remove them when L&A replaces hrabsence

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsencePeriodTest.php
davialexandre: You mean the create method of the BAO? If yes, that is not a good idea because of this https://gist.github.com/davialexandre/36c27089b2dbb8630fa4ba1a1d942c4f

hrabsence/hrabsence.php
omarabuhussein: the upgrader install() method only work when installing the extension same as install hook .. so moving the code their won't screw anything .. but as I said i won't push on anything here since this is going to be replaced .. but let me know what you decided to do.

davialexandre: In this case, I'm going with the option which will take less time: don't change anything :D

PR #1228, opened 2016-11-02
PCHR-1492: Redundant export fields

omarabuhussein:
## Problem

1. Go to Advanced Search.
2. Select contact and use Actions drop-down to select Export action
3. The following fields are legacy fields and should be removed:

- Addressee
- Communication style
- Current employer
- Current employer ID
- Do not email
- Do not mail
- Do not phone
- Do not SMS
- Do not trade
- Job role functional area
- Job role hours
- Job role hours unit
- Use for bulk mails
- No bulk emails
- On hold
- Preferred communication method
- Preferred language
- Preferred mail format
- Street number
- Street number suffix
- Street unit
- Open ID
- Unique ID

Also (CiviHR ID) Field is missing ,

## Solution 

Since all of these fields except job roles related fields are set in contact DAO class which cannot be altered ,  I had to hide these fields via JavaScript by using hrui_civicrm_alterContent hook ( most of these field where already hidden for import form but I've added the rest and added a new IF condition to allow these fields to be hidden for export form too ).

Regarding job roles fields :

- Job role functional area
- Job role hours
- Job role hours unit

I had to remove them completely since they are not used anywhere .

Also I found that the job roles XML schema file is outdated , so I've updated it and regenerated the DAO again but since the DAO generator is not able to generate foreign keys referencing columns from different extensions I had to drop the Foreign key from job_contract_id field and then added an index to it instead. This will break data integrity for now but I need to fix the DAO generator first and then come back to fix this .


davialexandre:
This has gone stale for more than 1 year. Closing it

PR #1233, opened 2016-11-04
PCHR-1703: Update change_reason field type

AkA84:
## Problem
The values of the "Job Contract Revision Change Reason" options can be edited by the CiviHR admin, meaning that they could be either integer or string. The problem is that `change_reason` field type is integer, so it won't accept any custom value that an admin has set, resulting in an error when creating a new revision

## Solution
Change the type of the field from integer to string, for both new and old installations


hrjobcontract/tests/phpunit/api/v3/HRJobContractRevisionTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Creates a revision via the api
+   *
+   * @param  [int] $contract_id
+   * @param  [Array] $reason    The reason change option
+   * @return [Array]
+   */
+  private function createRevision($contract_id, $reason) {
+    return civicrm_api3('HRJobContractRevision', 'create', array(
```
AkA84: @omarabuhussein @davialexandre the reason I didn't use a fabricator here is that this is explicitly about using the api to create a revision, while a fabricator hide the implementation details from you, so you don't know if it's using the api, the BAO, raw sql, etc.

I'm not sure this is the optimal approach, let me know what you think

PR #1234, opened 2016-11-04
PCHR-1674: Managers style fixes

deb1990:
**Requirement**:
1. Fix table styles
2. Fix pagination
3. Fix button padding

**Before**
image

**After**
image


JonScreat:
Pagination should be as seen here (05. Tables): https://invis.io/MU4RTI1S5#/111453672_Page_1_-_Common_Elements

PR #1236, opened 2016-11-07
PCHR-1707: Don't update absence period entitlement from contract after contract change

davialexandre:
**Problem**
The period entitlement could be updated in two different ways:
- Manually, by clicking on the Absence tab of the contact page and then setting the values for each type and period
- Automatically, after updating a contract. In this case, the period entitlement is updated with whichever value is set on the leave tab of the contract dialog

The problem with the automatic update is that it would overwrite any existing entitlement set manually.

**Solution**
No, in order to no overwrite things, the method which recalculates the entitlement, will only save the new value if there is no existing entitlement for a contact and absence type during a given period.

### DIFF
--- a/hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceEntitlementTest.php
+++ b/hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceEntitlementTest.php
@@ -63,11 +63,11 @@ class CRM_HRAbsence_BAO_HRAbsenceEntitlementTest extends PHPUnit_Framework_TestC
 
     CRM_HRAbsence_BAO_HRAbsenceEntitlement::recalculateAbsenceEntitlementsForPeriod($period->id);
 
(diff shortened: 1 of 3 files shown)


hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceEntitlementTest.php
```diff
@@ -90,4 +90,67 @@ public function testRecalculateAbsenceEntitlementsForPeriodCreatesTheContactsEnt
     $this->assertEquals(23, $contact2Entitlement->amount);
   }
 
+  public function testRecalculateAbsenceEntitlementDoesNotOverwriteExistingEntitlements() {
+    $absenceType = AbsenceTypeFabricator::fabricate();
+
+    $contact = ContactFabricator::fabricate([]);
```
omarabuhussein: you don't need to pass empty array here , contact fabricator already have an empty array as default parameter.

PR #1243, opened 2016-11-11
PCHR-1177: Fixed API Call Error When Marking Tasks as Complete

MiyaNoctem:
Issue affected admin part of CiviHR also.  The problem seems to happen because the activity's data is never queried from the DB.  Thus, the activity's type is null when the check is performed, and causes the error when the null value is used to bring a single result from the activitytype table.

Fixed by passing activity id instead of activity_type_id to activityCreatedByTaskandAssignments(), so data can be loaded from DB within function.

Should inspect the code to assure this same assumption will not cause further issues (assuming the object is loaded with all the information in DB).

### DIFF
--- a/hrcase/hrcase.php
+++ b/hrcase/hrcase.php
@@ -341,7 +341,10 @@ function hrcase_civicrm_alterContent( &$content, $context, $tplName, &$object ) 
 }
 
 function hrcase_civicrm_post( $op, $objectName, $objectId, &$objectRef ) {
-  if ($objectName == 'Activity' && isset($objectRef->case_id) && !activityCreatedByTaskandAssignments($objectRef->activity_type_id)) {
+  // PCHR-1177: $objectRef is just an instance of CRM_Activity_DAO_Activity, with values set from parameters sent through request.
+  // Thus, it cannot be used to obtain data for activity that could be in database for given activity id.  We need to load the info from DB first!
+  // Passing activity id instead of object, so data can be loaded from DB within function.
+  if ($objectName == 'Activity' && isset($objectRef->case_id) && !activityCreatedByTaskandAssignments($objectId)) {
     $component_id = CRM_Core_Component::getComponentID('CiviCase');
     $contact_id =  CRM_Case_BAO_Case::retrieveContactIdsByCaseId($objectRef->case_id);
     $hrjob = civicrm_api3('HRJobContract', 'get', array(
@@ -447,10 +450,16 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
 /**
  * function to check if the activity is created by task and assignments extension
  *
- * @param int $activity_type_id
+ * @param int $activity_id
  * @return boolean
  */
-function activityCreatedByTaskandAssignments($activity_type_id) {
+function activityCreatedByTaskandAssignments($activity_id) {
+  // PCHR-1177 - Status update from task lists caused a null activity type id to passed as parameter for this method.
+  // Passed activity_id instead to load data from db instead.
+  $params = array('id' => $activity_id);
+  $taskInDB = CRM_Activity_BAO_Activity::retrieve($params);
+  $activity_type_id = $taskInDB->activity_type_id;
+
   // check if task and assignments is enabled
   $isEnabled = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Extension', 'uk.co.compucorp.civicrm.tasksassignments', 'is_active', 'full_name');
   if(!$isEnabled) {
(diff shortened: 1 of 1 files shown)


hrcase/hrcase.php
```diff
@@ -341,7 +341,10 @@ function hrcase_civicrm_alterContent( &$content, $context, $tplName, &$object )
 }
 
 function hrcase_civicrm_post( $op, $objectName, $objectId, &$objectRef ) {
-  if ($objectName == 'Activity' && isset($objectRef->case_id) && !activityCreatedByTaskandAssignments($objectRef->activity_type_id)) {
+  // PCHR-1177: $objectRef is just an instance of CRM_Activity_DAO_Activity, with values set from parameters sent through request.
+  // Thus, it cannot be used to obtain data for activity that could be in database for given activity id.  We need to load the info from DB first!
+  // Passing activity id instead of object, so data can be loaded from DB within function.
+  if ($objectName == 'Activity' && isset($objectRef->case_id) && !activityCreatedByTaskandAssignments($objectId)) {
```
davialexandre: The right place for these type of comments is the commit message and/or the PR description. Comments on the code can easily get outdated and will give the wrong information to the developer

hrcase/hrcase.php
```diff
@@ -447,10 +450,16 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
...
  * function to check if the activity is created by task and assignments extension
  *
- * @param int $activity_type_id
+ * @param int $activity_id
  * @return boolean
  */
-function activityCreatedByTaskandAssignments($activity_type_id) {
+function activityCreatedByTaskandAssignments($activity_id) {
+  // PCHR-1177 - Status update from task lists caused a null activity type id to passed as parameter for this method.
+  // Passed activity_id instead to load data from db instead.
```
davialexandre: Same thing here. This should be on the commit message or PR description as well

hrcase/hrcase.php
```diff
@@ -447,10 +450,16 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
...
+  $params = array('id' => $activity_id);
```
davialexandre: Please use the short array syntax

hrcase/hrcase.php
```diff
@@ -447,10 +450,16 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
...
+  $taskInDB = CRM_Activity_BAO_Activity::retrieve($params);
+  $activity_type_id = $taskInDB->activity_type_id;
```
davialexandre: You don't need this assignment. You can use the value in $taskInDB directly.

hrcase/hrcase.php
```diff
@@ -447,10 +450,16 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
...
- * @param int $activity_type_id
+ * @param int $activity_id
  * @return boolean
  */
-function activityCreatedByTaskandAssignments($activity_type_id) {
+function activityCreatedByTaskandAssignments($activity_id) {
+  // PCHR-1177 - Status update from task lists caused a null activity type id to passed as parameter for this method.
+  // Passed activity_id instead to load data from db instead.
+  $params = array('id' => $activity_id);
+  $taskInDB = CRM_Activity_BAO_Activity::retrieve($params);
```
davialexandre: The `$taskInDB` name sounds a bit odd to me. What if, for some reason, the task is not stored on DB anymore? This is just an infrastructure details, so I don't think it should be part of the name. Probably it would be better to just name it as $activity, after all, the method name is `activityCreatedByTaskandAssignments()`

hrcase/hrcase.php
```diff
@@ -340,10 +340,15 @@ function hrcase_civicrm_alterContent( &$content, $context, $tplName, &$object )
   }
 }
 
+/**
+ * Implementation of hook_civicrm_post, executed after task creation/update
+ *
+ * @param $op string, the type of operation being performed;
```
davialexandre: The convention we use is the same one as Drupal's on which the @param description should be on a newline ([see here](https://www.drupal.org/node/1354#param)). Also, the type should come before the param name, so:

```
@param string $op
  The Type of the operation being performed
```

There's no need for period or semicolon at the end of the description.

By the way, if your IDE/Editor supports coding styles, it might be good to set it to use Drupal's to help you follow the standards.

hrcase/hrcase.php
```diff
@@ -454,11 +459,9 @@ function hrcase_getActionsSchedule($getNamesOnly = FALSE) {
...
  */
 function activityCreatedByTaskandAssignments($activity_id) {
-  // PCHR-1177 - Status update from task lists caused a null activity type id to passed as parameter for this method.
-  // Passed activity_id instead to load data from db instead.
-  $params = array('id' => $activity_id);
-  $taskInDB = CRM_Activity_BAO_Activity::retrieve($params);
-  $activity_type_id = $taskInDB->activity_type_id;
+  $params = ['id' => $activity_id];
+  $activity = CRM_Activity_BAO_Activity::retrieve($params);
+  $activity->activity_type_id;
```
davialexandre: What is this line doing?

PR #1244, opened 2016-11-11
PCHR-1709: Update the revision details after creating a new revision

sleekvick:
**The Problem:**

Basically the new revision details does not update, this is because the variable **isCurrentRevision** remains false if the **dateEffectiveRevisionCreated is < dateToday** and if the variable remains false it doesn't call the **updateContractView(results)** function however the **ContractService.updateHeaderInfo();** gets called that's why the header updates. 

**The Solution**

fixed_revision_bug

I see no reason why the **updateContractView(results)** should not be called if the **results** variable does not get mutated inside the if block. so I moved the execution of the **updateContractView(results)** to the of **ContractService.updateHeaderInfo();**


hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -249,9 +250,9 @@ define([
                                 (dateEffectiveRevisionCurrent > dateToday &&
                                 dateEffectiveRevisionCreated <= dateEffectiveRevisionCurrent);
 
+
```
davialexandre: On the other hand, I think you don't need this additional blank line here. Just one is enough :D

davialexandre:
So, the description on the ticket also mentions that the contract details displayed when you click on "View current revision" are not updated. The gif you added to the PR doesn't show you testing that. Can you please update it?

Also, please squash the commits. The second one doesn't is just adding/removing some empty lines, so there's no reason to keep a commit just for that

PR #1245, opened 2016-11-14
PCHR-1195: Implement the Public Holiday Leave Request creation/delete logic

davialexandre:
The Creation and Deletion logic were implemented as two services:
- CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreation
- CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion

The creation service has the `createForContact()` method which can create a Public Holiday Leave Request for a given Contact and PublicHoliday. The deletion service has the `deleteForContact`, which accepts the same parameters, but deletes an existing Public Holiday Leave Request.

Both logics are capable of handling regular LeaveRequests overlapping a public holiday. When creating a Public Holiday Leave Request, any overlapping LeaveRequest date will have it's deduction amount update to 0 (so we don't deduct the day twice, one from the LEave Request and other from the Public Holiday). When deleting a Public Holiday Leave Request, the process happens in the opposite direction and the deduction for the overlapping Leave Request Date will be update to it's proper value according to the Work Pattern.

In order to support all this logic, a few support methods and entities had to be added:
- ContactWorkPattern. A new entity to link contact with WorkPatterns. This was required for the deletion logic, in order to get the right WorkPattern to be used to calculation de deduction for a LeaveRequest date
- Methods to get the breakDown and TotalBalance for a LeaveRequest. Currently, these were used basically for tests, but they are methods that would have to be implemented in the future anyway
- Added fabricators for WorkPattern, LeaveRequest, LeaveBalanceChange and PublicHolidayLeaveRequest

Also, lots of refactoring and small improvements were introduced

### What are Public Holiday Leave Requests?
In CiviHR, there are some absence types with an option called "Must take public holiday as leave". Which means that, for that absence type, every public holiday should be a leave request. Since this is mandatory, the system should automatically create such requests, which are just regular Leave Request with a few different options:
- The status is "Admin Approved"
- Internally, the type of the balance change will be "Public Holiday" (a way to identify which leave request as from Public Holidays)


coldrunKacper:
Approved.

PR #1247, opened 2016-11-17
PCHR-1684: Increase field width

deb1990:
**Problem**
The Government ID field is not long enough when adding or editing a contact. The text is overlapping to the arrow.

**Before**
image

**After**
image

**Explanation**: 
Removed fixed width which was causing the fixed size issue for all dropdowns. Added padding right for only `select` type markup because otherwise the element was taking width based on content and the arrow section was overlapping with text(as arrow section is custom).


deb1990:
@AkA84 No dropdowns have fixed width, should I look into that?

AkA84:
@deb1990 if possible, otherwise what's the alternative? setting ad-hoc width to all dropdowns?

PR #1248, opened 2016-11-17
PCHR-1721: Modified conditions for updating the contract list

victorlenerd:
## Problem

Creating a past contract for someone who doesn't have a contract yet, and editing the contract's end date to be in the future  gets logged as a Past contract.

`updateContractList(period_end_date)` is the function that handles updating of the contract list. It’s invocation inside the `modalInstance` promise resolver first depends on if the there in a new revision date, 

```js
(dateEffectiveRevisionCreated <= dateToday &&
    dateEffectiveRevisionCreated >= dateEffectiveRevisionCurrent) ||
    (dateEffectiveRevisionCurrent > dateToday &&
    dateEffectiveRevisionCreated <= dateEffectiveRevisionCurrent);
```

Where:
```
dateEffectiveRevisionCreated = Is the date the revision was created
dateToday  = The current day 
dateEffectiveRevisionCurrent = The current contract revision date
```

When ever there’s a change in the contract’s end date the contract list won’t get updated the conditions don’t put in consideration the end date of the contract. Also inside the `updateContractList(period_end_date)`  function this condition would determine if the contract list get’s updated.

```js
($scope.details.period_end_date ?
  new Date($scope.details.period_end_date).getTime() != new Date(newEndDate).getTime() :
 !!$scope.details.period_end_date !== !!newEndDate)
```

However the expression will evaluate to false if the user had specified an end date for the contract, because the first expression of the ternary operator will be returned and we already called updateContractView(newScope) which updates the $scope.details.period_end_date to the newEndDate.
Also if the user had not specified an endDate when creating the contract, the expression would evaluate to undefined and the if block won’t get executed.

## Solution
Firstly I removed the `updateContractList(period_end_date)` function from the `if` block that checks if it’s the current revision, also I modified the `if` block in the `updateContractList(period_end_date)` to check if the `$scope.details.period_end_date` matches the `newEndDate`.
Note: For the header to update debug mode has to be OFF. 

job_contract


hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -73,9 +73,8 @@ define([
 
                 //Is contract end date changed
                 if ($scope.details.period_end_date ?
-                new Date($scope.details.period_end_date).getTime() !== new Date(newEndDate).getTime() :
-                !!$scope.details.period_end_date !== !!newEndDate) {
-
+                  new Date($scope.details.period_end_date).getTime() == new Date(newEndDate).getTime() :
```
deb1990: We should not use `==`, use `===` instead.

deb1990: We should not use `==`, use `===` instead.

PR #1249, opened 2016-11-17
PCHR-1196: Trigger Public Holiday Leave Requests creation/deletion when updating/deleting Absence Types

davialexandre:
The logic should be executed on the following situations:
- When creating a new Absence Type where "Must Take Public Holiday as Leave" is true
- When updating an existing Absence Type change the value of "Must Take Public Holiday as Leave" (either from true to false or from false to true)
- When deleting an Absence Type with "Must Take Public Holiday as Leave"

### How this is implemented?
The PublicHolidayLeaveRequestCreation service was augmented with a new method to create public holiday leave requests for all public holidays in the future. It works by fetching all the dates and all the contracts which overlap those dates. Finally, it loop through all this and add a leave request for each date for each of the contracts contacts.

The PublicHolidayLeaveRequestDeletion service was augmented with a new method to delete all the public holiday leave requests for public holidays in the future. It works by fetching all the dates and then delete the leave requests for those dates.

One more service was created. It is the PublicHolidayLeaveRequest service and it has a single method named updateAllLeaveRequestsInTheFuture(), which basically runs the deletion logic first, deleting everything for public holidays in the future, and then runs the creation logic to create new leave requests for the dates in the future.

Finally, the AbsenceType BAO was updated to enqueue a new background task whenever it gets updated and we identify that the public holiday leave requests must also be updated. This task, when consumed, will use the PublicHolidayLeaveRequest service to update everything. This is done in background because update things for all public holidays in the future and to all contacts might take some time, so it's better for to be executed later and not right after saving the changes. A new scheduled job was created to process this background task queue.

### Why always execute both the deletion and the creation logic?
The deletion only deletes existing leave requests and doesn't thrown an error is none exists. So it's safe (and fast) to execute it if there are no leave requests yet. The creation only creates leave request if there is one Absence Type with "Must Take public holiday as leave". So it's safe (and fast) to execute it if there are no MTPHL Absence Types.

This is how it works in practice:
- A new MTPHL Absence Type is created:
  - Deletion logic is executed. Doesn't delete anything (because there was no MTPHL type before)
  - Creation logic is executed. Leave Requests are created

- A non-MTPHL Absence Type is updated to be MTPHL:
  - Deletion logic is executed. Doesn't delete anything (because there was no MTPHL type before)
  - Creation logic is executed. Leave Requests are created

- A MTPHL Absence Type is updated to be non-MTPHL:
  - Deletion logic is executed. Existing Leave Requests are deleted
  - Creation logic is executed. Nothing is created (because there is no MTPHL type anymore)

- A MTPHL Absence is deleted:
  - Deletion logic is executed. Existing Leave Requests are deleted
  - Creation logic is executed. Nothing is created (because there is no MTPHL type anymore)

Note: This only works because we can have only one MTPHL Absence Type on the system

### What else is included on this PR?
I did some refactoring the the AbsenceType tests to use fabricators and follow our current code style. I also added a validation to the AbsenceType BAO to allow only one type to be MTPHL.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -1,5 +1,7 @@
 <?php
 
+use CRM_HRLeaveAndAbsences_Queue_PublicHolidayLeaveRequestUpdates as PublicHolidayLeaveRequestUpdatesQueue;
+
 class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_AbsenceType {
 
   const EXPIRATION_UNIT_DAYS = 1;
@@ -69,6 +71,8 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
 
     unset($params['is_reserved']);
 
+    $mustUpdatePublicHolidaysLeaveRequests = self::mustUpdatePublicHolidayLeaveRequests($params);
+
     $instance = new $className();
     $instance->copyValues($params);
     $transaction = new CRM_Core_Transaction();
@@ -78,6 +82,9 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
       self::saveNotificationReceivers($instance->id, $params['notification_receivers_ids']);
     }
 
+    if($mustUpdatePublicHolidaysLeaveRequests) {
+      self::enqueuePublicHolidayLeaveRequestUpdateTask();
+    }
     $transaction->commit();
     CRM_Utils_Hook::post($hook, $entityName, $instance->id, $instance);
 
@@ -424,6 +431,46 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    */
   private static function getNotificationReceiversIDs($typeId) {
     return CRM_HRLeaveAndAbsences_BAO_NotificationReceiver::getReceiversIDsForAbsenceType($typeId);
+  }
+
+  /**
+   * Checks, based on the $params array, if the Public Holiday Leave Requests
+   * must be updated after saving the Absence Type.
+   *
+   * They must be updated if one of these conditions is true:
+   *
+   * - If this is a new absence type and must_take_public_holiday_as_leave is true
+   * - If this is an existing absence type and the value of must_take_public_holiday_as_leave has changed
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private static function mustUpdatePublicHolidayLeaveRequests($params) {
+    if(!array_key_exists('must_take_public_holiday_as_leave', $params)) {
+      return false;
+    }
+
+    if(!empty($params['id'])) {
+      $oldValue = self::getFieldValue(self::class, $params['id'], 'must_take_public_holiday_as_leave');
+
+      return $oldValue != $params['must_take_public_holiday_as_leave'];
+    }
+
+    return (bool)$params['must_take_public_holiday_as_leave'];
+  }
+
+  /**
+   * Enqueue a new task to update the Public Holiday Leave Requests due to
+   * changes on one Absence Type
+   */
+  private static function enqueuePublicHolidayLeaveRequestUpdateTask() {
+    $task = new CRM_Queue_Task(
+      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllFuturePublicHolidayLeaveRequests', 'run'],
+      []
+    );
+
+    PublicHolidayLeaveRequestUpdatesQueue::createItem($task);
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -198,6 +198,10 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
       self::validateAddPublicHolidayToEntitlement($params);
     }
 
+    if(!empty($params['must_take_public_holiday_as_leave'])) {
+      self::validateMustTakePublicHolidayAsLeave($params);
+    }
+
     if (!empty($params['allow_request_cancelation']) &&
         !array_key_exists($params['allow_request_cancelation'], self::getRequestCancelationOptions())
     ) {
@@ -217,24 +221,65 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    * method checks if one such type already exists and throws an error if that
    * is the case.
    *
-   * @param array $params The params array received by the create method
+   * @param array $params
+   *  The params array received by the create method
    *
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
    */
   private static function validateAddPublicHolidayToEntitlement($params) {
+    if(!self::fieldIsUnique('add_public_holiday_to_entitlement', 1, $params)) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
+        'There is already one Absence Type where public holidays should be added to it'
+      );
+    }
+  }
+
+  /**
+   * Validates the must_take_public_holiday_as_leave field.
+   *
+   * There can be only one AbsenceType where this field is true. So this
+   * method checks if one such type already exists and throws an error if that
+   * is the case.
+   *
+   * @param array $params
+   *  The params array received by the create method
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
+   */
+  private static function validateMustTakePublicHolidayAsLeave($params) {
+    if(!self::fieldIsUnique('must_take_public_holiday_as_leave', 1, $params)) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
+        'There is already one Absence Type where "Must staff take public holiday as leave" is selected'
+      );
+    }
+  }
+
+  /**
+   * Checks if there's only one record with the given value for the given field.
+   *
+   * This method also considers the existence of an ID on the $params array when
+   * checking for uniqueness. When the ID is present, it searches for records
+   * with the given field and value, but with a different ID.
+   *
+   * @param string $fieldName
+   *  The field to be checked for uniqueness
+   * @param string $value
+   *  The value to be checked for uniqueness
+   * @param array $params
+   *  The params array passed to the create method
+   *
+   * @return bool
+   */
+  private static function fieldIsUnique($fieldName, $value, $params) {
     $dao = new self();
-    $dao->add_public_holiday_to_entitlement = 1;
+    $dao->$fieldName = $value;
 
     $id = empty($params['id']) ? null : intval($params['id']);
     if($id) {
       $dao->whereAdd("id <> {$id}");
     }
 
-    if($dao->count() > 0) {
-      throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
-          'There is already one Absence Type where "Must staff take public holiday as leave" is selected'
-      );
-    }
+    return $dao->count() == 0;
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -101,8 +101,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    *
    * @throws \CRM_HRLeaveAndAbsences_Exception_OperationNotAllowedException
    */
-  public static function del($id)
-  {
+  public static function del($id) {
     $absenceType = new CRM_HRLeaveAndAbsences_DAO_AbsenceType();
     $absenceType->id = $id;
     $absenceType->find(true);
@@ -112,6 +111,10 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
     }
 
     $absenceType->delete();
+
+    if($absenceType->must_take_public_holiday_as_leave) {
+      self::enqueuePublicHolidayLeaveRequestUpdateTask();
+    }
   }
 
   /**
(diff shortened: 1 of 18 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreation.php
```diff
@@ -109,4 +152,40 @@ private function zeroDeductionForOverlappingLeaveRequestDate(LeaveRequest $leave
...
+
+    return $result['values'];
+  }
+
+  /**
+   * Checks if the date of the given PublicHoliday overlaps the start and end
+   * dates of the given $contract
+   *
+   * @param array $contract
+   *  An contract as returned by the HRJobContract.getcontractswithdetailsinperiod API
```
omarabuhussein: two spaces

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -64,4 +67,38 @@ public function testItUpdatesTheBalanceChangeForOverlappingLeaveRequestDayToZero
 
     $this->assertEquals(0, LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest));
   }
+
+  public function testCanCreatePublicHolidayLeaveRequestsForAnAbsenceType() {
+    $contact = ContactFabricator::fabricate(['first_name' => 'Contact 1']);
```
omarabuhussein: you can fabricate a contact without passing any data since you didn't use the passed parameters

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreation.php
```diff
@@ -19,7 +19,7 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreation {
    */
   public function createForAbsenceType(AbsenceType $absenceType) {
     if(!$absenceType->must_take_public_holiday_as_leave) {
-      return;
+      throw new InvalidArgumentException("It's not possible to create Public Holidays for Absence Types where 'Must take public holiday as leave' is false");
```
omarabuhussein: use **ts()** function for the exception error message .

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -101,4 +102,17 @@ public function testCanCreatePublicHolidayLeaveRequestsForAnAbsenceType() {
 
     $this->assertEquals(-2, LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement));
   }
+
+  /**
+   * @expectedException InvalidArgumentException
+   * @expectedExceptionMessage It's not possible to create Public Holidays for Absence Types where 'Must take public holiday as leave' is false
```
omarabuhussein: if you added a ts to the exception above you may need to use **expectedExceptionCode** instead **expectedExceptionMessage** , but make sure to add a code when you through the exception

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -471,4 +471,25 @@ public static function getEnabledAbsenceTypes()
...
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsenceType|null
+   */
+  public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
+    $absenceType = new self();
+    $absenceType->must_take_public_holiday_as_leave = 1;
+    $absenceType->is_active = 1;
+
+    $absenceType->find();
+    if($absenceType->N == 1) {
```
omarabuhussein: why not :

```php
    if($absenceType->fetch()) {
      return $absenceType;
    }

  return null;
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -471,4 +471,25 @@ public static function getEnabledAbsenceTypes()
...
     return $absenceTypes;
   }
+
+  /**
+   * Returns the AbsenceType where the must_take_public_holiday_as_leave option
+   * is set
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsenceType|null
+   */
+  public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
```
omarabuhussein: the name is a bit odd to me .. maybe 
getPublicHolidayLeaveType

or maybe make it a generic one that return any LeaveType based on based passed parameters , not sure which is best here .

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreation.php
```diff
@@ -47,10 +47,8 @@ public function createForAbsenceType(AbsenceType $absenceType) {
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
    */
   public function createForContact($contactID, PublicHoliday $publicHoliday) {
-    $absenceTypes = $this->getAbsenceTypesWherePublicHolidaysMustBeTakenAsLeave();
-    foreach($absenceTypes as $absenceType) {
-      $this->create($contactID, $absenceType, $publicHoliday);
-    }
+    $absenceType = AbsenceType::getOneWithMustTakePublicHolidayAsLeaveRequest();
```
omarabuhussein: looking at it now , I think it is better to have a generic method to get absencType and based on a flag you get the one with "must take public holiday is leave" since it is a special leave type .

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
```diff
@@ -32,6 +32,26 @@ public function deleteForContact($contactID, PublicHoliday $publicHoliday) {
   }
 
   /**
+   * Deletes all the Public Holiday Leave Requests for Public Holidays in the
+   * future
+   */
+  public function deleteAllInTheFuture() {
+    $futurePublicHolidays = PublicHoliday::getAllInFuture();
+    $mostInTheFuturePublicHoliday = end($futurePublicHolidays);
```
omarabuhussein: $latestFuturePublicHoliday probably is more accurate name

omarabuhussein: $latestFuturePublicHoliday probably is easier to understand

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
```diff
@@ -78,6 +81,47 @@ public function testItUpdatesOverlappingLeaveRequestDatesAfterDeletingAPublicHol
     $this->assertEquals(-1, LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest));
   }
 
+  public function testCanDeleteAllPublicHolidayLeaveRequestsForFuturePublicHolidays() {
+    $contact = ContactFabricator::fabricate(['first_name' => 'Contact 1']);
```
omarabuhussein: no need to pass contact first_name

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequest.php
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Updates all the Leave Requests for Public Holidays in the future.
+   *
+   * Basically, this uses the Deletion Logic to delete all the leave requests
+   * for public holidays in the future and, then, uses the Creation Logic to
+   * create leave requests to public holidays in the future.
+   *
+   * @see CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion::deleteAllInTheFuture()
```
omarabuhussein: why not using the class alias instead of its full name

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -427,6 +434,46 @@ private static function getNotificationReceiversIDs($typeId) {
   }
 
   /**
+   * Checks, based on the $params array, if the Public Holiday Leave Requests
```
omarabuhussein: no need for the colon after checks

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
```diff
@@ -349,4 +350,32 @@ private static function checkIfDateOverlapsAnAbsencePeriod($params) {
   public static function getAllInFuture() {
     return self::getAllForPeriod(date('Ymd'));
   }
+
+  /**
+   * Process all the items on the PublicHolidayLeaveRequestUpdates Queue
+   *
+   * @return int
+   *  The number of items processed
```
omarabuhussein: two spaces

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
omarabuhussein: mmm .. not sure what else a suitable name would be ... okay no problem you can leave it as is

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: Why?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequest.php
davialexandre: I wish I could do that, but the alias works only for actual code, not for docblocks :(

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
davialexandre: Do you think so? If you take a look at this page https://en.oxforddictionaries.com/definition/late, you can see that latest, most of the time, is used to mean something like: "most recent". Examples:

_"The latest news"_
_"Residents say this is only the latest in a whole string of accidents at the same spot."_

How about `$lastPublicHoliday`?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
omarabuhussein: well , I Am not an English wizard :tophat:  but it sounds more correct to me without the colon .. but maybe I am wrong

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/PublicHoliday.php
```diff
@@ -339,4 +339,14 @@ private static function checkIfDateOverlapsAnAbsencePeriod($params) {
...
   }
+
+  /**
+   * Returns all the Public Holidays in the future. That is, all where date is
+   * >= today. Including Public Holidays that fall on weekends
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_PublicHoliday[]
+   */
+  public static function getAllInFuture() {
+    return self::getPublicHolidaysForPeriod(date('Ymd'));
```
davialexandre: It actually depends on the context. For this specific context, it has to be >=

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
davialexandre: To be honest, I'm not sure if it would be right to add a ts to an exception message. I think the translation responsibility should be somewhere else, not on the exception creation. Something like this:

```php
try {
//some code
} catch(Exception $e) {
   $errors[] = ts($e->getMessage())
}
```
Also, to use a code we need to spend some time creating a list of exception codes to make sure they're unique for each error message. That is not simple to do and might take some time.

Also, this is a headless test with just the default civi settings, so the language will always be english and adding a ts is unlikely to make the test fail.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: In English (and in Portuguese as well) we should use commas to separate nonessential phrases in a sentence. 

See that I can remove the "based on the $params array" entirely and the sentence still makes sense: "Checks if the Public Holiday Leave Requests must be updated after saving the Absence Type". So, since that part is nonessential, commas should be used (Just used them again here :D).

For reference, check rules 5 and 6 here: http://www.grammarbook.com/punctuation/commas.asp

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
omarabuhussein: mmm I see ... but chaning the expection message will ... I really don't like the idea of relying on exepction message I feel they are just wrong and codes are more robust ... but I am not gonna push on this since it will require a lot more work .

PR #1250, opened 2016-11-22
PCHR-1197: Trigger Public Holiday leave requests logic when updating contract

davialexandre:
When we update a contract, more specifically its details (HRJobDetails) we must also update the Public Holiday Leave Requests accordingly to the contract dates. The contract period might have changed and then we will need to add new leave requests for public holidays overlapping the new period.

### How this was implemented?
A new method was added to the PublicHolidayLeaveRequest service: `updateAllInTheFutureForContract()`. This method receives the ID of a contract and then:
1. Delete all leave requests for public holidays which are both in the future and overlaps the contract start and end dates
2. Creates leave requests for all public holidays which are both in the future and overlaps the contract start and end dates

Finally, using the `hook_civicrm_post()` hook, we call this new method every time a HRJobDetails is created or updated. The hook was used in order to tight couple the l&a extension with the hrjobcontract extension.

### What else has changed?
A new service, named JobContract, was created. It's kind of a wrapper around the HRJobContract API. Initially it was added as ways to remove some duplicated code among other services, but in the future it can be used to include other methods and be as a replacement in places where the civicrm_api3 is called directly, allowing us to mock API calls in tests.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/PublicHolidayLeaveRequestService.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/PublicHolidayLeaveRequestService.php
@@ -1,0 +1,28 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_JobContract as JobContractService;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequest as PublicHolidayLeaveRequestService;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreation as PublicHolidayLeaveRequestCreation;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion as PublicHolidayLeaveRequestDeletion;
+
+/**
+ * A factory for the PublicHolidayLeaveRequest service, which can be used
+ * to get instances of this service without having to manually create all of
+ * its dependencies
+ */
+class CRM_HRLeaveAndAbsences_Factory_PublicHolidayLeaveRequestService {
+
+  /**
+   * Returns a new instance of a PublicHolidayLeaveRequest Service
+   *
+   * @return \CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequest
+   */
+  public static function create() {
+    $jobContractService = new JobContractService();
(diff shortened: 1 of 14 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/PublicHolidayLeaveRequestService.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_JobContract as JobContractService;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequest as PublicHolidayLeaveRequestService;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreation as PublicHolidayLeaveRequestCreation;
+use CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion as PublicHolidayLeaveRequestDeletion;
+
+class CRM_HRLeaveAndAbsences_Factory_PublicHolidayLeaveRequestService {
```
MiyaNoctem: Could you add some documentation to the class? Like what it's for and how it's used?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/JobContract.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+class CRM_HRLeaveAndAbsences_Service_JobContract {
```
MiyaNoctem: Same here, documentation can really help us new comers get around code and understand what it's for.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -408,6 +410,48 @@ function hrleaveandabsences_civicrm_searchTasks($objectType, &$tasks) {
...
+ * HRJobDetails entity
+ *
+ * @param string $op
+ * @param int $objectId
+ * @param object $objectRef
+ */
+function _hrleaveandabsences_civicrm_post_hrjobdetails($op, $objectId, &$objectRef) {
+  if(in_array($op, ['create', 'edit'])) {
+
+    $revision = civicrm_api3('HRJobContractRevision', 'getsingle', [
```
MiyaNoctem: Maybe embed in a try /catch block to implement exception handling?

PR #1251, opened 2016-11-22
PCHR-1663: Changed Format of From Addresses

MiyaNoctem:
Implemented so e-mails sent on associated absence events have the format:
userName <SystemEMail>.  In that way, when the e-mail gets to users, they'll see tha name of the user that caused it to be sent, but the e-mail which it was sent from remains to be the one configured as the system-wide default sender.


coldrunKacper:
Approved.

PR #1256, opened 2016-11-25
PCHR-1123 Created API Endpoint:  LeavePeriodEntitlement.getBreakdown

tunbola:
**Created LeavePeriodEntitlement.getBreakdown API endpoint**

The API returns all of its leave balance changes for a LeavePeriodEntitlement when an entitlement_id is passed or leave balance changes for a set of LeavePeriodEntitlement when contact_id + period_id is passed to the API.

The API can be called from PHP like this with entitlement id: 

```php
civicrm_api3('LeavePeriodEntitlement', 'getbreakdown', array(
  'sequential' => 1,
  'entitlement_id' => 1,
));
```

**Sample Return values: **
```json
{
  "is_error": 0,
  "version": 3,
  "count": 1,
  "id": 0,
  "values": [
    {
      "id": "1",
      "breakdown": [
        {
          "amount": "3.50",
          "expiry_date": null,
          "type": {
            "id": 1,
            "value": "leave",
            "label": "Leave"
          }
        },
        {
          "amount": "4.00",
          "expiry_date": null,
          "type": {
            "id": 2,
            "value": "brought_forward",
            "label": "Brought Forward"
          }
        }
      ]
    }
  ]
}
```
With contact ID and Period ID:
```php
civicrm_api3('LeavePeriodEntitlement', 'getbreakdown', array(
  'sequential' => 1,
  'contact_id' => 109,
  'period_id' => 1,
));
```

**Sample return value:** 
```json
{
  "is_error": 0,
  "version": 3,
  "count": 3,
  "values": [
    {
      "id": "1",
      "breakdown": [
        {
          "amount": "3.50",
          "expiry_date": null,
          "type": {
            "id": 1,
            "value": "leave",
            "label": "Leave"
          }
        },
        {
          "amount": "4.00",
          "expiry_date": null,
          "type": {
            "id": 2,
            "value": "brought_forward",
            "label": "Brought Forward"
          }
        }
      ]
    },
    {
      "id": "2",
      "breakdown": [
        {
          "amount": "0.00",
          "expiry_date": null,
          "type": {
            "id": 1,
            "value": "leave",
            "label": "Leave"
          }
        }
      ]
    },
    {
      "id": "3",
      "breakdown": [
        {
          "amount": "0.00",
          "expiry_date": null,
          "type": {
            "id": 1,
            "value": "leave",
            "label": "Leave"
          }
        }
      ]
    }
  ]
}

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -140,7 +140,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
       $expiredBalanceWhereCondition = " AND expired_balance_change_id IS NULL";
     }
     if($returnExpiredOnly){
-      $expiredBalanceWhereCondition = " AND (expired_balance_change_id IS NOT NULL AND expiry_date < NOW())";
+      $expiredBalanceWhereCondition = " AND (expired_balance_change_id IS NOT NULL AND expiry_date < %1)";
     }
 
     $query = "
@@ -152,8 +152,11 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     ";
 
     $changes = [];
+    $params = [
+      1 => [date('Y-m-d'), 'String']
+    ];
 
-    $result = CRM_Core_DAO::executeQuery($query, [], true, self::class);
+    $result = CRM_Core_DAO::executeQuery($query, $params, true, self::class);
     while($result->fetch()) {
       $changes[] = clone $result;
     }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -127,6 +127,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
    * @param int $entitlementID
    *   The ID of the LeavePeriodEntitlement to get the Breakdown to
    * @param boolean $returnExpiredOnly
+   *   Passing true for this parameter will return only expired leave balance changes
+   *   Passing false will return only Non expired leave balance changes for the entitlement ID
    *
    * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
    */
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -123,12 +123,12 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
    * Forward and the Public Holidays. These are all balance changes, where the
    * source_id is the LeavePeriodEntitlement's ID and source_type is equal to
    * "entitlement", since they're are created during the entitlement calculation.
+   * Passing true for $returnExpiredOnly parameter will return only expired leave balance changes
+   * while Passing false will return only Non expired leave balance changes for the entitlement ID
    *
    * @param int $entitlementID
    *   The ID of the LeavePeriodEntitlement to get the Breakdown to
    * @param boolean $returnExpiredOnly
-   *   Passing true for this parameter will return only expired leave balance changes
-   *   Passing false will return only Non expired leave balance changes for the entitlement ID
    *
    * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
    */
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -129,6 +129,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
    * @param int $entitlementID
    *   The ID of the LeavePeriodEntitlement to get the Breakdown to
    * @param boolean $returnExpiredOnly
+   *   Whether to return Only Expired or Only Non Expired LeaveBalanceChanges
    *
    * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
    */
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php [hunk 1]
```diff
@@ -44,5 +44,42 @@ public function testGetRemainderWhenContactAndPeriodIdArePassed() {
...
+   * @expectedExceptionMessage You must include either the id of a specific entitlement, or both the contact and period id
+   */
+  public function testGetBreakdownWhenEntitlentIdAndPeriodIdArePassed() {
+    civicrm_api3('LeavePeriodEntitlement', 'getbreakdown', ['entitlement_id'=> 1, 'contact_id'=>1]);
+  }
+
+  public function testGetBreakdownWhenContactAndPeriodIdArePassed() {
+    $result = civicrm_api3('LeavePeriodEntitlement', 'getbreakdown', ['period_id'=> 1, 'contact_id'=>1]);
+    $this->assertEmpty($result['values']);
+  }
```
davialexandre: can you please add a test like this, but passing only the entitlement_id?

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -79,4 +79,38 @@ function _civicrm_api3_leave_period_entitlement_getremainder_spec(&$params) {
...
+ *
+ * @return array API result descriptor
+ *
+ * @throws API3_Exception
+ */
+function civicrm_api3_leave_period_entitlement_getbreakdown($params) {
+  $hasEntitlementID = !empty($params['entitlement_id']);
+  $hasContactAndPeriodID = !empty($params['contact_id']) && !empty($params['period_id']);
+  $hasContactOrPeriodID = !empty($params['contact_id']) || !empty($params['period_id']);
+  if(($hasEntitlementID && $hasContactAndPeriodID) || (!$hasEntitlementID && !$hasContactAndPeriodID) || ($hasEntitlementID && $hasContactOrPeriodID)) {
```
tunbola: Trying to catch the condition where entitlement_id is present and either of contact or period id is also present. This should not be so according to the ticket.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -839,6 +838,162 @@ public function testGetLeavePeriodEntitlementRemainderWithContactAndPeriodId() {
...
+    $periodEntitlement = LeavePeriodEntitlementFabricator::fabricate();
+    $this->createExpiredBroughtForwardBalanceChange($periodEntitlement->id, 9, 5);
+    $this->createExpiredBroughtForwardBalanceChange($periodEntitlement->id, 8, 3);
+    $this->createLeaveBalanceChange($periodEntitlement->id, 10);
+
+    $params = ['entitlement_id' => $periodEntitlement->id];
+    $result = LeavePeriodEntitlement::getBreakdown($params);
+    $this->assertCount(1, $result);
+    $this->assertArrayHasKey('id', $result[0]);
+    $this->assertEquals($periodEntitlement->id, $result[0]['id']);
```
tunbola: What if result does not have the id key by any chance?
Actually had that issue and it threw an error , thats why I added that.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -129,16 +129,22 @@ public static function getBalanceForEntitlement(LeavePeriodEntitlement $periodEn
    *
    * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
    */
-  public static function getBreakdownBalanceChangesForEntitlement($entitlementID) {
+  public static function getBreakdownBalanceChangesForEntitlement($entitlementID, $returnExpired=false) {
```
tunbola: I agree. Will make the change

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -573,4 +584,53 @@ public static function getRemainder($params){
...
+        'id' => $key,
+        'label' => $label,
+        'value' => CRM_Core_Pseudoconstant::getName('LeaveBalanceChange', 'type_id', $key)
+      ];
+    }
+
+    $results = [];
+    $returnExpired = !empty($params['expired']) ? true : false;
+    $i = 0;
+    foreach($leavePeriodEntitlements as $leavePeriodEntitlement){
```
davialexandre: For PhpStorm, check this help page: https://www.jetbrains.com/help/phpstorm/2016.2/code-style-php.html

Take a look at the "Set from" part. You'll need to set from Drupal.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
davialexandre: > This should not be so according to the ticket.
I'm not sure what does this mean ^

Anyway, I pulled your branch, changed the condition to as what I've suggested and the tests you added still pass. So I believe it will catch all the cases, unless your tests are not covering all of them :D

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
```diff
@@ -65,21 +65,28 @@ public function testOnGetBreakdownDoesNotAcceptPeriodIdWithoutContactId() {
...
+    $result = civicrm_api3('LeavePeriodEntitlement', 'getbreakdown', ['period_id' => 1, 'contact_id' => 1]);
     $this->assertEmpty($result['values']);
   }
 
-
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage Unable to find a CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement with id 1
+   */
+  public function testGetBreakdownWhenOnlyEntitlementIdIsPassed() {
```
davialexandre: This test is not good. The assertEmpty won't even be executed, because the api call will an exception, which will be caught by the @expectedException.

Please use the Fabricator to create a LeavePeriodEntitlement and remove the @expectedException and @expectedExceptionMessage

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -591,6 +591,9 @@ public static function getRemainder($params){
    * whether the expired parameter is true or false
    *
    * @param array $params
+   *   [
+   *    'entitlement_id' => 1
```
davialexandre: This description really doesn't help much. There are other values accepted here.

The purpose of the description is so that people can read it and understand what the function does and how to use it without having to read the code. Please keep that in mind when writing docblocks.

For this specific case, I think it would be better for the description to be something like this:

```php
@param array $param
  The param array passed to the LeavePeriodEntitlement.getBreakdown API endpoint.
  The supported values are:
  - entitlement_id: The id for a LeavePeriodEntitlement
  - contact_id: The id for a Contact
  - period_id: The id for a Absence Period
  - expired: A boolean flag. When it's true, only expired records will be returned. Otherwise, only non-expired 
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -647,4 +638,22 @@ public static function getBreakdown($params) {
     return $results;
   }
 
+  /**
+   * Returns LeaveBalanceChange Options for Type ID in a key-value pair
```
davialexandre: Just by reading the description of this method I can't understand what this method does. What is the key and what is the value for each pair?

PR #1258, opened 2016-11-29
PCHR-1251: My leave report html

dhavalsharma:
The html for my leave page as per PCHR-1251. 

The page will look like this 
image

A detailed description of breaking down this page into various components is described in [PR 117](https://github.com/compucorp/civihr-employee-portal-theme/pull/117)

This PR is created from [earlier PR](https://github.com/compucorp/civihr-employee-portal/pull/239) on employee portal.


AkA84:
There are several elements that still don't match (to various degrees) to the original design

### Paddings
paddings
padding-2

The paddings of the "main rows" of the table are visibly taller than what they're supposed to be

### Sub-tables headings height
heading-height
The height of the table headings is bigger than what it's supposed to be

## Custom dropdown
custom-dropdown-chrome
*(chrome)*
custom-dropdown-firefox
*(firefox)*

The dropdown looks completely broken in Chrome, while in Firefox it looks ok (except for the down-arrow not being vertically aligned) but there is a border overflowing to the bottom

Also, what's the reason for adding that `.chr-container__split-button` element in the `.chr_custom-select` component? And if you need it, given that it's definitely part of the `.chr_custom-select` component, why are you not calling it `.chr_custom-select__{something}`?

AkA84:
> PCHR-1251: Updated HTML to match design, add temporary report.html

You are not using the imperative mood

> PCHR-1251: Update HTML to match design, add temporary report.html

PR #1259, opened 2016-11-30
PCHR-1265: Add the LeaveRequest.getBalanceChangeByAbsenceType API endpoint

davialexandre:
This PR introduces the `LeaveRequest.getBalanceChangeByAbsenceType` API endpoint, which will return the total change in balance that is caused by the leave requests of a given absence type, or of all the absence types of a given contact and period.

This endpoint supports the following parameters:
- **contact_id**: Mandatory The ID of the Contact to get the balance change for
- **period_id**: Mandatory The ID of the Absence Period to get the balance change for
- **statuses**: Optional (Default: null) An array of OptionValue values which the list will be filtered by
- **public_holiday**: Optional (Default: false) Based on the value of this param, the calculation will include only the leave requests that aren't/are public holidays

The output is going to be a list of Absence Types ids followed by their respective total change in balance.

#### Examples
Say you have the following Leave Requests (all during the same absence period and for the same contact):

 type_id | status | public holiday | balance change
---------|------------------|----------------|----------------
 1 | Approved | false | -4
 1 | Cancelled | false | -1
 1 | Waiting Approval | false | -3
 2 | Admin Approved | true | -1
 2 | Rejected | false | -5

Calling this:
```php
civicrm_api3('LeaveRequest', 'getBalanceChangeByAbsenceType', [
  'contact_id' => 1,
  'period_id' => 1
]);
```

Would return this:
```javascript
{
   "1": -8, // including everything from absence type 1
   "2": -5 // including everything from absence type 2, except for the public holiday
}
```

Now, if we only want the public holidays, the API can be called like this:

```php 
civicrm_api3('LeaveRequest', 'getBalanceChangeByAbsenceType', [
  'contact_id' => 1,
  'period_id' => 1,
  'public_holiday' => true
]);
```

Which will return this:
```javascript
{
   "1": 0, // since there is no public holiday leave requests for absence type 1
   "2": -1
}
```

Using the `statuses` parameter, it's possible to return the balance change considering only Leave Requests with specific statuses:

```php
civicrm_api3('LeaveRequest', 'getBalanceChangeByAbsenceType', [
  'contact_id' => 1,
  'period_id' => 1,
  'statuses' => ['IN' => ['Cancelled', 'Rejected']]
]);
```

That would return:
```javascript
{
  "1": -1, //there only one cancelled leave request for absence type 1
  "2": -5 // there is rejected leave request for absence type 2
}
```

Note that, currently, `IN` is the only supported operator for the `statuses` parameter. That is enough for what we need at the moment and it would be extremely complex to support all the other operators. Trying to use any other operator with it will result in an error.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -91,19 +91,38 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * That is, the sum of all the Leave Balance Changes for Leave Requests of that
    * Absence Type, for the given $contactID during the given $periodID.
    *
+   * Balance Changes for Public Holiday Leave Requests won't be considered,
+   * except when $publicHolidays is true. In that case, on the balance changes
+   * for that type of request will be considered.
+   *
    * @param int $contactID
    * @param int $periodID
    * @param array $leaveRequestStatus
    *   And array of values from the Leave Request Status OptionGroup
+   * @param bool $publicHolidaysOnly
+   *   When true, will get the balance change only for the Public Holiday Leave Requests
    *
    * @return array
    */
-  public static function getBalanceChangeByAbsenceType($contactID, $periodID, $leaveRequestStatus = []) {
+  public static function getBalanceChangeByAbsenceType(
+    $contactID,
+    $periodID,
+    $leaveRequestStatus = [],
+    $publicHolidaysOnly = false
+  ) {
     $periodEntitlements = LeavePeriodEntitlement::getPeriodEntitlementsForContact($contactID, $periodID);
 
     $results = [];
+    $excludePublicHolidays = !$publicHolidaysOnly;
     foreach($periodEntitlements as $periodEntitlement) {
-      $balance = LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement, $leaveRequestStatus);
+      $balance = LeaveBalanceChange::getLeaveRequestBalanceForEntitlement(
+        $periodEntitlement,
+        $leaveRequestStatus,
+        null,
+        null,
+        $excludePublicHolidays,
+        $publicHolidaysOnly
+      );
       $results[$periodEntitlement->type_id] = $balance;
     }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -2,6 +2,7 @@
 
 use CRM_HRLeaveAndAbsences_BAO_PublicHoliday as PublicHoliday;
 use CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
+use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
 
 class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO_LeaveRequest {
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -86,6 +87,49 @@ public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $
...
+  public static function getBalanceChangeByAbsenceType(
+    $contactID,
+    $periodID,
+    $leaveRequestStatus = [],
+    $publicHolidays = false
+  ) {
+    $periodEntitlements = LeavePeriodEntitlement::getPeriodEntitlementsForContact($contactID, $periodID);
+
+    $results = [];
+    $excludePublicHolidays = $publicHolidays ? false : true;
```
tunbola: We can replace this with `$excludePublicHolidays = !($publicHolidays);`
Or what do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -443,6 +443,112 @@ public function testLeaveRequestBalanceForEntitlementCanSumBalanceChangesCreated
...
+      [],
+      null,
+      null,
+      false,
+      $includePublicHolidaysOnly
+    );
+    $this->assertEquals(-2, $balance);
+  }
+
+  public function testLeaveRequestBalanceForEntitlementWithParamsToBothExcludeAndIncludePublicHolidaysShouldReturnZero() {
```
tunbola: Been trying to wrap my end around why the balance change for this test is Zero, since a leave balance change that wasn't a public holiday was actually created.
Care to explain why? :)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -0,0 +1,79 @@
...
+      ['BETWEEN'],
+      ['NOT BETWEEN'],
+      ['LIKE'],
+      ['NOT LIKE'],
+      ['NOT IN'],
+      ['IS NULL'],
+      ['IS NOT NULL'],
+    ];
+  }
+}
```
tunbola: Do we need any API test where the public holiday parameter is passed either as true or false?

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -44,3 +44,88 @@ function civicrm_api3_leave_request_delete($params) {
...
+ */
+function civicrm_api3_leave_request_getbalancechangebyabsencetype($params) {
+  $statuses = _civicrm_api3_leave_request_get_statuses_from_params($params);
+
+  $values = CRM_HRLeaveAndAbsences_BAO_LeaveRequest::getBalanceChangeByAbsenceType(
+    $params['contact_id'],
+    $params['period_id'],
+    $statuses
+  );
+
```
tunbola: Not sure if I'm right, but don't we need to pass the public holiday parameter here too, if its present?

davialexandre: You're totally right! Nice catch. Thanks!

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: Do you mean checking that it only accepts a boolean value?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
davialexandre: It's because we're both excluding and including public holidays, which will result a q condition something like this:
```sql
WHERE leave_balance_change.type_id <> 'Public Holiday' AND leave_balance_change.type_id = 'Public Holiday'
```

There's no way for a balance change to be both a public holiday and not a public holiday, so nothing will be returned and the sum will be 0.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
davialexandre: Done. A bit difference though, as I didn't include the ()

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
tunbola: I see. Thanks for explaining.

PR #1263, opened 2016-12-01
PCHR-1663: Loaded Domain Sender E-mail from Database

MiyaNoctem:
Previous solution worked when using gmail as SMTP server, because gmail automatically replaces original "From" address for account being used to send the e-mail. Failed when testing aginst other SMTP servers
(sender should always be system-wide email). So, used call to CRM core Domain Entity to obtain default
system sender.

Tested on staging site by checking out branch directly. All tests were succesful, except test on line 38, which requires clarification on how to replicate, and test 39, which failed.

Results for tests can be seen here:
https://docs.google.com/a/compucorp.co.uk/spreadsheets/d/1n4Uovt9URyErQ2pMqt-hTGUBb2ezCVLR0J017xk_eDI/edit?usp=sharing

### DIFF
--- a/hrabsence/CRM/HRAbsence/Form/AbsenceRequest.php
+++ b/hrabsence/CRM/HRAbsence/Form/AbsenceRequest.php
@@ -561,6 +561,7 @@ class CRM_HRAbsence_Form_AbsenceRequest extends CRM_Core_Form {
     ));
     $mailprm[$this->_targetContactID]['display_name'] = $targetContactResult['values'][$this->_targetContactID]['display_name'];
     $mailprm[$this->_targetContactID]['email'] = $targetContactResult['values'][$this->_targetContactID]['email'];
+    $domainEmailAddress = CRM_Core_BAO_Domain::getNameAndEmail()[1];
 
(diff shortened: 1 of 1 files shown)


hrabsence/CRM/HRAbsence/Form/AbsenceRequest.php
```diff
@@ -561,6 +561,7 @@ public function postProcess() {
     ));
     $mailprm[$this->_targetContactID]['display_name'] = $targetContactResult['values'][$this->_targetContactID]['display_name'];
     $mailprm[$this->_targetContactID]['email'] = $targetContactResult['values'][$this->_targetContactID]['email'];
+    $domainEmailAddress = CRM_Core_BAO_Domain::getNameAndEmail()[1];
```
omarabuhussein: you can use the APi for this :

```php
$result = civicrm_api3('Domain', 'get', array(
  'sequential' => 1,
  'current_domain' => 1,
));

and you will get the from_email ... you can test it on API explorer to ensure it gives you the same result.
```

PR #1268, opened 2016-12-05
PCHR-1756: Filter out leave requests not overlapping contracts

davialexandre:
A Leave Request is not linked directly to a Contract. Instead it's linked to a Contact and we use the Leave Request dates in order to know with which contracts it overlaps. This means we can have "stale" requests, which don't overlap any existing contract (this can happen, for example, when a contract length is decreased).

A call to the LeaveRequest.get API endpoint, even without any params, should no include those "stale" requests, so this PR updates the query used by the API to filter out any Leave Request not overlapping a contract. This was implemented by updating the query to join Leave Request with Contracts using the contact_id. Next, we join the Contract with its details, through the Contract Revision, in order to finally be able to add the conditions to return only the Leave Requests where the from_date and to_date overlaps the Contract's period_start_date and period_end_date.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -1,9 +1,12 @@
 <?php
 
-use \Civi\API\SelectQuery;
-use \CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
-use \CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
-use \CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
+use Civi\API\SelectQuery;
+use CRM_Hrjobcontract_BAO_HRJobContract as HRJobContract;
+use CRM_Hrjobcontract_BAO_HRJobDetails as HRJobDetails;
+use CRM_Hrjobcontract_BAO_HRJobContractRevision as HRJobContractRevision;
+use CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
 
 /**
  * This class is basically a wrapper around Civi\API\SelectQuery.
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -277,6 +287,177 @@ public function testGetReturnsOnlyPublicHolidayLeaveRequestsIfThePublicHolidayIs
...
+    //This leave request matches the date params, but not the contract dates,
+    //so it will not be returned
+    LeaveRequestFabricator::fabricate([
+      'contact_id' => 1,
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2015-12-30'),
+      'status_id' => $leaveRequestStatuses['Cancelled']
+    ], true);
+
+    $result = civicrm_api3('LeaveRequest', 'get', ['to_date' => '2015-12-30']);
```
tunbola: Should this rather not be `'from_date' => '2015-12-30` ?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -277,6 +287,177 @@ public function testGetReturnsOnlyPublicHolidayLeaveRequestsIfThePublicHolidayIs
...
+      'to_date' => CRM_Utils_Date::processDate('2017-12-31'),
+      'status_id' => $leaveRequestStatuses['Waiting Approval']
+    ], true);
+
+    $result = civicrm_api3('LeaveRequest', 'get');
+    $this->assertCount(1, $result['values']);
+    $this->assertNotEmpty($result['values'][$leaveRequest2->id]);
+  }
+
+  public function testGetCanReturnALeaveRequestWhichOverlapsAContractWithEndDate() {
```
tunbola: Can we also have a test that returns multiple leave requests overlapping a contract rather than just returning a single leave request that overlaps the contract dates?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -42,15 +45,81 @@ public function __construct($params) {
...
+      'jcr',
+      [
+        'jcr.id = (SELECT id
+                    FROM ' . HRJobContractRevision::getTableName() . ' jcr2
+                    WHERE
+                    jcr2.jobcontract_id = jc.id
+                    ORDER BY jcr2.effective_date DESC
+                    LIMIT 1)'
+      ]
+    );
```
davialexandre: I'm not sure, but that could work. The problem is that the data structure of hrjobcontract is quite complex and it might not be so simple to change this. Also, the extension almost has no tests, so there's easy way to see if we'll break anything with such a change.
HRJobContract will be revisited/refactored some day and luckily we won't need those sub queries anymore. For now, we have to live with it :(

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: fixed

PR #1269, opened 2016-12-06
PCHR-1751: Fix Filtering of Contacts by Value of Region/Location on Search

MiyaNoctem:
When a local admin user, configured to have access to specific region(s)/location(s) tried to search for contacts, the resut included all contacts in the system, when it should only show contacts with current job roles in the region(s)/location(s) administered by the local admin.

The search filter was trying to use option_value ID to filter job roles, when option_value value was being stored in roles table.  Thus, they never matched.  This also caused a type mismatch error when optionvalue.value was a string (like "Headquarters") and not an integer, which caused an exception that made the filter ignore ACL $where and $wheretables being built in contactaccessrights extension, thus performing search on all contacts. 

Fixed by adding option_value.value to query to obtain access rights for current user and imploded
resulting values as strings, instead of integers, when building inner join to job roles table.

Problem was probably introduced in https://github.com/civicrm/civihr/pull/955, when Regions and Locations were changed to be pseudoconstants to be stored in job roles table.

### DIFF
--- a/contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
+++ b/contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * Fabricates rights for tests.
+ */
+class CRM_Contactaccessrights_Test_Fabricator_Rights {
+
+  /**
+   * Fabricates Rights, given specific values for contact_id, entity_id and 
+   * entity_type.
+   * 
+   * @param array $params
+   *   Associative array of parameters to be used to store rights information in 
+   *   database.
+   * @return array
+   *   Holds values stored in database for created access right
+   */
+  public static function fabricate($params) {
+    $result = civicrm_api3('Rights', 'create', [
+      'sequential' => 1,
+      'contact_id' => $params['contact_id'],
+      'entity_id' => $params['entity_id'],
+      'entity_type' => $params['entity_type']
+    ]);
+
+    return array_shift($result['values']);
+  }
+
+}
--- a/contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
+++ b/contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
@@ -1,0 +1,116 @@
+<?php
+
+use Civi\Test\HeadlessInterface;
+use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+use CRM_HRCore_Test_Fabricator_OptionValue as OptionValueFabricator;
+use CRM_Contactaccessrights_Test_Fabricator_Rights as RightsFabricator;
+
+/**
+ * Class CRM_Contactaccessrights_BAO_RightsTest
+ * Tests Rights BAO class for contact's access rights to civicrm.
+ * 
+ * @group headless
+ */
+class CRM_Contactaccessrights_Utils_ACLTest extends PHPUnit_Framework_TestCase implements
+  HeadlessInterface, TransactionalInterface {
+
+  /**
+   * Installs extensions required for test
+   * 
+   * @return \Civi\Test\CiviEnvBuilder
+   */
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
+      ->install('org.civicrm.hrjobcontract')
+      ->installMe(__DIR__)
+      ->apply();
+  }
+
+  /**
+   * Tests if ACL Utility class is building the Job Roles Clause appropriately,
+   * using option_value.value for Locations and Regions in the table join.
+   */
+  public function testJobRolesClauseWhereACL() {
+    // Setup local admin
+    $localAdmin = ContactFabricator::fabricate();
+
+    // Create Regions and Locations
+    $region1 = $this->createOptionValue('hrjc_region', 'Region 1');
+    $region2 = $this->createOptionValue('hrjc_region', 'Region 2');
+    $location1 = $this->createOptionValue('hrjc_location', 'Location 1');
+    $location2 = $this->createOptionValue('hrjc_location', 'Location 2');
+
+    // Add Regions / Locations to localAdmin's Rights
+    $this->setContactRights($localAdmin, $region1);
+    $this->setContactRights($localAdmin, $location1);
+    $this->setContactRights($localAdmin, $location2);
+
+    // Get Job Roles Clause
+    $aclUtil = new CRM_Contactaccessrights_Utils_ACL($localAdmin['id']);
+    $whereTables = $aclUtil->getWhereTables();
+    $jobsClause = $whereTables['car_3_jr'];
+
+    // Check Locations Clause
+    $locationMatches = [];
+    preg_match("/car_jr.location IN \(('[^']*'(, ?'[^']*')*)\)/", $jobsClause, $locationMatches);
+    $this->assertTrue(isset($locationMatches[1]), 'Jobs clause does not have expected format for locations.');
+
+    $locations = array_map('trim', explode(',', $locationMatches[1]));
+    $this->assertContains("'{$location1['value']}'", $locations, 'Expected value in clause for locations is not found.');
+    $this->assertContains("'{$location2['value']}'", $locations, 'Expected value in clause for locations is not found.');
+
+    // Check Regions Clause
+    $regionMatches = [];
+    preg_match("/car_jr.region IN \(('[^']*'(, ?'[^']*')*)\)/", $jobsClause, $regionMatches);
+    $this->assertTrue(isset($regionMatches[1]), 'Jobs clause does not have expected format for Regions.');
+
+    $regions = array_map('trim', explode(',', $regionMatches[1]));
+    $this->assertContains("'{$region1['value']}'", $regions);
+    $this->assertNotContains("'{$region2['value']}'", $regions);
+  }
+
+  /**
+   * Creates a new record for Rights entity for the given contact.
+   * 
+   * @param array $contact
+   *   Associative array with details for the contact
+   * @param array $entity
+   *   Details of either Region or Location for which the contact will be 
+   *   granted access
+   * @return array
+   *   Details of access right created in an associative array
+   */
+  private function setContactRights($contact, $entity) {
+    return RightsFabricator::fabricate([
+      'contact_id' => $contact['id'],
+      'entity_id' => $entity['id'],
+      'entity_type' => $entity['entity_type']
+    ]);
+  }
+
+  /**
+   * Creates a Record in Option Value for given Option Group
+   * 
+   * @param string $optionGroup
+   *   Name of the option group to which the option will be added
+   * @param string $value
+   *   Value for the new option value, used as its name, label and value
+   * @return array
+   *   Details of added option value + entity_type attribute
+   */
+  private function createOptionValue($optionGroup, $value) {
+    $params = [
+      'option_group_id' => $optionGroup,
+      'name' => $value,
+      'label' => $value,
+      'value' => $value
+    ];
+    $result = OptionValueFabricator::fabricate($params);
+    $result['entity_type'] = $optionGroup;
+    
+    return $result;
+  }
+
+}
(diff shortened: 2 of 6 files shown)


contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * Fabricates rights for tests.
+ *
+ * @author Camilo Rodriguez - camilo@compucorp.co.uk
+ */
+abstract class CRM_Contactaccessrights_Test_Fabricator_Rights {
```
davialexandre: Why abstract?

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * Fabricates rights for tests.
+ *
+ * @author Camilo Rodriguez - camilo@compucorp.co.uk
```
davialexandre: No need for the @author tag

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Fabricates rights for tests.
+ *
+ * @author Camilo Rodriguez - camilo@compucorp.co.uk
+ */
+abstract class CRM_Contactaccessrights_Test_Fabricator_Rights {
+	
+	/**
+	 * Fabricates Rights, given specific values forcontact_id, entity_id and
```
davialexandre: missing space between for and contact_id

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
...
+	
+	/**
+	 * Fabricates Rights, given specific values forcontact_id, entity_id and 
+	 * entity_type.
+	 * 
+	 * @param array $params
+	 *   Associative array of parameters to be used to store rights infrmation in 
+	 *   database.
+	 * @return array
+	 *   Holds values stored in datbase
```
davialexandre: s/datbase/database/

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @throws Exception
+	 * @throws CiviCRM_API3_Exception
+	 */
+	public static function fabricate($params) {
+    if (!isset($params['contact_id']) || !isset($params['entity_id']) 
+			|| !isset($params['entity_type'])) {
+      throw new Exception('Specify contact_id, entity_id and entity_type [' . json_encode($params) . ']');
+    }
```
davialexandre: I believe there's no need for this. Fabricators should not have validation

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,41 @@
...
+
+		try {
+			$result = civicrm_api3('Rights', 'create', [
+				'sequential' => 1,
+				'contact_id' => $params['contact_id'],
+				'entity_id' => $params['entity_id'],
+				'entity_type' => $params['entity_type']
+			]);
+		} catch (CiviCRM_API3_Exception $e) {
+			throw new Exception("Error creating Rights! Params: " . serialize($params), 0, $e);
```
davialexandre: I don't think we need such a specific exception here. Fabricator will only be used in tests. If a CiviCRM_API3_Exception is thrown, that specific test will fail and then it will be easy to figure out where was the fail and which param were passed to the API

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobContract as JobContractFabricator;
+use CRM_Contactaccessrights_Test_Fabricator_Rights as RightsFabricator;
+
+/**
+ * Class CRM_Contactaccessrights_BAO_RightsTest
+ * Tests Rights BAO class for contact's access rights to civicrm.
+ * 
+ * @group headless
+ * @author Camilo Rodriguez - camilo@compucorp.co.uk
```
davialexandre: No need for the @author tag here as well (If these are being added by your editor, you can configure it to not add them automatically)

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+	/**
+	 * Tests if ACL Utility class is building the Job Roles Clause appropriately,
+	 * using option_value.value for Locations and Regions in the table join.
+	 */
+  public function testJobRolesClauseWhereACL() {
+    // Setup local admin
+    $localAdmin = ContactFabricator::fabricate();
+		
+		// Create Regions and Locations
+		$region1 = $this->createOpValue('hrjc_region', 'Region 1');
```
davialexandre: There are fabricators for option values and option groups. You should use them instead

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+      ->installMe(__DIR__)
+      ->apply();
+  }
+  
+	/**
+	 * Tests if ACL Utility class is building the Job Roles Clause appropriately,
+	 * using option_value.value for Locations and Regions in the table join.
+	 */
+  public function testJobRolesClauseWhereACL() {
+    // Setup local admin
```
davialexandre: Most of the code on this method seems to be indented with only 1 space. Can you please check that and make sure to always use 2 spaces?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+		$this->setContactRights($localAdmin, $location1);
+		$this->setContactRights($localAdmin, $location2);
+		
+    // Get Job Roles Clause
+    $aclUtil = new CRM_Contactaccessrights_Utils_ACL($localAdmin['id']);
+		$whereTables = $aclUtil->getWhereTables();
+		$jobsClause = $whereTables['car_3_jr'];
+		
+		// Check Locations Clause
+		$locationMatches = array();
```
davialexandre: Please always use short array syntax

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+	 * Creates a Record in Option Value for given Option Group
+	 * 
+	 * @param string $opgroup
+	 *   Name of the option group to which the option will be added
+	 * @param string $value
+	 *   Value for the new option value, used as its name, label and value
+	 * @return array
+	 *   Details of added option value
+	 */
+	private function createOpValue($opgroup, $value) {
```
davialexandre: Just a tip, since you'll use the fabricator instead of this helper method, you don't really need to change this: avoid abbreviations always. Explicit is always better than implicit. Op can also be the abbreviation for Operations, for example. Saving 4 letters is not a good reason to justify someone getting confused with the method name.

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,118 @@
...
+	 *   Associative array with details for the contact
+	 * @param array $entity
+	 *   Details of either Region or Location for which the contact will be 
+	 *   granted access
+	 * @return array
+	 *   Details of access right created in an associative array
+	 */
+	private function setContactRights($contact, $entity) {
+		return RightsFabricator::fabricate([
+				'contact_id' => $contact['id'],
```
davialexandre: You're using 4 spaces here. Please always use 2

davialexandre:
Oh, I just noticed you didn't fix the PR title as per the first review. Please fix that

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
MiyaNoctem: Well, it's a class that will never be instantiated, and shouldn't be.  That, and most of the fabricators I checked were abstract classes also.  It just made sense to me.

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,35 @@
...
+  public static function fabricate($params) {
+    try {
+      $result = civicrm_api3('Rights', 'create', [
+        'sequential' => 1,
+        'contact_id' => $params['contact_id'],
+        'entity_id' => $params['entity_id'],
+        'entity_type' => $params['entity_type']
+      ]);
+    } catch (CiviCRM_API3_Exception $e) {
+      throw $e;
```
davialexandre: This just doesn't make sense. If you're just throwing the catched exception you don't need the try catch block at all

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Fabricates Rights, given specific values for contact_id, entity_id and 
+   * entity_type.
+   * 
+   * @param array $params
+   *   Associative array of parameters to be used to store rights infrmation in 
+   *   database.
+   * @return array
+   *   Holds values stored in database for created acess right
+   * @throws Exception
```
davialexandre: Since this method now only throws CiviCRM_API3_Exceptions, you can remove this @throws

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -0,0 +1,116 @@
...
+   * Creates a Record in Option Value for given Option Group
+   * 
+   * @param string $opgroup
+   *   Name of the option group to which the option will be added
+   * @param string $value
+   *   Value for the new option value, used as its name, label and value
+   * @return array
+   *   Details of added option value + entity_type attribute
+   */
+  private function createOptionValue($opgroup, $value) {
```
davialexandre: you're still using the abbreviation for $opgroup

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,35 @@
...
+ * Fabricates rights for tests.
+ */
+abstract class CRM_Contactaccessrights_Test_Fabricator_Rights {
+
+  /**
+   * Fabricates Rights, given specific values for contact_id, entity_id and 
+   * entity_type.
+   * 
+   * @param array $params
+   *   Associative array of parameters to be used to store rights infrmation in
```
davialexandre: you forgot to fix the typo in "information"

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Fabricates Rights, given specific values for contact_id, entity_id and 
+   * entity_type.
+   * 
+   * @param array $params
+   *   Associative array of parameters to be used to store rights infrmation in 
+   *   database.
+   * @return array
+   *   Holds values stored in database for created acess right
```
davialexandre: s/acess/access/

contactaccessrights/CRM/Contactaccessrights/Test/Fabricator/Rights.php
davialexandre: Sorry, but none of the existing fabricators are abstract classes. Only [BaseAPIFabricator](https://github.com/civicrm/civihr/blob/staging/hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/BaseAPIFabricator.php) is abstract, but that's because it's just a template for the other fabricators.
Using abstract to not allow a class to be instantiated is conceptually wrong. That is not what abstract classes are for

PR #1271, opened 2016-12-07
PCHR-1369: Add the LeaveRequest.getFull API endpoint

davialexandre:
This new API endpoint is an "extension" of the LeaveRequest.get and, in being so, it works exactly the same way and accepts the same params and options. The difference is on the returned values, where each Leave Request will include its balance and dates.

To call it, use:
```php
civicrm_api3('LeaveRequest', 'getFull');
```

And the response would be something like:
```json
{

    "is_error": 0,
    "version": 3,
    "count": 2,
    "values": [
        {
            "id": "17",
            "type_id": "1",
            "contact_id": "202",
            "status_id": "1",
            "from_date": "2016-02-01",
            "from_date_type": "1",
            "to_date": "2016-02-03",
            "to_date_type": "1",
            "balance_change": -3,
            "dates": [
                {
                    "id": "17",
                    "date": "2016-02-01"
                },
                {
                    "id": "18",
                    "date": "2016-02-02"
                },
                {
                    "id": "19",
                    "date": "2016-02-03"
                }
            ]
        },
        {
            "id": "19",
            "type_id": "1",
            "contact_id": "202",
            "status_id": "6",
            "from_date": "2016-01-30",
            "from_date_type": "1",
            "to_date": "2016-02-01",
            "to_date_type": "1",
            "balance_change": -3,
            "dates": [
                {
                    "id": "29",
                    "date": "2016-01-30"
                },
                {
                    "id": "30",
                    "date": "2016-01-31"
                },
                {
                    "id": "31",
                    "date": "2016-02-01"
                }
            ]
        }
   ]
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -30,6 +30,13 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
    */
   private $query;
 
+  /**
+   * @var bool
+   *   A flag indicating if the returned Leave Requests will also include the
+   *   balance change and the Leave Request dates
+   */
+  private $returnFullDetails = false;
+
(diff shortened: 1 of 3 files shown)


coldrunKacper:
Approved.

PR #1273, opened 2016-12-07
PCHR-1760: Replace class constants in WorkDay BAO with an OptionGroup

davialexandre:
The valid values for the Work Day type were hardcoded as a set of constants on the WorkDay BAO. This would make it harder to change the option labels, add or remove new options and to return them on API responses in a consistent way. So, in order to avoid those problems, they were replaced with an option group named "Work Day Type".

The option group options are:

| label   | value | name            |
|---------|-------|-----------------|
| No      | 1     | non_working_day |
| Yes     | 2     | working_day     |
| Weekend | 3     | weekend         |

Every call to one of the removed constants was replaced with a call to a method which returns the respective option value value:
- `WorkDay::WORK_DAY_OPTION_NO` was replaced by `WorkDay::getNonWorkingDayTypeValue()`
- `WorkDay::WORK_DAY_OPTION_YES` was replaced by `WorkDay::getWorkingDayTypeValue()`
- `WorkDay::WORK_DAY_OPTION_WEEKEND` was replaced by `WorkDay::getWeekendDayTypeValue()`

Because an Option Value value is a string, I changed the data type of the WorkDay type field to also be a string, with the same size of the Option Value value.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
@@ -1,5 +1,7 @@
 <?php
 
+use CRM_HRLeaveAndAbsences_Exception_InvalidWorkDayException as InvalidWorkDayException;
+
 class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_WorkDay {
 
   /**
@@ -17,7 +19,6 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
    *
    */
   public static function create($params) {
-    $className = 'CRM_HRLeaveAndAbsences_DAO_WorkDay';
     $entityName = 'WorkDay';
     $hook = empty($params['id']) ? 'create' : 'edit';
     self::validateParams($params);
@@ -36,7 +37,7 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
     }
 
     CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
-    $instance = new $className();
+    $instance = new self();
     $instance->copyValues($params);
     $instance->save();
     CRM_Utils_Hook::post($hook, $entityName, $instance->id, $instance);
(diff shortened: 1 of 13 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
```diff
@@ -42,6 +45,56 @@ public static function create($params) {
...
+      self::$workDayTypes = array_flip(self::buildOptions('type', 'validate'));
+
+      // The option value values are stored as strings on the database, so we
+      // need to make sure the values will always be returned as strings, even
+      // if they're numeric.
+      array_walk(self::$workDayTypes, function(&$type) {
+        $type = (string)$type;
+      });
+    }
+
```
tunbola: Do we really need to cast to string since the values are stored as strings in the db and I presume they will be retrieved as such?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkWeek.php
```diff
@@ -95,7 +95,7 @@ private function saveDays($days) {
         'type' => $day['type']
       ];
 
-      if($day['type'] == CRM_HRLeaveAndAbsences_BAO_WorkDay::WORK_DAY_OPTION_YES) {
+      if($day['type'] == CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()) {
```
tunbola: Why don't we use an aliased class name here `if($day['type'] == WorkDay::WORK_DAY_OPTION_YES)` ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/WorkPattern.php
```diff
@@ -48,13 +48,13 @@ public function setDefaultValues() {
...
-                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::WORK_DAY_OPTION_YES],
-                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::WORK_DAY_OPTION_WEEKEND],
-                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::WORK_DAY_OPTION_WEEKEND],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWeekendTypeValue()],
+                  ['type' => CRM_HRLeaveAndAbsences_BAO_WorkDay::getWeekendTypeValue()],
```
tunbola: Can we use aliased class name here also?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/WorkPattern.php
```diff
@@ -269,7 +269,7 @@ private function validateWorkDays($values, &$errors)
      */
     private function validateWorkDay($weekIndex, $dayIndex, $day, &$errors)
     {
-      if($day['type'] == CRM_HRLeaveAndAbsences_BAO_WorkDay::WORK_DAY_OPTION_YES) {
+      if($day['type'] == CRM_HRLeaveAndAbsences_BAO_WorkDay::getWorkingDayTypeValue()) {
```
tunbola: Aliased class name here also

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkDayTest.php
```diff
@@ -168,10 +168,22 @@ public function testNumberOfHoursShouldBeCalculatedFromTimesAndBreak($timeFrom,
     $this->assertEquals($expectedNumberOfHours, $entity->number_of_hours);
   }
 
+  public function testGetNonWorkingDayTypeValueReturnsTheOptionValueValueAsString() {
+    $this->assertSame('1', WorkDay::getNonWorkingDayTypeValue());
+  }
```
tunbola: Rather than assert against a value can we do something like
`$this->assertSame($this->WorkDayTypes['Weekend']['id'], WorkDay::getNonWorkingDayTypeValue());` ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
davialexandre: We need the cast. The problem is that buildOptions will use the option value value as the key on the returned array and PHP will convert an integer string to an actual int if it is used as a key: https://3v4l.org/vvoan

PR #1276, opened 2016-12-09
PCHR-459: Set up built-in assignments in shipped version

omarabuhussein:
Related PR :  https://github.com/compucorp/civihr-tasks-assignments/pull/131

## Problem

When installing hrcase the following case types have to be created (please check the wiki page) :

https://compucorp.atlassian.net/wiki/pages/viewpage.action?pageId=75857955

these activity types for these case types should have component ID = CiviTask and not the default CiviCase.

## Solution

First I've created these activity types via **hook_civicrm_managed** and defined the case and activity types inside **ManagedCaseAndActivityTypes.mgd.php** ( hook_civicrm_managed by default parse any file ends up with *.mgd and create any managed entity defined in these files) . I also created **CRM_HRCase_DefaultCaseAndActivityTypes** class which define these case and activity types so they are easily accessible anywhere inside the extension without the need to redefine them .

Using this hook will allow civicrm to handle creation of these case types and related activity types automatically via managed entities handler , And upgrader is also added to handle existing civihr installations.

Apart from this , hrcase is an old extension and contain some old logic that converts CiviCase component to be used as a component for Task & Assignments , but we have a separate
extension for that which is called ( Task & Assignments) which is also created on top of CiviCase component, But keeping both of these extensions in their current status may cause some problems later so I have cleaned all the unnecessary logic from hrcase ( though there are some parts that I didn't touch yet to avoid breaking any thing for now ).

At also worth mentioning that there is still some old logic inside some implemented hooks inside this extension like hrcase_civicrm_buildForm and hrcase_civicrm_Post
but they are leaved for now until the relationship between hrcase & hreccuriment extension is more clear since it is appear that they  still have some connection.

Also Apart from cleaning some of the old logic I have refactored the code and moved a lot of what it does inside hrcase.php to the upgrader class in more organized way and
forced the upgraders to run at the extension installation to follow other civihr extensions.


coldrunKacper:
Approved.

coldrunKacper:
Approved.

coldrunKacper:
Approved.

PR #1279, opened 2016-12-09
PCHR-1776: Fix Control on Contact Selection for Adding Funders to Existing Role

MiyaNoctem:
When editing a job role, the control to select the contact that would fund the role would fail to search on all contacts in system as the name was being typed. However, on role creation, adding funders would use a different control, which worked as expected, searching all available contacts as you type the name.  Fixed by implementing the same control used in add_new_role.html on job_role_panel.html

# Update 01/02
## Problem
When adding or amending a job roles' funder, the user is presented with a lookup dropdown that comes pre-filled with the first 25 contacts. If any of those 25 contact is selected as funder, the data is saved correctly. But if the user looks up (as in, makes ajax requests via the dropdown for additional contacts) a different contact and select it to be the funder, then it appears that the data is not saved correctly, as the new funder does not get listed among the old ones

before

The problem wasn't that the new funder wasn't saved (it actually was), but that it wasn't being displayed on the list.

The reason for that is that, in order to show the name of the funder, the pool of contacts pre-fetched by the lookup dropdown was used. And given that the pool contains only the first 25 contacts, if a funder happened to be linked to a contact outside of the pool, no match would be found, resulting in the funder not being listed
```js
// Set default funder rows funder rows
$scope.edit_data[role_id]['funders'].push({
  funder_id: {
    id: funder_contact_ids[i],
    // !! `job_roles.contactListObject` contains only the first 25 contacts !!
    sort_name: job_roles.contactListObject[funder_contact_ids[i]]['sort_name']
  },
  // ...
});
```

## Solution
The solution is to make sure to always fetch the contacts of all the funders of each job roles, and keep that list separated from the "volatile" list of the lookup dropdown
```js
// Right after fetching the job roles...
.then(function () {
  return getFundersContacts();
})
.then(function (contacts) {
  fundersContacts = contacts;
})

// ...then when initializing the funder list data of a given job role
jobRole.funders.push({
  // ...
  funder_id: {
    id: funder_contact_ids[i],
    // !! now using a separate list !!
    sort_name: fundersContacts[funder_contact_ids[i]].sort_name
  }
});
```
This internal list is also updated every time a new funder is added and saved
```js
$scope.saveNewRole = function () {
  // ..
 
  if (newRole.funders.length) {
    updateFundersContactsList(newRole.funders);
  }
});

$scope.updateRole = function (role_id, role_type) {
  // ...

  if (updatedRole.funders.length) {
    updateFundersContactsList(updatedRole.funders);
  }
}
```
### Refactoring
The controller was in an awful state. The code is bloated, all over the place, undocumented, with a massive (and redundant) interface. In order to start working on it, I had at least to reorder the variables and methods and functions, standardize the documentation, and split in smaller functions the `initData` scope method

### Augment `HRJobRolesService.getContactList()`
The method accepted only a string for searching contacts by partial name. Now it also accepts a list of contact ids (so that we can gather all the funder ids of the job roles, and then get their correspondent contacts)
```js
getContactList: function (sortName, idsList) {
  var deferred = $q.defer();

  CRM.api3('Contact', 'get', {
    "sequential": 1,
    "return": "id, sort_name",
    "id": _.isArray(idsList) ? { 'IN': idsList } : null, // <--
    "sort_name": sortName || null
  })
  // ...
}
```

### Mocked data
The mocked data has now been moved into the `mock/data` folder. Also the job role mocked data didn't represent the data as it is being returned directly from the API, but rather how it is after it has been processed by the controller. Fixing that meant breaking a lot of tests with no easy solution at end, so I ended up adding a different set of mocked job roles, this one representing the data as it given by the API

### Remove race condition
There was a pre-existing race condition between the request for fetching that initial pool of 25 names and the controller trying to display the funder names on the template.

Just by chance the first request always happened later, but given that the data for displaying the funder names now is coming from a different path (job contract -> job roles -> get funders ids -> fetch contacts matching with those ids) which is slower, the race condition became apparent, and resulted in a broken page

The solution was to use a simple `loading` flag that would be turned to `false` only after *all* the data (including the funder names) was fetched
```js
// Init block
(function init() {
  $scope.today();

  $q.all([
    getOptionValues(),
    getJobRolesList($scope.$parent.contactId),
    vm.getContactList()
  ])
  .then(function () {
    $scope.loading = false;
  });
})();
```


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -719,6 +715,22 @@ define([
       }
 
       /**
+       * Extracts, from each job role past and preset, the contact id of every funder
+       *
+       * @return {Array} a list of ids
+       */
+      function extractFundersContactIds() {
+        return _(vm.present_job_roles.concat(vm.past_job_roles))
```
bmnepali: Never came across `_().map().thru().compact(). uniq().value()`.  What is it trying to do here?

com.civicrm.hrjobroles/js/test/controllers/hr-job-roles-controller/hr-job-roles-controller_test.js
```diff
@@ -479,6 +471,42 @@ define([
...
+              });
+
+              it('fetches the contact for each job role funder', function () {
+                expect(HRJobRolesService.getContactList).toHaveBeenCalledWith(null, funderContactIds);
+              });
+
+              /**
+               * Method that mimics how the controller extracts the ids
+               *
+               * @return {[type]} [description]
```
bmnepali: `@return {[type]} [description]`  type  and description is missing.

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
AkA84: They are all [lodash](https://lodash.com/docs/3.10.1#includes) methods, will add a more detailed description in the documentation

PR #1287, opened 2016-12-14
PCHR-1389: Calculate Toil expiry date API

tunbola:
In this PR, The AbsenceType.calculateToilExpiryDate endpoint which, given a date, it calculates when the toil would expire is created. The endpoint supports the following parameters:

- id(required): The id of the absence type.
- date(required) : The date to calculate TOIL expiry for.

| absence_type_id  | allow_accruals_request  | accrual_expiration_duration|accrual_expiration_unit |
| ------------- | ------------- | ------------- |------------- |
| 1  | 0  |   |  |
| 2  | 1  |  |  |
| 3  | 1  | 10  |1  |

```php
$result = civicrm_api3('AbsenceType', 'calculatetoilexpirydate', array(
  'absence_type_id' => 1,
  'date' => "2016-10-11",
));
```
**$result =** 
```json 
{

    "is_error": 1,
    "error_message": "This Absence Type does not allow Accruals Request"
}
```

```php
$result = civicrm_api3('AbsenceType', 'calculatetoilexpirydate', array(
  'absence_type_id' => 2,
  'date' => "2016-10-11",
));
```
**$result['values'] =** 
```json 
{

    "expiry_date": false,
}
```

```php
$result = civicrm_api3('AbsenceType', 'calculatetoilexpirydate', array(
  'absence_type_id' => 3,
  'date' => "2016-11-10",
));
```
**$result['values'] =** 
```json 
{

    "expiry_date": "2016-11-20",
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -1,6 +1,7 @@
 <?php
 
 use CRM_HRLeaveAndAbsences_Queue_PublicHolidayLeaveRequestUpdates as PublicHolidayLeaveRequestUpdatesQueue;
+use CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException as InvalidAbsenceTypeException;
 
 class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_AbsenceType {
 
@@ -539,5 +540,46 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
     }
 
     return null;
+  }
+
+  /**
+   * Returns the date the TOIL will expire based on the date given to the method.
+   *
+   * @param \DateTime $date
+   *   The date to calculate TOIL expiry date
+   *
+   * @return \DateTime|null
+   *   returns null when TOIL never expires for the Absence Type
+   *
+   * @throws CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
+   */
+  public function calculateToilExpiryDate(DateTime $date) {
+
+    if (!$this->allow_accruals_request) {
+      throw new InvalidAbsenceTypeException("This Absence Type does not allow Accruals Request");
+    }
+    if ($this->toilNeverExpires()) {
+      return null;
+    }
+
+    if ($this->accrual_expiration_unit == self::EXPIRATION_UNIT_DAYS) {
+      $unit = 'day';
+    }
+    if ($this->accrual_expiration_unit == self::EXPIRATION_UNIT_MONTHS) {
+      $unit = 'month';
+    }
+    $expiryDate = clone $date;
+    $expiryDate->modify('+'.$this->accrual_expiration_duration. ' ' . $unit);
+
+    return $expiryDate;
+  }
+
+  /**
+   * This method determines whether the Absence Type TOIL expires or not
+   *
+   * @return bool
+   */
+  private function toilNeverExpires() {
+    return !$this->accrual_expiration_unit || !$this->accrual_expiration_duration;
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsenceType.php
@@ -44,3 +44,50 @@ function civicrm_api3_absence_type_delete($params) {
 function civicrm_api3_absence_type_get($params) {
   return _civicrm_api3_basic_get(_civicrm_api3_get_BAO(__FUNCTION__), $params);
 }
+
+
+/**
+ * AbsenceType.calculateToilExpiryDate Specification
+ *
+ * @param array $spec
+ */
+function _civicrm_api3_absence_type_calculatetoilexpirydate_spec(&$spec) {
+  $spec['absence_type_id'] = [
+    'name' => 'absence_type_id',
+    'title' => 'Absence Type ID',
+    'type' => CRM_Utils_Type::T_INT,
+    'api.required' => 1
+  ];
+
+  $spec['date'] = [
+    'name' => 'date',
+    'title' => 'Date to calculate TOIL expiry for',
+    'type' => CRM_Utils_Type::T_DATE,
+    'api.required' => 1
+  ];
+}
+
+/**
+ * AbsenceType.calculateToilExpiryDate API
+ *
+ * @param array $params
+ *   An array of parameters passed to the API
+ *
+ * @return array API result descriptor
+ *
+ * @throws API_Exception
+ */
+function civicrm_api3_absence_type_calculatetoilexpirydate($params) {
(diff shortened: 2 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsenceType.php
```diff
@@ -44,3 +44,47 @@ function civicrm_api3_absence_type_delete($params) {
...
   return _civicrm_api3_basic_get(_civicrm_api3_get_BAO(__FUNCTION__), $params);
 }
+
+
+/**
+ * AbsenceType.calculateToilExpiryDate Specification
+ *
+ * @param array $spec
+ *
+ * @return void
```
davialexandre: no need for @return void. If the method doesn't return anything, you can omit the @return tag

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -540,4 +540,39 @@ public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
...
+
+  /**
+   * Returns the date the TOIL will expire based on the date given to the method.
+   *
+   * @param int $typeId
+   *   The Absence type ID
+   * @param string $date
+   *   The date to calculate TOIL expiry date
+   *
+   * @return null|false|string
```
davialexandre: Whenever you have a @return which returns different values, please always document what each type means

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -540,4 +540,39 @@ public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
...
+   */
+  public static function calculateToilExpiryDate($typeId, $date) {
+    $absenceType = new self();
+    $absenceType->id = $typeId;
+    $absenceType->find(true);
+
+    if (!$absenceType->allow_accruals_request) {
+      return null;
+    }
+    if (!$absenceType->accrual_expiration_unit || !$absenceType->accrual_expiration_duration){
```
tunbola: What of changing to a variable instead of an extra function? 
```
$toilNeverExpires = !$this->accrual_expiration_unit || !$this->accrual_expiration_duration
if($this->toilNeverExpires) {
  return false;
}
```

tunbola:
I agree with your suggestions but I Was thinking the BAO method should return the expiry date values and any other method consuming it e.g the API should
be responsible for treating it as it deems fit either by throwing an exception or doing something else.
That way we can keep the BAO method easily reusable. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: the advantage of the method is that it can be reused on other places and it hides the logic of how we know that toil never expires. Say we change things and add a toil_never_expire future, then, instead of fixing all the places were we use accrual_expiration_unit and accrual_expiration_duration, we would only need to change the method

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsenceType.php
```diff
@@ -44,3 +44,54 @@ function civicrm_api3_absence_type_delete($params) {
...
+ *   An array of parameters passed to the API
+ *
+ * @return array API result descriptor
+ *
+ * @throws API_Exception
+ */
+function civicrm_api3_absence_type_calculatetoilexpirydate($params) {
+  $absenceType = CRM_HRLeaveAndAbsences_BAO_AbsenceType::findById($params['absence_type_id']);
+
+  if (!$absenceType) {
```
davialexandre: The findById() method throws an exception if it can't find a record, so this condition will never match. You can remove it

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -624,4 +624,62 @@ public function testSetIsSickFlagForAbsenceType() {
     ]);
     $this->assertEquals(1, $absenceType->is_sick);
   }
+
+  /**
+   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
+   * @expectedExceptionMessage This Absence Type does not allow Accruals Request
+   */
+
```
davialexandre: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/AbsenceTypeTest.php
```diff
@@ -0,0 +1,96 @@
...
+use CRM_HRLeaveAndAbsences_Test_Fabricator_AbsenceType as AbsenceTypeFabricator;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+
+/**
+ * Class api_v3_AbsenceTypeTest
+ *
+ * @group headless
+ */
+class api_v3_AbsenceTypeTest extends BaseHeadlessTest {
+
```
davialexandre: What about adding a test to assert that an error is thrown when theres no Absence Type with the given ID?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -540,4 +541,45 @@ public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
...
+    }
+
+    if ($this->accrual_expiration_unit == self::EXPIRATION_UNIT_DAYS) {
+      $unit = 'day';
+    }
+    if ($this->accrual_expiration_unit == self::EXPIRATION_UNIT_MONTHS) {
+      $unit = 'month';
+    }
+
+    $date->modify('+'.$this->accrual_expiration_duration. ' ' . $unit);
```
davialexandre: Instead of modify the received date, it's safer to clone it and modify the clone:
```php
$expiryDate = clone $date;
$expiryDate->modify('+'.$this->accrual_expiration_duration. ' ' . $unit);
```
The callers of this method probably won't expected that the passed date will be changed.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -540,4 +541,45 @@ public static function getOneWithMustTakePublicHolidayAsLeaveRequest() {
...
+
+    return $date;
+  }
+
+  /**
+   * This method determines whether the Absence Type TOIL expires or not
+   *
+   * @return bool
+   */
+  public function toilNeverExpires() {
```
davialexandre: If the method is public, you need to add a test to it. Another option is to make the method private, then you won't need the test. I prefer the second option, since it keeps the public interface of the class as small as possible. 

In the future, if we need to use the method outside the class, we can make it public.

PR #1288, opened 2016-12-14
PCHR-1689: Add the WorkPattern.getCalendar API endpoint

davialexandre:
This API endpoint accepts only 2 params: `contact_id` and `period_id`. Based on that, it returns the contact calendar for the given period. The calendar is a list of dates with information about it's type, wether it's a working day, non-working day or weekend. The type information is fetched using the WorkPattern.

The dates returned are only those between the start and end date of the given absence period AND between any of the contacts contracts within the absence period. Example:

Given an Absence Period from 2015-01-01 to 2015-12-31 and given a Contract from 2015-02-01 to 2015-06-20 and another Contract from 2015-10-11 to 2016-03-04. The returned dates will be:
- From 2015-02-01 to 2015-06-20 (the period for the first contract)
- From 2015-10-11 to 2015-12-31 (the days of the contract overlapping the 2016 period won't be returned)

### Calling the API
The API can be called as:
```php
civicrm_api3('WorkPattern', 'getCalendar', [
  'contact_id' => 2,
  'period_id' => 2
]);
```

The return format is:
```json
{

    "is_error": 0,
    "version": 3,
    "count": 3,
    "values": [
        {
            "date": "2015-08-04",
            "type": {
                "value": "2",
                "name": "working_day",
                "label": "Yes"
            }
        },
        {
            "date": "2015-08-05",
            "type": {
                "value": "2",
                "name": "working_day",
                "label": "Yes"
            }
        },
        {
            "date": "2015-08-06",
            "type": {
                "value": "2",
                "name": "working_day",
                "label": "Yes"
            }
        },
    ]
}
```

### About the implementation
Even though the endpoint is inside the WorkPattern API, the whole logic is inside a separate service named WorkPatternCalendar. The Reason for this is to keep the BAO only responsible for fetching WorkPattern information and doing simple calculations based on it's own data. Building the calendar involves much more than that (like dealing with ContactWorkPattern, effective date calculation, Contracts and Absence Periods), so it's better to keep this in a separate class.

There is a `WorkPattern.getCalendar` method though, but it's much more simpler than that. Basically, it receives a start date and an end date, and call `getWorkDayTypeForDate()` for each of the dates in the period.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
@@ -1,7 +1,9 @@
 <?php
 
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
 use CRM_HRLeaveAndAbsences_BAO_WorkPattern as WorkPattern;
 use CRM_HRLeaveAndAbsences_BAO_WorkDay as WorkDay;
+use CRM_HRLeaveAndAbsences_Service_JobContract as JobContractService;
 
 class CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern extends CRM_HRLeaveAndAbsences_DAO_ContactWorkPattern {
 
@@ -186,6 +188,40 @@ class CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern extends CRM_HRLeaveAndAbsenc
 
     $workDayTypeId = $workPattern->getWorkDayTypeForDate($date, $startDate);
     return $workDayTypeId;
+  }
+
+  /**
+   * Returns a list of ContactWorkPattern instances for the given $contactID,
+   * which overlap the period enclosed by the given $start and $end dates
+   *
+   * @param int $contactID
+   * @param \DateTime $start
+   * @param \DateTime $end
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern[]
+   */
+  public static function getAllForPeriod($contactID, DateTime $start, DateTime $end) {
+    $tableName = self::getTableName();
+
+    $query = "SELECT * FROM {$tableName}
+              WHERE contact_id = %1 AND 
+                    effective_date <= %2 AND 
+                    (effective_end_date >= %3 OR effective_end_date IS NULL)";
+
+    $params = [
+      1 => [$contactID, 'Integer'],
+      2 => [$end->format('Y-m-d'), 'String'],
+      3 => [$start->format('Y-m-d'), 'String']
+    ];
+
+    $result = CRM_Core_DAO::executeQuery($query, $params, true, self::class);
+
+    $contactWorkPatterns = [];
+    while($result->fetch()) {
+      $contactWorkPatterns[] = clone $result;
+    }
+
+    return $contactWorkPatterns;
   }
 
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
@@ -1,0 +1,316 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
+use CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern as ContactWorkPattern;
+use CRM_HRLeaveAndAbsences_BAO_WorkPattern as WorkPattern;
+use CRM_HRLeaveAndAbsences_Service_JobContract as JobContractService;
+
+/**
+ * This class calculates a calendar for a Contact and an Absence Period,
+ * based on the the contact's work pattern(s).
+ *
+ * A calendar is just a list of dates with information about its type, according
+ * to a work pattern (i.e. if it's a working day, non working day or weekend)
+ */
+class CRM_HRLeaveAndAbsences_Service_WorkPatternCalendar {
+
+  /**
+   * @var int
+   */
+  private $contactID;
+
+  /**
+   * @var \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod
+   */
+  private $absencePeriod;
+
+  /**
+   * @var array
+   *   An array to cache the WorkPattern instances loaded by getWorkPatternById()
+   */
+  private $workPatternCache;
+
+  public function __construct($contactID, AbsencePeriod $absencePeriod, JobContractService $jobContractService) {
+    $this->contactID = $contactID;
+    $this->absencePeriod = $absencePeriod;
+    $this->jobContractService = $jobContractService;
+  }
+
+  /**
+   * Returns a list of all the dates on this calendar, on the same format as
+   * WorkPattern.getCalendar(). The difference from the method on the WorkPattern,
+   * is that this one only returns the dates between the calendar's contact
+   * contracts during the calendar's Absence Period. Also, it deals with the
+   * possibility of a contact having multiple work patterns during a single
+   * Absence Period.
+   *
+   * @see CRM_HRLeaveAndAbsences_BAO_WorkPattern.getCalendar()
+   *
+   * @return array
+   */
+  public function get() {
+    $workPatternsPeriods = $this->getWorkPatternsPeriods();
+
+    $calendar = [];
+    foreach($workPatternsPeriods as $workPatternPeriod) {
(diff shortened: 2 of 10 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
```diff
@@ -0,0 +1,316 @@
...
+        ];
+
+        continue;
+      }
+
+      // If the first returned work pattern starts being effective after the
+      // contract start date, it means that, for some time, that contract's
+      // contact didn't have any work pattern assigned. In this case, we have to
+      // use the default work pattern for this period
+      $firstWorkPattern = reset($contactWorkPatterns);
```
tunbola: Is this possible to have this kind of lapses between other work patterns for the contract period(s). i.e a situation where a work pattern ends and there is a time lapse before another is added. Is the default work pattern used in this scenario also?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -0,0 +1,106 @@
...
+  public function testGetCalendarRequiresContactIdAndPeriodID() {
+    civicrm_api3('WorkPattern', 'getCalendar');
+  }
+
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage Mandatory key(s) missing from params array: contact_id
+   */
+  public function testGetCalendarRequiresContactIdIfPeriodIDIsNotEmpty() {
+    civicrm_api3('WorkPattern', 'getCalendar');
```
tunbola: As the test name suggests, isn't period_id supposed to be passed in here?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage Mandatory key(s) missing from params array: contact_id
+   */
+  public function testGetCalendarRequiresPeriodIdIfContactIDIsNotEmpty() {
```
tunbola: As the test name suggests, isn't contact_id supposed to be passed in here?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -0,0 +1,106 @@
...
+    civicrm_api3('WorkPattern', 'getCalendar');
+  }
+
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage Unable to find a CRM_HRLeaveAndAbsences_BAO_AbsencePeriod with id 99989389121.
+   */
+  public function testGetCalendarThrowsAnErrorIfThePeriodIDIsNotForAnExistentAbsencePeriod() {
+    civicrm_api3('WorkPattern', 'getCalendar', ['contact_id' => 1, 'period_id' => 99989389121]);
```
tunbola: is it possible to have a test with a valid period but with an invalid/non existent contact?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/WorkPatternHelpersTrait.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_WorkPatternHelpersTrait {
+
+  public function getWorkDayTypeOptionsArray() {
+    $result = $result = civicrm_api3('OptionValue', 'get', array(
+      'sequential' => 1,
+      'option_group_id' => "hrleaveandabsences_work_day_type",
```
tunbola: we can change the double quotes to single quotes here for consistency.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
davialexandre: This is not possible. When a new work pattern is assigned to a contact the the effective_end_date for the current one is automatically set to new pattern effective_date - 1 day

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/WorkPatternHelpersTrait.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_WorkPatternHelpersTrait {
+
+  public function getWorkDayTypeOptionsArray() {
+    $result = $result = civicrm_api3('OptionValue', 'get', array(
```
tunbola: please use short array syntax here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public function testGetCalendarShouldReturnEmptyIfTheGivenContactDoesntExists() {
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2015-01-05'),
+      'end_date'   => CRM_Utils_Date::processDate('2015-01-19'),
+    ]);
+
+    $result = civicrm_api3('WorkPattern', 'getCalendar', ['contact_id' => 321, 'period_id' => $absencePeriod->id]);
```
tunbola: Was wondering if the method should rather throw an exception or a more descriptive message if the contact does not exist rather than returning empty?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
```diff
@@ -188,4 +190,38 @@ public static function getWorkDayType($contactId, DateTime $date) {
...
 
+  /**
+   * Returns a list of ContactWorkPattern instances for the given $contactID,
+   * which overlap the period enclosed by the given $start and $end dates
+   *
+   * @param int $contactID
+   * @param \DateTime $start
+   * @param \DateTime $end
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern[]
```
tunbola: Its also possible for this method to return an empty array if there are no work patterns for the period.
Why not` \CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern[]|[]` ?

davialexandre: Even when it's empty, it's still an array of ContactWorkPatterns, just without any elements. Also, if you check here https://phpdoc.org/docs/latest/references/phpdoc/types.html#arrays you'll see that when no type is specified, then the correct usage is `array` and not `[]`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
davialexandre: Yeah, I thought about that, but the thing is... to fetch the contracts we really only need the contact_id, not the whole contact and it seems to be a bit too much to fetch the contact (which takes at least 4 SQL queries), just to make sure it exists. That's why I opted for the empty array instead of an exception. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/WorkPatternHelpersTrait.php
davialexandre: This is what happens when you just copy and paste the code from somewhere else :P

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
tunbola: In that case. Lets leave as it is then. Might not be worth the effort :)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
tunbola: OK

PR #1290, opened 2016-12-15
PCHR-1644: Delete Leave Entitlements on Absence Type Deletion

MiyaNoctem:
On absence type deletion, leave entitlements associated to deleted absence types were being kept in the dtabase, breaking the leave entitlement interface for contract revisions.  Also, though front end validates if an absence type is being used in existing leave requests to allow/disallow deletion, this could be circumvented if you clicked on the delete link for the absence type, created a leave request for that absence type on another screen, and then confirmed deletion of absence type.

Implemented functionality to delete leave entitlements associated to Absence Type being deleted. Also, added validation to check if Absence Type being deleted is used on existing leave requests.

Added tests to verify use of deletion method in absence type BAO throws an exception when deleting absence types used in existing leave requests, and to check succesful deletion of absence type deletes associated entitlements also.

### DIFF
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
@@ -154,12 +154,26 @@ class CRM_HRAbsence_BAO_HRAbsenceType extends CRM_HRAbsence_DAO_HRAbsenceType {
     }
     return $result;
   }
-
-  public static function del($absenceTypeId) {
+  
+  /**
+   * Delete Absence Type
+   * 
+   * @param int $absenceTypeId
+   *   ID of Absence Type to be deleted
+   * @throws CRM_Core_Exception
+   *   If Absence Type is being used on existing leave requests
+   */
+  public static function del($absenceTypeId) {    
     $absenceType = new CRM_HRAbsence_DAO_HRAbsenceType();
     $absenceType->id = $absenceTypeId;
     $absenceType->find(TRUE);
 
+    if (($countRequests = self::countLeaveRequestsForAbsenceType($absenceType)) > 0) {
+      throw new CRM_Core_Exception(ts("You can not delete this absence type -- it is assigned to %1 existing leave request(s). If you do not want this absence type to be used going forward, consider disabling it instead.", array(1 => $countRequests)));
+    }
+
+    self::deleteLeaveEntitlements($absenceTypeId);
+    
     $absenceActivities = CRM_Core_OptionGroup::values('activity_type', FALSE, FALSE, FALSE, " AND grouping = 'Timesheet'", 'id');
 
     if ($absenceType->debit_activity_type_id && $id = CRM_Utils_Array::value($absenceType->debit_activity_type_id, $absenceActivities)) {
@@ -173,6 +187,53 @@ class CRM_HRAbsence_BAO_HRAbsenceType extends CRM_HRAbsence_DAO_HRAbsenceType {
 
     CRM_Utils_Hook::post('delete', 'HRAbsenceType', $absenceTypeId, $absenceType);
   }
+  
+  /**
+   * Count leave requests for given absence type object
+   * 
+   * @param CRM_HRAbsence_BAO_HRAbsenceType $absenceType
+   *   Instance of Absence Type object for which leave requests need to be counted
+   * @return int
+   *   Number of activities found for given Absence Type object
+   */
+  private static function countLeaveRequestsForAbsenceType(CRM_HRAbsence_DAO_HRAbsenceType $absenceType) {
+    $count = 0;
+    
+    if (!empty($absenceType->credit_activity_type_id)) {
+      $count += civicrm_api3('Activity', 'getcount', [
+        'sequential' => 1,
+        'activity_type_id' => $absenceType->credit_activity_type_id
+      ]);
+    }
+    if (!empty($absenceType->debit_activity_type_id)) {
+      $count += civicrm_api3('Activity', 'getcount', [
+        'sequential' => 1,
+        'activity_type_id' => $absenceType->debit_activity_type_id
+      ]);
+    }
+    
+    return $count;
+  }
+
+  /**
+   * Deletes all leave entitlements for given absence type ID.
+   * 
+   * @param int $absenceTypeId
+   *   ID of Absence Type for which entitlements have to be deleted
+   */
+  private static function deleteLeaveEntitlements($absenceTypeId) {
+    $result = civicrm_api3('HRJobLeave', 'get', [
+      'sequential' => 1,
+      'leave_type' => $absenceTypeId,
+      'jobcontract_revision_id' => array('>' => 0)
+    ]);
+    foreach ($result['values'] as $currentLeave) {
+      $result = civicrm_api3('HRJobLeave', 'delete', [
+        'id' => $currentLeave['id'],
+      ]);
+    }
+  }
+
 
   /**
    * Get the total duration for given 'Source Absence ID'
--- a/hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsenceType.php
@@ -1,12 +1,24 @@
 <?php
 
 class CRM_HRAbsence_Test_Fabricator_HRAbsenceType {
-
+  /**
+   * Array of default parameters
+   * @var array 
+   */
   private static $defaultParams = [
     'is_active' => 1,
   ];
-
-  public static function fabricate($params = []) {
+  
+  /**
+   * Helper method that builds parameters array combining given parameters with
+   * default parameters.
+   * 
+   * @param arry $params
+   *   Given array of parameters
+   * @return array
+   *   Result of merging default parameters with given parameters
+   */
+  private static function buildParamsArray($params) {
     if(empty($params['title'])) {
       $params['title'] = 'Absence Type ' . microtime();
     }
@@ -15,9 +27,33 @@ class CRM_HRAbsence_Test_Fabricator_HRAbsenceType {
       $params['name'] = 'Absence Type ' . microtime();
     }
 
-    $params = array_merge(self::$defaultParams, $params);
-
-    return CRM_HRAbsence_BAO_HRAbsenceType::create($params);
+    return array_merge(self::$defaultParams, $params);
+  }
+  
+  /**
+   * Absence Type fabricator method, uses BAO directly
+   * 
+   * @param array $params
+   *   Parameters to be passed to BAO to create absence type
+   * @return CRM_HRAbsence_BAO_HRAbsenceType
+   */
+  public static function fabricate($params = []) {
+    return CRM_HRAbsence_BAO_HRAbsenceType::create(self::buildParamsArray($params));
+  }
+  
+  /**
+   * Absence Type fabricator method that uses API and returns values of created
+   * record.
+   * 
+   * @param array $params
+   *   Parameters to be passed to API call to create absence type
+   * @return array
+   *   Values of created absence type
+   */
+  public static function fabricateUsingAPI($params = []) {
+    $params['sequential'] = 1;
+    $result = civicrm_api3('HRAbsenceType', 'create', self::buildParamsArray($params));
+    return array_shift($result['values']);
   }
 
 }
--- a/hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
+++ b/hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
@@ -1,0 +1,84 @@
+<?php
+
+use Civi\Test\HeadlessInterface;
+use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+use CRM_HRAbsence_Test_Fabricator_HRAbsenceType as AbsenceTypeFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobContract as HRJobContractFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobLeave as HRJobLeaveFabricator;
+use CRM_HRCore_Test_Fabricator_Activity as ActivityFabricator;
+
+/**
+ * Class CRM_HRAbsence_BAO_HRAbsenceTypeTest
+ * 
+ * @group headless
+ */
+class CRM_HRAbsence_BAO_HRAbsenceTypeTest extends PHPUnit_Framework_TestCase 
+  implements HeadlessInterface, TransactionalInterface {
+  
+  protected $absenceType;
+  protected $contact;
+  protected $jobContract;
+  protected $leave;
+  
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
+      ->install('org.civicrm.hrjobcontract')
+      ->installMe(__DIR__)
+      ->apply();
+  }
+  
+  /**
+   * Creates absence type, contact, contract and leave entitlement to be used in
+   * tests.
+   */
+  protected function setUp() {
+    $this->absenceType = AbsenceTypeFabricator::fabricateUsingAPI();
+
+    $this->contact = ContactFabricator::fabricate();
+
+    $this->jobContract = HRJobContractFabricator::fabricate(
+      ['contact_id' => $this->contact['id']], 
+      ['period_start_date' => '2015-01-01']
+    );
+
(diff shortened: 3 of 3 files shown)


hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+use CRM_HRAbsence_Test_Fabricator_HRAbsenceType as AbsenceTypeFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobContract as HRJobContractFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobLeave as LeaveFabricator;
+use CRM_HRCore_Test_Fabricator_Activity as ActivityFabricator;
+
+/**
+ * Class CRM_HRAbsence_BAO_HRAbsenceTypeTest
+ * Runs tests on AbsenceType BAO Class
```
davialexandre: This comment it's a bit redundant. The name of the class itself makes it clear that these are tests for the HRAbsenceType BAO.

Also, technically, the class doesn't run the tests. The Phpunit classes are who do that :D

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
+      ->install('org.civicrm.hrjobcontract')
+      ->installMe(__DIR__)
+      ->apply();
+  }
+    
+  protected function setUp() {
+    // Create Absence Type
```
davialexandre: The code itself is clear enough, no need for this comment

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+    $this->absenceType = AbsenceTypeFabricator::fabricateUsingAPI();
+    
+    // Create Contact, Contract and Leave for Absence Type
```
davialexandre: Again, no need for this comment here. In general, we should avoid this type of comments. Please check the other methods on this commit for these comments and please remove them

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+    $this->contact = ContactFabricator::fabricate(['first_name' => 'john', 'last_name' => 'shepard']);
```
davialexandre: Tip: if you're not going to use the name of the contact anywhere on the test, you can just call the `fabricate()` method without params

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+use Civi\Test\HeadlessInterface;
+use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+use CRM_HRAbsence_Test_Fabricator_HRAbsenceType as AbsenceTypeFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobContract as HRJobContractFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobLeave as LeaveFabricator;
```
davialexandre: Why the HRJobContract fabricator was aliased to `HRJobContract` but the HRJobLeave was aliased to `LeaveFabricator`?

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+      'leave_type' => $this->absenceType['id']
+    ]);
+  }
+  
+  /**
+   * Tests that absence typs used in leave requests cannot be deleted.
+   * 
+   * @expectedException CRM_Core_Exception
+   */
+  public function testDeletionOfUsedAbsenceType() {
```
davialexandre: For tests, instead of describing what it tests on a docblock, it's better to do it on the test name:

```php
public function testDoesNotAllowAbsenceTypeInUseByLeaveRequestToBeDeleted() {
}
```

If you do that, you can make phpunit output the results in a more readable format, like this:
```
CRM_HRLeaveAndAbsences_Service_EntitlementCalculation
 [x] Brought forward should be zero if absence type doesnt allow carry forward
 [x] Brought forward should be zero if theres no previous period
 [x] Brought forward should be zero if there is no entitlement for previous period
 [x] Brought forward should be zero if expiration duration has expired
 [x] Brought forward should be zero if the previous period is not over yet
 [x] Brought forward should not be more than the max number of days allowed to be carried forward
 [x] Brought forward should not be more than the number of remaining days in previous entitlement
 [x] Pro rata should be zero if the contact has no contracts
 [x] Pro rata should be zero if the contract doesnt have start and end dates
 [x] Pro rata should be rounded to the nearest half day
 [x] Pro rata include public holidays between contract dates even if contract says they should be added
 [x] Pro rata should be the sum of the pro rata for each contract during the absence period
 [x] The proposed entitlement for a contact without a contract should be zero
 [x] The proposed entitlement for a contract without start and end dates should be zero
 [x] The proposed entitlement for a period with previously overridden entitlement should not be the the overridden value
```

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+      'assignee_contact_id' => $this->contact['id'],
+      'activity_type_id' => $this->absenceType['debit_activity_type_id'],
+    ];
+    ActivityFabricator::fabricate($activityParam);
+    
+    // Try to delete, expect exception
+    CRM_HRAbsence_BAO_HRAbsenceType::del($this->absenceType['id']);
+    
+    // Verify Absence Type is still in DB
+    $result = civicrm_api3('HRAbsenceType', 'get', [
```
davialexandre: If del() is expected to thrown an exception, anything after line 65 will never be executed.

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+      'sequential' => 1,
+      'id' => $this->absenceType['id'],
+    ]);
+    $this->assertEquals(1, $result['count']);
+    $this->assertEquals($this->absenceType['id'], $result['values'][0]['id']);
+  }
+  
+  public function testDeletionOfLeaveOnAbsenceDeletion() {
```
davialexandre: Please improve this test name as well

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+    CRM_HRAbsence_BAO_HRAbsenceType::del($this->absenceType['id']);
+    
+    // Check Leave has been deleted too.
+    $result = civicrm_api3('HRJobLeave', 'get', [
+      'sequential' => 1,
+      'id' => $this->leave['id'],
+      'jobcontract_id' => $this->jobContract['id'],
+    ]);
+    
+    $this->assertEquals(0, $result['count']);
```
davialexandre: The problem I see with this tests is that, if someone changes the setUp method (delete everything on it, for example) this test will still pass. To avoid that, you can add an assertion before the call to `del()` to make sure the HRJobLeave record exists

hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsenceType.php
```diff
@@ -15,9 +29,33 @@ public static function fabricate($params = []) {
...
+  /**
+   * Absence Type fabricator method that uses API and returns values of created
+   * record.
+   * 
+   * @param array $params
+   *   Parameters to be passed to API call to create absence type
+   * @return array
+   *   Values of created absence type
+   */
+  public static function fabricateUsingAPI($params = []) {
```
davialexandre: why did you need 2 fabricate methods?

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -154,12 +154,32 @@ public static function getActivityStatus($return = 'label') {
...
-
-  public static function del($absenceTypeId) {
+  
+  /**
+   * Delete Absence Type
+   * 
+   * @param int $absenceTypeId
+   *   ID of Absence Type to be deleted
+   * @throws CRM_Core_Exception
+   *   If Absence Type is Being used on existin leave requests
```
davialexandre: s/Being/being/
s/existin/existing/

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -154,12 +154,32 @@ public static function getActivityStatus($return = 'label') {
...
+   */
+  public static function del($absenceTypeId) {    
+    // Load Absence Type
     $absenceType = new CRM_HRAbsence_DAO_HRAbsenceType();
     $absenceType->id = $absenceTypeId;
     $absenceType->find(TRUE);
 
+    // Validate if Absence Type is used in existing requests
```
davialexandre: here, instead of a comment you can make this a lot more readable by extracting this logic to a method and then:
```php
if(self::thereAreLeaveRequestsForAbsenceType($absenceType)) {
  throw new CRM_Core_Exception(ts("You can not delete this absence type -- it is assigned to %1 existing leave request(s). If you do not want this absence type to be used going forward, consider disabling it instead.", array(1 => ($creditCount + $debitCount))))
}
```

With that, you wouldn't be able to include the number of leave requests on the exception message though. If you really need to do that you can use the countLeaveRequests instead, which should return all the leave requests for an Absence Type, and not for an activity type (see my other comment on that method)

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -169,10 +189,49 @@ public static function del($absenceTypeId) {
...
+  
+  /**
+   * Count leave requests for given activity type id
+   * 
+   * @param int $activityTypeID
+   *   ID of activity type to be used to count
+   * @return int
+   *   Number of activities found with given activity type ID
+   */
+  private static function countLeaveRequests($activityTypeID) {
```
davialexandre: Since the method counts based on a parameter, it's better to put that on the method name: `countLeaveRequestsByActivityType()`

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -169,10 +189,49 @@ public static function del($absenceTypeId) {
...
+  private static function deleteLeaveEntitlements($absenceTypeId) {
+    $result = civicrm_api3('HRJobLeave', 'get', [
+      'debug' => 1,
+      'sequential' => 1,
+      'leave_type' => $absenceTypeId,
+      'jobcontract_revision_id' => array('>' => 0)
+    ]);
+    foreach ($result['values'] as $currentLeave) {
+      $result = civicrm_api3('HRJobLeave', 'delete', [
+        'debug' => 1,
```
davialexandre: Why debug? Also, I don't think sequential will make any difference for a delete operation

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -169,10 +189,49 @@ public static function del($absenceTypeId) {
...
+
+  /**
+   * Deletes all leave entitlements for given absence type ID.
+   * 
+   * @param int $absenceTypeId
+   *   ID of Absence Type for which entitlements have to be deleted
+   */
+  private static function deleteLeaveEntitlements($absenceTypeId) {
+    $result = civicrm_api3('HRJobLeave', 'get', [
+      'debug' => 1,
```
davialexandre: why debug?

hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsenceType.php
```diff
@@ -1,12 +1,26 @@
 <?php
-
+/**
+ * Fabricator for Absence Types
+ */
```
davialexandre: The class name is already descriptive enough. You're just being redundant

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -173,6 +187,53 @@ public static function del($absenceTypeId) {
 
     CRM_Utils_Hook::post('delete', 'HRAbsenceType', $absenceTypeId, $absenceType);
   }
+  
+  /**
+   * Count leave requests for given absence type object
+   * 
+   * @param CRM_HRAbsence_BAO_HRAbsenceType $absenceType
+   *   Instance of Absence Type object for which leave reequests need to be counted
```
davialexandre: s/reequests/request/

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -173,6 +187,53 @@ public static function del($absenceTypeId) {
...
+   * @return type
```
davialexandre: @return int

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -173,6 +187,53 @@ public static function del($absenceTypeId) {
...
+   *   Number of activities found with given activity type ID
```
davialexandre: Not active type ID anymore. It accepts an HRAbsenceType instance now

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    $this->leave = HRJobLeaveFabricator::fabricate([
+      'jobcontract_id' => $this->jobContract['id'],
+      'leave_type' => $this->absenceType['id']
+    ]);
+  }
+  
+  /**
+   * Tests that absence typs used in leave requests cannot be deleted.
```
davialexandre: Since now the test name is descriptive enough, no need for the description on the docblock

hrabsence/tests/phpunit/CRM/HRAbsence/BAO/HRAbsenceTypeTest.php
```diff
@@ -0,0 +1,89 @@
...
+      'activity_type_id' => $this->absenceType['debit_activity_type_id'],
+    ];
+    ActivityFabricator::fabricate($activityParam);
+    
+    CRM_HRAbsence_BAO_HRAbsenceType::del($this->absenceType['id']);
+  }
+  
+  /**
+   * Tests if deletion of absence type deletes associated entitlements.
+   */
```
davialexandre: No need to the description here as well. The test name is descriptive enough

hrabsence/CRM/HRAbsence/Test/Fabricator/HRAbsenceType.php
davialexandre: If it's just a matter of the values returned, why not update the fabricate() method to return the values you need? Seems to be simpler than adding a new method (also, API calls are slow and will impact tests performance). You can change the fabricate method to something like this:

```php
public static function fabricate($params = []) {
   $absenceType = CRM_HRAbsence_BAO_HRAbsenceType::create(self::buildParamsArray($params));
   $absenceType = CRM_HRAbsence_BAO_HRAbsenceType::findByPk($absenceType->id);
   return $absenceType;
}
```

or, you can even make the findByPk part optional (to avoid unnecessary sql queries):

```php
public static function fabricate($params = [], $loadAfterSave = false) {
   $absenceType = CRM_HRAbsence_BAO_HRAbsenceType::create(self::buildParamsArray($params));
   
   if($loadAfterSave) {
     $absenceType = CRM_HRAbsence_BAO_HRAbsenceType::findByPk($absenceType->id);
   }
   
   return $absenceType;
}
```

PR #1292, opened 2016-12-16
PCHR-1784: Added Id and Class of parent tag to popup.

sleekvick:
# Scheduled Jobs Cannot Be Enable

The scheduled jobs section of administer would always break every time users try to enable a job.
enable-error

From the error in the console, the problem is  **$row[0].id** will be undefined because .crmentity is not a parent of the **this** which is the current anchor tag clicked by the user.

```
      el = this[0],
      ret = {},
      $row = this.first().closest('.crm-entity');
      ret.entity = $row.data('entity') || $row[0].id.split('-')[0];](url)
```

However due to changes made [in this PR](https://github.com/civicrm/civihr/pull/1173) the ul tag is positioned absolute relative to the position of the anchor and is not a child of .crm-entity has espected by the code above.

# Adding the Id and Class of parent tag to popup.

The solution takes the ID and the Class of the parent tag and adds it to the popup ul tag, before the position is set to absolute and moved outside the parent tag with the id expected by the code above. Essentially the solution deceives the **$row = this.first().closest('.crm-entity');** and also provides it the id it needs. 

Furthermore the solution does not have any css implications for now, but possible cons of the solution is when specific styling affects the tr affects the popup ul. Also when the user clicks outside the popup is detached from body so the id and the class does not remain there by avoiding any conflict later on.


hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -12,6 +12,11 @@ Prevent Popups to overflow tables.
     if ($button.hasClass('btn-slide-active')) {
       $popup = $button.children('ul.panel');
 
+      // Use id of parent tr tag
+      $popup.attr("id", $button.closest('.crm-entity')[0].id);
+      // Use class of parent tr tag
+      $popup.addClass($($button.closest('.crm-entity')[0]).attr("class"));
```
AkA84: Why do you need to add the class as well? I thought the crm script only looked for the `id`?

AkA84: Anyway, if you do really need the class, then please cache `$button.closest('.crm-entity')[0]`, since you're using it twice
```js
var $button = $(this), $body = $('body'), $popup, $closestEntity, buttonOffset;

if ($button.hasClass('btn-slide-active')) {
  // ...
  var $closestEntity = $button.closest('.crm-entity')[0];

  $popup.attr("id", $closestEntity.id);
  $popup.addClass($closestEntity.attr("class"));

```

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -12,6 +12,11 @@ Prevent Popups to overflow tables.
     if ($button.hasClass('btn-slide-active')) {
       $popup = $button.children('ul.panel');
 
+      // Use id of parent tr tag
```
AkA84: Both "Use id of parent tr tag", "Use class of parent tr tag" are pretty meaningless comments (I can already see what you are doing, but you're not telling me *why*).

Given that the reason why you do this is a bit complicated to explain in a comment, just leave the number of the ticket

```js
// PCHR-1784
$popup.attr(/* .. */);
$popup.addClass(/* .. */);

hrui/js/src/civihr-popup/civihr-popup.js
sleekvick: The class is necessary because the jquery plugin needs the class to get the id.

PR #1293, opened 2016-12-16
PCHR-1493: Fixing inability to export contracts properly

omarabuhussein:
## Problem

In civicrm 4.7 , the way exporting data is changed, So if you need to export directly via a URL only without going to fields selection page then you should supply the task parameter to the export URL . The task could hold many values depends on what are you trying to do with data, but if you want
to export the data in CSV format then the task parameter value should equal (**report_instance.csv**). Doing this will also result in fetching all contract revisions and data correctly since doing it through the fields selection page will cause some parameters like contract ID to be missing when fetching the result which will render empty data in the exported file.

There are two other requirments, first : the job contract revision rows should be orderd in descent order. the other is the revision editor name should be provided with the data for each revision. 

before

## Solution

1 - Adding **task=report_instance.csv**  paratmeter to export URL to automatically export contacts instead of directing to field selection page , and to fix the empty data issue.

2- changing civicrm_hrjobcontract_revision_revision_id order by parmenter to order data in descent order.

3- Altering the exporting query to join civicrm_uf_match and contact table to fetch the revision editor display name.

after


davialexandre:
This has been stale for more than one year and the related ticket has been closed, so  I'm closing this too

PR #1294, opened 2016-12-16
PCHR-1680: Create TOILRequest Entity

tunbola:
This entity is an extension of the LeaveRequest is used to store information specific to a TOIL request. These are its fields:
- *id*: type=int, primary_key
- *leave_request_id*: type=int, required=true, FK to this request's LeaveRequest 
- *duration*: type=int, required=true, The duration of the overtime work, in minutes

Some Validations are also added to the TOILRequest Entity BAO Create method:
- The "Toil to be accrued" cannot be more than the maximum defined (if any) for the absence type
- The user cannot request TOIL for past days if the absence type is not set up so
- The TOIL amount is one of the values in the TOIL Amount option group


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
```diff
@@ -143,8 +139,10 @@ private static function validateValidTOILPastDaysRequest($params) {
    *
    * @param CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
    *   The Leave Request created by this TOIL Request
+   * @param float $toil_to_accrue
+   *   The amount of TOIL to be accrued.
    */
-  private function saveBalanceChange(LeaveRequest $leaveRequest) {
+  private function saveBalanceChange(LeaveRequest $leaveRequest, $toil_to_accrue) {
```
davialexandre: can you use camelCase here, to keep it consistent with the other methods params?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -94,6 +94,18 @@ public function testCreateTOILRequestWithPastDatesAndAbsenceTypeAllowsIt() {
...
+    TOILRequest::validateParams([
+      'type_id' => $absenceType->id,
+      'contact_id' => 1,
+      'status_id' => 1,
+      'from_date' => $fromDate->format('YmdHis'),
+      'to_date' => $toDate->format('YmdHis'),
+      'toil_to_accrue' => $this->toilAmounts['2 Days']['value'],
+      'duration' => 120
+    ]);
+
```
davialexandre: I don't understand the reason for this.

It would be better to have a test to make sure the validateParams is called inside create when $validate = true. Something like this would be enough, I think:
```php
/**
* @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidTOILRequestException
* @expectedExceptionMessage You cannot request TOIL for past days
*/
public function testValidateParamsIsCalledOnCreate() {
  // code to request toil for past days
}
```

You don't need to test all possible exceptions. Only one is enough to make sure validaParams has been called

PR #1295, opened 2016-12-19
PCHR-1414: Make "Not Applicable" Payscale a Valid Value to Import

MiyaNoctem:
When importing job contracts, using "Not Applicable" as a pay scale resulted in an error.  This was because the label for all Pay Scales was expected to be of the form:
"<pay_scale> - <pay_grade> - <currency> <amount> per <periodicity>"

This meant the valid value to import a "Not Applicable" pay scale was, literally:
"Not Applicable - - per "

Fixed by building the expected label in CRM_Hrjobcontract_SelectValues::buildPayScales(), checking if pay_grade is empty, mirroring how the label is built on CRM_Hrjobcontract_BAO_Query::buildAdvancedSearchPaneForm().


coldrunKacper:
Approved.

PR #1296, opened 2016-12-19
PCHR-1828: Add details for LeaveRequest validation errors

tunbola:
In this PR, the exceptions thrown during the Leave Request validation is updated to include more details. Besides the error message, it now  contains the field where the error was found and the Exception/Error code code.

A new exception class CRM_HRLeaveAndAbsences_Exception_EntityValidationExceptionhas been added which has two private fields: field and exceptionCode. The InvalidLeaveRequestException class is the only class that extends this class for now.

The following Error messages and Exception/Error codes were implemented

| Error Message | Error Code |
| ------------- | ------------- |
|Leave Requests should have a start date|leave_request_empty_from_date|
|Leave Request start date cannot be greater than the end date|leave_request_from_date_greater_than_end_date|
|The Leave request dates is not contained within a valid absence period|leave_request_not_within_absence_period|
|The Leave request dates must not have dates in more than one contract period|leave_request_overlapping_multiple_contracts|
|Balance change for the leave request cannot be greater than the remaining balance of the period|leave_request_balance_change_greater_than_remaining_balance|
|Leave Request must have at least one working day to be created|leave_request_doesnt_have_working_day|
|This Leave request has dates that overlapps with an existing leave request|leave_request_overlaps_another_leave_request|

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -59,7 +59,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   private static function validateParams($params) {
     if (empty($params['from_date'])) {
       throw new InvalidLeaveRequestException(
-        "Leave Requests should have a start date"
+        'Leave Requests should have a start date',
+        'leave_request_empty_from_date',
+        'from_date'
       );
     }
 
@@ -92,7 +94,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     //either there was an overlap or the absence period does not simply exist.
     if (!$period) {
       throw new InvalidLeaveRequestException(
-        "The Leave request dates is not contained within a valid absence period"
+        'The Leave request dates are not contained within a valid absence period',
+        'leave_request_not_within_absence_period',
+        'from_date'
       );
     }
   }
@@ -120,7 +124,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
 
     if ($contract['count'] > 1) {
       throw new InvalidLeaveRequestException(
-        "The Leave request dates must not have dates in more than one contract period"
+        'The Leave request dates must not have dates in more than one contract period',
+        'leave_request_overlapping_multiple_contracts',
+        'from_date'
       );
     }
   }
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -92,7 +94,9 @@ private static function validateAbsencePeriod($params) {
     //either there was an overlap or the absence period does not simply exist.
     if (!$period) {
       throw new InvalidLeaveRequestException(
-        "The Leave request dates is not contained within a valid absence period"
+        'The Leave request dates is not contained within a valid absence period',
```
davialexandre: Shouldn't this be 'There Leave Request dates ARE not...'?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -228,7 +238,9 @@ private static function validateNoOverlappingLeaveRequests($params) {
 
     if ($overlappingLeaveRequests) {
       throw new InvalidLeaveRequestException(
-        "This Leave request has dates that overlapps with an existing leave request"
+        'This Leave request has dates that overlapps with an existing leave request',
```
davialexandre: s/overlapps/overlaps/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -92,7 +94,9 @@ private static function validateAbsencePeriod($params) {
     //either there was an overlap or the absence period does not simply exist.
     if (!$period) {
       throw new InvalidLeaveRequestException(
-        "The Leave request dates is not contained within a valid absence period"
+        'The Leave request dates is not contained within a valid absence period',
+        'leave_request_not_within_absence_period',
+        'type_id'
```
davialexandre: Instead of using type_id as the field, what about using from_date? I know sometimes the error will be because of the value in to_date, but at least this will be closer to what have caused the invalidation.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -120,7 +124,9 @@ private static function validateLeaveDatesDoesNotOverlapContracts($params) {
 
     if ($contract['count'] > 1) {
       throw new InvalidLeaveRequestException(
-        "The Leave request dates must not have dates in more than one contract period"
+        'The Leave request dates must not have dates in more than one contract period',
+        'leave_request_overlapping_multiple_contracts',
+        'to_date'
```
davialexandre: to_date is a non required field. What if we get this error on a single-day LeaveRequest? I think it would be better to use the from_date as the field here

PR #1297, opened 2016-12-20
PCHR-1267: my-leave-report component

AkA84:
_(Related to https://github.com/compucorp/civihr-employee-portal-theme/pull/126)_

This PR implement the business logic of the `<my-leave-report></my-leave-report>` component, and plugs it in the template

report

Business logic implemented:
* Absence Types, Absence Periods, Entitlements and balance changes are fetched on initialization
* When opening a sub-section, its data is fetched from the server and displayed
* When switching to another Absence Period, the data is refreshed
* Each leave request gets different possible actions depending on its status
* It is possible to cancel a leave request

### Add a `<my-leave-report-item>` component
The component will represent each item (leave request, breakdown) in the table, and will avoid unnecessary repetition in the template
### Allow async operation in `dialog` before closing
The `common/modules/dialog` module should be amended as to allow an async operation to be performed before the dialog closes itself

## Permissions
Despite granting the civihr_staff user the "access AJAX API" permission, the user still couldn't use the API endpoints from the SSP, due to the fact that CiviCRM still checks if the user has the "administer" permission anyway.

The solution unfortunately is to manually list all the endpoints that civihr_staff needs to use, and override their permissions check-list so that it includes only "access AJAX API"

```php
function hrleaveandabsences_civicrm_alterAPIPermissions($entity, $action, &$params, &$permissions) {
  $actionEntities = [
    'get' => ['absence_type', 'absence_period', 'option_group', 'option_value', 'leave_period_entitlement'],
    'getbalancechangebyabsencetype' => ['leave_request']
  ];

  foreach ($actionEntities as $action => $entities) {
    foreach ($entities as $entity) {
      $permissions[$entity][$action] = ['access AJAX API'];
    }
  }
}
```

### Module dependencies
There were module dependencies that weren't defined correctly, both on the AngularJS and RequireJS level

### `jasmine.clock()`
When `jasmine.clock()` is used, it also needs to be uninstalled via `jasmine.clock().uninstall()` after the test has run, otherwise the current time is kept changed throughout the whole test suite

### Use `OptionGroup` model in `LeaveRequestInstance`
The `LeaveRequestInstance` was using `api.optionGroup` instead of going through the `OptionGroup` model 

### Fix how statuses are passed to `LeaveRequest.getbalancechangebyabsencetype`
The statuses list was passed as a simple `=` query params
```js
var params = {
  // ...
  statuses: statuses || null
}
```
but that was refused by the endpoint, as it expected the statuses to be passed as an `IN` query param.
```js
var params = {
  // ...
  statuses: statuses ? { 'IN': statuses } : null,
}
```

### Fix response of .all() for LeaveRequest model and api
The `LeaveRequestAPI` service is using the `getAll()` method, which returns a composite object `{ list, total, allIds }`, but both the api and model services were ignoring that, treating the `getAll()` response as if it was composed only by the leave requests list

Before:
```js
return leaveRequestAPI.all(this.processFilters(filters), pagination, sort, params)
  .then(function (leaveRequests) {
    return leaveRequests.map(function (leaveRequest) {
      return instance.init(leaveRequest, true);
    });
  });
```

After:
```js
return leaveRequestAPI.all(this.processFilters(filters), pagination, sort, params)
  .then(function (response) {
    response.list = response.list.map(function (leaveRequest) {
      return instance.init(leaveRequest, true);
    });

    return response;
  });
```

### Ignore templates module in reqangular in `gulp watch`
When changing a single file in the reqangular extension, instead of running just the test for that single file, gulp was running the entire test suite. That was caused by the fact that `templates.js` (the file that contains the cached templates) was getting re-generated on every change, and the task runner detected the change and was trying to find a related test file for it to run (when gulp can't find a related test file for a changed file, it runs the whole test suite as a fallback)

Disabling the watch on `templates.js` solved the problem

### Add ability to fetch the collection of mocked option group
The mocked option group module didn't allow to simply fetch a whole collection, i've added
```js
getCollection: function (collection) {
  return mockedOptionValues[collection];
}
```
that takes care of that

### Integration between different mocked data
Given that all mocked data has links to each other in some way or another (entitlements are for specific absence types and periods, leave requests are of specific types, breakdown entires are of specific day types, etc), I've converted some of the manually-written mocked data into dynamically-generated collections based on those links. These ensure data-consistency, which is necessary to have in order to have the test running in realistic scenarios

```js
// generate a mock entitlement for every absence type, for every period
absencePeriodData.all().values.forEach(function (absencePeriod) {
  absenceTypeData.all().values.forEach(function (absenceType) {
    var id = i++;
    var entitlement = {
      "id": id,
      "period_id": absencePeriod.id,
      "type_id": absenceType.id,
      "contact_id": "202",
      "overridden": "0",
      "api.LeavePeriodEntitlement.getentitlement": {
        "is_error": 0,
        "version": 3,
        "count": 1,
        "values": [{
          "id": id,
          "entitlement": _.random(0, 30)
        }]
      }
    };

    // ...
```

### Implement `LeaveRequestInstance.toAPIFilter()`
When sending data to the API, a `LeaveRequestInstance` needs to get rid of its custom data, namely the `balance_change` and `dates` fields.
```js
  if (_.includes(['balance_change', 'dates'], key)) {
    return;
  } else {
    result[key] = this[key];
  }
}
```

### Disable cache for entitlement and leave absence calls
The calls to `LeavePeriodEntitlement.all` and `LeaveRequest. getbalancechangebyabsencetype` are now not cached, since we need to make sure to always get the latest values (for example, after we delete a leave request)

### Convert dates in Models only if format != YYYY-MM-DD
This is kind of a quickfix for a work-in-progress feature i wrote when i was working on the appraisals extension.

The parent `Model` service provides a `processFilters()` method that should be called by every model before passing params to the api, and this function among other things makes sure that dates are always passed to the `YYYY-MM-DD`
```js
function apiDateFormat(date) {
  return moment(date, 'DD/MM/YYYY').format('YYYY-MM-DD');
}
```
problem is that it was assuming always that the dates were in the DD/MM/YYYY format (which was true for appraisals, while in L&A the dates are always kept in the YYYY-MM-DD format, and so the conversion failed.

As a quick fix, we now just check first if the format is not already YYYY-MM-DD before attempting the conversion
```js
function apiDateFormat(date) {
  return isStandardDateFormat(date) ? date : moment(date, 'DD/MM/YYYY').format('YYYY-MM-DD');
}

function isStandardDateFormat(date) {
  return moment(date, 'YYYY-MM-DD').format('YYYY-MM-DD') === date;
}
```

### Encapsulate loading spinner style in directive itself
The `<crm-loading>` directive is used to show a loading spinner when data is loading.

The problem was that the style of the spinner (and the `background` property holding the encoded image) was in the `bootstrapcivicrm` extension, which didn't make this directive "portable" (as we would have had to copy that same style in the SSP theme).

Simply including the style directly in the directive template solved the issue


deb1990:
When there is no data for a section(eg. Expired), there should be an **Empty message**.

Right now its just an header with no body.
my_leave___civihr_1_6_demo

deb1990:
The action button's looks very small and hard to click, can we do something similar to 
https://github.com/compucorp/civihr-employee-portal-theme/pull/127
(refer the screengrab)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -575,5 +575,25 @@ define([
         return $q.resolve(value);
       }
     });
+
+    describe('toAPI()', function () {
```
deb1990: LeaveRequestInstance does not have a `toAPI` method. This test case should not be here

deb1990: Maybe you meant `toAPIFilter`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -186,6 +186,20 @@ define([
...
           return deferred.promise;
+        },
+
+        /**
+         * Override of parent method
+         *
+         * @param {object} result - The accumulator object
+         * @param {string} key - The property name
+         */
+        toAPIFilter: function (result, __, key) {
```
AkA84: The `_` or `__` as a function param indicates a param for which you have no use in the function implementation, but that you still need to declare since you need to get the parameter after that. I don't remember where I saw this before, but i learned that this is a convention. In fact I used `__` just to avoid the confusion with `_`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/apis/leave-request-api-mock.js
```diff
@@ -13,7 +13,15 @@ define([
...
           return $q(function (resolve, reject) {
-            resolve(mockData.all().values);
+            var list = mockData.all().values;
+
+            resolve({
+              list: list,
+              total: list.length,
+              allIds: list.map(function (leaveRequest) {
+                return leaveRequest.id;
+              }).join(',')
```
AkA84: It would be a syntax error, you are inside an object definition, it would be equivalent to doing
```js
{
  foo: 'foo',
  bar: 'bar';
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
AkA84: `toAPI()` is defined in the parent `Instance` service, you can see that `LeaveRequestInstance` uses it here
```js
toAPIData = leaveRequest.toAPI();
```
`toApi()` in turns uses the `toAPIFilter` in which you have the transformations. The reason that `toAPIFilter` is exposed in the object interface is that allows each instance to extend the original `toAPIFilter` defined in `Instance`
```js
/**
 * Function that filters data meant to be sent to the API, being
 * called in as the iteratee of a _.transform() call
 *
 * Can be overriden by children api for custom filtering
 *
 * @param {object} result - The accumulator object
 * @param {string} key - The property name
 */
  result[key] = this[key];
}
```

AkA84:
@deb1990 about the action button, let's leave it like it is. we can iterate over it later

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
AkA84: > In fact I used __ just to avoid the confusion with _

Although it seems that it doesn't really work given your comment :D

PR #1298, opened 2016-12-20
PCHR-1303: Augment leave request api model model instance

dhavalsharma:
# LeaveRequest 
Augment LeaveRequest api end points as per ticket [PCHR-1303](https://compucorp.atlassian.net/browse/PCHR-1303).

## API
User can `create`, `update` and `calculateBalanceChange` on leaverequests from `LeaveRequestAPI`. 
Also, added `isValid` for doing additional validation before `create` call.

## Model
Add `calculateBalanceChange` method.

## Instance 
Add `create`, `update` and `isValid` methods.

## Source Code location
It is located under `civihr/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular`.

The implementation is located under shared folder. 

The test cases are in tests folder. 

## Sample LeaveRequest
A sample response to `isValid` is presented below 

### `isValid` success response
```json
{
  "is_error": 0,
  "count": 0,
  "values": []
}
```
### `isValid` error  response
```json
{
  "is_error": 0,
  "count": 1,
  "values": {
    "from_date": [
      "overlaps_existing_leave_request"
    ]
  }
}
```


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+       * to_date and to_type.
+       *
+       * @return {Promise} containing the detailed breakdown of balance leaves
+       */
+      calculateBalanceChange: function (params) {
+        $log.debug('LeaveRequestAPI.calculateBalanceChange');
+        var deferred = $q.defer();
+
+        if (params && !('contact_id' in params) || !('from_date' in params) || !('from_date_type' in params)) {
+          deferred.resolve({
```
AkA84: This should be a `deferred.reject`. But why are you doing this validation? Isn't this supposed to be taken care of by the backend endpoint already (cc @davialexandre )? Aren't we duplicating validation logic?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+       * If to_date is given then to_date_type is also mandotory.
+       *
+       * @return {Promise} containing the leave request object additionally with id key set
+       * else return array object with is_error key set
+       */
+      create: function (params) {
+        $log.debug('LeaveRequestAPI.calculateBalanceChange');
+        var deferred = $q.defer();
+
+        if (params && 'to_date' in params && !('to_date_type' in params)) {
```
AkA84: Same as before

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+          });
+        } else if (params && !('contact_id' in params) || !('from_date' in params) ||
+          !('from_date_type' in params) || !('status_id' in params)) {
+
+          deferred.resolve({
+            is_error: 1,
+            error_message: 'contact_id, from_date, status_id and from_date_type params are mandatory'
+          });
+        } else {
+          this.isValid(params)
```
AkA84: Isn't the backend `create` endpoint already doing this internally (cc @davialexandre )? If yes, there is not reason to duplicate the logic

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+        }
+
+        return deferred.promise;
+      },
+
+      /**
+       * The method will validate params for a new new leave request.
+       *
+       *
+       * @param {Object} params matched the API end point params with
```
AkA84: Does it make sense to copy-paste the same description for every endpoint?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+       * and optional values for to_date and to_date_type.
+       * If to_date is given then to_date_type is also mandotory.
+       *
+       * @return {Promise} containing the status of isValid api call
+       */
+      isValid: function (params) {
+        $log.debug('LeaveRequestAPI.isValid');
+        var deferred = $q.defer();
+
+        deferred.resolve(this.sendGET('LeaveRequest', 'isValid', params)
```
AkA84: why are you using `deferred` at all here? `this.sendGET` already returns a promise

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -61,6 +61,125 @@ define([
...
+          .then(function (data) {
+            if (data.count > 0) {
+              return {
+                is_error: 1,
+                errors: data.values
+              }
+            }
+            return data.values;
```
AkA84: I'm not sure I understand what you're doing here. If there is something in `data.count` then you returned a custom object containing `data.values`, otherwise you return...`data.values`?

I understand if you want to return a custom response in case there are errors, but why returning `data.values` in case there are no errors?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -53,8 +53,34 @@ define([
          * @return {Promise} Resolved with {Object} Updated Leave request
          */
         update: function (attributes) {
-          return LeaveRequestAPI.sendPOST('LeaveRequest', 'create', _.assign({}, this.attributes(), attributes))
+          var updatedAttributes = _.assign({}, this.attributes(), attributes);
```
AkA84: I just realized that this is not how `update()` is supposed to work. It shouldn't accept any parameter, it should just send the internal state of the instance to the api
```js
update: function () {
  return LeaveRequestAPI.update(this.attributes);
});
```
Usage example:
```js
var instance = LeaveRequest.find(123);

instance.status_id; // 1
instance.status_id = 2;

instance.update().then(function () {
  instance.status_id; // 2
})
.catch(function (err) {

});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -53,8 +53,34 @@ define([
...
+          return LeaveRequestAPI.update(updatedAttributes);
+        },
+
+        /**
+         * This method is used to create a new leave request
+         *
+         * @param {object} attributes - Values which will be used to create new leave request
+         * @return {Promise} Resolved with {Object} Created Leave request
+         */
+        create: function (attributes) {
```
AkA84: Same for `update`, I don't think `create()` should accept any parameter, it should just send to the api the data already stored in the instance

```js
var instance = LeaveRequest.init({ /* some custom data */ });

instance.create().then(function () {
  instance.id; // the instance has now an id
});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -53,8 +53,34 @@ define([
...
+          var updatedAttributes = _.assign({}, this.attributes(), attributes);
+          return LeaveRequestAPI.create(updatedAttributes);
```
AkA84: You have created the instance on the database, you should store its id now

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -53,8 +53,34 @@ define([
...
+        },
+
+        /**
+         * This method is used to validate a leave request attributes.
```
AkA84: Please get rid of all these `This method is/This method does`
```js
/**
 * Validates itself
 */

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js [hunk 11]
```diff
@@ -53,8 +53,34 @@ define([
...
+         *
+         * @param {object} attributes - Values which are currently part of
+         *  leave request instance
+         * @return {Promise} empty array if no error found otherwise an object
+         *  with is_error set and array of errors
+         */
+        isValid: function (attributes) {
```
AkA84: Why are you passing `attributes` to `isValid`? The point of having all these methods (`create`, `update`, `isValid`) on an instance is so that the instance automatically passes its internal state to them

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -53,8 +53,34 @@ define([
...
+          var updatedAttributes = _.assign({}, this.attributes(), attributes);
+          return LeaveRequestAPI.isValid(updatedAttributes);
```
AkA84: As already stated, always leave an empty line before a `return`, if there is some code before it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: discussed to have basic validations on front end.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: params description is same and hence repeated

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -42,10 +42,12 @@ define([
         var deferred = $q.defer();
 
         if (!contactId || !periodId) {
-          deferred.resolve({
+          deferred.reject({
             is_error: 1,
             error_message: 'contact_id and period_id are mandatory'
           });
+
+          return deferred.promise;
```
AkA84: You were doing things right here. It's better to have one `return` in the end if possible, so wrapping `this.sendGET`  in `deferred.resolve` was correct

AkA84: This applies for every other instance in which you are doing this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -55,24 +57,35 @@ define([
...
+       * @param {object} params - Updated values of leave request
        * @return {Promise} Resolved with {Object} Updated Leave request
        */
-      update: function (updatedAttributes) {
+      update: function (params) {
         $log.debug('LeaveRequestAPI.update');
+        var deferred = $q.defer();
 
-        return this.sendPOST('LeaveRequest', 'create', updatedAttributes);
+        if (!('id' in params)) {
```
AkA84: Why are you using `in`? Isn't it better to do `if (!params.id`)?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -55,24 +57,35 @@ define([
...
+          deferred.reject({
+            is_error: 1,
+            error_message: 'id is mandatory field'
+          });
+          return deferred.promise;
+        }
+
+        return this.sendPOST('LeaveRequest', 'create', params)
+          .then(function (data) {
+            //returns array of single object hence getting first object
```
AkA84: Have you tried passing `sequential: 0` in the `params`? `sequential` is used by the API to determine whether `values` should be an array or an object. In the base `api` service it's set to `1`, but you can always override it if you need to do so

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -47,14 +47,37 @@ define([
...
+        /**
+         * Create a new leave request
+         *
+         * @return {Promise} Resolved with {Object} Created Leave request with
+         *  newly created id for this instance
+         */
+        create: function (attributes) {
+          return LeaveRequestAPI.create(this.attributes())
+            .then(function(result){
+              this.attributes = _.assign({}, this.attributes(), result);
```
AkA84: Why are you not simply doing `this.id = result.id` (or similar)? If you replace all the current properties of the instance with the object returned by the API, then that means that you would have to send that object through the `fromAPI` filter. I'm not sure it's worth to do it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+          'amount': 1,
+          'type': {
+            'id': 1,
+            'value': 1,
+            'label': 'All Day'
+          }
+        }]
+      }
+    },
+    randomLeaveRequestFromServer: {
```
AkA84: What's the point of this? You already have a collection of mocked leave requests, why don't you pick one at random?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+      'is_error': 0,
+      'version': 3,
+      'count': 3,
+      'values': {
+        '1': -21,
+        '2': -1,
+        '3': -11
+      }
+    },
+    calculateBalanceChangeRequestData: {
```
AkA84: I don't think I agree with this. Mocked data files are for mocked data coming *from* the server, not the other way around. If you need to mock the data to send to the server, just do that in the test

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+        'type_id': absenceTypeData.getRandomAbsenceType('id'),
+        'contact_id': '202',
+        'status_id': optionGroupMock.randomValue('hrleaveandabsences_leave_request_status', 'value'),
+        'from_date': '2016-02-01',
+        'from_date_type': optionGroupMock.randomValue('hrleaveandabsences_leave_request_day_type', 'value'),
+        'to_date': '2016-02-03',
+        'to_date_type': optionGroupMock.randomValue('hrleaveandabsences_leave_request_day_type', 'value'),
+      }]
+    },
+    randomLeaveRequestToServer: {
```
AkA84: Same as before. Mocked data is from data coming from the server only. I don't think it even makes sense conceptually to "mock" a request to the server, you just set it up in the test, you have full control over it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+      'type_id': absenceTypeData.getRandomAbsenceType('id'),
+      'contact_id': '202',
+      'status_id': optionGroupMock.randomValue('hrleaveandabsences_leave_request_status', 'name'),
+      'from_date': '2016-02-01',
+      'from_date_type': optionGroupMock.randomValue('hrleaveandabsences_leave_request_day_type', 'name'),
+      'to_date': '2016-02-03',
+      'to_date_type': optionGroupMock.randomValue('hrleaveandabsences_leave_request_day_type', 'name'),
+    },
+    randomUpdateLeaveRequestToServer: {
```
AkA84: Same as before

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+        'from_date_type': '1',
+        'to_date': '2016-02-03',
+        'to_date_type': '1'
+      }, {
+        'id': '18',
+        'type_id': '1',
+        'contact_id': '202',
+        'status_id': '1',
+        'from_date': '2016-08-17',
+        'from_date_type': '1',
```
AkA84: Didn't we discuss making the "day type" option values mocked as well?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -1,101 +1,207 @@
...
+      'count': 2,
+      'values': {
+        'amount': -4,
+        'breakdown': [{
+          'date': '2016-11-05',
+          'amount': 0,
+          'type': {
+            'id': 4,
+            'value': 4,
+            'label': 'Weekend'
```
AkA84: Isn't the "type" an option value that you can mock?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -0,0 +1,328 @@
+define([
```
AkA84: It would keep things more manageable if instead of having a single, huge "option-group" file, you had smaller, individual files for each option group. A file for the day types, one for the leave request statuses, and so on.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -111,6 +217,48 @@ define(function () {
...
+    getAllRequestStatusesLabels: function () {
+      return optionGroupMock.getAllValuesForKey('hrleaveandabsences_leave_request_status', 'label');
+    },
+    getAllRequestStatusesValues: function () {
+      return optionGroupMock.getAllValuesForKey('hrleaveandabsences_leave_request_status', 'value');
+    },
+    getAllAbsenceTypesTitles: function () {
+      return absenceTypeData.getAllAbsenceTypesByKey('title');
+    },
+    getAllAbsenceTypesIds: function () {
```
AkA84: Why are all these functions (`getAllRequestDayTypes`, `getAllRequestDayValues`, `getAllRequestStatusesNames`, `getAllRequestStatusesLabels`, `getAllRequestStatusesLabels`, `getAllRequestStatusesValues`, `getAllAbsenceTypesTitles`,  `getAllAbsenceTypesIds`)  here, in the leave request mock data file, if they are about option groups?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -0,0 +1,328 @@
...
+      }]
+    },
+    /**
+     * Pick a random value out of a collection
+     *
+     * @param {array} the option group collection key
+     * @param {string} key - The sub-collection key
+     * @return {object}
+     */
+    randomValue: function (collection, key) {
```
AkA84: I'm really not sure about this. These mock data files were supposed to be pretty simple: a collection of mocked data that the tests were to use. But now we're adding these utility functions that I don't know if they should be here at all. Why don't you let the tests themselves to take care of that? If they need some random data, they can do it themselves.

Actually i see that you're using it in the mocked leave requests file. That's fine, but then don't add "noise" here, just put a private function in that file

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -0,0 +1,328 @@
...
+      return randomOptionValue[key];
+    },
+    /**
+     * Gets all values for given key in collection object
+     *
+     * @param {array} the option group collection key
+     * @param {string} key - The sub-collection key
+     * @return {Array} of values
+     */
+    getAllValuesForKey: function (collection, key) {
```
AkA84: Same thing that i wrote for `randomValue`, applies here

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -21,6 +25,34 @@ define([
...
+
+          var uriFilter = uriParts.filter(function (item) {
+            var itemSplit = item.split('=');
+            if (itemSplit[0] === 'entity' || itemSplit[0] === 'action') {
+              uriEntityAction[itemSplit[0]] = itemSplit[1];
+              return true;
+            }
+          });
+
+          //'update' is also 'create' call with id set
```
AkA84: What do you mean by "also"? Update is *only* done via "create"

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -21,6 +25,34 @@ define([
...
         .respond(mockData.balanceChangeByAbsenceType());
+
+      //Intercept backend calls for LeaveRequest.calculateBalanceChange
+      $httpBackend.whenGET(/action\=calculatebalancechange&entity\=LeaveRequest/)
+        .respond(mockData.calculateBalanceChange());
+
+      //Intercept backend calls for LeaveRequest.create in POST
+      $httpBackend.whenPOST(/\/civicrm\/ajax\/rest/)
+        .respond(function (method, url, data, headers, params) {
+          var uriParts = decodeURIComponent(data).split('&');
```
AkA84: There's a lot of stuff going here, why's that? In any case, please put all this logic in a separate function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -115,5 +147,259 @@ define([
...
+        promise = LeaveRequestAPI.calculateBalanceChange(requestData);
+      });
+
+      afterEach(function () {
+        $httpBackend.flush();
+      });
+
+      it('calls endpoint', function () {
+        promise.then(function (result) {
+          expect(LeaveRequestAPI.calculateBalanceChange).toHaveBeenCalled();
```
AkA84: What does this test? You have just called manually `calculateBalanceChange`, what's the point if the method has actually been called? You did that manually!

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -115,5 +147,259 @@ define([
...
+          //returns an object(associative array) and not an array
+          expect(result.amount).toBeDefined();
+          expect(result.breakdown).toBeDefined();
+
+          var breakdown = result.breakdown[0];
+          expect(breakdown.date).toBeDefined();
+          expect(breakdown.amount).toBeDefined();
+          expect(breakdown.type).toBeDefined();
+
+          var breakdownType = breakdown.type;
```
AkA84: Put both the `vars` at the beginning

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -115,5 +147,259 @@ define([
...
+        promise = LeaveRequestAPI.create(requestData);
+      });
+
+      afterEach(function () {
+        $httpBackend.flush();
+      });
+
+      it('call endpoint', function () {
+        promise.then(function (result) {
+          expect(LeaveRequestAPI.create).toHaveBeenCalled();
```
AkA84: Again, same thing. You just manually called the method *in the test*, what's the point of testing if it has indeed been called?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -110,27 +111,72 @@ define([
...
+
+      it('id is appended to instance', function () {
+        expect(requestData.id).not.toBeDefined();
+        promise.then(function (result) {
+          expect(result.id).toBeDefined();
+        });
+      });
+    });
+
+    describe('isValid()', function () {
```
AkA84: At no point, neither in the test for `create()`, for `update()`, nor for `isValid()`, you are actually testing the instance, all you are doing is testing the API. Hence these tests are basically pointless if we keep them like this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/leave-request-model_test.js
```diff
@@ -53,12 +54,36 @@ define([
...
 
-      leaveRequestPromise.then(function () {
-        expect(LeaveRequestAPI.balanceChangeByAbsenceType).toHaveBeenCalled();
+      afterEach(function () {
+        //to excute the promise force an digest
+        $rootScope.$apply();
+      });
+
+      it('calls equivalent API method', function () {
+        var leaveRequestPromise = LeaveRequest.balanceChangeByAbsenceType();
```
AkA84: This should go in a `beforeEach` block

AkA84: Also, this is not representative of how you use the method. You would never call the method without parameters in the real code, so why are you doing it in the test?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/leave-request-model_test.js
```diff
@@ -53,12 +54,36 @@ define([
...
+
+    describe('calculateBalanceChange()', function () {
+
+      afterEach(function () {
+        //to excute the promise force an digest
+        $rootScope.$apply();
+      });
+
+      it('calls equivalent API method', function () {
+        var leaveRequestPromise = LeaveRequest.calculateBalanceChange();
```
AkA84: Not representative of how you would use the method in the real code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/leave-request-model_test.js
```diff
@@ -53,12 +54,36 @@ define([
...
+
+        leaveRequestPromise.then(function () {
+          expect(LeaveRequestAPI.calculateBalanceChange).toHaveBeenCalled();
+        });
+      });
+});
```
AkA84: Empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
AkA84: I didn't get any answer to this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: as discussed will keep basic validation in front end

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: for now will keep it as it is, theres no harm in having 2 return statements as long as it makes code clean to understand

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: if key id is not there then params.id will throw exception

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: sequential 0 will return 
thank you for the tip,
using sequential 0 will return object indexed by id, so its not very useful
```json
{

    "is_error": 0,
    "version": 3,
    "count": 1,
    "id": 24,
    "values": {
        "24": {
            "id": "24",
            "type_id": "3",
            "contact_id": "202",
            "status_id": "3",
            "from_date": "20161213000000",
            "from_date_type": "",
            "to_date": "",
            "to_date_type": ""
        }
    }
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
dhavalsharma: this is for array of leave requests coming from server, no point mocking them as they will correspond to real data and not some random combination of types

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
dhavalsharma: if i mock it then a valid Weekend can become a public holiday or anything else.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
dhavalsharma: its a random response from server and not an real one like status_id is randomly generated and may not correspond to valid status_id coming from server

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
dhavalsharma: ps see above responses

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
dhavalsharma: they are for the option values relevant for the leaverequest context so internally they do wrap a call to options group like here 

```javascript
    getAllRequestStatusesValues: function () {
      return optionGroupMock.getAllValuesForKey('hrleaveandabsences_leave_request_status', 'value');
    }
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
dhavalsharma: that will unnecessary bloat number of files and managing their inclusion instead of having a single file and obtain relevant data from them

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
dhavalsharma: then we will be adding that private function (with same code) in each file whereever we will have to pick some random values

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
dhavalsharma: then we will be adding that private function (with same code) in each file whereever we will have to pick some random values

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
dhavalsharma: that is to distinguish this POST request from other POST requests as all have the same path /civicrm/ajax/rest and the entities/action are lost somewhere in data

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
dhavalsharma: i thought it was to test the endpoint call and then callthrough to the function (if the then succeeds then simply verify if it was due to the calculateBalanceChange call ), will have to discuss this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/leave-request-model_test.js
dhavalsharma: this is not in scope of this ticket but still added params to it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -1,137 +1,284 @@
...
+      });
+    });
+  });
+
+  describe('create()', function () {
+    var instanceCreate;
+
+    beforeEach(function () {
+      requestData = helper.createRandomLeaveRequest();
+      instance = LeaveRequestInstance.init(requestData, true);
```
AkA84: When you create any instance, the data does not come from the api (since you're creating it for the first time), so you shouldn't pass `true` to `init()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -1,137 +1,284 @@
...
+
+    it('calls equivalent API method', function () {
+      instanceValid.then(function () {
+        expect(LeaveRequestAPI.isValid).toHaveBeenCalled();
       });
+    });
 
-      it('calls LeaveRequestAPI.sendPOST', function () {
-        expect(LeaveRequestAPI.sendPOST).toHaveBeenCalledWith('LeaveRequest', 'create', _.assign({}, LeaveRequestInstance.attributes(), attr));
+    describe('when leave request is valid', function () {
```
AkA84: I'm not sure how you are testing this? The result of `isValid()` is all done on the backend, you can really only test if `instance.isValid()` is calling the api's `isValid()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
dhavalsharma: in the call we also agreed that functions that belong to their respective context will be kept there , like all functions related to optiongroups will in optiongroups file , leaverequest in that leaverequest file etc., hence will keep it in optiongroup file only

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -0,0 +1,360 @@
...
+    },
+    /**
+     * Pick a specific value out of a collection
+     *
+     * @param {array} the option group collection key
+     * @param {string} key - The sub-collection key
+     * @param {string} value - The sub-collection key's value to match
+     * @return {object}
+     */
+    specificValue: function (collection, key, value) {
```
dhavalsharma: lets take example 
`
optionGroupMock.specificValue('hrleaveandabsences_leave_request_day_type', 'value', '1'),
`
here `value` is the key, `1` is the value and method is returning a value and hence is called `specificValue`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
dhavalsharma: testing the various responses which will be sent by the server for isValid calls

PR #1300, opened 2016-12-20
PCHR-1829: Improve Leave Request validation

tunbola:
In this PR, four additional validations are added to the LeaveRequest create BAO method.
They include:

- Check for empty contact_id
- Check for empty type_id
- Check for empty status_id
- Check to ensure that the to_date_type field should be mandatory if to_date is not empty

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -65,6 +65,38 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       );
     }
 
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -65,6 +65,38 @@ private static function validateParams($params) {
...
+    }
+
+    if (!empty($params['to_date']) && empty($params['to_date_type'])) {
+      throw new InvalidLeaveRequestException(
+        'The type of To Date should not be empty',
+        'leave_request_empty_to_date_type',
+        'to_date_type'
+      );
+    }
+
```
davialexandre: This method got quite big now with all those new rules. Can you please refactor it and put the validations into separate methods? You can group them by type of validation. For example, you can group all the validations for mandatory fields into a `validateMandatory()` method

PR #1302, opened 2016-12-20
PCHR-1768: Fix Contract Duration, and Pension Types in Summary

MiyaNoctem:
## Problem
Calculation of duration was one day less than it should be, noting that both first and last working day must be included.

Also, on saving contract, summary screen showed life insurance type twice and pension type was shown as value instead of label. 

pchr-1768-1-before

pchr-1768-2-before

## Solution
Fixed calculation of duration by adding 1 day to the result.  

Also, added joins to option_value table when building query to obtain contract details so pension type label can be used on summary.  Fixed double insurance type in contract summary view.

pchr-1768-1-after

pchr-1768-2-after


hrjobcontract/js/src/job-contract/controllers/form/form-general.js
```diff
@@ -89,7 +89,7 @@ define([
         months = m.diff(dateStart, 'months');
 
         m.add(-months, 'months');
-        days = m.diff(dateStart, 'days');
+        days = m.diff(dateStart, 'days') + 1;
```
davialexandre: Do you think you can add a test for this, to make sure we won't get a regression?

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -262,6 +262,17 @@ protected static function buildFullDetailsQuery($entities, $revision) {
...
+      if ($entity == 'pension') {
+        $query['select'][] = "pension_ov.label as {$entity}__pension_type_label";
+
+        $query['from'][] = 'civicrm_option_group pension_og';
+        $query['from'][] = 'civicrm_option_value pension_ov';
+
+        $query['where'][] = "pension_og.name = 'hrjc_pension_type'";
+        $query['where'][] = 'pension_ov.option_group_id = pension_og.id';
+        $query['where'][] = "pension_ov.value = {$table}.pension_type";
+      }
```
davialexandre: Did you run the tests for the hrjobcontract extension? It seems this change is making one of them fail:

screen shot 2016-12-21 at 08 32 45

omarabuhussein: sorry I missed this .. @MiyaNoctem  , can u please not just fixing this tests .. but also write some tests for fullDetails BAO method .. it would make more sense to test the BAO method directly

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
davialexandre: I was still thinking about this and wondering if this whole method could be implemented in a better way. Maybe [Moment.js Humanize](http://momentjs.com/docs/#/durations/humanize/) can help with that?

davialexandre:
@MiyaNoctem I just realized something. Since you updated one of the javascript files, you also need to update the minified version

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
MiyaNoctem: I played around a while with moment.js humanize method, but it really doesn't do what we'd need it to.  Even tried to humanize each part of the date (years, months, days) independently, but strange things happened on frontier cases, like a duration of 31 days, for example which got rendered as 30 days and a day.  You can use Relative Time Thresholds to set the value where a set of days gets rounded to a month, but you'd have to start to juggle it before and after you call humanize for months and days, which is kind of awkward.  There's a plugin for moment.js that does what we need, but it basically does the same thing our method does, so I don't know if it's woth it. http://momentjs.com/docs/#/plugins/preciserange/

omarabuhussein: handling date and time in JS is annoying generally

hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/HRJobPension.php
```diff
@@ -2,4 +2,37 @@
...
+      
+      if ($pensionTypeResult['count'] == 0) {
+        self::createPensionType($params['pension_type']);
+      }
+    }
+    
+    return parent::fabricate($params);
+  }
+  
+  public static function createPensionType($value) {
```
davialexandre: This should be private

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -15,7 +23,79 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevisionTest extends PHPUnit_Framework_
...
+
+  /**
+   * Obtains label for given pension type.
+   * 
+   * @param type $pension_type
+   *   ID of pension type
+   * @return string
+   *   Label for given pension type
+   */
+  private function getPensionTypeLabel($pension_type) {
```
davialexandre: Please use camelCase for the param

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -15,7 +23,79 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevisionTest extends PHPUnit_Framework_
...
+
+    $this->assertInternalType("array", $fullDetails['leave']);
+  }
+
+  /**
+   * Obtains label for given pension type.
+   * 
+   * @param type $pension_type
+   *   ID of pension type
+   * @return string
```
davialexandre: Add 1 space between @param and @return sections

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -15,7 +23,79 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevisionTest extends PHPUnit_Framework_
...
+
+    return $pensionTypeResult['label'];
+  }
+
+  /**
+   * Obtains current revision for given contract.
+   * 
+   * @param type $jobContractID
+   *   ID of job contract
+   * @return array
```
davialexandre: Add 1 space here as well

hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/HRJobPension.php
omarabuhussein: this is implementation details and should be hidden , it should be private

hrjobcontract/js/src/job-contract/controllers/form/form-general.js
davialexandre: handling date is annoying in every language :D 

@MiyaNoctem thanks for trying other options

davialexandre:
@MiyaNoctem looking at the screenshots, I see you not only removed the duplication, but also removed information from both life insurance and health insurance (description and number of dependents). Is that correct? Oy maybe the screenshots are from different contracts

hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/HRJobPension.php
omarabuhussein: this is implementation details and should be hidden , it should be private

MiyaNoctem:
@davialexandre, I was ussing screenshots from different contacts. I've updated them using one same contact, to avoid ambiguity.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -15,7 +23,81 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevisionTest extends PHPUnit_Framework_
...
+
+    $this->assertEquals($this->getPensionTypeLabel($pensionDetails['pension_type']), $fullDetails['pension']['pension_type_label']);
+
+    $this->assertInternalType("array", $fullDetails['leave']);
+  }
+
+  /**
+   * Obtains label for given pension type.
+   * 
+   * @param type $pension_type
```
davialexandre: you forgot to update the param name here

davialexandre: and also fix the @param type

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/BAO/HRJobContractRevisionTest.php
```diff
@@ -15,7 +23,81 @@ class CRM_Hrjobcontract_BAO_HRJobContractRevisionTest extends PHPUnit_Framework_
...
+      'value' => $pensionType
+    ]);
+
+    return $pensionTypeResult['label'];
+  }
+
+  /**
+   * Obtains current revision for given contract.
+   * 
+   * @param type $jobContractID
```
davialexandre: fix the type here as well. It should be int instead of type

hrjobcontract/js/test/controllers/form/form-general_test.js
```diff
@@ -73,6 +73,15 @@ define([
         it('sets duration calculating number of days including start and end dates', function() {
           expect($scope.duration).toEqual('17 days');
         });
+
+        it('calculates duration from 2017-02-01 to 2017-03-31 as two motnhs', function() {
```
davialexandre: I think this is too specific to be a good test name. It makes me think that it should only work for these 2 specific dates and not for everything else. Can you please make this more generic and try to actually describe what is the rule? As far as I understand, every 30 days should be considered a month, right?

PR #1303, opened 2016-12-21
PCHR-1761: Create Leave Request isValid API

tunbola:
This PR creates the LeaveRequest isValid API. The API uses the validateParams() method of the LeaveRequest BAO but does not make a call to LeaveRequest BAO create method.

The Response of the API includes the DAO field associated the error and also the error/Exception code for the exception.

Here are some sample request and responses for the API endpoint: 
_When from_date field is not present_
```php
$result = civicrm_api3('LeaveRequest', 'isvalid', array(
  'sequential' => 1,
  'contact_id' => 1,
  'status_id' => 1,
));
```
Response: 
```json
{
  "is_error": 0,
  "version": 3,
  "count": 1,
  "values": {
    "from_date": [
      "leave_request_empty_from_date"
    ]
  }
}
```
_When the absence type field is not included in the payload_
```php
$result = civicrm_api3('LeaveRequest', 'isvalid', array(
  'sequential' => 1,
  'contact_id' => 1,
  'status_id' => 1,
  'from_date' => "2016-12-10",
  'from_date_type' => 1,
));
```
Response: 
```json
{
  "is_error": 0,
  "version": 3,
  "count": 1,
  "values": {
    "type_id": [
      "leave_request_empty_type_id"
    ]
  }
}
```
When all validations pass the response is:

```json
{
  "is_error": 0,
  "version": 3,
  "count": 0,
  "values": {}
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
@@ -698,7 +698,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequestTest extends BaseHeadlessTest {
    * @expectedExceptionMessage Leave Requests should have a start date
    */
   public function testALeaveRequestShouldNotBeCreatedWithoutAStartDate() {
-    LeaveRequest::create([
+    LeaveRequest::validateParams([
       'type_id' => $this->absenceType->id,
       'contact_id' => 1,
       'status_id' => 1,
@@ -712,7 +712,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequestTest extends BaseHeadlessTest {
    */
   public function testALeaveRequestShouldNotBeCreatedWithoutContactID() {
     $fromDate = new DateTime('+4 days');
-    LeaveRequest::create([
+    LeaveRequest::validateParams([
       'type_id' => $this->absenceType->id,
       'status_id' => 1,
       'from_date' => $fromDate->format('YmdHis'),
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -698,7 +698,7 @@ public function testCalculateBalanceChangeForALeaveRequestForAContactWithMultipl
    * @expectedExceptionMessage Leave Requests should have a start date
    */
   public function testALeaveRequestShouldNotBeCreatedWithoutAStartDate() {
-    LeaveRequest::create([
+    LeaveRequest::validateParams([
```
davialexandre: Overall comment about the changes on this file: The test names do not match what it's being tested anymore. The test names mention the creation of a LeaveRequest, but now we're only testing the validation, so the names have to be updated. Also, we need at least one test to check that validateParam() will be called inside the create() method.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -696,4 +703,449 @@ public function testCalculateBalanceChangeWithAllRequiredParameters() {
     ]);
     $this->assertEquals($expectedResultsBreakdown, $result['values']);
   }
+
+  public function testLeaveRequestIsValidWithoutAStartDate() {
+    $result = civicrm_api3('LeaveRequest', 'isvalid', [
```
davialexandre: Overall comment about the tests on this class: The test names are confusing. They're saying the opposite of what is a valid Leave Request. Let's take this one here as an example. If you remove the "test" prefix, it says: "Leave Request Is Valid Without A Start Date", which is wrong. Note that the "test" prefix is only that so phpunit knows it's a test and it's not part of the test name.

Take a look of how the tests will look like when you run phpunit4 with the --testdox option:
screen shot 2016-12-21 at 12 29 31

A suggestion for a better names would be something like "testLeaveRequestIsValidShouldReturnErrorWhenStartDateIsEmpty"

PR #1304, opened 2016-12-22
PCHR-1300: Leave Request HTML (Static)

deb1990:
This PR contains the Static HTML, basic controller logic and updated test cases for Leave Request Modal

The styling for this PR can be found in https://github.com/compucorp/civihr-employee-portal-theme/pull/123

**Screencast**:
pchr1300

**Note**:
1. The date range picker mentioned in the mockup could not be implemented because Angular does not have a suitable and well-maintained plugin for the same.

2. The controller for the modal is done temporarily and is not final, as the scope of the ticket is for only the HTML + CSS.


uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-xs-12 form-horizontal">
+              <div class="form-group">
+                <label class="col-xs-4 control-label">From Date
+                  <i class="fa fa-asterisk"></i>
```
AkA84: That's a bit of an overkill, just use a "*"

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+          <option value="1" selected>Vacation (15 days available)</option>
+          <option value="2">Medical</option>
+        </select>
+      </div>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col-xs-12">
+      <uib-tabset class="chr-tab__outer">
```
AkA84: Why do we need this `.chr_tab__outer`? As far as I understand you're simply overriding the default style of bootstrap tabs, so then why you just don't do that, instead of creating a new component?

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+
+        <uib-tab heading="Details">
+          <div class="row">
+            <div class="col-xs-12 form-horizontal">
+              <div class="form-group">
+                <label class="col-xs-4 control-label">No. of Days</label>
+                <div class="col-xs-8">
+                  <div class="btn-group chr-toggle btn-group-justified" data-toggle="buttons">
```
AkA84: Same thing that I've said for `.chr_tab__outer`, here it seems you're just customizing `.btn-group`, then can't you just do that? Is it possible to imagine that every button group will have this styling?

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-xs-12 form-horizontal">
+              <div class="form-group">
+                <label class="col-xs-4 control-label">From Date
+                  <i class="fa fa-asterisk"></i>
+                </label>
+                <div class="col-xs-4">
+                  <div class="input-group chr-datepicker">
```
AkA84: About `.chr-datepicker`: this seems to me another instance of a custom component that doesn't really need to exist. What you are doing is styling `.uib-datepicker-popup`, so then just style that, instead of going through some custom classes

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+                      <option value="1" selected>1/2 - PM</option>
+                      <option value="2">1/2 - AM</option>
+                    </select>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+          <hr/>
+          <div class="row form-group">
```
AkA84: Here you are using `.form-group` in order to have the margin bottom, but it is semantically incorrect, as you do not have any form controls.

Why couldn't you have something like `.chr_leave-request-form` applied to the whole modal body, which you can then use to give layout rules? You could have then elements like `.chr_leave-request-form__balance` which then defines the layout rules of this section

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+            <div class="col-xs-4">Opening Balance</div>
```
AkA84: This is actually, semantically, a definition list
```html
<dl>
  <dt>Opening Balance</dt>
  <dd>24.0</dd>
</dl>
```
but i'll leave it to you to see how easy it is to style

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+            <div class="col-xs-8">17.0</div>
+          </div>
+          <div class="row form-group text-primary pointer"
+               ng-click="modal.leaveRequest.isChangeExpanded = !modal.leaveRequest.isChangeExpanded">
+            <div class="col-xs-4">
+              <i class="fa chr-inline-icon"
```
AkA84: Don't agree with this custom class as well. Just to be clear: a `chr_{xyz}` (please note that the prefix is `chr_`, not `chr-`) should be created only when you have to do something *from scratch*

Here what you are really doing is 1) having a font awesome icon smaller than the default size 2) giving layout rules to the icon

This is something that you could achieve easily if you used the `.chr_leave-request-form` class (or similar) that i mentioned above. You could have then something like:
```scss
.chr_leave-request-form__legend {
  
  .fa {

  }
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+                 aria-hidden="true"
+                 ng-class="{ 'fa-chevron-right': !modal.leaveRequest.isChangeExpanded, ' fa-chevron-down': modal.leaveRequest.isChangeExpanded }"></i>
+              Change
+            </div>
+            <div class="col-xs-8">2.5</div>
+          </div>
+          <table class="table chr-table" ng-show="modal.leaveRequest.isChangeExpanded">
```
AkA84: Another `chr-` (which should be `chr_`) that you are using for layout rules. Use `.chr_leave-request-form` for that

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
```diff
@@ -0,0 +1,177 @@
...
+            <thead>
+            <tr>
+              <th class="">Date</th>
```
AkA84: What's this `[class]` for?

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
deb1990: The Report/Calender tabs are using default bootstrap tabs which have different styling. So separated this one.

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
deb1990: Record New Absence would then need a specific class to style. Created `chr-toggle` for the toggle component, and normal button group is reserved for Dropdown buttons.

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
deb1990: We need a class or element type to style the asterisk, I tried with `span` before but its not a good idea because we can have span for normal text also. So went for `fontawesome` so that we can leverage `i` tag to style the asterisk.

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
deb1990: Changed the name, but keeping `chr_table` as this table style would be resued for other pages also.

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html
deb1990: I really do not want to add new markup and style it to change the look. Keeping it

PR #1305, opened 2016-12-22
PCHR-1762: Create TOIL Request isValid API

tunbola:
This PR creates the TOILRequest isValid API. The API uses the validateParams() method of the TOILRequest BAO but does not make a call to TOILRequest BAO create method.

The Response of the API includes the field associated the error and also the error/Exception code for the exception.

Here are some sample request and responses for the API endpoint:

_The TOIL amount is not valid_

```php
$result = civicrm_api3('TOILRequest', 'isvalid', array(
  'type_id' => 1,
  'contact_id' => 1,
  'status_id' => 1,
  'from_date' => '2016-11-14',
  'toil_to_accrue' => 200,
  'duration' => 120
));
```

Response: 
```json
{
  "is_error": 0,
  "version": 3,
  "count": 1,
  "values": {
    "toil_to_accrue": [
      "toil_request_toil_amount_is_invalid"
    ]
  }
}
```

When all validations pass the response is:

```json
{
  "is_error": 0,
  "version": 3,
  "count": 0,
  "values": {}
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
@@ -83,7 +83,11 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
   private static function validateTOILAmountIsValid($params) {
     $toilAmountOptions = self::toilAmountOptions();
     if (!in_array($params['toil_to_accrue'], $toilAmountOptions)) {
-      throw new InvalidTOILRequestException("The TOIL amount is not valid");
+      throw new InvalidTOILRequestException(
+        'The TOIL amount is not valid',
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
```diff
@@ -129,7 +135,9 @@ private static function validateValidTOILPastDaysRequest($params) {
 
     if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past) {
       throw new InvalidTOILRequestException(
-        "You cannot request TOIL for past days"
+        'You cannot request TOIL for past days',
+        'toil_request_toil_cannot_be_requested_for_past_days',
+        'toil_to_accrue'
```
davialexandre: wouldn't it make more sense to use `from_date` as the field here?

tunbola: You right.

PR #1311, opened 2016-12-26
PCHR-1681: Create/Delete LeaveBalanceChange related to Leave and Sickness Requests

davialexandre:
Whenever a Leave Request gets updated, we must also update the Leave Balance Changes caused by it. This should not be the responsibility of the Leave Request BAO though. That BAO should care only about itself and its direct dependents, the Leave Request Dates. Besides that, Leave Balance Changes are also used for other purposes, like storing information about the Entitlement and expired leaves. For these reasons, it's interesting to separate this from the Leave Request BAO and try to encapsulate the logic somewhere else.

The LeaveRequest service was created to achieve this isolation. It has two methods: `create` and `delete`. The former, basically, creates the LeaveRequest using the BAO's `create` method, and then uses the `LeaveRequest::calculateBalanceChange` method in order to get the balance change amounts for each of the request dates and saves the Balance Changes.  The `delete` method does the opposite, making sure to delete everything linked to a LeaveRequest (Balance Change and dates) and finally the Leave Request itself.

A SicknessRequest service was also created and it does pretty much the same thing. When creating a new SicknessRequest, an associated LeaveRequest is created (to store the dates, status, absence type etc). This is done by the BAO. The SicknessRequest service will be responsible to create the Balance Changes for the associated LeaveRequest.

Finally, the `create` and `delete` methods of both the LeaveRequest and SicknessRequest APIs where updated to call the respective service methods intead of the BAO methods.

### What else has changed?

How a SicknessRequest uses a LeaveRequest internally to store additional information is just an implementation detail and must be hidden from outside world. For that reason, the `SicknessRequest.create` API was customized a bit to work just like the `LeaveRequest.create` endpoint. That is, the response from `SicknessRequest.create` will be exactly like the one from LeaveRequest api and it will include of the LeaveRequest fields together with the SicknessRequest fields

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -227,9 +227,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   private static function calculateBalanceChangeFromCreateParams($params) {
     $leaveRequestOptionsValue = self::getLeaveRequestDayTypeOptionsGroupByValue();
 
-    $toDate = !empty($params['to_date']) ? new DateTime($params['to_date']) : null;
     $fromDateType = $leaveRequestOptionsValue[$params['from_date_type']];
-    $toDateType = !empty($params['to_date_type']) ? $leaveRequestOptionsValue[$params['to_date_type']] : '';
+    $toDate = !empty($params['to_date']) ? new DateTime($params['to_date']) : null;
+    $toDateType = !empty($params['to_date_type']) ? $leaveRequestOptionsValue[$params['to_date_type']] : null;
 
     $leaveRequestBalance = self::calculateBalanceChange(
       $params['contact_id'],
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -225,19 +225,20 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @return float
    */
   private static function calculateBalanceChangeFromCreateParams($params) {
-    $toDate = !empty($params['to_date']) ? $params['to_date'] : '';
     $leaveRequestOptionsValue = self::getLeaveRequestDayTypeOptionsGroupByValue();
+
+    $toDate = !empty($params['to_date']) ? new DateTime($params['to_date']) : null;
     $fromDateType = $leaveRequestOptionsValue[$params['from_date_type']];
     $toDateType = !empty($params['to_date_type']) ? $leaveRequestOptionsValue[$params['to_date_type']] : '';
-    $leaveRequestBalance =
-      self::calculateBalanceChange
-      (
-        $params['contact_id'],
-        $params['from_date'],
-        $fromDateType,
-        $toDate,
-        $toDateType
-      );
+
+    $leaveRequestBalance = self::calculateBalanceChange(
+      $params['contact_id'],
+      new DateTime($params['from_date']),
+      $fromDateType,
+      $toDate,
+      $toDateType
+    );
+
     return abs($leaveRequestBalance['amount']);
   }
 
@@ -534,22 +535,20 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @return array
    *   An array of formatted results
    */
-  public static function calculateBalanceChange($contactId, $fromDate, $fromType, $toDate = null, $toType = null) {
-
+  public static function calculateBalanceChange($contactId, DateTime $fromDate, $fromType, DateTime $toDate = null, $toType = null) {
     $leaveRequest = new self();
     $leaveRequest->contact_id = $contactId;
 
     //For single day leave requests
     if (!$toDate) {
-      $toDate = $fromDate;
+      $toDate = clone $fromDate;
     }
-    $startDate = new DateTime($fromDate);
-    $endDate = new DateTime($toDate);
-    $endDateUnmodified = new DateTime($toDate);
+
+    $endDateUnmodified = clone $toDate;
     // add one day to end date to include it in DatePeriod
-    $endDate->modify('+1 day');
+    $toDate->modify('+1 day');
     $interval   = new DateInterval('P1D');
-    $datePeriod = new DatePeriod($startDate, $interval, $endDate);
+    $datePeriod = new DatePeriod($fromDate, $interval, $toDate);
 
     $isHalfDay = ['half_day_am', 'half_day_pm'];
     $fromDateIsHalfDay = in_array($fromType, $isHalfDay);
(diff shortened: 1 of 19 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChange.php
```diff
@@ -0,0 +1,48 @@
...
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return array
+   */
+  private function calculateBalanceChanges(LeaveRequest $leaveRequest) {
+    return LeaveRequest::calculateBalanceChange(
+      $leaveRequest->contact_id,
+      new DateTime($leaveRequest->from_date),
+      $leaveRequest->from_date_type,
+      empty($leaveRequest->to_date) ? null : new DateTime($leaveRequest->to_date),
```
omarabuhussein: same here , use !empty

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChange.php
```diff
@@ -0,0 +1,48 @@
...
+      $leaveRequest->to_date_type
```
omarabuhussein: also shouldn't you check for to_date_type if exist as for to_date

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/SicknessRequest.php
```diff
@@ -0,0 +1,86 @@
...
+use CRM_HRLeaveAndAbsences_BAO_SicknessRequest as SicknessRequest;
+use CRM_HRLeaveAndAbsences_Test_Fabricator_LeaveBalanceChange as LeaveBalanceChangeFabricator;
+
+class CRM_HRLeaveAndAbsences_Test_Fabricator_SicknessRequest {
+
+  private static $leaveRequestStatuses;
+  private static $dayTypes;
+  private static $sicknessReasons;
+
+  public static function fabricate($params, $withBalanceChanges = false) {
```
omarabuhussein: why not  $withBalanceChanges just be a parameter ( you can unset it inside the method ) to keep consistency  ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/SicknessRequest.php
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Creates a new Sickness Request without running any validation
+   *
+   * @param array $params
+   * @param boolean $withBalanceChanges
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_SicknessRequest
+   */
+  public static function fabricateWithoutValidation($params = [], $withBalanceChanges = false) {
```
omarabuhussein: same here for $withBalanceChanges parameter

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/SicknessRequest.php
davialexandre: Because I would be using $params for something different from its original purpose. That parameter should only contain data to build the Entity and nothing else.

Also, using a separate param, I can easily declare a default value. If I as to use $params instead of $withBalanceChange, I would need to add it to a "defaultParams" array and then merge it into the $params array and, as you mentioned, remember to manually unset it.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChange.php
davialexandre: No. The check for to_date is required because:

- calculateBalanceChange expects a DateTime object for that param
- We don't want to pass a null to the DateTime constructor

On the other hand, calculateBalanceChange expects a string for the to_date_type param, including null, so there's no need to check for anything

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -0,0 +1,53 @@
...
+    $leaveRequestDateTypes = array_flip(LeaveRequest::buildOptions('from_date_type', 'validate'));
+
+    // a 7 days leave request, from monday to sunday
+    $leaveRequest = $leaveRequestService->create([
+      'type_id' => 1,
+      'contact_id' => $contact['id'],
+      'status_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'from_date_type' => $leaveRequestDateTypes['all_day'],
+      'to_date' => CRM_Utils_Date::processDate('2016-01-07'),
```
davialexandre: Yeah, I thought about that (and about adding a lot of tests for other different scenarios) but the problem is that this method is just a wrapper around LeaveRequestBAO.create(). If I add a test like that, I would only be duplicating the BAO's tests, increasing the time to run the tests and not really adding any value. Ideally, the only thing that would have to be tested here is if the service calls the BAO create method with the right data or not. Thanks to Civi that is a static method and cannot be mocked. So, the tests only focus on the things that the service does additionaly to creating the LeaveRequest: creating the balance changes.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -534,22 +535,20 @@ private function deleteDates() {
...
+
+    $endDateUnmodified = clone $toDate;
     // add one day to end date to include it in DatePeriod
-    $endDate->modify('+1 day');
+    $toDate->modify('+1 day');
     $interval   = new DateInterval('P1D');
-    $datePeriod = new DatePeriod($startDate, $interval, $endDate);
+    $datePeriod = new DatePeriod($fromDate, $interval, $toDate);
 
     $isHalfDay = ['half_day_am', 'half_day_pm'];
```
davialexandre: Don't think so. They are names of reserved option values, so the risk of change is very low.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -21,6 +22,27 @@ public function create($params, $validate = true) {
...
+    $transaction = new CRM_Core_Transaction();
+    try {
+      LeaveBalanceChange::deleteAllForLeaveRequest($leaveRequest);
+      LeaveRequestDate::deleteDatesForLeaveRequest($leaveRequest->id);
+      $leaveRequest->delete();
+
+      $transaction->commit();
+    } catch(Exception $e) {
+      $transaction->rollback();
+    }
```
davialexandre: Unneeded. This is kinda like a wrapper around the BAO `del()` method, which doesn't return anything

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -838,6 +838,26 @@ public function testCanDeleteTheBalanceChangeForALeaveRequestDate() {
...
+      'status_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'from_date_type' => 1,
+      'to_date' => CRM_Utils_Date::processDate('2016-01-07'),
+      'to_date_type' => 1,
+    ], true);
+
+    $balanceChanges = LeaveBalanceChange::getBreakdownForLeaveRequest($leaveRequest);
+    $this->assertCount(7, $balanceChanges);
+define( 'CIVICRM_DEBUG_LOG_QUERY', 1 );
```
davialexandre: This was removed later, in another commit and it's not on the final PR

PR #1314, opened 2016-12-27
PCHR-1770: Extend TOILRequest.get API to filter expired requests

davialexandre:
This PR adds the "expired" filter to the TOILRequest.get API. When this filter is set, only expired TOILRequests will be returned, otherwise all will be returned.

Example usage:
```php
civicrm_api3('TOILRequest', 'get', ['expired' => true]);
```

The `TOILRequestSelect` was introduced to make this filter work. It wraps the `Civi\API\SelectQuery` class in order to join the TOILRequest with its LeaveBalanceChanges and be able to check if it's expired or not.

### What else changed?

While working on this ticket, I found a few problems with TOILRequest and SicknessRequest apis and BAOs. They were all fixed:
- The spec for the contact_id field of the SicknessRequest.create API had the wrong class name. 
- The TOILRequest BAO was missing the validation of mandatory fields (duration and toil_to_accrue)
- The TOILRequest BAO did not support receiving an expiry date to set it on the LeaveBalanceChange]
- The TOILRequest.create API was nothing returning all of the LeaveRequest fields on the response
- The LeavePeriodEntitlement BAO tests had some bugs which would make them fail close to the end of the year

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/TOILRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/TOILRequestSelect.php
@@ -1,0 +1,104 @@
+<?php
+
+use Civi\API\SelectQuery;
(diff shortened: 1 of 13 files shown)


coldrunKacper:
Approved.

PR #1315, opened 2016-12-28
PCHR-1377: Updating current contract

sleekvick:
# Job contract should be current if end date is today

pchr 1377

When users edit the end date of a contract to be the today it shows up as the past contract, also when users create a new contract with the end date set to today it shows up to be a past contract. 

# Setting end date of new contract today

In the case of creating a new contract and setting the end date to today, the problem is not taking into concern the time difference in comparing the date set by the user and the current date.

`contract.is_current = !entityDetails.period_end_date || new Date(entityDetails.period_end_date) > new Date();` 

Javascript will set the time of every instance of the date object to the current time of the user's machine, also our users don't set the time for the end date. Finally, the using just > for comparing is not sufficient.

# Editing the end date of contract

In this case, whenever the user changes the end date, the modal-contract controller already modifies the `$scope.details.period_end_date`  to what the user has set when the post request sent to the server has been resolved. 

Therefore checking that the end date has changed will always return false.

`if ($scope.details.period_end_date ?  new Date($scope.details.period_end_date).getTime() !== new Date(newEndDate).getTime() : !!$scope.details.period_end_date !== !!newEndDate)`

# Updating contract list with revision 

```
if (isCurrentRevision) {		
   updateContractList(results.details.period_end_date);		
 }
```
If the contract is a new contract and the user decides to change the contract terms immediately with a revison date the if condition above evaluates to false because **$scope.revisionCurrent.effective_date** is undefined. 

# Setting current contract if the period end date is today.

By removing the if block in **updateContractList** function and changing the conditions for a current contract in modal-contract-new,  the bug is eliminated.

Furthermore, moving the execution of the **updateContractList** to the top of the on the same level with **updateContractView** and **ContractService.updateHeaderInfo();** ensures that the contact list update with or without revision.

pchr-1377


hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -69,25 +69,17 @@ define([
...
-                        if (isCurrentContract) {
-                            $scope.$parent.contract.is_current = '1';
-                            $scope.$parent.$parent.contractCurrent.push($scope.$parent.contract);
-                            $scope.$parent.$parent.contractPast.splice($scope.$parent.$parent.contractPast.indexOf($scope.$parent.contract),1);
-                        } else {
-                            $scope.$parent.contract.is_current = '0';
-                            $scope.$parent.$parent.contractPast.push($scope.$parent.contract);
-                            $scope.$parent.$parent.contractCurrent.splice($scope.$parent.$parent.contractCurrent.indexOf($scope.$parent.contract),1)
-                        }
+                var isCurrentContract = !newEndDate || new Date(newEndDate).setHours(0,0,0,0) >= new Date().setHours(0,0,0,0);
```
AkA84: Can't we use [Moment.js](http://momentjs.com/) for this kind of things? It would make the code more readable. You can require it by defining `common/moment` as part of the file dependencies

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -69,25 +69,17 @@ define([
...
+
+                if (isCurrentContract != !!+$scope.$parent.contract.is_current) {
```
AkA84: What's the need for `!!+`? `!!` should be sufficient

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -69,25 +69,17 @@ define([
             };
 
             function updateContractList(newEndDate){
```
AkA84: I think you understand now what this function is for and when it is used, please write the documentation for it
```js
/**
 * Lorem ipsum
 *
 * @param {type} newEndDate description
 */
function updateContractList(newEndDate) {

File: hrjobcontract/js/src/job-contract/controllers/contract.js [hunk 4]
```diff
@@ -69,25 +69,17 @@ define([
...
+                var isCurrentContract = !newEndDate || new Date(newEndDate).setHours(0,0,0,0) >= new Date().setHours(0,0,0,0);
+
+                if (isCurrentContract != !!+$scope.$parent.contract.is_current) {
+                    if (isCurrentContract) {
+                        $scope.$parent.contract.is_current = '1';
+                        $scope.$parent.$parent.contractCurrent.push($scope.$parent.contract);
+                        $scope.$parent.$parent.contractPast.splice($scope.$parent.$parent.contractPast.indexOf($scope.$parent.contract),1);
+                    } else {
+                        $scope.$parent.contract.is_current = '0';
+                        $scope.$parent.$parent.contractPast.push($scope.$parent.contract);
```
AkA84: This `$scope.$parent{1,}` is just awful. I remember encountering some time ago and in the end found out that the `$scope` already inherits all the data of the parent scope without having to access the `$parent` object, so in theory `$scope.contractPast` should work just as fine.

Could you check if that's true? Can we make this code better? Surely there's must be a better approach rather than accessing the `$parent` object an arbitrary number of times

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.js
```diff
@@ -152,7 +152,7 @@ define([
                                 uploader = $scope.uploader,
                                 revisionId;
 
-                            contract.is_current = !entityDetails.period_end_date || new Date(entityDetails.period_end_date) > new Date();
+                            contract.is_current = !entityDetails.period_end_date || new Date(entityDetails.period_end_date).setHours(0,0,0,0) >= new Date().setHours(0,0,0,0);
```
AkA84: Same thing about using Moment.js

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -69,25 +69,17 @@ define([
...
-                        } else {
-                            $scope.$parent.contract.is_current = '0';
-                            $scope.$parent.$parent.contractPast.push($scope.$parent.contract);
-                            $scope.$parent.$parent.contractCurrent.splice($scope.$parent.$parent.contractCurrent.indexOf($scope.$parent.contract),1)
-                        }
+                var isCurrentContract = !newEndDate || new Date(newEndDate).setHours(0,0,0,0) >= new Date().setHours(0,0,0,0);
+
+                if (isCurrentContract != !!+$scope.$parent.contract.is_current) {
+                    if (isCurrentContract) {
+                        $scope.$parent.contract.is_current = '1';
```
AkA84: Please convert the indentation to 2-spaces on whatever piece of old code you work on

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -68,28 +69,26 @@ define([
...
-                            $scope.$parent.$parent.contractCurrent.splice($scope.$parent.$parent.contractCurrent.indexOf($scope.$parent.contract),1)
-                        }
-                    }
-                }
+            /**
+              * Updates the contract list view,
+              * by sorting the contract into current or past
+              * depending on the period end date of the contract.
+              *
+              *
```
AkA84: Unnecessary empty line

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -68,28 +69,26 @@ define([
...
+              * @param {type} newEndDate the date specified by the user
```
AkA84: You are not specifying the `type` here. Is it `string`? `Date`?

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -68,28 +69,26 @@ define([
...
+            */
+            $scope.updateContractList = function(newEndDate){
```
AkA84: Why are you publishing this function on the `$scope`? You are not using it in the view, this is supposed to be an internal function only

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -233,26 +232,21 @@ define([
...
                                 pay: results.pay
                             },
-                            isCurrentRevision = (dateEffectiveRevisionCreated <= dateToday &&
-                                dateEffectiveRevisionCreated >= dateEffectiveRevisionCurrent) ||
-                                (dateEffectiveRevisionCurrent > dateToday &&
-                                dateEffectiveRevisionCreated <= dateEffectiveRevisionCurrent);
+                            isCurrentRevision = dateEffectiveRevisionCurrent.diff(dateToday, "day") <= 0 || dateEffectiveRevisionCurrent.diff(dateEffectiveRevisionCreated, "day") <= 0;
 
-                        if (isCurrentRevision) {
-                            updateContractList(results.details.period_end_date);
```
AkA84: I understand that this throws an error, but is it safe to remove it? It does not affect the application logic whatsoever?

AkA84: Also, why is this function throwing an error, but [this](https://github.com/civicrm/civihr/pull/1315/files#diff-6d37cd88016786f95a36f10caf1e4732R235)
```js
updateContractView(results);
$scope.updateContractList(results.details.period_end_date); // <---
```
that you just added, doesn't?

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -274,8 +268,6 @@ define([
                     } else {
                         var revisionListEntitiesView = ['details','hour','pay'], i, objExt;
 
-                        updateContractList(results.details.period_end_date);
```
AkA84: You don't need to call this function in the `else`?

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
     describe('ContractCtrl', function () {
-        var ctrl;
+        var ctrl, $rootScope, $scope, $controller;
 
         beforeEach(module('hrjc'));
-        beforeEach(inject(function ($controller, $rootScope) {
-            var $scope = $rootScope.$new();
-            $scope.contract = { id: '1' };
+        beforeEach(inject(function (_$controller_, _$rootScope_) {
+
```
AkA84: Empty line

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+          $scope.$parent.contract = {
+            id: '1',
+            contact_id: "84",
+            deleted: "0",
+            is_current: "0",
+            is_primary: "1"
+          };
+
+          $scope.$parent.contractCurrent = [];
+          $scope.$parent.contractPast = [];
```
AkA84: Can't you encapsulate all this code here
```js
$scope.contract = {
  id: '1',
  contact_id: "84",
  deleted: "0",
  is_current: "0",
  is_primary: "1"
};

$scope.$parent.contract = {
  id: '1',
  contact_id: "84",
  deleted: "0",
  is_current: "0",
  is_primary: "1"
};

$scope.$parent.contractCurrent = [];
$scope.$parent.contractPast = [];
```
in a function? Something like this:
```js
beforeEach(inject(function (_$controller_, _$rootScope_) {
  $controller = _$controller_;
  $rootScope = _$rootScope_;
  $scope = $rootScope.$new(); 

  setupData(); // <-- temporary name

  ctrl = $controller('ContractCtrl', { $scope: $scope });
}));
```
it would make the code more readable

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+            $scope: $scope
+          });
 
-            ctrl = $controller('ContractCtrl', { $scope: $scope });
         }));
 
-        it('example', function () {
-            expect(true).toBe(true);
+        it('should make current contract', function () {
+          var d = new Date().toISOString().split("T")[0];
```
AkA84: This is very hard to understand, why can't you use `moment`?

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+          $scope.updateContractList(d);
+          expect($scope.$parent.contract.is_current).toBe("1");
+        });
+
+        it('should make past contract', function () {
+          $scope.updateContractList("2016-01-06");
```
AkA84: Before you were doing `var d = new Date().toISOString().split("T")[0];`, but now you're just passing a string?

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+          ctrl = $controller('ContractCtrl', {
+            $scope: $scope
+          });
 
-            ctrl = $controller('ContractCtrl', { $scope: $scope });
         }));
 
-        it('example', function () {
-            expect(true).toBe(true);
+        it('should make current contract', function () {
```
AkA84: The test suite is supposed to be self-documenting, using the `describe` and `it` blocks to give meaningful description of what area of the code you are testing, how it is supposed to work, etc.

These "it should make current contract" and "it should make past contract" are not understandable. What does it mean? "What" should make current contract? What "make" does"? When does that should happen? It's unclear from the test.

You can use [this commit](https://github.com/civicrm/civihr/blob/bc6869cefce53efc96b2acbc792022752aa1bd6a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/my-leave/components/my-leave-report_test.js) as a reference to what i'm talking about

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+          var d = new Date().toISOString().split("T")[0];
+          $scope.updateContractList(d);
```
AkA84: An `it` block should contain, ideally, only one `expect` call and nothing else. Any setup needed for testing the expectation should go in a `beforeEach` block
```js
describe('lorem ipsum', function () { // <-- Dummy name because i don't know yet what you are testing here
  describe('when the date is ??, function () { // <-- I don't know, fill in the blank
    beforeEach(function () {
      $scope.updateContractList(/* some date */);
    });

    it('makes current contract', function () {
      expect($scope.$parent.contract.is_current).toBe("1");
    });
  });

  describe('when the date is ??, function () { // <-- I don't know, fill in the blank
    beforeEach(function () {
      $scope.updateContractList(/* some date */);
    });

    it('makes past contract', function () {
      expect($scope.$parent.contract.is_current).toBe("1");
    });
  });
});
```

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,57 @@
...
+          expect($scope.$parent.contract.is_current).toBe("1");
+        });
+
+        it('should make past contract', function () {
```
AkA84: You shouldn't use "should" in the `it` descriptions, you should use the imperative mood
```js
it('makes past contract', function () {

});
```

hrjobcontract/js/src/job-contract/controllers/contract.js
sleekvick: Yes because I already called it before the if statement.

hrjobcontract/js/src/job-contract/controllers/contract.js
sleekvick: As an internal function testing it would be difficult to test it, the other alternative would be to `this.updateContractList = function(newEndDate){`.

hrjobcontract/js/src/job-contract/controllers/contract.js
sleekvick: It's safe to remove. With or without a revision the contact list will update, this is because execution of the  `updateContractList` was moved tot he top.

hrjobcontract/js/test/controllers/contract_test.js
sleekvick: I need to test with a date in the past because It supposed to change from current to past contract.

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,79 @@
 define([
+    'common/moment',
     'common/angularMocks',
-    'job-contract/app'
-], function () {
+    'job-contract/app',
+    'job-contract/controllers/contract'
```
AkA84: You shouldn't include this, the `job-contract/controllers/contract` should already be included in `job-contract/app`

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,79 @@
...
+        beforeEach(inject(function (_$controller_, _$rootScope_) {
+          $controller = _$controller_;
+          $rootScope = _$rootScope_;
+          makeController();
         }));
 
-        it('example', function () {
-            expect(true).toBe(true);
+        describe("Update contract based on new end date", function () {
+
```
AkA84: Empty line

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,79 @@
...
+          describe("When end date is past", function () {
+            beforeEach(function () {
+              var date = moment().day(-7); // Seven days ago
+              $scope.updateContractList(date);
+            });
+            it("Make contract past", function () {
```
AkA84: Leave an empty line between the blocks

AkA84:
```js
it('marks the contract as not current', function () {
```

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,79 @@
...
+              expect($scope.$parent.contract.is_current).toBe("0");
+            });
+          });
+
+          describe("When end date is today", function () {
+            beforeEach(function () {
+              var date = moment();
+              $scope.updateContractList(date);
+            });
+            it("Make contract current", function () {
```
AkA84:
```js
it('marks the contract as current', function () {
```

hrjobcontract/js/test/controllers/contract_test.js
```diff
@@ -1,22 +1,79 @@
...
+              expect($scope.$parent.contract.is_current).toBe("1");
+            });
+          });
+
+          describe("When date is future", function () {
+            beforeEach(function () {
+              var date = moment().day(7); // Seven days from now
+              $scope.updateContractList(date);
+            });
+            it("Make contract current", function () {
```
AkA84:
```js
it('marks the contract as current', function () {
```

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -68,28 +69,25 @@ define([
...
-                            $scope.$parent.$parent.contractCurrent.splice($scope.$parent.$parent.contractCurrent.indexOf($scope.$parent.contract),1)
-                        }
-                    }
-                }
+            /**
+              * Updates the contract list view,
+              * by sorting the contract into current or past
+              * depending on the period end date of the contract.
+              *
+              * @param {string || date} newEndDate the date specified by the user
```
AkA84: So the param can be both a string and a date?

hrjobcontract/js/src/job-contract/controllers/contract.js
AkA84: Waiting for an answer on this

hrjobcontract/js/src/job-contract/controllers/contract.js
sleekvick: Yes

hrjobcontract/js/src/job-contract/controllers/contract.js
sleekvick: It's safe to remove. What is means is that with or without a revision the contact list will update, this is because execution of the  `updateContractList` has been moved to the top.

PR #1316, opened 2016-12-28
PCHR-1835: Add the LeaveManager Service

davialexandre:
This service encapsulates the logic to deal with who is the Leave Approver of an Admin of Leave and Absences. It has two methods:

### currentUserIsLeaveManagerOf()
It accepts a contactID and checks if the current logged in user is his/her Leave Approver. This is done by checking if there is an active "has Leave Approved by/is Leave Approver of" relationship between the two contacts.

### currentUserIsAdmin()
It just checks if the current logged in user has the "administer leave and absences" permission

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
@@ -1,0 +1,49 @@
+<?php
+
+class CRM_HRLeaveAndAbsences_Service_LeaveManager {
+
+  /**
+   * Checks if the current logged in user is the Leave Manager for the Contact
+   * with the given ID.
+   *
+   * In order to do this, we check if there's a Relationship between both
+   * contacts, where contact A ($contactID) "has Leave Approved By" contact B
+   * (the logged in), and that if the Relationship is active on the current date.
+   *
+   * Note about implementation: Due to the impossibility of making complex
+   * queries using the CiviCRM API (r.start_date IS NULL OR r.start_date <=
+   * CURDATE()), this method uses a SQL query to check the relationship.
+   *
+   * @param int $contactID
+   *
+   * @return bool
+   */
+  public function currentUserIsLeaveManagerOf($contactID) {
+    $currentUserID = CRM_Core_Session::getLoggedInContactID();
+
+    $relationshipTable = CRM_Contact_BAO_Relationship::getTableName();
+    $relationshipTypeTable = CRM_Contact_BAO_RelationshipType::getTableName();
+
+    $query = "SELECT r.id 
+                FROM {$relationshipTable} r
+                INNER JOIN {$relationshipTypeTable} rt ON rt.id = r.relationship_type_id
+              WHERE r.is_active = 1 AND 
+                    rt.is_active = 1 AND 
+                    rt.name_a_b = 'has Leave Approved By' AND
+                    r.contact_id_a = %1 AND
+                    r.contact_id_b = %2 AND
+                    (r.start_date IS NULL OR r.start_date <= %3) AND 
+                    (r.end_date IS NULL OR r.end_date >= %3)
+              ";
+
+    $params = [
+      1 => [$contactID, 'Integer'],
+      2 => [$currentUserID, 'Integer'],
+      3 => [date('Y-m-d'), 'String']
+    ];
+
+    $result = CRM_Core_DAO::executeQuery($query, $params);
+
+    return $result->N > 0;
+  }
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
@@ -1,5 +1,11 @@
 <?php
 
+/**
+ * This service contains some methods to deal with who is the manager of Leave
+ * Requests. Examples, checking if a contact is the "Leave Approver" of another
+ * contact of if a contact has permission to administer things under Leave and
+ * Absences.
+ */
 class CRM_HRLeaveAndAbsences_Service_LeaveManager {
 
   /**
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
```diff
@@ -0,0 +1,66 @@
...
+    $relationshipTable = CRM_Contact_BAO_Relationship::getTableName();
+    $relationshipTypeTable = CRM_Contact_BAO_RelationshipType::getTableName();
+
+    $query = "SELECT r.id 
+                FROM {$relationshipTable} r
+                INNER JOIN {$relationshipTypeTable} rt ON rt.id = r.relationship_type_id
+              WHERE r.is_active = 1 AND 
+                    rt.is_active = 1 AND 
+                    rt.name_a_b = 'has Leave Approved By' AND
+                    rt.name_b_a = 'is Leave Approver of' AND
```
omarabuhussein: this can be removed .. name_a_b is unique , so it would be enough to check it

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
```diff
@@ -0,0 +1,66 @@
...
+                    r.contact_id_a = %1 AND
+                    r.contact_id_b = %2 AND
+                    (r.start_date IS NULL OR r.start_date <= %3) AND 
+                    (r.end_date IS NULL OR r.end_date >= %3)
+              ";
+
+    $params = [
+      1 => [$contactID, 'Integer'],
+      2 => [$currentUserID, 'Integer'],
+      3 => [CRM_Utils_Date::processDate('today'), 'String']
```
omarabuhussein: I believe there is no need for using CRM_Utils_Date::processDate('today') here , date('Y-m-d') should be enough ...

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
```diff
@@ -0,0 +1,124 @@
...
+    $this->assertFalse($this->leaveManagerService->currentUserIsAdmin());
+    CRM_Core_Config::singleton()->userPermissionClass->permissions = ['administer leave and absences'];
+
+    $this->assertTrue($this->leaveManagerService->currentUserIsAdmin());
+  }
+
+  private function setContactAsLeaveApproverOf($leaveApprover, $contact, $startDate = null, $endDate = null, $isActive = true) {
+    $relationshipType = $this->getLeaveApproverRelationshipType();
+
+    civicrm_api3('Relationship', 'create', array(
```
omarabuhussein: why not to create a fabricator for it in civihr core extension .. same for RelationshipType below ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
davialexandre: I think, since "today" is explicit, that `CRM_Utils_Date::processDate('today')` is a bit more expressive than `date('Y-m-d')` and it also has the advantage of returning the date in the format expected by Civi (though the format won't really make a difference for this specific case).

Do you still think it should be replaced by date()?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
davialexandre: Cool. I really hate how RelatioshipType doesn't have a unique identifier

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
davialexandre: The Fabricator for RelationshipType already exists (you created it yourself!), but it has not been merged to la-milestone-3 yet. I would prefer not to use it for now and avoid a merge conflict once we merge staging into la-milestone-3.

As for the Relationship Fabricator, I agree we should create it, but would you think it's ok to do that on another PR? On PCHR-1798, I'm going to do some changes on this code and then I can add the fabricator. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
omarabuhussein: I think we should not rely much on CiviCRM utility classes as long as we can, you don't know when they will get changed or removed , there is no guarantee on that  .... 

> think, since "today" is explicit, that CRM_Utils_Date::processDate('today') is a bit more expressive than

everyone know what is date('Y-m-d') does , even the new comer to php

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
omarabuhussein: yup that's fine

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveManager.php
omarabuhussein: tbh I was thinking that we will maybe benefit from using a third party library for managing dates such as carbon http://carbon.nesbot.com/   .. we could add composor.json to  the hrcore extension and add this library to it ... we could also benfit from this approach to add any 3rd library we want instead of relying on civi unstable classes and methods .

PR #1318, opened 2016-12-29
PCHR-1412: Set FTE Values as 0 when Casual Type is Used on Import

MiyaNoctem:
## Problem
On import, use of empty hours amount or "casual" hours type was causing FTE values to be calculated as 1/1.  Expected result for wither empty hours or Casual hours type is 0/0 FTE.

As it turned out, default values for FTE values in database are 1.  On import, if hours amount was empty (""), no values were being set on insert query, thus using default values of 1 and hence calculating a 1/1 FTE.

## Solution
Added check to see if "Casual" hours type is being used on job contract import.  If either "Casual" type  or empty "Hours Amount" is used on import, FTE values are set to 0.

### DIFF
--- a/hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
+++ b/hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
@@ -2,6 +2,7 @@
 
 use Civi\Test\HeadlessInterface;
 use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
 
 /**
  * Class CRM_Hrjobcontract_Import_Parser_ApiTest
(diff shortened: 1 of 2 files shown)


hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -187,6 +196,53 @@ function testHourAutoPopulatedFields() {
...
+      'HRJobHour-location_standard_hours' => 'Small office - 36.00 hours per Week',
+      'HRJobContract-contact_id' => $contact2['id'],
+      'HRJobHour-hours_type' => 'part time',
+      'HRJobHour-hours_amount' => ''
+    ]);
+    $importEmptyContractResponse = $this->runImport($emptyHoursContract);
+    $this->assertEquals(CRM_Import_Parser::VALID, $importEmptyContractResponse);
+
+    // Assert FTE's
+    $expected = array('fte_num' => 0, 'fte_denom' => 0, 'hours_fte' => 0);
```
coldrunKacper: Can you use [] instead of array()?

PR #1319, opened 2016-12-29
PCHR-1798: Add the managed_by param to the LeaveRequest.get API

davialexandre:
This PR introduces the `managed_by` param for the `LeaveRequest.get` API. It accepts a Contact ID and, when present, it will make the API return only the Leave Requests of contacts managed by Contact with the given ID.

To know which contact is the manager of another contact, we use the "has Leave Approved by/is Leave Approver of" relationship. When the `managed_by` param is present, the LeaveRequestSelect class will add JOINs to the civicrm_relationship and civicrm_relationship_type.

The query also considers if the relationship is active today. That is, if the `start_date` and `end_date` are null or overlap the current date and if `is_active` is true

### Example usage
Say you have a manager with ID 1 and you have two contacts, with IDs 2 and 3. The manager is the Leave Approver only for the Contact with ID 2.

If you call:
```php
civicrm_api3('LeaveRequest', 'get');
```

It will return all the Leave Request for both contacts. 

Now, if you call:
```php
civicrm_api3('LeaveRequest', 'get', ['managed_by' => 1]);
```

Only the LeaveRequests for Contact ID 2 will be returned.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
@@ -22,6 +22,7 @@ class api_v3_LeaveRequestTest extends BaseHeadlessTest {
 
   use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
   use CRM_HRLeaveAndAbsences_LeaveBalanceChangeHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
 
   /**
    * @var CRM_HRLeaveAndAbsences_BAO_AbsenceType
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -1227,4 +1228,169 @@ public function testDeleteAlsoDeletesLeaveRequestAndItsBalanceChangesFor() {
...
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $staffMember3['id']],
+      ['period_start_date' => '2016-01-01']
+    );
+
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $staffMember4['id']],
+      ['period_start_date' => '2016-01-01']
+    );
+
```
MiyaNoctem: Contract fabrication could be easily refactored out into a single method, passing an array of staffmembers / staffmember id's and cycling through them, making the test easier to read.
```php
$this->createContracts([$staffmember1, $staffmember2, $staffmember3, $staffmember4]);
```
```php
private function createContracts($contacts) {
  foreach ($contacts as $staffMember) {
    HRJobContractFabricator::fabricate(
      ['contact_id' => $staffMember['id']],
      ['period_start_date' => '2016-01-01']
    );
  }
}
```

davialexandre: I don't think that that would be a good idea. It's just a coincidence (and a bit a laziness from my part) that all the contracts start at the same date. Having a method which creates contracts starting at 2016-01-01 is not very useful as it is too specific for just a few tests. 

With that said, instead of creating the method you have suggested, I changed the contract dates. Now each one start on a different day and we have some more "randomness" in the tests.

PR #1105, opened 2016-08-23
PCHR-1297: Fixing advanced-search styles

danilo-compucorp:
This PR fixes some layout issues spotted on the Advanced Search page, and it was also used to refactor some code, and improve the responsiveness of the page, in order to better fit on smaller screen sizes.

**_Help icons_ not aligned:**

**Before**:
helpicons_before

**After**:
helpicons_after

**Checkbox is overlapped**

**Before**:
checkbox_before

**After**:
checkbox_after

**Radiobuttons should be aligned with their options** 

**Before**:
radios_before

radios2_before

**After**:
radios_after

radios2_after

This PR also activates the "select2" dropdown plugin on standard select elements which still have the default look and feel.
In order to activate the script and target it only for the "Advanced Search" page, the "civicrm_alterContent" hook was used.

**Before**:
before

**After**:
after

**Final Result**

**Before
Screen width 1280px:**
advanced search_before

**Screen width 1920px:**
advanced search_large_before

**After:
Screen width 1280px:**
advanced search_after

**Screen width 1920px:**
advanced search_large_after

PR #1321, opened 2016-12-31
PCHR-1353: Bootstrap style improvements

colemanw:
- Adds bootstrap styles to select2 widgets
- Loads bootstrap scripts everywhere, not just on the style guide page
- Fixes a jQuery conflict
- Moves the bootstrap stylesheet to just under civicrm.css  for better overrides

Before (select2 looks wrong, bootstrap menu broken):
---------

bootstrapbefore

After:
-------

bootstrapafter


colemanw:
@totten I copied and modified this file from the select2 repo to make it look decent with our version of the bootstrap theme.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -123,22 +123,27 @@ function bootstrapcivicrm_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
...
 }
 
 /**
  * Implementation of hook_civicrm_buildForm
  */
 function bootstrapcivicrm_civicrm_buildForm($formName) {
-  CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/radio-checkbox.js');
-  CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/add-missing-date-addons.js');
   if($formName == 'CRM_Contact_Form_Search_Advanced') {
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/enable-select2.js');
```
colemanw: Aside... I scratched my head when I read the contents of this script. `Common.js` already takes care of initializing select2 elements. And if it's angular code you're targeting, there is a directive for that. So why is this needed?

totten:
+1

AkA84:
@colemanw please add the ticket number to the PR and commits title (*PCHR-1353: xyz*), and also please provide a description

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -123,22 +123,27 @@ function bootstrapcivicrm_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
 }
 
 /**
- * Implementation of hook_civicrm_pageRun
+ * Implementation of hook_civicrm_coreResourceList
  */
-function bootstrapcivicrm_civicrm_pageRun($page) {
-  if (!(isset($_GET['snippet']) && $_GET['snippet'] == 'json')) {
-    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css');
+function bootstrapcivicrm_civicrm_coreResourceList(&$items, $region) {
```
AkA84: Why did you change the hook? What's `_coreResourceList` for?

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
AkA84: You are referring to https://github.com/civicrm/civihr/pull/1105/commits/369d0389088bfad152ccb7b08bb8b47d471f1725 ? To be honest I haven't followed this, so I'm not sure if it's really needed or not. Just to give a bit of context: this is part of our effort to provide a civihr-like look-and-feel to core pages, when I'll have time to review what's been done so far I'll be glad to throw the script away if it's really not needed

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -123,22 +123,27 @@ function bootstrapcivicrm_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
...
+  if ($region == 'html-header') {
+    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css', -50, 'html-header');
+    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/select2-bootstrap.css', -49, 'html-header');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/radio-checkbox.js');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/add-missing-date-addons.js');
+    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrap', 'js/transition.js', 1000, 'html-header');
```
AkA84: Do you think it would be better to bundle all of these files?

org.civicrm.bootstrapcivicrm/css/select2-bootstrap.css
```diff
@@ -0,0 +1,94 @@
+#bootstrap-theme .form-control .select2-choice {
```
AkA84: The _css/_ folder should only contains two post-processed files: *bootstrap.css* and *custom-civicrm.css*

If you want this file to be always part of the bootstrap theme, then i think you can just place it in the `vendor/` folder and add it to the _bootstrap.scss_ file (make sure to remove the `#bootstrap-theme` namespace.

org.civicrm.bootstrapcivicrm/css/select2-bootstrap.css
```diff
@@ -0,0 +1,94 @@
...
+    border: 0;
+    border-radius: 2px;
+    text-decoration: none !important;
+}
+
+#bootstrap-theme .form-control .select2-choice .select2-arrow {
+    border-radius: 0 2px 2px 0;
```
AkA84: Use 2 spaces indentation

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
colemanw: I don't think it makes much difference. If there is a ready-made bundled version, that would be good to use. Otherwise, not a big deal.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
colemanw: I updated the commit message to explain. Basically, `hook_civicrm_coreResourceList` should be used for adding resources (css/js) which is exactly what we're doing here.

colemanw:
@AkA84 ready for re-review.

org.civicrm.bootstrapcivicrm/bootstrapcivicrm.php
```diff
@@ -123,22 +123,26 @@ function bootstrapcivicrm_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
...
  */
-function bootstrapcivicrm_civicrm_pageRun($page) {
-  if (!(isset($_GET['snippet']) && $_GET['snippet'] == 'json')) {
-    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css');
+function bootstrapcivicrm_civicrm_coreResourceList(&$items, $region) {
+  if ($region == 'html-header') {
+    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.bootstrapcivicrm', 'css/bootstrap.css', -50, 'html-header');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/radio-checkbox.js');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrapcivicrm', 'js/add-missing-date-addons.js');
+    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.bootstrap', 'js/transition.js', 1000, 'html-header');
```
AkA84: Just realized this, since you are checking `if ($region == 'html-header') {`, does it mean that you are placing all these js files in the `<head>`? If so, then this should be fixed and the js files should be right before `</body>`

colemanw: It's ok to place scripts in other regions via this hook. The reason for the conditional is just to filter out edge-cases where the hook might be called when CiviCRM is *not* loading its main styles and scripts.

colemanw: @AkA84 placing the scripts in the `<head>` is necessary due to our use of `jQuery.noConflict` - scripts placed in other regions of the page will not be able to find jQuery. You might be using a CMS which masks the problem by making jQuery a global variable, but on a stock install of Drupal/CiviCRM this bombs. I've fixed the bug by moving the scripts to the header.

AkA84:
@colemanw sorry for the delay on this. Please update the description explaining what you are doing/fixing (screenshots appreciated) and I can approve

colemanw:
@AkA84 done.

PR #1324, opened 2017-01-02
PCHR-1844: Set Leave Request to_date to be same as  from_date for single day leave requests

tunbola:
This PR updates the LeaveRequest DAO to_date field to be a required field.
The reason for this is that currently, a call to : 
```php
civicrm_api3('LeaveRequest', 'get', array(
  'sequential' => 1,
  'from_date' => array('>=' => "2016-01-01"),
  'to_date' => array('<=' => "2016-12-31"),
));
```
will not include Single day Leave Requests. This is because currently the to_date field for single day leave request is null. 
So in this PR, for single day leave requests, the to_date field is set to be same value as the from_date field.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -59,31 +59,12 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    */
   public static function validateParams($params) {
     self::validateMandatory($params);
-    self::validateToDateType($params);
     self::validateStartDateNotGreaterThanEndDate($params);
     self::validateAbsencePeriod($params);
     self::validateNoOverlappingLeaveRequests($params);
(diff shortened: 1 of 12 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -25,6 +25,14 @@ public static function create($params, $validate = true) {
...
 
     CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
+
+    //For single day leave requests
+    $hasFromDateAndFromDateType = !empty($params['from_date']) && !empty($params['from_date_type']);
+    $doesNotHaveToDateAndToDateType = empty($params['to_date']) && empty($params['to_date_type']);
+    if ($hasFromDateAndFromDateType && $doesNotHaveToDateAndToDateType){
+      $params['to_date'] = $params['from_date'];
+      $params['to_date_type'] = $params['from_date_type'];
+    }
```
davialexandre: I just realized that it doesn't make sense to make to_date required and don't do the same with to_date_type. Can you please make to_date_type required as well? Then I think you can remove these 6 lines here

PR #1325, opened 2017-01-02
PCHR-1790: Fix for updating Activity Types with Absence Types

bmnepali:
### Issue:
As for HRAbsenece Module, when updating the Absence Type didn't update its associated Activity Type. As a result, this issue arrises, as in absence navigation menu, filter dropdown and tables in entitlements, calendar, statistics in my absence page are displaying title or label from Activity Types. 

### Solution: 
**hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php's create function** has been modified to add Activity Type Update functionality when Absence Type is updated. Which solves the issue.

### DIFF
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
@@ -42,44 +42,68 @@ class CRM_HRAbsence_BAO_HRAbsenceType extends CRM_HRAbsence_DAO_HRAbsenceType {
     }
 
     $activityTypesResult = civicrm_api3('activity_type', 'get', array());
-    if (CRM_Utils_Array::value('allow_debits', $params) && empty($params['debit_activity_type_id'])) {
-      $weight = count($activityTypesResult["values"]);
-      $debitActivityLabel = $params['title'];
-      $debitActivityTypeId = array_search($debitActivityLabel, $activityTypesResult["values"]);
-      if (!$debitActivityTypeId) {
-        $weight = $weight + 1;
-        $paramsCreate = array(
-          'weight' => $weight,
-          'label' => $debitActivityLabel,
-          'filter' => 1,
-          'is_active' => 1,
-          'is_optgroup' => 0,
-          'is_default' => 0,
-          'grouping' => 'Timesheet',
-        );
-        $resultCreateActivityType = civicrm_api3('activity_type', 'create', $paramsCreate);
-        $debitActivityTypeId = $resultCreateActivityType['values'][$resultCreateActivityType["id"]]['value'];
+    if (CRM_Utils_Array::value('allow_debits', $params)) {
+      if (empty($params['debit_activity_type_id'])) {
+        $weight = count($activityTypesResult["values"]);
+        $debitActivityLabel = $params['title'];
+        $debitActivityTypeId = array_search($debitActivityLabel, $activityTypesResult["values"]);
+        if (!$debitActivityTypeId) {
+          $weight = $weight + 1;
+          $paramsCreate = array(
+            'weight' => $weight,
+            'label' => $debitActivityLabel,
+            'filter' => 1,
+            'is_active' => 1,
+            'is_optgroup' => 0,
+            'is_default' => 0,
+            'grouping' => 'Timesheet',
+          );
+          $resultCreateActivityType = civicrm_api3('activity_type', 'create', $paramsCreate);
+          $debitActivityTypeId = $resultCreateActivityType['values'][$resultCreateActivityType["id"]]['value'];
+        }
+        $params["debit_activity_type_id"] = $debitActivityTypeId;
+      } else {
+        $optionValue = civicrm_api3('OptionValue', 'get', array(
+          'sequential' => 1,
+          'option_group_id' => "activity_type",
+          'value' => $params['debit_activity_type_id'],
+        ));
+        civicrm_api3('OptionValue', 'create', array(
+          'id' => $optionValue['id'],
+          'label' => $params['title'],
+        ));
       }
-      $params["debit_activity_type_id"] = $debitActivityTypeId;
     }
-    if (CRM_Utils_Array::value('allow_credits', $params) && empty($params["credit_activity_type_id"])) {
-      $weight = count($activityTypesResult["values"]);
-      $creditActivityLabel = ts('%1 (Credit)', array(1 => $params["title"]));
-      $creditActivityTypeId = array_search($creditActivityLabel, $activityTypesResult["values"]);
-      if (!$creditActivityTypeId) {
-        $weight = $weight + 1;
-        $paramsCreate = array(
-          'weight' => $weight,
-          'label' => $creditActivityLabel,
-          'filter' => 1,
-          'is_active' => 1,
-          'is_optgroup' => 0,
-          'is_default' => 0,
-          'grouping' => 'Timesheet',
-        );
-        $resultCreateActivityType = civicrm_api3('activity_type', 'create', $paramsCreate);
-        $creditActivityTypeId = $resultCreateActivityType['values'][$resultCreateActivityType["id"]]['value'];
-        $params["credit_activity_type_id"] = $creditActivityTypeId;
+    if (CRM_Utils_Array::value('allow_credits', $params)){
+      if (empty($params['credit_activity_type_id'])) {
+        $weight = count($activityTypesResult["values"]);
+        $creditActivityLabel = ts('%1 (Credit)', array(1 => $params["title"]));
+        $creditActivityTypeId = array_search($creditActivityLabel, $activityTypesResult["values"]);
+        if (!$creditActivityTypeId) {
+          $weight = $weight + 1;
+          $paramsCreate = array(
+            'weight' => $weight,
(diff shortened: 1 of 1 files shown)


hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -42,44 +42,68 @@ public static function create($params) {
...
-          'is_default' => 0,
-          'grouping' => 'Timesheet',
-        );
-        $resultCreateActivityType = civicrm_api3('activity_type', 'create', $paramsCreate);
-        $creditActivityTypeId = $resultCreateActivityType['values'][$resultCreateActivityType["id"]]['value'];
-        $params["credit_activity_type_id"] = $creditActivityTypeId;
+    if (CRM_Utils_Array::value('allow_credits', $params)){
+      if (empty($params['credit_activity_type_id'])) {
+        $weight = count($activityTypesResult["values"]);
+        $creditActivityLabel = ts('%1 (Credit)', array(1 => $params["title"]));
```
davialexandre: What about moving this line to outside the if/else so you don't need to duplicate it when updating the Activity Type label?

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -42,44 +42,68 @@ public static function create($params) {
...
+      } else {
+        $optionValue = civicrm_api3('OptionValue', 'get', array(
+          'sequential' => 1,
+          'option_group_id' => "activity_type",
+          'value' => $params['credit_activity_type_id'],
+        ));
+        civicrm_api3('OptionValue', 'create', array(
+          'id' => $optionValue['id'],
+          'label' => ts('%1 (Credit)', array(1 => $params["title"]))
+        ));
```
davialexandre: Can you please extract those 2 API calls to a separate method which would accept the Option Value value and the Label as params? Then you can use it on both ifs and avoid the duplication

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
   }
 
   /**
+   * Creates Activitu Types based on $params and $activityTypes
```
davialexandre: s/Activitu/Activity

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+   * @param  $activityTypes Array
```
davialexandre: For PHP code we follow [Drupal's code standard](https://www.drupal.org/docs/develop/standards/coding-standards). These are the rules specific for docblocks: https://www.drupal.org/node/1354

- There should be a space between the method description and the @param section
- There should be a space between the @param and the @return section
- The correct order is @param TYPE $paramName Description
- Types are always all in lowercase (Except for objects). So it should be `array` instead of `Array` and `boolean` instead of `Boolean`

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+   * @param  $params        Array
+   * @param  $dr_cr         Boolean 'true' for 'debit' and 'false' for 'credit'
+   * @return Array          Created Activity Type as array
+   */
+  public static function createActivityType($activityTypes, $params, $dr_cr)
```
davialexandre: Please always use camelCase for variables in PHP. The only exception is for variables mapped to db fields, in which snake_case should be used instead.

Also, avoid using abbreviations as much as possible. Explicit is always better than implicit. The $dr_cr variable, for example, can be renamed to $isDebit

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+  {
```
davialexandre: To follow the Drupal code standard the opening brace should not be on a newline

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+    $paramsCreate = array(
+      'weight' => ($weight+1),
+      'label' => $valueLabel,
+      'filter' => 1,
+      'is_active' => 1,
+      'is_optgroup' => 0,
+      'is_default' => 0,
+      'grouping' => 'Timesheet'
+    );
+    return civicrm_api3('activity_type', 'create', $paramsCreate);
```
davialexandre: Please add one empty line before the return

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+  }
+
+  /**
+   * Updates Opton Value based on $params
+   * @param  $params  Array
+   * @param  $dr_cr   Boolean  'true' for 'debit' and 'false' for 'credit'
```
davialexandre: Please update this dockblock to follow the code standard as well

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,59 @@ public static function create($params) {
...
+    }
+    $optionValue = civicrm_api3('OptionValue', 'get', array(
+      'sequential' => 1,
+      'option_group_id' => "activity_type",
+      'value' => $valueId
+    ));
+    civicrm_api3('OptionValue', 'create', array(
+      'id' => $optionValue['id'],
+      'label' => $valueLabel
+    ));
```
davialexandre: Please always try to add some spaces to make the code easier to read. You could add on empty line after the if/else block and one between the civicrm_api3 calls

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
...
+    return civicrm_api3('activity_type', 'create', $paramsCreate);
+  }
+
+  /**
+   * Updates Option Value based on $params
+   *
+   * @param  array $params
+   * @param  boolean $isDebit  ('true' for 'debit' and 'false' for 'credit')
+   *
+   * @return void
```
davialexandre: Sorry, I missed this on my last review. If the method doesn't return anything, you should ommit the @return

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
...
+    );
+
+    return civicrm_api3('activity_type', 'create', $paramsCreate);
+  }
+
+  /**
+   * Updates Option Value based on $params
+   *
+   * @param  array $params
+   * @param  boolean $isDebit  ('true' for 'debit' and 'false' for 'credit')
```
davialexandre: I think that the new param name is self descriptive enough, so no need to the description. If you want to keep it, please remove the parenthesis

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
   }
 
   /**
+   * Creates Activity Types based on $params and $activityTypes
+   *
+   * @param  array $activityTypes
+   * @param  array   $params
```
davialexandre: please use only 1 space between array and $params

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
...
+   * @param  boolean $isDebit  ('true' for 'debit' and 'false' for 'credit')
```
davialexandre: The same thing applies here. You either remove the description or the parenthesis

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
...
+   *
+   * @return array Created Activity Type as array
+   */
+  public static function createActivityType($activityTypes, $params, $isDebit) {
```
davialexandre: Oh, another thing I only realized now: this method should be private

hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -117,6 +77,67 @@ public static function create($params) {
...
+
+  /**
+   * Updates Option Value based on $params
+   *
+   * @param  array $params
+   * @param  boolean $isDebit  ('true' for 'debit' and 'false' for 'credit')
+   *
+   * @return void
+   */
+  public static function updateOptionValue($params, $isDebit) {
```
davialexandre: this should be private as well

PR #1326, opened 2017-01-02
PCHR-1845: Add the LeavePeriodEntitlement.getEntitlement API endpoint

davialexandre:
This PR introduces the `LeavePeriodEntitlement.getEntitlement` API endpoint. It's a "complement" to `LeavePeriodEntitlement.getRemainder` and `LeavePeriodEntitlement.getBreakdown`, as it returns a list of LeavePeriodEntitlements IDs and its entitlement (the number of days available to be taken as leave at the start of the period).

The API supports 3 params:
* **entitlement_id** _Optional_. The id of a specific entitlement
* **contact_id** _Optional_. The contact we want to get the remainder of the entitlements
* **period_id** _Optional_. The period we want to get the remainder of entitlements

When calling the API, one should pass either only the entitlement_id or both the contact_id and period_id.

### Examples
Getting the entitlement for a single LeavePeriodEntitlement
```php
civicrm_api3('LeavePeriodEntitlement', 'getEntitlement', ['entitlement_id' => 1]);
```

The response will be 
```javascript
{
    "is_error": 0,
    "version": 3,
    "count": 1,
    "values": [
        {
            "id": "1",
            "entitlement": 31
        }
    ]
}
```

Getting the entitlement for all the LeavePeriodEntitlements of a Contact during an AbsencePeriod:
```php
civicrm_api3('LeavePeriodEntitlement', 'getEntitlement', ['contact_id' => 1, 'period_id' => 1]);
```

The response will be 
```javascript
{
    "is_error": 0,
    "version": 3,
    "count": 2,
    "values": [
        {
            "id": "1",
            "entitlement": 31
        },
        {
            "id": "2",
            "entitlement": 2
        }
    ]
}
```

### About the implementation
Different from the other APIs, this was all implemented inside the API function. The reasons for this choice are:
- The LeavePeriodEntitlment already has a instance method named `getEntitlement()` (which, by the way, is used by this new endpoint).
- All the code inside the function is very "API specific". It has some validation only to the API params and the remaining code is just a loop to build the results array.
- Putting the code in a separated class would lead to some tests duplication. We would need to test the method on this other class and also to duplicated some of these on the API tests, in order to make sure the API would call the right method and return the expected value.

### What else is included in this PR?
This PR also includes fixes for some tests related to `ContactWorkPatterns`. It was caused by a poorly choose default effective date on the `ContactWorkPattern` Fabricator. Please check the commit message of https://github.com/civicrm/civihr/commit/6e145298f40a490a0bf94246e8594118a95566c3 for more details

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
@@ -1,5 +1,7 @@
 <?php
 
+use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
+
 /**
  * LeavePeriodEntitlement.create API specification (optional)
  * This is used for documentation and validation.
@@ -102,7 +104,7 @@ function civicrm_api3_leave_period_entitlement_getbreakdown($params) {
 /**
  * LeavePeriodEntitlement.getbreakdown specification
  *
- * @param array $params
+ * @param array $spec
  *
  * @return void
  */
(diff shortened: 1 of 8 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -136,4 +138,107 @@ function _civicrm_api3_leave_period_entitlement_getbreakdown_spec(&$spec) {
...
+ * that specific LeavePeriodEntitlement is returned, other wise the API returns
+ * information about all the LeavePeriodEntitlements for the given contact during
+ * the given period.
+ *
+ * The return format is:
+ *
+ * [
+ *   'is_error' => 0,
+ *   'version' => 3,
+ *   'count' => 1,
```
tunbola: The value of count should be 2 here.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
```diff
@@ -106,4 +110,251 @@ public function testGetBreakdownWhenOnlyEntitlementIdIsPassed() {
...
+  public function testGetEntitlementWhenAllParametersArePassed() {
+    civicrm_api3('LeavePeriodEntitlement', 'getEntitlement', ['entitlement_id'=> 1, 'period_id' => 1, 'contact_id'=>1]);
+  }
+
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage You must include either the id of a specific entitlement, or both the contact and period id
+   */
+  public function testGetEntitlementWhenEntitlementIdAndPeriodIdArePassed() {
+    civicrm_api3('LeavePeriodEntitlement', 'getEntitlement', ['entitlement_id'=> 1, 'contact_id'=>1]);
```
tunbola: The description of this test is wrong. It should be `testGetEntitlementWhenEntitlementIdAndContactIdArePassed`.
We should also add an additional one for `testGetEntitlementWhenEntitlementIdAndPeriodIdArePassed`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
```diff
@@ -106,4 +110,251 @@ public function testGetBreakdownWhenOnlyEntitlementIdIsPassed() {
...
+        ],
+        [
+          'id' => $periodEntitlement2->id,
+          'entitlement' => 0,
+        ]
+      ]
+    ];
+
+    $this->assertEquals($expected, $result);
+
```
tunbola: Wouldn't it be better if we used `assertArraySubset()` ? That way the tests will not likely fail if Civicrm decides to add more stuff to the API response (hopefully)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
```diff
@@ -106,4 +110,251 @@ public function testGetBreakdownWhenOnlyEntitlementIdIsPassed() {
...
+          'entitlement' => 0,
+        ],
+        [
+          'id' => $periodEntitlement4->id,
+          'entitlement' => 0,
+        ]
+      ]
+    ];
+
+    $this->assertEquals($expected, $result);
```
tunbola: probably assertArraySubset() here also and the rest?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
davialexandre: I know I've said this was a good idea sometime ago, but I've been thinking about it and I changed my mind. The problem with `assertArraySubset` is that it will also not fail if the returned values change. For example, if we change this:

```php
[
  'values' => [
    'id' => 1,
    'contact_id'  => 1
  ]
]
```

To this:

```php
[
  'values' => [
    'id' => 1,
    'contact_id'  => 1,
    'blablabla' => 'blablabla', 
  ]
]
```

Testing only for a subset would not make the test fail, even though the response has a new field now.

tunbola: yeah, good point.

tunbola:
PR Approved

PR #1328, opened 2017-01-03
PCHR-1369: Allow custom fields to be filtered out on LeaveRequest.getFull

davialexandre:
### Problem

The passing an array of fields to be returned to the `LeaveRequest.getFull` API, the `balance_change` and `dates` fields would always be included, even if they were not present on the list of fields to be returned.

This happens because both `balance_change` and `dates` are not actual fields on of the LeaveRequest, but fields added manually to the API results, after they were fetched from the database.

## Solution

The LeaveRequestSelect class was updated to take into consideration if the `return` param passed to the API and, if it's not empty and doesn't includes `balance_change` and/or `dates`, they won't be includes on the results

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -188,16 +188,20 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
       $leaveRequestBao = new LeaveRequest();
       $leaveRequestBao->copyValues($leaveRequest);
 
-      $balanceChange = LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequestBao);
-      $results[$i]['balance_change'] = $balanceChange;
+      if($this->shouldReturnBalanceChange()) {
+        $balanceChange = LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequestBao);
+        $results[$i]['balance_change'] = $balanceChange;
+      }
 
-      $dates = $leaveRequestBao->getDates();
-      $results[$i]['dates'] = [];
-      foreach($dates as $date) {
-        $results[$i]['dates'][] = [
-          'id' => $date->id,
-          'date' => $date->date
-        ];
+      if($this->shouldReturnDates()) {
+        $dates = $leaveRequestBao->getDates();
+        $results[$i]['dates'] = [];
+        foreach($dates as $date) {
+          $results[$i]['dates'][] = [
+            'id' => $date->id,
+            'date' => $date->date
+          ];
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -540,6 +540,162 @@ public function testGetFullIncludesTheBalanceChangeAndDatesForTheReturnedLeaveRe
...
+      'contact_id' => 1,
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-03-02'),
+      'to_date' => CRM_Utils_Date::processDate('2016-03-02'),
+      'from_date_type' => 1,
+      'to_date_type' => 1,
+      'status_id' => $leaveRequestStatuses['Admin Approved']
+    ], true);
+
+    // This will be returned. The balance change will be -4
```
tunbola: The balance change will be -1 and not -4

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -540,6 +540,162 @@ public function testGetFullIncludesTheBalanceChangeAndDatesForTheReturnedLeaveRe
...
```
tunbola: This comment is not really relevant to this particular test.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -540,6 +540,162 @@ public function testGetFullIncludesTheBalanceChangeAndDatesForTheReturnedLeaveRe
...
+
+    HRJobContractFabricator::fabricate(
+      [ 'contact_id' => 1 ],
+      [
+        'period_start_date' => '2016-01-01',
+        'period_end_date' => '2016-10-01'
+      ]
+    );
+
+    // This will be returned. The balance change will be -1
```
tunbola: This comment is also not relevant to the this particular test

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -540,6 +540,162 @@ public function testGetFullIncludesTheBalanceChangeAndDatesForTheReturnedLeaveRe
...
+      'contact_id' => 1,
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-03-02'),
+      'to_date' => CRM_Utils_Date::processDate('2016-03-02'),
+      'from_date_type' => 1,
+      'to_date_type' => 1,
+      'status_id' => $leaveRequestStatuses['Admin Approved']
+    ], true);
+
+    // This will be returned. The balance change will be -4
```
tunbola: comment not relevant to the test

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -217,4 +221,39 @@ private function addGroupBy(CRM_Utils_SQL_Select $query) {
...
+   * a field will be returned only if "return" is not empty and it includes the
+   * field.
+   *
+   * @param string $field
+   *
+   * @return bool
+   */
+  private function shouldReturnField($field) {
+    return empty($this->params['return']) ||
+           (is_array($this->params['return']) && in_array($field, $this->params['return']));
```
tunbola: Why not `empty($this->params['return']) || $returnArrayHasField` ?
where `$returnArrayHasField = (is_array($this->params['return']) && in_array($field, $this->params['return']));`

tunbola:
Also in the PR description s/The LeaveRequestSelect class was update/The LeaveRequestSelect class was updated

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
davialexandre: The problem is that `$this->params['return']` might not defined, so I have to check if it's not empty first, otherwise I will get a E_NOTICE. If I do as you say, the method will have to look like this:

```php
  private function shouldReturnField($field) {
    $returnParamHasField = !empty($this->params['return']) && 
                           is_array($this->params['return']) && 
                           in_array($field, $this->params['return']);

    return empty($this->params['return']) || $returnParamHasField;
  }
```

Which is more complext then the original.

tunbola: I get your point. Lets leave as it is then.

PR #1329, opened 2017-01-03
PCHR-1677: Fix bug in overlapping leave request validation when a leave request is updated

tunbola:
This PR solves the bug encountered when a Leave Request is updated and the dates overlap with the originally created leave request. To solve this, when a leave request is updated and the the overlapping leave request validation runs, If the leave request being updated is the only one returned as an overlapping leave request, an exception is not thrown.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
@@ -129,6 +129,52 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequestTest extends BaseHeadlessTest {
     $this->assertEquals($toDate->format('Y-m-d'), $dates[1]->date);
   }
 
+  public function testUpdatingALeaveRequestShouldNotThrowOverLappingLeaveRequestExceptionWhenItOnlyOverlapsWithItself() {
+
+    $period = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date'   => CRM_Utils_Date::processDate('2016-12-31'),
+    ]);
+
+    $periodEntitlement = LeavePeriodEntitlementFabricator::fabricate([
+      'type_id' => $this->absenceType->id,
+      'contact_id' => 1,
+      'period_id' => $period->id
+    ]);
+
+    $this->createLeaveBalanceChange($periodEntitlement->id, 20);
+
+    $workPattern = WorkPatternFabricator::fabricateWithA40HourWorkWeek();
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -129,6 +129,57 @@ public function testUpdatingALeaveRequestShouldNotDuplicateTheLeaveRequestDates(
     $this->assertEquals($toDate->format('Y-m-d'), $dates[1]->date);
   }
 
+  public function testUpdatingALeaveRequestShouldNotThrowOverLappingLeaveRequestExceptionWhenTheDatesOverlap()
+  {
```
davialexandre: The opening { should not be on a newline

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -129,6 +129,57 @@ public function testUpdatingALeaveRequestShouldNotDuplicateTheLeaveRequestDates(
...
+
+    //updating leave request
+    $fromDate = new DateTime('2016-11-03');
+    $toDate = new DateTime('2016-11-05');
+    $leaveRequest2 = LeaveRequest::create([
+      'id' => $leaveRequest1->id,
+      'type_id' => $this->absenceType->id,
+      'contact_id' => 1,
+      'status_id' => 1,
+      'from_date' => $fromDate->format('YmdHis'),
```
davialexandre: you're not using fromDate and toDate anywhere else, so why not do this instead?

```php
 'from_date' => CRM_Utils_Date::processDate('2016-11-03'),
 'to_date' => CRM_Utils_Date::processDate('2016-11-05'),
```

You could do the same to fromDate1 and toDate1

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -129,6 +129,57 @@ public function testUpdatingALeaveRequestShouldNotDuplicateTheLeaveRequestDates(
     $this->assertEquals($toDate->format('Y-m-d'), $dates[1]->date);
   }
 
+  public function testUpdatingALeaveRequestShouldNotThrowOverLappingLeaveRequestExceptionWhenTheDatesOverlap()
```
davialexandre: I think this method name is a bit confusing. It gives me the impression that it should not throw the exception whenever it finds an overlapping leave request, even if that leave request is not the one being updated, which is not the case

PR #1330, opened 2017-01-04
PCHR-1687: Default CiviHR configurations

omarabuhussein:
In this PR we set the default CiviHR configurations , Here is the  details of these configs and how things are structured :

## HRCore
In HRCore extensions, 4 Upgraders are defined where each upgrader resides in a separate [trait](http://php.net/manual/en/language.oop5.traits.php), The reason for creating 4 upgraders instead of one is to better seperate things since there was many things to do , and the reason for putting each upgrader in a seperate trait is make each upgrader easier to read instead of having all upgraders defined in one class which is [CRM_HRCore_Upgrader](https://github.com/civicrm/civihr/blob/PCHR-1687-default-civihr-configs/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php#L6) and which will probably become very messy and long with time. 

To make sure the traits upgrader methods get executed you have to include them ( via use keyword ) to the main upgrader class [CRM_HRCore_Upgrader](https://github.com/civicrm/civihr/blob/PCHR-1687-default-civihr-configs/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php#L8)  and since traits will automatically get copied to the class (even its private methods) they will be just as they get defined main upgrader class itself. 

now to avoid method names collision in case two upgrader traits ( e.g CRM_HRCore_Upgrader_Steps_1000 and CRM_HRCore_Upgrader_Steps_1002) defined a method with the same name ( as I say php will blindly copy trait code to the class that use them ) then I had to add a prefix for each upgrader trait methods following this pattren : **upxxxx** ( where xxxx is the upgrader number ) . See these examples [[1](https://github.com/civicrm/civihr/blob/PCHR-1687-default-civihr-configs/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php#L27) , [2](https://github.com/civicrm/civihr/blob/PCHR-1687-default-civihr-configs/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1003.php#L29)]
 
Now and a side from technical side let's explain what each upgrader in hrcore extension do : 

1) **CRM_HRCore_Upgrader_Steps_1000** : In this upgrader we move what is done in this PR https://github.com/civicrm/civihr/pull/1174 to from drush-install.sh to hrcore except point 6 

> 6- Set the Cusom CSS URL to point to the new theme which is available at "[civicrm.root]/tools/extensions/civihr/org.civicrm.bootstrapcivicrm/css/custom-civicrm.css"

which is moved to HRUI extensions https://github.com/civicrm/civihr/blob/PCHR-1687-default-civihr-configs/hrui/CRM/HRUI/Upgrader/Steps/4701.php#L22 since it is a UI related configuration.

2) **CRM_HRCore_Upgrader_Steps_1001** : This upgrader removes some of the default (non-reserved) civicrm options and types which includes the following : 

- **Contact Types** :  'Student', 'Parent',  'Staff', 'Team',  'Sponsor'
- Case Types : 'adult_day_care_referral', 'housing_support'
- **Relationship Types** :    'Benefits Specialist', 'Case Coordinator', 'Parent of',  'Health Services Coordinator', 'Homeless Services Coordinator', 'Senior Services Coordinator',  'Sibling of',  'Supervisor',
'Volunteer is',  'Spouse of', 'Partner of'
- **Location Types** :  'Main', 'Other'
- **Activity Types** : 'Medical evaluation', 'Mental health evaluation',  'Secure temporary housing',
'Income and benefits stabilization', 'Long-term housing plan',  'ADC referral',
- **Mobile providers** : All of them
- **Ethnicity options** : All of them except Prefer_Not_to_Say and Not_Applicable .

3) **CRM_HRCore_Upgrader_Steps_1002** : This upgrader creates **'Miss.'**  individual prefix.

4) **CRM_HRCore_Upgrader_Steps_1003** : This upgrader Hide (Tag) and (Group) sections from contact related screens ( Setting Key : **contact_view_options** ) . It also Hide the following fields from the address (postal code suffix, county, longitude , latitude ), (Setting key : **address_options** ) .
(( Note : this upgrader probably fit hrui extension better , not sure really about it )

## HRUI
hrui extension contain the following upgraders : 
1- **CRM_HRUI_Upgrader_Steps_4700** : this is the just moving the old upgrader method upgrade_4700 from CRM_HRUI_Upgrader to this trait to follow this new method of defining upgraders.

2- **CRM_HRUI_Upgrader_Steps_4701** : This upgrader do the following :  

- sets Custom CSS URL  : as we explained above .
- Sorts Individual Prefixes alphabetically
- Moves Skype option value to the top  of IM list
- Replaces **'Website Type'** option group label by **'Social Account Type'** * and adds word replacement for **'Website'**, **'Website Type'** and **'Add another website'** .

- also I made some changes and removed the lines that disables some contact and relationship types in hrui.php file since we already removed them completely in one of hrcore upgraders.

## HR Emergency 

the following upgraders are added : 

1 - **CRM_Hremergency_Upgrader_Steps_1000** : This upgrader will create the following options for 'Relationship with employee' custom field : 

- Friend
- Spouse
- Partner

and this the upgrader is not executed automaticly for this extension , I updated the auto_install.xml too .

## HR absence ( legacy HR absence)

this upgrader is added **upgrade_1402**  to the main upgrader class ( need more work to migrate it to the new trait approach for upgraders)  which is responsible for renaming 'Vacation' absence type to 'Annual Leave' as well as its related activity type . 

**addDefaultPeriod()** method is updated too to create a period not for the current year only but for the next year too .

## HRIdent extension

upgrade_1501 upgrader is added to the main upgrader class to put (National Insurance) option value which belongs to Goverment ID option group to the top of the list (by setting its whieth to 0 ) and setting it as the default option. the related auto_install.xml is updated too since this extension upgraders are not auto-executed for now.

## HR Job Contract Extension

1- The following piece of code was written twice for no reason https://github.com/civicrm/civihr/pull/1330/files#diff-2aa40b2d89b2a4a5c05d280a6823ad78L205 so I removed one of them .

2- new upgrader **upgrade_1026** is added which : 

    - Remove unused pay scales except 'Not Applicable'
    - Remove unused Hour Locations except 'Head Office'
    - Remove Duplicated 'Employee - Permanent' Contract Type
    - Add 'Fixed Term' Contract Type
    - Sort contract types alphabetically
    - Add 'Retirement' contract end reason
    - Add new contract change reason options

## Other

- The [masquerade](https://www.drupal.org/project/masquerade) drupal module is added to be installed with civihr by default.  Added to drush makefile in this PR: https://github.com/civicrm/civicrm-buildkit/pull/309


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -164,10 +162,43 @@ private function up1001_civicrmActivityTypesList() {
...
+    $deleteableLocations = [];
+
+    $tableName = CRM_Core_BAO_LocationType::getTableName();
+    $references = CRM_Core_DAO::getReferencesToTable($tableName);
+
+    foreach ($locationsToDelete as $currentLocation) {
+      $deleteLocation = true;
+
+      foreach ($references as $currentReference) {
+        if (!is_a($currentReference, 'CRM_Core_Reference_Dynamic')) {
```
davialexandre: the `is_a`  function is deprecated since php 5.0. Use `instanceof` instead

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -164,10 +162,43 @@ private function up1001_civicrmActivityTypesList() {
...
+          $refTable = $currentReference->getReferenceTable();
+          $refKey = $currentReference->getReferenceKey();
+
+          $q = "
+            SELECT *
+            FROM $refTable, $tableName
```
davialexandre: to keep consistency with the rest of the code, please enclose the variables with `{}` (and no abbreviations!)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -9,32 +9,36 @@
...
+      'name', 
+      $this->up1001_civicrmEthnicityOptionsList(), 
+      ['option_group_id' => $this->up1001_getEthnicityGroupName()], 
+      'NOT IN'
+    );
 
     CRM_Core_BAO_Navigation::resetNavigation();
 
-    return TRUE;
+    return false;
```
omarabuhussein: why do you return false here ?

bin/drush-install.sh
```diff
@@ -56,3 +81,18 @@ fi
...
 
+# Scheduled Jobs Enabled by Default
+for currentJob in "${ENABLED_JOBS[@]}"
+do
+  JOB_ID=`drush cvapi Job.getvalue return=id name="$currentJob"`
+  echo "Enabling '$currentJob' Job (ID: $JOB_ID)"
+  drush cvapi Job.create id=$JOB_ID is_active=1
+done
+
+# Scheduled Jobs Disabled by Deffault
```
davialexandre: s/Deffault/Default

bin/drush-install.sh
```diff
@@ -35,6 +35,31 @@ org.civicrm.contactsummary,\
 org.civicrm.bootstrapcivicrm,\
 org.civicrm.bootstrapcivihr
 
+## List of Default Enabled Cron Jobs
+ENABLED_JOBS=( "CiviCRM Update Check" \
```
omarabuhussein: we have hrcore extension now and we could do such things in it , I see no reason to do it in drush-install.sh ... let's keep this sh file for things that cannot be done in using an extension ..

also if you moved this to hrcore , can you do a single array that contain the list of cron jobs along with its desired status in kay=>value pair like this : 

```
$cronJobs = [
'Clean-up Temporary Data and Files' => 1,
'Clone Documents' => 0,
etc ..
]
```

in then you can just loop through the result and set is_active value .

bin/drush-install.sh
```diff
@@ -35,6 +35,31 @@ org.civicrm.contactsummary,\
...
+"Clean-up Temporary Data and Files" \
+"Clone Documents" \
+"Disable expired relationships" \
+"Length of service updater" \
+"Mail Reports" \
+"Process Inbound Emails" \
+"Send Scheduled Mailings" \
+"Send Scheduled Reminders" \
+"Tasks and Assignments Daily Reminder" \
+"Documents Notification" \
```
davialexandre: For Jobs created by CiviCRM, I think this approach is good, but for Jobs created by extensions (e.g., Tasks and Assignments Daily Reminder and Lenght of Service updater), I'm not sure it this is the best way to do it.

For example, the Tasks and Assignments Daily Reminder probably uses code from the T&A extension (classes, BAOs etc). If the extension gets disabled the job will probably fail (because it will try to access an API that doesn't exist anymore). I feel it's better to keep the logic to enable and disable the scheduled job inside the extension and use hooks to disable/enable it when the extensions gets disabled/enabled (and remove it when the extension is removed). @omarabuhussein what do you think?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
```diff
@@ -113,19 +113,15 @@ private function up1000_setAvailableCountries() {
    * Sets Available Provinces to 'all provinces'
    */
   private function up1000_setAvailableProvinces() {
-    // ToDo : Fetch Provinces via API after upgrading to newer civicrm version since it is not available on civicrm 4.7.9
-    $tableName = CRM_Core_DAO_StateProvince::getTableName();
-    $query = CRM_Core_DAO::executeQuery("SELECT id FROM {$tableName}");
+    $countries = civicrm_api3('Country', 'get', ['sequential' => 1]);
```
davialexandre: Are you sure we should use the Country API instead of StateProvince? The previous version of the code was fetching the data using the StateProvince BAO

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -15,21 +15,21 @@ public function upgrade_1001() {
     $this->up1001_listDelete('LocationType', 'name', $this->up1001_civicrmLocationTypesList());
 
     $this->up1001_listDelete(
-      'civicrm_option_value', 
+      'OptionValue',
```
omarabuhussein: Okay I just saw that you fixed these here ... so forget my previous comment

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -38,7 +38,7 @@ public function upgrade_1001() {
 
     CRM_Core_BAO_Navigation::resetNavigation();
 
-    return false;
+    return true;
```
omarabuhussein: you fixed this too , so forget my previous comment

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
omarabuhussein: This is to set available Provinces not countries , so why you are fetching countries ?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -4,9 +4,9 @@
  * Collection of upgrade steps
  */
 class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
-
+  private $_enities;
```
davialexandre: s/enities/entities/

Also, never prefix private properties with `_`. This was something people did when PHP didn't have the access modifiers.

bin/drush-install.sh
omarabuhussein: I already left a comment about all of this , anything that could be done in an extension should be done in an extension and since we have hrcore extension now we can put these things there .. but you presented another important note here which is civihr extensions cron jobs .. I think you are right .. so only core cron jobs should be disabled/enabled in hrcore and the rest should be done in their related extensions .

omarabuhussein:
you also have a confilct @MiyaNoctem  that need to be fixed

omarabuhussein:
regarding this commit https://github.com/civicrm/civihr/pull/1330/commits/28606bd1321e42c331f0c20fe2833ab2005f6f06 

I didn't left any note on it since these methods will probably get removed , see my other comments

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
MiyaNoctem: Using province ID's was breaking the MySQL client I believe, because although the field where this is stored is a TEXT field, a fatal exception was being thrown due to the size of the string being stored.  I investigated how the ProvinceLimit setting is supposed to be used, but documentation for the field is somewhat obscure.  Still, the few on-line usages I found of the setting always used country codes, instead of province codes.  So basically, the setting is used to filter out provinces by providing the ID's for the countries for which you want provinces to be used in the system.

This is the available documentation for the setting in the API:
```
{
            "group_name": "Localization Preferences",
            "group": "localization",
            "name": "provinceLimit",
            "type": "Array",
            "quick_form_type": "Element",
            "html_type": "advmultiselect",
            "html_attributes": {
                "size": 5,
                "style": "width:150px",
                "class": "advmultiselect"
            },
            "default": [

            ],
            "add": "4.3",
            "title": "Available States and Provinces (by Country)",
            "is_domain": 1,
            "is_contact": 0,
            "description": "",
            "help_text": null,
            "pseudoconstant": {
                "callback": "CRM_Admin_Form_Setting_Localization::getAvailableCountries"
            }
        },
```

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1063,7 +1045,269 @@ public function createInsurancePlanTypes() {
...
+   * @return TRUE
+   */
+  public function upgrade_1026() {
+    $this->up1026_removeUnneededPayScales();
+    $this->up1026_removeUnneededHourLocations();
+    $this->up1026_removeDuplicateContractType();
+    $this->up1026_removeRevisionChangeReasons();
+
+    $optionValues = [
+      'hrjc_contract_type' => ['Fixed Term'],
```
MiyaNoctem: I'm not sure I follow... these values are being added, because they didn't exist.  I checked, and they are not in the xmls.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1047,50 +1047,158 @@ public function createInsurancePlanTypes() {
...
+    $this->deleteEntityRecords('HRHoursLocation', $deleteableReasons);
+  }
+
   /**
-   * Removes all Pay scales except 'Not Applicable'
+   * Removes duplicates for 'Employee - Permanent' contract type.
+   */
+  private function up1026_removeDuplicateContractType() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
```
MiyaNoctem: I need the sequential flag so that an ordered array is returned.  The idea is to only delete duplicates, so the first record must be skipped.  So if you look at the for, I'm cycling through the results starting at position 1 in the array.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1047,50 +1047,158 @@ public function createInsurancePlanTypes() {
...
+      'value' => 'Employee - Permanent',
+    ]);
+
+    if ($result['count'] > 1) {
+      for ($i = 1; $i < $result['count']; $i++) {
+        civicrm_api3('OptionValue', 'delete', [
+          'id' => $result['values'][$i]['id'],
+        ]);
+      }
+    }
```
MiyaNoctem: I can't chain it, because I only need to delete duplicates.  The first one I skip.

hrabsence/CRM/HRAbsence/Upgrader.php
```diff
@@ -73,7 +73,7 @@ public function installAbsenceTypes() {
       $leaves = FALSE;
       $options = array(
         'Sick' => 'Sick',
-        'Vacation' => 'Vacation',
+        'Annual Leave' => 'Annual Leave',
```
MiyaNoctem: Quite the catch!

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1047,50 +1047,158 @@ public function createInsurancePlanTypes() {
...
+
+    $this->deleteEntityRecords('HRHoursLocation', $deleteableReasons);
+  }
+
   /**
-   * Removes all Pay scales except 'Not Applicable'
+   * Removes duplicates for 'Employee - Permanent' contract type.
+   */
+  private function up1026_removeDuplicateContractType() {
```
MiyaNoctem: No need to do the check, since what is stored in DB on contract details table is not the ID of the contract type, but the value (ie. literally "Employee - Permanent").

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
MiyaNoctem: Using province ID's was breaking the MySQL client I believe, because although the field where this is stored is a TEXT field, a fatal exception was being thrown due to the size of the string being stored.  I investigated how the ProvinceLimit setting is supposed to be used, but documentation for the field is somewhat obscure.  Still, the few on-line usages I found of the setting always used country codes, instead of province codes.  So basically, the setting is used to filter out provinces by providing the ID's for the countries for which you want provinces to be used in the system.

This is the available documentation for the setting in the API:
```
{
            "group_name": "Localization Preferences",
            "group": "localization",
            "name": "provinceLimit",
            "type": "Array",
            "quick_form_type": "Element",
            "html_type": "advmultiselect",
            "html_attributes": {
                "size": 5,
                "style": "width:150px",
                "class": "advmultiselect"
            },
            "default": [

            ],
            "add": "4.3",
            "title": "Available States and Provinces (by Country)",
            "is_domain": 1,
            "is_contact": 0,
            "description": "",
            "help_text": null,
            "pseudoconstant": {
                "callback": "CRM_Admin_Form_Setting_Localization::getAvailableCountries"
            }
        },
```

bin/drush-install.sh
```diff
@@ -36,32 +36,6 @@ org.civicrm.shoreditch,\
 org.civicrm.bootstrapcivihr,\
```
davialexandre: There's a problem with this git commit message. The title says "Move Job enabling to Extensions", but then you look at the description and it's clear that you're doing more than that with this comment. You also made change to the upgraders, which have nothing to do with the scheduled jobs. In the future, for cases like this, please put the changes in separate commits.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
davialexandre: oh, I meant add instead of remove. As you mentioned yourself, these new options are not on the files and need to be added. 

Sorry for the confusion

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -0,0 +1,234 @@
...
+
+    foreach ($locationsToDelete as $currentLocation) {
+      $deleteLocation = true;
+
+      foreach ($references as $currentReference) {
+        if (!($currentReference instanceof CRM_Core_Reference_Dynamic)) {
+          $referredTableName = $currentReference->getReferenceTable();
+          $referenceKey = $currentReference->getReferenceKey();
+
+          $q = "
```
davialexandre: you forgot to update the variable name here

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1003.php
```diff
@@ -0,0 +1,73 @@
...
+    // update the required settings
+    if (!empty($newOptions)) {
+      civicrm_api3('Setting', 'create', [
+        'sequential' => 1,
+        $settingKey => $newOptions,
+      ]);
+    }
+  }
+
+}
```
davialexandre: missing newline

contactsummary/js/test/services/api_test.js
```diff
@@ -27,7 +27,14 @@ define([
             var expectedResult = {data: true}, actualResult = {};
 
             // Set up response from HTTP backend
-            httpBackend.expectPOST('/civicrm/ajax/rest').respond(expectedResult);
+            httpBackend.expectGET('/civicrm/ajax/rest?action=get&entity=test&json=1&rowCount=0&sequential=1').respond(expectedResult);
```
MiyaNoctem: The test was failing because the service to get information from the backend through the API was implemented using GET instead of POST.

https://github.com/civicrm/civihr/blob/ed0da6aecb7f1a140f8a40398aac8e1a48d356e6/contactsummary/js/src/contact-summary/services/api.js#L76

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php
```diff
@@ -12,7 +12,6 @@ public function upgrade_1001() {
     $this->up1001_listDelete('ContactType', 'name', $this->up1001_civicrmContactTypesList());
     $this->up1001_listDelete('CaseType', 'name', $this->up1001_civicrmCaseTypesList());
     $this->up1001_listDelete('RelationshipType', 'name_b_a', $this->up1001_civicrmRelationshipTypesList());
-    $this->up1001_listDelete('LocationType', 'name', $this->up1001_civicrmLocationTypesList());
```
omarabuhussein: can you remove the method itself 

https://github.com/civicrm/civihr/blob/474aa4a1603b616096716135ec3832306a3030ef/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1001.php#L162

since it is not used anymore

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -6,12 +6,33 @@
 class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
 
   public function install() {
-
     // $this->executeCustomDataFile('xml/customdata.xml');
     $this->executeSqlFile('sql/install.sql');
     $this->upgradeBundle();
   }
 
+  public function enable() {
```
MiyaNoctem: Removed the methods on Upgrader class and tested enbling and disabling extension.  Worked ok!

hrui/CRM/HRUI/Upgrader/Steps/4701.php
```diff
@@ -20,7 +20,7 @@ public function upgrade_4701() {
   }
 
   private function up4701_setCustomCSSURL() {
-    $bootstrapcivicrmDirectory = CRM_Core_Resources::singleton()->getPath('org.civicrm.bootstrapcivicrm', 'css/custom-civicrm.css');
+    $bootstrapcivicrmDirectory = CRM_Core_Resources::singleton()->getPath('org.civicrm.shoreditch', 'css/custom-civicrm.css');
```
davialexandre: well, this is not bootstrap anymore, so the variable name must be changed as well. I would suggest a more generic name, to avoid having to change it again in the future. Something like `$customCSSPath`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -14,26 +14,26 @@ class CRM_HRCore_Upgrader extends CRM_HRCore_Upgrader_Base {
    * @var array
    *   List of jobs in CiviCRM and their intended default status in the form [job_name => is_active]
```
davialexandre: Since the array format has changed, you have to update the docblock as well. Since now the variable names are more self descriptive, I believe only the `@var array` part is enough

omarabuhussein:
approved @MiyaNoctem

PR #1331, opened 2017-01-04
PCHR-1123: Group balance changes by type on LeavePeriodEntitlement.getBreakdown

davialexandre:
In cases when a LeavePeriodEntitlement has more than one LeaveBalanceChange with the same type_id, the `getBreakdown()`  method should return them as a single entry, where the amount will be the sum of the amounts of the grouped items.

So, say you have 2 balance changes of type "Leave", one with 5 days and another with 2 days. The returned date will be this single entry:

```json
{
      "amount": 7,
      "expiry_date": null,
      "type": {
      "id": 1,
        "value": "leave",
        "label": "Leave"
      }
}
```

There's a special handling for Balance Changes of type "Overridden". They will be grouped together with balance changes of type "Leave". The reason is that "Overridden" days are the same thing as Leave. The "Overridden" type is just used internally for the system to be able to know that the original calculated number of days for "Leave" was manually overidden during the entitlement calculation.

Say the calculated entitlement was 10 days, but then it was overridden to 15. This will create 2 balance changes:
- type: "Leave", amount: 10 (the calculated amount)
- type: "Overridden", amount: 5 (the difference between the calculated amount and the overridden value)

LeavePeriodEntitlement.getBreakdown will group both together and the returned data will be:
```json
{
      "amount": 15,
      "expiry_date": null,
      "type": {
      "id": 1,
        "value": "leave",
        "label": "Leave"
      }
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
@@ -192,9 +192,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement extends CRM_HRLeaveAndAb
 
       self::saveBroughtForwardBalanceChange($calculation, $periodEntitlement);
       self::savePublicHolidaysBalanceChanges($calculation, $periodEntitlement);
-
       self::saveLeaveBalanceChange($calculation, $periodEntitlement, $overriddenEntitlement);
-
 
       $transaction->commit();
     } catch(\Exception $ex) {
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -632,23 +683,33 @@ public static function getBreakdown($params) {
...
+        $amount = array_reduce($balanceChangeGroup, function($totalAmount, $balanceChange) {
+          return $totalAmount + (float)$balanceChange->amount;
+        }, 0);
+
         $periodEntitlementInfo['breakdown'][] = [
-          'amount' => $balanceChange->amount,
-          'expiry_date' => $balanceChange->expiry_date,
-          'type' => $leaveBalanceTypeIdOptionsGroup[$balanceChange->type_id],
+          'amount' => $amount,
+          'expiry_date' => $balanceChangeGroup[0]->expiry_date,
```
tunbola: Just clarifying something. Will all balance changes of the same type of the period entitlement have the same expiry date? Is it possible for their expiry date to be different?

PR #1333, opened 2017-01-04
PCHR-1811: Add the LeaveRequest.isManagedBy() API endpoint

davialexandre:
This is a new API endpoint that can be used to check if a given Contact is the manager of the contact of a given Leave Request.

It has two mandatory params:
- *contact_id*: The contact to check if the Leave Request is managed by 
- *leave_request_id*: The Leave Request to check if the contact is the manager of

### Usage:
```php
civicrm_api3('LeaveRequest', 'isManagedBy', ['contact_id' => 1, 'leave_request_id' => 1]);
```

If Contact with ID 1 is the manager of the contact for the the LeaveRequest with ID 1, this will be the response:
```json
{
    "is_error": 0,
    "version": 3,
    "count": 1,
    "values": true
}
```

Otherwise, the response will be:
```json
{
    "is_error": 0,
    "version": 3,
    "count": 1,
    "values": false
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
@@ -24,6 +24,73 @@ class CRM_HRLeaveAndAbsences_Service_LeaveManagerTest extends BaseHeadlessTest {
     $this->unregisterCurrentLoggedInContactFromSession();
   }
 
+  public function testIsContactManagedBy() {
+    $manager = ContactFabricator::fabricate();
+    $staffMember = ContactFabricator::fabricate();
+
+    $this->assertFalse($this->leaveManagerService->isContactManagedBy($staffMember['id'], $manager['id']));
+
+    $this->setContactAsLeaveApproverOf($manager, $staffMember);
+
+    $this->assertTrue($this->leaveManagerService->isContactManagedBy($staffMember['id'], $manager['id']));
+  }
+
+  public function testIsContactManagedByWhenTheRelationshipHasSpecificStartDate() {
+    $manager = ContactFabricator::fabricate();
+    $staffMember = ContactFabricator::fabricate();
+
+    $this->assertFalse($this->leaveManagerService->isContactManagedBy($staffMember['id'], $manager['id']));
+
+    $today = new DateTime('today');
+    $this->setContactAsLeaveApproverOf($manager, $staffMember, $today->format('Y-m-d'));
+
+    $this->assertTrue($this->leaveManagerService->isContactManagedBy($staffMember['id'], $manager['id']));
+  }
+
+  public function testIsContactManagedByWhenTheRelationshipHasSpecificDates() {
+    $manager = ContactFabricator::fabricate();
+    $staffMember = ContactFabricator::fabricate();
+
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
```diff
@@ -24,6 +24,48 @@ public function tearDown() {
     $this->unregisterCurrentLoggedInContactFromSession();
   }
 
+  public function testIsContactManagedBy() {
```
tunbola: What about another test/assertion for this but with valid dates set rather than the default null dates set in the `setContactAsLeaveApproverOf` method for the relationship?
i.e using the `setContactAsLeaveApproverOf` method but with valid dates.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -1594,6 +1594,69 @@ public function testGetShouldOnlyReturnTheLeaveRequestsOfStaffMembersManagedByMa
...
+
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage Mandatory key(s) missing from params array: leave_request_id, contact_id
+   */
+  public function testIsManagedByShouldThrowAnExceptionIfBothTheContactIDAndLeaveRequestIDAreMissing() {
+    civicrm_api3('LeaveRequest', 'isManagedBy');
+  }
+
+  public function testIsManagedByShouldReturnTrueIfTheContactOfTheGivenLeaveRequestIsManagedByTheGivenContactID() {
```
tunbola: This test seems to do two things which contradicts the description. It tests for both when the leave request is managed by the contact and when the leave request is not managed by the contact.
Perhaps it should be broken into two or the description modified to reflect what is being tested.

tunbola:
Also the second API response for when the contact is not the manager of the leave request needs to be updated.  values parameter should be  'false'

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: This first assertion is only to make sure the relationship doesn't exist prior the call to `setContactAsLeaveApproverOf()`. You can see I use the same approach for the other tests on LeaveManager. It's just a way to be sure the method would not return true even if the relationship between the two contacts doesn't exist.

Ideally this should be 2 tests, but we're starting to have performance problems with them (currently, it takes 1.35 minutes to run all of them on my computer), so I prefer to avoid splitting simple tests like this one. Of course, if this was something more complex with lots of changes between the first and last assertion, then I would definitely split it.

I don't see the contraction on the test name. It says it should return true if the contact of the leave request is managed by the given contact id, and that is what happens, the assertions make sure true is returned only on the describe scenario. What is the contradiction?

tunbola: I'm seeing it from the perspective that the API returns false also in the same test function. 
It might be a way to be sure the method would not return true even if the relationship between the two contacts doesn't exist but since logically the API can return either true or false, its like both scenarios is being tested in this same function and the test description gives the impression that only the the scenario when the API is to return true is tested.

PR #1334, opened 2017-01-05
PCHR-1693: Styling for all types of address

sleekvick:
# Cancel button is invisible
future first

Civihr custom theme provides styling for two addresses sections, however, there are six types of addresses so whenever the user adds a third address the appropriate styling does not reflect. 

Furthermore, the class is being generated by CiviCRM core scripts, other solutions would have required a modification of the core scripts.

# Styling for all types of address

Civihr provides the user option to add six different types, therefore providing styling for all six options solves the problem.

pchr-1639


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -232,7 +232,11 @@
 
       .crm-button-type-cancel,
       .crm-button_qf_Address_1_cancel,
-      .crm-button_qf_Address_2_cancel {
+      .crm-button_qf_Address_2_cancel,
```
AkA84: Could this be simply
```css
[class*="crm-button_qf_Address_"] {

}
```
?

This would work with `.crm-button_qf_Address_1_cancel`, `.crm-button_qf_Address_3_cancel`, etc..but also with `.crm-button_qf_Address_2_something_else` so be careful that it doesn't have unintended side effects

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/_detail.scss
```diff
@@ -281,7 +285,11 @@
 }
 
 #Address_1,
-#Address_2 {
+#Address_2,
```
AkA84:
```css
[class*="#Address_"] {

}
```

PR #1327, opened 2017-01-03
PCHR-1812: Augment LeaveRequest API

deb1990:
This PR adds the following functionality

**`isManagedBy()` to `LeaveRequestAPI`**
Given a `leave_request_id` and a `contact_id`, calls the same backend API.

**`roleOf()` to `LeaveRequestInstance`**
Check the role of a given contact in relationship to the leave request. It can be: `owner`, `manager`, `none`.

**Note**: 
1. Error handling of the API has been removed from all Leaves and Absence, as error handling needs to be correctly implemented later. 

2. `LeaveRequestInstance.update()` has been modified and resetting the Instance object to the data returned by API has been removed.

PR #1335, opened 2017-01-05
PCHR-1302 Angular JS leave request modal directive

dhavalsharma:
The directive implements the leave request model dialog which can be plugged in anywhere. It is managed by a controller.
leaverequest

## Parts of implementation

1. directive
leaveRequestPopup can be used in anywhere with the tag `<leave-request-popup contact-id="' + contactId + '"></leave-request-popup>'`
It will take the contact id from the enclosing scope. 
It is a wrapper over `$uibModal`.

2. controller
LeaveRequestPopupCtrl implements the business logic for the leave request creation process. It is located `uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js`

3. html conversion to angular app
It is the UI implementation for the angular app and is located at `civihr/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html`

## Implementation strategy
As leave request is a common modal required by staff/manager/admin it is kept in the shared directory so that it can be reused by individual implementation. Therefore a separate config has been created for the same. 

Used DateFormat to set the date formatting for `HR_settings`.

## Alerts
uib-alert has been used to alert users of messages

## Spinner 
The <crm-loading> directive is used to show a loading spinner when data is loading.

## Bugs fixed 

1. Using `var today = moment().format('YYYY-MM-DD');` for various server side calls in files like `uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/absence-period-model.js`, `uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/public-holiday-model.js` instead of using HR_settings

2. In `WorkPatternAPI` improved the call to take empty `{}` object when params where not defined, otherwise it was sending undefined params to server

3. Improvements in various mock apis and added few more mock apis like `work-pattern-api-mock`


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -52,10 +52,14 @@ define([
           public_holiday: isPublicHoliday || false
         };
 
-        deferred.resolve(this.sendGET('LeaveRequest', 'getbalancechangebyabsencetype', params, false)
+        this.sendGET('LeaveRequest', 'getbalancechangebyabsencetype', params)
           .then(function (data) {
-            return data.values;
-          }));
+            if(data && data.is_error == 1){
```
deb1990: 1. `if(data && !!data.is_error) {`
2. Leave a space before "{"
3. We should not send the error object via `reject`, instead we can just send  
`deferred.reject(data.error_message)`;

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -67,18 +71,22 @@ define([
...
         var deferred = $q.defer();
 
         if (!params.id) {
           deferred.reject('id is mandatory field');
         }
 
-        deferred.resolve(this.sendPOST('LeaveRequest', 'create', params)
+        this.sendPOST('LeaveRequest', 'create', params)
           .then(function (data) {
+            if(data && data.is_error == 1){
```
deb1990: 1. `if(data && !!data.is_error) {`
2. Leave a space before "{"
3. We should not send the error object via `reject`, instead we can just send  
`deferred.reject(data.error_message)`;

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -88,23 +96,27 @@ define([
...
+        if (params && (!params.contact_id || !params.from_date || !params.from_type)) {
+          deferred.reject('contact_id, from_date and from_type in params are mandatory');
         }
 
-        deferred.resolve(this.sendGET('LeaveRequest', 'calculatebalancechange', params)
+        this.sendPOST('LeaveRequest', 'calculatebalancechange', params)
           .then(function (data) {
-            return data.values;
-          }));
+            if(data && data.is_error == 1){
```
deb1990: SAme as before

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -133,11 +145,14 @@ define([
           }
         }
 
-        deferred.resolve(this.sendPOST('LeaveRequest', 'create', params)
+        this.sendPOST('LeaveRequest', 'create', params)
           .then(function (data) {
+            if(data && data.is_error == 1){
```
deb1990: SAme as before

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/public-holiday-model.js
```diff
@@ -36,15 +38,17 @@ define([
          * @return {Bool} returns true if date is a public holday else false
          */
         isPublicHoliday: function (whichDate) {
-          var dateFormat = HR_settings.DATE_FORMAT.toUpperCase();
-          var checkDate = moment(whichDate).format(dateFormat);
+          $log.debug('PublicHoliday isPublicHoliday whichDate', whichDate);
+
+          var checkDate = moment(whichDate).format('YYYY-MM-DD');
```
deb1990: Remove empty line after this line.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/apis/absence-period-api-mock.js
```diff
@@ -13,6 +13,7 @@ define([
         return $q(function (resolve, reject) {
           if (params && 'start_date' in params) {
             var dateFormat = HR_settingsMock.DATE_FORMAT.toUpperCase();
+            dateFormat = 'YYYY-MM-DD';
```
deb1990: Above two lines can be combined as 
`var dateFormat = 'YYYY-MM-DD'`;

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/apis/leave-request-api-mock.js
```diff
@@ -31,12 +31,19 @@ define([
         },
         calculateBalanceChange: function (params) {
           return $q(function (resolve, reject) {
+            if(params && params.from_type == 'half_day_am'){
```
deb1990: `params` should not be checked twice,
string equality should be checked using `===`

```
if(params) {
  if(params.from_type === 'half_day_am') {
    resolve(mockData.singleDayCalculateBalanceChange().values);
  }
  else if(params.from_type === 'all_day') {
    resolve(mockData.multipleDayCalculateBalanceChange().values);
  }
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -364,6 +364,20 @@ define([
...
+     * Pick a specific object out of a collection
+     *
+     * @param {array} the option group collection key
+     * @param {string} key - The sub-collection key
+     * @param {string} value - The sub-collection key's value to match
+     * @return {object}
+     */
+    specificObject: function (collection, key, value) {
+      var whichOptionGroup = mockedOptionValues[collection];
+      return _.find(whichOptionGroup, function (item) {
```
deb1990: Leave a empty line before return, if it is not the first line of the function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/option-group-mock-data.js
```diff
@@ -364,6 +364,20 @@ define([
      */
     getCollection: function (collection) {
       return mockedOptionValues[collection];
-    }
+    },
+    /**
+     * Pick a specific object out of a collection
+     *
+     * @param {array} the option group collection key
```
deb1990: Doc missing the param names

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -149,7 +148,18 @@ define([
 
       beforeEach(function () {
         requestData = helper.createRandomLeaveRequest();
-        spyOn(LeaveRequestAPI, 'sendGET').and.callThrough();
+        //todo --> will be removed once from_type will change to from_date_type
+        requestData = _.mapKeys(requestData, function (value, key) {
+          if (key == 'from_date_type') {
```
deb1990: Use === for string equality, its faster

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/directives/leave-request-popup_test.js
```diff
@@ -0,0 +1,60 @@
...
+          spyOn($uibModal, 'open');
+          directive.triggerHandler('click');
+          $controllerScope.$digest();
+        });
+
+        it('opens dependent popup', function () {
+          expect($uibModal.open).toHaveBeenCalledWith(jasmine.any(Object));
+        });
+
+        it('has expected markup', function () {
```
deb1990: Empty test, should be removed

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -0,0 +1,251 @@
+<div class="modal-header">
```
deb1990: The original file,
https://github.com/civicrm/civihr/blob/PCHR-1302-AngularJS-LeaveRequestModal-directive/uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-request.html

Should be removed

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
```diff
@@ -29,10 +35,16 @@ define([
     'common.models',
     'my-leave.config',
     'my-leave.components',
+    'leave-absences.directives',
     'leave-absences.models',
+    'common.angularDate',
```
AkA84: `common.angularDate` is already defined as a dependency

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
```diff
@@ -13,9 +13,15 @@ define([
...
   'leave-absences/shared/models/absence-type-model',
+  'leave-absences/shared/directives/leave-request-popup',
   'leave-absences/shared/models/entitlement-model',
   'leave-absences/shared/models/leave-request-model',
   'leave-absences/shared/models/calendar-model',
+  'leave-absences/shared/models/absence-period-model',
+  'leave-absences/shared/models/absence-type-model',
+  'leave-absences/shared/models/entitlement-model',
+  'leave-absences/shared/models/public-holiday-model',
+  'common/services/angular-date/date-format',
```
AkA84: `common/services/angular-date/date-format` is already defined as a dependency

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
dhavalsharma: 3. there is no error_message in this call's return so returning the entire object

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
```diff
@@ -29,10 +35,16 @@ define([
...
     'my-leave.components',
+    'leave-absences.directives',
     'leave-absences.models',
+    'common.angularDate',
   ])
-  .run(['$log', function ($log) {
+  .run(['$log','DateFormat', function ($log, DateFormat) {
     $log.debug('app.run');
+    //set HR_Settings
+    DateFormat.getDateFormat().then(function(result){
```
AkA84: Two things here:

1. The `then` function is useless and confusing, since you're not doing anything with it
2. `HR_Settings.DATE_FORMAT` is already assigned in the [`my-leave.config`](https://github.com/civicrm/civihr/blob/f9cb7301357a0e47bd363663c05fa6ef1dad83a9/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/modules/config.js#L27) module, an the reason for doing it in the `resolve` property is that until the value is set, the application is not loaded. Instead what you are doing here is calling an async operation in the `.run()` method, basically creating a race condition, hoping that it fetched the `DATE_FORMAT` value before it is actually used anywhere in your code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-report.js
```diff
@@ -13,12 +13,11 @@ define([
     controllerAs: 'report',
     controller: [
       '$log', '$q', 'AbsencePeriod', 'AbsenceType', 'Entitlement', 'LeaveRequest',
-      'OptionGroup', 'dialog', 'HR_settings', controller
+      'OptionGroup', 'dialog', 'HR_settings', '$rootScope', controller
```
AkA84: Please put the angular-native modules first
```js
controller: [
  '$log', '$q', '$rootScope', 'AbsencePeriod', 'AbsenceType', 'Entitlement', 'LeaveRequest',

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-report.js [hunk 17]
```diff
@@ -478,6 +477,10 @@ define([
       })['value'];
     }
 
+    $rootScope.$on('LeaveRequest::new', function(event, message){
```
AkA84: There is a `init()` block at your disposal in the component, please use it

AkA84: Also, mind the whitespace
```js
$rootScope.$on('LeaveRequest::new', function (event, message) {
```

AkA84: Also, you are not using `even`, nor `message`, so why including it?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave.js
```diff
@@ -10,8 +10,7 @@ define([
     }],
     controllerAs: 'myleave',
     controller: ['$log', '$scope', '$uibModal', 'settings', function ($log, $scope, $modal, settings) {
-      console.log('CRM id of the currently logged in user: ', this.contactId);
-      $log.debug('Component: my-leave');
+      $log.debug('Component: my-leave', this.contactId);
```
AkA84: I think `this.contactId` can be removed, there is no point at this stage to keep logging it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
AkA84:
```js
 if(data && data.is_error == 1){
```
should be
```js
 if (data && data.is_error) {
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -52,10 +52,14 @@ define([
           public_holiday: isPublicHoliday || false
         };
 
-        deferred.resolve(this.sendGET('LeaveRequest', 'getbalancechangebyabsencetype', params, false)
+        this.sendGET('LeaveRequest', 'getbalancechangebyabsencetype', params)
```
AkA84: You are reverting the `false` parameter, used to disabled caching on the ajax call (introduced [here](https://github.com/civicrm/civihr/pull/1297/commits/92c64494ae2bb1223fed05ce22de6a9b52edb468#diff-2580b71d3d2a79bd0a3154318a126cb7R58))

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
AkA84: You are doing this check over and over again, can't you put this in a function?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+  'common/moment',
+  'common/services/api/option-group',
+  'common/services/hr-settings',
+], function (components, _, moment) {
+  'use strict';
+
+  components.controller('LeaveRequestPopupCtrl', [
+    '$log', '$q', '$uibModalInstance', 'AbsencePeriod', 'AbsenceType', 'Entitlement',
+    'Calendar', 'LeaveRequestInstance', 'LeaveRequest', 'api.optionGroup', 'baseData',
+    'PublicHoliday', 'HR_settings', '$rootScope',
```
AkA84: As before, please let `$rootScope` in alphabetical order alongside the other angular-native modules

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+    function ($log, $q, $modalInstance, AbsencePeriod, AbsenceType, Entitlement,
+      Calendar, LeaveRequestInstance, LeaveRequest, OptionGroup, baseData, PublicHoliday,
+      HR_settings, $rootScope) {
+      $log.debug('LeaveRequestPopupCtrl');
+
+      var $ctrl = this,
```
AkA84: Let's use the same `vm` pattern
```js
var vm = {};
```
Do you need it to be equal to `this`? Is there any property that is automatically attached to `this` that you need to use? If no, use the above ^, if yes, you can do
```js
var vm = Object.create(this);
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+      };
+      $ctrl.error = undefined;
+      $ctrl.loading = {};
+
+      // Create an empty leave request
+      $ctrl.leaveRequest = LeaveRequestInstance.init({
+        contact_id: baseData.contactId //resolved from directive
+      }, false);
+
+      (function () {
```
AkA84: I assume this is the your `init()` code, right? Then why don't you put 
```js
// Create an empty leave request
$ctrl.leaveRequest = LeaveRequestInstance.init({
  contact_id: baseData.contactId //resolved from directive
}, false);
```
in it as well?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
deb1990: @dhavalsharma I have updated the below comment with proper example, the idea is to send just the error message instead of the whole object. So in this case as we dont have an error_message sent from the server, we can specify a custom error message.

Example:
`deferred.reject(<custom error message>);`

For reference
https://github.com/civicrm/civihr/pull/1327/files#diff-2580b71d3d2a79bd0a3154318a126cb7R74

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+            return initDayTypesAndStatus()
+          });
+      })();
+
+      /**
+       * Initializes values for absence types and entitlements when the model is loaded
+       *
+       * @returns {Promise}
+       */
+      function initAbsenceTypesAndEntitlements() {
```
AkA84: The ordering of functions is all over the place, making the code harder to read. You have some private functions first, the methods exposed on the view, then some private functions again. Please make the order consistent: first the variables, then the methods exposed, then the init method, then the private functions. Please use the [my report component](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-report.js) as reference

```js
var // ....

vm.methodExposed1 = function () {
};

vm.methodExposed2 = function () {
};

vm.methodExposed3 = function () {
};

// init
(function () {
});

function privateFunction1 () {
}

function privateFunction2 () {
}

function privateFunction3 () {
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+      $ctrl.loading = {};
+
+      // Create an empty leave request
+      $ctrl.leaveRequest = LeaveRequestInstance.init({
+        contact_id: baseData.contactId //resolved from directive
+      }, false);
+
+      (function () {
+        $ctrl.loading.absenceTypes = true;
+        AbsencePeriod.current().then(function (apInstance) {
```
AkA84: Always try to flatten the list of `then`s
```js
AbsencePeriod.current()
  .then(function (apInstance) {
    $ctrl.period = apInstance;
  })
  .then(function () {
    return initAbsenceTypesAndEntitlements();
  })
  .then(function () {
    $ctrl.loading.absenceTypes = false;
  })
  .then(function () {
    return initDayTypesAndStatus()
  });
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+            $ctrl.period = apInstance;
+          })
+          .then(function () {
+            return initAbsenceTypesAndEntitlements().then(function () {
+              $ctrl.loading.absenceTypes = false;
+            });
+          })
+          .then(function () {
+            return initDayTypesAndStatus()
```
AkA84: Is it important to `return` this? It doesn't seem you are doing anything with it?

AkA84: Also, missing a `;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+          });
+      })();
+
+      /**
+       * Initializes values for absence types and entitlements when the model is loaded
```
AkA84: > when the model is loaded

What model?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+       *
+       * @returns {Promise}
+       */
+      function initAbsenceTypesAndEntitlements() {
+        // Fetch all the absence types, except for the sickness ones
+        return AbsenceType.all({
+            is_sickness: false
+          })
+          .then(function (absenceTypes) {
+            var absenceTypesIds = absenceTypes.map(function (item) {
```
AkA84: use meaningful names for variables, it's not `item`, it's `absenceType`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+              return item.id;
+            })
```
AkA84: missing `;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+
+            // And then for each of them get the remaining balance from the
+            // entitlements linked to them
+            Entitlement.all({
+                contact_id: $ctrl.leaveRequest.contact_id,
```
AkA84: You are using a 4-spaces indentation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+                period_id: $ctrl.period.id,
+                type_id: { in: absenceTypesIds
+                }
```
AkA84: No reason to have this `}` on a new line? Please rewrite as
```js
Entitlement.all({
  contact_id: $ctrl.leaveRequest.contact_id,
  period_id: $ctrl.period.id,
  type_id: { in: absenceTypesIds }
}, true)
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+              }, true) // `true` because we want to use the 'future' balance for calculation
+              .then(function (entitlements) {
+                // create a list of absence types with a `balance` property
+                $ctrl.absenceTypes = entitlements.map(function (entitlementItem) {
```
AkA84: There is a lot going on here, from `return AbsenceType.all()` (line 90) to `$ctrl.balance.opening = $ctrl.selectedAbsenceType.remainder;` (line 125). please refactor this code and split it into different smaller functions

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+                  var absenceType = absenceTypes.find(function (absenceTypeItem) {
```
AkA84: [`Array.prototype.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) is not supported by IE

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+                    return absenceTypeItem.id === entitlementItem.type_id;
+                  });
+
+                  return {
+                    id: entitlementItem.type_id,
+                    title: absenceType.title + ' ( ' + entitlementItem.remainder.current + ' ) ',
```
AkA84: Do we really need to do this? You are putting a presentation-only related information in this object. Sometimes you can't avoid (i did it myself in the report component), but is it really unavoidable? Couldn't you create this string directly in the template?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+                $ctrl.leaveRequestDayTypes = optionValues;
+              });
+          })
+          .then(function () {
+            return OptionGroup.valuesOf('hrleaveandabsences_leave_request_status')
+              .then(function (optionValues) {
+                $ctrl.leaveRequestStatuses = optionValues;
+                $ctrl.leaveRequest.status_id = getSpecificValueFromCollection($ctrl.leaveRequestStatuses, 'name', 'waiting_approval');
+              });
+          })
```
AkA84: missing `;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+
+      /**
+       * Pick a specific value out of a collection
+       *
+       * @param {array} the option group collection key
+       * @param {string} key - The sub-collection key
+       * @param {string} value - The sub-collection key's value to match
+       * @return {object}
+       */
+      function getSpecificValueFromCollection(collection, key, value) {
```
AkA84: You are using this only once, is this really necessary? This looks awfully generic for a one-time specific purpose use

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+              .then(function (optionValues) {
+                $ctrl.leaveRequestStatuses = optionValues;
+                $ctrl.leaveRequest.status_id = getSpecificValueFromCollection($ctrl.leaveRequestStatuses, 'name', 'waiting_approval');
+              });
+          })
+      }
+
+      /**
+       * Change handler for change request type like multiple or single. It will
+       * reset dates, balanes types.
```
AkA84: typo

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+       *
```
AkA84: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,492 @@
...
+       */
+      $ctrl.changeInNoOfDays = function () {
+        //reset dates
+        $ctrl.uiOptions.toDate = $ctrl.uiOptions.fromDate = undefined;
+        //reset balance change
+        $ctrl.balance = {
```
AkA84: It seems to me that a lot here could be put in a private `reset()` function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: the ui code will be simpler and hence kept this logic in controller

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: its just a helper function like other helper functions which are also used only once

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+        //reinitialize opening balance
+        $ctrl.balance.opening = $ctrl.selectedAbsenceType.remainder;
+      };
+
+      /**
+       * Whenever the absence type changes, update the balance opening.
+       * Also the balance change needs to be recalculated, if the `from` and `to`
+       * dates have been already selected
+       */
+      $ctrl.onAbsenceTypeChange = function () {
```
AkA84: Please name the method in a way that explain *what* it does, rather than *when* it does it. Based on your comment, it could be something like
```js
vm.updateBalance = function () {

}
```
?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+      };
+
+      /**
+       * This should be called whenever a date has been changed
+       *
+       * First it syncs `from` and `to` date, if it's in 'single day' mode
+       * Then, if all the dates are there, it gets the balance change
+       *
+       */
+      $ctrl.onDateChange = function (date, isFrom) {
```
AkA84: Same thing as before
> Please name the method in a way that explain what it does, rather than when it does it.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+          });
+        }
+      };
+
+      /**
+       * This should be called whenever a date has been changed
+       *
+       * First it syncs `from` and `to` date, if it's in 'single day' mode
+       * Then, if all the dates are there, it gets the balance change
+       *
```
AkA84: You are missing `@param`s

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+            $ctrl.uiOptions.showBalance = true;
+          }
+        } else {
+          if ($ctrl.uiOptions.fromDate) {
+            $ctrl.uiOptions.showBalance = true;
+          }
+        }
+
+        $ctrl.filterLeaveRequestDayTypes(date, !!isFrom)
+          .then(function () {
```
AkA84: There is too much going on here, please refactor it by splitting it into smaller functions

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+       * This method will be used on the view to return a list of available
+       * leave request day types (1/2 PM, Non working day, etc) for the given date
+       * (which is the date selected by the user via datepicker)
+       *
+       * If no date is passed, then no list is returned
+       *
+       * @param  {Date} dateParam
+       * @return {Promise} of array
+       */
+      $ctrl.filterLeaveRequestDayTypes = function (dateParam, isFrom) {
```
AkA84: This method is not used in the view, there is no reason to expose it in the scope

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+        } else {
+          if ($ctrl.uiOptions.fromDate) {
+            $ctrl.uiOptions.showBalance = true;
+          }
+        }
+
+        $ctrl.filterLeaveRequestDayTypes(date, !!isFrom)
+          .then(function () {
+            if (isFrom) {
+              $ctrl.leaveRequest.from_date = convertDateFormatToServer($ctrl.uiOptions.fromDate);
```
AkA84: Why aren't you plugging `leaveRequest.from_date` directly in the view? Is it because of the datepicker? (i remember that it created some problems with the way it formats the dates)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+      };
+
+      /**
+       * This method will be used on the view to return a list of available
+       * leave request day types (1/2 PM, Non working day, etc) for the given date
+       * (which is the date selected by the user via datepicker)
+       *
+       * If no date is passed, then no list is returned
+       *
+       * @param  {Date} dateParam
```
AkA84: You are missing the `@param` for `isFrom`

Also, why not calling the first param just `date`? We know it's a param, what's the point of calling it `date*Param*`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+
+        if (!dateParam) {
+          deferred.reject([]);
+        }
+
+        var date = convertDateFormatToServer(dateParam);
+        // Make a copy of the list
+        var listToReturn = $ctrl.leaveRequestDayTypes.slice(0);
+
+        PublicHoliday.isPublicHoliday(date).then(function (result) {
```
AkA84: The `then` callback is massive, you need to split it into smaller chunks. This is not maintainable

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+          }
+        } else {
+          if ($ctrl.uiOptions.fromDate) {
+            $ctrl.uiOptions.showBalance = true;
+          }
+        }
+
+        $ctrl.filterLeaveRequestDayTypes(date, !!isFrom)
+          .then(function () {
+            if (isFrom) {
```
AkA84: An idea: instead of having a boolean `isFrom`, you could have something like `type`, which is either `"from"` or `"to"`
```js
$ctrl.onDateChange = function (date, type) {
  type = (typeof type === 'undefined' ? 'from');
```
this way the code could be simplified by a lot by referencing the fields dynamically 
```js
$ctrl.leaveRequest[type + '_date'] = convertDateFormatToServer($ctrl.uiOptions[type + 'Date']);
$ctrl.loading[type + 'DayTypes'] = false;

```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+       * This should be called whenever a date has been changed
+       *
+       * First it syncs `from` and `to` date, if it's in 'single day' mode
+       * Then, if all the dates are there, it gets the balance change
+       *
+       */
+      $ctrl.onDateChange = function (date, isFrom) {
+        if ($ctrl.uiOptions.multipleDays) {
+          if ($ctrl.uiOptions.toDate && $ctrl.uiOptions.fromDate) {
+            $ctrl.uiOptions.showBalance = true;
```
AkA84:
```js
if ($ctrl.uiOptions.multipleDays) {
  $ctrl.uiOptions.showBalance = !!$ctrl.uiOptions.toDate && !!$ctrl.uiOptions.fromDate;
} else {
  $ctrl.uiOptions.showBalance = !!$ctrl.uiOptions.fromDate;
}

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js (hunk 42 above)
Comment (AkA84) [inline uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js]: Also, does the content of the `then` callback (assigning the selected days to the leave request, matching fromDate to toDate in case it's a single day, calculating the balance change), really need to happen only after you fetched the day types of the selected days? Can't it happen in parallel? It doesn't seem you have any dependency on the result of the `filterLeaveRequestDayTypes` call?

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js [hunk 49]
```diff
@@ -0,0 +1,516 @@
...
+            }
+
+            if (!$ctrl.uiOptions.multipleDays) {
+              $ctrl.uiOptions.toDate = $ctrl.uiOptions.fromDate;
+              $ctrl.uiOptions.selectedToType = $ctrl.uiOptions.selectedFromType;
+              $ctrl.leaveRequest.to_date = $ctrl.leaveRequest.from_date;
+              $ctrl.leaveRequest.to_date_type = $ctrl.leaveRequest.from_date_type;
+            }
+
+            if (!$ctrl.uiOptions.showBalance) {
```
AkA84: I'm not sure that this is the best approach, you are basing your controller logic on a variable which is explicitly only about the UI. Why not deciding to return simply by checking if both `$ctrl.leaveRequest.from_date` and `$ctrl.leaveRequest.to_date` are present? You are actually doing it already on line `132`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+              return
```
AkA84: Missing `;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+            }
+          });
+      };
+
+      /**
+       * Calculate change in balance, it updates local balance variables
+       *
+       * @return {Promise}
+       */
+      $ctrl.calculateBalanceChange = function () {
```
AkA84: Another example of just *too much* going on in a given method. Methods should be single-purpose and as short as possible. Please refactor this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+
+        return deferred.promise;
+      };
+
+
+      /**
+       * Checks if submit button can be enabled for user and returns true if succeeds
+       *
+       * @returns {Boolean}
+       **/
```
AkA84: Wrong closing line
```js
*
*/
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+      $ctrl.canSubmit = function () {
+        if ($ctrl.leaveRequest.from_date && $ctrl.leaveRequest.to_date &&
```
AkA84:
```js
return $ctrl.leaveRequest.from_date && $ctrl.leaveRequest.to_date && $ctrl.leaveRequest.to_date_type && $ctrl.leaveRequest.from_date_type;

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js [hunk 54]
```diff
@@ -0,0 +1,516 @@
...
+        /* current absence type ($ctrl.leaveRequest.type_id) doesn't allow that */
+        if ($ctrl.balance.closing < 0 && $ctrl.selectedAbsenceType.allow_overuse == '0') {
+          // show an error
+          $ctrl.error = 'You are not allowed to apply leave in negative';
+          return;
+        }
+
+        $ctrl.error = undefined;
+        $ctrl.leaveRequest.isValid().then(function () {
+            $ctrl.leaveRequest.create()
```
AkA84: 4-spaces indentation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -0,0 +1,516 @@
...
+          //filter breakdowns
+          var begin = ((this.currentPage - 1) * this.numPerPage),
+            end = begin + this.numPerPage;
+
+          this.filteredbreakdown = $ctrl.balance.change.breakdown.slice(begin, end);
+        }
+      };
+      $ctrl.uiOptions = {
+        showDatePickerFrom: false,
+        showDatePickerTo: false,
```
dhavalsharma: it looks good but makes the reaching of end attributes longer like `$ctrl.uiOptions.datePicker.show.from` , i think too many dots there

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: its a change handler and they are named by _when_ and not _what_ they do

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: its a change handler and they are named by when and not what they do

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: calculateBalanceChange needs both day types to be defined for multiple days, it can change based on the date selected and therefore is in the callback, not parallel

similarly the setting of to date/types depends on whether its request for single/multiple days and which day type is selected

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: yes, included it before but formatting was messing so reverted to 2 step

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: The when is defined on the view, when you do stuff like `ng-change="ctrl.updateBalance()"`. The controller should define its api, while the view decides when to call it. Not to mention that `onAbsenceTypeChange` doesn't give any information whatsoever of what the method does. Please change it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+            expect($ctrl.balance.change.amount).toEqual(0);
+          });
+        });
+      });
+
+      describe('calendar', function () {
+
+        it('no dates are selected by default', function () {
+          expect($ctrl.uiOptions.fromDate).not.toBeDefined();
+          expect($ctrl.uiOptions.toDate).not.toBeDefined();
```
AkA84: Again, something that should be tested in the "initialization" block. Basically you should have one
```js
describe('when initialized', function () {

});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+          beforeEach(function () {
+            $ctrl.onDateChange(new Date(), true);
+            $scope.$digest();
+          });
+
+          it('can change date', function () {
+            expect(moment($ctrl.leaveRequest.from_date, serverDateFormat, true).isValid()).toBe(true);
+          });
+
```
AkA84: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          beforeEach(function () {
+            $ctrl.onDateChange(new Date(), false);
+            $scope.$digest();
+          });
+
+          it('can change date', function () {
+            expect(moment($ctrl.leaveRequest.to_date, serverDateFormat, true).isValid()).toBe(true);
+          });
+
```
AkA84: empty line (*applies for all next occurrences*)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+        });
+      });
+
+      describe('day types', function () {
+
+        it('by default no selection', function () {
+          expect($ctrl.uiOptions.selectedFromType).not.toBeDefined();
+          expect($ctrl.uiOptions.selectedToType).not.toBeDefined();
```
AkA84: Another test for initialization

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+          it('by default selects first day type', function () {
+            expect($ctrl.uiOptions.selectedFromType).toEqual($ctrl.leaveRequestFromDayTypes[0]);
+          });
+
+          it('will filter based on current date', function () {
+            expect($ctrl.filterLeaveRequestDayTypes).toHaveBeenCalledWith(jasmine.any(Date), jasmine.any(Boolean));
+          });
+
+          describe('is public holiday', function () {
```
AkA84: I'm not sure what you're testing here? You are checking what? In case the "from date" is a public holiday? What about the other day types?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+            beforeEach(function () {
+              $ctrl.uiOptions.fromDate = new Date('01/01/2016');
+              $ctrl.filterLeaveRequestDayTypes($ctrl.uiOptions.fromDate, true);
+              $scope.$digest();
+            });
+
+            it('should', function () {
+              expect($ctrl.leaveRequestFromDayTypes.length).toEqual(1);
```
AkA84: Not sure what is that you are testing

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+        describe('to', function () {
+
+          beforeEach(function () {
+            $ctrl.onDateChange(new Date(), false);
+            $scope.$digest();
+          });
+
+          it('by default selects first day type', function () {
+            expect($ctrl.uiOptions.selectedToType).toEqual($ctrl.leaveRequestToDayTypes[0]);
```
AkA84: Should go in initialization code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          });
+
+          describe('change selection', function () {
```
AkA84: This is confusing, you have "from" and "to" day types, but for each of them you are doing different tests? Like for "from" you are doing that "public holidays" test, but for "to" you are doing this "change selection" test?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+            beforeEach(function () {
+              before = $ctrl.uiOptions.selectedToType;
+              $ctrl.uiOptions.selectedToType = $ctrl.leaveRequestToDayTypes[1];
+              $ctrl.calculateBalanceChange();
+              after = $ctrl.uiOptions.selectedToType;
+              $scope.$digest();
+            });
+
+            it('success', function () {
```
AkA84: Please give meaningful titles to the tests

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          });
+        });
+      });
+
+      describe('calculate balance', function () {
+
+        describe('opening', function () {
+
+          it('has default', function () {
+            expect($ctrl.balance.opening).toEqual(0);
```
AkA84: initialization test

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          });
+
+          describe('when day type changed', function () {
+
+            describe('for single day', function () {
+
+              beforeEach(function () {
+                //select half_day_am  to get single day mock data
+                //$ctrl.uiOptions.multipleDays = false;
+                //$ctrl.uiOptions.fromDate = new Date();
```
AkA84: Commented code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+          it('has default', function () {
+            expect($ctrl.balance.opening).toEqual(0);
+          });
+
+        });
+
+        describe('change', function () {
+
+          it('has default', function () {
```
AkA84: What does it mean "change, it has default"?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+              it('will update balance', function () {
+                expect($ctrl.balance.change.amount).toEqual(-0.5);
+              });
+
+              it('will update closing balance', function () {
+                expect($ctrl.balance.closing).toEqual(-0.5);
+              });
+            });
+            describe('for multiple days', function () {
```
AkA84: Missing empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+                //select all_day to get multiple day mock data
+                $ctrl.onDateChange(new Date(), true);
+                $ctrl.onDateChange(new Date(), false);
+                $ctrl.uiOptions.selectedFromType = optionGroupMock.specificObject('hrleaveandabsences_leave_request_day_type', 'name', 'all_day');
+                $ctrl.calculateBalanceChange();
+                $scope.$digest();
+              });
+
+              it('will update change amount', function () {
+                expect($ctrl.balance.change.amount).toEqual(-2);
```
AkA84: But, where are you getting this very specific change amounts? You are not giving any specific dates, so how do you know the balance change is `-2`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+            });
+          });
+        });
+
+        describe('change details', function () {
+
+          describe('hidden', function () {
+
+            it('by default', function () {
+              expect($ctrl.uiOptions.isChangeExpanded).toBeFalsy();
```
AkA84: initialization test

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          });
+
+          describe('when expanded', function () {
+
+            beforeEach(function () {
+              $ctrl.uiOptions.isChangeExpanded = true;
+            });
+
+            it('will display balance breakdown', function () {
+              expect($ctrl.uiOptions.isChangeExpanded).toBeTruthy();
```
AkA84: Meaningless test, you have just changed manually the value of `$ctrl.uiOptions.isChangeExpanded`, what's the point testing it?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+            });
+
+            describe('during pagination', function () {
+
+              describe('init', function () {
+
+                it('should', function () {
+                  expect($ctrl.pagination.totalItems).toEqual(0);
```
AkA84: initialization code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+                  $ctrl.onDateChange($ctrl.uiOptions.toDate, false);
+                  $scope.$digest();
+                });
+
+                it('will select default page', function () {
+                  expect($ctrl.pagination.currentPage).toEqual(1);
+                });
+
+                it('should set totalItems', function () {
+                  expect($ctrl.pagination.totalItems).toBeGreaterThan(0);
```
AkA84: Where are you getting the items in the test?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          });
+        });
+      });
+
+      describe('save leave request', function () {
+
+        describe('with invalid fields', function () {
+
+          it('does not have required fields', function () {
+            expect($ctrl.leaveRequest.from_date).not.toBeDefined();
```
AkA84: I don't understand this test, you are not even submitting the leave request? The title seems incoherent ("save leave request / with invalid fields / does not have required fields") and in the end this looks simply an initialization test, since again you are not saving the leave request before testing the expectation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+
+          describe('on submit', function () {
+
+            beforeEach(function () {
+              $ctrl.submit();
+              $scope.$digest();
+            });
+
+            it('should fail', function () {
+              expect($ctrl.error).toEqual(jasmine.any(Object));
```
AkA84: Shouldn't you check what the error is? Or at least, given that `$ctrl.error` is an object, check what the structure is?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+        describe('with valid fields', function () {
+
+          beforeEach(function () {
+            $ctrl.onDateChange(new Date(), true);
+            $ctrl.onDateChange(new Date(), false);
+            $scope.$digest();
+          });
+
+          it('has all required fields', function () {
+            expect($ctrl.leaveRequest.id).not.toBeDefined();
```
AkA84: Checking that the `id` is *not* defined when the test is "it has all required fields" is very weird, do you even need to check this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+            expect($ctrl.leaveRequest.from_date).toBeDefined();
```
AkA84: Why is `from_date` required but not `to_date`? What about the status id? Or the contact id? Where are all the other mandatory fields?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+          describe('on submit', function () {
+
+            beforeEach(function () {
+              spyOn($rootScope, '$emit');
+              $ctrl.submit();
+              $scope.$digest();
+            });
+
+            it('is successful', function () {
+              expect($ctrl.error).not.toEqual(jasmine.any(Object));
```
AkA84: `not.toBeDefined()`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+                $ctrl.selectedAbsenceType = $ctrl.absenceTypes[1];
+                $ctrl.uiOptions.toDate = $ctrl.uiOptions.fromDate = new Date();
+                $ctrl.onDateChange($ctrl.uiOptions.fromDate, true);
+                $ctrl.onDateChange($ctrl.uiOptions.toDate, false);
+                $scope.$digest();
+                $ctrl.submit();
+                $scope.$digest();
+              });
+
+              it('will not save', function () {
```
AkA84: The way tests are structured is just as important as the content of the tests itself, it has a major impact on how easy they are to read
```js
describe('when balance change is negative', function () {


  describe('when the absence type does not allow overuse', function () {
    
    it('does not save', function () {

    });
  });

  describe('when the absence type allows overuse', function () {
    
    it('does save', function () {

    });
  });
});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/directives/leave-request-popup_test.js
```diff
@@ -0,0 +1,56 @@
...
+    'use strict';
+
+    describe('leaveRequestPopup', function () {
+      var $compile, $log, $rootScope, directive, $uibModal,
+        $controllerScope, $provide, DateFormat;
+
+      /**
+       * Creates and compiles the directive
+       */
+      function compileDirective() {
```
AkA84: Put this at the bottom, leverage hoisting to get simple utility functions out of the way

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
dhavalsharma: it is removed, the link gives 404

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: too many dots in object and moreover a separate object is required to set `datePickerOptions`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: they are change event handlers and so do what happens on change, i think its standard to name event handlers with on*

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: they are change event handlers and so do what happens on change, i think its standard to name event handlers with on*

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: not possible in parallel as day types and dates need to be set before calculate balance change can be called

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: yes

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/calendar-instance.js
```diff
@@ -6,8 +6,8 @@ define([
   'use strict';
 
   instances.factory('CalendarInstance', [
-    'ModelInstance',
-    function (ModelInstance) {
+    'ModelInstance', '$log',
```
dhavalsharma: not in scope

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+        describe('from and to dates are selected', function () {
+
+          beforeEach(function () {
+            $ctrl.uiOptions.toDate = $ctrl.uiOptions.fromDate = new Date();
+            $ctrl.onDateChange($ctrl.uiOptions.fromDate, true);
+            $ctrl.onDateChange($ctrl.uiOptions.toDate, false);
+            $scope.$digest();
+          });
+
+          it('balance is displayed', function () {
```
dhavalsharma: yes

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -0,0 +1,591 @@
...
+            beforeChangeAbsenceType = $ctrl.selectedAbsenceType;
+            $ctrl.selectedAbsenceType = $ctrl.absenceTypes[1];
+
+            $ctrl.onAbsenceTypeChange();
+            afterChangeAbsenceType = $ctrl.selectedAbsenceType;
+            $scope.$digest();
+          });
+
+          it('selects another absence type', function () {
+            expect(beforeChangeAbsenceType.id).not.toEqual(afterChangeAbsenceType.id);
```
dhavalsharma: also added test case to test for type_id

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: here the selection is changed so its not initialization, its something that happens after initialization and hence a separate block for 'single selection'

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: removed, this is already tested in model instance

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: could not locate this

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: moved to initialization

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: from test data

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: test data from api mock

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: $ctrl.error can be undefined or set to an object and therefore i think the test is sufficient

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
dhavalsharma: moved id into separate test case

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/directives/leave-request-popup.js
```diff
@@ -0,0 +1,44 @@
...
+
+          element.on('click', function (event) {
+            //to set HR_settings DateFormat
+            DateFormat.getDateFormat().then(function (result) {
+              scope.openComponentModal = function () {
+                $modal.open({
+                  templateUrl: settings.pathTpl + 'directives/leave-request-popup.html',
+                  animation: scope.animationsEnabled,
+                  controller: 'LeaveRequestPopupCtrl',
+                  controllerAs: '$ctrl',
```
dhavalsharma: from uibmodel documentation `controllerAs (Type: string, Example: ctrl) - An alternative to the controller-as syntax. Requires the controller option to be provided as well.`

controllerAs is string, not sure what you mean here?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/apis/option-group-api-mock.js
```diff
@@ -0,0 +1,17 @@
+define([
+  'mocks/module',
+  'mocks/data/option-group-mock-data',
+  'common/angularMocks',
+], function (mocks, mockData) {
+  'use strict';
+
+  mocks.factory('OptionGroupAPIMock', ['$q', function ($q) {
```
dhavalsharma: yes, but its not easily extensible in current form

PR #1338, opened 2017-01-05
PCHR-1850: Extend LeaveRequest.getFull API to include the TOIL Request balance changes

tunbola:
This PR updates the LeaveRequest.getFull() API to include the balance_change for TOIL leave request.
Currently, the API does not return the balance changes for TOIL requests because the balance change for a TOIL leave request is created differently from that of a normal leave request.
This PR takes this difference into consideration and makes changes that will allow balance changes for TOIL requests to returned in the LeaveRequest.getFull() API endpoint

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -3,9 +3,7 @@
 use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
-use CRM_HRLeaveAndAbsences_BAO_WorkPattern as WorkPattern;
 use CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern as ContactWorkPattern;
-use CRM_HRLeaveAndAbsences_BAO_WorkDay as WorkDay;
 
 class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsences_DAO_LeaveBalanceChange {
 
@@ -340,6 +338,23 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     }
 
     return $balance;
+  }
+
+  /**
+   * Returns the LeaveBalanceChange amount for the TOIL request ID
+   * Also note that a TOIL Request creates a single LeaveBalanceChange record.
+   *
+   * @param int $toilRequestID
+   *
+   * @return float
+   */
+  public static function getAmountForTOILRequest($toilRequestID) {
+    $balanceChange = new self();
+    $balanceChange->source_id = $toilRequestID;
+    $balanceChange->source_type = self::SOURCE_TOIL_REQUEST;
+    $balanceChange->find(true);
+
+    return $balanceChange->amount;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -343,6 +341,23 @@ public static function getTotalBalanceChangeForLeaveRequest(LeaveRequest $leaveR
...
 
   /**
+   * Returns the LeaveBalanceChange for the TOIL request ID
+   * Also note that a TOIL Request creates a single LeaveBalanceChange record.
+   *
+   * @param int $toilRequestID
+   *
+   * @return float
+   */
+  public static function getBalanceChangeForToilLeaveRequest($toilRequestID) {
```
davialexandre: There are 2 things that I think we can improve in this method name
- TOIL Leave Request makes me think this is something related to the Leave Request associated with the TOIL Request. So, to avoid confusion, I would probably be better to use TOILRequest
- Since it returns the amount of the balance change, it would be better to reflect that on the name

With that said, my suggestion for the method name is `getAmountForTOILRequest()`. Note that I also removed the BalanceChange from the name. The reason is that this is a static method, so we'll always call it like `LeaveBalanceChange::getAmountForTOILRequest()`, so there's no need for the redundancy on the method name.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -343,6 +341,23 @@ public static function getTotalBalanceChangeForLeaveRequest(LeaveRequest $leaveR
   }
 
   /**
+   * Returns the LeaveBalanceChange for the TOIL request ID
```
davialexandre: The method description is not accurate. It says it returns a LeaveBalanceChange, but it actually returns the amount of a LeaveBalanceChange

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -1038,5 +1039,25 @@ private function deleteDefaultWorkPattern() {
     $workPatternTable = WorkPattern::getTableName();
     CRM_Core_DAO::executeQuery("DELETE FROM {$workPatternTable} WHERE is_default = 1");
   }
+
+  public function testGetBalanceChangeForToilLeaveRequest() {
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'allow_accruals_request' => true,
+    ]);
```
davialexandre: since you're using `fabricateWithoutValidation()` I think you don't need this absence type here.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -687,6 +687,78 @@ public function testGetFullShouldNotIncludeTheDatesIfTheReturnOptionIsNotEmptyAn
...
+      [
+        'period_start_date' => '2016-01-01',
+        'period_end_date' => '2016-10-01'
+      ]
+    );
+
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'must_take_public_holiday_as_leave' => true,
+      'allow_accruals_request' => true,
+    ]);
```
davialexandre: you probably don't need this absence type here as well

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
tunbola: The saveBalanceChange method on the TOIL request BAO calls the calculateToilExpiryDate method of the AbsenceType BAO. This will throw an exception if the absence type does not allow accrual request.

davialexandre: It calls calculateToilExpiryDate only if the expiry_date is null (it's something I've changed recently in here https://github.com/civicrm/civihr/commit/e59c258c2bbbee75dbbe52cca39523525ed79777), which is not the case on the tests

PR #1337, opened 2017-01-05
PCHR-1720: Adds Styles form PCHR-1353 + fix for labels

bmnepali:
## Issue:
Styles form PCHR-1353 were not merged which halted css preprocessing (SASS). Labels for the Contact->Personal Details ->Extended Demographics edit form were hidden, as a result they were invisible. 

## Solution:
1. Styles form PCHR-1353 were extracted and applied in this fix.
2. Added style for labels for Extended Demographics to be visible.

### Before:
extdem

### After:
screen shot 2017-01-05 at 3 15 27 pm

PR #1339, opened 2017-01-05
PCHR-1827: Update main menu style

AkA84:
This PR changes the style of the main civicrm menu

**Before**
before

**After**
after

### Extension
The changes are stored in the org.civicrm.bootstrapcivicrm extension, in the civicrm/ folder, being part of the css suite that overrides the default styling of any core civicrm elements

### Arrows
The arrow shown in an item with sub-items, is an image that we can't get rid of without using javascript. So in hrui.js the image is swapped with a font-awesome icon
```js
// Remove the arrow for menu items with sub-items, and replaces it
// with a font awesome caret
$(document).ready(function () {
  $('#root-menu-div .menu-item-arrow').each(function ($element) {
    var $arrow = $(this);

    $arrow.before('<i class="fa fa-caret-right menu-item-arrow"></i>');
    $arrow.remove();
  });
});
```
Notice that the swapping happens on document.ready and not on `crmLoad`, as when the latter is fired, the arrows are still not in the DOM

**Before**
arrow-before

**After**
arrow-after


org.civicrm.bootstrapcivicrm/scss/civicrm/common/_buttons.scss
```diff
@@ -1,4 +1,6 @@
 @import '../../../org.civicrm.bootstrap/scss/vendor/bootstrap/buttons';
+@import '../../../org.civicrm.bootstrap/scss/vendor/bootstrap/button-groups';
```
AkA84: This is already done in #1337, but i needed to do it here as well because I don't which PR is going to be merged first

org.civicrm.bootstrapcivicrm/scss/civicrm/main-menu/_main-menu.scss
```diff
@@ -0,0 +1,76 @@
+#civicrm-menu {
```
deb1990: For all the colors in this file, can we move them to variables?

org.civicrm.bootstrapcivicrm/scss/civicrm/main-menu/_main-menu.scss
```diff
@@ -73,7 +75,7 @@
       }
 
       &.menu-separator {
-        border-color: $gray-darker;
+        border-color: #D3DEE2;
```
brunoqueiros: Use variable to store the colors.

PR #1341, opened 2017-01-07
PCHR-1265: Add the "expired" param to the LeaveRequest.getBalanceChangeByAbsenceType API

davialexandre:
This PR adds the `expired` param to the `LeaveRequest.getBalanceChangeByAbsenceType` API. 

This param is not mandatory and it's default value is `false`. When it's true, the API will return only the expired balance for that Absence Type. Note that, even though the endpoint is part of the LeaveRequest entity, it will also return any expired Brought Forward, which is not linked to any Leave Request. This was done in order to be able to get all the expired balance in a single API call.

To support this new API endpoint, a few changes had to be implemented on the LeaveBalanceChange BAO methods related to getting the balance for a Period Entitlement and LeaveRequests. More specifically, the methods `getLeaveRequestBalanceForEntitlement()` and `getBalanceForEntitlement()` were updated to also consider the balance changes caused by TOIL Requests. Next, the `getBalanceForEntitlement()` method received a new `expiredOnly` param, which, when it's true, will make the method return only the number of days expired for the given LeavePeriodEntitlement.

### Example
Say you have Absence Types 1 and 2. For Absence Type 1, you have 5 days Brought Forward with 3 days expired. For Absence Type 2, you have a TOIL Request with 2 days accrued and 1 day expired. If you call

```php
civicrm_api3('LeaveRequest', 'getBalanceChangeByAbsenceType', [
    'contact_id' => 1,
    'period_id' => 1
]);
```

It will  return:
```json
{
    "is_error": 0,
    "version": 3,
    "count": 3,
    "values": {
        "1": 0,
        "2": 2
    }
}
```

Note that, for Absence Type 1, the balance is 0 because there are no Leave Requests for it. The Brought Forward days are part of the entitlement and hence not returned here.

Now, if you call the API like this:

```php
civicrm_api3('LeaveRequest', 'getBalanceChangeByAbsenceType', [
    'contact_id' => 1,
    'period_id' => 1,
    'expired' => true
]);
```

It will  return:
```json
{
    "is_error": 0,
    "version": 3,
    "count": 3,
    "values": {
        "1": -3,
        "2": -1
    }
}
```

Note now that the -3 expired days from the Brought Forward were returned, as well as the 1 expired day from the TOIL Request.

### What else is included in this PR?
The last two commits include some cleanup, code style fixes and the removal of some unecessary code in order to make the tests run a bit faster. Basically, a lot of unecessary calls to the Contact and Absence Type fabricator were removed. This was possible because these specific tests run a an environment were fk check is disabled on the database, so intead of having to pass an actual Contact or Absence Type ID, we can just pass an integer.


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -562,6 +695,85 @@ public function testLeaveRequestBalanceForEntitlementWithParamsToBothExcludeAndI
...
+    //This balance change should not be included on the count
+    LeaveBalanceChangeFabricator::fabricate([
+      'source_id' => $toilBalanceChange->source_id,
+      'source_type' => $toilBalanceChange->source_type,
+      'expiry_date' => CRM_Utils_Date::processDate($toilBalanceChange->expiry_date),
+      'type_id' => $toilBalanceChange->type_id,
+      'amount' => -1,
+      'expired_balance_change_id' => $toilBalanceChange->id
+    ]);
+
```
tunbola: Why can't the `createExpiredTOILRequestBalanceChange` in LeaveBalanceChangeHelpersTrait.php be used here?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -451,28 +451,41 @@ public static function findOverlappingLeaveRequests($contactID, $fromDate, $toDa
...
    *
    * @return array
    */
   public static function getBalanceChangeByAbsenceType(
     $contactID,
     $periodID,
     $leaveRequestStatus = [],
-    $publicHolidaysOnly = false
+    $publicHolidaysOnly = false,
+    $expiredOnly = false
```
tunbola: Wondering if expired param should actually be on this method. After going through the method and the tests, its obvious that when the expiredOnly param is true, the method actually returns expired toil leave requests as well as expired brought forward entitlements which are not leave requests or related to one.
The method for returning expired balance changes might actually be well suited for the LeavePeriodEntitlement BAO. (just my thought) which is just the `getLeaveRequestBalance()` BAO method of the LeavePeriodEntitlement BAO method with an expiredOnly parameter to be added.

davialexandre: To be honest, I think this whole method does not belong to the LeaveRequest BAO. If you take a look at its implementation, it doesn't use any of that BAO's methods and gets all of it's data from LeaveBalanceChange and LeavePeriodEntitlement. There are two places where I think it should probably be:

- On the `civicrm_api3_leave_request_getbalancechangebyabsencetype`() function (yeah, directly on the API function. I think this is ok because this is the only place the method will be used anyway and it also avoids duplication of tests between BAO and API)
- On a separate class (like an API service/class, for example). The problem would only the the duplication of tests between API and Service

About using `getLeaveRequestBalance()` for it. It's not exactly the same thing. That method is specific about Leave Requests and including the Brought Forward on it would even be more confusing. That is the reason I also added the param to the `LeaveBalanceChange.getBalanceForEntitlement()` method. The Brought Forward is part of the entitlement and we can think of accrued TOIL as being part of the entitlement as well, since, event though it's stored as a Leave Request internally, it add days to the entitlement.

tunbola: Since the Brought Forward is part of the entitlement and we can think of accrued TOIL as being part of the entitlement as well, then probably we should have a method on the LeavePeriodEntitlement BAO then that returns the expired balance change. 
It being on the `civicrm_api3_leave_request_getbalancechangebyabsencetype()` makes it look as if the balance change is directly related to Leave Request and its not.
Or probably on an API method on leave_period_entitlement?

davialexandre: Yeah, that would be the perfect scenario for us, but not for the frontend guys :D If we opt for the new API in leave_period_entitlement then they would need to implement a new service (a lot of code and new tests) to call this new API, while with the param option all they have to do is add it to the current service and add a couple of tests to it. There's also the fact that with a new API, they would need 2 calls to fetch the data. 

Another reason to have this as a single API call is the way the information will be displayed:
selecao_002

As you can see, it's a single row displaying everything that has expired (currently Brought Forward + TOIL, but it can also include other things in the future).

So, what I'm trying to say is that even though I agree with you about separating things, from an API/frontend point of view it's better to keep them together, and the place that makes more sense to do that is on the LeaveRequest API (though it might be a bit confusing, most of the information comes from Leave Requests balance changes)

tunbola: As much as I don't like this too, but looking at it from the frontend perspective, It think it will be fine to go ahead with the first option you suggested earlier.

`On the civicrm_api3_leave_request_getbalancechangebyabsencetype() function (yeah, directly on the API function. I think this is ok because this is the only place the method will be used anyway and it also avoids duplication of tests between BAO and API)`

PR #1342, opened 2017-01-09
PCHR-1852: Install CiviCRM views integration module after civihr installation

omarabuhussein:
This installs (CiviCRM views integration module) by default when installing civihr . This module is part of civihr employee portal and it is added in this PR : https://github.com/compucorp/civihr-employee-portal/pull/258


davialexandre:
Closing this as it has gone stale

PR #1344, opened 2017-01-10
PCHR-1683: Update TOIL Requests according to changes in Absence Type settings

tunbola:
This PR takes care of updates/changes to TOIL requests when the TOIL accrual settings for an Absence Type is updated.

The following situations were taken care of in this PR:

- A test was added to verify that when an admin decreases the max number of days that can be accrued for an Absence Type, for open TOIL Request, the system "protects" itself and will not allow the request to be approved if the request amount is more than the new max.

- When an admin disables TOIL for an Absence Type, for all non-expired TOIL Requests, the TOIL requests + leave requests + balance changes + leave request dates associated with the TOIL is deleted

For situations when increase the TOIL expiry duration for an Absence Type is increase or decreased,
nothing is done because there's no way to distinguish between manually and automatically calculated TOIL expiry date.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
@@ -29,7 +29,7 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
     }
 
     $instance = new self();
-    //set from_date_type and to_date_type to be full day bxy default
+    //set from_date_type and to_date_type to be full day by default
     $dateTypeOptions = array_flip(LeaveRequest::buildOptions('from_date_type'));
     $params['from_date_type'] = $dateTypeOptions['All Day'];
     if(!empty($params['to_date'])){
@@ -234,11 +234,14 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
    *
    * @param int $absenceTypeID
    *   The absence Type that TOIL requests is to be deleted for.
-   * @param CRM_HRLeaveAndAbsences_BAO_AbsencePeriod $absencePeriod
+   * @param DateTime $startDate
+   *   Records linked to LeaveRequests with from_date >= this date will be deleted
+   * @param DateTime|null $endDate
+   *   If this is present records linked to LeaveRequests with to_date <= this date will be deleted
    * @param boolean $nonExpiredOnly
    *   Whether to delete only records linked to non expired balance changes
    */
-  public static function deleteAllForAbsenceType($absenceTypeID, AbsencePeriod $absencePeriod, $nonExpiredOnly = false) {
+  public static function deleteAllForAbsenceType($absenceTypeID, DateTime $startDate, DateTime $endDate = null, $nonExpiredOnly = false) {
     $leaveBalanceChangeTable = LeaveBalanceChange::getTableName();
     $toilRequestTable = TOILRequest::getTableName();
     $leaveRequestTable = LeaveRequest::getTableName();
@@ -248,21 +251,25 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
               INNER JOIN {$leaveBalanceChangeTable} bc ON bc.source_id = tr.id AND bc.source_type = %1
               INNER JOIN {$leaveRequestTable} lr ON tr.leave_request_id = lr.id
               INNER JOIN {$leaveRequestDateTable} lrd ON lr.id = lrd.leave_request_id
-              WHERE lr.from_date >= %2 AND lr.to_date <= %3
-              AND lr.type_id = %4";
+              WHERE lr.type_id = %2
+              AND lr.from_date >= %3
+              ";
 
-    if ($nonExpiredOnly) {
-      $query .= " AND bc.expiry_date >= %5";
-    }
-
-    $today = new DateTime('today');
     $params = [
       1 => [LeaveBalanceChange::SOURCE_TOIL_REQUEST, 'String'],
-      2 => [$absencePeriod->start_date, 'String'],
-      3 => [$absencePeriod->end_date, 'String'],
-      4 => [$absenceTypeID, 'Integer'],
-      5 => [$today->format('Y-m-d'), 'String'],
+      2 => [$absenceTypeID, 'Integer'],
+      3 => [$startDate->format('Y-m-d'), 'String'],
     ];
+
+    if ($endDate) {
+      $query .= " AND lr.to_date <= %5";
+      $params[5] = [$endDate->format('Y-m-d'), 'String'];
+    }
+
+    if ($nonExpiredOnly) {
+      $query .= " AND bc.expiry_date > %4";
+      $params[4] = [date('Y-m-d'), 'String'];
+    }
 
     CRM_Core_DAO::executeQuery($query, $params);
   }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
@@ -4,6 +4,9 @@ use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
 use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 use CRM_HRLeaveAndAbsences_Exception_InvalidTOILRequestException as InvalidTOILRequestException;
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
+use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
 
 class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_TOILRequest {
 
@@ -26,7 +29,7 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
     }
 
     $instance = new self();
-    //set from_date_type and to_date_type to be full day by default
+    //set from_date_type and to_date_type to be full day bxy default
     $dateTypeOptions = array_flip(LeaveRequest::buildOptions('from_date_type'));
     $params['from_date_type'] = $dateTypeOptions['All Day'];
     if(!empty($params['to_date'])){
@@ -223,5 +226,44 @@ class CRM_HRLeaveAndAbsences_BAO_TOILRequest extends CRM_HRLeaveAndAbsences_DAO_
       $toilAmounts[$toilAmount['name']] = $toilAmount['value'];
     }
     return $toilAmounts;
+  }
+
+  /**
+   * Deletes the TOIL Requests associated with an Absence Type (within the given Absence Period)
+   * and all the LeaveRequests, LeaveBalanceChanges and LeaveRequestDates of the TOIL Requests.
+   *
+   * @param int $absenceTypeID
+   *   The absence Type that TOIL requests is to be deleted for.
+   * @param CRM_HRLeaveAndAbsences_BAO_AbsencePeriod $absencePeriod
+   * @param boolean $nonExpiredOnly
+   *   Whether to delete only records linked to non expired balance changes
+   */
+  public static function deleteAllForAbsenceType($absenceTypeID, AbsencePeriod $absencePeriod, $nonExpiredOnly = false) {
+    $leaveBalanceChangeTable = LeaveBalanceChange::getTableName();
+    $toilRequestTable = TOILRequest::getTableName();
+    $leaveRequestTable = LeaveRequest::getTableName();
+    $leaveRequestDateTable = LeaveRequestDate::getTableName();
+
+    $query = "DELETE tr, bc, lr, lrd FROM {$toilRequestTable} tr
+              INNER JOIN {$leaveBalanceChangeTable} bc ON bc.source_id = tr.id AND bc.source_type = %1
+              INNER JOIN {$leaveRequestTable} lr ON tr.leave_request_id = lr.id
+              INNER JOIN {$leaveRequestDateTable} lrd ON lr.id = lrd.leave_request_id
+              WHERE lr.from_date >= %2 AND lr.to_date <= %3
+              AND lr.type_id = %4";
+
+    if ($nonExpiredOnly) {
+      $query .= " AND bc.expiry_date >= %5";
+    }
+
+    $today = new DateTime('today');
+    $params = [
+      1 => [LeaveBalanceChange::SOURCE_TOIL_REQUEST, 'String'],
+      2 => [$absencePeriod->start_date, 'String'],
+      3 => [$absencePeriod->end_date, 'String'],
+      4 => [$absenceTypeID, 'Integer'],
+      5 => [$today->format('Y-m-d'), 'String'],
+    ];
+
+    CRM_Core_DAO::executeQuery($query, $params);
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/AbsenceType.php
@@ -1,0 +1,31 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
+use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+
+class CRM_HRLeaveAndAbsences_Service_AbsenceType {
+
+  /**
+   * Actions that occurs when an AbsenceType is updated
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_AbsenceType $absenceType
+   */
+  public function postUpdateActions(AbsenceType $absenceType) {
+    $oldAbsenceType = $absenceType->oldAbsenceType;
+    if ($oldAbsenceType->allow_accruals_request == true && $absenceType->allow_accruals_request == false) {
+      $this->onToilDisable($absenceType);
+    }
+  }
+
+  /**
+   * The set of actions/updates that occurs when TOIL is disabled for an AbsenceType.
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_AbsenceType $absenceType
+   */
+  private function onToilDisable(AbsenceType $absenceType) {
+    $currentPeriod = AbsencePeriod::getCurrentPeriod();
+    $startDate = new DateTime($currentPeriod->start_date);
+    TOILRequest::deleteAllForAbsenceType($absenceType->id, $startDate, null, true);
+  }
+}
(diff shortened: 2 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -452,4 +452,51 @@ public function testCreateTOILRequestBalanceChangeWhenATOILExpiryDateIsGiven() {
...
+
+    // decrease the max leave accrual
+    AbsenceTypeFabricator::fabricate([
+      'id' => $absenceType->id,
+      'max_leave_accrual' => 1,
+      'allow_accruals_request' => true,
+    ]);
+
+    // update the TOIL request status
+    TOILRequest::validateParams([
```
davialexandre: There is a problem with this test. If, for some strange reason, the exception is thrown by the first call to TOILRequest::validateParams() the test would still pass, even though the validation would be wrong.

To make sure it will pass only if the second call throws an exception, please remove the `@expectedException` and `@expectedExceptionMessage` from the docblock and add them like this:

```php
// first TOILRequest::validateParams() calls

$this->expectedException('CRM_HRLeaveAndAbsences_Exception_InvalidTOILRequestException', 'The TOIL amount requested for is greater than the maximum for this Absence Type');

// Second call to TOILRequest::validateParams()
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -452,4 +452,51 @@ public function testCreateTOILRequestBalanceChangeWhenATOILExpiryDateIsGiven() {
...
+    ]);
+
+    $toilRequest = TOILRequest::create([
+      'type_id' => $absenceType->id,
+      'contact_id' => 1,
+      'status_id' => 3,
+      'from_date' => date('YmdHis'),
+      'to_date' => date('YmdHis'),
+      'to_date_type' => 1,
+      'from_date_type' => 1,
```
davialexandre: TOIL don't have to_date_type and from_date_type (they're set automatically)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -452,4 +452,51 @@ public function testCreateTOILRequestBalanceChangeWhenATOILExpiryDateIsGiven() {
...
+      'toil_to_accrue' => $this->toilAmounts['3 Days']['value'],
+      'duration' => 300,
+    ], false);
+
+    // decrease the max leave accrual
+    AbsenceTypeFabricator::fabricate([
+      'id' => $absenceType->id,
+      'max_leave_accrual' => 1,
+      'allow_accruals_request' => true,
```
davialexandre: Instead of fabricate, I think it would be better to use the BAO's create method. The fabricate gives me the impression you're creating a new instance instead of updating an existing one (I know, create is not much better, but still... ¯\_(ツ)_/¯)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -452,4 +452,51 @@ public function testCreateTOILRequestBalanceChangeWhenATOILExpiryDateIsGiven() {
...
+    // update the TOIL request status
+    TOILRequest::validateParams([
+      'id' => $toilRequest->id,
+      'type_id' => $absenceType->id,
+      'contact_id' => 1,
+      'status_id' => 1,
+      'from_date' => date('YmdHis'),
+      'to_date' => date('YmdHis'),
+      'to_date_type' => 1,
+      'from_date_type' => 1,
```
davialexandre: No need for to_date_type and from_date_type here as well

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
```diff
@@ -26,7 +29,7 @@ public static function create($params, $validate = true) {
     }
 
     $instance = new self();
-    //set from_date_type and to_date_type to be full day by default
+    //set from_date_type and to_date_type to be full day bxy default
```
davialexandre: woops

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/AbsenceType.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+  }
+
+  /**
+   * The set of actions/updates that occurs when TOIL is disabled for an AbsenceType.
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_AbsenceType $absenceType
+   */
+  private static function onToilDisable(AbsenceType $absenceType) {
+    $currentPeriod = AbsencePeriod::getCurrentPeriod();
```
davialexandre: We must delete every TOILRequests in the future and that include requests for periods also in the future. Basically, we need to delete everything with expiry_date > today.

I would change deleteAllForAbsenceType to accept a $startDate and an $endDate instead of an Absence Period

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/AbsenceType.php
```diff
@@ -0,0 +1,30 @@
...
+use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+
+class CRM_HRLeaveAndAbsences_Service_AbsenceType {
+
+  /**
+   * Actions that occurs when an AbsenceType is updated
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_AbsenceType $absenceType
+   */
+  public static function postUpdateActions(AbsenceType $absenceType) {
```
davialexandre: Can you please make the methods on this service non static? The reasons is that, in the future, if we need to inject this service somewhere, we would be able to mock the methods

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -681,4 +687,80 @@ public function testCalculateToilExpiryDateWhenAbsenceTypeAllowsAccrualsRequestA
...
   }
+
+  public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsenceType() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('-1 days'),
+      'end_date' => CRM_Utils_Date::processDate('+ 300days'),
+    ]);
+
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'title' => 'Title 1',
```
davialexandre: no need for title and is_active here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -681,4 +687,80 @@ public function testCalculateToilExpiryDateWhenAbsenceTypeAllowsAccrualsRequestA
...
+      'is_active' => 1,
+    ]);
+
+    TOILRequestFabricator::fabricateWithoutValidation([
+      'type_id' => $absenceType->id,
+      'contact_id' => 1,
+      'status_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('+3 days'),
+      'to_date' => CRM_Utils_Date::processDate('+3 days'),
+      'to_date_type' => 1,
```
davialexandre: no need for date types here and on all the other calls to TOILRequestFabricator and TOILRequest::validateParams

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -681,4 +687,80 @@ public function testCalculateToilExpiryDateWhenAbsenceTypeAllowsAccrualsRequestA
...
+      'from_date_type' => 1,
+      'toil_to_accrue' => 2,
+      'duration' => 120,
+      'expiry_date' => CRM_Utils_Date::processDate('2016-12-10')
+    ]);
+
+    //disable TOIL
+    AbsenceTypeFabricator::fabricate([
+      'id' => $absenceType->id,
+      'title' => 'Title 1',
```
davialexandre: only id and allow_accruals_request are enough here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -681,4 +687,80 @@ public function testCalculateToilExpiryDateWhenAbsenceTypeAllowsAccrualsRequestA
     $expiry = $absenceType2->calculateToilExpiryDate($date);
     $this->assertEquals('2017-09-10', $expiry->format('Y-m-d'));
   }
+
+  public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsenceType() {
```
davialexandre: I think this method name is a bit confusing. It says it tests that "Non expired TOIL Request are deleted", but the assertions are all about asserting that expired things were not deleted. Maybe you should change the test name or the assertions?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Class CRM_HRLeaveAndAbsences_Service_LeaveRequestTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Service_AbsenceTypeTest extends BaseHeadlessTest {
+
+  public function testPostUpdateActionsWhenToilIsDisabledForAbsenceType() {
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'title' => 'Title 1',
```
davialexandre: unnecessary fields here too

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+    //disable TOIL
+    $updatedAbsenceType = AbsenceTypeFabricator::fabricate([
+      'id' => $absenceType->id,
+      'title' => 'Title 1',
+      'allow_accruals_request' => false,
+      'is_active' => 1,
+    ]);
+
+    $updatedAbsenceType->oldAbsenceType = $absenceType;
```
davialexandre: isn't this duplicating the logic on the AbsenceType BAO?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
```diff
@@ -0,0 +1,97 @@
...
+use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+
+/**
+ * Class CRM_HRLeaveAndAbsences_Service_LeaveRequestTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Service_AbsenceTypeTest extends BaseHeadlessTest {
+
+  public function testPostUpdateActionsWhenToilIsDisabledForAbsenceType() {
```
davialexandre: can we have a more descriptive name here, like "testePostUpdateActionDeletesTOILRequestsWhenTOILGetsDisabled"?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
tunbola: Actually, the first call was to the TOILRequest::create method with validate set to false. 
So the call to ValidateParams is the only call for the validate method.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
tunbola: Yes, but this property need to be set in order to be able to test the `postUpdateActions` directly

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
```diff
@@ -248,22 +251,27 @@ public static function deleteAllForAbsenceType($absenceTypeID, AbsencePeriod $ab
...
+      4 => [$today->format('Y-m-d'), 'String'],
     ];
 
+    if ($endDate) {
+      $query .= " AND lr.to_date <= %5";
+      $params[5] = [$endDate->format('Y-m-d'), 'String'];
+    }
+
+    if ($nonExpiredOnly) {
+      $query .= " AND bc.expiry_date > %4";
```
davialexandre: You should only add param 4 here and not when declaring the $params array.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -718,21 +714,18 @@ public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsence
       'type_id' => $absenceType->id,
       'contact_id' => 1,
       'status_id' => 1,
-      'from_date' => CRM_Utils_Date::processDate('2016-03-10'),
-      'to_date' => CRM_Utils_Date::processDate('2016-03-10'),
-      'to_date_type' => 1,
-      'from_date_type' => 1,
+      'from_date' => CRM_Utils_Date::processDate('+1 days'),
```
davialexandre: please use '+1 day' (singular) for both dates

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -718,21 +714,18 @@ public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsence
       'type_id' => $absenceType->id,
       'contact_id' => 1,
       'status_id' => 1,
```
davialexandre: Oh, I just realized the TOILRequestFabricator also has a default value for the status_id, so you can remove it as well

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -718,21 +714,18 @@ public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsence
...
     ]);
 
     //disable TOIL
-    AbsenceTypeFabricator::fabricate([
+    AbsenceType::create([
       'id' => $absenceType->id,
-      'title' => 'Title 1',
       'allow_accruals_request' => false,
-      'is_active' => 1,
+      'color' => '#000000'
```
davialexandre: Do you really need the color?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -754,7 +747,8 @@ public function testNonExpiredToilRequestsAreDeletedWhenToilIsDisabledForAbsence
     $leaveRequestDate->find();
     $this->assertEquals($leaveRequestDate->N, 1);
     $leaveRequestDate->fetch();
-    $this->assertEquals($leaveRequestDate->date, '2016-03-10');
+    $date = date('Y-m-d', strtotime('+1 days'));
```
davialexandre: Whenever using 1, please always use day intestead of days

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/TOILRequestTest.php
```diff
@@ -676,4 +652,107 @@ public function testDeleteForAbsenceTypeWithinAGivenPeriodWhenNonExpiredOnlyFlag
...
+      'status_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-04-10'),
+      'to_date' => CRM_Utils_Date::processDate('2016-04-10'),
+      'toil_to_accrue' => 2,
+      'duration' => 120,
+      'expiry_date' => CRM_Utils_Date::processDate('2016-12-10')
+    ]);
+
+    $startDate = new DateTime('2016-03-10');
+    TOILRequest::deleteAllForAbsenceType($absenceType, $startDate, null, false);
```
davialexandre: This comment applies to all the "deleteSomething" method tests:

If you don't assert that the records we want to delete actually exist before the deletion, there's no way to know you actually deleted something. For example, if you simply remove the 2 calls to the TOILRequestFabricator, this test will still pass, even though nothing was actually deleted.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
tunbola: Yeah, got the error `undefined index 'color'` when this was not added.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
davialexandre: Right, but internally, won't the AbsenceType Fabricator call the AbsenceType::create, which will call the `loadOldAbsenceType` method and set the `oldAbsenceType` property?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/TOILRequest.php
```diff
@@ -270,6 +268,8 @@ public static function deleteAllForAbsenceType($absenceTypeID, DateTime $startDa
 
     if ($nonExpiredOnly) {
       $query .= " AND bc.expiry_date > %4";
+      $today = new DateTime('today');
+      $params[4] = [$today->format('Y-m-d'), 'String'];
```
davialexandre: instead of using a datetime object, what about `date('Y-m-d')`?

PR #1347, opened 2017-01-12
PCHR-1681: Save LeaveRequest balance changes as a negative number

davialexandre:
LeaveRequest always deduct days from the entitlement, so their balance changes should always have a negative number as the amount.

The associated tests were updated to reflected this. They were also updated in order to add some more "randomness" (they were all using the same dates and balance changes, so I picked some different dates in order to generate some differente balance changes in the end)

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/SicknessRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/SicknessRequestTest.php
@@ -38,14 +38,14 @@ class CRM_HRLeaveAndAbsences_Service_SicknessRequestTest extends BaseHeadlessTes
 
     $sicknessReasons = array_flip(SicknessRequest::buildOptions('reason', 'validate'));
 
-    // a 7 days request, from friday to thursday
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/SicknessRequestTest.php
```diff
@@ -38,29 +38,29 @@ public function testCreateAlsoCreateTheLeaveRequestBalanceChanges() {
...
-    // Since the 40 hours work pattern was used, and it this is a week long
-    // leave request, the balance will be 5 (for the 5 working days)
-    $this->assertEquals(5, $balance);
+    // Since the 40 hours work pattern was used and there are 7 working days during
+    // the 9 days period of the requests, the balance will be -7
+    $this->assertEquals(-7, $balance);
 
     $balanceChanges = LeaveBalanceChange::getBreakdownForLeaveRequest($leaveRequest);
-    // Even though the balance is 5, we must have 7 balance changes, one for
+    // Even though the balance is -5, we must have 7 balance changes, one for
```
tunbola: This should be "Even though the balance is -7, we must have 9 balance changes, one for"

PR #1348, opened 2017-01-12
PCHR-1814: Implement manager response feature

dhavalsharma:
This PR contains update to controller `LeaveRequestPopupCtrl`. 

When you login as manager you will be able to respond to staff leave request by changing the status as shown in gif below 
1814_manager_login

As part of implementation the `update` API is used for `LeaveRequest` end point.

## Implementation 

The implementation is for controller `LeaveRequestPopupCtrl` and directive `leaveRequestPopup`.

### Create Mode 
_leave request dialog opened by staff/manager to create_

Before the directive was taking in only contactId as additional attribute, so a sample usage will look like 
```html
<button type="button" class="btn btn-primary" leave-request-popup contact-id="myleave.contactId">Record New Absence</button>
```
where directive opens the leave request dialog in create mode with contact-id of the user.

### Modify mode 
_leave request dialog opened by manager of the staff to modify_

Now its extended to take existing leave request as additional attribute. 
Sample usage for manager looks like 
```html
<li><a href='#' leave-request-popup leave-request='request' contact-id="ctrl.contactId">Respond</a></li>
```
When used with leave request attribute the contactId attribute belongs to the manager. The leave request object will contain the contact id of the user.

## Initialization 

The initialization logic is modified to wait for all values to be retrieved before allowing user to perform action. A series of init* functions are added to manage the initialization when existing leave request is modified.

## Improvements

1.  changed isValid from sendGet to sendPOST for LeaveRequest API. 
2.  updates test cases for isManaged by from sendGet to sendPOST  for LeaveRequest API.
3.  modified `uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php` to include `'ismanagedby' => ['leave_request']` and `'isvalid' => ['leave_request']`
4. Added back contact with id 202 and 203 in contact mock js file


AkA84:
> When used with leave request attribute the contactId attribute belongs the the manager

@dhavalsharma i'm not sure i understand this, can you elaborate more?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -195,6 +214,11 @@ define([
         }
 
         vm.error = undefined;
+        //update leaverequest
+        if (updateRequest()) {
```
AkA84: This needs to be refactored, it's not good currently for several reason

The name of the function, combined with the fact that it is invoked in an `if` statement makes this code hard to understand. What does it mean _"if updateRequest then return"_? Is the function simply returning a boolean? Is it checking for something? Is it also doing an update? Why is it returning a boolean at all, given the name? You have to go and read the function implementation to find out

Why are we returning at all, in the middle of `vm.submit`, basically having a major branching between updating and creating a leave request? Can't this be done in a more elegant way? A leave request update and a leave request create are basically the same thing:
1. you send the current leave request status to the server
2. when the server is done, you emit an event
3. you cancel any `error`
4. you call `vm.ok()`
5. If any error happens during the process, you `catch` it and store the error message in `vm.error`

Given the above, is there any need for the branching at all? This surely can be improved

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -250,25 +274,45 @@ define([
...
-        vm.leaveRequest = LeaveRequestInstance.init({
-          contact_id: contactId //resolved from directive
-        }, false);
+      /**
+       * initializes the controller on loading the dialog
+       */
+      (function initController() {
+        if (directiveOptions.leaveRequest) {
+          //manager is responding to leave request
+          vm.uiOptions.isManager = directiveOptions.leaveRequest.contact_id != directiveOptions.contactId;
```
AkA84: But you are assigning the value to `isManager` later, when you call `isManagedBy()`, so why are you are doing it again here?

AkA84: Also, this assignment is not correct: the fact that the current user's id is different than the leave request's contact's id *does not* mean that the current user is the leave request manager. That's why we have `LeaveRequest.roleOf()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -250,25 +274,45 @@ define([
...
+          //owner is editing popup
+          vm.uiOptions.isOwner = directiveOptions.leaveRequest.contact_id == directiveOptions.contactId;
```
AkA84: Why are you not encapsulating this "check what role the current user has" logic in one single function? You have  it here, and then in `isManagedBy()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -250,25 +274,45 @@ define([
...
+        }
+
+        if (canEdit()) {
```
AkA84: Why can't you clone the leave request anyway, if it has been passed? Why do you need to check first if the user can edit? It seems unnecessary logic to me. Even if we were in "view" mode, cloning the leave request would not make any difference, and you would streamline the code logic

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -250,25 +274,45 @@ define([
...
+          //get a clone
+          var clone = _.cloneDeep(directiveOptions.leaveRequest);
+          vm.leaveRequest = LeaveRequestInstance.init(clone, true);
```
AkA84: 1. There is no need to re-init the leave request, you are already cloning an initialized leave request instance
2. the second parameter for `init()` is for when you are initializing the model instance with data coming straight from the API, which is not something that is happening here

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -250,25 +274,45 @@ define([
...
+        } else {
+          // Create an empty leave request
+          vm.leaveRequest = LeaveRequestInstance.init({
```
AkA84: Please encapsulate this whole "leave request initialization" in a separate function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -340,20 +384,14 @@ define([
...
+              }, true) // `true` because we want to use the 'future' balance for calculation
               .then(function (entitlements) {
                 // create a list of absence types with a `balance` property
                 vm.absenceTypes = filterAbsenceTypes(absenceTypes, entitlements);
-
-                // Assign the first absence type to the leave request
-                vm.selectedAbsenceType = vm.absenceTypes[0];
-                vm.leaveRequest.type_id = vm.selectedAbsenceType.id;
-                // Init the `balance` object based on the first absence type
-                vm.balance.opening = vm.selectedAbsenceType.remainder;
```
AkA84: Why did you remove all of this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -386,6 +424,10 @@ define([
        * @returns {Promise}
        */
       function initDayTypesAndStatus() {
+        if (vm.leaveRequestDayTypes.length && vm.leaveRequestStatuses.length) {
```
AkA84: Why are you doing this? This function seem to be called only once. Also, isn't the api request cached anyway?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -402,7 +444,6 @@ define([
             return OptionGroup.valuesOf('hrleaveandabsences_leave_request_status')
               .then(function (optionValues) {
                 vm.leaveRequestStatuses = optionValues;
-                vm.leaveRequest.status_id = valueOfRequestStatus('waiting_approval');
```
AkA84: Why did you remove this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -433,6 +474,34 @@ define([
...
 
       /**
+       * Pick a specific leave request day type
+       *
+       * @param {string} value - The leave request day type value to match
+       * @return {object}
+       */
+      function getStatusFromValue(value) {
+        var collection = vm.leaveRequestStatuses,
+          key = 'value';
```
AkA84: This is not readable, please change it
```js
function getStatusFromValue(value) {
  var collection = vm.leaveRequestStatuses, key = 'value';
  
  return _.find(collection, function (collectionItem) {
    return collectionItem[key] === value;
  });
}
```

AkA84: And actually there is no point defining `key` at all, just do 
```js
return collectionItem['value'] === value;
```

AkA84: There is no point defining `collection` either, the function body can be much more terse
```js
function getStatusFromValue(value) {  
  return _.find(vm.leaveRequestStatuses, function (status) {
    return status['value'] === value;
  });
}
```
Also please make sure to give meaningful names to the arguments of the callback (`status`, not a generic `collectionItem`)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -433,6 +474,34 @@ define([
       }
 
       /**
+       * Pick a specific leave request day type
+       *
+       * @param {string} value - The leave request day type value to match
+       * @return {object}
+       */
+      function getStatusFromValue(value) {
```
AkA84: I'm getting confused by this function

1. The documentation says "Pick a specific leave request day type", but you are using `vm.leaveRequestStatuses`?
2. This function seems extremely similar to `valueOfRequestStatus()`. Can you tell me what the difference is?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        }
+      }
+
+      /**
+       * sets leave requestion statuses for manager
+       *
+       */
+      function setStatusesForManager() {
+        var allowed_statuses = ['approved', 'more_information_requested', 'cancelled'];
+        if (vm.uiOptions.isManager) {
```
AkA84: You are already checking that `isManager` is true before you are calling this function, plus the name of the function contains `ForManager`. You can skip this check

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        // Init the `balance` object based on the first absence type
+        vm.balance.opening = vm.selectedAbsenceType.remainder;
+      }
+
+      /**
+       * initialize from and to dates, day types and contact
+       *
+       */
+      function initDates() {
+        initStatus();
```
AkA84: Why are you initializing the status(es?) in `initDates()`? It doesn't seem there is any dependency between them

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        initDayTypes();
+        initContact();
```
AkA84: Why are you initializing the contact in `initDates()`? It doesn't seem there is any dependency between them

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -433,6 +474,34 @@ define([
...
+        });
+      }
+
+      /**
+       * Pick a specific leave request day type
+       *
+       * @param {string} value - The leave request day type value to match
+       * @return {object}
+       */
+      function getDateTypeFromValue(value) {
```
AkA84: Is this a clone of `getStatusFromValue`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -64,13 +65,17 @@ define([
...
+        isManager: false, //when the dialog is opened by manager or admin
+        isOwner: false, //when the dialog is opened by the owner
         multipleDays: true,
+        contact: '',
         showDatePickerFrom: false,
         showDatePickerTo: false,
         userDateFormat: HR_settings.DATE_FORMAT,
+        selectedStatus: undefined,
         showBalance: false,
+        statusLabel: '',
```
AkA84: Why is this property even needed at all? You have access to the `leaveRequest.status_id` property, so in the view you can simply call one of those functions that you have in the component, that get you the OptionValue from the status_id. Something like
```html
<span class="col-xs-6">{{::$ctrl.getStatus($ctrl.leaveRequest.status_id).label}}</span>
```
This way you wouldn't have to deal with this additional property, and all the logic that it comes attached to

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+       */
+      function initStatus() {
+        if (canEdit()) {
+          vm.uiOptions.statusLabel = getStatusFromValue(vm.leaveRequest.status_id).label;
+          //waiting_approval is removed below so call the above before it
+          if (vm.uiOptions.isManager) {
+            setStatusesForManager();
+          }
+        } else {
+          vm.leaveRequest.status_id = valueOfRequestStatus('waiting_approval');
```
AkA84: This is something that should happen in the "initialize leave request" function that i mentioned earlier

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -443,6 +512,16 @@ define([
       }
 
       /**
+       * Converts given date to javascript date as expected by uib-datepicker
+       *
+       * @param {Date} date
+       * @return {Date} Javascript date
```
AkA84: The `@param` and `@return` types can't be right? From the documentation it seems that you are passing a `Date` object and getting a `Date` object as the output, so what's the point of the function?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
       /**
        * checks if all params are set to calculate balance
        *
-       * @param {Array} listOfDayTypes collection of available day types
+       * @param {Boolean} true if all present else false
        */
       function canCalculateChange() {
         return vm.leaveRequest.from_date && vm.leaveRequest.to_date &&
```
AkA84: Aren't you doing the exact same check in `canSubmit`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
 
+      /**
+       * checks if leaverequest is managed by given manager id
+       *
+       * @param {String} managerContactId
+       * @returns {Promise}
+       */
+      function isManagedBy(managerContactId) {
+        return vm.leaveRequest.roleOf({
+            id: managerContactId
```
AkA84: 4-spaces indentation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
       function canCalculateChange() {
         return vm.leaveRequest.from_date && vm.leaveRequest.to_date &&
           vm.leaveRequest.from_date_type && vm.leaveRequest.to_date_type;
       }
 
+      /**
+       * checks if leaverequest is managed by given manager id
+       *
+       * @param {String} managerContactId
+       * @returns {Promise}
```
AkA84: `@return`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+       */
+      function isManagedBy(managerContactId) {
```
AkA84: an `isSomething` function is expected to return a boolean value, but here instead it ends up with setting the value of `isManager`. So either you change the name of the function, or you change the implementation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        return vm.leaveRequest.roleOf({
+            id: managerContactId
+          })
+          .then(function (role) {
+            if (role === 'manager') {
+              vm.uiOptions.isManager = true;
+            } else {
+              vm.uiOptions.isManager = false;
```
AkA84:
```js
vm.uiOptions.isManager = role === 'manager';
```
please use terser code whenever possible

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+            }
+          });
+      }
+
+      /**
+       * initialize absence types
+       *
```
AkA84: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+       */
+      function initAbsenceType() {
+        if (canEdit()) {
```
AkA84: I really feel that this `canEdit()` is misused. The check that you should be doing is "are we in edit mode?" which is based on whether the leave request has an id or not.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+          vm.selectedAbsenceType = _.find(vm.absenceTypes, function (absenceType) {
+            return absenceType.id == vm.leaveRequest.type_id;
+          });
+        } else {
+          // Assign the first absence type to the leave request
+          vm.selectedAbsenceType = vm.absenceTypes[0];
+        }
+
+        vm.leaveRequest.type_id = vm.selectedAbsenceType.id;
```
AkA84: Shouldn't this happen only in the `else` branch? If the leave request already exists, why are you reassigning the `type_id`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+      function initDates() {
+        initStatus();
+        initDayTypes();
+        initContact();
+
+        if (canEdit()) {
+          vm.uiOptions.fromDate = convertDateFormatFromServer(vm.leaveRequest.from_date);
+          vm.onDateChange(vm.uiOptions.fromDate, 'from');
+          vm.uiOptions.toDate = convertDateFormatFromServer(vm.leaveRequest.to_date);
+          vm.onDateChange(vm.uiOptions.toDate, 'to');
```
AkA84: As always, please group the lines to make the code more readable
```js
vm.uiOptions.fromDate = convertDateFormatFromServer(vm.leaveRequest.from_date);
vm.uiOptions.toDate = convertDateFormatFromServer(vm.leaveRequest.to_date);

vm.onDateChange(vm.uiOptions.fromDate, 'from');
vm.onDateChange(vm.uiOptions.toDate, 'to');
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        }
+      }
+
+      /**
+       * initialize day types
+       *
```
AkA84: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        if (canEdit()) {
+          vm.uiOptions.fromDate = convertDateFormatFromServer(vm.leaveRequest.from_date);
+          vm.onDateChange(vm.uiOptions.fromDate, 'from');
+          vm.uiOptions.toDate = convertDateFormatFromServer(vm.leaveRequest.to_date);
+          vm.onDateChange(vm.uiOptions.toDate, 'to');
+        }
+      }
+
+      /**
+       * initialize day types
```
AkA84: Can you make sure to start every function documentation with a capital letter?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+       *
+       */
+      function initDayTypes() {
+        if (canEdit()) {
+          vm.uiOptions.selectedFromType = getDateTypeFromValue(vm.leaveRequest.from_date_type);
```
AkA84: I don't see the reason why you need `selectedFromType` and `selectedToType` to begin with. Why can't you plug `leaveRequest.from_day_type` and `leaveRequest.to_day_type` directly into the view? It would result in much less code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        }
+      }
+
+      /**
+       * initialize status
+       *
+       */
+      function initStatus() {
+        if (canEdit()) {
+          vm.uiOptions.statusLabel = getStatusFromValue(vm.leaveRequest.status_id).label;
```
AkA84: See what i wrote already about `statusLabel`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+          //waiting_approval is removed below so call the above before it
```
AkA84: This comment is a bit confusing, please remove it, the code is already self-explanatory as is

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        }
+      }
+
+      /**
+       * initialize contact
+       *
+       * {Promise}
+       */
+      function initContact() {
+        if (canEdit()) {
```
AkA84: This is wrong, only the manager should be able to see the contact's name

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: Leave an empty line after the `var` declarations

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+          return $q.resolve({});
+        }
+      }
+
+      /**
+       * sets leave requestion statuses for manager
+       *
+       */
+      function setStatusesForManager() {
+        var allowed_statuses = ['approved', 'more_information_requested', 'cancelled'];
```
AkA84: use `camelCase` in javascript, not `snake_case`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+        if (vm.uiOptions.isManager) {
+          //remove current status of leaverequest
+          _.remove(allowed_statuses, function (status) {
+            return status == getStatusFromValue(vm.leaveRequest.status_id).name;
+          });
+
+          vm.leaveRequestStatuses = vm.leaveRequestStatuses.filter(function (status) {
```
AkA84: What about
```js
vm.leaveRequestStatuses = vm.leaveRequestStatuses
  .filter(function (status) {
    return status == getStatusFromValue(vm.leaveRequest.status_id).name;
  })
  .filter(function (status) {
    return _.includes(allowed_statuses, status.name);
  });
```
would this work?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -545,13 +625,166 @@ define([
...
+      /**
+       * sets leave requestion statuses for manager
+       *
+       */
+      function setStatusesForManager() {
+        var allowed_statuses = ['approved', 'more_information_requested', 'cancelled'];
+        if (vm.uiOptions.isManager) {
+          //remove current status of leaverequest
+          _.remove(allowed_statuses, function (status) {
+            return status == getStatusFromValue(vm.leaveRequest.status_id).name;
```
AkA84: This should happen only if you are in "edit" mode

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/directives/leave-request-popup.js
```diff
@@ -22,8 +23,11 @@ define([
               controller: 'LeaveRequestPopupCtrl',
               controllerAs: '$ctrl',
               resolve: {
-                contactId: function () {
-                  return scope.contactId
+                directiveOptions: function () {
+                  return {
+                    contactId: scope.contactId,
+                    leaveRequest: scope.leaveRequest
+                  }
```
AkA84: missing `;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -175,7 +175,7 @@ define([
             deferred.resolve('owner');
           } else {
             LeaveRequestAPI.isManagedBy(this.id, contact.id)
-              .then(function (response) {
+              .then(function (response) {
```
AkA84: White space?

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
```diff
@@ -234,6 +234,142 @@ define([
...
               email: '',
               contact_type: 'Individual'
+            },
+            {
+              "contact_id": "202",
+              "contact_type": "Individual",
+              "contact_sub_type": "",
+              "sort_name": "civihr_staff@compucorp.co.uk",
+              "display_name": "civihr_staff@compucorp.co.uk",
+              "do_not_email": "0",
```
dhavalsharma: this was as before, not sure about the dependency elsewhere for rest of fields

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-requests.js
```diff
@@ -175,9 +176,9 @@ define([
      *
      * @param {string} page - page number of the pagination element
      */
-    vm.refresh = function (page) {
+    vm.refresh = function (page, cache) {
```
dhavalsharma: @deb1990 ps respond

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -64,13 +65,17 @@ define([
         }
       };
       vm.uiOptions = {
-        isAdmin: false,
         isChangeExpanded: false,
+        isManager: false, //when the dialog is opened by manager or admin
+        isOwner: false, //when the dialog is opened by the owner
```
dhavalsharma: I like the idea but here I think number of comparisons will still remain same, moreever a boolean is better and smaller comparison than string

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: its a one time call needed only in init like this `vm.uiOptions.statusLabel = getStatusFromValue(vm.leaveRequest.status_id).label;`

getStatusFromValue is a private function so either I will have to make that public or keep a public attribute, my preference would be for a simple attribute on UI rather than complex code there

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -146,6 +151,14 @@ define([
       };
 
       /**
+       * Updates leave request to currently selected status
+       *
+       */
+      vm.onStatusChanged = function () {
+        vm.leaveRequest.status_id = vm.uiOptions.selectedStatus.value;
```
dhavalsharma: tried it but its not coming up on initialization, shows empty value in dropdown

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -178,13 +191,19 @@ define([
        * @returns {Boolean}
        */
       vm.canSubmit = function () {
-        return vm.leaveRequest.from_date && vm.leaveRequest.to_date &&
+        var canSubmit = vm.leaveRequest.from_date && vm.leaveRequest.to_date &&
           vm.leaveRequest.to_date_type && vm.leaveRequest.from_date_type;
+
+        if (canEdit()) {
+          canSubmit = canSubmit && vm.uiOptions.selectedStatus;
```
dhavalsharma: changed logic to 
```javascript
        //check if manager has changed status
        if (vm.uiOptions.isManager) {
          canSubmit = canSubmit && vm.uiOptions.selectedStatus;
        }
```
its not exactly edit mode, its to check if manager can update or not.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: I think you have answered yourself in the 2 comments. 

1.  First `isManager` is set based on the id's coming from directive 
2. and then it can be reset in `isManagedBy` call if given manager is not of staff 

To call `roleOf` we need the leaverequest so first step essentially creates it.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: init to get methods like roleOf again on leaverequest instance as cloning removes them and the call is on object from server so setting flag to true

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: ps see `function initAbsenceType()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: future optimization may need to call the function again and again from say `onDateChange` and then this check will be useful

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: ps refer to `function initStatus()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: 1.  documentation changed
2.  The search key is different in both

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: But what you have here is a property that we could do without, since the all the data that you need to correctly show the status name is already on the `leaveRequest` object itself.

If you don't want to have a "utility" method on the scope (which i agree with), then you could simply have the statuses on the scope accessible by value (see [here](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-report.js#L230-L235)) for example, which mean you could simply do something like
```html
<span class="col-xs-6">{{::$ctrl.statuses[$ctrl.leaveRequest.status_id].label}}</span>
```
as a result you don't have neither unnecessary property, nor utility method on the scope

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: You know why?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: yes but they are on different collection

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: I don't find this clear

> 1. First isManager is set based on the id's coming from directive

But it's set based on *wrong* logic

> 2. and then it can be reset in isManagedBy call if given manager is not of staff

Why does it need to be reset at all? Why can't you simply set `isManager` and `isOwner` in one go, in a single function?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: > init to get methods like roleOf again on leaverequest instance as cloning removes them 

Why is cloning removing those methods?

> the call is on object from server so setting flag to true

Why is it an object from the server? Aren't you passing a `LeaveRequestInstance` to the controller?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
AkA84: is the request cached? If it is, then there's no need for the check

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: `Converts given date to javascript date as expected by uib-datepicker` is the purpose of the function so its taking a date in server format and reformatting it to a javascript date as uibdatepicker does not accepts moment date

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: `vm.statusLabel` is required to be set before `vm.leaveRequestStatuses` gets filtered for manager and removes `waiting_approval` or whatever is the current status of leaverequest, using the indexBy idea in controller

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: logic changed as its now uses indexBy

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -14,22 +15,25 @@
...
     'leave-absences/shared/controllers/leave-request-popup-controller',
-  ], function (_, moment, angular, optionGroupMock) {
+  ], function (_, moment, angular, optionGroupMock, mockData) {
     'use strict';
 
     describe('LeaveRequestPopupCtrl', function () {
       var $log, $rootScope, $ctrl, modalInstanceSpy, $scope, $q, $controller,
-        $provide, DateFormat, LeaveRequestInstance, serverDateFormat = 'YYYY-MM-DD';;
+        $provide, DateFormat, LeaveRequestInstance, Contact, ContactAPIMock,
+        serverDateFormat = 'YYYY-MM-DD';;
```
deb1990: remove one semicolon

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -40,29 +44,43 @@
         $provide.value('PublicHolidayAPI', _PublicHolidayAPIMock_);
       }));
 
+      beforeEach(inject(['api.contact.mock', function (_ContactAPIMock_) {
+        //ContactAPIMock = _ContactAPIMock_;
```
deb1990: Remove commented line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -282,13 +293,16 @@
 
       describe('day types', function () {
         describe('on change selection', function () {
+          var expectedDayType;
+
           beforeEach(function () {
+            expectedDayType = optionGroupMock.specificValue('hrleaveandabsences_leave_request_day_type', 'name', 'all_day')
```
deb1990: semicolon missing

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -458,20 +482,92 @@
         });
       });
 
+      describe('when manager opens leave request popup', function () {
+        beforeEach(function () {
+          //waiting approval request at index 3 with value 3
+          var leaveRequest = LeaveRequestInstance.init(mockData.all().values[3]);
+          leaveRequest.contact_id = '' + CRM.vars.leaveAndAbsences.contactId;
```
deb1990: `CRM.vars.leaveAndAbsences.contactId.toString()` is more descriptive

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/leave-request-popup-controller_test.js
```diff
@@ -458,20 +482,92 @@
         });
       });
 
+      describe('when manager opens leave request popup', function () {
+        beforeEach(function () {
+          //waiting approval request at index 3 with value 3
+          var leaveRequest = LeaveRequestInstance.init(mockData.all().values[3]);
```
deb1990: If we change order of mockData test case needs to be rewritten. instead write a find function to get the intended object.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -4,210 +4,206 @@
...
-                      <input type="text"
-                          class="form-control"
-                          uib-datepicker-popup
-                          show-button-bar="false"
-                          datepicker-options="$ctrl.uiOptions.datePickerOptions"
-                          ng-model="$ctrl.uiOptions.fromDate"
-                          is-open="$ctrl.uiOptions.showDatePickerFrom"
-                          ng-change="$ctrl.onDateChange($ctrl.uiOptions.fromDate, 'from')" />
-                      <span class="input-group-btn">
+                    <div class="col-xs-4">
```
deb1990: Indentation is 4 space it seems for all below changes

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -4,210 +4,206 @@
...
+            </tfoot>
+            </table>
+            <div class="row chr_leave-request-modal__form-group"
+                ng-class="{'text-danger': $ctrl.balance.closing < 0}">
+              <div class="col-xs-4">Closing Balance</div>
+              <div class="col-xs-8">{{$ctrl.balance.closing}}</div>
       </div>
     </div>
-    <div ng-show="$ctrl.uiOptions.isAdmin"
+    </crm-loading>
```
deb1990: Indentation missing 
```
 </div>
 -    <div ng-show="$ctrl.uiOptions.isAdmin"		
 +    </crm-loading>
 +    </uib-tab>
 +    <uib-tab heading="Comments">
 +    </uib-tab>
 +    <uib-tab heading="Files">
 +    </uib-tab>
 +    </uib-tabset>
 +    </div>
 +    </div>
```

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -4,210 +4,206 @@
...
+                <td>{{breakdown.date | date:'EEE '+ $ctrl.uiOptions.userDateFormat}}</td>
+                <td>{{breakdown.type.label}}</td>
+                <td>{{breakdown.amount}}</td>
+              </tr>
+            </tbody>
+            <tfoot>
+              <tr>
+                <td colspan="3"
+                    class="text-center">
+                  <uib-pagination data-ng-if="$ctrl.pagination.totalItems"
```
deb1990: can we use `ng-if` to keep consistency? As we are not using data prefix anywhere

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -4,210 +4,206 @@
...
-                            ng-change="$ctrl.calculateBalanceChange()">
-                      <option value="">Select</option>
-                    </select>
-                      </div>
-                    </crm-loading>
+                    <div class="col-xs-4">
+                      <crm-loading show="!$ctrl.loading.toDayTypes">
+                        <div class="chr_custom-select chr_custom-select--full">
+                          <select name="toDayTypeSelect"
+                              ng-disabled="$ctrl.leaveRequestToDayTypes.length == 1"
```
deb1990: `$ctrl.leaveRequestToDayTypes.length === 1`
=== should be used for Integer comparison

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -4,210 +4,206 @@
...
-                            ng-model="$ctrl.uiOptions.selectedFromType"
-                            ng-change="$ctrl.calculateBalanceChange()">
-                      <option value="">Select</option>
-                    </select>
-                    </crm-loading>
+                    <div class="col-xs-4">
+                      <crm-loading show="!$ctrl.loading.fromDayTypes">
+                        <div class="chr_custom-select chr_custom-select--full">
+                          <select name="fromDayTypeSelect"
+                              ng-disabled="$ctrl.leaveRequestFromDayTypes.length == 1"
```
deb1990: `$ctrl.leaveRequestToDayTypes.length === 1`
=== should be used for Integer comparison

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -4,34 +4,40 @@ define([
...
 
-      var vm = {},
-        serverDateFormat = 'YYYY-MM-DD';
+      var selectedAbsenceType = {},
+        serverDateFormat = 'YYYY-MM-DD',
+        vm = {};
 
       vm.absenceTypes = [];
       vm.calendar = {};
+      vm.contact = {};
```
deb1990: can we set `vm.error = null;`? setting to `undefined` does not look good.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -163,69 +166,50 @@ define([
...
-          .catch(function (errors) {
-            // show errors
-            if (errors.error_message)
-              vm.error = errors.error_message;
-            else {
-              vm.error = errors;
-            }
-          });
+        //update leaverequest
+        if (canEdit()) {
```
deb1990: Use ternary operator

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -294,16 +289,16 @@ define([
...
+        date = convertDateFormatToServer(date);
         PublicHoliday.isPublicHoliday(date)
           .then(function (result) {
             if (result) {
               listToReturn = listToReturn.filter(function (publicHoliday) {
                 return publicHoliday.name === 'public_holiday';
               });
             } else {
-              var inCalendarList = getDayTypesFromDate(date);
+              var inCalendarList = getDayTypesFromDate(date, listToReturn);
```
deb1990: Define all the variables in the first line of the function which is more optimized.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -323,12 +318,70 @@ define([
...
+        return $q.resolve(vm.mode);
+      }
+
+      /**
+       * Initialize user's role
+       *
+       * @return {Promise}
+       */
+      function initUserRole() {
+        if (directiveOptions.leaveRequest &&
```
deb1990: Instead of checking `ç` in many functions, we can check it in the first line of the controller and return if `false`. Or wrap to whole controller logic inside
`if(directiveOptions.leaveRequest){ ....controller....}`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -338,23 +391,30 @@ define([
...
+       * Loads entitlements and sets the absences type available for the user
+       *
+       * @param {Object} that contains all absencetypes and their ids
+       * @return {Promise}
+       */
+      function loadAbsenceTypesFromEntitlements(absenceTypesAndIds) {
+        return Entitlement.all({
+          contact_id: vm.leaveRequest.contact_id,
+          period_id: vm.period.id,
+          type_id: { in: absenceTypesAndIds.ids
```
deb1990: Indentation wrong

deb1990: Just a minor note. Can we use `IN` (caps) to make it more readable

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -530,26 +599,197 @@ define([
...
+      /**
+       * Sets leave requestion statuses
+       */
+      function setStatuses() {
+        var allowedStatuses = ['approved', 'more_information_requested', 'cancelled'];
+
+        if (vm.role === 'manager') {
+          //remove current status of leaverequest
+          _.remove(allowedStatuses, function (status) {
+            return status == vm.leaveRequestStatuses[vm.leaveRequest.status_id].name;
```
deb1990: === for string comparison

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -530,26 +599,197 @@ define([
...
+          });
+
+          //filter vm.leaveRequestStatuses to contain statues relevant for manager to act
+          for (var key in vm.leaveRequestStatuses) {
```
deb1990: Define `key` and `status` in the first line of the function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
```diff
@@ -530,26 +599,197 @@ define([
...
+      }
+
+      /**
+       * Called after successful submission of leave request
+       *
+       * @param {String} eventName name of the event to emit
+       */
+      function postSubmit(eventName) {
+        $rootScope.$emit(eventName, vm.leaveRequest);
+        vm.error = undefined;
```
deb1990: Setting `undefined` value to a variable does not look good, we should use `null`. `undefined` is when the variable is not initialized. So if we want to assign a value that means we are defining it. So we should use `null`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: null is for assignment but here we are specifying that it is declared but not yet assigned value

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: return value is not used so ternary is not used

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: its a local variable relevant in the block in which its declared, so its optimal

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: initially it was all kept in single place but alessandro suggested to split and have smaller functions, also the functions have an else initialization and sometimes additional code

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
dhavalsharma: why is indentation wrong, its 2 spaces?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/leave-request-popup-controller.js
deb1990: Javascript does not support block variable's, any variable declared in a function would be available throughout the function.

PR #1352, opened 2017-01-13
PCHR-1590: Fix Advanced Search for Job Roles Multi-selects

MiyaNoctem:
## Problem
Advanced search throws no result if you select a Region or Location for a job role. Multi-selects in general for job role section of dvanced search were being built with option_value.id as values, but query was being built comparing these option_value.id's with option_value.value's stored in job roles table. 

## Solution
Fixed by changing the way selects are built, so that they now use option_value.value, just as they are built on job contract section of advanced search.


com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
```diff
@@ -335,7 +335,7 @@ private function getLevelTypeOptions() {
     }
 
     private function getOptionsWithIDAsKeyFor($daoName, $field) {
```
davialexandre: Since now value is used as key, you also have to change the method name to reflect this

PR #1357, opened 2017-01-17
PCHR-1869: Modify Comment Entity Get and Delete API endpoints

tunbola:
This PR modifies the default Get and Delete API endpoints of the Comment entity. The Get endpoint is modified to return only non deleted comments.
The Delete endpoint is modified to only soft delete comments by setting the is_deleted column of the comment to 1.

Given a Comment table

| id | entity_id | entity_name | contact_id| text | created_at | is_deleted |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |------------- |
|1|1|LeaveRequest|2|Randon Comment 1 |2017-01-17 12:51:13|0|
|2|1|LeaveRequest|2|Random Comment 2|2017-01-18 12:51:13|0|
|3|1|LeaveRequest|2|Random Comment 2|2017-01-19 12:51:13|1|

A call to
```php
$result = civicrm_api3('Comment', 'get', []);
```
Will return only non deleted values 
```json
{
  "is_error": 0,
  "version": 3,
  "count": 2,
  "values": [
    {
      "id": "1",
      "entity_name": "LeaveRequest",
      "entity_id": "1",
      "text": "Randon Comment 1",
      "contact_id": "2",
      "created_at": "2017-01-17 12:51:13",
      "is_deleted": "0"
    },
    {
      "id": "2",
      "entity_name": "LeaveRequest",
      "entity_id": "1",
      "text": "Randon Comment 2",
      "contact_id": "2",
      "created_at": "2017-01-18 12:51:13",
      "is_deleted": "0"
    }
  ]
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/API/Query/CommentSelect.php
+++ b/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/API/Query/CommentSelect.php
@@ -1,0 +1,81 @@
+<?php
+
+use Civi\API\SelectQuery;
+use CRM_HRComments_BAO_Comment as Comment;
+
+/**
+ * This class is basically a wrapper around Civi\API\SelectQuery.
+ */
+class CRM_HRComments_API_Query_CommentSelect {
+
+  /**
+   * @var array
+   *   An array of params passed to an API endpoint
+   */
+  private $params;
+
+  /**
+   * @var \Civi\API\SelectQuery
+   *  The SelectQuery instance wrapped by this class
+   */
+  private $query;
+
+  public function __construct($params) {
+    $this->params = $params;
+    $this->buildCustomQuery();
+  }
+
+  /**
+   * Build the custom query.
+   */
+  private function buildCustomQuery() {
+    $customQuery = CRM_Utils_SQL_Select::from(Comment::getTableName() . ' as a');
+
+    $this->addWhere($customQuery);
+    $this->filterReturnFields();
+    $this->query = new SelectQuery(Comment::class, $this->params, false);
+    $this->query->merge($customQuery);
+  }
+
+  /**
+   * Add the conditions to the query.
+   *
+   * This where it is ensured that only non soft-deleted comments are returned
+   *
+   * @param \CRM_Utils_SQL_Select $customQuery
+   */
+  private function addWhere(CRM_Utils_SQL_Select $customQuery) {
+    $conditions[] = 'a.is_deleted = 0';
+    $customQuery->where($conditions);
+  }
+
+  /**
+   * Executes the query
+   *
+   * @return array
+   */
+  public function run() {
+    $results = $this->query->run();
+    return $results;
+  }
+
+  /**
+   * This function allows some fields to be filtered out of the query results.
+   * Currently only the is_deleted field is filtered out
+   */
+  private function filterReturnFields() {
+    if (empty($this->params['return'])) {
+      $allFields = array_keys(Comment::fields());
+      $key = array_search('is_deleted', $allFields);
+      unset($allFields[$key]);
+      $this->params['return'] = $allFields;
+    }
+
+    if (!empty($this->params['return'])) {
+      if (in_array('is_deleted', $this->params['return'])){
+        $key = array_search('is_deleted', $this->params['return']);
+        unset($this->params['return'][$key]);
+      }
+    }
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
+++ b/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
@@ -20,6 +20,7 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
       self::validateParams($params);
     }
     unset($params['created_at']);
+    unset($params['is_deleted']);
 
     $instance = new self();
     $instance->copyValues($params);
@@ -44,6 +45,26 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
    */
   public static function validateParams($params) {
     self::validateMandatory($params);
+    self::validateCommentSoftDeleteDuringUpdate($params);
+  }
+
+  /**
+   * A method for validating that a comment cannot be soft deleted
+   * during an update on the BAO
+   *
+   * @param array $params
+   *   The params array received by the create method
+   *
+   * @throws \CRM_HRComments_Exception_InvalidCommentException
+   */
+  public static function validateCommentSoftDeleteDuringUpdate($params) {
+    if (isset($params['id']) && $params['is_deleted'] == 1) {
+      throw new InvalidCommentException(
+        'Comment can not be soft deleted during an update, use the delete method instead!',
+        'comment_cannot_soft_delete_comment',
+        'is_deleted'
+      );
+    }
   }
 
   /**
@@ -87,5 +108,18 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
         'contact_id'
       );
     }
+  }
+
+  /**
+   * Soft Deletes the comment with the given ID by setting the is_deleted column to 1
+   *
+   * @param int $id The ID of the comment to be soft deleted
+   *
+   * @return boolean
+   */
+  public static function softDelete($id) {
+    $comment = self::findById($id);
+    $comment->is_deleted = 1;
+    $comment->save();
   }
 }
--- a/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/Test/Fabricator/Comment.php
+++ b/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/Test/Fabricator/Comment.php
@@ -1,0 +1,20 @@
+<?php
+
+use CRM_HRComments_BAO_Comment as Comment;
+
+class CRM_HRComments_Test_Fabricator_Comment {
+
+  public static function fabricate($params) {
+    $params = self::mergeDefaultParams($params);
+
+    return Comment::create($params);
+  }
+
+  private static function mergeDefaultParams($params) {
+    $defaultParams = [
+      'text' => 'This is some random comment',
+      'entity_name' => 'DefaultEntity',
+    ];
+    return array_merge($defaultParams, $params);
+  }
+}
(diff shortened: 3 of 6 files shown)


uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/Test/Fabricator/Comment.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+use CRM_HRComments_BAO_Comment as Comment;
+
+class CRM_HRComments_Test_Fabricator_Comment {
+
+  public static function fabricate($params) {
+    $params = self::mergeDefaultParams($params);
+
+    $comment = Comment::create($params);
```
davialexandre: since you're not using `$comment` for anything, what about `return Comment::create($params)`?

uk.co.compucorp.civicrm.hrcomments/api/v3/Comment.php
```diff
@@ -31,16 +31,60 @@ function civicrm_api3_comment_create($params) {
...
-  return _civicrm_api3_basic_get(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+  $results =  _civicrm_api3_basic_get(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+
+  if ($results['count'] > 0){
+    $nonDeleted = array_filter($results['values'], function($item) {
+      return !$item['is_deleted'];
+    });
+    $results['values'] = array_values($nonDeleted);
+  }
+
```
davialexandre: mmmm... there are a few problems with this implementation:
- It will mess with `$results['count']`, because you removing items from returned values but the count number will still be the original number of items returned
- It will mess with pagination (limit and offset). Example: say you have 6 comments and only the first one of them is deleted. Then you use `limit` to only get the first 5, which will include the deleted one. The code will remove the deleted comment and will actually return only 4 comments, even though, in the database, we have 5 non deleted comments.

In order to avoid these problems, you'll need to play with the `Civi\API\SelectQuery` class, which is what is used internally by the `get` API. Take a look at [API/Query/LeaveRequestSelect.php](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php) and [api/v3/LeaveRequest.php](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php#L108-L111) to see how something similar was done for the LeaveRequest API

uk.co.compucorp.civicrm.hrcomments/api/v3/Comment.php
```diff
@@ -31,16 +31,60 @@ function civicrm_api3_comment_create($params) {
  * @throws API_Exception
  */
 function civicrm_api3_comment_delete($params) {
-  return _civicrm_api3_basic_delete(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+  return _civicrm_api3_comment_soft_delete(_civicrm_api3_get_BAO(__FUNCTION__), $params);
```
davialexandre: The `_civicrm_api3_basic_delete` functions call the BAO's `del` method internally. So, instead of implementing this soft delete logic on the API level, please implement on the BAO. This is also important because, if we don't want to use the API for some reason, the BAO will work the same way

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/API/Query/CommentSelect.php
```diff
@@ -0,0 +1,79 @@
...
+    $this->filterReturnFields();
+    $this->query = new SelectQuery(Comment::class, $this->params, false);
+    $this->query->merge($customQuery);
+  }
+
+  /**
+   * Add the conditions to the query.
+   *
+   * This where it is ensured that only non soft-deleted comments are returned
+   * deleted contracts and with balance changes.
```
davialexandre: I guess this line is a leftover from the comment on the class you based yourself on

davialexandre: I guess this line is a leftover from the comment on the class you based yourself on

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/API/Query/CommentSelect.php
```diff
@@ -0,0 +1,79 @@
...
+    return $results;
+  }
+
+  /**
+   * This function allows some fields to be filtered out of the query results.
+   * Currently only the is_deleted field is filtered out
+   */
+  private function filterReturnFields() {
+    if (empty($this->params['return'])) {
+      $this->params['return'] = ['id', 'entity_name', 'entity_id', 'text', 'contact_id', 'created_at'];
```
davialexandre: what about getting the list of the filters with `Comment::fields()`, so you don't have to update this manually every time a new field gets added to the DAO?

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
```diff
@@ -88,4 +109,27 @@ private static function validateMandatory($params) {
...
+   * @param int $id The ID of the comment to be soft deleted
+   *
+   * @return boolean
+   */
+  public static function softDelete($id) {
+    $comment = new self();
+    $comment->id = $id;
+
+    if ($comment->find()) {
+      while ($comment->fetch()) {
```
davialexandre: Does is really need to a while? You're using the $id, so if we get more than one comment it means there's something really wrong happening :D 

To avoid the secondary call to `->fetch()` you can pass `true` to `find()` and it will call `fetch()` internally

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
```diff
@@ -88,4 +109,27 @@ private static function validateMandatory($params) {
...
+        $comment->is_deleted = 1;
+        $comment->save();
+      }
+      return true;
+    }
+    else{
```
davialexandre: This else is unecessary. You can rewrite this as
```php
if($comment->find(true)) {
   // do things
   return true;
}

return false;
```

uk.co.compucorp.civicrm.hrcomments/api/v3/Comment.php
```diff
@@ -31,7 +39,14 @@ function civicrm_api3_comment_create($params) {
...
  */
 function civicrm_api3_comment_delete($params) {
-  return _civicrm_api3_comment_soft_delete(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+  civicrm_api3_verify_mandatory($params, NULL, ['id']);
+
+  if (CRM_HRComments_BAO_Comment::softDelete($params['id'])) {
+    return civicrm_api3_create_success();
+  }
+  else{
+    throw new API_Exception('Comment cannot be deleted because it does not exist');
```
davialexandre: The `findByPK` method already throws an exception when it can't find a record with the given ID. So, if inside `Comment::softDelete` you fetch the comment with `$comment = self::findByPk($id)` you won't need to manually throw the exception here

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -76,6 +77,28 @@ public function testValidateParamsIsCalledOnCreate() {
...
+   * @expectedExceptionMessage Comment can not be soft deleted during an update use the delete method instead!
+   */
+  public function testTryingToSoftDeleteCommentOnUpdateThrowsException() {
+    $comment1 = Comment::create([
+      'entity_id' => 1,
+      'entity_name' => 'LeaveRequest',
+      'text' => 'This is a sample comment',
+      'contact_id' => 1,
+    ]);
+
```
davialexandre: What about using `$this->setExpectedException()` here, just to ensure it will check the exception only for the second create?

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -76,6 +77,28 @@ public function testValidateParamsIsCalledOnCreate() {
...
+    Comment::create([
+      'id' => $comment1->id,
+      'entity_id' => 1,
+      'entity_name' => 'LeaveRequest',
+      'text' => 'This is a sample comment',
+      'contact_id' => 1,
```
davialexandre: Since all the other fields are the same, on this call you only need to pass the `id` and `is_deleted`

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -127,4 +150,69 @@ public function testCreatedAtCannotBeChangedWhenUpdatingExistingComment() {
...
+    $comment->fetch();
+    $this->assertEquals($comment->id, $comment1->id);
+    $this->assertEquals(1, $comment->is_deleted);
+
+    $comment->fetch();
+    $this->assertEquals($comment->id, $comment2->id);
+    $this->assertEquals(0, $comment->is_deleted);
+  }
+
+  public function testIsDeletedAtWillBeZeroIrrespectiveOfTheValuePassedViaIsDeletedParameterOnCreate() {
```
davialexandre: The test name says "testIsDeleted**At**" but it should probably be "testIsDeleted"

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -127,4 +150,69 @@ public function testCreatedAtCannotBeChangedWhenUpdatingExistingComment() {
...
+      'created_at' => CRM_Utils_Date::processDate('2016-01-01'),
+      'is_deleted' => 1
+    ], false);
+
+    $comment = new Comment();
+    $comment->find();
+    $this->assertEquals($comment->N, 1);
+    $comment->fetch();
+
+    $this->assertEquals(0, $comment->is_deleted);
```
davialexandre: On this test, can you try to change some other field and assert that it was not changed? Just to make sure that the updated was fine and only the `is_deleted` value was the one not changed

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
tunbola: This call to create will run validateParams internally and will throw an exception if these fields are not present

PR #1358, opened 2017-01-17
PCHR-1666: Standalone Basic Installation script for CiviHR.

hitesh-jain:
## Purpose:
Need to create a unified installation script for CiviHR that will setup an CiviHR site with an single command line without having to depend on buildkit to spin up sites. Install script needs to be  executed without having to download any dependencies and would be an generic script that can ran to setup basic CiviHR site using and all its required installations. Script needs to be able to run from any instance ie within buildkit , as a standalone install script , with Drupal profiler and with Drush command to execute and setup the CiviHR site without having to know from where it has been execute from.

## Implementation:
Created a basic install script that implements below points -
1. Sets up databases
2. Fetches the latest CiviCRM (from https://download.civicrm.org/civicrm-4.7.9-drupal.tar.gz)
3. Fetches a remote drush.make file (from https://raw.githubusercontent.com/civicrm/civihr/PCHR-1666/bin/drush.make)
4. Installs Drupal, CiviCRM, CiviHR.

## Installation Steps:
1. Download Executable file from bin/civihr-install and to require location.(https://raw.githubusercontent.com/civicrm/civihr/PCHR-1666/bin/civihr-install)
2. Add civihr-install command to your $PATH variable , run below command from command prompt replacing location of file:
```html
export PATH=$PATH:/<location-of-file>
```
3. Now its time to execute installation , run below command replacing all the required parameters 
```html
civihr-install <civihr-site-name>  --dbuser <mysql-username> --dbpass <mysql-password> --cmsdb <name-of-drupal-database> --crmdb <name-of-civicrm-database> --url <site-url>
```
4. Setup Apache vhosts manually and hosts file to point to desired directory and restart apache (provides you details when you are finish with installation)

## Dependencies:
1. Requires up and running Apache , MySQL and PHP configuration.
2. Required Git to be installed.

### DIFF
--- a/bin/civihr-install
+++ b/bin/civihr-install
@@ -286,9 +286,9 @@ function _amp_install_cms() {
   else
     amp create -f --root="$CMS_ROOT" --name="$amp_name" --prefix=CMS_ --output-file="$amp_vars_file_path" --perm="$CMS_DB_PERM"
   fi
-  echo $amp_vars_file_path;exit
   source "$amp_vars_file_path"
-  rm -f "$amp_vars_file_path"
+
+  #rm -f "$amp_vars_file_path"
 }
 
 function _amp_install_civi() {
@@ -580,11 +580,18 @@ function civihr_parse() {
       #   
       #   ;;
 
-      --type)
-        SITE_TYPE="$1"
+      --dbuser)
+        DB_USER="$1"
         shift
         ;;
-
+      --dbpass)
+        DB_PASS="$1"
+        shift
+        ;;
+      --dbport)
+        DB_PORT="$1"
+        shift
+        ;;
       --url)
         CMS_URL="$1"
         shift
@@ -598,17 +605,14 @@ function civihr_parse() {
     esac
   done
 
-#[ -z "$CACHE_DIR" ]          && CACHE_DIR="$TMPDIR/git-cache"
 [ -z "$WEB_ROOT" ]           && WEB_ROOT="$BLDDIR/$SITE_NAME"
 [ -z "$CMS_ROOT" ]           && CMS_ROOT="$WEB_ROOT"
 [ -z "$CIVI_SITE_KEY" ]      && CIVI_SITE_KEY=$(civihr_makepasswd 16)
 [ -z "$ADMIN_PASS" ]         && ADMIN_PASS=$(civihr_makepasswd 12)
 [ -z "$DEMO_PASS" ]          && DEMO_PASS=$(civihr_makepasswd 12)
-[ -z "$SITE_TYPE" ]          && SITE_TYPE="$SITE_NAME"
 [ -z "$CMS_TITLE" ]          && CMS_TITLE="$SITE_NAME"
 [ -z "$CMS_HOSTNAME" ]       && CMS_HOSTNAME=$(php -r '$p = parse_url($argv[1]); echo $p["host"];' "$CMS_URL")
 [ -z "$CMS_PORT" ]           && CMS_PORT=$(php -r '$p = parse_url($argv[1]); echo $p["port"];' "$CMS_URL")
-[ -z "$SITE_CONFIG_DIR" ]    && SITE_CONFIG_DIR="$PRJDIR/app/config/$SITE_TYPE"
 [ -z "$SITE_TOKEN" ]         && SITE_TOKEN=$(civihr_makepasswd 16)
 
 }
@@ -777,7 +781,7 @@ EOSQL
 ###############################################################################
 ## Run the download scripts if necessary
 function civihr_app_download() {
-  civihr_assertvars civihr_app_download WEB_ROOT PRJDIR SITE_NAME SITE_TYPE TMPDIR #CACHE_DIR
+  civihr_assertvars civihr_app_download WEB_ROOT PRJDIR SITE_NAME TMPDIR #CACHE_DIR
 
   echo "[[Download $SITE_NAME ( in '$WEB_ROOT')]]"
 
@@ -799,7 +803,7 @@ function civihr_app_download() {
 ###############################################################################
 ## Runs the installation process for CMS, CiviCRM and CiviHR.
 function civihr_app_install() {
-  civihr_assertvars civihr_app_install WEB_ROOT  SITE_NAME SITE_ID SITE_TYPE
+  civihr_assertvars civihr_app_install WEB_ROOT  SITE_NAME SITE_ID
 
   echo "[[Install $SITE_NAME ( in '$WEB_ROOT')]]"
 
--- a/bin/civihr-install
+++ b/bin/civihr-install
@@ -71,13 +71,12 @@ CMS_URL=
 CMS_ROOT=
 
 ## DB credentials for CMS
-## (suggested: autogenerate via 'amp create -f --root="$WEB_ROOT" --prefix=CMS_ --url="$CMS_URL"')
 CMS_DB_DSN=
 CMS_DB_ARGS=
-CMS_DB_HOST=
+CMS_DB_HOST=127.0.0.1
 CMS_DB_NAME=
 CMS_DB_PASS=
-CMS_DB_PORT=
+CMS_DB_PORT=3306
 CMS_DB_USER=
 CMS_DB_PERM=admin
 
@@ -85,13 +84,12 @@ CMS_DB_PERM=admin
 CIVI_CORE=
 
 ## DB credentials for Civi
-## (suggested: autogenerate via 'amp create -f --root="$WEB_ROOT" --name=civi --prefix=CIVI_ --skip-url')
 CIVI_DB_DSN=
 CIVI_DB_ARGS=
-CIVI_DB_HOST=
+CIVI_DB_HOST=127.0.0.1
 CIVI_DB_NAME=
 CIVI_DB_PASS=
-CIVI_DB_PORT=
+CIVI_DB_PORT=3306
 CIVI_DB_USER=
 CIVI_DB_PERM=super
 
@@ -205,6 +203,18 @@ function civihr_mkdir() {
 }
 
 ###############################################################################
+## Creates host and port to a single string
+function civihr_build_hostport() {
+  local host=$1
+  local port=$2
+  if [ -z "$port" ]; then
+    echo "$host"
+  else
+    echo "$host:$port"
+  fi
+}
+
+###############################################################################
 ## Appends the civibuild settings directives to a file
 ## usage: civihr_inject_settings <php-file> <settings-dir-name>
 ## example: civihr_inject_settings "/var/www/build/drupal/sites/foo/civicrm.settings.php" "civicrm.settings.d"
@@ -264,46 +274,42 @@ EOF
 }
 
 ###############################################################################
-## Setup HTTP and MySQL services
-## This outputs several variables: CMS_URL, CMS_DB_*, CIVI_DB_*, and TEST_DB_*
-function amp_install() {
-  _amp_install_cms
-  _amp_install_civi
+## Setup MySQL services
+function database_install() {
+  _database_install_cms
+  _database_install_civi
 }
 
-function _amp_install_cms() {
-  echo "[[Setup MySQL and HTTP for CMS]]"
+function _database_install_cms() {
+
+  echo "[[Setup MySQL for CMS]]"
   civihr_assertvars _amp_install_cms CMS_ROOT SITE_NAME SITE_ID TMPDIR
+  
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CMSdbName" >/dev/null 2>&1 || { echo "[[Please provide CMS database name ( i.e --crmdb ) or Check if its already exists.]]";exit 1;  }
 
-  civihr_maketempfile
+  [ -z "$CMS_DB_USER" ]           && CMS_DB_USER="$DB_USER"
+  [ -z "$CMS_DB_PASS" ]           && CMS_DB_PASS="$DB_PASS"
+  [ -z "$CMS_DB_NAME" ]           && CMS_DB_NAME="$CMSdbName"
 
-  local amp_vars_file_path=$(php $TMPDIR/mktemp.php ampvar)
-  local amp_name="cms$SITE_ID"
-  [ "$SITE_ID" == "default" ] && amp_name=cms
-
-  if [ -n "$CMS_URL" ]; then
-    amp create -f --root="$CMS_ROOT" --name="$amp_name" --prefix=CMS_ --url="$CMS_URL" --output-file="$amp_vars_file_path" --perm="$CMS_DB_PERM"
-  else
-    amp create -f --root="$CMS_ROOT" --name="$amp_name" --prefix=CMS_ --output-file="$amp_vars_file_path" --perm="$CMS_DB_PERM"
-  fi
-  source "$amp_vars_file_path"
-
-  #rm -f "$amp_vars_file_path"
+  CMS_DB_HOSTPORT=$(civihr_build_hostport "$CMS_DB_HOST" "$CMS_DB_PORT")
+  [ -z "$CMS_DB_DSN" ]           && CMS_DB_DSN="mysql://${DB_USER}:${DB_PASS}@${CMS_DB_HOSTPORT}/${CMSdbName}?new_link=true"
 }
 
-function _amp_install_civi() {
+function _database_install_civi() {
+
   echo "[[Setup MySQL for Civi]]"
   civihr_assertvars _amp_install_civi CMS_ROOT SITE_NAME SITE_ID TMPDIR
 
-  local amp_vars_file_path=$(php $TMPDIR/mktemp.php ampvar)
-  local amp_name="civi$SITE_ID"
-  [ "$SITE_ID" == "default" ] && amp_name=civi
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CIVIdbName" >/dev/null 2>&1 || { echo "[[Please provide CRM database name ( i.e --crmdb ) or Check if its already exists.]]";exit 1;  }
 
-  amp create -f --root="$CMS_ROOT" --name="$amp_name" --prefix=CIVI_ --skip-url --output-file="$amp_vars_file_path" --perm="$CIVI_DB_PERM"
+  [ -z "$CIVI_DB_USER" ]           && CIVI_DB_USER="$DB_USER"
+  [ -z "$CIVI_DB_PASS" ]           && CIVI_DB_PASS="$DB_PASS"
+  [ -z "$CIVI_DB_NAME" ]           && CIVI_DB_NAME="$CIVIdbName"
 
-  source "$amp_vars_file_path"
-  rm -f "$amp_vars_file_path"
+  CIVI_DB_HOSTPORT=$(civihr_build_hostport "$CIVI_DB_HOST" "$CIVI_DB_PORT")
+  [ -z "$CIVI_DB_DSN" ]           && CIVI_DB_DSN="mysql://${DB_USER}:${DB_PASS}@${CIVI_DB_HOSTPORT}/${CIVIdbName}?new_link=true"
 }
+
 
 ###############################################################################
 ## Generate config files and setup database
@@ -313,30 +319,24 @@ function civicrm_install() {
     echo "Failed to locate valid civi root: $CIVI_CORE"
     exit 1
   fi
-
   ## Create CiviCRM data dirs
-  amp datadir "$CIVI_FILES" "$CIVI_TEMPLATEC"
+  civihr_mkdir "$CIVI_FILES" "$CIVI_TEMPLATEC"
   if [ -n "$CIVI_EXT_DIR" ]; then
-    amp datadir "$CIVI_EXT_DIR"
+    civihr_mkdir "$CIVI_EXT_DIR"
   fi
 
   ## Create CiviCRM config files
   civicrm_make_settings_php
   civicrm_make_setup_conf
-  
 
   pushd "$CIVI_CORE" >> /dev/null
-    ## Does this build include development support (eg git or tarball-based)?
-    if [ -e "xml" -a -e "bin/setup.sh" ]; then
-      env SITE_ID="$SITE_ID" bash ./bin/setup.sh
-    elif [ -e "sql/civicrm.mysql" -a -e "sql/civicrm_generated.mysql" ]; then
-      cat sql/civicrm.mysql sql/civicrm_generated.mysql | eval mysql $CIVI_DB_ARGS
+    if [ -e "sql/civicrm.mysql" -a -e "sql/civicrm_generated.mysql" ]; then
+      cat sql/civicrm.mysql sql/civicrm_generated.mysql | mysql -u$CIVI_DB_USER -p$CIVI_DB_PASS $CIVI_DB_NAME
     else
       echo "Failed to locate civi SQL files"
     fi
   popd >> /dev/null
-
-  eval mysql $CIVI_DB_ARGS <<EOSQL
+  eval mysql -u$CIVI_DB_USER -p$CIVI_DB_PASS $CIVI_DB_NAME <<EOSQL
     UPDATE civicrm_domain SET name = '$CIVI_DOMAIN_NAME';
     SELECT @option_group_id := id
       FROM civicrm_option_group n
@@ -348,17 +348,7 @@ function civicrm_install() {
 EOSQL
 }
 
-###############################################################################
-## Creates host and port to a single string
-function civihr_build_hostport() {
-  local host=$1
-  local port=$2
-  if [ -z "$port" ]; then
-    echo "$host"
-  else
-    echo "$host:$port"
-  fi
-}
+
 
 ###############################################################################
 ## Generate a "civicrm.settings.php" file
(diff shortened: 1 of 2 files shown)


MiyaNoctem:
Could you add to installation instructions where you should be located (what directory), before you execute civihr-install, and explain that the install script will create a directory for the installation with <civihr-site-name> as name in it?

bin/civihr-install
```diff
@@ -0,0 +1,808 @@
...
+  _database_install_civi
+}
+
+## Setting up Database for CMS i.e Drupal
+function _database_install_cms() {
+
+  echo "[[Setup MySQL for CMS]]"
+  civihr_assertvars _database_install_cms CMS_ROOT SITE_NAME SITE_ID TMPDIR
+  # Create database.
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CMSdbName" >/dev/null 2>&1 || { echo "[[Please provide CMS database name ( i.e --crmdb ) or Check if its already exists.]]";exit 1;  }
```
MiyaNoctem: The echo mentions --crmdb, when the check is for CMS DB.  Probably should be something like:

`echo "[[Please provide CMS database name ( i.e --cmsdb ) or Check if its already exists.]]"`

bin/civihr-install
```diff
@@ -0,0 +1,808 @@
...
+  fi
+
+  civihr_makeparent "$lock"
+  if pidlockfile.php "$lock" $$ $CACHE_LOCK_WAIT ; then
+    php -r 'echo time();' > "$lastrun"
+    if [ ! -d "$cachedir" ]; then
+      ## clone
+      echo "[[Initialize cache dir: $cachedir]]"
+      civihr_makeparent "$cachedir"
+      timeout.php $SCM_TIMEOUT git clone --mirror "$url" "$cachedir"
```
davialexandre: Instead of cloning the repository, which makes mandatory to have git installed, can't we download the tarballs from github? Github provides those nice URLs where you can get them. For example, the tarball for civihr staging can be downloaded at https://github.com/civicrm/civihr/archive/staging.zip

bin/civihr-install
hitesh-jain: @MiyaNoctem Fixed.

bin/civihr-install
hitesh-jain: @davialexandre I think we would anyways require Git to be installed before executing as we are cloning all the CiviHR required repos you can see in drush.make file because I think we would require cloning process.

MiyaNoctem:
You should update the link to the civihr-install file in the PR description, since the current link doesn't include the most recent fixes.

bin/civihr-install
```diff
@@ -596,8 +609,7 @@ export PATH="$PATH:$BINDIR"
 civihr_parse "$@"
 
 # Checking for Drush dependencies if not present downloading into temp.
-command -v drush >/dev/null 2>&1 || ( echo "[[Downloading dependencies.........]]" && curl -Lo "$TMPDIR/drush" "https://github.com/drush-ops/drush/releases/download/8.1.8/drush.phar"  && chmod +x "$TMPDIR/drush" && export PATH="$PATH:$TMPDIR/drush" )
-
+command -v drush >/dev/null 2>&1 || ( echo "[[Downloading dependencies.........]]" && curl -Lo "$TMPDIR/drush" "https://github.com/drush-ops/drush/releases/download/8.1.8/drush.phar" && chmod +x "$TMPDIR/drush" && export PATH="$TMPDIR:$PATH" )
```
davialexandre: As an user, I would be VERY mad to find out a script modified my PATH without at least asking me if I'm ok with that.

I think we should either:
- Make drush mandatory. So we just inform the user that drush has to be installed and available and stop the script if it's not
- Tell the user that drush is not present and ask if it's ok to download it and add it to PATH
- Download it silently, but instead of adding it to PATH, add the absolut path to a variable (DRUSH=$TMPDIR/drush) and then use it to call drush. I'm not sure if that is possible though

bin/civihr-install
davialexandre: Yeah, but if instead of cloning you download the tarball, then git will not be necessary.

For the drush.make file, I believe we could change it to something like this:

```
libraries[civihr][destination] = modules
libraries[civihr][directory_name] = civicrm/tools/extensions/civihr
libraries[civihr][download][type] = get
libraries[civihr][download][url] = https://github.com/civicrm/civihr/archive/master.tar.gz
```

bin/civihr-install
```diff
@@ -0,0 +1,820 @@
...
+
+# Assigning values.
+[ -z "$SITE_NAME" ]          && civihr_app_usage && exit 99
+[ -z "$WEB_ROOT" ]           && WEB_ROOT="$BLDDIR/$SITE_NAME"
+[ -z "$CMS_ROOT" ]           && CMS_ROOT="$WEB_ROOT"
+[ -z "$CIVI_SITE_KEY" ]      && CIVI_SITE_KEY=$(civihr_makepasswd 16)
+[ -z "$ADMIN_PASS" ]         && ADMIN_PASS=$(civihr_makepasswd 12)
+[ -z "$CMS_TITLE" ]          && CMS_TITLE="$SITE_NAME"
+[ -z "$CMS_HOSTNAME" ]       && CMS_HOSTNAME=$(php -r '$p = parse_url($argv[1]); echo $p["host"];' "$CMS_URL")
+[ -z "$CMS_PORT" ]           && CMS_PORT=$(php -r '$p = parse_url($argv[1]); echo $p["port"];' "$CMS_URL")
```
MiyaNoctem: This line is issuing a PHP notice when port is not defined:
> PHP Notice:  Undefined index: port in Command line code on line 1

We could have a default value if it's not set, or try something like:
`[ -z "$CMS_PORT" ]           && CMS_PORT=$(php -r '$p = parse_url($argv[1]); if (isset($p["port"])) {echo $p["port"];}' "$CMS_URL")`

bin/civihr-install
hitesh-jain: @davialexandre But I think we require this to clone repos , as buildkit also clones and  then we can use those to work on right because if we download that directly then I think it won't have any git initialized/history so they have to done manually for each of the repos ?

hitesh-jain:
@MiyaNoctem what link are you referring to can you please help to understand thanks

bin/civihr-install
davialexandre: That's right, it won't be git repository and hence it won't have the repository history. But why do you need that? If I understand things correctly, this is a general purpose script, right? So, if I only want to install civihr, without doing any development on top of it, I don't need the repository. 
Maybe we can have an additional flag/option that we can pass to the script in order to tell if we want the repositories or not

davialexandre:
@hitesh-jain  Regarding what @MiyaNoctem said about updating the script link on the PR, please use this: https://github.com/civicrm/civihr/blob/PCHR-1666/bin/civihr-install it points to your branch instead of a specific commit, so it will always be up to date

bin/civihr-install
```diff
@@ -0,0 +1,820 @@
+#!/bin/bash
```
davialexandre: Why is the script name civihr-install and not civihr-install.sh? I think it's always a good practice to keep the extension. If the user doesn't want it, they can always removed it

hitesh-jain:
@davialexandre @MiyaNoctem Thanks for noticing incorrect link  !!

davialexandre:
@hitesh-jain actually the correct link to the raw file is https://raw.githubusercontent.com/civicrm/civihr/PCHR-1666/bin/civihr-install. Sorry by giving you the wrong link on the last comment

hitesh-jain:
@davialexandre aha right !! thanks

bin/civihr-install
```diff
@@ -0,0 +1,820 @@
...
+
+Common Options :
+  <site-name>        The name of the site-directory to build
+
+Syntax: $APP <site-name> [required-options]
+Description: Downloads and/or installs the application.
+
+Required options:
+  --dbuser <mysql-username>     Username for mysql
+  --dbpass <mysql-password>     Password for mysql
```
davialexandre: Why is the password required? What if I want to use an user that doesn't have a password?

bin/civihr-install
hitesh-jain: I haven't used extension here because it needs to be executable as a command and as I've read executable files should be without any extension (did some research  :https://google.github.io/styleguide/shell.xml#File_Extensions ). Also I think using sh command to execute a .sh file won't be much user friendly.

davialexandre: This is the text from the link you just shared: 

> Executables should have no extension (strongly preferred) **or a .sh extension**

The extension really doesn't have anything to do if the script is executable or not. I can put a .txt extension to file and it will be executable if give it the right permission.

You don't need to use sh to run a bash command. Even it it has the extension, you can install call it like:
```
$ ./civihr-install.sh
$ /full/path/to/civihr-install.sh
$ civihr-install.sh #if the script can be found in $PATH
```

What I said about the user removing the extension is that the user will have to download this script, right? Most will probably do it with curl. Something like this:

```
curl -O https://raw.githubusercontent.com/civicrm/civihr/PCHR-1666/bin/civihr-install.sh
```

And if they want it without the extension, they can just:

```
curl -o civihr-install https://raw.githubusercontent.com/civicrm/civihr/PCHR-1666/bin/civihr-install.sh
```

On the repository, I think it's better to keep the extension. The reason is that you can just look at the file and know that it is a script without having to open it

hitesh-jain: So if we add a extension don't you think it will cause an extra step to end user to rename file in-case of user doesn't want to use curl command to rename and directly save the file to use it. And also executing civihr-install.sh <parameters> would be not so user friendly.

bin/civihr-install
hitesh-jain: Fixed

bin/civihr-install
hitesh-jain: Yes I think you are right this is a general purpose script and we should not be bothering about providing github repository for development.
But providing additional flag/option would need additional efforts to maintain two different drush.make files one with get type that will download tar and another with git type to provide github repos to development.
@AkA84 what you think on this ?

bin/civihr-install
hitesh-jain: I'm not sure whether its a bad practice to modify temporary PATH variable within script and also I've seen Buildkit modifying PATH variable so thought that we can use same process.
And I've investigated on the 3 option you suggested in that case I think we need to then replace all the drush commands with the absolute variable substitution. Something like
```
drush -y make --concurrency=5 --working-copy "$MAKEFILE" "$WEB_ROOT"
```
would be replaced with 
```
$DRUSH -y make --concurrency=5 --working-copy "$MAKEFILE" "$WEB_ROOT"
```
Please correct if I'm wrong as I'm very new to shell scripting. thanks 
@aljazceru @AkA84 what you think about above suggestions by Davi ?

bin/civihr-install
hitesh-jain: I've changed to accept Password now. thanks

bin/civihr-install
jamienovick: I think that we perhaps should have 2 modes - standard install i.e. when this is used in a Drupal installation profile, should not need git and if you are doing a developer mode install then it could get from github.

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+    echo "$host"
+  else
+    echo "$host:$port"
+  fi
+}
+
+###############################################################################
+## Making Random temp files
+function civihr_maketempfile(){
+ cat > "$TMPDIR/mktemp.php" << EOF
```
davialexandre: Do we really need PHP for this?

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  civihr_assertvars _database_install_cms CMS_ROOT SITE_NAME SITE_ID TMPDIR
+  # Create database.
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CMSdbName" || { echo "[[Please fix above MySQL error.]]"; exit 1;  }
+  # Assign values to global variables.
+  [ -z "$CMS_DB_USER" ]           && CMS_DB_USER="$DB_USER"
+  [ -z "$CMS_DB_PASS" ]           && CMS_DB_PASS="$DB_PASS"
+  [ -z "$CMS_DB_NAME" ]           && CMS_DB_NAME="$CMSdbName"
+
+  CMS_DB_HOSTPORT=$(civihr_build_hostport "$CMS_DB_HOST" "$CMS_DB_PORT")
+  [ -z "$CMS_DB_DSN" ]           && CMS_DB_DSN="mysql://${DB_USER}:${DB_PASS}@${CMS_DB_HOSTPORT}/${CMSdbName}?new_link=true"
```
davialexandre: I think this will create an invalid dsn if `DB_PASS` is empty, won't it?

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  _database_install_civi
+}
+
+## Setting up Database for CMS i.e Drupal
+function _database_install_cms() {
+
+  echo "[[Setting up MySQL for CMS]]"
+  civihr_assertvars _database_install_cms CMS_ROOT SITE_NAME SITE_ID TMPDIR
+  # Create database.
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CMSdbName" || { echo "[[Please fix above MySQL error.]]"; exit 1;  }
```
davialexandre: I stumbled upon a very annoying error: If an error happens at the middle of the installation, but after the database has been created, in order to run it again, I'll have to go to MySQL and manually drop the database the script created, otherwise I won't be able to proceed because this line will fail since the database already exists.

I can think of 2 possible improvements here:
- Implement some sort of rollback in case of errors (maybe using traps). Not sure of how difficult/complex it would be to implement this
- Add an option to force droping existing databases during install

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  civihr_assertvars _database_install_civi CMS_ROOT SITE_NAME SITE_ID TMPDIR
+  # Create database.
+  mysql -u$DB_USER -p$DB_PASS -e "create database $CIVIdbName" || { echo "[[Please fix above MySQL error.]]"; exit 1;  }
+  # Assign values to global variables.
+  [ -z "$CIVI_DB_USER" ]           && CIVI_DB_USER="$DB_USER"
+  [ -z "$CIVI_DB_PASS" ]           && CIVI_DB_PASS="$DB_PASS"
+  [ -z "$CIVI_DB_NAME" ]           && CIVI_DB_NAME="$CIVIdbName"
+
+  CIVI_DB_HOSTPORT=$(civihr_build_hostport "$CIVI_DB_HOST" "$CIVI_DB_PORT")
+  [ -z "$CIVI_DB_DSN" ]           && CIVI_DB_DSN="mysql://${DB_USER}:${DB_PASS}@${CIVI_DB_HOSTPORT}/${CIVIdbName}?new_link=true"
```
davialexandre: The dns will probably be invalid here too if `DB_PASS` is empty

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+    git_cache_setup "https://github.com/civicrm/civicrm-core.git" "$CACHE_DIR/civicrm/civicrm-core.git"
+    git_cache_setup "https://github.com/civicrm/civicrm-drupal.git"  "$CACHE_DIR/civicrm/civicrm-drupal.git"
+    git_cache_setup "https://github.com/civicrm/civicrm-packages.git" "$CACHE_DIR/civicrm/civicrm-packages.git"
+    git_cache_setup "https://github.com/eileenmcnaughton/civicrm_developer.git" "$CACHE_DIR/eileenmcnaughton/civicrm_developer.git"
+  fi
+}
+
+###############################################################################
+## Parse options
+function civihr_parse() {
```
davialexandre: I think `civihr_parse_options` would be a better name for this method

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+    | sed "s;%%CMS_VERSION%%;${CMS_VERSION};" \
+    | sed "s;%%HR_VERSION%%;${HR_VERSION};" \
+    > "$MAKEFILE"
+
+  drush -y make --concurrency=5 --working-copy "$MAKEFILE" "$WEB_ROOT"
+}
+
+
+# Creates the default CiviHR users
+function create_default_users() {
```
davialexandre: This function is civihr specific, shouldn't it be named `civihr_create_default_users`? (also applies to the other functions below which don't have the civihr prefix)

bin/civihr-install
davialexandre: It's an extra step, but not a required step. They can still execute it just fine without removing the extension.

The thing with the extension is just a matter of organization of the project. As I said, with the extension you don't have to open the file to figure out if it's a bash script or not. What if, in the future, we also have self executable python scripts on the same dir?

One interesting option though, is what civicrm-buildkit does. It has a `bin` dir and a `lib` dir. Both have bash scripts, but those inside `bin` are supposed to be self-executable and don't have any extension. Scripts on `lib` should only be "included" and have the `.sh` extension. Maybe we can do the same for our script (related to this, check my comment about splitting the file)

bin/civihr-install
davialexandre: So, I forgot one important thing here: `export` will only affect the environment of the script itself and of its children. That is, any change you make will only have effect inside the script. So I think it's fine to call `export`.

The fix won't work though. The `export` is inside of a subshell (in between `( )`), meaning that the export will have effect only inside that subshell enviroment. My suggestion here is:

Create two new functions (it will also improve readbility):
```bash
function check_if_drush_is_available() {
  return command -v drush > /dev/null 2>&1
}

function install_drush() {
  echo "[[Downloading drush.........]]"
  curl -Lo "$TMPDIR/drush" "https://github.com/drush-ops/drush/releases/download/8.1.8/drush.phar"
  chmod +x "$TMPDIR/drush"
  export PATH="$TMPDIR:$PATH"
}
```

And then you can replace this line with:
```bash
check_if_drush_is_available || install_drush
```

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+CIVI_EXT_URL=
+
+
+
+###############################################################################
+## Display usage message
+function civihr_app_usage() {
+  APP=$(basename "$0")
+
+  php -r 'echo file_get_contents("php://stdin");' <<EOT
```
hitesh-jain: Fixed

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+
+Common Options :
+  <site-name>        The name of the site-directory to build
+
+Syntax: $APP <site-name> [required-options]
+Description: Downloads and/or installs the application.
+
+Required options:
+  --dbuser <mysql-username>     Username for mysql
+  --dbpass <mysql-password>     Password for mysql ( Optional )
```
hitesh-jain: Fixed

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  --url    <url>                The public URL of the site
+
+EOT
+  exit 99;
+}
+
+
+###############################################################################
+## Creates random password.
+function civihr_makepasswd() {
```
hitesh-jain: Removed php . Thanks

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+?>
+EOF
+# Calling temp make password php file.
+php "$TMPDIR/mkpasswd.php" "$1"
+}
+
+###############################################################################
+## Assert that shell variables are defined. If not defined, exit with an error.
+##  If a variable missing, the application dies.
+function civihr_assertvars() {
```
hitesh-jain: Fixed

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+    eval "val=\$$var"
+    echo " - $var: $val"
+  done
+}
+
+
+###############################################################################
+## Creates and ensure parent directory exists.
+## civihr_makeparent <file>
+function civihr_makeparent() {
```
hitesh-jain: Fixed.

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  for a in "$@" ; do
+    if [ ! -d "$a" ]; then
+      mkdir -p "$a"
+    fi
+  done
+}
+
+###############################################################################
+## Creates host and port to a single string.
+function civihr_build_hostport() {
```
hitesh-jain: Fixed

bin/civihr-install
```diff
@@ -0,0 +1,823 @@
...
+  if [ -z "$port" ]; then
+    echo "$host"
+  else
+    echo "$host:$port"
+  fi
+}
+
+###############################################################################
+## Making Random temp files
+function civihr_maketempfile(){
```
hitesh-jain: Fixed

bin/civihr-install
hitesh-jain: Removed.

bin/civihr-install
hitesh-jain: Added --force-drop option , so this will drop databases even if they exists.

bin/civihr-install
hitesh-jain: Fixed

bin/civihr-install
hitesh-jain: Fixed

bin/civihr-install
hitesh-jain: Changed.

bin/civihr-install
hitesh-jain: Changed, as this is the helper function so I've prefixed with '_'

bin/civihr-install
```diff
@@ -633,6 +650,12 @@ function civihr_parse_options() {
         shift
         ;;
 
+      --force-drop)
+        mysql -u$DB_USER $PASS_PARAM -e "drop database $CMSdbName" || { echo "[[Please fix above MySQL error.]]"; exit 1;  }
+        mysql -u$DB_USER $PASS_PARAM -e "drop database $CIVIdbName" || { echo "[[Please fix above MySQL error.]]"; exit 1;  }
```
davialexandre: I don't its a good idea to actually execute an action while parsing the options. We should just set a flag if the option is present and, probably on the functions which install the databases, drop them if the flag is set.

Putting the drop command here, also make it impossible to add the `--force-drop` option before the `--dbpass` param, because it depends on the $PASS_PARAM variable which is only set when the `--dbpass` option is parsed. Try this to reproduce this issue:

- Create a new site using a db user with a password
- After the site is created, use the same command from the previous step, but add the `--force-drop` option **BEFORE** the `--dbpass` option

The installation will fail saying it was not possible to connect to the database without a password, even though you passed the password as a param to the script

bin/civihr-install
```diff
@@ -200,9 +241,16 @@ function civihr_mkdir() {
...
-## Creates host and port to a single string.
-function civihr_build_hostport() {
+##########################################################
+# Creates host and port to a single string.
+# Arguments:
+#   Folder Name.
+# Returns:
+#   Creates directory.
+# Usage:
+#   civihr_mkdir <directory-name>
```
hitesh-jain: Fixed

bin/civihr-install
hitesh-jain: Fixed.

bin/civihr-install
```diff
@@ -809,11 +835,11 @@ function civihr_install_site(){
...
   database_install
 
-  ## Grant access for the Drupal database user to access the Civi Database too.
-eval mysql -u"$CIVI_DB_USER" -p"$CIVI_DB_PASS" <<EOSQL
-    GRANT ALL PRIVILEGES ON $CIVI_DB_NAME.* TO $CMS_DB_USER@'%';
-EOSQL
-  
+#   ## Grant access for the Drupal database user to access the Civi Database too.
+# eval mysql -u"$CIVI_DB_USER" $PASS_PARAM <<EOSQL
+#     GRANT ALL PRIVILEGES ON $CIVI_DB_NAME.* TO $CMS_DB_USER@'%';
```
hitesh-jain: Fixed.

davialexandre:
Everything seems to be working fine! 
- The script works when drush is not available
- It creates the database automatically when the given ones don't exist
- I can pass --force-drop at any position and it will work
- --dbpass is not required
- I can have a working site in just 2 min!!!

There's only one thing that I think we can improve: We should prevent the installation if git is not available. I forgot to install it before running the script and, to my surprise, it went all the way to the end and even said the site was installed successfully. Of course the site didn't work, so I had to check the installation output and only then I noticed errors on every step that required git.

hitesh-jain:
@davialexandre I've fixed git dependencies.

bin/civihr-install
```diff
@@ -713,10 +713,13 @@ export PATH="$PATH:$BINDIR"
 
 civihr_parse_options "$@"
 
+# Checking for Git dependencies.
+command -v git >/dev/null 2>&1 || ( echo "Please download git to continue." && exit 99 )
```
davialexandre: Did you test this? It doesn't seem to be working. It prints the "Please download git to continue" message but the installation continues:

captura de tela de 2017-02-16 07-32-18

I believe the problem is that the `exit` command is inside a subshell, so it will exit that subshell and not the one where the script is running

hitesh-jain: Fixed

bin/civihr-install
```diff
@@ -0,0 +1,987 @@
...
+#   None
+# Returns:
+#   None
+# Usage:
+#   civihr_setup_structure
+##########################################################
+function civihr_setup_structure() {
+
+  [ -z "$CMS_VERSION" ] && CMS_VERSION=7.x
+  [ -z "$CIVI_VERSION" ] && CIVI_VERSION=4.4
```
jamienovick: Should this be civi version 4.4 or higher?

aljazceru:
Script works but I do have several comments:
- all parameters need to be tested befor anything else is executed/downloaded (so test if mysql creds are ok etc, if not stop the script and notify the user. Ideally also check for installed dependencies)
- using --force-drop shouldn't create an error if the databases dont exist
- there should be an option for target directory 
- -dbpass prompt is not working (already discussed with @davialexandre )

davialexandre:
Closing this as the script now has its own repository https://github.com/compucorp/civihr-installer

PR #1359, opened 2017-01-18
PCHR-1682: Make the extension capable of expiring TOIL Requests

davialexandre:
The `LeaveBalanceChange.createExpirationRecords()` method was created a long time ago and with all the changes in the data structure, it was not working anymore. The first thing this PR does is to fix all the existing tests for that method and make them all pass. Next, the method itself was updated to also be able to expire TOIL Requests and to be able to handle some complex cases, like when a TOIL Request overlaps a Brought Forward and we have to deduct a LeaveRequest which overlaps both, like on this example:

expiring_balance_changes

The logic is as follow:
- First we load all the LeaveBalanceChanges where the expiry_date is in the past but still don't have a expiry record yet (that is an associated LeaveBalanceChange where `expired_balance_changed_id` is `NOT NULL`)
- To each expired balance change, we also fetch a start date (when that balance change is considered to have started being active). For balance changes of type 'entitlement', the start date will be the `start_date` of the AbsencePeriod of the LeavePeriodEntitlement it's associated to. For those of type 'toil_request', the start date will be the `from_date` of the LeaveRequest of the TOILRequest it's associated to.
- Next we load all the LeaveRequestDates and the amount of days deducted for them (only for approved leave requests) between the `start_date` and `expiry_date` of expired balance changes
-  We loop through all the expired balance changes (in ascending order, according to their expiry date) and check if there's any data overlapping it. If there's, and the remaining no. of days for that balance change is >= than the number of days deducted for the date, we deduct it from the expired balance change and remove the date from the list (so it won't be deducted again from other balance change)
- Finally, we create a new balance change with the remaining no. of days (it will be 0 if all the days were used) and save it as a new balance change where the `expired_balance_change_id` will point to the expired balance change.

In the end, the `createExpirationRecords()` was renamed to `createExpiryRecords()` in order to make it consistent with `expiry_date` field and also to make the code a little more British :D 

### What else has changed
The commit 79b9674 has a small change not really related to this task. It's just the addition of a description to some of the LeaveRequest API specifications.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -17,7 +17,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
    *
    * @param array $params key-value pairs
    *
-   * @return CRM_HRLeaveAndAbsences_DAO_LeaveBalanceChange|NULL
+   * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange|NULL
    */
   public static function create($params) {
     $entityName = 'LeaveBalanceChange';
@@ -487,7 +487,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
 
     $balanceChange->amount = (float)$balanceChange->amount;
 
-    $entitlement = LeavePeriodEntitlement::findById($balanceChange->source_id);
+    $entitlement = $balanceChange->getLeavePeriodEntitlement();
     $expiredAmount = self::getLeaveRequestBalanceForEntitlement(
       $entitlement,
       $approvedStatuses,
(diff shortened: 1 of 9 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -1232,6 +1233,76 @@ public function testGetBreakdownWillGroupTogetherOverriddenBalanceChangeWithLeav
...
+    $this->assertEquals($periodEntitlement1->id, LeavePeriodEntitlement::getForLeaveRequest($leaveRequest1)->id);
+    $this->assertEquals($periodEntitlement2->id, LeavePeriodEntitlement::getForLeaveRequest($leaveRequest2)->id);
+    $this->assertEquals($periodEntitlement3->id, LeavePeriodEntitlement::getForLeaveRequest($leaveRequest3)->id);
+  }
+
+  /**
+   * @expectedException RuntimeException
+   * @expectedExceptionMessage It was not possible to find an AbsencePeriod containing the given LeaveRequest
+   */
+  public function testGetForLeaveRequestShouldThrowAnExceptionIfThereIsNoAbsencePeriodContainingTheGivenLeaveRequest() {
```
tunbola: We can change the test name to `testGetForLeaveRequestShouldThrowAnExceptionIfThereIsNoAbsencePeriodContainingTheGivenLeaveRequestDates` to be more descriptive

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -410,94 +411,98 @@ public static function getAmountForTOILRequest($toilRequestID) {
...
+      $remainingAmount = $balanceChangeToExpire['amount'];
+
+      foreach($datesOverlappingBalanceChangesToExpire as $i => $date) {
+        if($date['date'] >= $balanceChangeToExpire['start_date'] && $date['date'] <= $balanceChangeToExpire['expiry_date']) {
+          if($remainingAmount >= abs($date['amount'])) {
+            // Date already deducted, so new we remove it from the
+            // array so it won't be deducted again from another
+            // balance change
+            unset($datesOverlappingBalanceChangesToExpire[$i]);
+            $remainingAmount += $date['amount'];
```
tunbola: There will be some instances when `$remainingAmount` would have become zero and the dates loop will still run for all the remaining dates even though the this condition `if($remainingAmount >= abs($date['amount']))` is there, it will still run through the whole date nonetheless to check that condition. We can short circuit by checking if `$remainingAmount = 0` and then escape the loop altogether and move to the next `$balanceChangeToExpire` in line.

PR #1361, opened 2017-01-20
PCHR-1878: Fix adding a contract revision causes the contract to be m…

bmnepali:
…arked as past

### Issue: 
On adding a revision on a contract with no date, It was marked as   "Past Contract". As on creating new revision, if no end date of a contract is found then it was kept in the list of past contracts.

### Solution:
If we do not have a end date of a contract, on creating a new revision, keeping the contract as "Current Contract" .i.e not as "Past Contract".

### Screenshots:
1. Adding a contract with no end date to a contact.
screen shot 2017-01-20 at 4 18 26 pm

2. Go to contract "change Terms" -> click -> change position -> to Creating New Revision with date to next day or current day.
screen shot 2017-01-20 at 4 20 34 pm

3. The contract should not be in the list of "Past contracts""
screen shot 2017-01-20 at 4 20 57 pm


hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -77,7 +77,7 @@ define([
               * @param {string || date} newEndDate the date specified by the user
             */
             function updateContractList(newEndDate){
-              var isCurrentContract = moment().diff(newEndDate, "day") <= 0;
+              var isCurrentContract = !newEndDate ? true : (moment().diff(newEndDate, "day") <= 0);
```
dhavalsharma: 1. check if !newEndDate will pass through for empty strings or empty objects, what are the various expected values of newEndDate?
2. also check if test cases are updated for this

PR #1370, opened 2017-01-24
PCHR-1839: Improvements to Sickness BAO validations

tunbola:
This PR improves the validations on SicknessRequest BAO. It adds an additional validation related to the is_sick flag on an Absence Type i.e

- SicknessRequests can no longer be created/updated for AbsenceTypes where is_sick is false

This PR also fixes a loophole in SicknessRequest.isValid API endpoint.

**Problem:**  

A sickness request is created from a combination of parameters needed to create a leave request and additional parameters specific to a sickness request. Before this PR, SicknessRequest.isValid only runs the SicknessRequest validation function only(which only runs validation for SicknessRequest specific parameters), the fallout of this is that if the validations pass and for some reason some required conditions for creating a leave request was not met, this gives a false assurance to whatever calls the SicknessRequest.isValid API that a subsequent call to create a SicknessRequest will go through.

**Solution:**

In addition to calling the SicknessRequest validation function, the LeaveRequest Validation function is also called inside the validation function of the SicknessRequest.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/SicknessRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/SicknessRequest.php
@@ -170,7 +170,12 @@ function civicrm_api3_sickness_request_isvalid($params) {
     $result[$e->getField()] = [$e->getExceptionCode()];
   }
 
-  return civicrm_api3_create_success($result);
+  $results =  civicrm_api3_create_success($result);
+  if (isset($results['id'])) {
+    unset($results['id']);
+  }
+
+  return $results;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/SicknessRequest.php
```diff
@@ -169,8 +169,16 @@ function civicrm_api3_sickness_request_isvalid($params) {
   catch (CRM_HRLeaveAndAbsences_Exception_InvalidSicknessRequestException $e) {
     $result[$e->getField()] = [$e->getExceptionCode()];
   }
+  catch (CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException $e) {
```
davialexandre: Since both exception extend `CRM_HRLeaveAndAbsences_Exception_EntityValidationException`, you only need one catch:

```php
   try {
     CRM_HRLeaveAndAbsences_BAO_SicknessRequest::validateParams($params);
   }
    catch (CRM_HRLeaveAndAbsences_Exception_EntityValidationException $e) {
      $result[$e->getField()] = [$e->getExceptionCode()];
    }
```

PR #1384, opened 2017-01-27
AI-6: Warning messages

meladawy:
The Problem :- 
Popup messages usually takes 2 seconds to be loaded via javascript, this makes the error message appear in the website for 2 seconds then javascript popup take the lead and hide the message from the page itself. 
 

Requirements :- 
Keep messages only appear in popup and hide it from the website itself 

Solution :- 
To make the message inside the page invisible, this will guarantee that messages will not show for 2 seconds until javascript to be loaded.


org.civicrm.bootstrapcivicrm/scss/civicrm/common/_notifications.scss
```diff
@@ -59,3 +59,14 @@
     padding: 15px;
   }
 }
+
+#crm-main-content-wrapper {
+  form {
+    .messages {
+      &.crm-error {
+        position: absolute;
```
omarabuhussein: is setting the position really important here ?

omarabuhussein:
can u also please post a pre and post gifs

org.civicrm.bootstrapcivicrm/scss/civicrm/common/_notifications.scss
meladawy: visibility:hidden; hide the messge but keep the height in the page. After 2 seconds of activating the popup messages the height disappear. Without absolute, users will see the height of the page changes within 2 seconds of loading the page. 

If i replace visibility: hidden with display:none; => javascript popup library will not be able to detect the message in the page. so the solution is prevent such height changes using absolute position

PR #1385, opened 2017-01-30
PCHR-1804: Disabled location region department level options still appear on job role dropdown

bmnepali:
### Issue:
Currently, all available options values  are fetched both enabled or disabled, which are displayed in dropdown options in job role form without filtering the option values with their activeness.

### Solution:
Modified, job role controller and job role service to fetch the option groups with their activeness indicator `is_active` attribute. Which is used in filtering the active options form the group and displayed in the dropdown options.

### Screenshots:
1. List of locations both enabled and disabled
screen shot 2017-01-30 at 1 12 25 pm

2. Displays only the enabled/active locations in job role form
screen shot 2017-01-30 at 1 14 12 pm

**Note**: Similarly, it works with departments, region, lavel as well.


com.civicrm.hrjobroles/js/src/job-roles/filters/option-group-filter.js
```diff
@@ -0,0 +1,19 @@
...
+  'job-roles/filters/filters'
+], function(filters) {
+  'use strict';
+
+  filters.filter('getActiveGroup', ['$log', function($log) {
+    $log.debug('Filter: getActiveGroup');
+
+    return function(groupData, param) {
+      var filteredGroupData = [];
+      angular.forEach(groupData, function(group, key) {
```
AkA84: Always leave a space between the `var` declarations and the rest of the code
```js
var filteredGroupData = []; 

angular.forEach(groupData, function (group, key) {
```

AkA84: Also, mind the whitespace. It's not
```js
angular.forEach(groupData, function(group, key) {
```
it's
```js
angular.forEach(groupData, function (group, key) {
```

AkA84: Also, please always use lodash instead of angular for utility functions like `forEach` on objects
```js
define([
  'common/lodash'
], function (_) {

  _.each(groupDate, function (group) {

  });
});

Comment (AkA84) [inline com.civicrm.hrjobroles/js/src/job-roles/filters/option-group-filter.js]: Finally, you are not using the `key` parameter, so don't declare it

File: com.civicrm.hrjobroles/js/src/job-roles/filters/option-group-filter.js [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+        if (group.is_active === param) {
+          filteredGroupData.push(group);
+        }
+      });
+      return filteredGroupData;
```
AkA84: Always leave an empty line before `return` if there is some code before it
```js
  if (group.is_active === param) {
    filteredGroupData.push(group);
  }
});

return filteredGroupData;
```

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
+define([
+    'common/angular',
```
AkA84: you're using 4-spaces indentation instead of 2-spaces indentation

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
...
+    'common/angularMocks',
+    'job-roles/filters/option-group-filter',
+    'job-roles/filters/filters',
```
AkA84: There is no need to declare this requirejs module here, it's already being declared inside `job-roles/filters/option-group-filter`

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
...
+], function(angular, moment) {
+  'use strict';
+
+  describe('Using getActiveGroup filter', function() {
```
AkA84: The outermost `describe` should contain the name of what you are testing
```js
describe('getActiveGroup', function () {

})
```

AkA84: Also, mind the whitespace. It's not
```js
describe('Using getActiveGroup filter', function () {
```
it's
```js
describe('Using getActiveGroup filter', function () {
```

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
...
+    var $filter, locations, getActiveGroup;
+
+    beforeEach(module('hrjobroles.filters'));
+    beforeEach(inject(['$filter', function(_$filter_) {
```
AkA84:
```js
function (_$filter_) {

File: com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js [hunk 7]
```diff
@@ -0,0 +1,59 @@
...
+        $filter = _$filter_;
```
AkA84: You are using 4-spaces indentation

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
...
+        getActiveGroup = $filter('getActiveGroup');
+      }
+    ]));
+
+    it('should define getActiveGroup filter ', function() {
```
AkA84: `it` titles should never contain "should", they must always contain statements in the imperative mood
```js
it('should do something', function () {}); // wrong
it('does something', function () {}); // correct
```
in your case it can simply be
```js
it('is defined', function () {});

File: com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+      expect($filter('getActiveGroup')).toBeDefined();
+    });
+
+    describe('When filtering locations, departments, region and lavels ', function() {
```
AkA84:
```js
function () {
```
please fix this in all the other instances

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,59 @@
...
+          }
+        }
+      });
+
+      it('should have 2 locations after filtration ', function() {
+        expect(getActiveGroup(locations, '1').length).toBe(2);
+      });
+
+      it('should have all data with is_active set to 1 ', function() {
+        expect(getActiveGroup(locations, '1')[0].is_active).toBe('1');
```
AkA84: To be honest I don't like this filter signature.

First of all the name. You're not picking the active "group", you're picking the active "value", aren't you? If that's the case, why isn't the filter then called `getActiveValues`?

Second, if the filter is explicitly about picking the active values, what's the need for the second parameter? You are always passing `1` anyway! And given the name of the filter (get**Active**Group), one would expect that the filter automatically picks the active values by default, without having to pass a parameter to specify that. I don't think the second parameter is required at all

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,58 @@
...
+
+    it('is defined', function () {
+      expect($filter('getActiveValues')).toBeDefined();
+    });
+
+    describe('when filtering locations, departments, region and lavels ', function () {
+      beforeEach(function () {
+        locations = {
+          "Headquarters": {
+            "id": "Headquarters",
```
AkA84: These are not really ids, i believe. Ids are supposed to be numbers, aren't they?

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
```diff
@@ -0,0 +1,58 @@
+define([
+  'common/angular',
+  'common/lodash',
```
AkA84: What do you need lodash for?

com.civicrm.hrjobroles/js/src/job-roles/filters/option-group-filter.js
```diff
@@ -0,0 +1,22 @@
...
+
+  filters.filter('getActiveValues', ['$log', function($log) {
+    $log.debug('Filter: getActiveValues');
+
+    return function(groupData) {
+      var filteredGroupData = [];
+
+      _.each(groupData, function (group) {
+        if (group.is_active === '1') {
+          filteredGroupData[group.id] = group;
```
AkA84: Why do you need an object instead of an array?

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
bmnepali: We wont need lodash, will remove.

com.civicrm.hrjobroles/js/test/filters/option-group-filter_test.js
bmnepali: Previously, `value` is kept `id`, but definitely they should be contain the value of `id`. I will change them.

com.civicrm.hrjobroles/js/src/job-roles/filters/option-group-filter.js
bmnepali: For API call to create new job role, for location, department, level and region, `id` are  no used rather than `value` attribute is used. So in select options in new/edit job role form, we need the option values to contain value as **contents of value property in a optionValue**. So, filter is required to return the key value pair objects. 

`  <select ng-options="k as v.title for (k, v) in jobroles.LocationsData | getActiveValues" ... ></select>`
screen shot 2017-01-31 at 4 37 03 pm

bmnepali:
Closing this PR in favor of refactoring and adding these changes again.

PR #1400, opened 2017-02-02
PCHR-1859: Fix Calculation of Estimated Annual Salary on Contract Revision

MiyaNoctem:
## Problem
Calculation of estimated annual salary was assuming 2080 hours per year always, without taking into account the actual number of hours configured in Hours tab.
before

## Solution
Fixed by using FTE fraction to calculate annual salary.
after


davialexandre:
The ticket for this has been closed, so I'm closing this as well

PR #1401, opened 2017-02-03
PCHR-1893: Add contact.js to load on contact page load

bmnepali:
### Issue:
Currently contact header information was not updated on creating  or deleting existing contract for the given contact. Which was due to not loading `contact.js` form **hrui** module which si responsible to update the contact header.

### Solution:
Added `contract.js` in `hrui.php` of **hrui** module  to load on contact page load.
`CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/contact.js');`

Added logic to _civicrm_hrjobcontract_api3_get_current_revision so that if no current revision is found, it searches for a future revision.

Included clauses to query on HRJobContract::getCurrentContract() to take into account revisions in the future and limited to current contracts.

Added an assertion to test use of HRJobContract::getCurrentContract() does not return null when there are current contracts defined.

### Screenshots:
1. Contract header gets updated on adding new contract for a contact.
add-header-change

2. Contract header gets updated on deleting new contract for a contact.
contact-header

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
+++ b/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
@@ -50,7 +50,7 @@ class CRM_Hrjobcontract_BAO_HRJobContract extends CRM_Hrjobcontract_DAO_HRJobCon
   public function delete($useWhere = false) {
       $id = $this->id;
       $result = parent::delete($useWhere);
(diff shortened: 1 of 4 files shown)


hitesh-jain:
Approved.

coldrunKacper:
Approved.

coldrunKacper:
Approved.

PR #1391, opened 2017-02-01
PCHR-1791: Add Leave and Absence General Settings Page

tunbola:
This PR creates the Leave and absence general settings page. The general settings page hinges on the core Setting API to create, save and retrieve settings. 

The only setting added in this PR is the 'Relationships that are allowed to approve a Leave Request' setting. This field/setting gets its values from the RelationshipType.get API endpoint. It is a multi select field that allows selection of one or more relationship type. The values for this field is saved in civicrm_setting table.

generalsettings

PR #1414, opened 2017-02-06
PCHR-1894: Build ACL filter using Leave Approver relationship types in settings

davialexandre:
This PR updates the `ACLQueryHelper` to build the ACL filter using the leave approver relationship types from the General Settings (https://github.com/civicrm/civihr/pull/1391). Before this, the filter would always use the 'has Leave Approved by' relationship.

The relationship types are stored in settings as a list of IDs. This IDs are added to the where clauses as `WHERE relationship_type.id IN(list, of, ids)`.

Testing code that uses the Setting API is a PAIN. You have to flush the cache/system on every test, making things really, really slow. In order to avoid that, I've created a `SettingsManager` interface and two implementations: `InMemorySettingsManager` and `APISettingsManager`. Together with that, there's a SettingsManager Factory, which is able to create instances of both, depending on a given param. Finally, using the civicrm container, I added a `hrleaveandabsences.settings_manager` service, which uses the factory and will always return a InMemorySettingsManager if we are in a test environment (if CIVICRM_UF == 'UnitTests').  The In Memory version doesn't call the API and hence there's no need to flush caches and everything.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/SettingsManager.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/SettingsManager.php
@@ -1,0 +1,16 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_InMemorySettingsManager as InMemorySettingsManager;
+use CRM_HRLeaveAndAbsences_Service_APISettingsManager as APISettingsManager;
+
+class CRM_HRLeaveAndAbsences_Factory_SettingsManager {
+
+  public static function create($inMemory = false) {
+    if ($inMemory) {
+      return new InMemorySettingsManager();
+    }
+
+    return new APISettingsManager();
+  }
+
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/SettingsManager.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/SettingsManager.php
@@ -3,8 +3,22 @@
 use CRM_HRLeaveAndAbsences_Service_InMemorySettingsManager as InMemorySettingsManager;
 use CRM_HRLeaveAndAbsences_Service_APISettingsManager as APISettingsManager;
 
+/**
+ * Creates CRM_HRLeaveAndAbsences_Service_SettingsManager instances
+ */
 class CRM_HRLeaveAndAbsences_Factory_SettingsManager {
 
+  /**
+   * Instantiate a new CRM_HRLeaveAndAbsences_Service_SettingsManager
+   * implementation and returns it.
+   *
+   * By default, it will create an APISettingsManager instance, but if $inMemory
+   * is true, an InMemorySettings instance will be returned
+   *
+   * @param bool $inMemory
+   *
+   * @return \CRM_HRLeaveAndAbsences_Service_SettingsManager
+   */
   public static function create($inMemory = false) {
     if ($inMemory) {
       return new InMemorySettingsManager();
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/APISettingsManager.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/APISettingsManager.php
@@ -2,8 +2,15 @@
 
 use CRM_HRLeaveAndAbsences_Service_SettingsManager as SettingsManager;
 
+/**
+ * A CRM_HRLeaveAndAbsences_Service_SettingsManager implementation that uses the
+ * CiviCRM Setting API to store and retrieve settings.
+ */
 class CRM_HRLeaveAndAbsences_Service_APISettingsManager implements SettingsManager {
 
+  /**
+   * {@inheritdoc}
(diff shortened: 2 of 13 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/SettingsManager.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_InMemorySettingsManager as InMemorySettingsManager;
+use CRM_HRLeaveAndAbsences_Service_APISettingsManager as APISettingsManager;
+
+class CRM_HRLeaveAndAbsences_Factory_SettingsManager {
+
```
tunbola: The method in the class is missing doc block

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/APISettingsManager.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_SettingsManager as SettingsManager;
+
+class CRM_HRLeaveAndAbsences_Service_APISettingsManager implements SettingsManager {
+
```
tunbola: The methods in this class are missing doc blocks

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/InMemorySettingsManager.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Service_SettingsManager as SettingsManager;
+
+class CRM_HRLeaveAndAbsences_Service_InMemorySettingsManager implements SettingsManager {
+
```
tunbola: Methods in this class are missing doc blocks

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -515,3 +515,21 @@ function _hrleaveandabsences_civicrm_post_absencetype($op, $objectId, &$objectRe
...
+ * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+ */
+function hrleaveandabsences_civicrm_container(\Symfony\Component\DependencyInjection\ContainerBuilder $container) {
+  $settingsManagerDefinition = new Symfony\Component\DependencyInjection\Definition(
+    CRM_HRLeaveAndAbsences_Service_SettingsManager::class
+  );
+  $settingsManagerDefinition->setFactoryClass(CRM_HRLeaveAndAbsences_Factory_SettingsManager::class);
+  $settingsManagerDefinition->setFactoryMethod('create');
+  // If we running unit tests, this will make the factory return an InMemorySettingsManager
+  $settingsManagerDefinition->setArguments([CIVICRM_UF == 'UnitTests']);
```
tunbola: Is there a way to add a tests for this so that we can always ensure that the right settings manager class was instantiated as per the environment?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SettingsManager.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+interface CRM_HRLeaveAndAbsences_Service_SettingsManager {
+  public function get($setting);
```
tunbola: missing doc blocks and we can probably add a brief description of what the interface does.
Maybe If the doc blocks are defined here, the other child classes implementing it can inherit that.

tunbola:
In the PR description
This sentence ending with 'depending on a given' is not complete.
s/InMemorySettingsManager if we in a test enviroment/InMemorySettingsManager if we are in a test environment

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeavePeriodEntitlementTest.php
```diff
@@ -423,6 +423,49 @@ public function testGetReturnsOnlyDataLinkedToContactsThatLoggedInUserManagesWhe
...
+      'approves leaves for',
+      'manages things for',
+    ]);
+
+    $manager1 = ContactFabricator::fabricate();
+    $manager2 = ContactFabricator::fabricate();
+    $contact1 = ContactFabricator::fabricate();
+    $contact2 = ContactFabricator::fabricate();
+
+    $this->setContactAsLeaveApproverOf($manager1, $contact2, null, null, true, 'manage things for');
```
tunbola: Is it possible to also add a test where a manager manages two contacts but with different relationships?
i.e manages contactA with 'approves leaves for' and contactB with 'manages things for'

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
davialexandre: There isn't. The problem is that this relies on the CIVICRM_UF constant, which is automatically set to 'UnitTests' when running tests. Since it's a constant, we cannot change its value and then there's no way to test the scenario where it has a different value.

PR #1416, opened 2017-02-07
PCHR-1806: Job role still visible when contract is removed until page reload

bmnepali:
### Issue:
As, Job roles module never knew about deleting existing contract, In job role tab job roles list  for a contract is still displayed until the page is reloaded.

### Solution:
A Pub-Sub service is created in reqangular module to facilitate communication between different angular module. Using it to communicate  between hrjobroles and hrjobcontract module to fetch the list of job roles and contracts on deleting the job contract. Using following steps:

1. Subscribing for `contract:deleted` event on job roles controller using `pubSub.subscribe('contract:deleted',...)` which will call function to fetch the latest contracts and  job roles for given contact.
2. Then, when deleting the contract successfully, publish a event as pubSub.publish('contract:deleted',data).

### Screenshots:
1. No job roles associated with a contract is shown in job roles tab on deleting a contract.
no jobroles


com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -1233,17 +1232,16 @@ define([
...
         contractIdsFromContact(contactId);
       });
 
-      // Triggers when a contract is deleted for a contact.
-      pubSub.subscribe('contract:deleted', function(data){
-        contractIdsFromContact(data.contactId);
-        jobRolesFromContracts(data.contractId);
+      pubSub.subscribe('contract:deleted', function(data) {
+        contractIdsFromContact(data.contactId).then(function(contractIds) {
+          jobRolesFromContracts(contractIds);
```
dhavalsharma: do we need to check if contractIds.length exists before calling `jobRolesFromContracts`

bmnepali: Yeah! its safe to check contract Ids before fetching the job roles. If we do not have contact ids. we don't have to job roles, which doesn't exist.

com.civicrm.hrjobroles/js/src/job-roles/controllers/hr-job-roles-controller.js
```diff
@@ -1256,16 +1256,22 @@ define([
...
       pubSub.subscribe('contract:created', function(contactId){
         contractIdsFromContact(contactId);
       });
 
       // Triggers when a contract is deleted for a contact.
-      pubSub.subscribe('contract:deleted', function(contactId){
-        contractIdsFromContact(contactId);
+      pubSub.subscribe('contract:deleted', function(data) {
+        contractIdsFromContact(data.contactId).then(function(contractIds) {
+          if(!contractIds.length){
```
dhavalsharma: can `contractIds` be null or undefined?

PR #1418, opened 2017-02-07
PCHR-1798: Use Leave Approver relationship types for the managed_by param

davialexandre:
When the `managed_by` param is present in a call to `LeaveRequest.get` or `LeaveRequest.getFull`, the `LeaveRequestSelect` class changes the database query to include filters to return only the leave requests of contacts managed by the contact with the ID given by `managed_by`.

`LeaveRequestSelect` was only filtering this by the 'has Leave Approved by' relationship type, though. Now that the extensions allows the configuration of multiple relationship types as "Leave Approvers", the class was updated to fetch the types from the settings and use them to build the query.

All the tests for the `managed_by` param were updated to also make sure they will work for both `get` and `getFull` endpoints.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -92,9 +92,10 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
     if(!empty($this->params['managed_by'])) {
       $managerID = (int)$this->params['managed_by'];
       $today =  '"' . date('Y-m-d') . '"';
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -300,4 +301,14 @@ private function getToilLeaveRequests($result) {
...
     return $toilLeaveRequestIDs;
   }
+
+  /**
+   * Returns a list of relationship types stored on the
+   * 'relationship_types_allowed_to_approve_leave' setting
+   *
+   * @return array
+   */
+  private function getLeaveApproverRelationshipTypes() {
```
tunbola: Can we extract this function to a class seeing that the function is also duplicated in ACLQueryHelper.php class?
Will be easy for us to make changes if its in a central place.

PR #1425, opened 2017-02-08
PCHR-1835: Change LeaveManager Service to use relationship types in general settings

davialexandre:
This PR updates the `LeaveManager.isContactManagedBy` method to use the leave approver relationship types in general settings in order to check if the a contact is managed by another. Before this, it would only consider the 'has Leave Approved by' relationship type as the only one valid for a leave approver.

The PR also includes some overall refactorings:
- Lots of removed duplication on the LeaveManager tests
- A SettingsManagerTrait was added in order to centralize the code to fetch the leave approver relationship types from settings
- The ACLQueryHelper was replaced by a trait, so it can be more easily reused in BAOs. Also, is make possible for this trait to use another trait (the SettingsManagerTrait) and avoid some code duplication

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/ACL/LeaveApproverTrait.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/ACL/LeaveApproverTrait.php
@@ -1,24 +1,21 @@
 <?php
 
-/**
- * Access Control Lists Custom Query generator class
- *
- * This class is basically an Utility class used to generate queries needed for overriding addSelectWhereClause
- * for some specific BAO's
- *
- */
-class CRM_HRLeaveAndAbsences_API_Query_ACLQueryHelper {
+trait CRM_HRLeaveAndAbsences_ACL_LeaveApproverTrait {
+
+  use CRM_HRLeaveAndAbsences_Service_SettingsManagerTrait;
 
   /**
-   * This method returns a query string that limits contacts records that are available to a logged in user
-   * by taking into account user relationships. It checks whether the logged in user is a leave approver
-   * to some contacts or not. If yes the logged in user has access to contact records for the contacts/users he manages,
-   * If not the logged is limited to his/her own records only.
+   * This method returns a query string that limits contacts records that are
+   * available to a logged in user by taking into account user relationships.
+   * It checks whether the logged in user is a leave approver to some contacts
+   * or not. If yes the logged in user has access to contact records for the
+   * contacts/users he manages. If not the logged is limited to his/her own
+   * records only.
    *
    * @return string
    *   Query String
    */
-  public static function limitContactsByTakingUserRelationShipsIntoAccount() {
+  public function getLeaveApproverACLClauses() {
     $contactsTable = CRM_Contact_BAO_Contact::getTableName();
     $relationshipTable = CRM_Contact_BAO_Relationship::getTableName();
     $relationshipTypeTable = CRM_Contact_BAO_RelationshipType::getTableName();
@@ -28,7 +25,8 @@ class CRM_HRLeaveAndAbsences_API_Query_ACLQueryHelper {
     $conditions = [];
     $conditions[] = "(c.id = {$loggedInUserID})";
 
-    $leaveApproverRelationships = self::getLeaveApproverRelationships();
+    $leaveApproverRelationships = $this->getLeaveApproverRelationshipsTypes();
+
     if(!empty($leaveApproverRelationships)) {
       $conditions[] = "(
         r.is_active = 1 AND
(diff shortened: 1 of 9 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/ACL/LeaveApproverTrait.php
```diff
@@ -1,24 +1,21 @@
...
+   * It checks whether the logged in user is a leave approver to some contacts
+   * or not. If yes the logged in user has access to contact records for the
+   * contacts/users he manages. If not the logged is limited to his/her own
+   * records only.
    *
    * @return string
    *   Query String
    */
-  public static function limitContactsByTakingUserRelationShipsIntoAccount() {
+  public function getLeaveApproverACLClauses() {
```
tunbola: This function name makes me think that the query only takes the leave approver into consideration whereas it also takes the leave contact into consideration too. i.e it checks based on whether its a leave approver or a leave contact. Perhaps we can give it a better name?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
```diff
@@ -25,109 +29,109 @@ public function tearDown() {
...
-    $staffMember = ContactFabricator::fabricate();
+    $this->assertFalse($this->leaveManagerService->isContactManagedBy($this->staffMember['id'], $this->manager['id']));
+
+    $this->setContactAsLeaveApproverOf($this->manager, $this->staffMember, null, null, false);
+
+    $this->assertFalse($this->leaveManagerService->isContactManagedBy($this->staffMember['id'], $this->manager['id']));
+  }
 
-    $this->assertFalse($this->leaveManagerService->isContactManagedBy($staffMember['id'], $manager['id']));
+  public function testIsContactManagedByWhenCurrentUserHasMoreThanOneRelationshipWithTheEmployee() {
```
tunbola: The test name is a bit misleading. The test actually has nothing to do with the current user. The `isContactManagedBy` takes two contact parameters irrespective of if any of the contact is the current/loggen in user or not.
Also the phrase `HasMoreThanOneRelationshipWithTheEmployee` makes one think its possible for the manager to have more than one relationship with the employee at a given time which seems not possible?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -13,6 +13,8 @@
  */
 class CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement extends CRM_HRLeaveAndAbsences_DAO_LeavePeriodEntitlement {
 
+  use CRM_HRLeaveAndAbsences_ACL_LeaveApproverTrait;
+
```
tunbola: We should remove the reference to `CRM_HRLeaveAndAbsences_API_Query_ACLQueryHelper as ACLQueryHelper` at the beginning of this file since its no longer necessary.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -13,6 +13,8 @@
 
 class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO_LeaveRequest {
 
+  use CRM_HRLeaveAndAbsences_ACL_LeaveApproverTrait;
+
```
tunbola: We should also remove the reference to `CRM_HRLeaveAndAbsences_API_Query_ACLQueryHelper as ACLQueryHelper` at the beginning of this file since its no longer necessary.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveManagerTest.php
davialexandre: It's possible to have more than one active relationship at the same time. I added a new test to cover this scenario

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/ACL/LeaveApproverTrait.php
```diff
@@ -1,24 +1,28 @@
...
 
-/**
- * Access Control Lists Custom Query generator class
- *
- * This class is basically an Utility class used to generate queries needed for overriding addSelectWhereClause
- * for some specific BAO's
- *
- */
-class CRM_HRLeaveAndAbsences_API_Query_ACLQueryHelper {
+trait CRM_HRLeaveAndAbsences_ACL_LeaveApproverTrait {
```
tunbola: We should probably change the name of this trait also to `CRM_HRLeaveAndAbsences_ACL_LeaveInformationTrait`

PR #1427, opened 2017-02-08
PCHR-1858: Implement Management of Insurance Plan Types

MiyaNoctem:
## Problem
Insurance plan types were hardcoded and were not manageable from UI.

## Solution
Included ugrade to create option group and option values for insurance plan types. Also, changed implementation of CRM_Hrjobcontract_SelectValues::planType() method to obtain values from option values in database.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php
@@ -557,14 +557,6 @@ class CRM_Hrjobcontract_Import_Parser_Api extends CRM_Hrjobcontract_Import_Parse
   private function validateField($key, $value)  {
     $errorMessage = NULL;
     $convertedValue = $value;
-    $optionReturnType = 'id';
-    // TODO: all values should be stored by values and this part should be removed
-    if ( in_array($key, array('HRJobPension-pension_type', 'HRJobDetails-contract_type')) )  {
-      $optionReturnType = 'label';
-    }
-    elseif ( in_array($key, array('HRJobHour-hours_type', 'HRJobPay-pay_cycle', 'HRJobDetails-location', 'HRJobDetails-end_reason')) )  {
-      $optionReturnType = 'value';
-    }
 
     switch($key)  {
       case 'HRJobContractRevision-jobcontract_id':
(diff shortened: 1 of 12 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1021,6 +1021,49 @@ function upgrade_1023() {
     return true;
   }
 
+  /**
+   * Creates Option Group and Option Values for Health an Life Insurance Plan 
+   * Types.
+   * 
+   * @return boolean TRUE
+   */
+  public function upgrade_1025() {
```
omarabuhussein: also can you rename the method to upgrade_1024 since there is no upgrader here with that number instead of jumping to 1025

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1021,6 +1021,49 @@ function upgrade_1023() {
...
+    $this->installInsurancePlanTypes();
```
omarabuhussein: maype this could be renamed to createInsurePlanTypes  since install keyword is rarely used for method names that create sutff in civihr

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1021,6 +1021,49 @@ function upgrade_1023() {
...
+    return TRUE;
+  }
+
+  /**
+   * Creates Option Group for Insurance Plan Types.
+   */
+  public function installInsurancePlanTypes() {
+    try {
+      civicrm_api3('OptionGroup', 'create', [
+        'sequential' => 1,
```
omarabuhussein: no need for 'sequential' => 1, in the API calls inside this method

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
MiyaNoctem: I tried using upgrade_1024, but for some reason it would not get identified as a pending upgrade with this method name, so I just bumped it to 1025, where it started to work.  I saw Davi ran into a similar problem, but couldn't understand why it was happening.

https://github.com/civicrm/civihr/commit/cbe0b5b6c0a6caca78700387b500f18cd5a4cd5f

hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php
```diff
@@ -597,15 +589,15 @@ private function validateField($key, $value)  {
       case 'HRJobPay-pay_cycle':
       case 'HRJobDetails-location':
       case 'HRJobDetails-end_reason':
-        $optionID = $this->getOptionID($key, $value, $optionReturnType);
+      case 'HRJobHealth-plan_type':
+      case 'HRJobHealth-plan_type_life_insurance':
+        $optionID = $this->convertOptionValue($key, $value);
```
omarabuhussein: I feel disgusting each time I look  at the job contract importer code which I significantly updated  on my first days at compucorp , and keep saying what the hell I was doing :smile:  .. thanks for making it a bit better :D

PR #1420, opened 2017-02-07
PCHR-1836: Modify Leave Request Service

tunbola:
This PR actually modifies the existing LeaveRequest Service. This create method of this class is a wrapper around the LeaveRequest::create BAO method. This PR adds some checks/validations to the already existing LeaveRequest Service create method. 

The validations are:
- Status can only be changed according to the LeaveRequestStatusMatrix: [#1399](https://github.com/civicrm/civihr/pull/1399)
- Status can only be updated only if the user has the permission required for the status  according to the LeaveRequestRights [#1407](https://github.com/civicrm/civihr/pull/1407)
- Managers cannot change the dates or absence type of their managees leave requests

PR #1429, opened 2017-02-09
PCHR-1838: Modify SicknessRequest Service

tunbola:
This PR modifies the already existing SicknessRequest Service by making the it to extend LeaveRequest Service [#1420](https://github.com/civicrm/civihr/pull/1420). 

Some necessary methods that needed a different implementation in SicknessRequest Service child class  were overridden and modified.

One of such is the fact that Leave Approvers and Admins can modify SicknessRequest dates on behalf of the Leave Contact, this was not possible in LeaveRequest Service. So the `canChangeDatesFor` function of Sickness Request service was added as an overridden function of the parent class to reflect this change.

Also the way a leave request is created and deleted is a bit different from that of SicknessRequest, this also was taken into consideration and the methods required for this in the parent Leave Request Service class was overridden and modified in the child Sickness Request Service class.

Hopefully, this is a better way than reinventing the wheel and duplicating the methods of the LeaveRequest Service in the SicknessRequest Service.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
@@ -62,6 +62,10 @@ class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends LeaveRequestService
    * {@inheritDoc}
    */
   protected function runValidation($params) {
+    if (isset($params['id'])) {
+      $params['leave_request_id'] = $this->getOldLeaveRequest($params['id'])->id;
+    }
+
     SicknessRequest::validateParams($params);
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
@@ -4,7 +4,7 @@ use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_SicknessRequest as SicknessRequest;
 use CRM_HRLeaveAndAbsences_Service_LeaveRequest as LeaveRequestService;
 
-class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends CRM_HRLeaveAndAbsences_Service_LeaveRequest {
+class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends LeaveRequestService {
 
   /**
    * {@inheritDoc}
@@ -21,7 +21,7 @@ class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends CRM_HRLeaveAndAbsen
    *
    * @return \CRM_HRLeaveAndAbsences_BAO_SicknessRequest|NULL
    */
-  protected function createLeaveRequestWithBalanceChange($params) {
+  protected function createRequestWithBalanceChanges($params) {
     $sicknessRequest = SicknessRequest::create($params, false);
     $leaveRequest = LeaveRequest::findById($sicknessRequest->leave_request_id);
     $this->leaveBalanceChangeService->createForLeaveRequest($leaveRequest);
@@ -56,5 +56,12 @@ class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends CRM_HRLeaveAndAbsen
     $sicknessRequest = SicknessRequest::findById($sicknessRequestID);
     parent::delete($sicknessRequest->leave_request_id);
     $sicknessRequest->delete();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  protected function runValidation($params) {
+    SicknessRequest::validateParams($params);
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
@@ -2,44 +2,47 @@
 
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_SicknessRequest as SicknessRequest;
-use CRM_HRLeaveAndAbsences_Service_LeaveBalanceChange as LeaveBalanceChangeService;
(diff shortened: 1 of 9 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
```diff
@@ -2,47 +2,50 @@
 
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_SicknessRequest as SicknessRequest;
-use CRM_HRLeaveAndAbsences_Service_LeaveBalanceChange as LeaveBalanceChangeService;
 use CRM_HRLeaveAndAbsences_Service_LeaveRequest as LeaveRequestService;
 
-class CRM_HRLeaveAndAbsences_Service_SicknessRequest {
+class CRM_HRLeaveAndAbsences_Service_SicknessRequest extends CRM_HRLeaveAndAbsences_Service_LeaveRequest {
```
davialexandre: you can use the class alias here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/SicknessRequestTest.php
```diff
@@ -23,12 +23,20 @@ class SicknessRequestTest extends BaseHeadlessTest {
   use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
   use CRM_HRLeaveAndAbsences_LeaveBalanceChangeHelpersTrait;
 
+  use CRM_HRLeaveAndAbsences_SessionHelpersTrait;
```
davialexandre: put this statement together with the other traits on this class

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
```diff
@@ -2,47 +2,50 @@
...
    *
    * @param array $params
-   * @param bool $validate
    *
-   * @return \CRM_HRLeaveAndAbsences_BAO_SicknessRequest
+   * @return \CRM_HRLeaveAndAbsences_BAO_SicknessRequest|NULL
    */
-  public function create($params, $validate = true) {
-    $sicknessRequest = SicknessRequest::create($params, $validate);
+  protected function createLeaveRequestWithBalanceChange($params) {
```
davialexandre: maybe we should rename this method to `createWithBalanceChanges` or `createRequestWithBalanceChanges`, because it doesn't create only the leave request, but also the sickness request

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
```diff
@@ -2,47 +2,50 @@
...
+    $sicknessRequest = SicknessRequest::create($params, false);
```
davialexandre: what about the call `SicknessRequest::validateParams()`? I don't see it anywhere, but we need it

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/SicknessRequest.php
```diff
@@ -51,16 +54,18 @@ public function create($params, $validate = true) {
...
+  /**
+   * {@inheritDoc}
+   */
+  protected function runValidation($params) {
+    if (isset($params['id'])) {
+      $params['leave_request_id'] = $this->getOldLeaveRequest($params['id'])->id;
     }
-  }
 
+    SicknessRequest::validateParams($params);
```
davialexandre: `validatesParams` throws exceptions, so we need to add this to the dockblock

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/SicknessRequestTest.php
```diff
@@ -3,6 +3,9 @@
 use CRM_HRLeaveAndAbsences_BAO_SicknessRequest as SicknessRequest;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_Test_Fabricator_AbsenceType as AbsenceTypeFabricator;
+use CRM_Hrjobcontract_Test_Fabricator_HRJobContract as HRJobContractFabricator;
```
davialexandre: Just as a matter of organization, can you please put all the classes for L&A together? Something like:

use other_extesion_class;
use some_other_extension_class;
use all_the_leave_and_absences_classes;

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/SicknessRequestTest.php
```diff
@@ -207,4 +219,60 @@ public function testValidateParamsCallsLeaveRequestValidateParams() {
       'required_documents' => $requiredDocuments
     ]);
   }
+
+  public function testCreateAndUpdateSicknessRequest() {
```
davialexandre: Is this test really necessary? We don't do any special handling for reason or required_documents, so there's nothing that needs to be tested. I understand you added this because of the possible bug reported by Dhaval, but since there isn't really a bug (at least, not one we could reproduce), then no test is needed I would say.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/SicknessRequestTest.php
```diff
@@ -13,13 +16,22 @@ class CRM_HRLeaveAndAbsences_BAO_SicknessRequestTest extends BaseHeadlessTest {
...
     CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
 
     $this->requiredDocumentOptions = $this->getSicknessRequestRequiredDocumentsOptions();
     $this->leaveRequestDayTypes = $this->getLeaveRequestDayTypes();
     $this->sicknessRequestReasons = $this->getSicknessRequestReasons();
+    $this->leaveRequestStatuses = $this->getLeaveRequestStatuses();
+
+    $this->leaveContact = 1;
+    $this->registerCurrentLoggedInContactInSession($this->leaveContact);
+    CRM_Core_Config::singleton()->userPermissionClass->permissions = [];
```
davialexandre: Is this really necessary?

PR #1430, opened 2017-02-09
PCHR-1833: Set "Has Leave Approved By" as the default Leave Approver relationship type

davialexandre:
With this PR, the L&A extension will now have a default Leave Approver relationship type right after it gets installed. The new relationship type is named "Has Leave Approved By" and it's the default value on the General Settings.

The enable/disable/install/uninstall hooks were all updated to make sure the RelationshipType gets updated properly on each of this events. Originally, the idea was to use Managed Entities to create the relationship type but was not possible. The reason is that the `hrabsence` extension also creates the same relationship type. So, if that extension is installed before l&a, the managed entity creation will fail because the type already exists. Using hook, we can first check if it exists and only create it if it doesn't.

Due to the fact that now the extensions creates the relationship type and it has the same name as the default one created by the `LeaveManagerHelpersTrait`, some tests started to fail (because the relationship already existed and the trait would set it as a leave approver relationship type). I changed the relationship type name used by the trait in order to fix this.

### Other changes includes with this PR
- Refactoring and reorganization of the code inside hrleaveandabsences.php
- Fix the positioning of the "Leave and Absences" menu item position on the main navigation menu. It used the Reports menu as a reference, but it doesn't exist anymore, so the Vacancies menu is now the reference.
- Use managed entities to handle the scheduled jobs

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
@@ -354,7 +354,7 @@ function _hrleaveandabsences_create_administer_menu() {
   $administerMenuId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'Administer', 'id', 'name');
   $maxWeightOfAdminMenuItems = _hrleaveandabsences_get_max_child_weight_for_menu($administerMenuId);
 
-  $params = array(
+  $params = [
     'label'      => ts('Leave and Absences'),
     'name'       => 'leave_and_absences',
     'url'        => null,
@@ -363,7 +363,7 @@ function _hrleaveandabsences_create_administer_menu() {
     'parent_id'  => $administerMenuId,
     'weight'     => $maxWeightOfAdminMenuItems + 1,
     'permission' => 'administer leave and absences'
-  );
+  ];
 
   $leaveAndAbsencesAdminNavigation = _hrleaveandabsences_add_navigation_menu($params);
 
@@ -425,9 +425,9 @@ function _hrleaveandabsences_create_administer_menu_tree($leaveAndAbsencesAdminN
  */
 function _hrleaveandabsences_get_max_child_weight_for_menu($menu_id) {
   $query = "SELECT MAX(weight) AS max FROM civicrm_navigation WHERE parent_id = %1";
-  $params = array(
-    1 => array($menu_id, 'Integer')
-  );
+  $params = [
+    1 => [$menu_id, 'Integer']
+  ];
   $dao = CRM_Core_DAO::executeQuery($query, $params);
   $dao->fetch();
   if($dao->max) {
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -517,19 +538,91 @@ function _hrleaveandabsences_civicrm_post_absencetype($op, $objectId, &$objectRe
...
+ * @return mixed|null
+ */
+function _hrleaveandabsences_get_has_leave_approved_by_relationship_type() {
+  $result = civicrm_api3('RelationshipType', 'get', [
+    'sequential' => 1,
+    'name_a_b' => 'has Leave Approved by',
+  ]);
+
+  if (!empty($result['values'])) {
+    return array_shift($result['values']);
```
tunbola: Perhaps a small comment/note as to why `array_unshift` is needed here.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -477,6 +316,188 @@ function hrleaveandabsences_civicrm_post($op, $objectName, $objectId, &$objectRe
...
+    'label'      => ts('Leave and Absences'),
+    'name'       => 'leave_and_absences',
+    'url'        => null,
+    'operator'   => null,
+    'is_active'  => 1,
+    'parent_id'  => $administerMenuId,
+    'weight'     => $maxWeightOfAdminMenuItems + 1,
+    'permission' => 'administer leave and absences'
+  );
+
```
tunbola: please use short array syntax here.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -477,6 +316,188 @@ function hrleaveandabsences_civicrm_post($op, $objectName, $objectId, &$objectRe
...
+  $params = array(
+    'label'      => ts('Leave and Absences'),
+    'name'       => 'leave_and_absences',
+    'url'        => 'civicrm/leaveandabsences/dashboard',
+    'operator'   => null,
+    'weight'     => $vacanciesWeight + 1,
+    'is_active'  => 1,
+    'permission' => 'access leave and absences'
+  );
+
```
tunbola: short array syntax here also

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -477,6 +316,188 @@ function hrleaveandabsences_civicrm_post($op, $objectName, $objectId, &$objectRe
...
+ *
+ * @param $menu_id
+ *
+ * @return int
+ */
+function _hrleaveandabsences_get_max_child_weight_for_menu($menu_id) {
+  $query = "SELECT MAX(weight) AS max FROM civicrm_navigation WHERE parent_id = %1";
+  $params = array(
+    1 => array($menu_id, 'Integer')
+  );
```
tunbola: short array syntax here also

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
davialexandre: tbh, array_shift is not really needed. I'll fix this

PR #1439, opened 2017-02-13
PCHR-1681: Fix the saving of balance changes for half days

davialexandre:
The LeaveRequest.calculateBalanceChange returns a list of dates for a leave request and the amount of days to be deducted for each one of them. It accepts the from and to dates as params, as well as the types for these dates (if it's a half day, weekend, working day or non-working day). The types are values from the `hrleaveandabsences_leave_request_day_type` Option Group. Internally, this method was working based on the Option Value name instead of the Option Value value. That is, it was expecting to receive the something like 'half_day_am' instead of "2" (the value for the option "1/2 AM").

A few places of the code (mainly, the frontend part) was passing the name to it (using the API) and other (the backend) was passing values, and this was making the results inconsistent. When calling the method through the API, it would return the correct value for half days (0.5), but when called from any backend code, which would pass the value instead of the name, it would return the wrong amount of days, since it didn't know to which option the received value refer to.

This is how it was fixed:
- The BAO LeaveBalanceChange.calculateBalanceChange method was fixed to work based on the Option Value value instead of the names
- The API spec for the LeaveBalanceChange.calculateBalanceChange API was updated. The to_type and from_type fields were renamed to to_date_type and from_date_type, to keep consistency with the LeaveRequest fields. Their spec was updated to say they are option values. With this, the API layer will automatically convert a given Option Value name to an Option Value value, which will then be passed to the BAO method
- All the frontend code making references to the old field names was updated. Some mapping code, put in place to map from_date_type and to_date_type to from_type and to_type was removed.

Finally, I also did a small cleanup and refactor on calculateBalanceChange.


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Date/BasicDatePeriod.php
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * @param mixed $start
+   *   Can either be a DateTimeInterface instance of any value acceptable by the
+   *   DateTimeImmutable constructor
+   * @param mixed $end
+   *   Can either be a DateTimeInterface instance of any value acceptable by the
+   *   DateTimeImmutable constructor
+   */
+  public function __construct($start, $end) {
+    if($start instanceof DateTimeInterface) {
```
tunbola: Is there any reason why we are not checking for `instanceof DateTime` here rather than  `DateTimeInterface` since most likely a DateTimeImmutable object will probably not be passed.
Then we can use the object passed directly if its a date time rather than instantiating a new DateTime object for that purpose?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Date/BasicDatePeriodTest.php
```diff
@@ -0,0 +1,65 @@
...
+
+    $this->assertCount(5, $dates);
+    $this->assertEquals('2016-01-01', $dates[0]->format('Y-m-d'));
+    $this->assertEquals('2016-01-02', $dates[1]->format('Y-m-d'));
+    $this->assertEquals('2016-01-03', $dates[2]->format('Y-m-d'));
+    $this->assertEquals('2016-01-04', $dates[3]->format('Y-m-d'));
+    $this->assertEquals('2016-01-05', $dates[4]->format('Y-m-d'));
+  }
+
+  public function testItDoesNotModifyTheEndDate() {
```
tunbola: Don't think we actually need this test. The `testItIncludesTheEndDate` actually checks all the dates and confirms that none was modified.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -545,21 +546,7 @@ public function getDates() {
...
-      $endDate = new DateTime($this->to_date);
-    }
-
-    // We need to add 1 day to the end date to include it
-    // when we loop through the DatePeriod
-    $endDate->modify('+1 day');
-
-    $interval   = new DateInterval('P1D');
-    $datePeriod = new DatePeriod($startDate, $interval, $endDate);
+    $datePeriod = new BasicDatePeriod($this->from_date, $this->to_date ?: $this->from_date);
```
tunbola: The from_date and to_date parameter is now compulsory for a LeaveRequest, so we can do away with the ternary operator to check if to_date is present or not.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -582,72 +569,70 @@ private function deleteDates() {
...
-   * @param string $fromType
+   * @param string $fromDateType
    * @param null|\DateTime $toDate
-   * @param null|string $toType
+   * @param null|string $toDateType
    *
    * @return array
    *   An array of formatted results
    */
-  public static function calculateBalanceChange($contactId, DateTime $fromDate, $fromType, DateTime $toDate = null, $toType = null) {
```
tunbola: The to_date and to_date_type is now required on the BAO. I think it should apply on this method also?
If then, we should probably update the parameters to required on the `LeaveRequest.calculateBalanceChange` endpoint and also on this method.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Date/BasicDatePeriod.php
davialexandre: The reasons are:
- There's no reason to limit the method to only DateTime objects
- Creating a new object is important to avoid changing the values of the given objects. Here is what happens if we don't create new objects: https://3v4l.org/UBvr2

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Date/BasicDatePeriodTest.php
davialexandre: Maybe the name of the method is not clear enough. What I'm testing two different aspects here:

- On testItIncludesTheEndDate I'm asserting the the given $endDate will be included on the list of dates.
- On testItDoesNotModifyTheEndDate I'm asserting that the given $endDate is not modified by the class. Check this comment for more reference https://github.com/civicrm/civihr/pull/1439#discussion_r100775750

Maybe I should rename this method to something like `testItDoesNotModifyThe**Given**EndDate`

PR #1445, opened 2017-02-13
PCHR-1837: Create TOILRequest Service

tunbola:
This PR creates the TOILRequest Service by making the it to extend LeaveRequest Service #1420.

Some necessary methods that needed a different implementation in TOILRequest Service child class were overridden and modified.

One of such is the fact that a TOILRequest creates its own balance changes, so the `createRequestWithBalanceChanges` method of the TOILRequest service was created to override that of the parent LeaveRequest Service class.

Also a bug that prevents a TOIL request from being updated due to an `overlapping leave request` error was fixed in this PR.

The methods for Creating and Deleting LeaveRequestBalance Changes were removed completely from the TOILRequest BAO and moved to the LeaveBalanceChange Service. Also TOILRequest Fabricator was updated to create its balance changes when needed.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
@@ -54,19 +54,9 @@ class CRM_HRLeaveAndAbsences_Service_TOILRequest extends LeaveRequestService {
    */
   public function delete($toilRequestID) {
     $toilRequest = TOILRequest::findById($toilRequestID);
-    $leaveRequest = LeaveRequest::findById($toilRequest->leave_request_id);
-
-    $transaction = new CRM_Core_Transaction();
-    try {
-      $this->leaveBalanceChangeService->deleteForTOILRequest($toilRequest);
-      LeaveRequestDate::deleteDatesForLeaveRequest($toilRequest->leave_request_id);
-      $leaveRequest->delete();
-      $toilRequest->delete();
-
-      $transaction->commit();
-    } catch(Exception $e) {
-      $transaction->rollback();
-    }
+    parent::delete($toilRequest->leave_request_id);
+    $this->leaveBalanceChangeService->deleteForTOILRequest($toilRequest);
+    $toilRequest->delete();
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
@@ -1,0 +1,82 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
+use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequest as LeaveRequestService;
+
+class CRM_HRLeaveAndAbsences_Service_TOILRequest extends LeaveRequestService {
+
+  /**
+   * This method creates/updates the TOILRequest along with the
+   * associated LeaveRequest and it's LeaveBalanceChanges
+   *
+   * @param array $params
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_TOILRequest|NULL
+   */
+  protected function createRequestWithBalanceChanges($params) {
+    $toilRequest = TOILRequest::create($params, false);
+    $expiryDate = !empty($params['expiry_date']) ? new DateTime($params['expiry_date']) : null;
+
+    $this->leaveBalanceChangeService->createForTOILRequest(
+      $toilRequest,
+      $params['type_id'],
+      $params['toil_to_accrue'],
+      $expiryDate
+    );
+
+    return $toilRequest;
+  }
+
+  /**
+   * Returns the LeaveRequest object associated with the TOILRequest
+   * in its current state (i.e before it gets updated)
+   *
+   * @param int $toilRequestID
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest
+   */
+  protected function getOldLeaveRequest($toilRequestID) {
+    $toilRequest = TOILRequest::findById($toilRequestID);
+    if (!$this->oldLeaveRequest) {
+      $this->oldLeaveRequest = LeaveRequest::findById($toilRequest->leave_request_id);
+    }
+    return $this->oldLeaveRequest;
+  }
+
+  /**
+   * Deletes the TOILRequest with the given $toilRequestID, including
+   * its associated LeaveRequest and all of its LeaveRequestDates and
+   * TOIL LeaveBalanceChanges
+   *
+   * @param int $toilRequestID
+   */
+  public function delete($toilRequestID) {
+    $toilRequest = TOILRequest::findById($toilRequestID);
+    $leaveRequest = LeaveRequest::findById($toilRequest->leave_request_id);
+
+    $transaction = new CRM_Core_Transaction();
+    try {
+      $this->leaveBalanceChangeService->deleteForTOILRequest($toilRequest);
+      LeaveRequestDate::deleteDatesForLeaveRequest($toilRequest->leave_request_id);
+      $leaveRequest->delete();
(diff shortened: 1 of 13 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChangeTest.php
```diff
@@ -15,6 +18,15 @@
  */
 class CRM_HRLeaveAndAbsences_Service_LeaveBalanceChangeTest extends BaseHeadlessTest {
 
+  use CRM_HRLeaveAndAbsences_LeaveBalanceChangeHelpersTrait;
+
+  private $leaveBalanceChangeService;
+
+  public function setUp() {
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
```
davialexandre: use single quotes here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/TOILRequestTest.php
```diff
@@ -0,0 +1,150 @@
...
+  use CRM_HRLeaveAndAbsences_TOILRequestHelpersTrait;
+
+  private $leaveContact;
+
+  private $absenceType;
+
+  private $leaveBalanceChangeService;
+
+  public function setUp() {
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
```
davialexandre: single quotes

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public function delete($toilRequestID) {
+    $toilRequest = TOILRequest::findById($toilRequestID);
+    $leaveRequest = LeaveRequest::findById($toilRequest->leave_request_id);
+
+    $transaction = new CRM_Core_Transaction();
+    try {
+      $this->leaveBalanceChangeService->deleteForTOILRequest($toilRequest);
+      LeaveRequestDate::deleteDatesForLeaveRequest($toilRequest->leave_request_id);
+      $leaveRequest->delete();
```
davialexandre: instead of this, It would probably be better to call `parent::delete($toilRequest->leave_request_id)` to delete the leave request and the dates

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/TOILRequestTest.php
```diff
@@ -0,0 +1,150 @@
...
+      'allow_accruals_request' => true,
+      'max_leave_accrual' => 10
+    ]);
+
+    $params['id'] = $toilRequest->id;
+    $params['type_id'] = $absenceType->id;
+
+    $toilRequestService = $this->getTOILRequestServiceWhenCurrentUserIsLeaveManager();
+    $toilRequestService->create($params, false);
+  }
```
davialexandre: I'm not sure if this method is really necessary, since the behavior is the same as the LeaveRequest Service, so we're just duplicating the test from that service

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/TOILRequestTest.php
```diff
@@ -0,0 +1,150 @@
...
+    $leaveRequestRightsService = new LeaveRequestRightsService($leaveManagerService);
+
+    return new TOILRequestService(
+      $this->leaveBalanceChangeService,
+      $leaveRequestStatusMatrixService,
+      $leaveRequestRightsService
+    );
+  }
+
+  private function getTOILRequestServiceWhenCurrentUserIsAdmin() {
```
davialexandre: I think you're not using this anywhere

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/TOILRequestTest.php
```diff
@@ -20,12 +20,19 @@ class api_v3_TOILRequestTest extends BaseHeadlessTest {
   use CRM_HRLeaveAndAbsences_LeaveBalanceChangeHelpersTrait;
   use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
   use CRM_HRLeaveAndAbsences_TOILRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_SessionHelpersTrait;
+
+  private $leaveContact;
 
   public function setUp() {
     CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
```
davialexandre: single quotes

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/TOILRequest.php
```diff
@@ -2,29 +2,42 @@
...
 use CRM_HRLeaveAndAbsences_BAO_TOILRequest as TOILRequest;
+use CRM_HRLeaveAndAbsences_Test_Fabricator_LeaveBalanceChange as LeaveBalanceChangeFabricator;
 
 class CRM_HRLeaveAndAbsences_Test_Fabricator_TOILRequest {
 
   private static $leaveRequestStatuses;
   private static $toilAmounts;
 
-  public static function fabricate($params) {
+  public static function fabricate($params, $withBalanceChanges = false) {
```
tunbola: I tried to follow same format for LeaveRequest Fabricator which use false as default. Besides I think its better not to create Balance changes by default, it will create unnecessary overhead for tests. The tests that were using the fabricator before, once I moved the balance change logic for the BAO, they started failing so I only added the true parameter only for those that failed. If it were true by default, all the tests would have passed but there would be some that didn't need balance changes to work.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
tunbola: Thought of that too but there was an extra logic in the parent method to delete LeaveBalance changes for the Leave Request. Since LeaveBalance changes for TOIL is created differently, I thought I should do it this way. although I know calling the parent::delete function won't break the function. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/TOILRequestTest.php
tunbola: Its just a single test added to ensure that the rule that applies to the TOILRequest service also applies to the Leave Request Service

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/TOILRequest.php
davialexandre: My impression was that almost (i would say something like 90%) of the tests required that you add the true param. But, I think that the argument in favor of consistency is a good one, so let's keep it as false

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/TOILRequest.php
davialexandre: Since the logic to delete the balance change is harmless, I think it's fine to use the parent's method. I prefer to call one additional method to delete balance changes instead of duplicating the logic to delete the dates and leave request

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/TOILRequestTest.php
davialexandre: I still think it's not necessary. TOILRequest service extends LeaveRequest service, which is the same to say that TOILRequest service IS A LeaveRequest service. So, if they're the same, there's no need to duplicate the tests

PR #1489, opened 2017-02-20
PCHR-1868:  Modify LeaveRequest.AddComment API and Comment.create 

tunbola:
This PR modifies the Create method of the Comment Entity BAO. 
The changes are as follows:

- The create method now allows the created_at date parameter to be passed when creating a new comment although this is still disallowed when updating a comment.
- As a security check for allowing this (1st change), a new comment will not be allowed to be created when the created_at date is less than the created_at date of the last comment for that entity.

The created_at is also added as an optional parameter for the LeaveRequest.addComment API endpoint.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
+++ b/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
@@ -155,11 +155,17 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
    */
   private static function validateCommentCreatedAtDateDuringAnUpdate($params) {
     if (isset($params['id']) && !empty($params['created_at'])) {
-      throw new InvalidCommentException(
-        'You cannot update the created_at date of a comment',
-        'comment_cannot_update_created_at',
-        'created_at'
-      );
+      $comment = self::findById($params['id']);
+      $commentDate = new DateTime($comment->created_at);
+      $createdAt = new DateTime($params['created_at']);
+
+      if ($commentDate != $createdAt) {
+        throw new InvalidCommentException(
+          'You cannot update the created_at date of a comment',
+          'comment_cannot_update_created_at',
+          'created_at'
+        );
+      }
     }
   }
 
--- a/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
+++ b/uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
@@ -19,14 +19,16 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
     if($validate){
       self::validateParams($params);
     }
-    unset($params['created_at']);
+
     unset($params['is_deleted']);
 
     $instance = new self();
     $instance->copyValues($params);
 
+    $instance->created_at = CRM_Utils_Date::processDate('now');
+
     if ($hook == 'create') {
-      $instance->created_at = CRM_Utils_Date::processDate('now');
+      $instance->created_at = isset($params['created_at']) ? $params['created_at'] : CRM_Utils_Date::processDate('now');
     }
 
     $instance->save();
@@ -46,6 +48,7 @@ class CRM_HRComments_BAO_Comment extends CRM_HRComments_DAO_Comment {
   public static function validateParams($params) {
     self::validateMandatory($params);
     self::validateCommentSoftDeleteDuringUpdate($params);
+    self::validateCommentCreatedAtDate($params);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
```diff
@@ -122,4 +157,26 @@ public static function softDelete($id) {
...
+  /**
+   * This method fetches the last/latest comment for the given entityName and entityID
+   *
+   * @param string $entityName
+   * @param int $entityID
+   *
+   * @return \CRM_HRComments_BAO_Comment|null
+   */
+  private static function getLastCommentForEntity($entityName, $entityID) {
+    $comment  = new self();
```
davialexandre: there are 2 spaces before the = sign

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
```diff
@@ -19,14 +19,16 @@ public static function create($params, $validate = true) {
...
       self::validateParams($params);
     }
-    unset($params['created_at']);
+
     unset($params['is_deleted']);
 
     $instance = new self();
     $instance->copyValues($params);
 
+    $instance->created_at = CRM_Utils_Date::processDate('now');
```
davialexandre: This will always set the `created_at` date to now on updates, which I don't think is correct. I think we should throw and error if someone tries to change the date on an update

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -96,7 +96,7 @@ public function testTryingToSoftDeleteCommentOnUpdateThrowsException() {
     ]);
   }
 
-  public function testCreatedDateIsInsertedCorrectly() {
+  public function testCreatedDateIsInsertedCorrectlyWhenNotPassedAsAParameter() {
```
davialexandre: please rename this test to something more specific like `testCreatedDateIsInsertedAsTheCurrentDateWhenNotPassedAsAParameter`

uk.co.compucorp.civicrm.hrcomments/tests/phpunit/CRM/HRComments/BAO/CommentTest.php
```diff
@@ -216,4 +213,45 @@ public function testIsDeletedCannotBeChangedWhenUpdatingExistingComment() {
     $this->assertEquals('SickRequest', $comment->entity_name);
     $this->assertEquals('This is a random sample comment', $comment->text);
   }
+
+  public function testValidateCommentCreatedAtDate() {
```
davialexandre: I think that maybe we should test this using the `create` method instead of `validateParams`. The reason is to make sure that `validateParams` will be called inside the create method.

By changing that, please also change the test name to something more descriptive like `testCommnetCanNotBeCreatedWithACreatedDateLessThanTheCreatedDateOfTheLastCommentForThisEntity` (please feel free to use a better a name if you can think of one)

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -469,6 +469,14 @@ function _civicrm_api3_leave_request_addcomment_spec(&$spec) {
     'FKClassName'  => 'CRM_Contact_DAO_Contact',
     'FKApiName'    => 'Contact',
   ];
+
+  $spec['created_at'] = [
+    'name' => 'created_at',
+    'type' => CRM_Utils_Type::T_TIME,
```
davialexandre: The type should be `CRM_Utils_Type::T_DATE + CRM_Utils_Type::T_TIME`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCommentTest.php
```diff
@@ -58,6 +58,37 @@ public function testAddCanCreateCommentForLeaveRequest() {
...
+    $comment->fetch();
+
+
+    $expected = [
+      [
+        'comment_id' => $comment->id,
+        'leave_request_id' => $comment->entity_id,
+        'text' => $comment->text,
+        'contact_id' => $comment->contact_id,
+        "created_at"=> $created_at->format('YmdHis')
```
davialexandre: using single quotes here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCommentTest.php
```diff
@@ -58,6 +58,37 @@ public function testAddCanCreateCommentForLeaveRequest() {
...
+      'sequential' => 1
+    ];
+
+    $result = $this->leaveRequestCommentService->add($params);
+
+    $comment = new Comment();
+    $comment->find();
+    $this->assertEquals(1, $comment->N);
+    $comment->fetch();
+
```
davialexandre: remove one of those 2 blank lines

uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/BAO/Comment.php
```diff
@@ -146,6 +145,25 @@ private static function validateCommentCreatedAtDate($params) {
...
+   * Validates that the created_at parameter should not be allowed
+   * to be changed during an update.
+   *
+   * @param array $params
+   *   The params array received by the create method
+   *
+   * @throws \CRM_HRComments_Exception_InvalidCommentException
+   */
+  private static function validateCommentCreatedAtDateDuringAnUpdate($params) {
+    if (isset($params['id']) && !empty($params['created_at'])) {
```
davialexandre: So here, the docblock says something but the method is doing another thing. The doc says the `created_at` cannot be changed (that is, the value was 2016-01-01 and now they are trying to change it to 2016-01-02). But what the method is doing is, even if the given date is the same as the current one, it will throw an exception.

I think the better option is to load the comment with the given ID and then compare if the dates are different. The reason is that, depending on how the frontend code works, it's probably easier for them to always pass the created_at value, even if it has not changed. So doing this would avoid unexpected errors

PR #1492, opened 2017-02-21
PCHR-1804: Hide disabled location region department level options on job role dropdown

bmnepali:
### Issue:
Currently, all available options values  are fetched both enabled or disabled, which are displayed in dropdown options in job role form without filtering the option values with their activeness.

### Solution:
Modified, job role controller and job role service to fetch the option groups with their activeness indicator `is_active` attribute. Which is used in filtering the active options form the group and displayed in the dropdown options.

#### 1. List of locations both enabled and disabled
screen shot 2017-01-30 at 1 12 25 pm

##### a) In new job role form
screen shot 2017-02-21 at 1 00 40 pm

##### b) In Edit Jb role form
screen shot 2017-02-21 at 1 00 18 pm

**Note**: 
Similarly, it works with departments, region, level options both in new jobrole form and edit job role form as well.


com.civicrm.hrjobroles/js/src/job-roles/filters/option-values-filter.js
```diff
@@ -0,0 +1,22 @@
...
+    return function(optionValues) {
+      var filteredOptionValues = {};
+
+      _.each(optionValues, function (optionValue, idValue) {
+        if (optionValue.is_active === '1') {
+          filteredOptionValues[idValue] = optionValue;
+        }
+      });
+
+      return filteredOptionValues;
```
dhavalsharma: `return _.filter(optionValues, function(o) { return !o.is_active === '1'; });`

bmnepali: Here in this case i need a `{}` but `_.filter(...)` return a `[]`. So, lodash filter can't be used.
Ref: https://jsfiddle.net/bmnepali/85tnzg7e/

PR #1509, opened 2017-02-23
PCHR-459: Manual creation for activity and case types instead of manged entitites

omarabuhussein:
Civicrm managed entities handler are unable to clean cases before removing the case type no matter what you put in 'cleanup' field ( refer to civi doc to know more about
cleanup field) because case type cannot be deleted when it contain any attached case and managed entity handler canoot clean case records.

This PR removes the previous upgrader and create a new one that creates all needed data manually , in the same time I left the definition of these case types as managed
entities in **hrcase_civicrm_managed** hook since civicrm will still cause an error if you removed them from there ( it will try to delete them and will fail if any case type contain re
related cases ) .


hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -248,16 +269,144 @@ public static function isExtensionEnabled($key)  {
...
+    foreach($caseTypesToRemove as $caseType) {
+      $this->removeUnusedCaseType($caseType);
+    }
+  }
+
+  /**
+   * Creates/Updates default case types to be shipped
+   * with this extension
+   *
+   * @param $defaultCaseTypes
```
davialexandre: missing param type

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -248,16 +269,144 @@ public static function isExtensionEnabled($key)  {
...
+
+      civicrm_api3('CaseType', 'create', $caseType);
+    }
+  }
+
+  /**
+   * Creates default activity types to be shipped
+   * with this extension if they are not exits
+   *
+   * @param $defaultActivityTypes
```
davialexandre: missing param type here too

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -248,16 +269,144 @@ public static function isExtensionEnabled($key)  {
...
+   */
+  private function up1402_createDefaultActivityTypes($defaultActivityTypes) {
+    foreach ($defaultActivityTypes as $activityType) {
+      $type = civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
+        'name' => $activityType,
+        'option_group_id' => "activity_type",
```
davialexandre: please make sure to use single quotes for all the keys and values on this array and on the next one

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -248,16 +269,144 @@ public static function isExtensionEnabled($key)  {
...
+  }
+
+  /**
+   * Changes Activity type component
+   *
+   * @param $activityTypeName
+   *   Activity type name
+   * @param $currentComponent
+   *   Current activity type component
+   * @param $newComponent
```
davialexandre: missing param types here

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -248,16 +269,144 @@ public static function isExtensionEnabled($key)  {
...
+   *  New activity type component
+   */
+  private function changeActivityTypeComponent($activityTypeName, $currentComponent, $newComponent) {
+    civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => "activity_type",
```
davialexandre: use single quotes for value

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -266,83 +415,67 @@ private function removeManagedEntityRecord($name, $type) {
...
-    if (!empty($componentActivityTypes['values'])) {
-      $componentActivityTypes = $componentActivityTypes['values'];
-    }
-    else {
-      $componentActivityTypes = [];
-    }
+      if (!empty($caseTypeRow['id'])) {
+        // Removing all cases related to this case type , we using
+        // DAO instead of API because API deletion does not supports
+        // batch delete and this should be much faster
```
davialexandre: this comment doesn't make sense to me. The code is just using a simple delete, even though the comment talks about a batch delete. Is this correct?

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -266,83 +415,67 @@ private function removeManagedEntityRecord($name, $type) {
...
-        }
       }
     }
   }
 
+  private function removeActivityTypes($activityTypes) {
+    civicrm_api3('OptionValue', 'get', [
+      'name' => ['IN' => $activityTypes],
+      'component_id' => 'CiviTask',
+      'option_group_id' => "activity_type",
```
davialexandre: single quotes for the value

hrcase/CRM/HRCase/Upgrader.php
omarabuhussein: nope , hence ($caseDAO->case_type_id = $caseTypeRow['id'];) which means delete all cases belongs to a specific case type which may be one or thousands of records

PR #1510, opened 2017-02-23
AI-16: new activity

meladawy:
Fix Comments

- add extra padding for all fields (55px left) —> to look symmetrical with the padding from the email text editor's right side
- To make sure the full width panels are not affected, Jamie suggested to remove outline and then push individual things in
- extend Subject, Assigned to and With Contact fields. Assigned to and With Contact fields to be extended only if this doesn't affect other screens.

ai-16

AFTER

ai-16-after


jamienovick:
like!

PR #1524, opened 2017-02-24
AI-23: Add right padding

meladawy:
After: 
ai-23-after


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_memberships.scss
```diff
@@ -55,6 +55,15 @@
 }
 
 .page-civicrm-contact-view {
+  .ui-widget-content {
+    &.ui-tabs-panel {
+      padding-right: 15px !important;
```
omarabuhussein: but specific it like that won't work on all different screens without using media queries , right ?

PR #1533, opened 2017-02-27
AI-22: Fix export contacts styling

brunoqueiros:
**Before**

screenshot-1

**After**

screencapture-localhost-amnesty-civicrm-contact-search-1488198080273


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_export-contact.scss
```diff
@@ -31,6 +31,8 @@
 
     h3 {
       padding-left: 0;
+      color: $gray-darker;
+      font-size: 16px;
```
meladawy: Can u use $font-size-h3 instead ?

PR #1536, opened 2017-02-27
AI-26: contacts events tab

meladawy:
Before: 
ai-26-before-1
ai-26-before-2
ai-26-before-3
ai-26-before-4

After:
ai-26-after-1
ai-26-after-2
ai-26-after-3
ai-26-after-4


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_events.scss
```diff
@@ -207,12 +214,34 @@
     }
   }
 
+  .price_set-section {
+
+  }
```
bmnepali: Its a empty section, why are we keeping it?

PR #1539, opened 2017-02-28
PCHR-1917: Create LeaveRequest.deleteAttachment API

tunbola:
This PR creates the LeaveRequest.deleteAttachment API endpoint.
The API uses the delete method exposed by the newly created LeaveRequestAttachment Service.
Also only Leave Approvers and admins can delete attachments from a Leave Request.

Sample Request
```php
$result = civicrm_api3('LeaveRequest', 'deleteattachment', array(
  'sequential' => 1,
  'leave_request_id' => 1,
  'attachment_id' => 1,
));
```
Response
```json
{
    "is_error": 1,
    "version": 3,
    "count" : 0,
    "values: [],
}
```

**Note** that this API only works currently when called in php/smarty but not via REST. When called via REST the check_permissions parameter is set to true and cannot be overriden and triggers some checks in core.  There is a blocker in core that prevents an attachment from being created or deleted due to this [HERE](https://github.com/civicrm/civicrm-core/blob/master/Civi/API/Subscriber/DynamicFKAuthorization.php#L226). Core tries to check if the entity associated with the attachment can be created/modified using the parameters sent to the Attachment API. However we have some BAO level validations on Leave and Absence Extension which prevents this check from going through and a validation error is thrown.

Also note that a PR for CORE also currently blocks the creation of attachments for entities other than core entities. [HERE](https://github.com/civicrm/civicrm-core/pull/9875)

For the purpose of this PR since the tests call the API within PHP with check_permissions set to false, the tests passed successfully but a solution needs to be found for this issue because when creating attachments using the internal callback, Core manually sets check_permission to be true [HERE](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Page/AJAX/Attachment.php#L101).

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
@@ -1,0 +1,75 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_Service_LeaveManager as LeaveManagerService;
+
+class CRM_HRLeaveAndAbsences_Service_LeaveRequestAttachment {
+
+  /**
+   * Uses the Attachment API to delete an attachment associated with a LeaveRequest.
+   * This method also implement some checks to ensure that only the LeaveRequest Approver
+   * or an Admin can delete attachments for a leave request
+   *
+   * @param array $params
+   *
+   * @throws UnexpectedValueException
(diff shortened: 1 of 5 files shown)


colemanw:
I believe the solution for the first blocker might be instead of doing a rollback, switch to using `\Civi::service('civi_api_kernel')->runAuthorize()`. What do you think @totten?

davialexandre:
@colemanw @totten `\Civi::service('civi_api_kernel')->runAuthorize()` looks like a much better option to me . Even if it wasn't for the blocker we found, I don't think it's a good idea to actually update a entity just to check if you have permission to do it.
Just so we can keep track of this, I've created a ticket for this https://issues.civicrm.org/jira/browse/CRM-20210. If the `runAuthorize()` solution looks good, we can create a PR for this

PR #1548, opened 2017-03-01
PCHR-1875: Fixed Action menu design break

mukeshcompucorp:
## Problem 1
Text in action menu was longer which caused it to break on staging. Check different highlighted content.

ticket-1

ticket1 1

## Problem 2
Due to fixed height, the menu was overflowing as highlighted

ticket-2

## Solution 1
Increased width so that normal text gets covered within width.

resolved-ticket1

## Solution 2
Add overscroll-y:auto in action menu to prevent overflowing menu.

resolved-ticket2


mi-kon:
@mukeshcompucorp  there should not be a scrollbar in dropdown, the dropdown height should be dynamic.

PR #1555, opened 2017-03-02
AI-58: Membership import

meladawy:
Before:
ai-58-before-1
ai-58-before-2
ai-58-before-3
ai-58-before-4

After:
ai-58-after-1
ai-58-after-2
ai-58-after-3
ai-58-after-4


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_member-import.scss
```diff
@@ -0,0 +1,90 @@
+.page-civicrm-member-import {
+  .help {
+    padding: 0px 20px;
```
brunoqueiros: Remove `px` from `0px`.

PR #1557, opened 2017-03-02
PCHR-1820: Convert Pension Provider to Contact

omarabuhussein:
## Problem
Pension providers in contractual info are currently handled as dropdown options, stored as option values in 'hrjc_pension_type' option group. They should be contacts of a reserved contact type (pension providers) based on organizations.

## Solution
Implementing pension providers as contacts required creating a new contact sub-type.  This implied changing how pension provider was being used throughout the whole application.  In particular:
- Created upgrader method that replaces hrjc_pension_type option group for a new contact subtype called "Pension Provider", which:
  - Creates new pension_provider sub-type for contacts
  - Replaces old option value references in contract revisions for contact id's
  - Adds a foreign key on pension_type field to contacts' table
  - Removes hrjc_pension_type option group
- Altered HRJobPension DAO to implement reference to contacts table on pension_type field.
- Replaced inputs on forms working with option values with input boxes working with contacts with pension_provider sub-type.
- Changed validation on imports to use contacts instead of option values.
- Implemented use of new pension provider reference to contacts on advanced search, to enable queries using complete or partial organization name.
- Changed tests so they use contacts as pension providers.

pchr-1820-after-2


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,67 +1065,153 @@ public function createInsurancePlanTypes() {
...
+      CRM_Core_DAO::executeQuery("
+        ALTER TABLE {$pensionsTableName} 
+        MODIFY pension_type INT(10) unsigned,
+        ADD CONSTRAINT pension_type_contact_id_fk
+        FOREIGN KEY(pension_type)
+        REFERENCES civicrm_contact(id)
+        ON DELETE SET NULL ON UPDATE CASCADE;
+      ");
+    } catch (Exception $e) {
+      $messages[] = "Error altering table $pensionsTableName: " . $e->getMessage();
```
davialexandre: you can use single quotes here and in the other messages

hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/HRJobPension.php
```diff
@@ -13,26 +13,39 @@ class CRM_Hrjobcontract_Test_Fabricator_HRJobPension extends BaseAPIFabricator {
    */
   public static function fabricate($params) {
     if (!empty($params['pension_type'])) {
-      $pensionTypeResult = civicrm_api3('OptionValue', 'get', [
+      $pensionTypeResult = civicrm_api3('Contact', 'get', [
         'sequential' => 1,
-        'option_group_id' => 'hrjc_pension_type',
-        'value' => $params['pension_type'],
+        'contact_type' => "Organization",
```
davialexandre: use single quotes for all the params

hrjobcontract/CRM/Hrjobcontract/Test/Fabricator/HRJobPension.php
```diff
@@ -13,26 +13,39 @@ class CRM_Hrjobcontract_Test_Fabricator_HRJobPension extends BaseAPIFabricator {
...
+   * @return array
+   *   Result of creating new provider.
+   */
   private static function createPensionType($value) {
-    $result = civicrm_api3('OptionValue', 'create', [
-      'value' => $value, 
-      'name' => 'Test Pension Type ' . mt_rand(1000, 9000), 
-      'option_group_id' => 'hrjc_pension_type'
+    $result = civicrm_api3('Contact', 'create', [
+      'contact_type' => "Organization",
```
davialexandre: single quotes

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -380,13 +380,27 @@ function testPensionImport() {
...
     $this->validateResult($contactID, 'HRJobPension');
+
+    $params['HRJobPension-pension_type'] = 'Wrong Pension Provider';
+    $failedImportResponse = $this->runImport($params);
+    $this->assertEquals(CRM_Import_Parser::ERROR, $failedImportResponse);
+  }
+
+  private function createPensionProvider($providerName) {
+    $provider = ContactFabricator::fabricateOrganization([
+      'contact_type' => "Organization",
```
davialexandre: single quotes

hrjobcontract/CRM/Hrjobcontract/BAO/Query.php
```diff
@@ -139,6 +139,12 @@ function whereClauseSingle(&$values, &$query) {
         $query->_tables['civicrm_hrjobcontract_pension'] = $query->_whereTables['civicrm_hrjobcontract_pension'] = 1;
         return;
 
+      case 'hrjobcontract_pension_pension_type':
+        $query->_qill[$grouping][]  = 'Pension Provider contains "' . $value . '"';
+        $query->_where[$grouping][] = CRM_Contact_BAO_Query::buildClause("pension_provider.sort_name", 'LIKE', '%' . $value . '%');
```
davialexandre: single quotes

hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContractRevision.php
```diff
@@ -315,19 +315,6 @@ private static function buildEntityQueryArray($entity, $revision, $query) {
         ";
         break;
 
-      case 'pension':
```
davialexandre: don't we need to replace this with something else (fetch the pension provider name from contact)?

MiyaNoctem: No need, this is handled in front-end, just as it's done for health and life insurance providers.  The only place where this label was used was on contract summary page, and now that it's handled by obtining the contact, is unnecessary.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1072,23 +1072,19 @@ public function createInsurancePlanTypes() {
...
-        MODIFY pension_type INT(10) unsigned,
-        ADD CONSTRAINT pension_type_contact_id_fk
-        FOREIGN KEY(pension_type)
-        REFERENCES civicrm_contact(id)
-        ON DELETE SET NULL ON UPDATE CASCADE;
-      ");
-    } catch (Exception $e) {
-      $messages[] = "Error altering table $pensionsTableName: " . $e->getMessage();
-    }
+    CRM_Core_DAO::executeQuery("
```
davialexandre: I think you should also remove the exception handling from the other helper methods (createPensionProvider, replacePensionProviders and removePensionTypeOptionGroup). It will make the upgrader stop on the first error it find.

On the other hand, I wonder if all this upgrade logic should be inside a transaction. My concern is that we might have an error in the middle of the process and then the system might get in a state that is not easily recoverable (unless you go to the database and manually fix things). I don't know if API calls would work inside a transaction though.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+      CRM_Core_DAO::executeQuery("
+        ALTER TABLE $pensionsTableName
+        MODIFY pension_type INT(10) unsigned,
+        ADD CONSTRAINT pension_type_contact_id_fk
+        FOREIGN KEY(pension_type)
+        REFERENCES civicrm_contact(id)
+        ON DELETE SET NULL ON UPDATE CASCADE;
+      ");
+    } catch (Exception $e) {
+      $civi_db_settings = parse_url(CIVICRM_DSN);
```
davialexandre: for some variables you're using snake_case, but for other you're using camelCase. Can you please use camelCase for everything?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+    }
 
     if (!empty($messages) && function_exists('drupal_set_message')) {
       foreach ($messages as $message) {
         drupal_set_message($message);
       }
     }
 
-    return true;
+    return false;
```
davialexandre: shouldn't this be true?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+      $this->createPensionProviderType($messages);
+      $this->replacePensionProviders($pensionsTableName, $messages);
+      $this->removePensionTypeOptionGroup($messages);
+      $this->alterPensionsTable($pensionsTableName, $messages);
+    } catch (Exception $ex) {
+      $tx->rollback();
+      throw new Exception('Cannot do upgrade - ' . $ex->getMessage());
+    }
 
     if (!empty($messages) && function_exists('drupal_set_message')) {
```
davialexandre: I think this will never get executed. If $messages is not empty, it means some of the methods have thrown an exception, right? And if that is the case, then the block inside the catch will be executed. This block also throws an Exception and the method flow will stop there, never reaching this if

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+      
+      if ($fkCheck->N == 0) {
+        throw new Exception("Error updating $pensionsTableName table: " . $e->getMessage());
+      }
+
+      $pensionTypeColumnCheck = CRM_Core_DAO::executeQuery("DESC $pensionsTableName");
+      while ($pensionTypeColumnCheck->fetch()) {
+        if ($pensionTypeColumnCheck->Field == 'pension_type' && $pensionTypeColumnCheck->Type != 'int(10) unsigned') {
+          throw new Exception("Error updating $pensionsTableName table: " . $e->getMessage());
+        }
```
davialexandre: I don't understand the need for all this complexity here. If the ALTER table command results in an error, the only thing that matters is that something went wrong and we will rollback. What exactly went wrong it not that important (I think). Why are you doing all these checks about the table structure?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
MiyaNoctem: Actually, I was using messages to notify about minor incidents, for example, trying to create a contact type that already exists.  The API will throw an exception, but I capture the exception and let it pass, so long as there is a contact type with 'Pension_Provider' name.  I know the probability of someone creating a "Pension_Provider" contact type before running this upgrade is slim, but then again, should the upgrade fail if this happens? Same idea applies for the rest of the messages: if an exception is thrown on an API call, but I verify everything is as it's supposed to be, I use the $messages array to warn, but let the upgrade finish succesfully.

davialexandre: Oh, I see. I this case, do you think the warning will make any difference? Think you're the user updating the extension, then the upgrade runs successfully but you will still see those warnings. So what? Should I be worried? Should I do something about it? Is everything really fine? I'm afraid it will just make the users more confused. If everything went fine, I don't think we should warn them about anything.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
davialexandre: Good point about ALTER table committing the PR. I didn't think of that.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+        ADD CONSTRAINT pension_type_contact_id_fk
+        FOREIGN KEY(pension_type)
+        REFERENCES civicrm_contact(id)
+        ON DELETE SET NULL ON UPDATE CASCADE;
+      ");
+    } catch (Exception $e) {
+      $civi_db_settings = parse_url(CIVICRM_DSN);
+      $civi_db_name = trim($civi_db_settings['path'], '/');
+
+      $fkCheck = CRM_Core_DAO::executeQuery($q = "
```
davialexandre: Please remove the `$q = `

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1065,41 +1065,91 @@ public function createInsurancePlanTypes() {
...
+   * key referencing civicrm_contact.
+   * 
+   * @param type $pensionsTableName
+   *   Name of pension data table
+   * @param type $messages
+   *   Array of messages for salvageable errors found
+   * 
+   * @throws Exception
+   */
+  function alterPensionsTable($pensionsTableName, &$messages) {
```
davialexandre: you're not using $messages anywhere in this method. Please remove the param

davialexandre: please make this and the other helper methods private

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1117,9 +1167,7 @@ private function removePensionTypeOptionGroup(&$messages) {
         'name' => 'hrjc_pension_type',
       ]);
 
-      if ($pensionOptionGroup['count'] == 0) {
-        $messages[] = 'Error deleting hrjc_pension_type option group: ' . $e->getMessage();
-      } else {
+      if ($pensionOptionGroup['count'] > 0) {
```
davialexandre: with this, you're not using $messages in this method anymore. Please remove the param from the method

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1131,7 +1179,10 @@ private function removePensionTypeOptionGroup(&$messages) {
    * contracts.
    * 
    * @param array $messages
-   *   Array of messages for errors found
+   *   Array of messages for salvageable errors found
+   * 
+   * @throws Exception
+   *   
    */
   private function replacePensionProviders($pensionsTableName, &$messages) {
```
davialexandre: Please remove $messages from here as well

hrjobcontract/js/src/job-contract/controllers/form/form-pension.js
```diff
@@ -1 +1,34 @@
...
+      $log.debug('Controller: FormPensionCtrl');
+
+      $scope.contacts = {
+        Pension_Provider: []
+      };
+
+      $scope.refreshContacts = function(input, contactSubType){
+          if (!input) {
+            return;
+          }
```
bmnepali: Please main 2 spaced indentation.

hrjobcontract/js/src/job-contract/controllers/form/form-pension.js
```diff
@@ -1 +1,34 @@
...
+
+        ContactService.search(input, {
+          contact_type: 'Organization',
+          contact_sub_type: contactSubType
+        }).then(function(results){
```
bmnepali: not `function(results){` . use ` function (results) {` instead. Also please give`results` a meaningfull name maybe ` contactSubTypes`

hrjobcontract/js/src/job-contract/controllers/form/form-pension.js
```diff
@@ -1 +1,34 @@
...
+          $scope.contacts[contactSubType] = results;
+        });
+      };
+
+      if ($scope.entity.pension.pension_type) {
+        ContactService.getOne($scope.entity.pension.pension_type).then(function(result){
+          $scope.contacts.Pension_Provider.push(result);
```
bmnepali: Here also, rename `result` to more meaningful name.

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+      });
+    });
+
+    beforeEach(function() {
+      $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond({});
+      $httpBackend.whenGET(/views.*/).respond({});
+    });
+
+    describe('On initialization', function() {
+      describe('Case I: when $scope.entity.pension.pension_type is truthy', function() {
```
dhavalsharma: reword
`when pension type is true`

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+        beforeEach(function() {
+          initController(1);
```
dhavalsharma: what is 1?

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+          $scope.$digest();
+        });
+
+        it('defines contact object', function() {
```
dhavalsharma: defines contacts collection

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+          expect($scope.contacts).toBeDefined();
+        });
+
+        it("calls getOne()", function() {
```
dhavalsharma: calls contact's service api to get one contact? or similar

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+          expect(ContactService.getOne).toHaveBeenCalled();
+        });
+
+        it("sets value for $scope.contacts.Pension_Provider", function() {
```
dhavalsharma: in test cases describe the variables and ps avoid using them directly so this could be 

> sets the values for contact's pension provider

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+          expect($scope.contacts.Pension_Provider.length).toBe(1);
+          expect($scope.contacts.Pension_Provider[0].contact_id).toBe(ContactMock.contact.values[0].contact_id);
+          expect($scope.contacts.Pension_Provider[0].contact_type).toBe(ContactMock.contact.values[0].contact_type);
+        });
+      });
+
+      describe('Case II: when $scope.entity.pension.pension_type is falsy', function() {
```
dhavalsharma: when pension type is false

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+        beforeEach(function() {
+          initController(null);
```
dhavalsharma: null can be default in controller?

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+        it('returns resoponse to be falsy', function() {
+          expect(response).toBeFalsy();
+        });
+
+        it('does not call search()', function() {
+          expect(ContactService.search).not.toHaveBeenCalled();
+        });
+      });
+
+      describe('when input is truthy', function() {
```
dhavalsharma: describe input?

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+
+        it('sets contact_sub_types data in $scope.contacts', function() {
+          expect(ContactService.search).toHaveBeenCalled();
+          expect($scope.contacts[params.contact_sub_type].length).toBe(1);
+          expect($scope.contacts[params.contact_sub_type]).toEqual(ContactMock.contactSearchData.values);
+        });
+      });
+    });
+
+    function initController(pension_type) {
```
dhavalsharma: document function

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -0,0 +1,136 @@
...
+          contact_type: 'Organization',
+          contact_sub_type: 'Life_Insurance_Provider'
+        };
+        contactServiceSpy();
+      });
+    });
+
+    beforeEach(function() {
+      $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond({});
+      $httpBackend.whenGET(/views.*/).respond({});
```
dhavalsharma: are they used anywhere in testing?

dhavalsharma:
add gif for the working functionality

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: oops! not need! will remove it.

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: Since `pension_type` is not a boolean value rather it is a numerical value. I would rather prefer to call `when pension type is truthy`

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: `1` is the value of `pension_type`

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: will update!

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: Since `pension_type` is not a boolean value rather it is a numerical value. I would rather prefer to call `when pension type is falsy`

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: will update the controller.

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: will do!

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: will do!

hrjobcontract/js/test/controllers/form/form-pension_test.js
dhavalsharma: ps add comment for same

hrjobcontract/js/test/controllers/form/form-pension_test.js
dhavalsharma: instead of truthy how about 
'is available'
or
'is valid' etc

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: added comments in controller creation function.

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -24,49 +24,44 @@ define([
...
     });
 
-    beforeEach(function() {
-      $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond({});
-      $httpBackend.whenGET(/views.*/).respond({});
-    });
-
     describe('On initialization', function() {
-      describe('Case I: when $scope.entity.pension.pension_type is truthy', function() {
+      describe('Case I: when pension_type is valid', function() {
```
dhavalsharma: `Case I:`  is not required

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -24,49 +24,44 @@ define([
...
         beforeEach(function() {
           initController(1);
```
dhavalsharma: please add a comment for what 1 is

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -24,49 +24,44 @@ define([
...
-        it("sets value for $scope.contacts.Pension_Provider", function() {
+        it("sets the values for contact's pension provider", function() {
           expect($scope.contacts.Pension_Provider.length).toBe(1);
           expect($scope.contacts.Pension_Provider[0].contact_id).toBe(ContactMock.contact.values[0].contact_id);
           expect($scope.contacts.Pension_Provider[0].contact_type).toBe(ContactMock.contact.values[0].contact_type);
         });
       });
 
-      describe('Case II: when $scope.entity.pension.pension_type is falsy', function() {
+      describe('Case II: when pension_type is not valid', function() {
```
dhavalsharma: `Case II: ` not required

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -77,12 +72,12 @@ define([
         initController(1);
       });
 
-      describe('when input is falsy', function() {
+      describe('when search text is not available', function() {
```
dhavalsharma: what search text?

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -77,12 +72,12 @@ define([
...
         beforeEach(function() {
           response = $scope.refreshContacts('', {});
         });
 
-        it('returns resoponse to be falsy', function() {
+        it('returns resoponse to be empty', function() {
```
dhavalsharma: typo in `resoponse`

hrjobcontract/js/test/controllers/form/form-pension_test.js
```diff
@@ -91,28 +86,32 @@ define([
...
           expect(ContactService.search).toHaveBeenCalled();
           expect($scope.contacts[params.contact_sub_type].length).toBe(1);
           expect($scope.contacts[params.contact_sub_type]).toEqual(ContactMock.contactSearchData.values);
         });
       });
     });
 
+    /**
+     * Creates FormPensionCtrl Controller
+     * @param  integer pension_type
```
dhavalsharma: also add that if its not passed then sets it to null

bmnepali: done!

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: fixed!

hrjobcontract/js/test/controllers/form/form-pension_test.js
bmnepali: fixed

PR #1572, opened 2017-03-06
PCHR-2009: Custom return page for "confirm calculation" page

tunbola:
This PR changes the way the Confirm Calculations page determines the page to return to after clicking 'save new entitlements' on the page. A new URL parameter called 'returnUrl' is passed to the confirm calculations page from the three main entry points to the confirm calculations page. 

Namely:

- Manage entitlements link on the Absence Periods list 
- After creating a new contract or making changes to an existing one
- After searching for contacts on the adv search and then selecting the "Manage entitlements" action.

The Confirm calculation page uses the value of the returnUrl parameter to determine the page to return to after the 'save new entitlements' button is clicked.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/Task/ManageEntitlements.php
```diff
@@ -69,4 +76,24 @@ private function getAbsencePeriods() {
     return $rows;
   }
 
+  /**
+   * This method builds and saves the URL the user was before coming to this form
+   * to the session
+   */
+  private function setPreviousUrl() {
```
davialexandre: When you select the "Manage Entitlements" action on the adv search, you are taken to a page that displays a list of absence periods, right? This page is `CRM_HRLeaveAndAbsences_Form_Task_ManageEntitlements` and the redirect to the entitlement calculation page happens here: https://github.com/civicrm/civihr/blob/97793470a199d82deba2b7bdbfd0d69add44692b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/Task/ManageEntitlements.php#L37-L51 

I think the logic you put here should actually be on that page and then you just pass the adv search page url to the returnUrl param

PR #1582, opened 2017-03-07
PCHR-1968: Modify Contact.getleavemanagees API

tunbola:
This PR modifies the Contact.getleavemanagees API.

A new required parameter: 'managed_by' is added.  Previously the API returns contacts information for contacts managed by the currently logged in user. The API has been modified to now return contact information for contacts managed by the given contactID passed in via the managed_by parameter.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
@@ -23,10 +23,9 @@ class api_v3_ContactTest extends BaseHeadlessTest {
   public function setUp() {
     CRM_Core_DAO::executeQuery('SET foreign_key_checks = 0;');
 
-    $this->contact1 = ContactFabricator::fabricate();
-    $this->contact2 = ContactFabricator::fabricate();
+    $this->contact1 = ContactFabricator::fabricate(['first_name' => 'ContactA']);
+    $this->contact2 = ContactFabricator::fabricate(['first_name' => 'ContactB']);
     $this->manager = ContactFabricator::fabricate();
-    $this->registerCurrentLoggedInContactInSession($this->manager['id']);
 
     $this->setLeaveApproverRelationshipTypes([
       'has leaves approved by',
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -83,21 +83,53 @@ public function testGetLeaveManageesDoesNotReturnFilteredOutFieldsWhenFilteredOu
...
 
-    //only the two contacts who are managed by the current logged in manager are returned
     $this->assertEquals(2, $result['count']);
     $this->assertEquals($result['values'][$this->contact1['id']]['id'], $this->contact1['id']);
     $this->assertEquals($result['values'][$this->contact1['id']]['display_name'], $this->contact1['display_name']);
     $this->assertEquals($result['values'][$this->contact2['id']]['id'], $this->contact2['id']);
     $this->assertEquals($result['values'][$this->contact2['id']]['display_name'], $this->contact2['display_name']);
   }
+
+  public function testGetLeaveManageesCanBeFilteredByDifferentFields() {
```
davialexandre: can we also have a test where we search for a partial name? Like, for example, searching for "civi" and then checking that it includes civihr_staff

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -83,21 +83,53 @@ public function testGetLeaveManageesDoesNotReturnFilteredOutFieldsWhenFilteredOu
...
-    $result = civicrm_api3('Contact', 'getleavemanagees');
+    $result = civicrm_api3('Contact', 'getleavemanagees', ['managed_by' => $this->manager['id']]);
 
-    //only the two contacts who are managed by the current logged in manager are returned
     $this->assertEquals(2, $result['count']);
     $this->assertEquals($result['values'][$this->contact1['id']]['id'], $this->contact1['id']);
     $this->assertEquals($result['values'][$this->contact1['id']]['display_name'], $this->contact1['display_name']);
     $this->assertEquals($result['values'][$this->contact2['id']]['id'], $this->contact2['id']);
     $this->assertEquals($result['values'][$this->contact2['id']]['display_name'], $this->contact2['display_name']);
   }
```
davialexandre: can you also add a security test? If I'am logged in as manager 1 but then I call the API passing the manager 2 ID I should get an empty response (Because I should not have permission to see other manager's managees)

PR #1589, opened 2017-03-08
PCHR-1714: Drag and drop files feature moved up and overlaps other fields

bmnepali:
### Issue:
Currently the styles for the drag and drop is dependent on the Drupal generated ids. Which  may change and break the styles in Application Form.

### Solution:
1. I tried to find and use css to identify the specific way to add styles to the drag and drop feature, but could not do using css alone. So had to use jQuery.
2. Used jQuery to find the label with file field in the form and add unique attribute `data-file="true"`. Which is used to add styles for the drag and drop feature. This code is added in /hrrecruitment/templates/CRM/HRRecruitment/Form/Application.tpl.
 `
  CRM.$(function ($) {
    $("input[type='file']")
    .closest("div")
    .parent()
    .find('label')
    .attr('data-file','true');
  });
`

3. Now modify the following in 'org.civicrm.bootstrapcivicrm/scss/civicrm/vacancies/_vacancies.scss'. 
`
      label[for="custom_87"] {
        @extend %label-input-file;
      }
`
to 
`
      label[data-file="true"] {
        @extend %label-input-file;
      }
`

#### Before:
screen shot 2017-03-08 at 11 40 13 am

#### After:
screen shot 2017-03-08 at 11 39 14 am


dhavalsharma:
`dragon`  is `drag` 
 `**` etc in pr description can be removed or used correctly

PR #1590, opened 2017-03-08
PCHR-1967: Augment Contact Instance

deb1990:
This PR Augments `ContactInstance` to add `leaveManagees` function.

**Note:** 
Previously ContactModel was augmented because the API only returned managees for currently logged in user. 
Now same functionality is added to the Instance. This is done as now the API supports sending `managed_by` parameter so `leaveManagees` can be called to get managees for any specific user.


org.civicrm.reqangular/src/common/models/instances/contact-instance.js
```diff
@@ -1,10 +1,22 @@
...
+    function (ModelInstance, ContactAPI) {
+      return ModelInstance.extend({
+
+        /**
+         * Finds the contacts who are managed this contact
+         *
+         * @return {Promise}
+         */
+        leaveManagees: function () {
+          return ContactAPI.leaveManagees(this.id);
```
AkA84: It could happen that this contact instance is not on the DB yet (for example if you are about to call `.create()` on it), thus it could be that it doesn't have an `id`. So please check that `id` exists before you call the api

deb1990: Error handling architecture must be finalised before implementing this change. So skipping it.

PR #1591, opened 2017-03-08
PCHR 2053: common service for file upload

dhavalsharma:
Implement a common file upload service that can upload multiple files to civicrm endpoint. 

`FileUpload` has 2 methods

1. `uploader` which creates a new instance of [FileUploader](https://github.com/nervgh/angular-file-upload)
2. `uploadAll`  is called to upload all file items which are in uploader's queue. This will trigger ajax calls to civicrm endpoint (`/civicrm/ajax/attachment` by default) to upload files.

## Sample Usage

The `FileUpload` service can be included by the entity instances to create and upload files.  A sample is shown below 
```javascript 
app.factory('LeaveRequestInstance', ['FileUpload', function(FileUpload) {
  return {
    create: function() {
        //below call will be in callback
      this.uploadAll({
      	entityID: '12'
      }).then();
    },
    init: function() {
      this.uploader = FileService.uploader({
          crmAttachmentToken: '123abc'      
      });
      return this;
    }
  }
}]);
```
The controller can create the file uploader object. A sample is shown below 
```javascript
app.controller('RequestCtrl', ['LeaveRequestInstance', function(LeaveRequestInstance) {
  var vm = this;
  vm.request = LeaveRequestInstance.init();
  vm.save = function() {
    vm.request.create();
  }
  return this;
}]);
```

The uploader object can be attached to `file` input type. A sample is shown below
```html
  <input type="file" nv-file-select uploader="vm.request.uploader" />
```

The above implementation is based on this [gist](https://gist.github.com/AkA84/a2eaddc00b0c1b0104a15c53237e9452).


org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
+define([
+  'common/modules/services',
+], function (module) {
+  'use strict';
+
+  module.factory('FileUploadService', ['$q', 'FileUploader',
```
AkA84: there is no need for `Service` in the name, it's clear that it is a service

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+    function ($q, FileUploader) {
+      var fileUploader;
```
AkA84: Why do you have this variable here? That means that you're gonna override any file uploader that was created with the newly created one, since you're storing each file uploader in the same variable

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+
+        /**
+         * Sets up the file uploader service
+         *
+         * @param {Object} customSettings can have keys like url for file upload path,
+         * formData for updating form data, queueLimit to limit number of files that can be uploaded at onCompleteAll
+         *
+         * @return {Object} an instance of FileUploader
+         **/
+        uploader: function (customSettings) {
```
AkA84: You could rewrite this as something like
```js
uploader: function (customSettings) {
  return new FileUploader({
    url: customSettings.url || '/civicrm/ajax/attachment',
    queueLimit: +customSettings.queueLimit || some_default?,
    formData: [{
      entity_table: customSettings.entityTable || some_default_or_throw_error?
      crm_attachment_token: customSettings.crm_attachment_token || some_default_or_throw_error?
    }]
  });
},
```

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+      return {
+
+        /**
+         * Sets up the file uploader service
+         *
+         * @param {Object} customSettings can have keys like url for file upload path,
+         * formData for updating form data, queueLimit to limit number of files that can be uploaded at onCompleteAll
+         *
+         * @return {Object} an instance of FileUploader
+         **/
```
AkA84: A documentation is closed by `*/`, not by `**/`
```js
/**
 *
 */
```

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+          }
+
+          fileUploader = new FileUploader(uploaderSettings);
+          return fileUploader;
+        },
+
+        /**
+         * Uploads all files in queue updating with additional form data
+         *
+         * @param {Object} additionalFormData that has keys like entityID or description
```
AkA84: you are missing the `@return` line

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+         **/
+        uploadAll: function (additionalFormData) {
+          var deferred = $q.defer(),
+            results = [];
+
+          fileUploader.onBeforeUploadItem = function (item) {
```
AkA84: Can you explain what is happening here?

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+            for (key in additionalFormData) {
+              var snakeCaseKey = _.snakeCase(key);
+              item.formData.push({
+                snakeCaseKey: additionalFormData[key]
+              });
+            }
+          };
+
+          fileUploader.onCompleteItem = function (item, response) {
```
AkA84: Shouldn't this be part of the `upgrader` method?

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+            results.push(response);
+          };
+
+          fileUploader.onErrorItem = function (item, response, status, headers) {
```
AkA84: Same as above, shouldn't this be part of the `upgrader` method? Why do you need to redefine it every time you trigger the upload?

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -0,0 +1,81 @@
...
+            $log.error(' ===== Item Error: ' + status + ' ======');
+            $log.error(' =====  - item ======');
+            $log.error(item);
+            $log.error(' =====  - response ======');
+            $log.error(response);
+            $log.error(' =====  - headers ======');
+            $log.error(headers);
+          };
+
+          fileUploader.onCompleteAll = function () {
```
AkA84: You are missing the documentation for all of this methods

org.civicrm.reqangular/test/mocks/services/file-uploader-mock.js
```diff
@@ -0,0 +1,54 @@
+define([
+  'common/mocks/module'
+], function (mocks) {
+  'use strict';
+
+  mocks.factory('FileUploaderMock', function () {
+    var fileUploaderConstructor = function (settings) {
+      var response = {
```
AkA84: This mocked response should be in the `data/` folder, i think

org.civicrm.reqangular/test/mocks/services/file-uploader-mock.js
```diff
@@ -0,0 +1,54 @@
...
+            }
+          }
+        }
+      };
+
+      return {
+        uploadAll: function () {
+          this._onCompleteItem();
+        },
+        _onCompleteItem: function () {
```
AkA84: This is an internal function, there's no point in publishing it on the object interface

org.civicrm.reqangular/test/mocks/services/file-uploader-mock.js
```diff
@@ -0,0 +1,54 @@
...
+          this.onCompleteItem({}, response);
+          this.onCompleteAll();
+        },
+        onCompleteItem: function (item, response, status, headers) {},
+        onCompleteAll: function () {}
```
AkA84: same goes for `onCompleteItem` and `onCompleteAll `, you are using them internally only correct?

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -0,0 +1,77 @@
...
+
+        promise = fileUploadService.uploadAll(param);
+      });
+
+      afterEach(function () {
+        $rootScope.$apply();
+      });
+
+      it('calls file upload service endpoint', function () {
+        expect(fileUploadService.uploadAll).toHaveBeenCalledWith(param);
```
AkA84: What's the point of this test? You are manually calling `uploadAll(param)` on line 53
```js
promise = fileUploadService.uploadAll(param);
```
so what's the purpose of testing if that method had indeed been called?

org.civicrm.reqangular/src/common/services/file-upload.js
AkA84: Isn't `FileUploader` an external dependencies registered on a different module? Doesn't it mean then that `common.service` should have a dependency on that module?

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    describe('uploadAll()', function () {
+      var param;
+
+      beforeEach(function () {
+        spyOn(fileUploadService, 'uploadAll').and.callThrough();
+        spyOn(uploader, 'uploadAll').and.callThrough();
+        param = {
+          entityID: '12'
```
AkA84: Small objects can be one-liners
```js
param = { entityID: '12' };
```

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -0,0 +1,77 @@
...
+      afterEach(function () {
+        $rootScope.$apply();
+      });
+
+      it('calls file upload service endpoint', function () {
+        expect(fileUploadService.uploadAll).toHaveBeenCalledWith(param);
+      });
+
+      it('calls fileuploader endpoint', function () {
+        expect(uploader.uploadAll).toHaveBeenCalledWith();
```
AkA84: you are testing that that method had been called with something, but you're not testing the "something"

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -0,0 +1,77 @@
...
+      });
+
+      it('returns success', function () {
```
AkA84: this description is not helpful, you are obviously testing more than that, specifically what is the structure of the resolved value of the promise

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: * FileUploader callback that gets fired before each file item gets uploaded, useful 
          * to insert additional form params like entityId which are not available 
          * at the time of uploader creation.

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: * FileUploader callback that gets fired after each file item gets
          * uploaded, useful to obtain result of upload. The result of each
          * upload are aggregated in local results array.

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: * FileUploader callback to capture error during upload.

org.civicrm.reqangular/test/mocks/services/file-uploader-mock.js
dhavalsharma: there is no data folder in reqangular

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -1,48 +1,65 @@
...
+  module.factory('FileUpload', ['$q', 'FileUploader',
     function ($q, FileUploader) {
-      var fileUploader;
+
+      /**
+      * Helper to call throw on expressions by wrapping it on function
+      *
+      * @param {String} error message to throw
+      */
+      function _throw (msg) {
```
AkA84: The `_` prefix is not necessary, we're not using it anywhere to mark "private" functions

AkA84: Also, this function as it is is a bit pointless, given that it's just calling `throw` with the given parameter (as in, what is it's purpose then? You might as well call `throw msg` directly

The way you could make it useful would be if it was something like

```js
function error(property) {
  throw property + 'missing from the uploader settings';
}
```

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -1,48 +1,65 @@
...
-          uploaderSettings.url = 'url' in customSettings ? customSettings.url : '/civicrm/ajax/attachment';
-          uploaderSettings.formData = [{
-            entity_table: customSettings.entityTable,
-            crm_attachment_token: customSettings.crmAttachmentToken
-          }];
-
-          if (customSettings.queueLimit && typeof customSettings.queueLimit === 'number') {
-            uploaderSettings.queueLimit = customSettings.queueLimit;
+          if (!customSettings) {
+            throw 'custom settings object need to be defined in paramter';
```
AkA84: typo

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -67,10 +101,19 @@ define([
             $log.error(headers);
           };
 
+          /**
+          * FileUploader callback that gets fired after uploading each file item
+          * in a queue or a single file item.
+          *
```
AkA84: empty line

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -67,10 +101,19 @@ define([
...
+          */
           fileUploader.onCompleteAll = function () {
             deferred.resolve(results);
           };
 
+          /**
+          * FileUploader method to start uploading files.
+          *
```
AkA84: empty line

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -17,58 +17,80 @@ define([
...
       });
+
+      describe('missing params', function () {
+        it('throws error if no param is passed', function () {
+          expect(fileUpload.uploader).toThrow('custom settings object need to be defined in paramter');
+        });
+
+        it('throws error if entityTable is not defined', function () {
+          expect(function () {
+              fileUpload.uploader({ crmAttachmentToken: '123abc' })
```
AkA84: wrong indentation

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -17,58 +17,80 @@ define([
...
+            })
+            .toThrow('entityTable missing from custom settings');
+        });
+
+        it('throws error if crmAttachmentToken is not defined', function () {
+          expect(function () {
+              fileUpload.uploader({ entityTable: 'civicrm_hrleaveandabsences_leave_request' })
```
AkA84: wrong indentation

org.civicrm.reqangular/test/test-main.js
```diff
@@ -60,7 +63,8 @@ require.config({
         'common/lodash': srcPath + '/vendor/lodash.min',
         'common/ui-select': srcPath + '/vendor/angular/select',
         'common/vendor/perfect-scrollbar': srcPath + '/vendor/perfect-scrollbar.min',
-        'common/text-angular-sanitize': srcPath + '/vendor/angular/textAngular-sanitize.min'
+        'common/text-angular-sanitize': srcPath + '/vendor/angular/textAngular-sanitize.min',
+        'common/angularFileUpload': srcPath + '/vendor/angular/angular-file-upload'
```
AkA84: use a dasherized format
```
common/angular-file-upload
```

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -1,48 +1,65 @@
 define([
   'common/modules/services',
+  'common/angularFileUpload'
 ], function (module) {
   'use strict';
 
-  module.factory('FileUploadService', ['$q', 'FileUploader',
+  module.factory('FileUpload', ['$q', 'FileUploader',
```
AkA84: Same question I didn't get an answer for

> Isn't FileUploader an external dependencies registered on a different module? Doesn't it mean then that common.service should have a dependency on that module?

dhavalsharma: added `  'common/angularFileUpload'` here only and not in common.services as angularFileUpload is not relevant to other common services

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: without prefix it will confuse with throw keyword

dhavalsharma: throw msg cannot be called directly in expression like `entity_table: customSettings.entityTable || throw 'entityTable missing from custom settings'` will not work

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -25,96 +44,76 @@ define([
...
-                snakeCaseKey: additionalFormData[key]
-              });
-            }
-          };
+           * Uploads all files in queue updating with additional form data.
+           * Wraps the default `uploadAll` in a custom method so we can
+           * return a promise
+           *
+           * @param {Object} additionalFormData that has keys like entityID or description
+           *
```
AkA84: Unnecessary empty line

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -25,96 +44,76 @@ define([
...
-          };
+            return function (additionalFormData) {
+              /**
+               * FileUploader callback that gets fired before each file item gets
+               * uploaded, useful to insert additional form params like entityId
+               * which are not available at the time of uploader creation.
+               *
+               * @param {Object} item file item being uploaded
+               */
+              uploader.onBeforeUploadItem = function (item) {
```
AkA84: Please always prefer a functional approach if possible, make it for a terser code
```js
uploader.onBeforeUploadItem = function (item) {
  _.each(additionalFormData, function (val, key) {
    item.formData[_.snakeCase(key)] = val;
  });
};
```

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -25,96 +44,76 @@ define([
...
+               * FileUploader callback to capture error during upload.
+               *
+               * @param {Object} item file item being uploaded
+               * @param {Object} response obtained from server for upload
+               * @param {Number} status
+               * @param {Object} headers
+               */
+              uploader.onErrorItem = function (item, response, status, headers) {
+                deferred.reject('Could not upload file: ' + item.file.name);
+                logError(item, response, status, headers);
```
AkA84: First log, then reject

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -90,6 +89,32 @@ define([
...
+
+      describe('onBeforeUploadItem()', function () {
+        beforeEach(function () {
+          spyOn(uploader, 'onBeforeUploadItem').and.callThrough();
+          uploader.onBeforeUploadItem({ formData: [] });
+        });
+
+        it('gets called', function () {
+          //verify if method is getting called and hence a simple case
+          expect(uploader.onBeforeUploadItem).toHaveBeenCalled();
```
AkA84: Again, what's the point of this? You are calling this method yourself, how can this test ever fail?

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -90,6 +89,32 @@ define([
...
+        var fileItem = { file: { name: 'filename' } };
+
+        beforeEach(function () {
+          spyOn(uploader, 'onErrorItem').and.callThrough();
+          promise = uploader.onErrorItem(fileItem);
+        });
+
+        it('gets called', function () {
+          //verify if method is getting called and hence a simple case
+          expect(uploader.onErrorItem).toHaveBeenCalled();
```
AkA84: Again, what's the point of this? You are calling this method yourself, how can this test ever fail?

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -90,6 +89,32 @@ define([
           expect(firstObject.file).toBeDefined();
         });
       });
+
+      describe('onBeforeUploadItem()', function () {
```
AkA84: You are testing only if this method gets called (and you're testing that with un-falsifiable test), but not what the method is supposed to be doing

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -90,6 +89,32 @@ define([
...
+          uploader.onBeforeUploadItem({ formData: [] });
+        });
+
+        it('gets called', function () {
+          //verify if method is getting called and hence a simple case
+          expect(uploader.onBeforeUploadItem).toHaveBeenCalled();
+        });
+      });
+
+      describe('onErrorItem()', function () {
```
AkA84: You are testing only if this method gets called (and you're testing that with un-falsifiable test), but not what the method is supposed to be doing

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -71,27 +70,23 @@ define([
            * return a promise
            *
            * @param {Object} additionalFormData that has keys like entityID or description
-           *
            * @return {Promise} that resolves to result if successful else error
            */
           uploader.uploadAll = (function () {
-            oldUploadAll = uploader.uploadAll;
+              oldUploadAll = uploader.uploadAll;
```
AkA84: 4 spaces indentation

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -89,32 +89,6 @@ define([
           expect(firstObject.file).toBeDefined();
         });
       });
-
-      describe('onBeforeUploadItem()', function () {
```
AkA84: You need to test both `onBeforeUploadItem` and `onErrorItem`

PR #958, opened 2016-06-01
Pchr 773 custom ui select

danilo-compucorp:
This PR introduces a new theme called "civihr-ui-select" for the "ui-select" directive.
It's based on the "select2" theme, with bootstrap look-and-feel, and it contains styles overrides in order to look properly in civihr, either in "appraisals" (which uses the common bootstrap theme) or tasks and assignments (which for the moment, has it's own version of the theme).

Examples:

* Appraisals (multi select):
appraisals-multiple

* Tasks and Assignments (multi select):
t a-multiple

* Appraisals (single select):
appraisals-single

* Tasks and Assignments (single select):
t a-single

PR #1592, opened 2017-03-08
PCHR-1912: Upgrade ui.select

AkA84:
# Problem
The current `ui-select` dropdowns flicker when looking up data via AJAX. This is a known issue found in the older version of the library

_(There is no gif showing the problem because the flicker does not get recorded on a gif)_

# Solution
Upgrading `ui-select` to the latest version `0.19.5` solves the problem

### Bug fix for when there are no items to show
The later versions of the library has a bug that causes the search bar not to be shown in case there are no items to display in the first place, see [here](https://github.com/angular-ui/ui-select/issues/1652). The fix outlined in the linked issue solves the problem

### Support custom theme in xeditable
The [angular-xeditable](https://vitalets.github.io/angular-xeditable) library apparently has a bug that causes the `[theme]` attribute given to the `[editable-ui-select]` directive to be ignored. This was a problem for us given that we have a custom theme for ui-select (`civihr-ui-select`). I could no find any issue mentioned in the repo, but at the same time i could not make this work as shown in the [documentation](https://vitalets.github.io/angular-xeditable/#uiselect), so I fixed the library directly
```js
render: function () {
  this.parent.render.call(this);
  this.inputEl.append(editableUtils.rename('ui-select-match', this.attrs.$matchElement));
  this.inputEl.append(editableUtils.rename('ui-select-choices', this.attrs.$choicesElement));
  this.inputEl.removeAttr('ng-model');
  this.inputEl.attr('ng-model', '$parent.$parent.$data');

  /* CUSTOM */
  this.inputEl.attr('theme', this.attrs.theme);
}
```

### Hide original caret on "clear" icon
There is a caret icon showing underneath our custom "clear" icon. It has now been hidden

#### Before
before

#### After
after

# Problem
The scenarios in BackstopJS tended to time out randomly

# Solution
When building the temporary config file that is then feed to BackstopJS, we are now forcing a 6000ms delay to each scenario (unless there is already a delay set for it). The delay is sufficient to avoid times out
```js
.map(function (scenario) {
  return _.assign(scenario, { delay: scenario.delay || 6000 });
})
```

# Update 16/03
The custom `civihr-ui-select` theme had two problems that went unnoticed

1. The result lists would not adapt the height on further search refinements
unresponsive-height

2. The result lists would show a horizontal scrollbar
horizontal scroll

Both issues were caused by the usage of the perfect-scrollbar plugin on `ui.select`, introduced in #958 , via decorators. I guess this was an attempt to make the scrollbars look the same across all browsers (but i think it was mainly done for IE).

The plugin inserts a couple of new DOM elements in the select dropdown, causing those issues. Removing the plugin from `ui.select` solves the issue


AkA84:
@deb1990 i was thinking about it, but then i remember that we are using that library somewhere else. so it's safer to just not use it on ui.select instead of removing it completely

PR #1594, opened 2017-03-08
PCHR-1892: Cleanup orphan contract leave

mickadoo:
## Problem

A leave type can be deleted only if it has no approved absences related to it. However if the leave type is used only in a contract then it can be deleted. In a [related PR](https://github.com/civicrm/civihr/pull/1290) this was fixed by deleting all entries in `civicrm_hrjobcontract_leave` related to the deleted type. The problem here was that the API call to fetch them was being limited to 25 entries.

## Solution

- Remove the limit on the `civicrm_api3('HRJobLeave', 'get')` API call
- Add an upgrader to remove all entries from `civicrm_hrjobcontract_leave` that do not have a related entry in `civicrm_hrabsence_type`

### Notes

- I removed `sequential` from the API call since it's not relevant

### DIFF
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
@@ -88,10 +88,8 @@ class CRM_HRAbsence_BAO_HRAbsenceType extends CRM_HRAbsence_DAO_HRAbsenceType {
   private static function createActivityType($activityTypes, $params, $isDebit) {
     $weight = count($activityTypes["values"]);
     if ($isDebit) {
-      $valueId = $params['debit_activity_type_id'];
       $valueLabel = $params['title'];
     } else {
-      $valueId = $params['credit_activity_type_id'];
       $valueLabel = ts('%1 (Credit)', array(1 => $params["title"]));
     }
 
--- a/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
@@ -267,7 +267,7 @@ class CRM_HRAbsence_BAO_HRAbsenceType extends CRM_HRAbsence_DAO_HRAbsenceType {
   private static function deleteLeaveEntitlements($absenceTypeId) {
     $result = civicrm_api3('HRJobLeave', 'get', [
       'leave_type' => $absenceTypeId,
-      'jobcontract_revision_id' => array('>' => 0),
+      'jobcontract_revision_id' => ['>' => 0],
       'options' => ['limit' => 0]
     ]);
     foreach ($result['values'] as $currentLeave) {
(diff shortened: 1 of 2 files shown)


hrabsence/CRM/HRAbsence/BAO/HRAbsenceType.php
```diff
@@ -266,18 +266,17 @@ private static function countLeaveRequestsForAbsenceType(CRM_HRAbsence_DAO_HRAbs
    */
   private static function deleteLeaveEntitlements($absenceTypeId) {
     $result = civicrm_api3('HRJobLeave', 'get', [
-      'sequential' => 1,
       'leave_type' => $absenceTypeId,
-      'jobcontract_revision_id' => array('>' => 0)
+      'jobcontract_revision_id' => array('>' => 0),
```
mickadoo: sure, wasn't sure since this is what it was but definitely prefer short syntax

hrabsence/CRM/HRAbsence/Upgrader.php
```diff
@@ -551,4 +551,18 @@ public function upgrade_1401() {
...
+   * entries for that type. This upgrade is to remove legacy data where civicrm_hrjobcontract_leave points to a
+   * civicrm_hrabsence_type that does not exist
+   *
+   * @see https://compucorp.atlassian.net/browse/PCHR-1892
+   */
+  public function upgrade_1402() {
+    $deleteQuery = "DELETE FROM civicrm_hrjobcontract_leave WHERE leave_type NOT IN (
+      SELECT id FROM civicrm_hrabsence_type
+    );";
+    CRM_Core_DAO::executeQuery($deleteQuery);
```
mickadoo: ah, nice. I was hacking some api calls to test it but this is much nicer

hrabsence/CRM/HRAbsence/Upgrader.php
```diff
@@ -551,4 +551,20 @@ public function upgrade_1401() {
...
+
+  /**
+   * Previously when a leave type was deleted it was only deleting the first 25 civicrm_hrjobcontract_leave
+   * entries for that type. This upgrade is to remove legacy data where civicrm_hrjobcontract_leave points to a
+   * civicrm_hrabsence_type that does not exist
+   *
+   * @see https://compucorp.atlassian.net/browse/PCHR-1892
+   */
+  public function upgrade_1402() {
+    $deleteQuery = "DELETE FROM civicrm_hrjobcontract_leave WHERE leave_type NOT IN (
```
omarabuhussein: @davialexandre  here he is talking to another extension (job contract extension) .. I know it will hurt the performance but this an upgrader that will be executed  only once .. so I suggest to use the API here .. and for absence types list , he can fetch them via BAO .

hrabsence/CRM/HRAbsence/Upgrader.php
```diff
@@ -560,10 +560,22 @@ public function upgrade_1401() {
...
+    );
+
+    if (empty($absenceTypes['values'])) {
+      return true;
+    }
+
+    $absenceTypeNames = array_column($absenceTypes['values'], 'name');
+
+    civicrm_api3('HRJobLeave', 'get', [
+      'leave_type' => ['NOT IN' => $absenceTypeNames],
```
davialexandre: This doesn't look right. In leave_type we store the Absence Type ID, not the name

PR #1605, opened 2017-03-09
AI-37: Adjust accordion header font weight

brunoqueiros:
**Before**

screenshot-3

**After**

captura de tela 2017-03-09 as 11 13 24


org.civicrm.bootstrapcivicrm/.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
meladawy: @brunoqueiros do u need this change ?

PR #1606, opened 2017-03-09
AI-38: Adjust sticky header styling

brunoqueiros:
**Before**

screenshot-3 1

**After**

captura de tela 2017-03-09 as 12 33 11


meladawy:
Can u adjust the vertical align for table cells to be in the middle?

PR #1608, opened 2017-03-09
PCHR-2044: Extend LeaveRequest.get and LeaveRequest.getFull API to support the 'unassigned' Parameter

tunbola:
This PR extends LeaveRequest.get and LeaveRequest.getFull API endpoints to support the 'unassigned' Boolean Parameter.

The 'unassigned' parameter returns Leave Requests for contacts without active Leave managers, when false, it returns Leave Requests for contacts with active Leave managers for LeaveRequest.get and LeaveRequest.getFull
When the parameter is absent, LeaveRequest.get and LeaveRequest.getFull returns Leave Requests for all contacts irrespective of whether they have an active Leave manager or not.

Note that contacts without Leave managers and contacts having an inactive relationship with leave managers both fall under the category of contacts without active leave managers.

Different tests were also added to ensure that ACL's and other existing parameters on the API's were respected.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -74,6 +74,9 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
    * @param \CRM_Utils_SQL_Select $customQuery
    */
   private function addWhere(CRM_Utils_SQL_Select $customQuery) {
+    $hasUnassignedAsTrue = isset($this->params['unassigned']) && $this->params['unassigned'] == true;
+    $hasUnassignedAsFalse = isset($this->params['unassigned']) && $this->params['unassigned'] == false;
+
     $conditions = [
       'jc.deleted = 0',
       '(
@@ -86,17 +89,27 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
         )'
     ];
 
+    if($hasUnassignedAsFalse) {
+      $conditions[] = implode(' AND ', $this->hasActiveLeaveManagerCondition());
+    }
+
+    if($hasUnassignedAsTrue) {
+      $conditions[] = "NOT (" . implode(' AND ', $this->hasActiveLeaveManagerCondition()) . ") 
+                       OR (r.is_active IS NULL AND rt.is_active IS NULL)";
+    }
+
     if(!empty($this->params['managed_by'])) {
       $managerID = (int)$this->params['managed_by'];
-      $today =  '"' . date('Y-m-d') . '"';
-      $leaveApproverRelationshipTypes = $this->getLeaveApproverRelationshipsTypesForWhereIn();
 
-      $conditions[] = 'rt.is_active = 1';
-      $conditions[] = 'rt.id IN(' . implode(',', $leaveApproverRelationshipTypes) . ')';
-      $conditions[] = 'r.is_active = 1';
-      $conditions[] = "r.contact_id_b = {$managerID}";
-      $conditions[] = "(r.start_date IS NULL OR r.start_date <= {$today})";
-      $conditions[] = "(r.end_date IS NULL OR r.end_date >= {$today})";
+      if($hasUnassignedAsFalse) {
+        $conditions[] = "r.contact_id_b = {$managerID}";
+      }
+
+      if(!isset($this->params['unassigned'])) {
+        $activeLeaveManagerCondition = $this->hasActiveLeaveManagerCondition();
+        $activeLeaveManagerCondition[] = "r.contact_id_b = {$managerID}";
+        $conditions[] = implode(' AND ', $activeLeaveManagerCondition);
+      }
     }
 
     if(!empty($this->params['expired'])) {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -150,6 +158,15 @@ function _civicrm_api3_leave_request_getfull_spec(&$spec) {
     'type'         => CRM_Utils_Type::T_BOOLEAN,
     'api.required' => 0
   ];
+
+  $spec['unassigned'] = [
+    'name' => 'unassigned',
+    'title' => 'Unassigned only?',
+    'description' => 'Include only Leave Requests of contacts without active leave managers?',
+    'type' => CRM_Utils_Type::T_BOOLEAN,
+    'api.default' => FALSE,
```
davialexandre: Please remove this. It will set unassigned to false when unassigned is not present

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -91,6 +91,15 @@ function _civicrm_api3_leave_request_get_spec(&$spec) {
...
```
davialexandre: Please remove this

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -1312,6 +1312,406 @@ public function testGetFullReturnsOnlyTheOriginalAmountWhenTheExpiredParamIsNotP
...
+    $this->assertNotEmpty($result['values'][$leaveRequest1->id]);
+
+    $this->assertEquals(1, $resultGetFull['count']);
+    $this->assertNotEmpty($resultGetFull['values'][$leaveRequest1->id]);
+  }
+
+  public function testGetAndGetFullShouldReturnEmptyResponseForALoggedInLeaveManagerWhenUnassignedIsTrue() {
+    $manager1 = ContactFabricator::fabricate();
+    $this->registerCurrentLoggedInContactInSession($manager1['id']);
+    CRM_Core_Config::singleton()->userPermissionClass->permissions = ['access AJAX API'];
```
davialexandre: Please check my other comment about permissions and also apply it here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -1312,6 +1312,406 @@ public function testGetFullReturnsOnlyTheOriginalAmountWhenTheExpiredParamIsNotP
...
+      'from_date' => CRM_Utils_Date::processDate('2016-01-05'),
+      'to_date' => CRM_Utils_Date::processDate('2016-01-05'),
+      'from_date_type' => 1,
+      'to_date_type' => 1
+    ], true);
+
+    //No result is returned because the values on both managed_by and unassigned parameters contradict
+    //i.e you cannot get results for contacts with active managers with unassigned true at the site time
+    //its a contradiction.
+    $result = civicrm_api3('LeaveRequest', 'get', ['managed_by' => $manager1['id'], 'unassigned' => true]);
```
davialexandre: for this test, I think only one contact, contract and leave request is enough, since no matter the number of records, this combination of params will always result in an empty result set

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -141,9 +154,9 @@ private function addJoins(CRM_Utils_SQL_Select $query) {
       'INNER JOIN ' . HRJobDetails::getTableName() . ' jd ON jd.jobcontract_revision_id = jcr.details_revision_id'
     ];
 
-    if(!empty($this->params['managed_by'])) {
-      $joins[] = 'INNER JOIN ' . Relationship::getTableName() . ' r ON r.contact_id_a = a.contact_id';
-      $joins[] = 'INNER JOIN ' . RelationshipType::getTableName() . ' rt ON rt.id = r.relationship_type_id';
+    if(!empty($this->params['managed_by']) || isset($this->params['unassigned'])) {
+      $joins[] = 'LEFT JOIN ' . Relationship::getTableName() . ' r ON r.contact_id_a = a.contact_id';
```
davialexandre: I might be wrong here because I didn't expect the generated queries, but I have the impression that, since now this is a LEFT JOIN, the whole logic can be simplified:

- For unassigned = true: We filter out all the records where r.id IS NOT NULL (That is, those that have one relationship)
- For unassigned = false: We filter out all the records where r.id IS NULL (That is, those that don't have one relationship)
- For when unassigned is not present: Don't do anything. The query should just work like before.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -86,17 +89,27 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
         )'
     ];
 
+    if($hasUnassignedAsFalse) {
+      $conditions[] = implode(' AND ', $this->hasActiveLeaveManagerCondition());
```
tunbola: Yes because without the `$this->hasActiveLeaveManagerCondition())` returns an array and adding it to the conditions array makes it become an array of arrays.

davialexandre: can't we use array_merge instead and leave the concatenation job to civi (That is, we just pass an array of conditions and let civi do the job)?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -171,6 +184,12 @@ public function setReturnFullDetails($value) {
    * @return array
    */
   public function run() {
+    $hasUnassignedAsTrue = isset($this->params['unassigned']) && $this->params['unassigned'] == true;
+    $hasManagedBy = isset($this->params['managed_by']);
+
+    if ($hasManagedBy && $hasUnassignedAsTrue) {
```
tunbola: :)

PR #1626, opened 2017-03-14
PCHR-1980: Add upgrade method to add 'other task' task type

mickadoo:
## Overview

To give a more generic option for task types a 'Other Task' type will be added

### Notes

- I'm not sure if it should also be added to CRM_HRCase_DefaultCaseAndActivityTypes::$defaultActivityTypes and would like to hear from the reviewer what they think. If it is added here the upgrade will need to be changed since 'create' will create a duplicate.

### DIFF
--- a/hrcase/CRM/HRCase/DefaultCaseAndActivityTypes.php
+++ b/hrcase/CRM/HRCase/DefaultCaseAndActivityTypes.php
@@ -26,6 +26,7 @@ class CRM_HRCase_DefaultCaseAndActivityTypes {
       'Probation appraisal (start probation workflow)',
       'Confirm End of Probation Date',
       'Start Probation workflow',
+      'Other Task',
     ],
     'CiviDocument' => [
       'Government Photo ID',
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -142,6 +142,16 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
    * Upgrader to add new default task type: 'Other Task'
    */
   public function upgrade_1403() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'component_id' => "CiviTask",
+      'name' => "Other Task",
+      'option_group_id' => "activity_type",
+    ]);
+
+    if ($result['is_error'] || $result['count'] != 0) {
+      return true;
+    }
+
     civicrm_api3('OptionValue', 'create', [
       'option_group_id' => "activity_type",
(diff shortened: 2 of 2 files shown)


omarabuhussein:
would be better to add it CRM_HRCase_DefaultCaseAndActivityTypes::$defaultActivityTypes  and keep the upgrader , but inside the upgrader ensure that this task type is not already exist (via another get API call )

mickadoo:
@omarabuhussein sure, I can do this. I just want to be clear thought that this upgrader will only add the single 'Other Task' type, and not ensure that all other task types are present, because this would possibly revert  changes for existing clients.

omarabuhussein:
not sure if I get your question correctly but you just need to do this inside the upgrader : 

```php
  public function upgrade_1403() {
    $otherTaskType = civicrm_api3('OptionValue', 'get', [
      'option_group_id' => "activity_type",
      'component_id' => "CiviTask",
      'label' => "Other Task",
      'name' => "Other Task",
      'options' => ['limit' => 1]
    ]);
    
   if (empty($otherTaskType['id']) {
    civicrm_api3('OptionValue', 'create', [
      'option_group_id' => "activity_type",
      'component_id' => "CiviTask",
      'label' => "Other Task",
      'name' => "Other Task",
    ]);
   }

    return true;
  }
```php


the whole thing works like this : 

1- for exsiting site , only upgrade_1403 will get executed ( after applying pending upgrader) which will create this new task type.

2- for new sites and if you added this task type to  CRM_HRCase_DefaultCaseAndActivityTypes::$defaultActivityTypes , then this lines https://github.com/civicrm/civihr/blob/83f2b0e4bc897be6dfca89ff0ef166f94ac7e65c/hrcase/CRM/HRCase/Upgrader.php#L21 ( from 21~27) will run all the existing upgraders one by one .. so upgrade_1402 will be excuted first and this new task type will be created (see https://github.com/civicrm/civihr/blob/83f2b0e4bc897be6dfca89ff0ef166f94ac7e65c/hrcase/CRM/HRCase/Upgrader.php#L120 ) ... then upgrade_1403 will be executed , so in case you didn't checked if the task type is exist or not the a duplicate will be created .. so you just need to add the check for duplicate as you can see in the code snippet above to ensure this will not happen .

mickadoo:
@omarabuhussein yep, that was the plan. I just meant I don't want to loop through all `$defaultActivityTypes` inside the upgrader and make sure they're present - I will only focus on the new 'Other Task' type.

Thanks for the overview of the upgrader functionality

omarabuhussein:
yup exactly , you don't need to loop through all of them

PR #1638, opened 2017-03-15
PCHR-2026: Add is_deleted field to LeaveRequest.

tunbola:
This PR adds adds the is_deleted field to the LeaveRequest Entity.
The field is a boolean field and allows a LeaveRequest to be soft deleted by setting it to true.
A new softDelete function is created on the LeaveRequest BAO to soft delete leave requests and the leave request service delete method was updated to use this new method.

Also functions that get balance for entitlements and for leave requests for a given status were updated not to count soft deleted leave requests including functions for calculating and recalculating expired balance changes.

Also a new method was added to the leaveRequestRights service 'canDeleteFor' to check access as to who can delete a leave request.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -646,6 +646,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $leaveRequest->contact_id = (int)$contactID;
     $leaveRequest->from_date = date('Ymd', strtotime($publicHoliday->date));
     $leaveRequest->request_type = self::REQUEST_TYPE_PUBLIC_HOLIDAY;
+    $leaveRequest->is_deleted = 0;
 
     $leaveRequest->find(true);
     if($leaveRequest->id) {
@@ -681,15 +682,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
         ON lrd.leave_request_id = lr.id
       INNER JOIN {$leaveBalanceChangeTable} lbc
         ON lbc.source_id = lrd.id AND lbc.source_type = %1
-      WHERE lr.contact_id = %2
+      WHERE lr.contact_id = %2 AND lr.is_deleted = 0
+      AND lrd.date BETWEEN %3 AND %4
     ";
-
-    if (!empty($toDate)){
-      $query .= ' AND lrd.date BETWEEN %3 AND %4';
-    }
-    else{
-      $query .= ' AND lrd.date = %3';
-    }
 
     if (is_array($leaveRequestStatus) && !empty($leaveRequestStatus)) {
       array_walk($leaveRequestStatus, 'intval');
@@ -704,11 +699,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       1 => [LeaveBalanceChange::SOURCE_LEAVE_REQUEST_DAY, 'String'],
       2 => [$contactID, 'Integer'],
       3 => [CRM_Utils_Date::processDate($fromDate, null, false, 'Y-m-d'), 'String'],
+      4 => [CRM_Utils_Date::processDate($toDate, null, false, 'Y-m-d'), 'String'],
       5 => [self::REQUEST_TYPE_PUBLIC_HOLIDAY, 'String']
     ];
-    if (!empty($toDate)) {
-      $params[4] = [CRM_Utils_Date::processDate($toDate, null, false, 'Y-m-d'), 'String'];
-    }
 
     $leaveRequest = CRM_Core_DAO::executeQuery($query, $params, true, self::class);
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -37,6 +37,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     if($validate){
       self::validateParams($params);
     }
+    unset($params['is_deleted']);
+
     $instance = new self();
     $instance->copyValues($params);
 
@@ -67,6 +69,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    */
   public static function validateParams($params) {
     self::validateMandatory($params);
+    self::validateLeaveRequestSoftDeleteDuringUpdate($params);
     self::validateRequestType($params);
     self::validateTOILFieldsBasedOnRequestType($params);
     self::validateSicknessFieldsBasedOnRequestType($params);
@@ -111,6 +114,25 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
           $field
         );
       }
+    }
+  }
+
+  /**
+   * A method for validating that a LeaveRequest cannot be soft deleted
+   * during an update on the BAO.
+   *
+   * @param array $params
+   *   The params array received by the create method
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
+   */
+  public static function validateLeaveRequestSoftDeleteDuringUpdate($params) {
+    if (isset($params['id']) && !empty($params['is_deleted'])) {
+      throw new InvalidLeaveRequestException(
+        'Leave Request can not be soft deleted during an update, use the delete method instead!',
+        'leave_request_cannot_soft_delete',
+        'is_deleted'
+      );
     }
   }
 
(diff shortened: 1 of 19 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -966,6 +967,19 @@ public static function deleteAllNonExpiredTOILRequestsForAbsenceType($absenceTyp
   }
 
   /**
+   * Soft Deletes the LeaveRequest with the given ID by setting the is_deleted column to 1
+   *
+   * @param int $id The ID of the LeaveRequest to be soft deleted
```
davialexandre: Please put the param description in a new line

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -966,6 +967,19 @@ public static function deleteAllNonExpiredTOILRequestsForAbsenceType($absenceTyp
...
+   *
+   * @return boolean
```
davialexandre: This method doesn't return anything

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -794,6 +794,21 @@ public function testTheLeaveRequestBreakdownReturnsOnlyTheLeaveBalanceChangesOfT
     }
   }
 
+  public function testTheLeaveRequestBreakdownReturnsEmptyArrayForSoftDeletedLeaveRequests() {
```
davialexandre: A better name for this would be "testTheLeaveRequestBreakdownShouldNotReturnSoftDeletedLeaveRequests"

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -115,38 +115,32 @@ public function testCreateDoesNotDuplicateLeaveBalanceChangesOnUpdate() {
     $this->assertCount(11, $balanceChanges);
   }
 
-  public function testDeleteDeletesTheLeaveRequestItsBalanceChangesAndDates() {
+  public function testDeleteSoftDeletesTheLeaveRequestAndThatFetchingItsBalanceChangesAndDatesReturnsEmpty() {
```
davialexandre: This test looks very weird to me. The only thing that the delete() method does is call the softDelete() method on the BAO, so that should be the only thing being tested here. There should be only a single assert to check if the is_deleted value was updated to 1. The method name should not mention anything about dates or balance changes (it was ok before, because the previous version also deleted these things)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -974,12 +974,35 @@ public static function deleteAllNonExpiredTOILRequestsForAbsenceType($absenceTyp
...
+   * finding a leave request record.
+   *
+   * @param int $id
+   *   ID of leave request to be found
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest
+   *
+   * @throws \Exception
+   */
+  public static function findRecordById($id) {
```
davialexandre: is the old findById still being used anywhere?

If you had to replace all the `findById` (for LeaveRequests) occurrences with `findRecordById`, then I think it would make more sense to override the `findById` method in the LeaveRequest BAO to only return non deleted leave requests

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestDateTest.php
```diff
@@ -41,48 +41,42 @@ public function testALeaveRequestCantHaveMoreThanOneLeaveRequestDateWithTheSameD
...
-
-    LeaveRequestDate::create([
-      'date' => date('YmdHis', strtotime('+1 day')),
-      'leave_request_id' => 2,
-    ]);
-
-    LeaveRequestDate::create([
-      'date' => date('YmdHis', strtotime('+2 days')),
-      'leave_request_id' => 2,
+    $leaveRequest1 = LeaveRequestFabricator::fabricateWithoutValidation([
```
davialexandre: The fact that this test broke with the changes is an indicator that `getDatesForLeaveRequest` should not have changed. If you compare the previous version of it with the new one, you can clearly see that now the method logic depends on details specific to the LeaveRequest (the reason why you now need the LeaveRequest Fabricator, where before all you needed was dates).

Also, if we think about this as an hierarchy, we can say that LeaveRequest is the parent and LeaveRequestDate is the child. Usually, in situations like this, it's ok for the parent to know details about the child, but it's not ok for the child to know details about the LeaveRequest (in this case, to know that it's parent has been deleted).

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestDateTest.php
```diff
@@ -92,31 +86,32 @@ public function testGetDatesForLeaveRequestReturnsEmptyArrayForANonExistingLeave
...
   public function testDeleteDatesDeletesTheDatesForASpecificLeaveRequest() {
-    LeaveRequestDate::create([
-      'date' => date('YmdHis'),
-      'leave_request_id' => 1,
-    ]);
-
-    LeaveRequestDate::create([
-      'date' => date('YmdHis'),
-      'leave_request_id' => 2,
+    $leaveRequest1 = LeaveRequestFabricator::fabricateWithoutValidation([
```
davialexandre: you should not need the LeaveRequest Fabricator here as well (see my comment about dependencies on the previous test)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -109,7 +109,7 @@ public function testGetBalanceForEntitlementCanSumAllTheBalanceChangesForAGivenE
     $this->assertEquals(0.8, LeaveBalanceChange::getBalanceForEntitlement($entitlement));
   }
 
-  public function testBalanceForEntitlementCanSumOnlyTheBalanceChangesForLeaveRequestWithSpecificStatuses() {
+  public function testBalanceForEntitlementCanSumOnlyTheBalanceChangesForLeaveRequestWithSpecificStatusesAndDoesNotSumForSoftDeletedLeaveRequests() {
```
davialexandre: Let's try to make tests as short as possible (I know it's a bit difficult sometimes) and avoid testing a lot of things at once. It's probably better to add a new test to only check that `balanceForEntitlement` will not include deleted leave requests.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -119,49 +119,65 @@ public function testBalanceForEntitlementCanSumOnlyTheBalanceChangesForLeaveRequ
     $this->createLeaveBalanceChange($entitlement->id, 10);
     $this->assertEquals(10, LeaveBalanceChange::getBalanceForEntitlement($entitlement));
 
-    $this->createLeaveRequestBalanceChange(
-      $entitlement->type_id,
-      $entitlement->contact_id,
-      $leaveRequestStatuses['Cancelled'],
-      date('Y-m-d', strtotime('-10 days'))
-    );
+    LeaveRequestFabricator::fabricateWithoutValidation([
```
davialexandre: The move from helper methods to Fabricator was a good idea! Please keep it when adding a new test to check that the method does not include deleted records

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -721,6 +738,16 @@ public function testLeaveRequestBalanceForEntitlementShouldIncludeTOILBalanceCha
...
+    $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $entitlement->contact_id,
+      'type_id' => $entitlement->type_id,
+      'from_date' => CRM_Utils_Date::processDate('+11 days'),
+      'to_date' => CRM_Utils_Date::processDate('+11 days')
+    ], true);
+
+    //The leave request is soft deleted and therefore will nnot be accounted for in the calculation
+    LeaveRequest::softDelete($leaveRequest->id);
+
```
davialexandre: Same thing here and for the other tests. Probably better to create a separate test for testing things related to deleted records

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -830,14 +830,16 @@ public function testFindOverlappingLeaveRequestsForOneOverlappingLeaveRequest()
     $this->assertEquals($leaveRequest1->id, $overlappingRequests[0]->id);
   }
 
-  public function testFindOverlappingLeaveRequestsForMoreThanOneOverlappingLeaveRequests() {
+  public function testFindOverlappingLeaveRequestsDoesNotCountSoftDeletedLeaveRequestAsOverlappingLeaveRequest() {
```
davialexandre: Separate the test about delete records here as well

PR #1640, opened 2017-03-16
AI-98: Manage Groups styling

meladawy:
Improve the styling of manage groups table

Before:
ai-98-before

After:
ai-98-after


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_manage-groups.scss
```diff
@@ -163,39 +163,15 @@
...
+    .crm-no-children {
+      padding-left: 0;
     }
 
-    .crm-editable-enabled {
-      white-space: normal !important;
-    }
-
-    .crm-group-group_links {
-      > .crm-hover-button,
```
deb1990: `crm-hover-button`and `btn-slide` styles are removed. That is going to change the look of the buttons right?

meladawy: All these styling created in a mixing "mx-expandable-table()" that i have included in line #166

PR #1643, opened 2017-03-16
PCHR-2046:  Don't allow deleting a Work Pattern that has ever been used by a contact

tunbola:
This PR changes the way a WorkPattern is deleted:

A WorkPattern that fulfils any of the following criteria is not allowed to be deleted, instead an exception is thrown.

- If the WorkPattern is the default one
- If there's at least one ContactWorkPattern where the pattern_id is the ID of the WorkPattern being deleted.

A new WorkPattern service is created to implement these new changes

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPattern.php
@@ -17,7 +17,7 @@ class CRM_HRLeaveAndAbsences_Service_WorkPattern {
    */
   public function workPatternHasEverBeenUsed($workPatternID) {
     if ($this->isDefaultWorkPattern($workPatternID) ||
-      $this->workPatternIsLinkedToAContactWorkPattern($workPatternID))
+      $this->workPatternIsLinkedToAContact($workPatternID))
     {
       return true;
     }
@@ -63,7 +63,7 @@ class CRM_HRLeaveAndAbsences_Service_WorkPattern {
    *
    * @return boolean
    */
-  private function workPatternIsLinkedToAContactWorkPattern($workPatternID) {
+  private function workPatternIsLinkedToAContact($workPatternID) {
     $contactWorkPattern = new ContactWorkPattern();
     $contactWorkPattern->pattern_id = $workPatternID;
     $contactWorkPattern->find();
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPattern.php
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Checks if the work pattern is linked to at least one contact work pattern
+   *
+   * @param int $workPatternID
+   *
+   * @return boolean
+   */
+  private function workPatternIsLinkedToAContactWorkPattern($workPatternID) {
```
davialexandre: Let's make this method name a bit shorter. The important thing here is to know if a work pattern is linked to a contact. How it's linked (using a ContactWorkPattern) is not really important. With that said, I believe a better name would be `workPatternIsLinkedToAContact`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternTest.php
```diff
@@ -0,0 +1,67 @@
...
+  public function tearDown() {
+    CRM_Core_DAO::executeQuery('SET foreign_key_checks = 1;');
+  }
+
+  public function testWorkPatternHasEverBeenUsedReturnsTrueWhenWorkPatternIsTheDefaultWorkPattern() {
+    $workPattern = WorkPatternFabricator::fabricate(['is_default' => 1]);
+    $this->assertTrue($this->workPatternService->workPatternHasEverBeenUsed($workPattern->id));
+  }
+
+  public function testWorkPatternHasEverBeenUsedReturnsTrueWhenWorkPatternIsLinkedToAContactWorkPattern() {
```
davialexandre: Given the method name change I suggested, you'll probably want to change this related test name as well

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternTest.php
```diff
@@ -0,0 +1,67 @@
...
+    $workPattern = WorkPatternFabricator::fabricate(['is_default' => 0]);
+    $this->assertFalse($this->workPatternService->workPatternHasEverBeenUsed($workPattern->id));
+  }
+
+  /**
+   * @expectedException UnexpectedValueException
+   * @expectedExceptionMessage Work pattern cannot be deleted because it is used by one or more contacts
+   */
+  public function testDeleteThrowsAnExceptionWhenAttemptingToDeleteAWorkPatternThatHasBeenUsed() {
+    $workPattern = WorkPatternFabricator::fabricate(['is_default' => 1]);
```
davialexandre: What do you think about adding a similar test, but to check that a work pattern cannot be deleted if it's not the default but it's linked to a contact?

I know that the logic revolves around the the `workPatternHasEverBeenUsed` method, but since we're not mocking it, it's probably better to have both cases covered in case the logic inside `workPatternHasEverBeenUsed` changes.

PR #1644, opened 2017-03-16
PCHR-2082: Comments ACL Improvements

davialexandre:
This PR improves the way ACLs are handled for Comments.

In CiviCRM, you can add ACL clauses using `hook_civicrm_selectWhereClause`. But there is a limitation with that: All the ACLs clauses added by all the extensions will always be added to the final query. That is, one clause specific to Leave Requests will also be added when searching for Comments linked to Contracts, for example.

Given that Comments can be linked to multiple different entities, we need a way to add more complex ACLs rules and that it can be possible to add them conditionally (for example, only add them when searching for Comments linked to a Leave Request). For this reason, a new custom hook named `hook_hrcomments_selectWereClause` was added.

The hook is invoked inside the `CommentSelect` class, right before the moment it builds the where clause to be added to the SELECT query. As parameters, it accepts:
-` $conditions`: an array of conditions that will be added to the query. This param is expected to be received as a reference, so it can be modified by other hooks
- `$params`: an array of params passed to the query. Ideally, this should not be changed, but in case that happens, the query will not be affected and the query will work with the original params.

We then implement this hook in leave and absences in order to filter the comments linked to leave requests according to these rules:
- staff members: can see any comments added to their Leave Requests (including admin and manager comments)
- leave approvers: can see any comments added to the Leave Requests of their managees
- admins: can see any comments added to any Leave Requests

Finally, in order for all this to work, we also use `hook_civicrm_selectWhereClause` to remove all the default ACLs added to Comments. We need this in order to make sure only the more specific ones (added with hook_hrcomments_selectWhereClause) will be applied. 

Some missing actions were added to alterAPIPermissions. Some old ones were removed.


uk.co.compucorp.civicrm.hrcomments/CRM/HRComments/API/Query/CommentSelect.php
```diff
@@ -45,6 +45,8 @@ private function buildCustomQuery() {
    * @param \CRM_Utils_SQL_Select $customQuery
    */
   private function addWhere(CRM_Utils_SQL_Select $customQuery) {
+    $conditions = $this->invokeSelectWhereClauseHook();
```
tunbola: We should probably add another function to return the conditions here rather than calling the `invokeSelectWhereClauseHook` directly. Something like:

`public function getConditionsForACL() {
 return $this->invokeSelectWhereClauseHook();
}`

The reason for this is that:
(1) Its more readable
(2) In the future the `invokeSelectWhereClauseHook()` can return more than just the conditions array and may need to return other parameters.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/ACL/LeaveRequestCommentsWhereClause.php
```diff
@@ -0,0 +1,59 @@
...
+      return $conditions;
+    }
+
+    $whereClauses = $this->getLeaveInformationACLWhereConditions('lr.contact_id');
+    $whereClauses .= ' AND (lr.is_deleted = 0)';
+
+    $conditions[] = "
+      a.entity_id IN (
+       SELECT lr.id
+       FROM civicrm_hrleaveandabsences_leave_request lr
```
tunbola: It will be better to use the `getTableName()` function of the respective BAO's for this query.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCommentTest.php
```diff
@@ -175,20 +176,268 @@ public function testGetReturnsAssociatedCommentsForLeaveRequest() {
     $this->assertEquals($expected1, $result['values']);
   }
 
+  public function testGetReturnsOnlyCommentsOfLeaveRequestsOfTheCurrentLoggedInUserWhenTheLoggedInUserIsAStaffMember() {
+    $entityName = 'LeaveRequest';
+
+    $contact1 = ContactFabricator::fabricate();
+    $contact2 = ContactFabricator::fabricate();
```
tunbola: We don't really need Contact fabricators here. We can just use static ids. Will save some test time.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCommentTest.php
```diff
@@ -175,20 +176,268 @@ public function testGetReturnsAssociatedCommentsForLeaveRequest() {
...
+    ]);
+    $this->assertEmpty($result['values']);
+  }
+
+  public function testGetReturnsAllTheCommentsOfLeaveRequestsWhenTheLoggedInUserIsAnAdmin() {
+    $entityName = 'LeaveRequest';
+
+    $admin = ContactFabricator::fabricate();
+    $contact1 = ContactFabricator::fabricate();
+    $contact2 = ContactFabricator::fabricate();
```
tunbola: We can do without Contact fabricators here too.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCommentTest.php
```diff
@@ -175,20 +176,268 @@ public function testGetReturnsAssociatedCommentsForLeaveRequest() {
...
+    $this->assertCount(3, $result['values']);
+    $this->assertNotEmpty($result['values'][$comment1->id]);
+    $this->assertNotEmpty($result['values'][$comment2->id]);
+    $this->assertNotEmpty($result['values'][$comment3->id]);
+  }
+
+  public function testGetDoesNotReturnsCommentsOfSoftDeletedLeaveRequests() {
+    $entityName = 'LeaveRequest';
+
+    $contact1 = ContactFabricator::fabricate();
```
tunbola: We can do without Contact fabricators here.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -357,6 +360,47 @@ function hrleaveandabsences_civicrm_tabset($tabsetName, &$tabs, $context) {
   }
 }
 
+/**
+ * Implementation of hook_civicrm_selectWhereClause
+ *
+ * @param $entity
+ * @param $clauses
+ */
```
tunbola: Missing Param type

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -110,17 +110,20 @@ function hrleaveandabsences_civicrm_permission(&$permissions) {
...
-    'isvalid' => ['leave_request', 'sickness_request', 't_o_i_l_request'],
-    'getfull' => ['leave_request', 'sickness_request'],
+    'isvalid' => ['leave_request'],
+    'getfull' => ['leave_request'],
     'calculatetoilexpirydate' => ['absence_type'],
-    'getleavemanagees' => ['contact']
+    'getleavemanagees' => ['contact'],
+    'getcomment' => ['leave_request'],
+    'addcomment' => ['leave_request'],
+    'deletecomment' => ['leave_request']
```
tunbola: We should probably add the new actions here to `api_v3_ApiPermissionTest`

tunbola:
Typo in PR description: s/ it accetps/ it accepts

PR #1657, opened 2017-03-20
PCHR-1994: Add New Pay Cycles

mickadoo:
## Problem

Users should be able to choose Bi-weekly and Bi-Monthly pay cycles

## Solution

- Add the pay cycles in the option
- Update the salary calculation in the frontend for the new pay cycle options

## Notes

- Backend changes were just adding an upgrader to insert the new options.
- FE changes were really authored by @MiyaNoctem who outlined how to do them in the comments below


MiyaNoctem:
Please capitalize the title for the PR, so it reads "PCHR-1994: Add New Pay Cycles".

mickadoo:
@MiyaNoctem I changed the title

mickadoo:
@MiyaNoctem actually Mike said to just do the backend changes. I had seen that form-pay.js file but was going to leave it to a proper JS dev to handle it. Since you gave me all the changes I needed I just added them and pushed my changes now. If you're happy with that then I guess we won't need a separate FE dev for this ticket - thanks!

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1034,6 +1034,29 @@ public function upgrade_1025() {
   }
 
   /**
+   * Add options to choose bi-monthly or bi-weekly for pay cycle
+   *
+   * @return bool
+   */
+  public function upgrade_1026() {
```
MiyaNoctem: Sorry I missed this earlier, but you should call this method  at the end of upgradeBundle() function in this file.  This is so that on new installations your upgrade gets run also and you don't have to worry about making the same change twice (installation vs upgrade scripts). Just call your upgrader function after this line:

https://github.com/civicrm/civihr/blob/d335e1d80c9c68fc5e06a88c37a890d0fcfd1fd1/hrjobcontract/CRM/Hrjobcontract/Upgrader.php#L545

On some other extensions this is done in a little smarter way, so we don't have to maintain this upgradeBundle method by hand, but for now, hrjobcontract still requires this.

PR #1658, opened 2017-03-21
PCHR-1387: Updates to edit and regression issues

dhavalsharma:
## Change List

1. Balance change was showing empty upon directly selecting single day
2. Made toil request to always have positive balance 
3. Add isValid to update call
4. Show Single day for single day leave request in edit 
5. Disable (not allow user to edit) all controls in view mode
6. Set default for hours/minutes instead of empty drop down
7. Editing requests that have half days as date type are now shown correctly instead of showing the default all_day
8. For multiple day request user will get to select to date from next day onwards and for same day request he has to go to single day tab
9. Show balance for toil request in edit mode
10. Staff responds to more_information_requested status request and it saves with waiting_approval status change. Test pending completion of PCHR-1834.
11. On edit required_documents for sickness request does not get set
12. Manager allowed to have statuses 'approved', 'more_information_requested', 'cancelled', 'rejected'

## gifs
for staff 
staff

for manager
manager


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -547,6 +557,10 @@ define([
 
             if (self.isMode.call(self, 'edit')) {
               initialLeaveRequestAttributes = _.cloneDeep(self.request.attributes());
+
+              if (self.request.from_date == self.request.to_date) {
```
deb1990: String comparison, use ===

PR #1663, opened 2017-03-21
PCHR-1834: Allow staff member to put leave request in "Waiting for Approval"

davialexandre:
Whenever the staff member updates a non approved/cancelled/rejected leave request, the status should be moved back to _"Waiting Approval"._ There were two conflicting checks about this in the system:

- `LeaveRequestStatusMatrix`: would return that this status change is allowed
- `LeaveRequestRights`: would return that this status change is **not** allowed

One possible fix would be changing `LeaveRequestRights` to follow the same rules as `LeaveRequestStatusMatrix`. In fact, this was the first thing I did, only to realize we would have the same logic duplicated between the two classes. For this reason, I removed all the checks related to statuses from `LeaveRequestRights` and the only class responsible for this now is `LeaveRequestStatusMatrix`.

The `LeaveRequestRights` is still responsible for non status related checks (for example,
to check which user can change the dates of a request).

#### Other changes included with this PR

There are couple of minor fixes included, that I didn't think it would be worth a separate PR:

- Change the order on which the extensions are installed during tests. Since l&a depends on the other, they should be installed first. This might prevent errors in the future, in case there's some code in the l&a upgraders that depend on something of other extension
- Fix contract info in sample data. The sample data was missing some required information and this would prevent the user to edit the contract in the UI.


uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -391,9 +391,9 @@ SET foreign_key_checks = 0;
...
 INSERT INTO `civicrm_hrjobcontract_revision` VALUES (1,1,1,'2017-02-23 15:04:46','2017-02-23 15:04:47','2016-01-01','2016-01-01',NULL,NULL,1,NULL,NULL,NULL,NULL,NULL,1,1,0),(2,1,1,'2017-02-23 15:04:47','2017-02-23 15:04:48','2016-01-01','2016-01-01',NULL,NULL,1,NULL,NULL,2,NULL,NULL,1,1,0),(3,1,1,'2017-02-23 15:04:49','2017-02-23 15:04:49','2016-01-01','2016-01-01',NULL,NULL,1,NULL,3,2,NULL,NULL,1,1,0),(4,1,1,'2017-02-23 15:04:49','2017-02-23 15:04:50','2016-01-01','2016-01-01',NULL,NULL,1,NULL,3,2,4,NULL,1,1,0),(5,1,1,'2017-02-23 15:04:50','2017-02-23 15:04:50','2016-01-01','2016-01-01',NULL,NULL,1,5,3,2,4,NULL,1,1,0),(6,1,1,'2017-02-23 15:04:50','2017-02-23 15:04:50','2016-01-01','2016-01-01',NULL,NULL,1,5,3,2,4,6,1,1,0),(7,1,1,'2017-02-23 15:04:50','2017-02-23 15:04:50','2016-01-01',NULL,NULL,NULL,1,5,3,2,4,7,1,0,0);
 INSERT INTO `civicrm_hrjobcontract_leave` VALUES (1,1,20,2,1),(2,2,0,2,0),(3,3,0,2,0);
-INSERT INTO `civicrm_hrjobcontract_details` VALUES (1,NULL,NULL,NULL,NULL,'2016-01-01',NULL,NULL,0,NULL,0,NULL,NULL,1);
+INSERT INTO `civicrm_hrjobcontract_details` VALUES (1,'Developer','Developer',NULL,'Employee - Permanent','2016-01-01',NULL,NULL,0,NULL,0,NULL,NULL,1);
```
mickadoo: Is this for testing or used in new installations?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/BaseHeadlessTest.php
```diff
@@ -8,10 +8,10 @@
 
   public function setUpHeadless() {
     return \Civi\Test::headless()
-      ->installMe(__DIR__)
       ->install('uk.co.compucorp.civicrm.hrcore')
       ->install('org.civicrm.hrjobcontract')
       ->install('uk.co.compucorp.civicrm.hrcomments')
+      ->installMe(__DIR__)
```
mickadoo: No problem here, just curious why you re-ordered it

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
davialexandre: This is for the 1.7 test and staging sites, where we need some sample data in it (mainly for demonstration purposes). When we deploy this branch to the test site, the extension will be reinstalled and this SQL script will be executed, creating the tables structure and the sample data.

This is just a temporary measure though, while the extension is still in development.  Once the extension is done, the sample data will be removed.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/BaseHeadlessTest.php
davialexandre: It's all a matter of dependencies. The l&a extension (installed with `installMe()`) depends on the other three, so it makes sense to have them installed first. Currently, the order of installation would not cause any problems, but it can in the future. For example, part of the l&a installation process might call an API from any of the other extensions and this call will fail because that extension is not available yet.

mickadoo: Ah gotcha, makes sense now

PR #1664, opened 2017-03-22
PCHR-2085: Contract Date Fallback for Role Import

mickadoo:
## Problem

When importing job roles the logic should be that if the role start / end are not set then use the related contract start / end. However currently for imports with no role start / end dates the dates are undefined after import.

This is caused by improper formatting of the existing contract dates, which is done by `CRM_Utils_Date::formatDate`. It takes a $dateType argument which refers to the format of the date being passed in. Currently we are using the $dateType the user selects for their import file. However, for the existing contract dates we do not want to use this. We should use the $dateType that matches the format of dates stored in the database, i.e. YYYY-mm-dd.

## Solution

Change the formatting of the contract dates to use the correct $dateType.

## Notes
- I encountered some undefined errors during the test import. I think it's important that we fix these warnings and I would recommend all developers to have the same php.ini settings for error reporting
- I'm using type-hinting, for example on [this line](https://github.com/civicrm/civihr/compare/PCHR-2085-use-contract-date-for-role-import?expand=1#diff-0b10d27247ff6ffc7bfc5598a88c589cR150). I prefer not to have any warnings in my code, but I'm open to discussion on this.
- My IDE is trimming whitespace in files ever since I enabled the editorconfig plugin. I know it's annoying to have these extra changes to look at. If people are doing things different normally just tell me and I'll try to turn it off.
- I didn't find any documentation on the $dateType and to be honest, it's a mess looking at their functions. If you know where I can find documentation on this please let me know.

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
@@ -147,7 +147,6 @@ class CRM_Hrjobroles_Import_Parser_HrJobRoles extends CRM_Hrjobroles_Import_Pars
     $session = CRM_Core_Session::singleton();
     $dateType = $session->get('dateTypes');
 
-    /** @var object $contractDetails */
     $contractDetails = NULL;
     if (!empty($params['job_contract_id']))  {
       $contractDetails = CRM_Hrjobcontract_BAO_HRJobContract::checkContract($params['job_contract_id']);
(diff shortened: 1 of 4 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
```diff
@@ -147,6 +147,7 @@ function summary(&$values) {
     $session = CRM_Core_Session::singleton();
     $dateType = $session->get('dateTypes');
 
+    /** @var object $contractDetails */
```
davialexandre: Wouldn't it make more sense to add the type hint to the checkContract (which is wrongly saying it returns an array instead of an object)?

davialexandre:
About your notes:
- No problem with the whitespaces. This is just the plugin cleaning up the code a bit. When viewing changes in github, you can ignore them by adding a ?w=1 to the url: https://github.com/civicrm/civihr/pull/1664/files?w=1
- Sadly, I doubt there's any documentation for the $dateType. The only thing I can think of is to search places in civicrm-core that use this method, check what are the values they are passing to it and maybe submit a patch to core adding so constants to avoid having this magic number in the code. For now, the change you did for the PR is fine

com.civicrm.hrjobroles/CRM/Hrjobroles/Import/Parser/HrJobRoles.php
mickadoo: Sure, that makes more sense.

PR #1666, opened 2017-03-23
PCHR-1961: Merge existing and new contract data if it already exists

bmnepali:
### Issue:
Currently, when a contract is edited and saved, it is also added to the list of contracts (in both current and past contracts). Which is causing the duplicate contract error, halting to display other contracts in the page.

### Solution:
1. If we are editing the contract, the new data,  extend/merge with the existing one. 
2. Add the contract to the list of contracts if we are adding new one only.

### Screen capture:
1. Contracts don't get disappear when any one of the contract is deleted or edited.
editing-contract-makes-other -disappear


hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -78,16 +78,29 @@ define([
        * @param {string || date} newEndDate the date specified by the user
        */
       function updateContractList(newEndDate) {
-        var isCurrentContract = !newEndDate ? true : (moment().diff(newEndDate, "day") <= 0);
+        var isCurrentContract = !newEndDate ? true : (moment().diff(newEndDate, "day") <= 0),
+          contract = $scope.$parent.contract,
+          contractContracts = $scope.$parent.contractCurrent,
```
deb1990: contractContracts?

hrjobcontract/js/src/job-contract/controllers/contract.js
```diff
@@ -78,16 +78,29 @@ define([
...
         } else {
-          $scope.$parent.contract.is_current = '0';
-          $scope.$parent.contractPast.push($scope.$parent.contract);
-          $scope.$parent.contractCurrent.splice($scope.$parent.contractCurrent.indexOf($scope.$parent.contract), 1)
+          contract.is_current = '0';
+          if (pastContractIndex + 1) {
+            angular.extend(pastContracts[pastContractIndex], contract);
+          } else {
+            pastContracts.push(contract);
+            contractContracts.splice(currentContractIndex, 1);
```
deb1990: Are you sure this is correct? Looking at the old code, it seems that it should be
`contractContracts.splice(currentContractIndex);`

hrjobcontract/js/src/job-contract/controllers/contract.js
bmnepali: It should be currentContracts. will correct.

hrjobcontract/js/src/job-contract/controllers/contract.js
bmnepali: `contractContracts.splice(currentContractIndex);` and `contractContracts.splice(currentContractIndex, 1);` means same. But `contractContracts.splice(currentContractIndex);` seems more appropriate.

PR #1667, opened 2017-03-23
PCHR-1428: Implement attachments UI

dhavalsharma:
Implements the UI for adding file attachments to request model.

## gif
attach

## Testing And API Limitation
A custom [patch](https://compucorp.atlassian.net/wiki/display/PCHR/Working+with+CiviCRM+Attachments) is applied on civicrm to make attachments work for now. 

The `LeaveRequestAPI.getAttachments` api end point does not return file uploader's contact id. Hence it is not displayed in UI.

## Implementation 
The `LeaveRequestInstance.files` array is used to manage the already uploaded attachments. The removal of files is from server side upon save. 

The `LeaveRequestInstance.uploader.queue` array is used to manage the currently uploading attachments. It's addition is either by clicking on `Browse` button or by drag-drop file in the upload area. Its removal is from local `queue` array.

When the user's saves the leave request `LeaveRequestInstance.uploadAttachments` is called. It will either resolve or reject the promise based on state of uploading of files to server.


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -163,69 +163,82 @@ <h4 class="modal-title">
...
-                    <div class="row">
-                      <div class="chr_leave-request-files_icon text-primary col-xs-6">
-                        <i class="fa fa-file-pdf-o" aria-hidden="true"></i>
-                        <span>filename.loremipsum</span>
+                  <span ng-if="$ctrl.request.uploader" ng-disabled="$ctrl.isMode('view')">
+                    <li ng-repeat="file in $ctrl.request.uploader.queue">
+                      <div class="chr_leave-request-files_meta">
+                        <span class="chr_leave-request-files_author">Me</span>
+                        <span>-</span>
+                        <span class="chr_leave-request-files_date">{{$ctrl.today | date: $ctrl.uiOptions.userDateFormat + ' HH:mm'}}</span>
```
AkA84:
```html
$ctrl.uiOptions.userDateFormat + ' HH:mm'
```
I think this should be stored as a controller's scope property

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -125,7 +125,8 @@ define([
             request_type: 'leave',
             uploader: FileUpload.uploader({
               entityTable: 'civicrm_hrleaveandabsences_leave_request',
-              crmAttachmentToken: sharedSettings.attachmentToken
+              crmAttachmentToken: sharedSettings.attachmentToken,
+              queueLimit: 5 //allow to upload upto 5 files for now
```
AkA84: Can this property be encapsulated in the `FileUpload` service? it seems something more of a generic config setting rather than something specific that `LeaveRequestInstance` should be concerned about

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -125,7 +125,8 @@ define([
             request_type: 'leave',
             uploader: FileUpload.uploader({
               entityTable: 'civicrm_hrleaveandabsences_leave_request',
-              crmAttachmentToken: sharedSettings.attachmentToken
+              crmAttachmentToken: sharedSettings.attachmentToken,
```
AkA84: The following comment is unrelated to this line, but I can't comment on lines that are not marked as diff

When the instance uploads the attachments via `this.uploader`, shouldn't it store them afterwards in the `this.files` property? By looking at
```js
function uploadAttachments() {
  if (this.uploader.queue && this.uploader.queue.length > 0) {
    return this.uploader.uploadAll({ entityID: this.id });
  } else {
    return $q.resolve([]);
  }
}
```
once the files are uploaded nothing happens, the instance is completely unaware of the files that are now associated with it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -386,7 +397,7 @@ define([
         //if set indicates self leaverequest is either being managed or edited
         if (this.directiveOptions.leaveRequest) {
           //get a clone so self it is not the same reference as passed from callee
-          attributes = _.cloneDeep(this.directiveOptions.leaveRequest.attributes());
+          attributes = this.directiveOptions.leaveRequest.attributes();
```
AkA84: Why the need for `cloneDeep` is not there anymore?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -541,6 +552,7 @@ define([
             return $q.all([
               self._loadAbsenceTypes.call(self),
               loadCommentsandContactnames.call(self),
```
AkA84: This should be
```js
loadCommentsAndContactNames
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -932,6 +944,20 @@ define([
       }
 
       /**
+       * Loads the attachmentss for current leave request
+       *
+       * @return {Promise}
+       */
+      function loadAttachments() {
+        //In CREATE mode dont fetch comments
```
AkA84: I don't like this tbh, same goes for the comments. The controller shouldn't concern itself with whether it's safe to call or not to call `instance.loadAttachments` or `instance.loadComments`. The only reason for why we are doing this check is only because on "create" mode the leave request instance wouldn't have the `id` property, and so the api call would fail.

But again, this is something that needs to be handled within the leave request instance itself, not on the controller. The controller should only call
```js
this.request.loadComments()
```
and then that either returns the array of comments or an empty array, either because it has none, or because the leave request is not on the DB yet

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Also in the call we talked about using `angular.copy` as `cloneDeep` was not making a perfect copy, dont we need that?

dhavalsharma: Because uploader gets stuck while calling uploadAll and the promise is never resolved if its cloneDeep'ed or angular.copy'ed. Also, its not affecting the report view so reverted to direct copy of attributes

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: I think its a customization for the client's of the service , by default it is set to 1

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: added `implementation` note in PR above.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -123,9 +123,12 @@ define([
...
             files: [],
             request_type: 'leave',
-            uploader: FileUpload.uploader({
+            //FileUpload.uploader has uploader property which was causing circular reference issue
+            //hence renamed this uploader to fileUploader
+            fileUploader: FileUpload.uploader({
               entityTable: 'civicrm_hrleaveandabsences_leave_request',
-              crmAttachmentToken: sharedSettings.attachmentToken
+              crmAttachmentToken: sharedSettings.attachmentToken,
+              queueLimit: 5 //allow to upload upto 5 files for now
```
AkA84: this comment is unnecessary and kind of confusing with that "for now", it seems this is temporary?

PR #1670, opened 2017-03-24
PCHR-1406: Fix FTE Calculation for Empty Standard Hours

MiyaNoctem:
## Problem
Default value for FTE in database was 1, for both numerator and denominator.  This caused an FTE of 1 to be calculated on import for contracts with an empty standard hours value, since the contract was being created with these values empty and hence defaulted to 1.

image-2016-08-09-16-37-57-904

Also, 0 hours amount with empty period was showing as "(0h per )".

1406_imported contract

## Solution
Fixed by altering contract hours table so default values are zero and implemented default handling of FTE values on import, also defaulting to 0.  Also implemented condition to show hours per period only if hours is more than 0 and period is not empty.

image


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1034,6 +1035,24 @@ public function upgrade_1025() {
   }
 
   /**
+   * Alters civicrm_hrjobcontract_hour table to make default values for fte_num
+   * and fte_denom equal to 0.
+   * 
+   * @return boolean
+   *   True on success
+   * @todo Rename upgrade!!  To upgrade_1026()  Include in upgrade bundle!
```
davialexandre: I don't understand what you meant here. Can you please clarify?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1034,6 +1035,24 @@ public function upgrade_1025() {
...
+   */
+  public function upgrade_1026() {
+    $query = '
```
davialexandre: Just to make this a bit cleaner, use double quotes around the string and then you can remove the escape chars inside the query

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1034,6 +1035,24 @@ public function upgrade_1025() {
...
 
   /**
+   * Alters civicrm_hrjobcontract_hour table to make default values for fte_num
+   * and fte_denom equal to 0.
+   * 
+   * @return boolean
+   *   True on success
+   * @todo Rename upgrade!!  To upgrade_1026()  Include in upgrade bundle!
+   */
+  public function upgrade_1026() {
```
davialexandre: Whenever you change the definition of a table with an upgrader method, you also need to update the sql script used during the extension installation. For hrjobcontract, this file is `sql/install.sql `.

The DAO has to be updated as well. Since there are no XML schemas for this extension, please update it manually.

hrjobcontract/api/v3/jobcontract_utils.php
```diff
@@ -123,8 +123,10 @@ function _civicrm_hrjobcontract_api3_get_current_revision($params) {
         'details_revision_id.period_end_date'
       ],
     ]);
-    $contractEnd = $revision['values'][0]['details_revision_id.period_end_date'];
-    $revisionEndDate = $revision['values'][0]['effective_end_date'];
+    if ($revision['count'] > 0) {
+      $contractEnd = isset($revision['values'][0]['details_revision_id.period_end_date']) ? $revision['values'][0]['details_revision_id.period_end_date'] : null;
```
davialexandre: Let's try not to create such long lines. Here's a shorter version:

```php
$firstRevision = $revision['values'][0];
$contractEnd = CRM_Utils_Array::value('details_revision_id.period_end_date', $firstRevision);
$revisionEndDate = CRM_Utils_Array::value('effective_end_date', $firstRevision);
```

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -226,10 +226,22 @@ public function testFTECalculationOnCasualTypeAndEmptyAmount() {
     $importEmptyContractResponse = $this->runImport($emptyHoursContract);
     $this->assertEquals(CRM_Import_Parser::VALID, $importEmptyContractResponse);
 
+    // Only Mandatory Fields
```
davialexandre: With this new addition, the method name doesn't reflect what it's being tested anymore. The options here are either:
- Split the test into 3 tests (one for casual type, one for empty amount and one for mandatory fields)
- Leave the existing test alone and create a separate one just for the mandatory fields
- Update the test name to reflect what is being tested now

My order of preference if from the first option to the last one. Honestly, the last option should only be used in case the performance impact of adding a separate test is too high. Otherwise, it's always best to make each test to assert a single scenario.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
MiyaNoctem: It was a note for me, I had to use a function name different to upgrade_1026, since I had already used that name for another upgrade, on a different branch.  Sorry about that!

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -224,9 +231,17 @@ public function testFTECalculationOnCasualTypeAndEmptyAmount() {
...
+    $expected = ['fte_num' => 0, 'fte_denom' => 0, 'hours_fte' => 0];
+    $this->validateHourAutoFields($contact2['id'], $expected);
+  }
 
-    // Only Mandatory Fields
+  /**
+   * Verifies if import with only mandatory fields set calculates FTE values as 
+   * 0 for FTE value, FTE numerator and FTE denominator.
+   */
+  public function testFTECalculationOnOnlyMandatoryFieldsImport() {
```
davialexandre: Thanks for splitting this into separate tests. Looks a lot better now. 

Before adding comments like this, always try to make the test name more descriptive. For example, this test can be renamed to something like `testFteFieldsAreSetToZeroWhenImportingOnlyTheMandatoryFields`

PR #1676, opened 2017-03-28
PCHR-1108:  Fix issues brought up in PCHR-970

tunbola:
This PR fixes issues brought up in the PCHR-970 user story.

These issues are mainly: 

1. Inability to save and override the 'New Proposed Period Entitlement' column on the confirm calculations page. 

Issue (1) was fixed once the code for creating public holiday leave requests when creating balance changes for public holidays after saving entitlements on the confirm calculations page was removed. (This code was never needed and has been removed).
It turned out that the public holiday leave requests were created without passing the false (don't run validation) parameter to the LeaveRequest BAO create method thereby causing validation errors which prevented the all the methods in the transaction to successfully execute..

Some other issues/cleanup was also done:

- Removing old balance changes related to a LeavePeriodEntitlement after it gets overridden/re-saved and a new LeavePeriodEntitlement is created in its place.

- Some changes to creating LeaveBalanceChanges for LeavePeriodEntitlement after saving on the confirm calculations page. Currently the proRata amount is used to create the balance change, but proRata actually includes public holiday days [View Code Here](https://github.com/civicrm/civihr/blob/255f80bafac49bf0826637b53f0224e2783e6673/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/ContractEntitlementCalculation.php#L60) (i.e when the job leave allows this) and we are also creating LeaveBalance changes for the Public Holidays separately.
This resulted in duplication and causing the numbers not to add up. So the number of public holiday is deducted from the proRata before saving the balance change.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
@@ -325,8 +325,6 @@ class CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement extends CRM_HRLeaveAndAb
     LeavePeriodEntitlement $periodEntitlement
   ) {
     $balanceChangeTypes = array_flip(LeaveBalanceChange::buildOptions('type_id'));
-    $leaveRequestStatuses = array_flip(LeaveRequest::buildOptions('status_id'));
-    $leaveRequestDateTypes  = array_flip(LeaveRequest::buildOptions('from_date_type'));
 
     $publicHolidays = $calculation->getPublicHolidaysInEntitlement();
 
@@ -337,28 +335,6 @@ class CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement extends CRM_HRLeaveAndAb
         'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT,
         'amount'      => count($publicHolidays)
       ]);
-
-      foreach ($publicHolidays as $publicHoliday) {
-        $leaveRequest = LeaveRequest::create([
-          'type_id'        => $periodEntitlement->type_id,
-          'contact_id'     => $periodEntitlement->contact_id,
-          'status_id'      => $leaveRequestStatuses['Admin Approved'],
-          'from_date'      => CRM_Utils_Date::processDate($publicHoliday->date),
-          'to_date'        => CRM_Utils_Date::processDate($publicHoliday->date),
-          'from_date_type' => $leaveRequestDateTypes['All Day'],
-          'to_date_type'   => $leaveRequestDateTypes['All Day'],
-          'request_type'   => LeaveRequest::REQUEST_TYPE_PUBLIC_HOLIDAY
-        ]);
-
-        $requestDate  = LeaveRequestDate::getDatesForLeaveRequest($leaveRequest->id)[0];
-
-        LeaveBalanceChange::create([
-          'type_id'        => $balanceChangeTypes['Public Holiday'],
-          'amount'         => -1,
-          'source_id'      => $requestDate->id,
-          'source_type'    => LeaveBalanceChange::SOURCE_LEAVE_REQUEST_DAY
-        ]);
-      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -250,13 +251,16 @@ private static function saveLeaveBalanceChange(
...
+    //since public holiday balance changes are saved differently, we need to deduct it from the pro rata
+    $adjustedProRata = $calculation->getProRata() - $calculation->getNumberOfPublicHolidaysInEntitlement();
+
 
     LeaveBalanceChange::create([
       'type_id' => $balanceChangeTypes['Leave'],
       'source_id' => $periodEntitlement->id,
       'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT,
-      'amount' => $proRata
+      'amount' => $adjustedProRata
```
davialexandre: Since this is the only place were we're using this variable, we can ditch it and just do
```php
'amount' => $calculation->getProRata() - $calculation->getNumberOfPublicHolidaysInEntitlement()
```

But please keep the comment

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -337,28 +338,6 @@ private static function savePublicHolidaysBalanceChanges(
         'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT,
         'amount'      => count($publicHolidays)
       ]);
```
davialexandre: The method description mentions that it creates one leave request for each public holiday, but this is not the case anymore. Please update it

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -2540,5 +2540,29 @@ public function testCalculateAmountForDateForAContactUsingTheDefaultWorkPattern(
     $amount = LeaveBalanceChange::calculateAmountForDate($leaveRequest, new DateTime('2016-08-03'));
     $this->assertEquals(0, $amount);
   }
+
+  public function testCanDeleteTheBalanceChangesForALeavePeriodEntitlement() {
+    $leavePeriodEntitlement = $this->createLeavePeriodEntitlement();
+
+    LeaveBalanceChangeFabricator::fabricateForLeavePeriodEntitlement($leavePeriodEntitlement);
+    LeaveBalanceChangeFabricator::fabricateForLeavePeriodEntitlement($leavePeriodEntitlement);
+    LeaveBalanceChangeFabricator::fabricateForLeavePeriodEntitlement($leavePeriodEntitlement);
```
davialexandre: Instead of creating a new fabricator method that it's unlikely to be used anywhere else other than here. I would probably be better to do this:

```php
$params = ['source_id' => $leavePeriodEntitlement->id, 'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT];
LeaveBalanceChangeFabricator::fabricate($params);
LeaveBalanceChangeFabricator::fabricate($params);
LeaveBalanceChangeFabricator::fabricate($params);
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -382,8 +382,8 @@ public function testCanSaveALeavePeriodEntitlementFromAnEntitlementCalculation()
     // 10 + 1 + 3 (Pro Rata + Brought Forward + No. Public Holidays)
     $this->assertEquals(14, $periodEntitlement->getEntitlement());
 
-    //The 3 days deducted because of the Public Holidays
-    $this->assertEquals(-3, $periodEntitlement->getLeaveRequestBalance());
+    //No leave request should be created
+    $this->assertEquals(0, $periodEntitlement->getLeaveRequestBalance());
```
davialexandre: You can delete this assertion completely. No need to assert something that the method doesn't do anymore (and it was never supposed to do)

PR #1677, opened 2017-03-29
PCHR-2081: Add non-EEA joining timeline

mickadoo:
## Problem

- The "Joining" process for non-European citizens can differ, but this is not reflected in the joining tasks.
- Default tasks do not have a default status
- There are no documents attached to the default Joining timelines

## Solution

- Add a new "Non-EEA" joining timeline
- Add the default status of "Scheduled" for all default tasks
- Add the required documents to each Joining timeline

### DIFF
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -116,9 +116,22 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
     $this->up1402_removeUnusedCaseTypes();
     //Removes CiviCase activity types which should belong to CiviTask component
     $this->removeActivityTypesList($defaultActivityTypes, 'CiviCase');
-    $this->up1402_createDefaultCaseTypes($defaultCaseTypes);
+    $this->createOrUpdateDefaultCaseTypes($defaultCaseTypes);
     $this->up1402_createDefaultActivityTypes($defaultActivityTypes);
     $this->changeActivityTypeComponent('Open Case', 'CiviCase', 'CiviTask');
+
+    return TRUE;
+  }
+
+  /**
+   * Resets all default case types discarding any customization to match new
+   * activity workflow
+   *
+   * will be 1404
+   */
+  public function upgrade_1412() {
+    $defaultTypes = CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultCaseTypes();
+    $this->createOrUpdateDefaultCaseTypes($defaultTypes);
 
     return TRUE;
   }
@@ -329,10 +342,12 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
    * Creates/Updates default case types to be shipped
    * with this extension
    *
+   * WARNING: Resets activity types for an assignment, discarding customization
+   *
    * @param array $defaultCaseTypes
    *   A list of case types with their related data
    */
-  private function up1402_createDefaultCaseTypes($defaultCaseTypes) {
+  private function createOrUpdateDefaultCaseTypes($defaultCaseTypes) {
     foreach ($defaultCaseTypes as $caseType) {
       $type = civicrm_api3('CaseType', 'get', [
         'sequential' => 1,
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -126,10 +126,8 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
   /**
    * Resets all default case types discarding any customization to match new
    * activity workflow
-   *
-   * will be 1404
    */
-  public function upgrade_1412() {
+  public function upgrade_1404() {
     $defaultTypes = CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultCaseTypes();
     $this->createOrUpdateDefaultCaseTypes($defaultTypes);
 
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -32,23 +32,23 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
   public function uninstall() {
     self::activityTypesWordReplacement(true);
     self::removeRelationshipTypes();
-    self::removeCaseTypesWithData(array_column(CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultCaseTypes(), 'name'));
-    $this->removeActivityTypesList(CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultActivityTypes(), 'CiviTask');
+    self::removeCaseTypesWithData(array_column(DefaultCaseAndActivityTypes::getDefaultCaseTypes(), 'name'));
+    $this->removeActivityTypesList(DefaultCaseAndActivityTypes::getDefaultActivityTypes(), 'CiviTask');
   }
 
   public function enable() {
     self::toggleRelationshipTypes(1);
-    self::toggleCaseTypes(array_column(CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultCaseTypes(), 'name'), 1);
-    self::toggleCaseTypes(CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultCiviCRMCaseTypes(), 0);
-    self::toggleActivityTypes(CRM_HRCase_DefaultCaseAndActivityTypes::getDefaultActivityTypes(), 1);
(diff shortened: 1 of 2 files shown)


hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -116,14 +116,26 @@ public function upgrade_1402() {
...
 
     return TRUE;
   }
 
   /**
+   * Resets all default case types discarding any customization to match new
+   * activity workflow
+   */
+  public function upgrade_1404() {
+    $this->installActivityTypes('CiviDocument', ['P45']);
```
omarabuhussein: you could have used up1402_createDefaultActivityTypes and changed it to createOrUpdateDefaultActivityType similar to what you did with createOrUpdateDefaultCaseTypes .

so the whole thing would be : 

1) inside civihr/hrcase/CRM/HRCase/DefaultCaseAndActivityTypes.php , change $defaultActivityTypes  to the following :

```php
  private static $defaultActivityTypes = [
    'Schedule Exit Interview' => 'CiviTask',
    'Get "No Dues" certification' => 'CiviTask',
    'Conduct Exit Interview' => 'CiviTask',
    'Revoke Access to Database' => 'CiviTask',
    'Block work email ID' => 'CiviTask',
    'Background Check' => 'CiviTask',
    'References Check' => 'CiviTask',
    'Schedule joining date' => 'CiviTask',
    'Issue appointment letter' => 'CiviTask',
    'Fill Employee Details Form' => 'CiviTask',
    'Submission of ID/Residence proofs and photos' => 'CiviTask',
    'Program and work induction by program supervisor' => 'CiviTask',
    'Enter employee data in CiviHR' => 'CiviTask',
    'Group Orientation to organization, values, policies' => 'CiviTask',
    'Probation appraisal (start probation workflow)' => 'CiviTask',
    'Confirm End of Probation Date' => 'CiviTask',
    'Start Probation workflow' => 'CiviTask'
    'P45' =>'CiviDocument',
  ];
```

and here change up1402_createDefaultActivityTypes  to :

```php
  private function createOrUpdateActivityTypes($defaultActivityTypes) {
    foreach ($defaultActivityTypes as $activityType => $componentName) {
      $type = civicrm_api3('OptionValue', 'get', [
        'sequential' => 1,
        'name' => $activityType,
        'option_group_id' => 'activity_type',
        'component_id' =>  $componentName,
      ]);

      if (empty($type['id'])) {
        civicrm_api3('OptionValue', 'create', [
          'sequential' => 1,
          'option_group_id' => 'activity_type',
          'label' => $activityType,
          'name' => $activityType,
          'component_id' =>  $componentName,
        ]);
      }
    }
  }
```

omarabuhussein: also I see that you added VISA and Passport activity types but you did not created them first , are they created in different extension ?

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -329,10 +341,12 @@ private function up1402_removeUnusedCaseTypes() {
    * Creates/Updates default case types to be shipped
    * with this extension
    *
+   * WARNING: Resets activity types for an assignment, discarding customization
```
omarabuhussein: not sure what do you mean by this , this method only creates a Case Type if the case type is not already exists , so nothing will get discarded here , or am I missing something  ?

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -471,4 +485,62 @@ private function removeCaseTypesWithData($caseTypes) {
     }
   }
 
+  /**
+   * @param $component
+   * @param array $types
+   */
+  private function installActivityTypes($component, array $types) {
```
omarabuhussein: so if you followed my suggestion above you can remove both **installActivityTypes** & **fetchActivityTypeParams** methods .. by the way , civicrm understands component and option group names , so you can just pass the component name and option group name without the need to fetch their IDs first , so you can simply just use : 

```php
    civicrm_api3('OptionValue', 'create', array(
 +        'sequential' => 1,
 +        'option_group_id' => 'activity_type',
 +        'component_id' => 'CiviTask',
 +        'label' => $type,
 +        'name' => $type,
 +      ));
```

you can use API explorer to create you API calls and better understanding for how things work

hrcase/CRM/HRCase/Upgrader.php
mickadoo: It's just a warning that if the client has customized the workflow then this customization will be discarded by calling this method. When I saw it first I thought "oh nice, an easy way to add new case types" but didn't realize it will completely reset all of them.

hrcase/CRM/HRCase/Upgrader.php
mickadoo: Yes, which led to a bit of confusion for me about which repo to do this in since the other types are created in the [T&A repo](https://github.com/compucorp/civihr-tasks-assignments/blob/master/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php#L501)

omarabuhussein: ah okay , I think that's fine , we probably need to move them from T&A to this one later

hrcase/CRM/HRCase/Upgrader.php
omarabuhussein: ops I just relized this ,  I think you can just change pass a flag to this method :

```php
function createOrUpdateDefaultCaseTypes($defaultCaseTypes, $doNotUpdate = false )
```

and inside the method do : 

```php
      if (!empty($type['id'])) {
        $caseType['id'] = $type['id'];
      }
      
      if ((!empty( $caseType['id']) && !$doNotUpdate) || empty( $caseType['id'])) {
        civicrm_api3('CaseType', 'create', $caseType);
      }
```

and in your new upgrader you can pass (true) to it : 

`$this->createOrUpdateDefaultCaseTypes($defaultTypes, true);`

mickadoo: I think by adding this flag doesn't match what's requested for this ticket. Mike told me that since assignments are not widely customized yet it's fine to override them. By adding a flag to not update them if they exist I think it would fail to pass testing (documents would not be added). 

Since all case types will already be added by previous upgrades then calling this method with `true` would have no effect.

hrcase/CRM/HRCase/Upgrader.php
mickadoo: Good point about the duplicate functions, and I've changed them to use the existing ones. To avoid duplication I changed the structure of the `$defaultActivityTypes`. This meant changing some methods in the Upgrader class too.

What is your opinion on using the `up1402_` style of naming methods in upgraders? I think it makes it harder to find duplicate functions if the name has a unique code in it.

hrcase/CRM/HRCase/Upgrader.php
omarabuhussein: aah right ... but since Mike said it is fine to override then then I have no problem with that. you can keep things as is.

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -367,25 +376,27 @@ private function createOrUpdateDefaultCaseTypes($defaultCaseTypes) {
...
-      $type = civicrm_api3('OptionValue', 'get', [
-        'sequential' => 1,
-        'name' => $activityType,
-        'option_group_id' => 'activity_type',
-        'component_id' => 'CiviTask',
-      ]);
-
-      if (empty($type['id'])) {
-        civicrm_api3('OptionValue', 'create', [
+  private function createOrUpdateActivityTypes($defaultActivityTypes) {
```
omarabuhussein: there is no update happening here just create .. so it should be **createActivityTypes**

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -289,11 +294,15 @@ public static function isExtensionEnabled($key)  {
    *   (e.g : CiviCase, CiviTask .. etc)
    */
   private function removeActivityTypesList($activityTypes, $componentName) {
+    $allActivityTypes = [];
+    foreach ($activityTypes as $componentActivities) {
+      $allActivityTypes = array_merge($allActivityTypes, $componentActivities);
+    }
```
omarabuhussein: please add a space after this

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -3,6 +3,8 @@
 require_once 'Upgrader/Base.php';
 require_once 'DefaultCaseAndActivityTypes.php';
 
+use CRM_HRCase_DefaultCaseAndActivityTypes as DefaultCaseAndActivityTypes;
```
omarabuhussein: since you defined this alias , can you replace any use for CRM_HRCase_DefaultCaseAndActivityTypes  in this class to DefaultCaseAndActivityTypes

PR #1679, opened 2017-03-29
PCHR-2089: Recalculate 'hours_unit' on change

mickadoo:
## Problem

In the contract edit modal, under the "Hours" tab the "Actual hours" unit is not updated when the "Location/Standard hours" is changed. 

This leads to the incorrect situation where "Location/Standard hours" uses "XX per day" but the unit says "X per week" or vice-versa

screenshot-5

The hours should still be reset if the user chooses "full time" as the new hours type, according to specs from Mike:

>Edit the contract (for example hours = 20 per week) > select hours > select full time
Expected:
hours are not editable
hours are populated from selected location "Standard Hours" field value

## Solution

Recalculate the hours unit every time the "Location/Standard hours" option is changed


hrjobcontract/js/src/job-contract/controllers/form/form-hour.js
```diff
@@ -20,10 +20,9 @@ define([
       function updateHours(locStandHrs, hrsTypeId) {
         $scope.hrsTypeDefined = !!entityHour.hours_type;
         $scope.hrsAmountDefined = !!entityHour.hours_amount;
+        entityHour.hours_unit = locStandHrs.periodicity;
 
         if ($scope.hrsTypeDefined && !$scope.hrsAmountDefined) {
-          entityHour.hours_unit = locStandHrs.periodicity;
```
AkA84: Any idea why this line was in the `if ` in the first place?

mickadoo: Originally it was just a conditional checking if the hours type was set since if it's not these fields are hidden. In this case it made no sense to recalculate hours type.

The most [recent commit](https://github.com/civicrm/civihr/commit/61b297e1bd4de676a195747dbfe02810cd894af2) on this file changed it to also check if the hours amount was defined. This is because if it's not set (on the first time you load this) it will use the default standard hours in the "hours amount".

I think it was just an oversight, the hours_unit doesn't need to be updated if the hours_type is not selected, but it should always be updated regardless of whether the hours_amount is set.

PR #1680, opened 2017-03-29
PCHR-2055: Disallow Duplicate Absence Types

mickadoo:
## Problem

Creating two absence types with the same name is permitted, but they will use the same activity_type_id. When one is deleted the activity type for it is also deleted. The whole system becomes unusable and the following error is displayed:

screenshot from 2017-03-09 11-53-28

## Solution

Add a rule to check for absence types with duplicate names, but only when creating a new type

## Notes

- I wasn't sure which was the best option for checking if it's a 'create' or 'update'. I used the `$self->getAction()` but I'm open to changing it to a `isset($fields['id'])` if that makes more sense.

### DIFF
--- a/hrabsence/CRM/HRAbsence/Form/AbsenceType.php
+++ b/hrabsence/CRM/HRAbsence/Form/AbsenceType.php
@@ -103,7 +103,7 @@ class CRM_HRAbsence_Form_AbsenceType extends CRM_Core_Form {
    $this->add('checkbox', 'allow_debits', ts('Allow Debits?'), CRM_Core_DAO::getAttribute('CRM_HRAbsence_DAO_HRAbsenceType', 'allow_debits'));
     $this->add('checkbox', 'is_active', ts('Enabled?'), CRM_Core_DAO::getAttribute('CRM_HRAbsence_DAO_HRAbsenceType', 'is_active'));
 
-    $this->addFormRule(array('CRM_HRAbsence_Form_AbsenceType', 'formRule'), $this);
+    $this->addFormRule([$this, 'formRule']);
   }
 
   /**
@@ -113,22 +113,23 @@ class CRM_HRAbsence_Form_AbsenceType extends CRM_Core_Form {
    *  An array of fields from the form
    * @param array $files
    *  An array of uploaded files
-   * @param self $self
-   *  An instance of this class
    * @return array
    *   An array of errors
    */
-  public static function formRule($fields, $files, $self) {
+  public function formRule($fields, $files) {
     $errors = [];
     if (!array_key_exists('allow_debits', $fields) && !array_key_exists('allow_credits', $fields)) {
       $errors['allow_debits'] = $errors['allow_credits'] = ts("Please choose either 'Allow Debits' and/or 'Allow Credits'");
     }
 
     $title = CRM_Utils_Array::value('title', $fields);
-    if ($self->getAction() === CRM_Core_Action::ADD) {
+    $isCreation = $this->getAction() === CRM_Core_Action::ADD;
+
+    if ($isCreation) {
       $existing = civicrm_api3('HRAbsenceType', 'get', ['name' => $title]);
       if (isset($existing['count']) && $existing['count'] > 0) {
-        $errors['title'] = ts("An absence type with title '%1' already exists", [1 => $title]);
+        $errorMessage = "An absence type with title '%1' already exists";
+        $errors['title'] = ts($errorMessage, [1 => $title]);
       }
     }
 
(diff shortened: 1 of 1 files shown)


hrabsence/CRM/HRAbsence/Form/AbsenceType.php
```diff
@@ -106,11 +106,32 @@ public function buildQuickForm() {
...
+   * @param array $fields
+   *  An array of fields from the form
+   * @param array $files
+   *  An array of uploaded files
+   * @param self $self
+   *  An instance of this class
+   * @return array
+   *   An array of errors
+   */
+  public static function formRule($fields, $files, $self) {
```
davialexandre: If you need $this inside formRule, do this
- Change the method to be an instance method instead of static
- On line 106, change `array('CRM_HRAbsence_Form_AbsenceType', 'formRule')` to `array($this, 'formRule')`

and then you can use `$this` and get rid of `$self`

hrabsence/CRM/HRAbsence/Form/AbsenceType.php
```diff
@@ -106,11 +106,32 @@ public function buildQuickForm() {
...
+    $errors = [];
     if (!array_key_exists('allow_debits', $fields) && !array_key_exists('allow_credits', $fields)) {
       $errors['allow_debits'] = $errors['allow_credits'] = ts("Please choose either 'Allow Debits' and/or 'Allow Credits'");
     }
+
+    $title = CRM_Utils_Array::value('title', $fields);
+    if ($self->getAction() === CRM_Core_Action::ADD) {
+      $existing = civicrm_api3('HRAbsenceType', 'get', ['name' => $title]);
```
davialexandre: do we need to care about update? That is, I create an Absence Type with a different name but later update it to the same name of an existing one. Will this cause problems?

hrabsence/CRM/HRAbsence/Form/AbsenceType.php
mickadoo: Nice, I should have checked the method signature

hrabsence/CRM/HRAbsence/Form/AbsenceType.php
mickadoo: It won't cause the error I mentioned in the description, but it's a good point. We should probably disallow editing absence types names to match existing names

mickadoo: I've tried implementing this and had a working solution. However I noticed if I click on the type name and edit directly on the /civicrm/absence/type page and the request is sent by AJAX the validation is not called. This is really not what I expected here and I think it could cause trouble later on. 

Did I put the validation in the wrong place or do we have somewhere we usually put it so it will always be called?

davialexandre: The problem is that there's no entity validation in Civi. So, if the update comes from a form, we can use the form rules to check things, but when it comes from the API (which is the case when you edit the name directly on the list) that validation will never be called.
There's no out-of-the box solution for this in Civi, unfortunately. What we've been doing in L&A is [this](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L207-L226). We call this validateParams [inside the create](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L67) method and then we can be sure the validation will be called either if you use the API or if you call the BAO method directly (which is usually what happens on a form).

This approach has its own problems though:
- Since it throws exceptions, it will stop on the first error instead of doing all the validations and showing all the error messages just once
- The error messages cannot be linked to a form field
- In case of an error, you'll not be redirected to the form to correct whatever is wrong. Instead, you 'll be back to the list, lose what you've entered into the form and just see an error message
- Depending on how the validation methods are implemented, you might lose the ability to update a single field of the entity (what happens when you edit the name from the list)

Anyway, since you said that it won't cause the problem mentioned on the ticket and that we're going to replace this extension soon, I would say it's ok to leave this as it is.

mickadoo: I can leave this for now, but it would definitely be nice to have some way to combine the logic. I think the forms offer some handy functionality - like returning multiple errors linked to each field. 

As a pure API user (like someone integrating with the site) it would be good to have an array of descriptive errors in the result, similar to what the form validation is producing.

PR #1681, opened 2017-03-29
PCHR-1968: Fix ACL issues with Contact.getLeaveManagees API Endpoint

tunbola:
This PR fixes ACL issues with Contact.getLeaveManagees API endpoint.

Currently the API returns an empty result set for users without 'edit all contacts' or 'view all contacts' permissions.

To fix this, the Contact.getLeaveManagees is modified to ignore the default civicrm ACL checks by setting check_permissions parameter to false.
The 'ACL' needed for this endpoint was already implemented in the ContactSelect Query class:

- A Leave Approver only gets to see his/her managees and can't view the managees of another Manager.
- An Admin can view the managees of any Manager.
- A staff member can't view managees of any manager.

Some of the tests were modified to set check_permissions to true just to be sure that civicrm 'ACL' checks will be ignored.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/Contact/Getleavemanagees.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/Contact/Getleavemanagees.php
@@ -29,6 +29,8 @@ function _civicrm_api3_contact_getleavemanagees_spec(&$spec) {
  * @throws API_Exception
  */
 function civicrm_api3_contact_getleavemanagees($params) {
+  // We need to set check_permissions to false so as to disable default Civi ACL checks for
+  // this endpoint. ACL checks needed are already in the ContactSelect Query class.
   $params['check_permissions'] = false;
   $leaveManagerService = new CRM_HRLeaveAndAbsences_Service_LeaveManager();
   $query = new CRM_HRLeaveAndAbsences_API_Query_ContactSelect($params, $leaveManagerService);
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/Contact/Getleavemanagees.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/Contact/Getleavemanagees.php
@@ -29,6 +29,7 @@ function _civicrm_api3_contact_getleavemanagees_spec(&$spec) {
  * @throws API_Exception
  */
 function civicrm_api3_contact_getleavemanagees($params) {
+  $params['check_permissions'] = false;
   $leaveManagerService = new CRM_HRLeaveAndAbsences_Service_LeaveManager();
   $query = new CRM_HRLeaveAndAbsences_API_Query_ContactSelect($params, $leaveManagerService);
   return civicrm_api3_create_success($query->run(), $params);
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
@@ -33,7 +33,7 @@ class api_v3_ContactTest extends BaseHeadlessTest {
     ]);
 
     $this->registerCurrentLoggedInContactInSession($this->manager['id']);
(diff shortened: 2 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/Contact/Getleavemanagees.php
```diff
@@ -29,6 +29,7 @@ function _civicrm_api3_contact_getleavemanagees_spec(&$spec) {
  * @throws API_Exception
  */
 function civicrm_api3_contact_getleavemanagees($params) {
+  $params['check_permissions'] = false;
```
davialexandre: Please add a comment here explaining why we need this

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -45,7 +45,7 @@ public function testGetLeaveManageesDoesNotReturnDeceasedORDeletedContacts() {
     $this->setContactAsLeaveApproverOf($this->manager, $contact3, null, null, true, 'has leaves approved by');
     $this->setContactAsLeaveApproverOf($this->manager, $contact4, null, null, true, 'has things managed by');
 
-    $result = civicrm_api3('Contact', 'getleavemanagees', ['managed_by' => $this->manager['id']]);
+    $result = $this->callGetLeaveManagees(['managed_by' => $this->manager['id']]);
```
davialexandre: Why not use check_permissions=true for all the tests? In fact, I was just thinking we can have a BaseAPITestCase class, which would serve as base class to all the tests under tests/phpunit/api/v3. This class would:
- automatically give the 'Access Ajax API' permission to the current logged in user on its setup method
- have a helper method to call civicrm_api3 passing check_permission to true

The idea is to avoid having these ACL/permissions issues again and again. 

We don't need to refactor all the other tests now, but having it for this class would be a good start.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/SessionHelpersTrait.php
```diff
@@ -11,4 +11,8 @@ private function unregisterCurrentLoggedInContactFromSession() {
     $session = CRM_Core_Session::singleton();
     $session->set('userID', null);
   }
+
+  private function addPermissions(Array $permissions = []) {
```
davialexandre: s/Array/array/

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/ApiHelpersTrait.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_ApiHelpersTrait {
+
+  private function callAPI($entity, $action, $params, $checkPermissions = false) {
```
davialexandre: Since the idea is to enforce check_permission=true I think it would probably be better to swap things here a bit:
- The default value for `$checkPermissions` would be true
- `callAPIWithCheckPermissionsOn` would be renamed to `callAPIWithoutCheckPermissions` and it would pass false to `callAPI`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/SessionHelpersTrait.php
```diff
@@ -11,4 +11,8 @@ private function unregisterCurrentLoggedInContactFromSession() {
...
+    CRM_Core_Config::singleton()->userPermissionClass->permissions = $permissions;
```
davialexandre: This is not how I would expect this method to work. Given the `add` prefix, I would assume that, in case I already have some permissions set, it would just add the given ones instead of replacing it. Something more like this:

```php 
private function addPermission($permission) {
   if (!is_array(CRM_Core_Config::singleton()->userPermissionClass->permissions)) {
     CRM_Core_Config::singleton()->userPermissionClass->permissions = [];
   }
   CRM_Core_Config::singleton()->userPermissionClass->permissions[] = $permission;
}
```

PR #1696, opened 2017-04-07
PCHR-2003: Absence tab Module setup

deb1990:
As part of this PR and (https://github.com/compucorp/civihr-employee-portal/pull/293, https://github.com/compucorp/org.civicrm.shoreditch/pull/13 and [https://github.com/civicrm/civihr/pull/1746](https://github.com/civicrm/civihr/pull/1746)

1. Use non minified files when debugging is on (First Commit)
2. `absence-tab` module is setup
3. `contact id` is set in the settings file.
4. Testing framework is setup

**Improvements**
1. Ability to specify custom header class is added to `uibTabset` by decorating it.
To add custom header class
```html
<uib-tabset 
  custom-header-class="nav-pills-horizontal nav-pills-horizontal-default">
```
2. When debugg mode is turned on, error is thrown saying 'Uncaught Error: Mismatched anonymous define() module: function...'. This is because we had
```
sites/all/modules/civicrm/tools/extensions/civihr/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/my-leave.js
sites/all/modules/civicrm/tools/extensions/civihr/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave.js
```
 which had same file name. 
So changed the file name to `my-leave-container` and subsequent HTML, Test files too.
Same thing is done for `ManagerLeave` and `AbsenceTab`.

**Note**
This PR does not form a complete page, so `backstopjs` tests are not added. When individual pages(report, calendar etc) will be developed, `backstopjs` tests would be added.

**Screengrab**


org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
```diff
@@ -0,0 +1,18 @@
+// Decorates uibTabset to add 'customHeaderClass' and
+// to change the temnplateUrl
```
AkA84: typo

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
```diff
@@ -0,0 +1,18 @@
...
+define([
+  'common/lodash'
+], function(_) {
```
AkA84: whitespace

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
```diff
@@ -0,0 +1,18 @@
...
+  'use strict';
+
+  return ['$delegate', function($delegate) {
```
AkA84: whitespace

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
```diff
@@ -0,0 +1,18 @@
...
+    var directive = $delegate[0];
+
+    directive.bindToController = _.extend(directive.bindToController, {
```
AkA84: isn't this equivalent to
```js
directive.bindToController.customHeaderClass = '@';
```

org.civicrm.reqangular/src/common/modules/directives.js
```diff
@@ -2,18 +2,21 @@ define([
...
   'common/decorators/angular-date/datepicker-addon',
+  'common/decorators/ui-bootstrap/uib-tabset',
+  'common/angularBootstrap',
   'common/ui-select',
   'common/modules/templates',
   'common/modules/controllers',
   'common/modules/services',
   'common/modules/apis'
-], function (angular, uiSelectChoicesCustomScrollBars, datepickerAddon) {
+], function (angular, uiSelectChoicesCustomScrollBars, datepickerAddon, uibTabset) {
```
AkA84: There are several points to make here

1. I think it would be better to split this module in two, `common.directives` and `common.decorators`
2. Can you find out what `common/modules/services` and `common/modules/apis` are required for in this case? Why does this module needs to have access to them?
3. Same goes with the angular dependencies that are being injected, is `common.apis` and `ngSanitize` really needed?

org.civicrm.reqangular/src/common/templates/tab-outer.html
```diff
@@ -0,0 +1,14 @@
+<div>
+  <!--Overrides the 'uib/template/tabs/tabset.html'
+to add '{{tabset.customHeaderClass}}' which is used for custom header class-->
```
AkA84: keep the indentation

AkA84:
```html
<!-- ... header class -->
```

org.civicrm.reqangular/src/common/templates/tab-outer.html
```diff
@@ -0,0 +1,14 @@
+<div>
+  <!--Overrides the 'uib/template/tabs/tabset.html'
```
AkA84:
```html
<!-- Overrides
```

org.civicrm.reqangular/src/common/templates/tab-outer.html
```diff
@@ -0,0 +1,14 @@
...
+to add '{{tabset.customHeaderClass}}' which is used for custom header class-->
+  <ul class="nav nav-{{tabset.type || 'tabs'}} {{tabset.customHeaderClass}}"
+      ng-class="{'nav-stacked': vertical, 'nav-justified': justified}"
```
AkA84: since it's our template, use 2 spaces indentation

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceTab.php
```diff
@@ -7,7 +7,20 @@ class CRM_HRLeaveAndAbsences_Page_AbsenceTab extends  CRM_Core_Page {
   public function run() {
     CRM_Utils_System::setTitle(ts('Absence'));
 
-    CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
+    CRM_Core_Resources::singleton()
+      ->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css')
+      ->addScriptFile('org.civicrm.reqangular', 'dist/reqangular.min.js', 1010)
```
AkA84: this is always added in every civihr page, why do you need it?

org.civicrm.reqangular/src/common/templates/tab-outer.html
```diff
@@ -0,0 +1,14 @@
...
+      ng-transclude></ul>
+  <div class="tab-content">
+    <div class="tab-pane"
+         ng-repeat="tab in tabset.tabs"
+         ng-class="{active: tabset.active === tab.index}"
+         uib-tab-content-transclude="tab">
+    </div>
+  </div>
+</div>
```
AkA84: Actually I think there is a better way to do this, without having to re-write the entire template just to add single string.

If you look [here](https://github.com/civicrm/civihr/blob/staging/org.civicrm.reqangular/src/common/modules/angular-date.js) you can see that you can simply fetch the template, modify it, and then put it back in the template cache.

I think this is more maintainable because we keep using the default template instead of providing our own

AkA84: (also notice that here we're decorating a module inside a module definition, while all the other decorators are being applied in `common.directives`. This is another sign that the whole decorator approach needs to be cleaned up)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
```diff
@@ -0,0 +1,27 @@
...
+  require.config({
+    urlArgs: 'bust=' + (new Date()).getTime(),
+    paths: {
+      'leave-absences/shared': srcPath + '/shared',
+      'leave-absences/absence-tab': srcPath + '/absence-tab'
+    }
+  });
+
+  require([
+      'leave-absences/shared/config',
```
AkA84: wrong indentation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
```diff
@@ -0,0 +1,27 @@
+(function (CRM, require) {
+  CRM.vars.leaveAndAbsences = {
+    baseURL: CRM.absenceTabApp.path
+  };
+
+  var srcPath = CRM.vars.leaveAndAbsences.baseURL + '/js/angular/src/leave-absences';
```
AkA84: Why all these different steps? Can't you simply define `CRM.vars.leaveAndAbsences.baseURL` in the backend?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
```diff
@@ -0,0 +1,27 @@
...
+      'leave-absences/absence-tab': srcPath + '/absence-tab'
+    }
+  });
+
+  require([
+      'leave-absences/shared/config',
+      'leave-absences/absence-tab/app',
+    ],
+    function () {
+      document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('absenceTabReady') : (function(){
```
AkA84: `function () {}`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/app.js
```diff
@@ -0,0 +1,28 @@
+define([
+  'common/angular',
+  'common/angularBootstrap',
+  'common/services/angular-date/date-format',
```
AkA84: what do you need this for at this point?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/modules/settings.js
```diff
@@ -0,0 +1,11 @@
+(function (CRM) {
+  define([
+    'common/angular'
+  ], function (angular) {
+    return angular.module('absence-tab.settings', []).constant('settings', {
+      contactId: decodeURIComponent((new RegExp('[?|&]cid=([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null,
```
AkA84: is this from some other app?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/app.js
```diff
@@ -0,0 +1,28 @@
...
+  'leave-absences/shared/ui-router',
+  'leave-absences/absence-tab/modules/settings',
```
AkA84: you are missing a `modules/config`, that in turn requires `modules/settings`. please see the other angular apps in L&A

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
```diff
@@ -8,7 +8,7 @@ define([
   'common/services/angular-date/date-format',
   'leave-absences/shared/ui-router',
   'leave-absences/my-leave/modules/config',
-  'leave-absences/my-leave/components/my-leave',
+  'leave-absences/my-leave/components/my-leave-container',
```
AkA84: why this change? (same for `manager-leave-container`)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/modules/shared-settings.js
```diff
@@ -3,12 +3,13 @@
     'common/angular'
   ], function (angular) {
     return angular.module('leave-absences.settings', []).constant('shared-settings', {
+      attachmentToken: CRM.vars.leaveAndAbsences.attachmentToken,
       debug: CRM.debug,
-      pathTpl: CRM.vars.leaveAndAbsences.baseURL + '/views/shared/',
       managerPathTpl: CRM.vars.leaveAndAbsences.baseURL + '/views/manager-leave/',
+      pathTpl: CRM.vars.leaveAndAbsences.baseURL + '/views/shared/',
+      reqAngular: CRM.vars.reqAngular,
```
AkA84: what is this for?

uk.co.compucorp.civicrm.hrleaveandabsences/views/absence-tab/components/absence-tab.html
```diff
@@ -0,0 +1,36 @@
...
+    <ul class="dropdown-menu" role="menu">
+      <li>
+        <a href="#">Record staff sickness</a>
+      </li>
+      <li>
+        <a href="#">Record staff overtime</a>
+      </li>
+    </ul>
+  </div>
+  <uib-tabset active="active"
```
AkA84:
```html
<uib-tabset active="active"
  type="pills"
  custom-header-class="nav-pills-horizontal nav-pills-horizontal-default">
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/absence-tab/components/absence-tab_test.js
```diff
@@ -0,0 +1,39 @@
+(function (CRM) {
+  define([
```
AkA84: at this point you might as well include the tests for all the other components

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
deb1990: Sorry I dont get it

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-tabset.js
deb1990: Sorry I dont get it

org.civicrm.reqangular/src/common/templates/tab-outer.html
deb1990: Here 2 space indentation only is used. What you are pointing out is the line which is maintained by IDE's for attributes(same with other HTML's)
tab-outer_html_-_hr17_-____buildkit_build_hr17_

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
deb1990: Not sure what you mean by backend. For My/manager Leave there is a html page which is used by drupal, and that html defines these variables. In absence tab we dont have a html like that that is why defined the variable in the JS.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/app.js
deb1990: For other apps config contains the routing, which we dont need for absence tab. Thats why removed it. If it is required in future we can create.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/modules/settings.js
deb1990: Yes, found that this how the `contact-id`is getting fetched. Found in job roles and contact summary.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
deb1990: Got this error 'reqangular.min.js?ooqxmb:91 Uncaught Error: Mismatched anonymous define() module: function...' when debugging is turned on. 
Forgot to document, doing it.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/modules/shared-settings.js
deb1990: Removed it, not needed anymore

org.civicrm.reqangular/src/common/templates/tab-outer.html
deb1990: > Actually I think there is a better way to do this, without having to re-write the entire template just to add single string.

If you look here you can see that you can simply fetch the template, modify it, and then put it back in the template cache.

I think this is more maintainable because we keep using the default template instead of providing our own

I think having separate HTML is much more clean than fetching an HTML and modifying it with regex which makes it more complicated. Lets keep the HTML separated from JS

AkA84: The advantage of modifying the existing template over completely rewriting it, is that if we want to update the library, we have to keep syncing the original template with our overriding version. Using the original template would make things more maintainable

org.civicrm.reqangular/src/common/modules/directives.js
deb1990: > I think it would be better to split this module in two, common.directives and common.decorators

Not sure if this is a good idea. Because, `common.directives` is importing all the directives , so If we want to override some of them it is better to do there itself? The idea is `common.directives` imports all common directives and overrides them before exporting the same.

deb1990: > Can you find out what common/modules/services and common/modules/apis are required for in this case? Why does this module needs to have access to them?

1. `common/modules/services`is not required - removed
2. `common/modules/apis`is required by https://github.com/civicrm/civihr/blob/e95e63f7836e7e20b163e87754dd0652c143f1af/org.civicrm.reqangular/src/common/directives/contact-actions/contact-actions.js which internally calls a Controller and that needs `api.contactActions`

deb1990: > Same goes with the angular dependencies that are being injected, is common.apis and ngSanitize really needed?

1. `ngSanitize` not required - removed
2. `common.apis` is required by https://github.com/civicrm/civihr/blob/e95e63f7836e7e20b163e87754dd0652c143f1af/org.civicrm.reqangular/src/common/directives/contact-actions/contact-actions.js which internally calls a Controller and that needs `api.contactActions`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/app.js
AkA84: There are also things like
```js
$logProvider.debugEnabled(settings.debug);
$resourceProvider.defaults.stripTrailingSlashes = false;
$httpProvider.defaults.headers.common["X-Requested-With"] = 'XMLHttpRequest';
```
in general, even if it's just for the log, it's better to keep things consistent across all L&A apps

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
```diff
@@ -14,14 +14,13 @@
...
-          var e = document.createEvent('Event');
-          e.initEvent('absenceTabReady', true, true);
-          return e;
-        })());
-    });
+    'leave-absences/shared/config',
+    'leave-absences/absence-tab/app'
+  ], function () {
+    document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('absenceTabReady') : (function () {
+        var e = document.createEvent('Event');
```
AkA84: wrong indentation

org.civicrm.reqangular/src/common/modules/directives.js
AkA84: 1. agreed
2. if `common/modules/api`/`common.apis` is needed by a controller, then we can move it to the `common.controllers` module

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
AkA84: > manager-leave and my-leave components renamed to manager-leave-container and my-leave-container to fix the JS error when Debugging is turned on.

but why this error happens in the first place? why does it matter what a component is called? Also, aren't you changing the name of the file without changing the name of the component?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
AkA84: > Not sure what you mean by backend

You're doing this in your PR
```php
return array(
  'absenceTabApp' => array(
    'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer'),
    'path' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.hrleaveandabsences'),
  )
);
```

so why do you define `absenceTabApp` instead of directly `leaveAndAbsences`, since it's the property that you're using anyway?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
deb1990: The error happens when we have same file name for multiple components.
Added `-container` to all files(Html, Component, test etc.)

org.civicrm.reqangular/src/common/modules/uib-tabset.js
```diff
@@ -0,0 +1,18 @@
+define([
+  'common/angular',
+  'common/angularBootstrap',
+  'common/modules/templates'
+], function (angular) {
+  'use strict';
+
+  return angular.module("common.uibTabset", [])
```
AkA84: this doesn't need to be a module, you should simply doing this in the `.run()` of the already existing `common.directive` module

deb1990: We should not add this to the `common.directives` because at this moment we need to override a single directive so it looks fine, but when there will be multiple overrides the `common.directives` would get polluted easily.

AkA84: you are adding an entire new angular module specifically to simply alter the html of a single directive, it is not necessary (let's not overcomplicate things, especially if we don't even have yet a consistent approach with decorators, see `common.angularDate`) nor scalable (because you called the module with a very specific name)

Please put this in `common.directives`, if will need improvements then we'll take it from there

PR #1703, opened 2017-04-10
PCHR-1197: Fix issues with Public Holiday Leave Request Balance changes when a new Contract is created

tunbola:
When a new contract is created for a contact and there is an Absence Type with 'Must take public Holiday As Leave' set to true, The Public Holiday Leave requests for Public holidays in the future are created for that contact for the absence Type. 

## **Problem**
The problem is that the Public Holiday Leave Requests are all being created with zero balance change amounts rather than the amount being derived from the work pattern that is being used by the contact. This issue is observed for contacts using the default Work Pattern.

To calculate the amount to be deducted for the public holiday leave requests, The work pattern used by the contact and the effective start date of the Work pattern or start date of the job contract revision of the contact in the case of the contact using the default work pattern is needed.

Since the effective/start date of the job contract revision is saved right after the hook to create public holiday leave requests is triggered(when a job details is saved), the effective/start date of the job contract revision is not found when the public holiday leave requests are being created which makes the start date null. Due to the logic  [Here](https://github.com/civicrm/civihr/blob/44b189c99335b41553a5c16f6776e3417c6079f6/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php#L458-L458), The amount to be deducted returns zero.

## **Solution**
This PR fixes the issue by attempting to set the effective/start date date for the corresponding job revision contract for the contact Job contract before the call to calculate the Public holiday leave amount is triggered.

The logic to set the effective/start date for the job contract revision is moved to a point just before the job details is saved which was in the parent create method for the Job Details BAO.
 Since the start date for the job contract revision is now available at the point the public Holiday leave request creation is triggered, the correct amount as defined by the Work pattern is now deducted as balance change.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/Base.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/Base.php
@@ -92,7 +92,7 @@ class CRM_Hrjobcontract_DAO_Base extends CRM_Core_DAO
 
   /**
    * Contract revision id
-   * 
+   *
    * @var int
    */
   public $jobcontract_revision_id;
@@ -122,7 +122,7 @@ class CRM_Hrjobcontract_DAO_Base extends CRM_Core_DAO
     }
     return self::$_links;
   }
-  
+
   /**
    * returns all the column names of this table
    *
@@ -133,7 +133,7 @@ class CRM_Hrjobcontract_DAO_Base extends CRM_Core_DAO
   {
     return self::$_fields;
   }
-  
+
   static function setFields(array $fields)
   {
     self::$_fields = array_merge($fields, array(
@@ -148,7 +148,7 @@ class CRM_Hrjobcontract_DAO_Base extends CRM_Core_DAO
 
     return self::$_fields;
   }
-  
+
   /**
    * Returns an array containing, for each field, the arary key used for that
    * field in self::$_fields.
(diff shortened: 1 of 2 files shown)


hrjobcontract/CRM/Hrjobcontract/DAO/Base.php
```diff
@@ -299,12 +299,30 @@ public static function create($params)
                 $tableName . '_revision_id' => $params['jobcontract_revision_id'],
             ));
         }
-        
+
+        //setting the 'effective_date' if it's not set: Only executed for the
+        //HRJobDetails BAO.
+        if ($baoName == 'CRM_Hrjobcontract_BAO_HRJobDetails') {
```
davialexandre: you can use the `$entityName` variable here and make this shorter.

The "Only executed for the HRJobDetails BAO" part is not really necessary, since that is clear from the `if`

hrjobcontract/CRM/Hrjobcontract/DAO/Base.php
```diff
@@ -299,12 +299,30 @@ public static function create($params)
...
+            $revisionToUpdate = civicrm_api3('HRJobContractRevision', 'get', [
+                'sequential' => 1,
+                'jobcontract_id' => $jobContractId,
+                'id' => $params['jobcontract_revision_id'],
+            ]);
+
+            $revisionToUpdate = $revisionToUpdate['values'][0];
+            if (empty($revisionToUpdate['effective_date']) &&  !empty($params['period_start_date'])) {
```
davialexandre: remove one of the 2 spaces after the `&&` operator

PR #1707, opened 2017-04-11
PCHR-2124: Remove styleguide backstop scenario + misc

AkA84:
This PR contains the following:

1) Removed the styleguide backstopjs scenario, as now the styleguide extension itself will have that scenario (see [here](https://github.com/compucorp/org.civicrm.styleguide/pull/1))
2) Renamed `backstop:action` tasks in `backstopjs:action`
3) Added `backstopjs:approve` task
4) Now on page ready backstop will close the notification (if any) before taking a screenshot
5) Using the cid = 2 in all scenarios
6) Removed some of the scenarios for the Documents page (T&A), because of [this issue](https://github.com/ariya/phantomjs/issues/11151). PhantomJS has a bug with how it parses dates, and that leads the Documents page to not show any documents, which then leads to errors when CasperJS tries to click on documents that are not there. I don't know of any solutions yet, so i could only removed those scenarios

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/page.js
+++ b/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/page.js
@@ -20,6 +20,24 @@ function closeAnyModal() {
   return this;
 }
 
+/**
+ * [closeNotifications description]
+ * @return {[type]} [description]
+ */
+function closeNotifications() {
+  var casper = this.casper;
+  var notificationSelector = 'a.ui-notify-cross.ui-notify-close';
+
+  casper.then(function () {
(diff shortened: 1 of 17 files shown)


uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/page.js
```diff
@@ -20,6 +20,24 @@ function closeAnyModal() {
   return this;
 }
 
+/**
+ * [closeNotifications description]
+ * @return {[type]} [description]
+ */
```
bmnepali: If the comment block is necessary, it is  does not have a correct description.

PR #1710, opened 2017-04-12
PCHR-2113: Fix warning on the Activity.getAbsences API

davialexandre:
After upgrading to CiviCRM 4.7.18, this error started being displayed on the contact summary page:
captura de tela de 2017-04-12 17-07-51

The was caused by a call to the `getAbsences` custom action of the Activity API. It uses the `_civicrm_api3_activity_get_formatResult` function, which had its signature changed by this commit:
https://github.com/civicrm/civicrm-core/commit/bc4b6f0f5694d1a5b8e2151e6dcf9a394c300df4#diff-7673c7d055caeab2a3ec5d05a1553293R395

Due to this change, any call to this action was resulting in a warning, because `getAbsences` was not passing the new third param added to the function.

This fixes the warning by fetching the options from the `$params` array and passing it to the function.

### DIFF
--- a/hrabsence/api/v3/Activity/Getabsences.php
+++ b/hrabsence/api/v3/Activity/Getabsences.php
@@ -127,6 +127,7 @@ function civicrm_api3_activity_getabsences($params) {
   $entity = _civicrm_api3_get_BAO(__FUNCTION__);
   $bao = CRM_Core_DAO::executeQuery($select->toSQL(), array(), TRUE, 'CRM_Activity_BAO_Activity');
   $activities = _civicrm_api3_dao_to_array($bao, $params, FALSE, $entity, FALSE);
-  $activities = _civicrm_api3_activity_get_formatResult($params, $activities);
+  $options = _civicrm_api3_get_options_from_params($params, FALSE, 'Activity', 'get');
+  $activities = _civicrm_api3_activity_get_formatResult($params, $activities, $options);
   return civicrm_api3_create_success($activities, $params, $entity, 'getAbsences');
 }


mickadoo:
Do you think it makes sense to let Coleman know about the breaking change, I'm not sure if we're using semantic versioning fully but seems like a patch release shouldn't change method signatures, or if it does it should provide a default?

hrabsence/api/v3/Activity/Getabsences.php
```diff
@@ -127,6 +127,7 @@ function civicrm_api3_activity_getabsences($params) {
   $entity = _civicrm_api3_get_BAO(__FUNCTION__);
   $bao = CRM_Core_DAO::executeQuery($select->toSQL(), array(), TRUE, 'CRM_Activity_BAO_Activity');
   $activities = _civicrm_api3_dao_to_array($bao, $params, FALSE, $entity, FALSE);
-  $activities = _civicrm_api3_activity_get_formatResult($params, $activities);
+  $options = _civicrm_api3_get_options_from_params($params, FALSE, $entity, 'get');
```
mickadoo: In this case won't `$entity` be the name of the BAO class? In other examples where this is called I see it using the short entity name, e.g.

```
_civicrm_api3_get_options_from_params($params, TRUE, 'contribution', 'get');
_civicrm_api3_get_options_from_params($params, TRUE, 'Membership', 'get');
_civicrm_api3_get_options_from_params($params, TRUE, 'ReportTemplate', 'get');
```

davialexandre:
@mickadoo unfortunately, Civi doesn't use semantic versioning. In every release they just ship anything that was merged into master and barely notify people about the changes. It's not the first time something like this happens and I doubt it will be the last :(

Also note that the function starts with an underscore, which means it's private and we're not supposed to use it in our custom code. Ideally, the fix would be to remove all the usages of private functions like that, but since the hrabsence extension is about to be replace by L&A, I went to the fastest/easiest path.

PR #1720, opened 2017-04-14
AI-109: Fix styleguide error

meladawy:
Loop over current defined menus and initializ 'child' index for the menu items that doesn't has the index defined. 

Before:
ai-109

After:
screen shot 2017-04-14 at 4 14 48 pm


coldrunKacper:
Approved.

PR #1725, opened 2017-04-18
PCHR-1439: Migrate Job Roles to shoreditch theme

AkA84:
## Overview
In this PR we get rid of the local theme of the Job Roles extension and implement the common Shoreditch theme.

Also minor changes were made to the following:
* loading spinners logic
* `org.civicrm.bootstrapcivihr` extension
* BackstopJS

### Overall style
before

### Loading spinner
The way the spinners were used didn't make much sense: there was one spinner that appeared below the "Current Roles and Projects" header (which was visible from the start), and after the data was loaded the spinner was still being visible for a second or so. Also, we have two different data sets, current and past job roles, but only one spinner

loading-spinner-before

### Overall style
after

### Loading spinner
loading-spinner-after

### Unused files
The whole local bootstrap theme (`scss/civihr/`, `scss/hrjobroles/`) had been deleted. Also the `config.rb` file that was used by compass was removed as well, since from now on `gulp-sass` will be used to process the scss partials

Finally, we were explicitly including the `angular-xeditable` style on the page
```php
// Page/JobRoles.php
CRM_Core_Resources::singleton()->addStyleFile('com.civicrm.hrjobroles', 'angular-xeditable/css/xeditable.css');
```
this is not required anymore, given that the xeditable style is automatically included in Shoreditch
 

### Gulpfile
`gulpfile.js` had been updated to use the `civicrm-scssroot` module (to fetch scss partials from other extensions, like `org.civicrm.shoreditch`) and `gulp-sass`

### Markup
The entire markup of the extension had been cleaned up, refactored, and made it adhere to the Shoreditch theme.

### Custom css
A small amount of custom css was still necessary, but just for layout purposes:

#### Margin between lists
A bottom margin was added between lists, just to give some breathing spaces

#### Positioning of actions
The job roles actions (Edit, Save, Cancel) had been positioned absolutely on the top right corner of the tabs (see screenshots above)

#### Display of `.editable-textarea`
The `.editable-textarea` element had been forced to be displayed as `block` as opposed to the default `inline-block` (this is a default style coming from the original `angular-xeditable` style). This was necessary so that the "Description" textarea could go full-width

**Before**
description-before

**After**
description-after

#### Default CasperJS timeout
The default CasperJS timeout had been raised to 1 minute. This was possible by defining the value of `this.casper.options.waitTimeout` in `page.init()`
```js
// page-objects/page.js
init: function (casper, clearDialogs) {
  // ...
  this.casper = customCasperJS(casper);
  this.casper.options.waitTimeout = 60000;
  // ...
}
```

#### Job Roles page object
Given that the markup of the job roles extension has changed, the BackstopJS page object was updated as well to use the correct css selectors

#### `civihr-ui-select` theme
The `civihr-ui-select` theme had been moved from the Shoreditch theme to the bootstrapcivihr theme (see [here](https://github.com/compucorp/org.civicrm.shoreditch/pull/14))

#### SASS variables import chain
The chain of variables import had been now encapsulated in `overrides/_variables.scss`, so that now the main sass partial needs only to import that partial instead of having to explicitly declare the whole import chain

**Before**
```scss
/* civihr.scss */
@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/overrides/variables";
@import "overrides/variables";
@import "SCSSROOT/org.civicrm.shoreditch/base/scss/overrides/variables";

@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/mixins/ALL";
@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/mixins";
```

**After**
```scss
/* civihr.scss */
@import "overrides/variables"; // contains the whole import chain within

@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/mixins/ALL";
@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/mixins";
```

---

- [x] Tests Pass


deb1990:
@AkA84 
4ae66216-439d-11e7-8cf9-9a6e5d21fa29_png__1058x2000_
Here the style looks different than before. Is it intentional?

com.civicrm.hrjobroles/CRM/Hrjobroles/Page/JobRoles.php
```diff
@@ -19,6 +19,5 @@ private static function registerScripts() {
 
     CRM_Core_Resources::singleton()->addScriptFile('com.civicrm.hrjobroles', CRM_Core_Config::singleton()->debug ? 'js/src/job-roles.js' : 'js/dist/job-roles.min.js', 1010);
     CRM_Core_Resources::singleton()->addStyleFile('com.civicrm.hrjobroles', 'css/hrjobroles.css');
-    CRM_Core_Resources::singleton()->addStyleFile('com.civicrm.hrjobroles', 'angular-xeditable/css/xeditable.css');
```
deb1990: Why this file is removed?

com.civicrm.hrjobroles/views/include/job_role_panel.html
```diff
@@ -64,7 +64,7 @@
                 </div>
                 <div class="form-group" ng-init="jobroles.initData(job_roles_data.id, 'job_contract_id', job_roles_data.job_contract_id)">
                   <label for="contract" class="col-sm-4 control-label">Contract:</label>
-                  <div class="col-sm-8">
+                  <div class="col-sm-6">
```
deb1990: Why this is `6` and rest are `5`?

deb1990:
@AkA84 What is this `First pass` that you have mentioned in many of the commits.

com.civicrm.hrjobroles/scss/hrjobroles.scss
```diff
@@ -1,124 +1,19 @@
...
-  @import "civihr/partials/forms";
-  @import "civihr/partials/iframe";
-  @import "civihr/partials/modals";
-  @import "civihr/partials/page-contact";
-  @import "civihr/partials/pagination";
-  @import "civihr/partials/panels";
-  @import "civihr/partials/tables";
-  @import "civihr/partials/tooltip";
-  @import "civihr/partials/wells";
+  &__list {
```
deb1990: Should not they go to their individual scss files?

com.civicrm.hrjobroles/scss/hrjobroles.scss
```diff
@@ -17,3 +17,10 @@
     position: relative;
   }
 }
+
+
+// workaround to make sure that the texarea always go full-width
+// the !important is necessary to override the #bootstrap-theme style
+.editable-textarea {
+  display: block !important;
```
deb1990: Instead of `!important`, if we add this class inside `.job-roles` namespace it would mean the same thing it seems. Then `!important` can be removed.

org.civicrm.bootstrapcivihr/scss/components/civihr-ui-select/_civihr-ui-select.scss
```diff
@@ -0,0 +1,40 @@
+@import 'overrides';
```
deb1990: Where did we move it from?

AkA84:
> What is this First pass that you have mentioned in many of the commits.

@deb1990 there were so many changes to do to the markup that I didn't know how many different times I had to come back and fix/adjust/amend stuff. So when I reached a point with the cleanup/refactoring that I was satisfied with as "first pass", I commited it. It's just a way to say "i've done a lot of changes, but this is just a first pass, there could be more later"

AkA84:
> Here the style looks different than before. Is it intentional?

@deb1990 yes, we're using after all a different theme (local vs Shoreditch), so it wasn't expected that the before/after would be exactly the same

com.civicrm.hrjobroles/views/modalDialog.html
```diff
@@ -1,11 +1,18 @@
 <div class="modal-header">
-    <button type="button" class="close" ng-click="cancel()"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
-    <h4 class="modal-title">{{title}}</h4>
+  <button type="button" class="close" ng-click="cancel()">
+    <span aria-hidden="true">&times;</span>
```
AkA84: `&times;` is included in the default [Bootstrap markup](http://getbootstrap.com/javascript/#modals) for modals. We use it even for L&A

com.civicrm.hrjobroles/CRM/Hrjobroles/Page/JobRoles.php
AkA84: We don't need it anymore, the style for xeditable is not included directly in Shoreditch

org.civicrm.bootstrapcivihr/scss/components/civihr-ui-select/_civihr-ui-select.scss
AkA84: I wrote it in the PR description

> The civihr-ui-select theme had been moved from the Shoreditch theme to the bootstrapcivihr theme (see here)

com.civicrm.hrjobroles/scss/hrjobroles.scss
AkA84: no, cause the `#bootstrap-theme` shoreditch namespace, which contains the default style for xeditable, will still have a higher priority (because of the ID)

com.civicrm.hrjobroles/scss/hrjobroles.scss
AkA84: well this css is so minimal that i wanted to keep things simple, we're talking about 19 lines of code here. If for some reason things get more complicated we can still split it in multiple files

com.civicrm.hrjobroles/views/include/job_role_panel.html
AkA84: The columns width was kept from the original design. It might need some rework but for now I've decided to keep it as is (the scope of the ticket was switching to a different them, not to fix/amend the layout of job roles)

com.civicrm.hrjobroles/gulpfile.js
```diff
@@ -1,21 +1,28 @@
...
-var fs = require('fs');
 
-gulp.task('sass', function (done) {
-  // The app style relies on compass's gems, so we need to rely on it
-  // for the time being
-  exec('compass compile', function (_, stdout, stderr) {
-    console.log(stdout);
-    done();
-  });
+gulp.task('sass', ['sass:sync'], function () {
```
AkA84: Was thinking of having some npm package under the compucorp's github user that we can then declare as a dependency, or something like that. But yeah we'll have to figure out something to avoid this copy-paste

com.civicrm.hrjobroles/views/mainTemplate.html
```diff
@@ -1,26 +1,30 @@
...
       ng-repeat="job_roles_data in jobroles.present_job_roles"
       ng-include
       src="jobroles.job_role_panel_url">
     </li>
   </ul>
-  <h3 ng-show="jobroles.past_job_roles.length > 0" class="header-bar header-bar-danger">Past Roles and Projects:</h3>
-  <ul class="{{prefix}}list-role">
+</section>
+<section ng-show="jobroles.past_job_roles.length">
+  <div class="page-header">
```
AkA84: I think you left a comment on an old commit. Both sections have a loader, see [here](https://github.com/civicrm/civihr/blob/c8dc86255fa87f7f18a1c797c5ff5b4b45ed117f/com.civicrm.hrjobroles/views/mainTemplate.html) and the gif i've attached in the PR description

com.civicrm.hrjobroles/CRM/Hrjobroles/Page/JobRoles.php
deb1990: Could you please document this. (Unless you already did, could not find it)

com.civicrm.hrjobroles/views/include/job_role_panel.html
deb1990: You are saying that the kept the original width but here I can see you have changed the numbers from 8 to 6?

com.civicrm.hrjobroles/scss/hrjobroles.scss
deb1990: Still I think its better to keep the style inside `.job-roles` namespace with `!important`. I mean thats what the namespace is for, to contain all styles related to `.job-roles`

com.civicrm.hrjobroles/gulpfile.js
deb1990: Yes exactly, may be we can take this up while migrating next extension's styles.

com.civicrm.hrjobroles/views/include/job_role_panel.html
AkA84: yeah it's a bit complicated, because of the grid system. The columns width had changed but visually (the end result) is kept the same between the old and new theme

com.civicrm.hrjobroles/CRM/Hrjobroles/Page/JobRoles.php
AkA84: yep, done

PR #1729, opened 2017-04-18
PCHR-2113: hrvisa activity handling fixes

davialexandre:
One of the features of the hrvisa extension is to check for expired visas and, in case this is a required document and the existing one is expired, it creates Activities to "notify" the user that it has to be updated.

In order to do this, it first checks if such Activity already exists (to avoid spamming the user with duplicated activities) and it does that by looking for activities where the Contact of the expired visa is the target. This check was done by inspecting the `targets` property returned by the API, but since the changes introduced by https://github.com/civicrm/civicrm-core/pull/9841, `targets` is not returned anymore and the code was failing.

To fix this, I first thought about replacing `targets` with something else, but then I realized that this check is not required at all. Looking at the code, there was this comment: 

> note : using filter 'activity_type_id' in combination with 'contact_id' filter doesn't work

It means, the check was only added because it was not possible, by the time the code was created back in 2013, to get an activity by both the type and the contact. Fortunately, this is not the case anymore. The API works just fine now, so I completely remove the check, which fixed the error.

Finally, I did some cleanup to fix some typos and remove the dependency on CiviUnitTestCase, which we're not using anymore on the new tests.

### DIFF
--- a/hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
+++ b/hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
@@ -153,18 +153,15 @@ class CRM_HRVisa_ActivityTest extends CiviUnitTestCase implements HeadlessInterf
       'activity_type_id' => $activityTypeId,
       'sequential' => 1,
     );
-    // note : using filter 'activity_type_id' in combination with 'contact_id' filter doesn't work
     $activities = civicrm_api3('activity', 'get', $activityGetParams);
 
     $activityId = NULL;
     $count = 0;
     foreach($activities['values'] as $val) {
-      if ($val['activity_type_id'] != $activityTypeId || !array_key_exists('targets', $val)) {
-        continue;
-      }
       $activityId = $val['id'];
       $count++;
     }
+
     return array($count, $activityId);
   }
 
--- a/hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
+++ b/hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
@@ -2,21 +2,23 @@
 
 use Civi\Test\HeadlessInterface;
 use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
 
 /**
  * Class CRM_HRVisa_ActivityTest
  *
  * @group headless
  */
-class CRM_HRVisa_ActivityTest extends CiviUnitTestCase implements HeadlessInterface , TransactionalInterface {
+class CRM_HRVisa_ActivityTest extends PHPUnit_Framework_TestCase implements HeadlessInterface , TransactionalInterface {
 
   protected $customFields;
 
   public function setUpHeadless() {
-    // check phpunitPopulateDB() to know why org.civicrm.hrdemog is installed here
     return \Civi\Test::headless()
       ->installMe(__DIR__)
+      // hrdemog is necessary because it creates the Immigration fields used by the tests
       ->install('org.civicrm.hrdemog')
+      ->install('uk.co.compucorp.civicrm.hrcore')
       ->apply();
   }
 
@@ -37,21 +39,6 @@ class CRM_HRVisa_ActivityTest extends CiviUnitTestCase implements HeadlessInterf
     $this->createLoggedInUser();
   }
 
-  protected static function _populateDB($perClass = FALSE, &$object = NULL) {
-    self::phpunitPopulateDB();
-
-    //also create 'Visa Expiration' activity type
-    $params = array(
-      'weight' => 1,
-      'label' => 'Visa Expiration',
-      'filter' => 0,
-      'is_active' => 1,
-      'is_default' => 0,
-    );
-    civicrm_api3('activity_type', 'create', $params);
-    return TRUE;
-  }
-
   /**
    * CASE 1 : is_visa_required = TRUE, and 2 migration records,
    * activity of type 'Visa Expiration' created with target contact as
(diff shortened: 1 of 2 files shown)


mickadoo:
Just to check, but the PCHR-2113 issue number is used in [multiple](https://github.com/civicrm/civihr/pull/1729) [pull](https://github.com/civicrm/civihr/pull/1710) [requests](https://github.com/civicrm/civihr/pull/1730) - this is intentionally splitting the ticket into smaller PRs?

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -20,7 +20,7 @@ public function setUpHeadless() {
       ->apply();
   }
 
-  function setUp() {
+  public function setUp() {
```
mickadoo: +1 for visibility on all methods

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -37,30 +37,27 @@ function setUp() {
...
     $params = array(
       'weight' => 1,
       'label' => 'Visa Expiration',
       'filter' => 0,
       'is_active' => 1,
       'is_default' => 0,
     );
-    $result = civicrm_api3('activity_type', 'create', $params);
+    civicrm_api3('activity_type', 'create', $params);
     return TRUE;
```
mickadoo: I've seen a mix of both, do you think we should have a newline before `return`

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -145,26 +144,23 @@ function testSyncSncenario2() {
     $this->assertEquals(CRM_Core_OptionGroup::getValue('activity_status', 'Cancelled', 'name'), $activity['values'][$activity['id']]['status_id'], 'in line ' . __LINE__ . ' Status of \'Visa Expiration\' activity should be \'Cancelled\' but wrongly is ' . $activity['values'][$activity['id']]['status_id']);
   }
 
-  function _getTargetContactActivity($contactId) {
+  private function _getTargetContactActivity($contactId) {
```
mickadoo: If we're making this private can we remove the underscore in front of it?

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -83,7 +80,7 @@ function testSyncScenario1() {
     // sync activity with contact of above details
     CRM_HRVisa_Activity::sync($cid);
 
-    // calling a common function for getting acivity a particular target contact and acitvity type
```
mickadoo: two different ways to misspell activity in the one sentence, impressive :-D

davialexandre:
I decided for submitting separate PRs for a few reasons:
- To have smaller PRs
- The issues were caused by different changes and affected different parts of the system
- https://github.com/civicrm/civihr/pull/1730 is a fix for CiviHR 1.7, and has to be merged into a different branch

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -2,25 +2,27 @@
...
 use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
 
 /**
  * Class CRM_HRVisa_ActivityTest
  *
  * @group headless
  */
-class CRM_HRVisa_ActivityTest extends CiviUnitTestCase implements HeadlessInterface , TransactionalInterface {
+class CRM_HRVisa_ActivityTest extends PHPUnit_Framework_TestCase implements HeadlessInterface , TransactionalInterface {
```
mickadoo: I like the idea of using vanilla PHPUnit framework tests

hrvisa/tests/phpunit/CRM/HRVisa/ActivityTest.php
```diff
@@ -145,42 +131,40 @@ function testSyncSncenario2() {
...
-    $import->run(
-      CRM_Extension_System::singleton()->getMapper()->keyToBasePath('org.civicrm.hrvisa')
-      . '/xml/auto_install.xml'
-    );
-    // this had to be done as demographics consists of is_visa_required field (used in unit test)
-    $import->run(
-      CRM_Extension_System::singleton()->getMapper()->keyToBasePath('org.civicrm.hrdemog')
-      . '/xml/auto_install.xml'
-    );
+  private function createLoggedInUser() {
```
mickadoo: This method is nice, clean and descriptive. I wonder would it be worth putting in a trait to be used in other tests?

davialexandre: That's a great a idea. In fact, in L&A we do have a trait that does almost the same thing: https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/SessionHelpersTrait.php

I even thought about adding that to this test, but then I found a few reasons to not do it (for now):
- That trait it's on the 1.7 branch, so I would have to copy it to the 1.6 branch in order to be able to use it here
- It works on a slightly different way than the one we have in SessionHelpersTrait
- I could create a new trait just for this test, but then I would have 2 different traits doing almost the same thing, which is not a good thing

So, my idea is to wait for this until we have 1.7 merged into staging. Also, once we have Jenkins, we'll have to fix and improve a lot of things on the existing old tests. One of these things we'll definitely be extracting this to a generic trait that can be used both by hrvisa and L&A

mickadoo: Sounds reasonable. If you're merging 1.6 into 1.7 as you're going along you might consider merging there two traits (unifying the functionality) but leaving it till later sounds fine too. I'm sure there are a lot of other things that can be cleaned up if we spend some time focusing on tests.

PR #1730, opened 2017-04-18
PCHR-2113: Make CiviHR 1.7 compatible with CiviCRM 4.7.18

davialexandre:
Up until now, CiviHR 1.7 was based on CiviCRM 4.7.9. This PR updates it to follow the changes introduced since then and make it compatible with 4.7.18.

The only problem found was related to the usage of the `SelectQuery` class, used to create some custom get* APIs (e.g. `LeaveRequest.getFull`, `Contact.getManagess`, etc). On civicrm/civicrm-core#7725, a lot of changes to support APIv4 were introduced, including changes to this class, which now is abstract and cannot be instantiated, breaking any code which was using the `SelectQuery` class.

To fix this, the *Select classes now use the Api3SelectQuery class instead. For L&A, a new `CRM_HRLeaveAndAbsences_API_Query_Select` was added. It extends `Api3SelectQuery`, but introduces a new constructor which accepts the params array. Internally, it uses this array to set all the necessary properties. This new class is now used by `ContactSelect` and `LeaveRequestSelect`, exactly where `SelectQuery` was being used before.

For hrcomments, a similar approach was used, but instead of adding a new class, a new method was added.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/Select.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/Select.php
@@ -1,0 +1,49 @@
+<?php
+
+use Civi\API\Api3SelectQuery;
+
+/**
+ * This class is just a wrapper around Civi\API\Api3SelectQuery.
+ *
+ * It introduces a new constructor which accepts the API params and then uses it
+ * to prepare the internal query properties.
+ */
+class CRM_HRLeaveAndAbsences_API_Query_Select extends Api3SelectQuery {
+
+  /**
+   * CRM_HRLeaveAndAbsences_API_Query_Select constructor.
+   *
+   * @param string $entity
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/Select.php
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Setup things like limit, order, offset and fields to return
+   *
+   * @param array $params
+   */
+  private function setupQueryFromParams($params) {
+    $this->where = $params;
+
+    $options = _civicrm_api3_get_options_from_params($params);
+    $this->select = array_keys(array_filter($options['return']));
```
tunbola: Wondering if we need to make provision for 'is_count' option like its done here: [Here](https://github.com/civicrm/civicrm-core/pull/7725/files#diff-3076917760efeb467459726ebb6af282R1344)
?

PR #1738, opened 2017-04-21
PCHR-2027: Add delete method to leave request instance

dhavalsharma:
## Overview
Add delete method to leave request instance

- [ X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -741,7 +741,8 @@ define([
     });
 
     describe('attachments', function () {
-      var promise, test_id = '63', attachments, numberOfFiles;
+      var promise, test_id = '63',
```
deb1990: Please keep the variables which has values in the end, rest in alphabetical order
```
var attachments, numberOfFiles, promise, 
test_id = '63';
```

PR #1743, opened 2017-04-21
PCHR-2142: Send E-mail notifications when a Leave Request is created or updated

tunbola:
## Overview
This PR introduces sending of Email-notifications when a Leave Request is created or updated. The Emails are sent out to the Leave contact and the Leave approvers(s) (and If there are no leave approvers for the contact, then the default leave approver for the absence type for that request). The emails are sent out for Leave/TOIL/Sickness Request Types. Each of the leave types has its own pre-configured email templates that are inserted into the database on installation of the L&A extension and can be edited at _civicrm/admin/messageTemplates?reset=1_. 

## Before
_No E-mail was sent when creating or updating Leave/Sick/TOIL requests_

## After
Emails are sent out when a Leave/Sick/TOIL request is created or updated. The emails are based on the pre-configured email templates for each of the request types.
Below are screenshots of the emails sent for each of the request types.

https--mailtrap io-api-v1-inboxes-202193-messages-414989642-body html-api_token 3077ffb668cebf9b4384ae681703064f 2017-04-24 11-49-45

https--mailtrap io-api-v1-inboxes-202193-messages-414993042-body html-api_token 3077ffb668cebf9b4384ae681703064f 2017-04-24 11-52-11

https--mailtrap io-api-v1-inboxes-202193-messages-414990394-body html-api_token 3077ffb668cebf9b4384ae681703064f 2017-04-24 11-52-37

## Technical Details
The email sending process  is handled by the Template classes, The Message class and the LeaveRequestMailNotificationSender Service. The function of each of the classes are highlighted below:

- The Template classes: SicknessRequestNotificationTemplate, LeaveRequestNotificationTemplate and TOILRequestNotificationTemplate extend the base BaseRequestNotificationTemplate class and are responsible for building the template parameters needed in the smarty email template and also for fetching the email templateID in the 'civicrm_msg_template' table.
- The Message class is responsible for building the email information such as the recipients, from email and uses a template factory class to get the required Template class needed to get template parameters and templateID for the template.
- The LeaveRequestMailNotificationSender service is responsible for sending the email, It accepts a Message object in its send method and uses that to get the information it needs before finally sending the mail through the `CRM_Core_BAO_MessageTemplate::sendTemplate` core method

`CRM_Core_BAO_MessageTemplate::sendTemplate` core method  is also used internally to send emails by the Civi Events component. A detailed explanation can be found in the prior investigation done here [gist](https://gist.github.com/tunbola/0c95c063fd42ebb4592b7ec47219a6eb).
 
For unit tests to test email sending functionality, `define('CIVICRM_MAILER_SPOOL', 1);` was set in the phpunit bootstrap file. This will make civi redirect all emails to be sent to the database.

## Comments

For this PR, the email templates are basic stripped down version with little or no styling but with the logic and all necessary information already available in the templates

The leave request link in this email currently leads to the leave requests reports page. [PCHR-2162](https://compucorp.atlassian.net/browse/PCHR-2162) has been created to allow a leave modal to be opened by default for a particular leave request from a URL since this functionality is not available currently.

Also the From email for the Email notifications currently picks the email address of the contact attached to the default domain created in the `civicrm_domain` table. This logic will soon be changed to pick the From email address and name from the L&A general settings configuration

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+   * @return array
+   */
+  public function getTemplateParameters(LeaveRequest $leaveRequest) {
+    $templateParameters =  [
+      'leaveComments' => $this->getLeaveComments($leaveRequest),
+      'leaveFiles' => $this->getAttachments($leaveRequest),
+      'leaveRequestStatuses' => $this->getLeaveRequestStatuses(),
+      'fromDate' => $this->formatLeaveRequestDate($leaveRequest->from_date),
+      'toDate' => $this->formatLeaveRequestDate($leaveRequest->to_date),
+      'fromDateType' => $this->getLeaveRequestDayTypes()[$leaveRequest->from_date_type],
```
davialexandre: Instead of using array dereferencing here, I think this will look better if you passe the values as a parameter, and then the method returns the label (the method name also has to be updated to say it returns a label. Something like this:

```php
$this->getLeaveRequestDayTypeLabel($leaveRequest->from_date_type);
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+        $templateParameters['sicknessRequiredDocuments'] = $this->getSicknessRequiredDocuments();
+        $templateParameters['leaveRequiredDocuments'] =  explode(',', $leaveRequest->sickness_required_documents);
+      }
+    }
+
+    return $templateParameters;
+  }
+
+  /**
+   * get Email template to be used based on request type
```
davialexandre: s/get/Get/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+        $templateName = 'CiviHR Leave Request Notification';
+        break;
+      case LeaveRequest::REQUEST_TYPE_TOIL:
+        $templateName = 'CiviHR TOIL Request Notification';
+        break;
+    }
+
+    $params = ['msg_title' => $templateName, 'is_default' => 1];
+    $defaults = [];
+    $template = MessageTemplate::retrieve($params, $defaults);
```
davialexandre: Please use the `MessageTemplate.get` API instead

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+
+    return $template;
+  }
+
+  /**
+   * Gets the From Email address configured on the site
+   * Currently It gets the from email address and name of the contact linked to the
+   * default domain in civicrm_domain table.
+   *
+   * This logic will be changed later to fetch from L&A general settings
```
davialexandre: Add a `@TODO` for this paragraph

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+      'is_primary' => 1,
+      'api.Contact.get' => ['id' => '$value.contact_id', 'return' => ['display_name']],
+      'return' => ['email','contact_id']
+    ]);
+
+    return $result['values'];
+  }
+
+  /**
+   * Gets the contactID's of Leave approvers for the current leave request
```
davialexandre: I think this will describe this method a bit better: "Gets the Contact ID's of the currently active Leave Approvers of the given Leave Request"

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
+<?php
+
```
davialexandre: This Service has 3 responsibilities: 
- Build the e-mail template
- Build the message (which is the template + from/to addresses)
- Send the e-mail

We should separate these into different classes. What I'm thinking is something like this:
- 3 message template classes (one for each request type), with methods to build the necessary params. These will contain all the template logic that currently is in this class
- A factory to return the right template for a Leave Request instance
- The LeaveRequestMailNotification will represent the item `#2` on the list (it will be the message), and hence it will make sense to have public methods like getRecipients. We should inject template factory here.
- Finally, we'll have a LeaveRequestMailNotificationSender service, which will send the notifications

Since neither the notification nor the template are services, I think we can have a new Email namespace for this: `CRM_HRLeaveAndAbsences_Mail`. Inside it, we can have `CRM_HRLeaveAndAbsences_Mail_Template` for the templates and `CRM_HRLeaveAndAbsences_Mail_Message` for the message class. 

What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotificationTest.php
```diff
@@ -0,0 +1,303 @@
...
+  use CRM_HRLeaveAndAbsences_SessionHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+
+
+  private $leaveRequestMailNotification;
+
+  private $leaveContact;
+
+  public function setUp() {
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0;");
```
davialexandre: use single quotes here

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Contact.php
```diff
@@ -21,4 +21,14 @@ public static function fabricate($params = []) {
 
     return array_shift($result['values']);
   }
+
+  public static function fabricateWithEmail($params, $email) {
+    $contact = self::fabricate($params);
```
davialexandre: Is this method really necessary? Did you try just pass the e-mail among the params? Like this: `fabricate(['name' => 'bla bla bla', 'email' => 'bla@blabla.bla'])`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  /**
+   * Returns an array of emails of eligible recipients for this leave Request Notification
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return array
+   */
+  public function getRecipientEmails(LeaveRequest $leaveRequest) {
```
davialexandre: Why is this method public?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  /**
+   * get Email template to be used based on request type
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return \CRM_Core_DAO_MessageTemplate
+   */
+  public function getTemplate(LeaveRequest $leaveRequest) {
```
davialexandre: Why is this method public?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  /**
+   * Return parameters to be used in the Email smarty template
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return array
+   */
+  public function getTemplateParameters(LeaveRequest $leaveRequest) {
```
davialexandre: Why is this public?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2065,4 +2070,59 @@ public function testLeaveRequestIsDeletedValueCanNotBeSetWhenCreatingALeaveReque
...
+
+    $leaveContact = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Staff1', 'last_name' => 'Staff1'], 'staffmember@dummysite.com'
+    );
+
+    $this->setLeaveApproverRelationshipTypes(['has Leaves Approved By']);
+
+    // Set manager1 and manager2 only to be leave aprovers for the leave contact
+    $this->setContactAsLeaveApproverOf($manager1, $leaveContact, null, null, true, 'has Leaves Approved By');
+    $this->setContactAsLeaveApproverOf($manager2, $leaveContact, null, null, true, 'has Leaves Approved By');
```
davialexandre: You're already testing all the recipients logic on the email classes, so keep it as simple as possible here (that is, use just a single approver)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2065,4 +2070,59 @@ public function testLeaveRequestIsDeletedValueCanNotBeSetWhenCreatingALeaveReque
...
+      'to_date' => CRM_Utils_Date::processDate('tomorrow'),
+      'to_date_type' => 1,
+      'toil_to_accrue' => 2,
+      'toil_duration' => 120,
+      'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE
+    ], false);
+
+    //emails redirected to the database are stored in the message spool table
+    $messageSpoolTable = CRM_Mailing_BAO_Spool::getTableName();
+    $query = "SELECT * FROM {$messageSpoolTable} WHERE recipient_email
```
davialexandre: please add a helper trait/method for this

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2065,4 +2070,59 @@ public function testLeaveRequestIsDeletedValueCanNotBeSetWhenCreatingALeaveReque
     $leaveRequestRecord = LeaveRequest::findById($leaveRequest->id);
     $this->assertEquals(0, $leaveRequestRecord->is_deleted);
   }
+
+  public function testEmailGetsSentWhenLeaveRequestIsCreated() {
```
davialexandre: what about adding a similar test, but for when the Leave Request is updated?

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -385,6 +385,33 @@ CREATE TABLE `civicrm_hrleaveandabsences_contact_work_pattern` (
 )  ENGINE=InnoDB DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci  ;
 
 -- ********************************************************
+-- *  Insert Leave Request Email Notification Templates
+-- ********************************************************
+
+INSERT INTO `civicrm_msg_template` (
```
tunbola: Will check this out.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
tunbola: Sounds like a great idea!

Except I'm not sure IF having one template class for each request type will be the best for now.
- Currently the only two template parameters that is added conditionally is just the sickness reason and required documents.
 All the other template logic to fetch parameters are actually shared by all the request types.
- Since all the request types are in the same table, an attribute peculiar to each request type can be accessed in the template once the leave request object is available in each template.
What do you think?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Contact.php
tunbola: You right. That actually worked.
My initial thought was that since emails are stored in separate tables, a separate API call is needed for that.
Thanks.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: I wanted to, but I thought it will be a repetition, basically the hook gets triggered, either for an update or create. 
I'll add it though.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
davialexandre: For the classes what you think about this:
- have an abstract BaseLeaveRequestNotificationTemplate, with all the logic that is shared among the templates
- have a LeaveRequestNotificationTemplate and a TOILRequestNotificationTemplate extending this base abastract class (it will just be an extends, without actually overriding anything)
- have a SicknessRequestNotificatioTemplate which will have the logic for parameters specific for this type

About all the templates having access to the whole LeaveRequest, I think that this is fine. The template idea is just to separate the different logic for each type, but they're all still based on a LeaveRequest.

tunbola: Ok. Sounds good.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotification.php
```diff
@@ -0,0 +1,391 @@
...
+
+  /**
+   * Gets the Attachments associated with this LeaveRequest
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return array
+   */
+  private function getAttachments(LeaveRequest $leaveRequest) {
+    $result = civicrm_api3('LeaveRequest', 'getAttachments', [
```
tunbola: The attachment service does not have a get method.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Contact.php
tunbola: I'll probably keep this method, For one reason or the other when I pass in the email parameter, the display_name is set to the email irrespective of what ! pass in the display_name parameter or first_name and last_name. Looks like a bug.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2065,4 +2071,106 @@ public function testLeaveRequestIsDeletedValueCanNotBeSetWhenCreatingALeaveReque
...
+    $manager1 = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Manager1', 'last_name' => 'Manager1'], 'manager1@dummysite.com'
+    );
+
+    $leaveContact = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Staff1', 'last_name' => 'Staff1'], 'staffmember@dummysite.com'
+    );
+
+    $this->setLeaveApproverRelationshipTypes(['has Leaves Approved By']);
```
davialexandre: Do we really need to specify the relationship name? Can't we use the default one?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
davialexandre: mmmmm... since both tests are pretty much the same, what about using the same test for both? After checking that the messages were sent on create, we empty the spool table, update the leave request and then check if new messages were sent.

Usually it's better to tests things separately, but given the amount of code duplicate here and that we're actually testing the same lines of code, I think it's fine to make an exception

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Factory/RequestMailNotificationTemplateTest.php
```diff
@@ -0,0 +1,38 @@
...
+ * Class CRM_HRLeaveAndAbsences_Factory_RequestMailNotificationServiceTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Factory_RequestMailNotificationServiceTest extends BaseHeadlessTest  {
+
+  private $leaveRequestTemplateFactory;
+
+  public function setUp() {
+    CRM_Core_DAO::executeQuery('SET foreign_key_checks = 0;');
```
davialexandre: We're not touching the DB on this class's tests, so we don't need this here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/LeaveRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Class CRM_HRLeaveAndAbsences_Mail_LeaveRequestNotificationTemplateTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Mail_LeaveRequestNotificationTemplateTest extends BaseHeadlessTest {
+
+  use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+
```
davialexandre: please remove one of these two empty lines here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/LeaveRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,101 @@
...
+    $params = [
+      'leave_request_id' => $leaveRequest->id,
+      'text' => 'Random Commenter',
+      'contact_id' => $leaveRequest->contact_id,
+      'sequential' => 1
+    ];
+
+    $leaveRequestCommentService = new LeaveRequestCommentService();
+    $leaveRequestCommentService->add($params);
+    $leaveRequestCommentService->add(array_merge($params, ['text' => 'Sample text']));
```
davialexandre: maybe we should have a `createCommentForLeaveRequest()` helper method as well and simplify this a little bit?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_NotificationReceiver as NotificationReceiver;
+use  CRM_HRLeaveAndAbsences_Mail_Message as Message;
```
davialexandre: please remove the additional space on this line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+ * Class CRM_HRLeaveAndAbsences_Mail_MessageTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Mail_MessageTest extends BaseHeadlessTest {
+
+  use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+  use CRM_HRLeaveAndAbsences_MessageHelpersTrait;
+
```
davialexandre: please remove one of these two empty lines

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+      'first_name' => 'Manager1', 'last_name' => 'Manager1'], 'manager1@dummysite.com'
+    );
+    $manager2 = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Manager2', 'last_name' => 'Manager2'], 'manager2@dummysite.com'
+    );
+    $manager3 = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Manager3', 'last_name' => 'Manager3'], 'manager3@dummysite.com'
+    );
+
+    $this->setLeaveApproverRelationshipTypes(['has Leaves Approved By']);
```
davialexandre: can't we use the default relationship type here too?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+
+    // Set manager1 and manager2 only to be leave aprovers for the leave contact
+    $this->setContactAsLeaveApproverOf($manager1, $this->leaveContact, null, null, true, 'has Leaves Approved By');
+    $this->setContactAsLeaveApproverOf($manager2, $this->leaveContact, null, null, true, 'has Leaves Approved By');
+
+    $leaveRequest = LeaveRequest::create([
```
davialexandre: Please use the LeaveRequest Fabricator here and make this a bit shorter by relying on its default params

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+    $message = new Message($leaveRequest, $this->leaveRequestTemplateFactory);
+    $recipientEmails = $message->getRecipientEmails($leaveRequest);
+
+    $this->assertCount(3, $recipientEmails);
+
+    //The contact and the leave approvers are eligible recipients for this email notification.
+    foreach($recipientEmails as $value) {
+      $this->assertContains($value['email'], [
+        'staffmember@dummysite.com', 'manager1@dummysite.com', 'manager2@dummysite.com'
+      ]);
```
davialexandre: There's a problem with this assertion. If all the e-mail on the recipients is the same (which is a bug), the test will still pass. 

A safer way to handle this would be:
- get all the emails (probably using array_column())
- sort them, so we can always have them in an expected order
- finally compare the array of expected emails with the array of returned emails

This applies to all the other tests where you're doing similar checks

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+    );
+    $defaultApprover2 = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Approver2', 'last_name' => 'Approver2'], 'approver2@dummysite.com'
+    );
+
+    $absenceType = 1;
+    //add two default leave approvers for the absence type
+    NotificationReceiver::addReceiversToAbsenceType($absenceType, [$defaultApprover1['id'], $defaultApprover2['id']]);
+
+    $leaveRequest = LeaveRequest::create([
```
davialexandre: use the fabricator here too

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+    $message = new Message($leaveRequest, $this->leaveRequestTemplateFactory);
+    $this->assertEquals($leaveRequest->contact_id, $message->getLeaveContactID());
+  }
+
+  public function testGetTemplateParameters() {
+    $notificationTemplateFactory = $this->createRequestNotificationTemplateFactoryMock();
+
+    $leaveRequest = new LeaveRequest();
+    $message = new Message($leaveRequest, $notificationTemplateFactory);
+    $message->getTemplateParameters();
```
davialexandre: I might be missing something, but it looks like you're not really testing anything here.

What I would expect from this test is:
- Mock the template and make it return the expected params
- Mock the factory to return the mocked template
- Assert that the message returns the expected params

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  public function testGetTemplateID() {
+    $notificationTemplateFactory = $this->createRequestNotificationTemplateFactoryMock();
+
+    $leaveRequest = new LeaveRequest();
+    $message = new Message($leaveRequest, $notificationTemplateFactory);
+    $message->getTemplateID();
```
davialexandre: Looks like you're not testing anything here as well

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/SicknessRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Mail_SicknessRequestNotificationTemplate as SicknessRequestNotificationTemplate;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+
```
davialexandre: please remove one of these 2 empty lines here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/SicknessRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Class RM_HRLeaveAndAbsences_Mail_SicknessRequestNotificationTemplateTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Mail_SicknessRequestNotificationTemplateTest extends BaseHeadlessTest {
+
+  use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+
```
davialexandre: please remove one of these two empty lines here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/SicknessRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,108 @@
...
+    $this->leaveRequestDayTypes = $this->getLeaveRequestDayTypes();
+  }
+
+  public function testGetTemplateReturnsTheCorrectTemplate() {
+    $template = $this->sicknessRequestNotificationTemplate->getTemplate();
+    $this->assertEquals($template['msg_title'], 'CiviHR Sickness Record Notification');
+  }
+
+  public function testGetTemplateParametersReturnsTheExpectedParametersForTheTemplate() {
+    $leaveRequest = LeaveRequest::create([
```
davialexandre: use the fabricator

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/SicknessRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,108 @@
...
+    $params = [
+      'leave_request_id' => $leaveRequest->id,
+      'text' => 'Random Commenter',
+      'contact_id' => $leaveRequest->contact_id,
+      'sequential' => 1
+    ];
+
+    $leaveRequestCommentService = new LeaveRequestCommentService();
+    $leaveRequestCommentService->add($params);
+    $leaveRequestCommentService->add(array_merge($params, ['text' => 'Sample text']));
```
davialexandre: if you create the helper I've suggested, then you can use it here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/TOILRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,102 @@
...
+    $this->leaveRequestDayTypes = $this->getLeaveRequestDayTypes();
+  }
+
+  public function testGetTemplateReturnsTheCorrectTemplate() {
+    $template = $this->toilRequestNotificationTemplate->getTemplate();
+    $this->assertEquals($template['msg_title'], 'CiviHR TOIL Request Notification');
+  }
+
+  public function testGetTemplateParametersReturnsTheExpectedParametersForTheTemplate() {
+    $leaveRequest = LeaveRequest::create([
```
davialexandre: use the fabricator

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/TOILRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Class CRM_HRLeaveAndAbsences_Mail_TOILRequestNotificationTemplateTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Mail_TOILRequestNotificationTemplateTest extends BaseHeadlessTest {
+
+  use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+
```
davialexandre: remove empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/TOILRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_Mail_TOILRequestNotificationTemplate as TOILRequestNotificationTemplate;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+
```
davialexandre: remove empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/TOILRequestNotificationTemplateTest.php
```diff
@@ -0,0 +1,102 @@
...
+    $params = [
+      'leave_request_id' => $leaveRequest->id,
+      'text' => 'Random Commenter',
+      'contact_id' => $leaveRequest->contact_id,
+      'sequential' => 1
+    ];
+
+    $leaveRequestCommentService = new LeaveRequestCommentService();
+    $leaveRequestCommentService->add($params);
+    $leaveRequestCommentService->add(array_merge($params, ['text' => 'Sample text']));
```
davialexandre: use the helper method, if you have created it

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotificationSenderTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+use  CRM_HRLeaveAndAbsences_Mail_Message as Message;
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestMailNotificationSender as LeaveRequestMailNotificationSenderService;
+use CRM_HRLeaveAndAbsences_Factory_RequestNotificationTemplate as RequestNotificationTemplateFactory;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+
```
davialexandre: remove empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotificationSenderTest.php
```diff
@@ -0,0 +1,71 @@
...
+ * Class CRM_HRLeaveAndAbsences_Service_LeaveRequestMailNotificationSenderTest
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_Service_LeaveRequestMailNotificationSenderTest extends BaseHeadlessTest {
+
+  use CRM_HRLeaveAndAbsences_LeaveRequestHelpersTrait;
+  use CRM_HRLeaveAndAbsences_SessionHelpersTrait;
+  use CRM_HRLeaveAndAbsences_LeaveManagerHelpersTrait;
+
```
davialexandre: remove empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotificationSenderTest.php
```diff
@@ -0,0 +1,71 @@
...
+      'first_name' => 'Staff1', 'last_name' => 'Staff1'], 'staffmember@dummysite.com'
+    );
+  }
+
+  public function testSendRetunsTrue() {
+    $manager1 = ContactFabricator::fabricateWithEmail([
+      'first_name' => 'Manager1', 'last_name' => 'Manager1'], 'manager1@dummysite.com'
+    );
+
+    $this->setLeaveApproverRelationshipTypes(['has Leaves Approved By']);
```
davialexandre: do we need this?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestMailNotificationSenderTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    $leaveMailSenderService = new LeaveRequestMailNotificationSenderService();
+
+    $mailStatus = $leaveMailSenderService->send($message);
+
+    //The email was sent to the leave contact and leave approver
+    $this->assertCount(2, $mailStatus);
+
+    //since the emails are redirected to the database, the status is expected to be true for each email address
+    foreach($mailStatus as $email => $status) {
```
davialexandre: To me, it makes more sense to check if the was actually sent, which, in this case, means: check if it's in the spool table

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/MessageHelpersTrait.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_MessageHelpersTrait {
```
davialexandre: Since this helper doesn't have only Message related helpers, but helpers related to all the classes inside the Mail namespace (and relatives), why don't we rename it to `MailHelpersTrait`?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
```diff
@@ -0,0 +1,152 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
+
+abstract class CRM_HRLeaveAndAbsences_Mail_BaseRequestNotificationTemplate {
```
davialexandre: Just to keep things a bit more organized, can we move the templates to their own folder? That is, rename them to: `CRM_HRLeaveAndAbsences_Mail_Template_<Name>`.

Example: `CRM_HRLeaveAndAbsences_Mail_Template_BaseRequestNotification`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
```diff
@@ -0,0 +1,152 @@
...
+    $this->leaveRequestCommentService = $leaveRequestComment;
+  }
+
+  /**
+   * Gets the message template for this Template Type.
+   *
+   * @return array
+   *  An array containing the message template attributes/values
+   */
+  abstract public function getTemplate();
```
tunbola: I initially named it getTemplateID() but then I found it difficult to test the method in the child template classes since the ID is not fixed but just an auto increment ID in the database. 
I need it to return all the other details of the template so that the test can use that to ascertain the right template was returned. 
What about changing the name to getTemplateDetails?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Gets the Attachments associated with this LeaveRequest
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   *
+   * @return array
+   */
+  private function getAttachments(LeaveRequest $leaveRequest) {
+    $result = civicrm_api3('LeaveRequest', 'getAttachments', [
```
tunbola: The LeaveAttachment service does not have a get Method. We agreed to this because the way attachments are created is different so we did not put the get method on the service. So the service only has the delete method.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/Message.php
```diff
@@ -0,0 +1,185 @@
...
+   * Gets the From Email address configured on the site
+   * Currently It gets the from email address and name of the contact linked to the
+   * default domain in civicrm_domain table.
+   *
+   * @TODO This logic will be changed later to fetch from L&A general settings
+   *
+   * @return string
+   */
+  public function getFromEmail() {
+    if (is_null($this->fromEmail)) {
```
tunbola: Well, I copied the code from Civi Events in CORE to use temporarily.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
tunbola: Well its actually asserting that the getTemplateParameters of the Template class is at least called. 
All the method does is this
`  public function getTemplateParameters() {
    return $this->getTemplate()->getTemplateParameters($this->leaveRequest);
  }`
and since the `getTemplateParameters` had already been tested, I only wanted to make sure that the message class calls the method. 
What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
davialexandre: I still think it should return the ID, because that's what the actual code needs. On the tests, can't you load the Message Template by name (using the API) and then compare its ID with the value returned by the `getTemplateID()` method?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
davialexandre: I saw your previous reply about this but I forgot it and asked again... (facepalm)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Mail/BaseRequestNotificationTemplate.php
tunbola: Sounds good.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/MessageTest.php
davialexandre: The problem is that you're not fully testing it. Currently, you only assert that the method was called, but not if it returned the expected value. So, if for some reason `getTemplateParameters()` stops returning anything or returns something different, the test will still pass

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Factory/RequestNotificationTemplate.php
```diff
@@ -2,9 +2,9 @@
 
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
-use CRM_HRLeaveAndAbsences_Mail_SicknessRequestNotificationTemplate as SicknessRequestNotificationTemplate;
-use CRM_HRLeaveAndAbsences_Mail_TOILRequestNotificationTemplate as TOILRequestNotificationTemplate;
-use CRM_HRLeaveAndAbsences_Mail_LeaveRequestNotificationTemplate as LeaveRequestNotificationTemplate;
+use CRM_HRLeaveAndAbsences_Mail_Template_SicknessRequestNotificationTemplate as SicknessRequestNotificationTemplate;
```
davialexandre: The idea of adding a new Template folder/namespace was to be able to remove it from the class name. So, instead of 
`CRM_HRLeaveAndAbsences_Mail_Template_SicknessRequestNotificationTemplate` you would have `CRM_HRLeaveAndAbsences_Mail_Template_SicknessRequestNotification`

PR #1746, opened 2017-04-24
PCHR-2003: Margin bottom for nav-pills

deb1990:
## Overview
As per the design of CiviHR the `nav-pills-horizontal`should have a bottom margin, which has been added.
This PR is meant for Leave and Absence but as this design should be part of 1.6 version also, PR has been opened to the staging branch.

## Before
civihr_staff_compucorp_co_uk___hr17

## After
civihr_staff_compucorp_co_uk___hr17


org.civicrm.bootstrapcivihr/scss/overrides/style/_navs.scss
```diff
@@ -0,0 +1,3 @@
+.nav-pills {
```
AkA84: not `.nav-pills`, but `.nav-pills-horizontal`

PR #1747, opened 2017-04-24
AI-91: Fix contact record membership tab issues

mukeshcompucorp:
## Before
membership-tab-issues

## After
resolved-membership-tab-issues

## Comments
Scrollable modal issue are already resolved by AI-101 task and deployed.

---

- [ ] Tests Pass


org.civicrm.bootstrapcivicrm/scss/civicrm/common/_base.scss
```diff
@@ -76,6 +76,13 @@
       }
     }
   }
+
+  .btn-slide:after {
+    background: none;
+    content: $fa-var-ellipsis-v;
```
meladawy: Can you use fa-icon mixin instead? This is how to use it.
https://github.com/civicrm/civihr/blob/14630873545ad5816a42ab1bda585b867a322541/org.civicrm.bootstrapcivicrm/scss/civicrm/common/_datepickers.scss#L48 (also please change font size with a variable)

mukeshcompucorp: Fixed.

PR #1761, opened 2017-04-28
AI-98: Fix manage groups issues

mukeshcompucorp:
## Before
group-issues

## After
resolved-group-issues

---

- [ ] Tests Pass


org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_manage-groups.scss
```diff
@@ -242,33 +242,44 @@
...
 
   #DataTables_Table_0 {
     @include mx-expandable-table();
 
+    .crm-group-group_links {
+      text-align: right;
+
+      .btn-slide.crm-hover-button {
+        display: inline-block;
+        text-indent: -999999px;
```
meladawy: Do you need overflow:hidden in this? Some browsers just convert the text-indent to scroll so its important to have a wrapping element that hide any overflow.

bhupatbheda: I have tested this in all browser now and it's not creating problem,

mukeshcompucorp: I have tested in all browser all looks good no scrolling.

PR #1762, opened 2017-04-28
PCHR-2161: Fix issues with Leave Days amount is not saved correctly on a Work Pattern

tunbola:
## Overview
When saving/editing a Work Pattern, it's possible to set the Leave Days amount for a working day. Currently, if you select any amount other than 1 Day, the saved amount will be incorrect.

## Before
When saving/editing the leave day amount for a working day in a Work Pattern other than 1 Day, the saved amount will be incorrect.
The reason for this is that the column for leave_days in the Work Day table is integer and some of the leave day values are float e.g 0.5 for Half day. When these values are saved, they get rounded up to integers in the database i.e 0.5 becomes 1 and 1.5 becomes 2.

## After
This PR fixes that by changing the leave_days column to VARCHAR and also binding the column to the leave_days_amount option group. The obvious solution would have been to change to change the column type to decimal but since its linked to the leave_days_amount option group, the column has to be VARCHAR to be same as the option value column type storing the leave day amounts.

When saving/editing the leave day amount for a working day in a Work Pattern, It gets saved correctly.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkDayTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkDayTest.php
@@ -183,6 +183,37 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDayTest extends BaseHeadlessTest {
     $this->assertSame((string)$workDayTypes['weekend'], WorkDay::getWeekendTypeValue());
   }
 
+  public function testCreateWorkDayWithLeaveDaysFromLeaveDaysAmountOptionGroup() {
+    $leaveDayAmounts = array_flip(WorkDay::buildOptions('leave_days', 'validate'));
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkDayTest.php
```diff
@@ -183,6 +183,37 @@ public function testGetWeekendTypeValueReturnsTheOptionValueValueAsString() {
     $this->assertSame((string)$workDayTypes['weekend'], WorkDay::getWeekendTypeValue());
   }
 
+  public function testCreateWorkDayWithLeaveDaysFromLeaveDaysAmountOptionGroup() {
```
davialexandre: This test is too complex and in the end it's not clear what we're really testing. To make it simpler, we can focus on what is really important, which is saving decimal amounts.

We can just check that it can save different decimals amounts like 0.34, 1.5, 1.25 etc (just a few variations are enough)

PR #1771, opened 2017-05-02
AI-83: Fix blue and red bars on groups tab

mukeshcompucorp:
## Before
bars-issues

## After
resolved-bars-issues

---

- [x] Tests Pass


meladawy:
PR title ticket number AI-83 is different than commit title ticket number AI-85. Please unify this

org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_manage-groups.scss
```diff
@@ -132,55 +126,51 @@
...
     } 
 
     .description {
-      margin: 15px;
       font-family: $font-family-base;
+      margin: 15px;      
     }
 
     #_qf_GroupContact_next {
+      background: $brand-primary url(../../../../../i/plus-icon.png) 8px 11px no-repeat;
```
meladawy: Well, i believe this should be changed. I can see that you are trying to get image from outside the extension itself "org.civicrm.bootstrapcivicrm". Images should be come from the same extension and shouln't rely on any other extension. Please try to use fontawesome if possible or create image directory in the same extension and put ur images inside it. I do recommend the font awesome solution if possible

mukeshcompucorp: Actually, It's used sometime before. It's not recent change. Though, I have updated it and use the image from same extension folder. We can't use font-awesome because it's used for input button and font-awesome will not work for input button.

PR #1776, opened 2017-05-02
PCHR-1826: Hide Option Value Edit Field

mickadoo:
## Overview
All option values have a "value" field which acts like their ID. If this value is changed it could cause major problems since it acts like a database foreign key. For example changing the "value" for an absence type would change all absences created in the system that point to that absence type. The aim of this PR is to hide the option to edit the value.

## Before
From the admin option edit screen (/civicrm/admin/options) you can change the value.
selection_181

## After
From the admin option edit screen (/civicrm/admin/options) the value field is not displayed.

selection_180

- [x] Tests Pass

### DIFF
--- a/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
+++ b/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
@@ -13,11 +13,6 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
 
   use HRUITrait;
 
-  protected $requiredExtensions = [
-    'uk.co.compucorp.civicrm.tasksassignments',
-    'org.civicrm.hrcase'
-  ];
-
   /**
    * @return CiviEnvBuilder
    */
@@ -27,7 +22,6 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
       ->install('org.civicrm.hrident')
       ->install('uk.co.compucorp.civicrm.tasksassignments')
       ->installMe(__DIR__)
-      ->install($this->requiredExtensions)
       ->apply();
   }
 
--- a/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
+++ b/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
@@ -62,9 +62,11 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
    * @return int
    */
   private function createContact($firstName, $lastName) {
-    $params = ['first_name' => $firstName, 'last_name' => $lastName];
-
-    return ContactMaker::fabricate($params)['id'];
+    return ContactMaker::fabricate([
+        'first_name' => $firstName,
+        'last_name' => $lastName
+      ]
+    )['id'];
   }
 
 }
--- a/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
+++ b/hrui/tests/phpunit/CRM/HRUI/HelperTest.php
@@ -3,6 +3,7 @@
 use Civi\Test\HeadlessInterface;
 use Civi\Test\TransactionalInterface;
 use Civi\Test\CiviEnvBuilder;
+use CRM_HRCore_Test_Fabricator_Contact as ContactMaker;
 
 /**
  * Class CRM_HRUI_HelperTest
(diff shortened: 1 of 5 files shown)


hrui/tests/phpunit/HRUITrait.php
```diff
@@ -35,11 +37,27 @@ protected function createRelationship($contactA, $contactB, $relationship_type)
...
+
     civicrm_api3('Relationship', 'create', array(
       'contact_id_a' => $contactA,
       'contact_id_b' => $contactB,
       'relationship_type_id' => $relationshipType['id'],
     ));
   }
 
+  /**
+   * @param $labelAtoB
```
tunbola: Missing param type

hrui/tests/phpunit/HRUITrait.php
```diff
@@ -35,11 +37,27 @@ protected function createRelationship($contactA, $contactB, $relationship_type)
...
+   * @param $labelBtoA
```
tunbola: Missing param type

hrui/tests/phpunit/HRUITrait.php
```diff
@@ -8,17 +8,19 @@
...
-   * @return int return the contact ID
-   * @throws \CiviCRM_API3_Exception
+   * @param $firstName
+   * @param $lastName
+   *
+   * @return mixed
    */
-  protected function createContact($params) {
+  protected function createContact($firstName, $lastName) {
     $result = civicrm_api3('Contact', 'create', array(
```
tunbola: A problem with this function:

- You can only create a Contact with first_name and last_name, In case there's a need to pass in other parameters when creating a contact, It can't be passed.

I think the function is supposed to work like a fabricator with some default parameters that can be overridden and the parameters passed straight to the contact API.
I think you should use the contact fabricator [Here](https://github.com/civicrm/civihr/blob/8602dcdb97bd5aa4471db45af68b70d4aac58262/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Fabricator/Contact.php#L12-L12), you will need to install the hrcore extension to be able to use it.
 Another alternative is to refactor the method to support passing in any parameter to create a contact.
 I will go with using the fabricator.

hrui/tests/phpunit/HRUITrait.php
```diff
@@ -35,11 +37,27 @@ protected function createRelationship($contactA, $contactB, $relationship_type)
       'return' => array("id"),
```
tunbola: Please use single quotes here.

hrui/tests/phpunit/HRUITrait.php
```diff
@@ -35,11 +37,27 @@ protected function createRelationship($contactA, $contactB, $relationship_type)
...
       'name_a_b' => $relationship_type,
     ));
+
     civicrm_api3('Relationship', 'create', array(
```
tunbola: Please use the short array syntax here.

hrui/tests/phpunit/HRUITrait.php
mickadoo: I've changed it to use the fabricator

hrui/tests/phpunit/HRUITrait.php
mickadoo: Updated all code in this class to use short array syntax

hrui/tests/phpunit/CRM/HRUI/HelperTest.php
```diff
@@ -12,36 +14,57 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
...
   public function testGetLineManagersList() {
-    $contactParamsA = array("first_name" => "chrollo", "last_name" => "lucilfer");
-    $contactParamsB = array("first_name" => "hisoka", "last_name" => "morou");
-    $contactA = $this->createContact($contactParamsA);
-    $contactB = $this->createContact($contactParamsB)
-    ;
-    $this->createRelationship($contactA, $contactB, 'Line Manager is');
+    $relationshipType = 'Line Manager is';
+
+    $contactA = $this->createContact('chrollo', 'lucilfer');
```
tunbola: We should probably call the method this way:
`$this->createContact(['first_name'=>'chrollo', 'last_name' => 'lucilfer']);`
That way in case we need to fabricate with more parameters in the future we can do
`$this->createContact(['first_name'=>'chrollo', 'last_name' => 'lucilfer' 'nick_name' => 'BlaBla', 'is_deleted' => 0]);`

hrui/tests/phpunit/CRM/HRUI/HelperTest.php
```diff
@@ -12,36 +14,57 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
...
     $this->assertCount(2, $managers);
   }
 
+  /**
+   * @param string $firstName
+   * @param string $lastName
+   *
+   * @return int
+   */
+  private function createContact($firstName, $lastName) {
```
tunbola: This will just be `private function createContact($params) {`

hrui/tests/phpunit/CRM/HRUI/HelperTest.php
tunbola: Or we can ditch the method altogether and use the fabricator only

mickadoo: It's purely to keep things under 80 characters, the alternatives are removing the `createContact` method and using

```
   $contactA = ContactMaker::fabricate(['first_name' => 'chrollo', 'last_name' => 'lucilfer'])['id'];
```

Which is 103 chars with whitespace. Or:

```
    $params = ['first_name' => 'chrollo', 'last_name' => 'lucilfer'];
    $contactA = ContactMaker::fabricate($params)['id'];
```

Which is short, but takes two lines each time and the array structure is duplicated.

I agree with your point that it's less future proof but think that the method makes things more readable for now and it should be easy to change if we ever need to.

PR #1777, opened 2017-05-03
PCHR-2201: Manager Leave pagination Fix

deb1990:
## Overview
Manager was not navigated to the correct page by using First/Last/Previous/Next buttons

## Comments
This PR does not include BackstopJS tests as Manager Leave Report section is already covered.


deb1990:
@dhavalsharma Right now its not possible as we are not doing e2e tests also unit tests cannot cover this scenario.

dhavalsharma:
the question was on unit testing and as you have pointed to [unit test](https://github.com/civicrm/civihr/blob/a519212fc214dddc29476724e51e62911b760e75/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/manager-leave/components/manager-leave-requests_test.js#L398), its fine

PR #1780, opened 2017-05-03
PCHR-2212: Bump version to 1.7.0-beta1

davialexandre:


uk.co.compucorp.civicrm.appraisals/info.xml
```diff
@@ -8,8 +8,8 @@
     <author>coldrunKacper</author>
     <email>kacper@coldrun.pl</email>
```
tunbola: Do we need to standardise/unify the maintainer author and email across the CIVIHR extensions?

davialexandre: Maybe in the future, but not at the moment

PR #1781, opened 2017-05-03
PCHR-2200: Fix Uncaught Exception When Attempting to Create an Absence Period With Same Title

tunbola:
## Overview
When attempting to create/update an absence period with same title as an already existing absence period, an error is displayed and the site crashes.

## Before
The title column of the absence period table is a unique field and this is what causes a exception of type `PEAR_Exception`to be thrown. Since the Absence Period form is expecting an error of type `InvalidAbsencePeriodException` to be thrown, the exception does not get to be handled.

localhost-8900-civicrm-admin-leaveandabsences-periods 2017-05-03 16-51-27
## After

A validation function to check for the existence of an absence period with same title as the one being created/updated was added and an exception of type `InvalidAbsencePeriodException` is thrown if such absence type is found. The exception is caught and the error message displayed appropriately.

absence periods - civihr 1 6 demo 2017-05-03 16-52-08

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -67,6 +67,8 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
         'This Absence Period overlaps with another existing Period'
       );
     }
+
+    self::validateAbsencePeriodTitle($params);
   }
 
   /**
@@ -99,6 +101,34 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
     if($startDate >= $endDate) {
       throw new InvalidAbsencePeriodException(
         'Start Date should be less than End Date'
+      );
+    }
+  }
+
+  /**
+   * Checks if another absence period exists with same title as
+   * the absence period being created/updated and throws an exception if found.
+   *
+   * @param array $params
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsencePeriodException
+   */
+  private static function validateAbsencePeriodTitle($params) {
+    $title = !empty($params['title']) ? $params['title'] : '';
+    $absencePeriod = new self();
+    $absencePeriod->title = $title;
+    $absencePeriod->find(true);
+
+    if (!$absencePeriod->id) {
+      return;
+    }
+
+    $throwExceptionOnCreate = empty($params['id']);
+    $throwExceptionOnUpdate = !empty($params['id']) && $absencePeriod->id != $params['id'];
+
+    if ($throwExceptionOnCreate || $throwExceptionOnUpdate) {
+      throw new InvalidAbsencePeriodException(
+        'Absence Period with same title already exists!'
       );
     }
   }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -111,10 +111,10 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
    *
    * @param array $params
    *
-   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsencePeriodException
+   * @throws InvalidAbsencePeriodException
    */
   private static function validateAbsencePeriodTitle($params) {
-    $title = !empty($params['title']) ? $params['title'] : '';
+    $title = CRM_Utils_Array::value('title', $params);
     $absencePeriod = new self();
     $absencePeriod->title = $title;
     $absencePeriod->find(true);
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -104,6 +106,34 @@ private static function validateDates($params) {
   }
 
   /**
+   * Checks if another absence period exists with same title as
+   * the absence period being created/updated and throws an exception if found.
+   *
+   * @param array $params
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsencePeriodException
```
mickadoo: It might be nice to add an alias for this very long exception name just using the last part `InvalidAbsencePeriodException`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -104,6 +106,34 @@ private static function validateDates($params) {
...
+   */
+  private static function validateAbsencePeriodTitle($params) {
+    $title = !empty($params['title']) ? $params['title'] : '';
```
mickadoo: Personal preference, but I quite like `CRM_Utils_Array::value()` for things like this (the third parameter is a default)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -104,6 +106,34 @@ private static function validateDates($params) {
...
+    $absencePeriod = new self();
```
mickadoo: Interesting method to check if it exists, just curious why you chose this instead of an api call.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -104,6 +106,34 @@ private static function validateDates($params) {
...
+    $absencePeriod->title = $title;
+    $absencePeriod->find(true);
+
+    if (!$absencePeriod->id) {
+      return;
+    }
+
+    $throwExceptionOnCreate = empty($params['id']);
```
mickadoo: `$throwExceptionOnCreate` sounds like it will decide whether to throw an exception or not, but really once we get past the previous `if` statement we will be sure to throw an exception every time for creation of a new entity. How about using `$isCreate` for this, and for the other one `$isSeparateUpdate`. I think this would make the following `if` block more understandable.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -104,6 +106,34 @@ private static function validateDates($params) {
...
+    $throwExceptionOnUpdate = !empty($params['id']) && $absencePeriod->id != $params['id'];
+
+    if ($throwExceptionOnCreate || $throwExceptionOnUpdate) {
+      throw new InvalidAbsencePeriodException(
+        'Absence Period with same title already exists!'
```
mickadoo: It might be nice for the user to see the name of the duplicate title.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
```diff
@@ -72,8 +72,8 @@ public function testIfWeightIsEmptyItWillBeMaxWeightPlusOne()
   }
 
   /**
-   * @expectedException PEAR_Exception
-   * @expectedExceptionMessage DB Error: already exists
+   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidAbsencePeriodException
```
mickadoo: alias again

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
```diff
@@ -88,6 +88,25 @@ public function testPeriodsTitlesShouldBeUnique() {
     ]);
   }
 
+  public function testNoExceptionIsThrownWhenUpdatingAnAbsencePeriodWithSameTitle() {
```
mickadoo: nice to see new tests :grinning:

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
tunbola: This will be faster than the API. Will eliminate a lot of checks that the API has to make before the actual query to the db.
Besides, API will make more sense if the data to be retrieved is in another extension but since its within the same extension.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
tunbola: I agree but looking at the other exceptions we throw in L&A, we don't usually do this, so I'll just leave it like this to keep it consistent.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
tunbola: The alias wasn't working on the tests. I've not tried it before anyway. 
Seems phpunit needs the fully qualified class name.

PR #1782, opened 2017-05-03
PCHR-1928: Remove 'Application' CaseType

mickadoo:
## Overview
The Application CaseType is used by the HR Recruitment extension, but no clients are using it now and it conflicts with our use of CaseTypes as "Assignments". As a result you can see it in the "Assignment Edit" screen and add it as an assignment to someone, but most of the tasks are not selectable.

## Before
"Application" appears in both the "Assignment Edit" page (/civicrm/a/#/caseType) and it appears in the dropdown when you choose "Add Assignment" from the home page.

## After
"Application" does not appear in either the "Assignment Edit" page, nor is it available in the list of possible assignment types to assign to someone.

## Technical Details
Previously we were overriding the CaseType DAO. This was because the `add()` method calls a method to reconcile managed entities. Since the Assignment CaseType was a managed entity it would always be reset, no matter what you chose on the edit page. This fix was short-sighted since clearing the cache would reset the entity anyway. 

In this PR the override is removed, which also has the benefit that we don't have to keep it up-to-date with changes in the core file.

---

- [ ] Tests Pass

There is only one Selinium test in this extension directory and Davi mentioned that these tests should be redone at a later point.


hrrecruitment/CRM/HRRecruitment/Upgrader.php
```diff
@@ -116,4 +116,47 @@ public function upgrade_1400() {
     CRM_Core_DAO::executeQuery("UPDATE civicrm_case_type SET weight = 7 WHERE name = 'Application'");
     return TRUE;
   }
```
mickadoo: I don't know why, but github seems to think this code is an error :smiley_cat:

tunbola: A lot of commented code in this file. Do we still need them?

hrrecruitment/CRM/HRRecruitment/Upgrader.php
```diff
@@ -116,4 +116,47 @@ public function upgrade_1400() {
...
+      3 => ['CaseType', 'String']
+    ];
+    CRM_Core_DAO_Managed::executeQuery($query, $params);
+
+    $result = civicrm_api3('CaseType', 'get', [
+      'name' => 'Application',
+    ]);
+
+    // Application type doesn't exist, so our job is done
+    if ($result['count'] < 1) {
```
tunbola: Do we also need to be sure that Assignment of type 'Application' does not exist before we do a return here?
What do you think?

hrrecruitment/CRM/HRRecruitment/Upgrader.php
mickadoo: good point, I'll remove it since it's 3 years old

hrrecruitment/CRM/HRRecruitment/Upgrader.php
mickadoo: I was thinking about this, but it seems like assignments are deleted when you delete the parent CaseType. The code after this to fetch the assignments won't work if type "Application" CaseType doesn't exist.

hrrecruitment/CRM/HRRecruitment/Upgrader.php
```diff
@@ -5,115 +5,57 @@
...
-    CRM_Core_DAO::executeQuery('UPDATE foo SET bar = "whiz"');
-    CRM_Core_DAO::executeQuery('DELETE FROM bang WHERE willy = wonka(2)');
-    return TRUE;
-  } // */
-
+   * Remove the "Application" assignment type and the managed entity.
+   *
+   * @return bool
+   */
+  public function upgrade_1402() {
```
omarabuhussein: since the managed entity handler as you said is working fine now and clean everything , is this upgrader needed ?

mickadoo: ah, maybe I wasn't clear. I understood that we would need to execute SQL before upgrading because of the bug you found. It was always my plan to have the upgrader to remove the entities, but I just wanted to check if we needed to execute SQL or not.

omarabuhussein: yeah but still what is the value of that if the managed entity handler can already handle that  ?

mickadoo: @omarabuhussein after checking, the manager entity handler does not actually handle the deletion well, and will thrown an exception about foreign keys as you mentioned in your [pull request](https://github.com/civicrm/civicrm-core/pull/9896/files).

However the upgrader doesn't call the method to clear the cache; `CRM_Admin_Page_ExtensionsUpgrade::run` just handles the upgrades. It's just the API call that will clear the cache and cause and exception.

So I need to know the order of how things are done. There are two situations where I think I will need to prepare a pre-release script to delete the assignments:

- If we clear the cache _before_ running the upgrader
- If we run the upgrader using the API call

In these cases we can remove the upgrader because the Manager Entities handler will do the deletion work for us.

But if the API method is not used, and the cache is not cleared before the upgraders are run then we can use the upgrader.

PR #1783, opened 2017-05-04
PCHR-2197: CiviHR Emails, Foundation for Emails framework setup

AkA84:
## Overview
This PR contains the first proper setup of the Foundation for Emails framework in the "CiviHR Emails" extension

### Removed unused files
All the example files and .gitkeep files have been removed

### Remove HTML minification
In the gulpfile provided out of the box, when running `npm run build` (used to create the production-ready email templates), the whole email html was minified, in order to keep the email size at a minimum. Unfortunately this would create errors due to lines being randomly broken when CiviCRM sends the emails (maybe due to Smarty template processing?). In order to avoid these errors, the minification have been removed

### Use Shoreditch values
Any HTML email should follow the style of the [Shoreditch theme](https://github.com/civicrm/org.civicrm.shoreditch), and as such `civicrm-root` was added as a dependency so we can import the theme's variables and use them in the `_settings.scss` partial, where we have all the global variables defined for things like colors, paddings, typography etc
```scss
// app.scss

@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/overrides/variables";

@import 'settings';
// ...
```
```scss
// settings.scss

// 0. Shoreditch values
// ---------

$primary-color:   $brand-primary;
$secondary-color: $gray-light;
$success-color:   $brand-success;
// ...
$global-font-size: $font-size-base;
$h1-font-size:     $font-size-h1;
// ....
$global-font-color:  $text-color;
$body-font-family:   $font-family-base;
$global-line-height: $line-height-base;
// ...
```

### Base partials
A new scss partials have been introduces, `_base.scss`, where we can put all the styling for the generic tags like `<p>`, `<hr>`, `<h1>`, etc. For now it only contains the styling for `<hr>`
```scss
hr {
  border: 0;
  border-top: 1px solid $crm-dark-grayblue;
  height: 0;
  margin: $hr-margin;
  Margin: $hr-margin;
}
```

### Callout-header partials
Foundation for Emails provides a "callout" element, which is basically a panel. Unfortunately though the panel doesn't have any header (it's basically just a box where you put content in it). So a custom partial called "callout-header" have been added. The partial accepts a `title` parameter to set the title to show in the header

callout-with-header

```html
{{> callout-header title="This is a callout title"}}
<callout>
  lorem ipsum dolor sit amet
</callout>
```

### Components
The callout header is also defined as an html component, styled in the related `_callout-header.scss` partial, contained in the new `scss/components/` folder.

Another component is `email`, which is supposed to be used to define the style of the email as a whole. Currently the component has the style for the heading via the `email-heading` element

email-header

```html
<center class="email">
  <!-- header -->
  <container class="email-heading">
    <row class="collapse">
      <columns small="12" large="6">
        <h1 class="email-title">CiviHR Leave Request</h1>
      </columns>
      <columns small="12" large="6">
        <p class="email-date small-text-left text-right">Thu 04 May 2017</p>
      </columns>
    </row>
  </container>
  <!-- body -->
</center>
```

### Production-ready templates in the repo
The out of the box `.gitignore` file was ignore the whole `dist/` folder, but we need to store in the repo also the production-ready templates, so the following was added to `.gitignore`
```
dist/*
!dist/*.html
```

### `margin` mixin
In outlook the `margin` property needs to be used in the uppercase version, `Margin`. See here https://www.emailonacid.com/blog/article/email-development/outlook.com-does-support-margins for more details.

To avoid duplication, a `margin` mixin has been added
```scss
.selector1 {
  @include margin(10px);
}

.selector2 {
  @include margin(20px, bottom);
}
```
```css
.selector1 {
  margin: 10px;
  Margin: 10px;
}

.selector2 {
  margin-bottom: 20px;
  Margin-bottom: 20px;
}
```


uk.co.compucorp.civicrm.hremails/email-templates/src/assets/scss/components/_email.scss
```diff
@@ -0,0 +1,28 @@
+$email-heading-line-height: 40px;
+
+.email {
+  margin: 40px 0;
```
deb1990: Add `Margin`

uk.co.compucorp.civicrm.hremails/email-templates/src/assets/scss/app.scss
```diff
@@ -1,4 +1,7 @@
+@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/overrides/variables";
```
AkA84: no, in bootstrap sass the overrides come first, because of the usage of the `!default` property (https://robots.thoughtbot.com/sass-default)

uk.co.compucorp.civicrm.hremails/email-templates/src/assets/scss/components/_email.scss
AkA84: done

uk.co.compucorp.civicrm.hremails/email-templates/src/assets/scss/_base.scss
```diff
@@ -0,0 +1,7 @@
+hr {
+  border: 0;
+  border-top: 1px solid $crm-dark-grayblue;
+  height: 0;
+  margin: $hr-margin;
+  Margin: $hr-margin;
```
AkA84: done

uk.co.compucorp.civicrm.hremails/email-templates/src/assets/scss/components/_email.scss
deb1990: To be honest adding multiple properties is not a good solution, can we create a mixin which would do this for us?

PR #1786, opened 2017-05-04
PCHR-2167: Allow manager to edit all in sickness requests

dhavalsharma:
## Overview
Allow manager to edit all in sickness requests

## After
manager-actions

## Comments
Backstopjs test case is added to display sickness request by manager

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
@@ -5,13 +5,13 @@
       <div class="col-xs-8">
         <div class="btn-group btn-group-justified">
           <label
-            ng-disabled="$ctrl.isMode('view') || $ctrl.isRole('manager')"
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
```diff
@@ -5,13 +5,13 @@
       <div class="col-xs-8">
         <div class="btn-group btn-group-justified">
           <label
-            ng-disabled="$ctrl.isMode('view') || $ctrl.isRole('manager')"
+            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isLeaveType('sickness'))"
```
deb1990: According to current condition, all the items will be disabled for manager, if the leave type is not `sickness`. But the everything should be enabled for a manager regardless of the leave type?

dhavalsharma: ps refer to PCHR-1101 and it clearly specifies that only for sickness even dates are enabled otherwise not

PR #1789, opened 2017-05-04
PCHR-2209: Fix bug with requesting TOIL when current Entitlement Balance is Zero

tunbola:
## Overview
When trying to accrue TOIL for an absence type whose current balance is Zero for that period, It returns an error that the Leave Request Balance change is greater that the remaining balance for the absence type. This error is actually valid for a leave request especially if the absence type does not allow overuse but employee should be able to accrue TOIL even if current balance is Zero provided the other validations for accruing TOIL passes.

This PR also fixes an issue whereby TOIL can not be accrued when there is no working day for the TOIL request. Since an employee may want to accrue TOIL for days that are non-working days.

## Before
Trying to accrue TOIL when the current balance change for the absence type for the period is Zero returns an error.

Trying to accrue TOIL when the TOIL request has no working day returns an error.

## After
TOIL can be accrued even when the current balance for the absence type for the period is Zero provided all the other TOIL validations pass.

TOIL can be accrued even when the TOIL request has no working day provided all the other TOIL validations pass.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -80,8 +80,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     self::validateAbsenceTypeAllowRequestCancellationForLeaveRequestCancellation($params);
     self::validateAbsencePeriod($params);
     self::validateNoOverlappingLeaveRequests($params);
-    self::validateBalanceChange($params);
     self::validateWorkingDay($params);
+    self::validateBalanceChange($params);
     self::validateLeaveDatesDoesNotOverlapContracts($params);
   }
 
@@ -267,7 +267,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    */
   private static function validateTOILToAccruedAmountIsValid($params) {
     $absenceType = AbsenceType::findById($params['type_id']);
-    $unlimitedAccrual = empty($absenceType->max_leave_accrual) && $absenceType->max_leave_accrual != 0;
+    $unlimitedAccrual = empty($absenceType->max_leave_accrual) && $absenceType->max_leave_accrual !== 0;
 
     $periodContainingToilDates = AbsencePeriod::getPeriodContainingDates(
       new DateTime($params['from_date']),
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -471,8 +476,13 @@ private static function calculateBalanceChangeFromCreateParams($params) {
...
    */
   private static function validateWorkingDay($params) {
+    //TOIL accrual is independent of Working days.
+    if($params['request_type'] == self::REQUEST_TYPE_TOIL) {
+      return;
+    }
+
     $leaveRequestBalance = self::calculateBalanceChangeFromCreateParams($params);
-    if ($leaveRequestBalance == 0) {
+    if ($leaveRequestBalance <= 0) {
```
davialexandre: Please add a test to this new condition

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -267,7 +267,7 @@ private static function validateTOILPastDays($params) {
    */
   private static function validateTOILToAccruedAmountIsValid($params) {
     $absenceType = AbsenceType::findById($params['type_id']);
-    $unlimitedAccrual = empty($absenceType->max_leave_accrual) && $absenceType->max_leave_accrual != 0;
+    $unlimitedAccrual = empty($absenceType->max_leave_accrual) && $absenceType->max_leave_accrual !== 0;
```
tunbola: I tried requesting a TOIL when the max_leave_accrual was NULL and  could not. Discovered that the variable `$unlimitedAccrual` has a value of false. That was because NULL was type casted to Zero, So I had o change the comparison operator.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2151,4 +2151,99 @@ public function testEmailGetsSentWhenLeaveRequestIsCreatedAndUpdated() {
...
+      'from_date' => CRM_Utils_Date::processDate('today'),
+      'from_date_type' => 1,
+      'to_date' => CRM_Utils_Date::processDate('tomorrow'),
+      'to_date_type' => 1,
+      'toil_to_accrue' => 3,
+      'toil_duration' => 120,
+      'request_type' => LeaveRequest::REQUEST_TYPE_TOIL
+    ]);
+
+    $this->assertInstanceOf(LeaveRequest::class, $leaveRequest);
```
davialexandre: Having a second look at this, I think that a better way to confirm that the LeaveRequest was created is to check that the ID is not null or even checking getBalance() again to see if it will include the accrued days.

Also applies to the other test

tunbola: I've changed it to check the that the ID of the request is Not Null. 
To check for the balance change, that requires switching to use the leave request service to create the request and not sure we actually need to do that?

PR #1790, opened 2017-05-05
PCHR-1421: Error handling when comment throws error related to timezone

deb1990:
## Overview
If a comment is created at a time which is before an already existing comment(this could occur due to timezone), an error is thrown from backend but frontend did not show that error to the user. The timezone issue needs to be fixed from backend, but error handling for the same has been added so that the user knows about comments not being saved.

## Before
Bug was reported if a comment is created at a time which is before an already existing comment(this could occur due to timezone), an error is thrown from backend but frontend did not show that error to the user.

## After

All error messages will now be shown in a new line.
manager_leave___hr16-4

Total Workflow - Staff creating Leaves

Total Workflow - Manager Responding and Staff Reviewing
1421-manager

## Technical Details

1. The logic of how the `role`was getting set is changed, as when we are in Manager Leave Page, it automatically means that the role is `manager` so removed calling `this.request.roleOf` function which was anyway not working for a new request as the `leave_id`is not generated yet and the same is required for the `this.request.roleOf` function. 
2. `$rootScope.role = '<manager/owner/admin>'`is added to the `manager-leave-container`, `my-leave-container`, `absence-tab-container` which is passed by the `leave-request-popup` directive to the `request-ctrl` which is how the role of the user is decided.
3. Also there was an error in the `LeaveRequest.isManagedBy` function as for new LeaveRequest there is no `leave_id`until you save, which we never noticed but after implementing error handling we got frozen UI, that got fixed by removing the call altogether and using role sent from markup to determine the role of the user.
4. Removed `LeaveRequestInstance.roleOf`and `LeaveRequestAPI.isManagedBy` as both of them is not required anymore. 
5. The `error` property of `request-ctrl` has been changed to an array to support display of multiple errors.

## Comments

Error handling and loading all records has been implemented in Base API service in https://github.com/civicrm/civihr/pull/1824

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -121,9 +122,13 @@ define([
...
-
-          return response.data;
+          if(!!response.data.is_error) {
+            deferred.reject(response.data.error_message);
+          }
+          deferred.resolve(response.data);
         });
+
+        return deferred.promise;
       },
```
dhavalsharma: missing catch when $http calls itself fails

deb1990: It is not done intentionally, because the catch will be caught by the actual controller/service(example: `RequestCtrl`). The error automatically gets caught by the first `catch` in the promise chain.

dhavalsharma: you are not returning $http, you are only resolving/rejecting in then call, so how will clients of the call know when $http itself throws error

dhavalsharma:
also noticed in request ctrl 

```
      function handleError(errors) {
        // show errors
        if (errors.error_message)
          this.error = errors.error_message;
        else {
          if (_.isObject(errors)) {
            this.error = _.map(errors, function (value, key) {
              var errorMessage = _.isArray(value) ? value.join(', ') : JSON.stringify(value);

              return errorMessage + ' for key ' + key;
            }).join(', ');
          } else {
            this.error = errors.toString();
          }
        }

        //reset loading Checks
        this.loading.showBalanceChange = false;
        this.loading.absenceTypes = false;
        this.loading.fromDayTypes = false;
        this.loading.toDayTypes = false;
      }
```

but you are returning error string instead of error object? sometimes the errors can be collection of string in error object

deb1990:
> you are not returning $http, you are only resolving/rejecting in then call, so how will clients of the call know when $http itself throws error

I dont think the client needs to know if the error is thrown by `$http` itself or it is thrown by the code. At the end the client only cares about the error, so it might be from $http itself or it might be the error thrown by the code using `deferred.reject(response.data.error_message);`

org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -108,8 +108,9 @@ define([
        */
       sendGET: function (entity, action, params, cache) {
         $log.debug('api.sendGET');
+        var deferred = $q.defer();
 
-        return $http({
+        $http({
```
AkA84: I'm not 100% sure, but i think this should be equivalent
```js
sendGET: function (entity, action, params, cache) {
  $log.debug('api.sendGET');

  return $http({
    // ...
  })
  .then(function (response) {
    if(!!response.data.is_error) {
      return $q.reject(response.data.error_message);
    }

    return response.data;
  });
},
```

org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -121,9 +122,13 @@ define([
             action: action
           }
         }).then(function (response) {
-
-          return response.data;
+          if(!!response.data.is_error) {
```
AkA84: the whole `then` function is the same in both `sendGET` and `sendPOST`, meaning you can put it into a separate function and just pass that function to then
```js
function commonHandler(response) {
  // ..
}

// ...

$http.get({
  // ...
})
.then(commonHandler);

Comment (AkA84) [inline org.civicrm.reqangular/src/common/services/api.js]: also, mind the white space

File: org.civicrm.reqangular/src/common/services/api.js (hunk 1 above)
Comment (AkA84) [inline org.civicrm.reqangular/src/common/services/api.js]: i think @dhavalsharma 's correct, that's why you should return `$http` (see my previous comments)

File: org.civicrm.reqangular/test/services/api_test.js [hunk 4]
```diff
@@ -0,0 +1,122 @@
+define([
+  'common/angular',
+  'common/angularMocks',
+  'common/services/api'
+], function () {
+  'use strict';
+
+  describe('api test', function () {
```
AkA84: the first describe of a test is always the name of what you are testing

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    describe('sendGET', function () {
+      var entity = 'send_get_entity',
+        action = 'send_get_action',
+        params = {key: 'value'},
+        cache = true,
+        promise;
+
+      afterEach(function () {
+        $httpBackend.flush();
```
AkA84: can you probably do this outside of any sub-`describe` given that you need to do it for all of the tests anyway

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+      api = _api_;
+      $rootScope = _$rootScope_;
+      $httpBackend = _$httpBackend_;
+    }));
+
+    describe('sendGET', function () {
+      var entity = 'send_get_entity',
+        action = 'send_get_action',
+        params = {key: 'value'},
+        cache = true,
```
AkA84: you could have a single function encapsulating all the call parameters and the `whenGET` call, would make for a more organized code
```js
function expectAndSendGET(returnValue) {
  var action = 'get', entity = 'DummyEntity';

  $httpBackend
    .whenGET(new RegExp('action=' + action + '&entity=' + entity))
    .respond(returnValue);

  return api.sendGET(entity, action, { filter_name: 'filter_value' }, true);
}
```
```js
beforeEach(function () {
  promise = expectAndSendGET({ /* .. */ });
});
```

also please stick with the real naming convention of entities and actions

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+        promise;
+
+      afterEach(function () {
+        $httpBackend.flush();
+      });
+
+      describe('when there is no error', function () {
```
AkA84: maybe it's better "when the API doesn't return an error"

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+        };
+
+        beforeEach(function () {
+          $httpBackend.whenGET(/action\=send_get_action&entity\=send_get_entity/)
+            .respond(returnValue);
+
+          promise = api.sendGET(entity, action, params, cache);
+        });
+
+        it('proper value is returned', function () {
```
AkA84: `it` titles should always try to make a full sentence starting with "it": "it returns the proper value"

AkA84: anyway this title is not clear. What is a "proper" value? are we implying that there is an "improper" value? aren't you simply saying that when there is no error the api service returns the data as is?

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+        promise;
+
+      afterEach(function () {
+        $httpBackend.flush();
+      });
+
+      describe('when there is no error', function () {
+        var returnValue = {
+          is_error: 0,
+          somekey: 'someval'
```
AkA84: `sendGET` and sendPOST` always return `response.data` when the response is good, where is the `.data` property here?

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+        });
+
+        it('proper value is returned', function () {
+          promise.then(function (response) {
+            expect(response).toEqual(returnValue);
+          });
+        });
+      });
+
+      describe('when there is error', function () {
```
AkA84: "when the API returns an error"

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -0,0 +1,122 @@
...
+        };
+
+        beforeEach(function () {
+          $httpBackend.whenGET(/action\=send_get_action&entity\=send_get_entity/)
+            .respond(returnValue);
+
+          promise = api.sendGET(entity, action, params, cache);
+        });
+
+        it('error message is thrown', function () {
```
AkA84: "it rejects the promise with the error message provided by the API"

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -67,11 +57,7 @@ define([
 
         this.sendGET('LeaveRequest', 'getbalancechangebyabsencetype', params, false)
           .then(function (data) {
-            if (checkError(data)) {
-              deferred.reject(data.error_message);
-            } else {
-              deferred.resolve(data.values);
-            }
+            deferred.resolve(data.values);
```
AkA84: you don't even need `deferred` anymore, you can simply return `data.values`
```js
this.sendGET()
  .then(function (data) {
    return data.values;
  });
```

AkA84: applies for all other instances

org.civicrm.reqangular/test/services/api_test.js
deb1990: This `data` is different, it gets added by the ajax handler like
```
{data: <response from api>, url:'', headers:''}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
deb1990: This wont work as we are still returning `deferred.promise`, so we need to resolve.

org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -9,6 +9,20 @@ define([
 
     var API_ENDPOINT = '/civicrm/ajax/rest';
 
+    /**
+     * Common handler for API calls
+     *
+     * @param {object} response - response from backend
+     * @return {Promise/Any}
+     */
+    function commonHandler(response) {
```
AkA84: `commonHandler` was just a placeholder name that i was using in my example....it should go without saying that you should give it a descriptive name

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -16,107 +16,107 @@ define([
...
 
-        it('error message is thrown', function () {
+        it('rejects the promise with the error message provided by the API', function () {
           promise.catch(function (response) {
             expect(response).toBe(returnValue.error_message);
           });
         });
       });
+
+      function expectAndSendGET(returnValue) {
```
AkA84: always document every function

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -16,107 +16,107 @@ define([
...
+        var action = 'get', entity = 'DummyEntity';
+
+        $httpBackend
+          .whenGET(new RegExp('action=' + action + '&entity=' + entity))
+          .respond(returnValue);
+
+        return api.sendGET(entity, action, {filter_name: 'filter_value'}, true);
```
AkA84:
```js
{ filter_name: 'filter_value' }
```
always leave a white space between the brackets (applies everywhere)

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -16,107 +16,107 @@ define([
...
-        params = {key: 'value'},
-        promise;
-
-      afterEach(function () {
-        $httpBackend.flush();
-      });
+      var promise;
 
-      describe('when there is no error', function () {
+      describe('when the API doesnt return an error', function () {
```
AkA84: doesn't

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -16,107 +16,107 @@ define([
...
+        it('rejects the promise with the error message provided by the API', function () {
           promise.catch(function (response) {
             expect(response).toBe(returnValue.error_message);
           });
         });
       });
+
+      function expectAndSendPOST(returnValue) {
+        var entity = 'send_post_entity',
+          action = 'send_post_action';
```
AkA84: there is no need for any of these variables

AkA84: also, as already said you need to stick with the naming convention for both entities and actions in civicrm

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -16,107 +16,107 @@ define([
...
+
+        $httpBackend
+          .whenPOST('/civicrm/ajax/rest')
+          .respond(returnValue);
```
AkA84: You can save lines here
```js
$httpBackend.whenPOST('/civicrm/ajax/rest').respond(returnValue);
```
the line is still short and readable.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -157,7 +157,11 @@ define([
         this.sendPOST('LeaveRequest', 'isValid', params)
           .then(function (data) {
             if (data.count > 0) {
-              deferred.reject(data.values);
+              var errorMessage = _.map(data.values, function (value) {
+                return _.isArray(value) ? value.join('</br>') : JSON.stringify(value);
```
AkA84: you are doing something related to the presentation layer (putting every message in a new line), which something that *should not* happen on the api service, as it shouldn't care how the messages will be displayed on the page. Leave the job to the controller, or even better to the template

AkA84:
@deb1990 there is a lot of empty space between the first comment and the textarea, why is that?

deb1990:
> @deb1990 there is a lot of empty space between the first comment and the textarea, why is that?
here is the fix https://github.com/compucorp/civihr-employee-portal-theme/pull/159

org.civicrm.reqangular/test/services/api_test.js
```diff
@@ -108,14 +108,9 @@ define([
...
-        var entity = 'send_post_entity',
-          action = 'send_post_action';
-
-        $httpBackend
-          .whenPOST('/civicrm/ajax/rest')
-          .respond(returnValue);
+        $httpBackend.whenPOST('/civicrm/ajax/rest').respond(returnValue);
 
-        return api.sendPOST(entity, action);
+        return api.sendPOST('send_post_entity', 'send_post_action');
```
AkA84: be consistent, you use these names in expectAndSendGET
```js
var action = 'get', entity = 'DummyEntity';
```
so do the same here. Even better, just use the same variables for both expectAndSendGET and expectAndSendPOST, no?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -157,11 +157,15 @@ define([
...
-              deferred.reject(errorMessage);
+              var errors = [];
+              _.map(data.values, function (value) {
+                if(_.isArray(value)) {
+                  errors = errors.concat(value);
+                } else {
+                  errors.push(JSON.stringify(value));
+                }
+              });
+              deferred.reject(errors);
```
AkA84:
```js
deferred.reject(_(data.values).map().flatten().value());
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -322,12 +322,17 @@ define([
           });
 
           it('returns validation errors as string', function () {
```
AkA84: does it?

AkA84: also, somehow mention that it rejects the promise

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -322,12 +322,17 @@ define([
...
-            var errorResponse = _.map(mockData.getNotIsValid().values, function (value) {
-              return _.isArray(value) ? value.join('</br>') : JSON.stringify(value);
-            }).join('</br>');
+            var errors = [];
+            _.map(mockData.getNotIsValid().values, function (value) {
```
AkA84: i think this whole thing can be simplified, you simply expect to reject the promise with a flatten array of the mocked errors

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -353,8 +353,9 @@ <h4 class="modal-title">
     uib-alert
     type="danger"
     close="$ctrl.closeAlert()"
-    ng-bind-html="$ctrl.error"
     dismiss-on-timeout="5000">
+    <div ng-if="$ctrl.isArray($ctrl.error)" ng-repeat="errMsg in $ctrl.error track by $index">{{errMsg}}</div>
+    <div ng-if="!$ctrl.isArray($ctrl.error)">{{$ctrl.error}}</div>
```
AkA84: the errors will always be an array, you don't need the `isArray` on the scope

davialexandre:
@deb1990 Please fix the PR description:
- Instead of writing "Bug -" in the overview section, add the "bug" tag to the PR
- Please add the Before section. It doesn't make sense to have an After section without first explaining how things worked before
- I don't think the "Other changes" sections is really necessary. Item `#1` there is actually a technical detail and `#2` can probably be added to the After section

Also, the description altogether is a bit confusing and it's hard to understand what is being changed. Did you fix the timezone problem only? Did you added error handling for comments only? Did you change how we handle errors for everything?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -27,14 +27,14 @@ define([
       var absenceTypesAndIds,
         initialLeaveRequestAttributes = {}, //used to compare the change in leaverequest in edit mode
         mode = '', //can be edit, create, view
-        role = '', //could be manager, owner or admin
+        role = $rootScope.isManager ? 'manager' : 'owner', //could be manager, owner or admin
```
dhavalsharma: how will it set for admin?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
```diff
@@ -5,13 +5,13 @@
       <div class="col-xs-8">
         <div class="btn-group btn-group-justified">
           <label
-            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isLeaveType('sickness'))"
+            ng-disabled="$ctrl.isMode('view') || (!$ctrl.isNewRequest() && !$ctrl.isLeaveType('sickness'))"
```
dhavalsharma: why is manager check removed from here? and not sure why `(!$ctrl.isNewRequest()` test is added across file here?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/toil.html
```diff
@@ -5,7 +5,7 @@
       <div class="col-xs-4 chr_leave-request-modal__input-shift-left">
         <div class="chr_custom-select chr_custom-select--full">
           <select name="hours" ng-model="$ctrl.request.toilDurationHours"
-            ng-disabled="$ctrl.isMode('view') || $ctrl.isRole('manager')"
+            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isNewRequest())"
```
dhavalsharma: why is `(!$ctrl.isNewRequest()` added here?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-container.js
```diff
@@ -9,12 +9,14 @@ define([
...
     }],
     controllerAs: 'managerLeave',
-    controller: ['$log', function ($log) {
+    controller: ['$log', '$rootScope', function ($log, $rootScope) {
       console.log('CRM id of the currently logged in user: ', this.contactId);
       $log.debug('Component: manager-leave-container');
 
       var vm = Object.create(this);
 
+      $rootScope.isManager = true;
```
deb1990: When we have a way to know how admin can be recognised, we can set another variable like this, and using that same variable request-ctrl can decide, which role it is.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: When we have a way to know how admin can be recognised, we can set another variable like this, and using that same variable request-ctrl can decide, which role it is.

Also added `initUserRole` where the code can be placed in future.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/toil.html
deb1990: Its not required anymore, using `isMode('create')` instead

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
deb1990: Manager check was not returning proper values before in case of a new request. So removed the same and modified the logic to keep the disabled state as before.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-container.js
deb1990: Its implemented now, check description of PR please.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-container.js
```diff
@@ -9,9 +9,11 @@ define([
       return settings.pathTpl + 'components/my-leave-container.html';
     }],
     controllerAs: 'myleave',
-    controller: ['$log', '$scope', '$uibModal', 'settings', function ($log, $scope, $modal, settings) {
+    controller: ['$log', '$rootScope', '$scope', '$uibModal', 'settings', function ($log, $rootScope, $scope, $modal, settings) {
       $log.debug('Component: my-leave-container');
 
+      $rootScope.role = 'owner';
```
AkA84: At this point i think it makes more sense to use "staff" instead of "owner"

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -406,13 +406,14 @@ define([
        * @return {Object} attributes
        */
       this._initRequestAttributes = function () {
+        initUserRole.call(this);
```
AkA84: This should not be there, look at what this method is doing. This method is for the leave request attributes, not for the settings of the popup.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -968,15 +942,21 @@ define([
       }
 
       /**
+       * Initialize user's role to either owner, manager or admin
+       */
+      function initUserRole () {
+        role = this.directiveOptions.userRole;
```
AkA84: Why do you even need this at all? Either you do it at the top of the controller (does it make sense to initialize `role` to an empty string?) or you do it in the `_.init()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/toil-request-ctrl.js
```diff
@@ -46,8 +46,8 @@ define([
        */
       vm.calculateToilExpiryDate = function () {
         if (!vm.request.from_date) {
-          vm.error = 'Please select from date to find expiry date';
-          return $q.reject(vm.error);
+          vm.errors = ['Please select from date to find expiry date'];
+          return $q.reject(vm.errors);
```
AkA84: add an empty line before it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/directives/leave-request-popup.js
```diff
@@ -34,12 +34,18 @@ define([
...
           leaveType: '@',
-          forStaff: '<'
+          userRole: '@'
         },
         restrict: 'EA',
         link: function (scope, element) {
           var controller = _.capitalize(getLeaveType(scope.leaveType, scope.leaveRequest)) + 'RequestCtrl';
 
+          if (!scope.userRole) {
+            $log.error("User role not specified.");
```
AkA84: this is pointless, either you throw an error or you fallback to a default. Given that you are falling back to a default, you can simply do

```js
directiveOptions: function () {
  return {
    userRole: scope.userRole || 'staff';
  }
});

Comment (AkA84) [inline uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/directives/leave-request-popup.js]: although i think it would make more sense to set the default fallback in the controller, instead of relying on the directive to give the controller the default

File: uk.co.compucorp.civicrm.hrleaveandabsences/views/manager-leave/components/manager-leave-container.html [hunk 34]
```diff
@@ -1,16 +1,28 @@
...
         <span class="caret"></span>
       </button>
       <ul class="dropdown-menu" uib-dropdown-menu>
-        <li><a leave-request-popup contact-id="managerLeave.contactId" leave-type="sickness" for-staff="true" href="#">
-          Record Sickness</a></li>
-        <li><a leave-request-popup contact-id="managerLeave.contactId" leave-type="toil" for-staff="true" href="#">
-          Record Overtime</a></li>
+        <li>
+          <a leave-request-popup
+             user-role="{{$root.role}}"
```
AkA84: it's not 2-spaces

uk.co.compucorp.civicrm.hrleaveandabsences/views/manager-leave/components/manager-leave-container.html
```diff
@@ -1,16 +1,28 @@
...
+             contact-id="managerLeave.contactId"
+             leave-type="sickness"
+             href="#">Record Sickness
+          </a>
+        </li>
+        <li>
+          <a leave-request-popup
+             user-role="{{$root.role}}"
```
AkA84: it's not 2-spaces

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-container.html
```diff
@@ -1,13 +1,31 @@
 <div class="my-leave-page">
   <div class="button-container">
     <div class="btn-group" uib-dropdown>
-      <button type="button" class="btn btn-primary" leave-request-popup leave-type="leave" contact-id="myleave.contactId">Record New Absence</button>
+      <button type="button" class="btn btn-primary" leave-request-popup
+              leave-type="leave"
```
AkA84: it's not 2-spaces

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-container.html
```diff
@@ -1,13 +1,31 @@
...
+      </button>
       <button type="button" class="btn btn-primary dropdown-toggle-split" uib-dropdown-toggle>
         <span class="caret"></span>
       </button>
       <ul class="dropdown-menu" uib-dropdown-menu>
-        <li><a leave-request-popup contact-id="myleave.contactId" leave-type="sickness" href="#">Record Sickness</a></li>
-        <li><a leave-request-popup contact-id="myleave.contactId" leave-type="toil" href="#">Record Overtime</a></li>
+        <li>
+          <a leave-request-popup
+             user-role="{{$root.role}}"
```
AkA84: it's not 2-spaces

uk.co.compucorp.civicrm.hrleaveandabsences/views/my-leave/components/my-leave-container.html
```diff
@@ -1,13 +1,31 @@
...
+             contact-id="myleave.contactId"
+             leave-type="sickness"
+             href="#">Record Sickness
+          </a>
+        </li>
+        <li>
+          <a leave-request-popup
+             user-role="{{$root.role}}"
```
AkA84: it's not 2-spaces

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -11,14 +11,14 @@ <h4 class="modal-title">
 </div>
 <crm-loading show="!$ctrl.loading.absenceTypes">
   <div
-    ng-show="$ctrl.isRole('manager')"
+    ng-show="$ctrl.isRole('manager') && !$ctrl.isMode('create')"
```
AkA84: why this change? why do you need to show this bit only when the role is manager *and* when we are creating a new leave request?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
```diff
@@ -5,13 +5,13 @@
       <div class="col-xs-8">
         <div class="btn-group btn-group-justified">
           <label
-            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isLeaveType('sickness'))"
+            ng-disabled="$ctrl.isMode('view') || (!$ctrl.isMode('create') && !$ctrl.isLeaveType('sickness'))"
```
AkA84: before it was:
"disable if we're in view mode, or if the user is a manager and we are not looking at a sickness leave request"

now it's:
"disable if we're in view mode, or if the mode is not create and we are not looking at a sickness leave request"

two questions:
1. Why the change from "is a manager" to "it's not create mode"?
2. Why the focus on the sickness type?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
```diff
@@ -37,11 +37,11 @@
             ng-model="$ctrl.uiOptions.fromDate"
             is-open="$ctrl.uiOptions.date.from.show"
             ng-readonly="true"
-            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isLeaveType('sickness'))"
+            ng-disabled="$ctrl.isMode('view') || ($ctrl.isRole('manager') && !$ctrl.isMode('create') && !$ctrl.isLeaveType('sickness'))"
```
AkA84: I need you to explain also this change in logic

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: I tried that, but `this`does not have child directives values when role is defined. 
`directiveOptions`is set after calling parent controller. That is why moved the same to a common place which is called by the child, i.e. `_initRequestAttributes `

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Please refer the reply below

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/directives/leave-request-popup.js
deb1990: I have kept the error so the developers can know if they have missed specifying the role. I can change it to `warn`?

deb1990: I think the opposite, because `userRole`is the directives property, so it his job to decide what the default value should be.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
deb1990: This is the block which displays the users Name and Status of the Request, and it should only be visible when It a manager and he is reviewing an old request. Previously the `$ctrl.isRole('manager')` was only true when the user is manager but not in create mode. As the `$ctrl.isRole('manager')` is fixed now, the `!$ctrl.isMode('create')` is necessary to be added.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
deb1990: This is the button which display the Multiple Days button, and it should only be disabled when the user is not in `create` mode regardless of the `role`. 
Previously the $ctrl.isRole('manager') was only true when the user is manager but not in create mode. As the $ctrl.isRole('manager') is fixed now, the !$ctrl.isMode('create') is necessary to be added otherwise the button will be disabled for Create mode for manager which we dont want.

Also Manager can edit everything if its a sickness request.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/request-modal/details/leave.html
deb1990: The reason is specified above.

PR #1791, opened 2017-05-05
PCHR-2189: Load only approved and waiting_approval requests

deb1990:
## Overview
Leave Request is not shown in the Calendar if overlaps the cancelled  request

## After

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/calendar-ctrl.js
```diff
@@ -266,7 +259,7 @@ define([
     };
 
     /**
-     * Loads all the leave requests and calls calendar load function
+     * Loads the approved and waiting approval leave requests and calls calendar load function
```
dhavalsharma: ps add admin_approved here as well

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/calendar-ctrl.js
```diff
@@ -358,6 +352,19 @@ define([
...
+     * Returns leave status value from name
+     * @param {String} name - name of the leave status
+     * @returns {int/boolean}
+     */
+    function getLeaveStatusValuefromName(name) {
+      var leaveStatus = _.find(leaveRequestStatuses, function (status) {
+        return status.name === name;
+      });
+
+      return leaveStatus ? leaveStatus.value : false;
```
dhavalsharma: why return 2 different data types?

deb1990: When `leaveStatus` is not present, something needs to be returned, it will be either false/undefined/null. Returned false so that I can use ternary operator instead of an if condition.

PR #1801, opened 2017-05-08
PCHR-2195: Fix Issues With Toil To Accrue Amount Not Being Saved Correctly.

tunbola:
## Overview
When creating a TOIL accrual request either as a staff or as a manager on behalf of the staff, If the TOIL to be accrued amount selected is a float e.g 2.5, It does not get saved correctly and the value is not displayed when trying to edit the request after it has been created

## Before
The Toil to be accrued amount does not get saved correctly when a floating number value is to be accrued.

The reason for this is that the column for toil_to_accrue in the Leave Request table is integer type and some of the toil_to_accrue values are float e.g 0.5 for Half day. When these values are saved, they get rounded up to integers in the database i.e 0.5 becomes 1 and 1.5 becomes 2.

## After
This PR fixes that by changing the toil_to_accrue column to VARCHAR. The obvious solution would have been to change to change the column type to decimal but since its linked to the hrleaveandabsences_toil_amounts group, the column has to be VARCHAR to be same as the option value column type storing the toil_to_accrue amounts.

The Toil to be accrued amount gets saved correctly irrespective of the amount to be accrued.

An extra validation is also added to ensure that the toil to be accrued amount should be numeric even if the option value for the toil amount passed in is valid.

## Comments
The Civicrm version for L&A has been bumped to 4.7.18, While regenerating the DAO for Leave Request BAO, Civi also regenerated the code for all the DAO's in L&A extension probably due to a change in the code for generating the DAOs in the newer version.

Manual tests were done after an uninstall and reinstall of the extension to make sure everything works fine as before. 

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1778,6 +1778,67 @@ public function testLeaveRequestCanNotBeCreatedWhenRequestTypeIsToilButToilToAcc
...
+      'toil_to_accrue' => "4 days",
+      'request_type' => LeaveRequest::REQUEST_TYPE_TOIL
+    ]);
+  }
+
+  /**
+   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
+   * @expectedExceptionMessage The TOIL to accrue amount is not valid
+   */
+  public function testLeaveRequestCanNotBeCreatedWhenRequestTypeIsToilAndToilToAccrueIsAValidAmountButNotNumeric() {
```
mickadoo: Wow, long name :smile: 

How about just something a little shorter like

`testValidNonNumericToilRequest`

And if it needs more information you could put it in the method description. It's just a bit of a headache trying to read such long camel case method names.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -221,8 +221,8 @@ private static function validateTOILRequest($params) {
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
    */
   private static function validateTOILToAccrueIsAValidOptionValue($params) {
-    $toilAmountOptions = self::buildOptions('toil_to_accrue', 'validate');
-    if(!array_key_exists($params['toil_to_accrue'], $toilAmountOptions) || !is_numeric($params['toil_to_accrue'])) {
+    $toilAmountOptions = array_flip(self::buildOptions('toil_to_accrue', 'validate'));
+    if(!in_array($params['toil_to_accrue'], $toilAmountOptions) || !is_numeric($params['toil_to_accrue'])) {
```
mickadoo: nice solution :+1:

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: I get. For L&A the test names are usually very descriptive(and long sometimes). So that you can read the method name at a glance and know what is being tested without bothering to read the description.

PR #1806, opened 2017-05-09
PCHR-1687: Fix Failing Tests

MiyaNoctem:
## Overview
Some tests of hrjobcontract and hrui extension were failing due to the changes introduced in PCHR-1687.

## Before
Tests for hrjobcontract and hrui extensions were failing, since they relied on values that were removed in upgrades created for PCHR-1687.  Also, some warnings were being thrown by upgrades, when trying to access non-existing array keys.  Upgrade 1000 of hrcore extension was also failing on tests.

## After
 Fixed by using constructors to create missing values for tests, implementing checks prior to access array values and updating tests setup to install required extensions.  Fixed upgrade 1000 of hrcore installation by including the name of the file being downloaded.

---

- [x] Tests Pass

### DIFF
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -345,8 +345,10 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
       $entitiesToRemove[$caseType] = 'CaseType';
     }
 
-    foreach ($activityTypes as $activityType) {
-      $entitiesToRemove['civitask:act:' . $activityType] = 'OptionValue';
+    foreach ($activityTypes as $extension) {
+      foreach ($extension as $activityType) {
+        $entitiesToRemove['civitask:act:' . $activityType] = 'OptionValue';
+      }
     }
 
     foreach ($entitiesToRemove as $name => $entity) {
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -157,19 +157,18 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
 
     CRM_Core_DAO::executeQuery("DROP TABLE IF EXISTS civicrm_hrhours_location");
     CRM_Core_DAO::executeQuery("
-        CREATE TABLE IF NOT EXISTS `civicrm_hrhours_location` (
-          `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-          `location` varchar(63) DEFAULT NULL,
-          `standard_hours` int(4) DEFAULT NULL,
-          `periodicity` varchar(63) DEFAULT NULL,
-          `is_active` tinyint(4) DEFAULT '1',
-          PRIMARY KEY(id)
-        ) ENGINE=InnoDB  DEFAULT CHARSET=utf8 AUTO_INCREMENT=1
-      ");
-      CRM_Core_DAO::executeQuery("
-        INSERT INTO `civicrm_hrhours_location` (`id`, `location`, `standard_hours`, `periodicity`, `is_active`) 
-        VALUES (1, 'Head office', 40, 'Week', 1)
-      ");
+      CREATE TABLE IF NOT EXISTS `civicrm_hrhours_location` (
+        `id` int(10) unsigned NOT NULL,
+        `location` varchar(63) DEFAULT NULL,
+        `standard_hours` int(4) DEFAULT NULL,
+        `periodicity` varchar(63) DEFAULT NULL,
+        `is_active` tinyint(4) DEFAULT '1'
+      ) ENGINE=InnoDB  DEFAULT CHARSET=utf8 AUTO_INCREMENT=1
+    ");
+    CRM_Core_DAO::executeQuery("
+      INSERT INTO `civicrm_hrhours_location` (`id`, `location`, `standard_hours`, `periodicity`, `is_active`) 
+      VALUES (1, 'Head office', 40, 'Week', 1)
+    ");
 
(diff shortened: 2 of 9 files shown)


davialexandre:
@MiyaNoctem I believe this is still in progress, right? Can you please describe this somewhere in the PR description and/or title?

Also, the overview section still has part of the template text. Please remove it

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -345,8 +345,10 @@ private function up1402_removedUnusedManagedEntities($caseTypes, $activityTypes)
       $entitiesToRemove[$caseType] = 'CaseType';
     }
 
-    foreach ($activityTypes as $activityType) {
-      $entitiesToRemove['civitask:act:' . $activityType] = 'OptionValue';
+    foreach ($activityTypes as $extension) {
```
mickadoo: I think this is probably my fault from changing the structure of the default activity types.

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -345,8 +345,10 @@ private function up1402_removedUnusedManagedEntities($caseTypes, $activityTypes)
...
+      foreach ($extension as $activityType) {
+        $entitiesToRemove['civitask:act:' . $activityType] = 'OptionValue';
+      }
```
mickadoo: Just a question, since HRCase doesn't seem to have any tests in it did you find this bug from another test failing?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -158,11 +158,12 @@ public function upgradeBundle() {
     CRM_Core_DAO::executeQuery("DROP TABLE IF EXISTS civicrm_hrhours_location");
     CRM_Core_DAO::executeQuery("
         CREATE TABLE IF NOT EXISTS `civicrm_hrhours_location` (
-        `id` int(10) unsigned NOT NULL,
+          `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
```
mickadoo: If we're adding this to an existing upgrader won't it be a problem for installations where this upgrader has already been run?

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -47,6 +53,22 @@ function setUp() {
     ];
   }
 
+  function createPayScale() {
```
mickadoo: I think this is a nice solution to create the data that's required, but shouldn't it be removed in a tearDown method too?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
```diff
@@ -38,7 +38,7 @@ private function up1000_downloadUKEnglish() {
       mkdir($downloadPath, 0755, true);
     }
 
-    file_put_contents($downloadPath, fopen($localizationURL, 'r'));
+    file_put_contents($downloadPath . 'civicrm.mo', fopen($localizationURL, 'r'));
```
mickadoo: Same question, was this broken and if it was, then maybe it could be added to the PR description. Also same question about editing an upgrader when it may have already been run.

hrcase/CRM/HRCase/Upgrader.php
MiyaNoctem: Yes, that's right, test fot hrui extension requires installation of hrcase.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
MiyaNoctem: You're right!  What I found is it's impossible to add records to this table, so I don't think it's ever worked.  Good point.

MiyaNoctem: Ok, no one seems to know about this bug, so I'll fix it within this ticket.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
MiyaNoctem: As I understand, we don't need teardown methods because the test database gets re-installed and upgraded at the beginning of each test, to start with a clean slate, so to speak.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1000.php
MiyaNoctem: This one is part of the upgrades done for this ticket, so I don't think it's a problem, since it hasn't been deployed to any live sites yet.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
mickadoo: That would be nice, I've heard different things on this since in another review I was asked to add a teardown method. Do you know if it's reset at the end of each test method, or is it by class?

MiyaNoctem: It's by class, but it actually happens at the start, on setup.  I've never been asked to do a teardown before, but maybe we should check with @davialexandre ?

mickadoo: From what Omar said in the chat, as long as it implements TransactionalInterface it should be reset after each test.

MiyaNoctem: Ok, so the real reason teardowns are not required is this class implements TransactionalInterface interface, which handles all clean-up at the end of the test.

PR #1812, opened 2017-05-10
PCHR-2207: Fix issues with Admin or Leave Manager not being able to create Leave Request for themselves

tunbola:
## Overview
Currently when an Admin or Leave manager tries to request leave, an error is returned that the leave request cannot be created. This is due to the fact that the Manager Status transition matrix is used when making the request and it does not allow a transition from 'No status' to 'Awaiting Approval' Status(The default status set for a new leave request).  [See Matrix Here](https://compucorp.atlassian.net/browse/PCHR-1825)

## Before
When an Admin or Leave Manager requests leave for themselves or on behalf of a Staff member, the Manager Status Matrix is used.

An Admin or Leave Manager cannot request leave for themselves.

## After
Changes were made such that when an Admin or Leave Manager requests  leave for themselves, the Staff Status transition matrix is used since at this point they are requesting leave as a staff. When they are making modifications to a Leave Request on behalf of a Staff Member, the Manager Status Matrix is used.

An Admin or Leave Manager can now request leave for themselves.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrix.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrix.php
@@ -151,9 +151,7 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequestStatusMatrix {
 
     if ($currentUserID == $leaveRequestContactID) {
       $statusMatrix = $this->getStaffStatusMatrix();
-    }
-
-    if ($this->shouldUseManagerMatrixForCurrentUser($leaveRequestContactID)) {
+    } elseif ($this->shouldUseManagerMatrixForCurrentUser($leaveRequestContactID)) {
       $statusMatrix = $this->getManagerStatusMatrix();
     }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
@@ -73,28 +73,28 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequestStatusMatrixTest extends BaseHe
   }
 
   public function testCanTransitionToForAdminForReturnsTrueAllPossibleTransitionStatuses() {
-    $manager = ContactFabricator::fabricate();
-    $leaveContact = ContactFabricator::fabricate();
-    $this->registerCurrentLoggedInContactInSession($manager['id']);
-    CRM_Core_Config::singleton()->userPermissionClass->permissions = ['administer leave and absences'];
+    $adminID = 5;
+    $leaveContact = 2;
+    $this->registerCurrentLoggedInContactInSession($adminID);
+    $this->setPermissions(['administer leave and absences']);
 
     $possibleTransitions = $this->allPossibleStatusTransitionForLeaveApprover();
 
     foreach($possibleTransitions as $transition) {
-      $this->assertTrue($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $leaveContact['id']));
+      $this->assertTrue($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $leaveContact));
     }
   }
 
   public function testCanTransitionToForAdminReturnsFalseForAllNonPossibleTransitionStatuses() {
-    $manager = ContactFabricator::fabricate();
-    $leaveContact = ContactFabricator::fabricate();
-    $this->registerCurrentLoggedInContactInSession($manager['id']);
-    CRM_Core_Config::singleton()->userPermissionClass->permissions = ['administer leave and absences'];
+    $adminID = 5;
+    $leaveContact = 2;
+    $this->registerCurrentLoggedInContactInSession($adminID);
+    $this->setPermissions(['administer leave and absences']);
 
     $nonPossibleTransitions = $this->allNonPossibleStatusTransitionForLeaveApprover();
 
     foreach($nonPossibleTransitions as $transition) {
-      $this->assertFalse($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $leaveContact['id']));
+      $this->assertFalse($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $leaveContact));
     }
   }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
@@ -98,6 +98,79 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequestStatusMatrixTest extends BaseHe
     }
   }
 
+  public function testCanTransitionToReturnsTrueForAllPossibleStaffTransitionStatusesWhenLeaveApproverIsTheLeaveContact() {
+    $manager = ContactFabricator::fabricate();
+    $leaveContact = ContactFabricator::fabricate();
+    $this->registerCurrentLoggedInContactInSession($manager['id']);
+    $this->setContactAsLeaveApproverOf($manager, $leaveContact);
+
+    $possibleTransitions = $this->allPossibleStatusTransitionForStaffDataProvider();
+
+    foreach($possibleTransitions as $transition) {
+      $this->assertTrue($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $manager['id']));
+    }
+  }
+
+  public function testCanTransitionToReturnsFalseForAllNonPossibleStaffTransitionStatusesWhenLeaveApproverIsTheLeaveContact() {
+    $manager = ContactFabricator::fabricate();
+    $leaveContact = ContactFabricator::fabricate();
+    $this->registerCurrentLoggedInContactInSession($manager['id']);
+    $this->setContactAsLeaveApproverOf($manager, $leaveContact);
+
+    $nonPossibleTransitions = $this->allNonPossibleStatusTransitionForStaffDataProvider();
+
+    foreach($nonPossibleTransitions as $transition) {
+      $this->assertFalse($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $manager['id']));
+    }
+  }
+
+  public function testCanTransitionToReturnsFalseForPossibleManagerExclusiveStatusTransitionsWhenLeaveApproverIsTheLeaveContact() {
+    $manager = ContactFabricator::fabricate();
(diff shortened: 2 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -185,4 +258,18 @@ public function allNonPossibleStatusTransitionForLeaveApprover() {
       ['', $leaveRequestStatuses['cancelled']],
     ];
   }
+
+  /**
+   * Return possible status transitions that is only exclusive to the Manager or Admin
+   *
+   * @return array
+   */
+  public function getManagerExclusivePossibleStatusTransitions() {
```
mickadoo: While it's fine like this, it might be worth looking into [data providers](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) for this kind of test in the future.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -73,28 +73,101 @@ public function testCanTransitionToForLeaveApproverReturnsFalseForAllNonPossible
   }
 
   public function testCanTransitionToForAdminForReturnsTrueAllPossibleTransitionStatuses() {
+    $adminID = 5;
```
mickadoo: It's not very clear where these values come from

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -73,28 +73,101 @@ public function testCanTransitionToForLeaveApproverReturnsFalseForAllNonPossible
...
+    $possibleTransitions = $this->allPossibleStatusTransitionForLeaveApprover();
+
+    foreach($possibleTransitions as $transition) {
+      $this->assertTrue($this->leaveRequestStatusMatrix->canTransitionTo($transition[0], $transition[1], $leaveContact));
+    }
+  }
+
+  public function testCanTransitionToForAdminReturnsFalseForAllNonPossibleTransitionStatuses() {
+    $adminID = 5;
+    $leaveContact = 2;
```
mickadoo: I see the fabricator is used in tests below - why not here too?

`$leaveContact = ContactFabricator::fabricate();`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -185,4 +258,18 @@ public function allNonPossibleStatusTransitionForLeaveApprover() {
...
+  /**
+   * Return possible status transitions that is only exclusive to the Manager or Admin
+   *
+   * @return array
+   */
+  public function getManagerExclusivePossibleStatusTransitions() {
+    $possibleStaffTransitions = $this->allPossibleStatusTransitionForStaffDataProvider();
+    $possibleManagerTransitions = $this->allPossibleStatusTransitionForLeaveApprover();
+
+    $results = array_diff(array_map('serialize', $possibleManagerTransitions), array_map('serialize', $possibleStaffTransitions));
```
mickadoo: That's one way to do it :smile: 

Would this not be possible with `array_udiff`?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -185,4 +258,18 @@ public function allNonPossibleStatusTransitionForLeaveApprover() {
...
+    $managerExclusiveStatusTransition = array_map('unserialize', $results);
+    return $managerExclusiveStatusTransition;
```
mickadoo: empty line before return

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrix.php
```diff
@@ -151,9 +151,7 @@ private function getStatusMatrixForCurrentUser($leaveRequestContactID) {
 
     if ($currentUserID == $leaveRequestContactID) {
       $statusMatrix = $this->getStaffStatusMatrix();
-    }
-
-    if ($this->shouldUseManagerMatrixForCurrentUser($leaveRequestContactID)) {
+    } elseif ($this->shouldUseManagerMatrixForCurrentUser($leaveRequestContactID)) {
```
mickadoo: just about the logic here, if `($currentUserID == $leaveRequestContactID)` the fix was to change this to else if because `shouldUseManagerMatrixForCurrentUser` would return true. But does it make sense to change the logic in `shouldUseManagerMatrixForCurrentUser` to return false in this case since that's what seems to be the root cause?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
tunbola: They are just random contact ID's. Normally the ID is supposed to be gotten from the contact fabricator but since Its not really needed for the contact to exist in DB before registering it in the session. This allows the test run time to be significantly faster.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
tunbola: Some of the tests in this class actually use data providers. The reason why I did not add it for the newer tests is because, some of them use contact fabricators and these contacts have to be fabricated each time the data provider supplies values to the test, this caused the tests to run slower.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
tunbola: array_diff still seems better though. I had to serialize because it can't handle more than a one-dimensional array.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrix.php
tunbola: The Admin when requesting for leave is a staff member as well as an admin. Using the elseif makes sure that once the logic detects that you are requesting leave for yourself, use the status matrix irrespective of any role you may have. This helps eliminate unnecessary checks in `shouldUseManagerMatrixForCurrentUser` because it also runs some queries to determine  if the user is a leave approver or admin. More like a short circuit.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
mickadoo: I can see how it might run faster but I think that tests should be responsible for creating and then destroying all data they use. If tests rely on data created by other tests it could lead to one change breaking lots of tests in other extensions.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
mickadoo: I see, sounds like there are some bigger issues with testing. Hopefully we'll get them running fast and clean someday

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
mickadoo: Why more suited? If you have to serialize it, it sounds like it might not be the best choice. I don't know of course, if there was some unique property or was to easily compare transitions then it might be cleaner and faster with `array_udiff`. If not then I guess serialization and immediate deserialization is an option, although it seems a bit wasteful.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
mickadoo: After taking to @tunbola it seems the tests do not rely on existing data. These values are random and only tests that actually require a contact to be created will use the fabricator.

tunbola: Well it won't because all that is isn't interfering with any other test. All it does is log in the contact ID supplied in session and we don't need an actual contact fabricated for that

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestStatusMatrixTest.php
```diff
@@ -266,9 +266,9 @@ public function getManagerExclusivePossibleStatusTransitionsDataProvider() {
...
     $possibleManagerTransitions = $this->allPossibleStatusTransitionForLeaveApprover();
 
-    $results = array_diff(array_map('serialize', $possibleManagerTransitions), array_map('serialize', $possibleStaffTransitions));
-    $managerExclusiveStatusTransition = array_map('unserialize', $results);
-
-    return $managerExclusiveStatusTransition;
+    $comparator = function ($a, $b) {
+      return strcmp(sprintf('%d-%d', $a[0], $a[1]), sprintf('%d-%d', $b[0], $b[1]));
+    };
+    return array_udiff($possibleManagerTransitions,$possibleStaffTransitions, $comparator);
```
mickadoo: You're probably going to hate me, but I did a check of the performance for a small array.

```
100000 runs:

serialization: 1.0214629173279
udiff: 3.2251260280609
```

Maybe for large, multidimensional arrays `udiff` could be faster, but if the arrays are all just arrays of 2 integers it seems like serialization has the edge.

PR #1822, opened 2017-05-12
PCHR-2164: Add 10 file as attachments,restrict file formats

dhavalsharma:
## Overview
Now user can add upto 10 files as attachments. The file upload is restricted to `txt,jpg,png,jpeg,bmp,gif,pdf,doc,docx,xls,xlsx,ppt,pptx` file formats.

## Before
User could only add upto 5 files as attachment to requests. 
User could upload any file type.

## After
User can add upto 10 files as attachment to requests. 
The file types that can be uploaded are restricted.

## Technical Details

1. Changed `queueLimit: 10` for fileuploader in leave request instance `uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js`.
2. Added `filters` during creation of `FileUploader` in `org.civicrm.reqangular/src/common/services/file-upload.js`
3. The file types are checked using mime types as its supported by the `FileUploader` library. It internally adds mime type metadata to each file. 

### Reason for not using file extension
Sometimes its possible to rename an unsupported file extension to supported file extension and upload it. To prevent such scenarios mime types are used to filter file types. 

### Future recommendation
Currently the number of files that can be uploaded is set to 10 but in future it may be required to be configured number as all clients may not be comfortable with a limitation of 10.
`queueLimit: 10,` in `uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js` needs to be configured sometimes in future.

- [x] Tests Pass


org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -65,9 +65,28 @@ define([
             formData: [{
               entity_table: customSettings.entityTable || error('entityTable'),
               crm_attachment_token: customSettings.crmAttachmentToken || error('crmAttachmentToken')
+            }],
+            filters: [{
+              name: 'fileFormatFilter',
+              fn: function (item /*{File|FileLikeObject}*/ , options) {
```
AkA84:
```js
function (item /*{File|FileLikeObject}*/ , options) {
```
what is this syntax?

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -65,9 +65,28 @@ define([
...
+                var type = '|' + item.type.slice(item.type.lastIndexOf('/') + 1) + '|';
+
+                //checks for mime types and not file name extensions
```
AkA84: why do you need to check for mime types? There is no mention of this in the technical details

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -65,9 +65,28 @@ define([
             formData: [{
               entity_table: customSettings.entityTable || error('entityTable'),
               crm_attachment_token: customSettings.crmAttachmentToken || error('crmAttachmentToken')
+            }],
+            filters: [{
+              name: 'fileFormatFilter',
+              fn: function (item /*{File|FileLikeObject}*/ , options) {
+                var type = '|' + item.type.slice(item.type.lastIndexOf('/') + 1) + '|';
```
AkA84: can you write here an example of a value of `item.type`?

AkA84: also why do you need to wrap the `type` in "|"s? Isn't it sufficient to use the type to look in the list? why are the "|"s necessary for the search?

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -65,9 +65,28 @@ define([
...
+
+                //checks for mime types and not file name extensions
+                //|txt|jpg|png|jpeg|bmp|gif|pdf|doc|docx|xls|xlsx|ppt|pptx|
+                return '|plain|png|jpeg|bmp|gif|pdf|msword|vnd.openxmlformats-officedocument.wordprocessingml.document|vnd.ms-excel|vnd.openxmlformats-officedocument.spreadsheetml.sheet|vnd.ms-powerpoint|vnd.openxmlformats-officedocument.presentationml.presentation|'.indexOf(type) !== -1;
```
AkA84: I don't think this is a scalable approach, you are basically imposing the list of allowed file types for any application that wants to use the file uploader service.

What instead should happen is that the application itself defines the list of file types that it can accept, and then pass the information to the file uploader

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -128,7 +128,7 @@ define([
             fileUploader: FileUpload.uploader({
               entityTable: 'civicrm_hrleaveandabsences_leave_request',
               crmAttachmentToken: sharedSettings.attachmentToken,
-              queueLimit: 5
+              queueLimit: 10 //TODO for now set the limit to 10 files
```
AkA84: why TODO? you are not explaining why this is supposed to be a temporary solution

AkA84:
@dhavalsharma also, this needs at least a new unit test

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: comment copy pasted from sample

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: changed name to mimeType

dhavalsharma: need to use a separator to search in a collection and dont think | is used in any of mime type

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: this is taken from file uploader's sample where it creates meta data for the files. 
where are the technical details?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: how else can it explain that it's a temp fix and currently being set to 10 files?

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: now the user can pass mime types to be filtered

dhavalsharma:
@AkA84 filters are part of library itself and do we need to write test case to test the library functions or do you need a test case in `LeaveRequestInstance` to check if file types are getting passed or not?

org.civicrm.reqangular/src/common/services/file-upload.js
AkA84: > where are the technical details?

The section in the PR template.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
AkA84: You are not explaining *why* it's a temporary fix. Temporary fix in place of what permanent solution?

org.civicrm.reqangular/src/common/services/file-upload.js
AkA84: if you have 
```js
var types = '|plain|png|jpeg|bmp|gif|pdf|msword|vnd.openxmlformats-officedocument.wordprocessingml.document|vnd.ms-excel|vnd.openxmlformats-officedocument.spreadsheetml.sheet|vnd.ms-powerpoint|vnd.openxmlformats-officedocument.presentationml.presentation';
```
why something like `types.indexOf('png')` shouldn't work?

Also, since you are using separators, why don't you do

```js
_.includes(types.split('|'), typeToFind);
```

AkA84:
@dhavalsharma if you are now passing the allowed types to the service, then that needs a unit test

org.civicrm.reqangular/src/common/services/file-upload.js
AkA84: that's not related to my comment, i wasn't commenting on the name of the variable. I was simply asking if you can give here an example of what `item.type` looks like

org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -53,6 +53,8 @@ define([
             return error('custom settings');
           }
 
+          customSettings.allowedMimeTypes = customSettings.allowedMimeTypes || '|plain|png|jpeg|bmp|gif|pdf|';
```
AkA84: write a private function where you set the default of `customSettings`, and you use `_.defaults`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -128,7 +128,9 @@ define([
             fileUploader: FileUpload.uploader({
               entityTable: 'civicrm_hrleaveandabsences_leave_request',
               crmAttachmentToken: sharedSettings.attachmentToken,
-              queueLimit: 5
+              queueLimit: 10, //TODO for now set the limit to 10 files
+              //set the mime types which are allowed to be separated by pipe | character
```
AkA84: the | character is expected by the library, i assume?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -128,7 +128,9 @@ define([
...
+              allowedMimeTypes: '|plain|png|jpeg|bmp|gif|pdf|msword|vnd.openxmlformats-officedocument.wordprocessingml.document|vnd.ms-excel|vnd.openxmlformats-officedocument.spreadsheetml.sheet|vnd.ms-powerpoint|vnd.openxmlformats-officedocument.presentationml.presentation|'
```
AkA84: I think this should go in the `config` module of the app, not buried in the leave request instance

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: type looks like any other mime type like `image/png`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: got it, improved comment

org.civicrm.reqangular/src/common/services/file-upload.js
dhavalsharma: indexOf alone will not give correct result because sometimes a mimetype may be part of another type (for eg. x-bzip and x-bzip2) and may not give correct index. 

The includes logic also will fail for above condition.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: no its useful in separating and setting boundary for mime type

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
dhavalsharma: moved it to `shared-settings`

org.civicrm.reqangular/test/services/file-upload_test.js
```diff
@@ -139,4 +139,55 @@ define([
       });
     });
   });
+
+  describe('FileUpload without mocked FileUploader service', function () {
```
AkA84: do you need any of this? you have a `FileUploaderMock` that you can extend by simply (i think, based on what i see you already did) adding
```js
return {
  // ..
  onCompleteItem: settings.onCompleteItem,
  onErrorItem: settings.onErrorItem,
  filters: settings.filters, // <-- added
};
```

this way you should be able to simply get the content of `uploader.filters` and run directly the `fileFormatFilter` function
```js
describe('allowed mime types', function () {
  var filterFn, allowedMimeTypes;

  beforeEach(function () {
    allowedMimeTypes = [ /* ... */ ];

    uploader = fileUpload.uploader({ allowedMimeTypes: allowedMimeTypes });
    filterFn = // find the `fileFormatFilter` function in `uploader.filters`
  });

  it('filters out files with not allowed mime types', function () {
    expect(filterFn(/* file allowed */)).toBe(true);
    expect(filterFn(/* file not allowed */)).toBe(false);
  });
});
```
so you don't need to use the real file service and you don't need to have two main branches in the test file

PR #1824, opened 2017-05-15
PCHR-1421: Error handling and load all records

deb1990:
## Overview
Error handling has been implemented at the Base API class, loading all records instead of default 25 is implemented in base api.

## Technical Details

1. Error handling has been implemented at the Base API class.
2. The base API checks if the response contains `is_error` property, if true the API rejects the promise with the `error_message` sent from Backend. 
3. Loading all records instead of default 25 is implemented in base api. But a explicit specification of `limit` parameter will have greater priority.
4. `$httpParamSerializer`is used in place of the previous `transformRequest`function, as it does the same thing.

## Comments
This is related to https://github.com/civicrm/civihr/pull/1790

---

- [x] Tests Pass


deb1990:
> how are promises getting resolved without digest being fired?

```
afterEach(function () {
  $httpBackend.flush();
});

PR #1829, opened 2017-05-15
PCHR-2199: Sickness Request email template

AkA84:
## Overview
This PR adds the Sickness Request email template to the CiviHR Email extension

screenshot

## Technical Details
The leave request email template files have been renamed to include the `.civicrm.` suffix, see https://github.com/civicrm/civihr/pull/1836

---

The TOIL request email template had the `msg_html` property twice in the .mgd file. Fixed that.

---

Some of handlebar templates now use the `{{{ }}}` syntax instead of `{{ }}`
```html
<span>{{{value}}}</span>
```
because some of the param values can have characters that can be escaped by Handlebars, for example, see the `value` here:
```html
{{> request-data value="{$leaveRequest->toil_to_accrue} {if $leaveRequest->toil_to_accrue > 1}days{else}day{/if}"
```
it would be escaped like this
```html
{$leaveRequest-&gt;toil_to_accrue} {if $leaveRequest-&gt;toil_to_accrue &gt; 1}days{else}day{/if}
```
Using `{{{}}}` makes Handlebars skip the character escaping

---

The screenshot doesn't show the date of the request. This is because of a backend bug, which is likely going to be addressed in https://github.com/civicrm/civihr/pull/1838


uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
```diff
@@ -0,0 +1,64 @@
...
+    <columns>
+      {{> callout-header title="Leave Request Type"}}
+      <callout class="request-data">
+        {{> request-data key="Status" value="{$leaveStatus}"}}
+        {{> request-data key="Staff Member" value="{contact.display_name}"}}
+
+        {if $leaveRequest->from_date eq $leaveRequest->to_date}
+          {{> request-data key="Date" value="{$fromDate|truncate:10:''|crmDate} {$fromDateType}"}}
+        {else}
+          {{> request-data key="From Date" value="{$fromDate|truncate:10:''|crmDate} {$fromDateType}"}}
```
dhavalsharma: why are dates truncated to 10? can 10 come from a variable which can be changed at a single place?

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
```diff
@@ -0,0 +1,64 @@
...
+          {{> request-data key="To Date" value="{$toDate|truncate:10:''|crmDate} {$toDateType}"}}
+        {/if}
+        <hr>
+        {{> request-data key="Additional Details" value=""}}
```
dhavalsharma: is `Additional Details` a label as value is always empty?

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
```diff
@@ -0,0 +1,64 @@
...
+      </callout>
+      <spacer></spacer>
+      <button href="{$leaveRequestLink}" class="expanded alert">View This Request</button>
+      {if $leaveComments}
+        <hr>
+        {{> callout-header title="Request comments"}}
+        <callout class="callout-no-padding">
+          <div class="request-comments">
+            {foreach from=$leaveComments item=value key=label}
+              {{> request-comment commenter="{$value.commenter}" createdAt="{$value.created_at|crmDate}" text="{$value.text}"}}
```
dhavalsharma: why is date not truncated to 10 here?

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
AkA84: this is taken from @tunbola 's raw template. @tunbola could reply to this?

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
AkA84: @tunbola ?

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
AkA84: this was done simply for layout purposes and to reuse what we have instead of adding some exception to the style. Basically additional details has the same look of any label in a label/value couple, so i reused the request-data partial and kept the value empty

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
tunbola: Well I saw this code a couple of places in L&A. Basically crmDate function formats a date in this form `2017-04-21 22:17:54` , the truncate removes the time part and leaves only the date. The from_date column is a DATE type.

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/sickness-request.civicrm.html
tunbola: The reason why is that the created_at is a DATETIME column and one will be interested in the date and also the time the comment is made.

PR #1839, opened 2017-05-17
PCHR-2225: Prevent multiple save clicks

dhavalsharma:
## Overview
Prevent user from submitting again while a request is being saved after clicking submit.

## Before
A user was allowed to submit while the leave request is being saved. 

## After
A user is not allowed to submit while the leave request is being saved. A spinner is shown instead of the save button.

manager-actions

## Technical Details
A flag `didSubmit` was added to observe when user submits requests. It is reset after request is successfully submitted or returns and error.

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -34,6 +34,7 @@ define([
       this.absenceTypes = [];
       this.calendar = {};
       this.contactName = null;
+      this.didSubmit = false;
```
AkA84: The name of this property is misleading, as by this name you would expect to always stay `true` after the user has clicked "submit" (instead, it can go back to `false` if some error happens)

Probably a better name is `submitting`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -629,6 +629,16 @@
         });
 
         describe('save leave request', function () {
+          describe('allows multiple save', function() {
```
AkA84: we are doing the exact opposite, we are *not* allowing multiple save*s*

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -629,6 +629,16 @@
...
+            beforeEach(function() {
+              $ctrl.submit();
+            });
+
+            it('does not allow user to submit again', function () {
+              expect($ctrl.didSubmit).toBeTruthy();
```
AkA84: This does not test whether the controller tries to save the leave request again, it simply tests that when i click "submit" this flag gets set to `true`. You should test instead that when that happen, "submit" doesn't save the request

AkA84: actually i just realized that the reason that you are not testing for that ("when the flag is `true`, it doesn't save the request again") is because you are not doing it at all. You are simply relying on the view to hide the buttons so that the user can't click on them.

Please add also a logic to the controller that actually prevents multiple saves when the flag is `true`

dhavalsharma: even though the button is not visible its still disabled
`ng-disabled="!$ctrl.canSubmit() || $ctrl.didSubmit"` from html

dhavalsharma: even though the button is not visible its still disabled
ng-disabled="!$ctrl.canSubmit() || $ctrl.didSubmit" from html

dhavalsharma: even though the button is not visible its still disabled
ng-disabled="!$ctrl.canSubmit() || $ctrl.didSubmit" from html

AkA84: Delegating a data-integrity check mechanism to the view it's not a good idea. Use the flag on the view to show/hide the buttons, but put the actual multiple-saves-preventing mechanism in the controller

PR #1844, opened 2017-05-17
PCHR-2164: Prevent more than 10 file uploads

dhavalsharma:
## Overview
Now user is allowed to upload upto 10 files in edit mode as well as in create mode.

## Before
User could not upload more than 10 files in create mode but could do so in edit mode.

## After
Now user is allowed to upload upto 10 files in edit mode as well as in create mode.
image

## Technical Details
Added check to see if the sum of already uploaded files and those which are in queue is less than allowed limit of files that can be uploaded.
Also updated UI to give user feedback after 10 files add.

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -333,6 +333,16 @@ define([
...
           return $q.resolve();
+        },
+
+        /**
+         * Checks if user can upload more file, it totals the number of already
+         * uploaded files and those which are in queue and compares it to limit.
+         *
+         * @return {Boolean} true is user can upload more else false
+         */
+        canUploadMore: function () {
```
AkA84: I don't think this logic should be the responsibility of the instance, rather it should be the controller's. All that the instance knows and should know is that it can own some attachments, but it's the whole app that , knowing what the limit is on the max allowed files (in the settings module), should enforce said limit

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -184,6 +184,16 @@ define([
       };
 
       /**
+       * Checks if user can upload more file, it totals the number of already
+       * uploaded files and those which are in queue and compares it to limit.
+       *
+       * @return {Boolean} true is user can upload more else false
+       */
+      this.canUploadMore = function () {
```
AkA84: all good, but please keep the methods in alphabetical order, first `this.canSubmit, then `this.canUploadMore`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -184,6 +184,16 @@ define([
...
+        return this.request.files.length + this.request.fileUploader.queue.length < sharedSettings.fileUploader.queueLimit;
```
AkA84: also, please use parenthesis to make the code more readable
```js
return (this.request.files.length + this.request.fileUploader.queue.length) < sharedSettings.fileUploader.queueLimit;
```

PR #1851, opened 2017-05-18
PCHR-2206: Fix contact selection dropdown styles

dhavalsharma:
## Overview
Manager can now select contact with properly styles dropdown.

## Before
The styles of contact were messed up after migration to 1.7.  It looked like below before.

captura de tela de 2017-05-02 08-01-04

## After
This is how it looks with the changes.
eidt-comment

## Technical Details

1.  Removed theme `civihr-ui-select` from dropdown.
2. After deselection of contact id was set to undefined resulting in failure of some APIs, therefore added a check to see if contact id is set or not and then call chain of APIs.

## Comments
The CSS was also updated as per this [PR](https://github.com/compucorp/civihr-employee-portal-theme/pull/160).

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -509,7 +509,7 @@ define([
           .then(function (usersCalendar) {
             self.calendar = usersCalendar;
           });
-      };
+      }
```
AkA84: the semicolon was correct, and actually it should be present in every
```js
this.method = function () {

};

**Comment by AkA84** [comment]
@dhavalsharma i believe this was a regression issue, so if it was please point out where (and why) the issue was introduced

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js [hunk 2]
```diff
@@ -633,6 +633,11 @@ define([
       this.initAfterContactSelection = function () {
         var self = this;
 
+        //when manager deselects contact it is called without a selected contact_id
+        if (!self.request.contact_id) {
```
AkA84: you already have, just some lines above
```js
if (self.request.contact_id) {
  return self.initAfterContactSelection();
}
```
why then is this check needed?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -23,7 +23,6 @@ <h4 class="modal-title">
         autofocus
         ng-model="$ctrl.request.contact_id"
         ng-change="$ctrl.initAfterContactSelection()"
-        theme="civihr-ui-select"
```
AkA84: why does removing the theme works? If the theme was specified, and if before the dropdown looked correct, doesn't it mean the theme was necessary?

AkA84:
> PCHR-2206: Fix contact selection dropdown styles	

The name of the commit has nothing to do with its content

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
dhavalsharma: this is again called after manager selects contact

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
dhavalsharma: good questions, tried to to git bisect and changed to earlier version of civicrm, could not locate the cause of regression yet

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
AkA84: ok so do you need it here
```js
.then(function () {
   if (self.request.contact_id) {
     return self.initAfterContactSelection();
   }
 });
```
there is duplication here that should be avoided

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -633,6 +633,11 @@ define([
...
+          return $q.reject("The contact id or period id was not set");
```
AkA84: Why is the error message mentioning the period id if the check is on `self.request.contact_id`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -633,6 +633,11 @@ define([
       this.initAfterContactSelection = function () {
         var self = this;
 
+        //when manager deselects contact it is called without a selected contact_id
```
AkA84: leave a space after `//`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -1234,6 +1235,18 @@
...
+
+        describe('after contact is deselected', function() {
+          beforeEach(function() {
+            $ctrl.request.contact_id = undefined;
+            $ctrl.initAfterContactSelection();
+            $scope.$digest();
+          });
+
+          it('does not call calendar APIs', function () {
+            expect(WorkPatternAPI.getCalendar).not.toHaveBeenCalled();
```
AkA84: shouldn't you test that it throws an error?

PR #1862, opened 2017-05-19
PCHR-1687: Catch Exception When Creating PK for Hours Location Table

MiyaNoctem:
## Overview
An upgrade for hrjobcontract extension was throwing an exception when being ran on some environments, where the particular fix it was trying to apply was already implemented.

## Before
On some installations, civicrm_hrhours_location table was being created without defining the id field as a primary key and not setting it as autonumeric, which made it all but impossible to add new records to it, via the front-end of civihr. To solve this problem, an upgrader was created to add the constraints to the table. However, staging environment, among others, already had the fix, which caused a DB error when the upgrader tried to add the primary key again.

## After
Fixed by catching the exception, and throwing it only if primary key and autonumeric has not been set for the id field in the table.

---

- [x] Tests Pass

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1275,7 +1275,7 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
   /**
    * Makes Hour Location id field autonumeric and adds id as a primary key.
    */
-  public function upgrade_2017051901() {
+  public function upgrade_1030() {
     try {
       CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` ADD PRIMARY KEY (`id`)');
       CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` CHANGE `id` `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT');
@@ -1289,7 +1289,7 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
       }
     }
 
-    return false;
+    return true;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1276,8 +1276,18 @@ public function upgrade_1029() {
...
-    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` ADD PRIMARY KEY (`id`)');
-    CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` CHANGE `id` `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT');
+    try {
+      CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` ADD PRIMARY KEY (`id`)');
+      CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` CHANGE `id` `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT');
+    } catch (Exception $e) {
+      $tableStructure = CRM_Core_DAO::executeQuery('DESC civicrm_hrhours_location');
+
+      while ($tableStructure->fetch()) {
+        if ($tableStructure->Field == 'id' && ($tableStructure->Key != 'PRI' || stripos($tableStructure->Extra, 'auto_increment') === false)) {
```
davialexandre: can't you use the exception message to tell what was the cause of the error instead of inspecting the table description?

If no, then please make this more readable. Assign the whole condition to a variable that describes what is being checked and use this variable inside the if

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1279,13 +1279,9 @@ public function upgrade_1030() {
...
       CRM_Core_DAO::executeQuery('ALTER TABLE `civicrm_hrhours_location` CHANGE `id` `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT');
-    } catch (Exception $e) {
-      $tableStructure = CRM_Core_DAO::executeQuery('DESC civicrm_hrhours_location');
-
-      while ($tableStructure->fetch()) {
-        if ($tableStructure->Field == 'id' && ($tableStructure->Key != 'PRI' || stripos($tableStructure->Extra, 'auto_increment') === false)) {
-          throw $e;
-        }
+    } catch (PEAR_Exception $e) {
+      if (stripos($e->getCause()->userinfo, 'nativecode=1068') === false) {
```
davialexandre: This looks better, but the readability is still bad. No one knows what nativecode 1068 means. Assigning the `stripos` result to a variable with a descriptive name will make it easier for other devs to understand what you're trying to check here.

Also, I don't understand the reason for throwing a new exception with a new message. Is there any reason for that?

MiyaNoctem: Yeah, the original message for the exception is very obscure, it just says 'Unknown Error'. $e->getCause()->userinfo holds actual details of error, which are a lot more helpful if an error occurs.

PR #1881, opened 2017-05-24
PCHR-1344: Add env and global options for semistandard

dhavalsharma:
## Overview
Updated pre commit hook from 
`semistandard  "$files" --verbose | snazzy`
to
`semistandard --global inject --global CRM --verbose "$files" | snazzy`

This will avoid declaring inject/CRM globals manually or in files.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -22,7 +22,7 @@ if [ -n "$files" ]; then
   command_exists semistandard "https://www.npmjs.com/package/semistandard"
   command_exists snazzy "https://www.npmjs.com/package/snazzy"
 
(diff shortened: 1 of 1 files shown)


AkA84:
```
--env jasmine --env amd
```
i don't think this is correct, you are basically defining that the environment for *any* file is going to be jasmine *and* amd, which is obviously not true

dhavalsharma:
i think it is inside if condition
```
files=`git diff --name-only --cached -- '*.js' ':(exclude)**/dist/**' ':(exclude)*min.js'`
if [ -n "$files" ]; then
```
which will run the semistandard command only for those files

AkA84:
that condition is to exclude any minified or distribution-ready files, not sure how it applies

PR #1888, opened 2017-05-25
PCHR-2241: Fix issues with WorkPattern.getCalendar API returning inconsistent results

tunbola:
## Overview
WorkPattern.getCalendar API returns the calendar dates for a contact for a given period. The dates returned are dates for which the contact has an active contract.
For a contact that has an active contract for a given period that spans over a year, 365 calendar dates are supposed to be returned but the API returns inconsistent results for this on different environments like staging, test sites, local.

## Before
The WorkPattern.getCalendar API returns inconsistent results.
The reason for this is that the API fetches all the existing contracts for all contacts in the system for the period and uses that to calculate calendar dates for a contact instead of fetching just the contracts for the contact.

## After
Results returned by the The WorkPattern.getCalendar API is now consistent.
The API was changed to fetch only the contracts of the given contact and use that to generate the calendar dates.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkPatternTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkPatternTest.php
@@ -695,6 +695,7 @@ class CRM_HRLeaveAndAbsences_BAO_WorkPatternTest extends BaseHeadlessTest {
   private function assertPublicHolidayQueueTask($numberOfItems, $class = null, $expectedArgument = null) {
     $queue = PublicHolidayLeaveRequestUpdatesQueue::getQueue();
     $this->assertEquals($numberOfItems, $queue->numberOfItems());
+    $item = '';
 
     if($class || $expectedArgument) {
       $item = $queue->claimItem();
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkPatternTest.php
```diff
@@ -704,6 +707,9 @@ private function assertPublicHolidayQueueTask($numberOfItems, $class = null, $ex
     if($expectedArgument) {
       $this->assertEquals($expectedArgument, $item->data->arguments[0]);
     }
-    $queue->deleteItem($item);
+
+    if($class || $expectedArgument) {
```
davialexandre: I think `if($item)` makes more sense here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendarTest.php
```diff
@@ -426,4 +426,34 @@ public function testItCanGenerateTheCalendarWhenThereAreMultiplesWorkPatternsWit
...
   }
+
+  public function testItGeneratesAccurateNumberOfCalendarDatesUsingOnlyTheContractOfTheContact() {
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date'   => CRM_Utils_Date::processDate('2016-12-31'),
+    ]);
+
+    WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => true]);
+    $contact2 =ContactFabricator::fabricate();
```
davialexandre: missing space after the = sign

PR #1892, opened 2017-05-26
PCHR-2196: Add permissions for accessing my leave and leave manager page on SSP

tunbola:
## Overview
When the Leave and Absence civiHR extension is disabled/uninstalled, the Leave manager can still see Manager Leave and My Leave page on the SSP. The reason is that it is the Employee portal module that creates the permissions that is needed to see this pages. This PR allows the L&A extension to create these permissions itself.

## Before
The L&A extension does not create the permissions for accessing Manager Leave and My Leave page on the SSP.

## After
The L&A extension creates the permissions for accessing Manager Leave and My Leave page on the SSP.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
@@ -104,6 +104,8 @@ function hrleaveandabsences_civicrm_permission(&$permissions) {
   $prefix = ts('CiviHRLeaveAndAbsences') . ': '; // name of extension or module
   $permissions['access leave and absences'] = $prefix . ts('Access Leave and Absences');
   $permissions['administer leave and absences'] = $prefix . ts('Administer Leave and Absences');
+  $permissions['access leave and absences in ssp'] = $prefix . ts('Access Leave and Absences in SSP');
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -104,6 +104,8 @@ function hrleaveandabsences_civicrm_permission(&$permissions) {
   $prefix = ts('CiviHRLeaveAndAbsences') . ': '; // name of extension or module
   $permissions['access leave and absences'] = $prefix . ts('Access Leave and Absences');
   $permissions['administer leave and absences'] = $prefix . ts('Administer Leave and Absences');
+  $permissions['access leave and absences in ssp'] = $prefix . ts('Access CiviHR Leave and Absences');
+  $permissions['manage leave and absences in ssp'] = $prefix . ts('Manage CiviHR Leave and Absences');
```
davialexandre: If you check the permissions page on Drupal, you'll see that, before this change, the permissions were grouped into 2 different sections. The first 2 were grouped under CiviCRM and the last 2 were groups under civiHR employee portal. This helped the user identify which ones were applied to the SSP and which ones to the Admin. Now there's basically no difference.

I suggested you rename the SSP permissions to:
- Access Leave and Absences in SSP
- Manage Leave and Absences in SSP

PR #1893, opened 2017-05-29
PCHR-1687: Fix Problematic Upgrades

MiyaNoctem:
## Overview
Some upgrades caused problems on staging site.  Thes problems should be addressed so they don't occur.

## Before
- Upgrader step 1000 of hremergency extension ran without errors, but wasn't creating the values for 'relationship with employee' field.  Also, values could get duplicated, if they were already added by users.
- 'Miss' value was being duplicated, since it was already added manually.  Alson reordering didn't seem to work.
- 'National Insurance' for ID types was being duplicated, since it had already been added manually.

## After
- Upgrader step 1000 of hremergency extension was trying to add relationship values to an unexisting option group, calculated form the custom field column name. Removed logic to obtain option group name and replaced for real value being used.  Also added check to verify values being added don't already exist.
- Included functionality in upgrader for HRCore to remove duplicates for 'Miss', in case it finds them. Also, moved reordering of prefixes to hrcore extension to make sure they are run in the proper order.
- Value 'National Insurance' was already being added by a prior upgrade (1500), so 1501 was adding it again.  There is no guarantee that on other sites the value will be found, so left its creation, after validating it doesn't exist. Also, deleted duplicates if found.


hrui/CRM/HRUI/Upgrader/Steps/4702.php
```diff
@@ -30,31 +33,6 @@ private function up4702_setCustomCSSURL() {
...
-          'weight' => $weight++
-        ]);
-      }
-    }
-  }
-
-  /**
    * Moves Skype option value to the top of IM list
    */
   private function up4702_moveSkypeTop() {
```
omarabuhussein: are you sure that this wouldn't make a duplicate IM , create only update when the ID is provided so you have to fetch the ID for skype option value first (you can use API chaining)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1002.php
```diff
@@ -8,21 +8,64 @@
    * @return bool
    */
   public function upgrade_1002() {
+
```
omarabuhussein: remove the space

hrident/CRM/HRIdent/Upgrader.php
```diff
@@ -100,16 +100,40 @@ public function upgrade_1500(){
    * to be on the top.
    */
   public function upgrade_1501() {
-    // hence that type_20130502144049 is
-    // the hardcoded name for this option group
-    civicrm_api3('OptionValue', 'create', [
+
```
omarabuhussein: remove the space

org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1000.php
```diff
@@ -12,30 +12,42 @@ public function upgrade_1000() {
...
+          'value' => strtolower($relationship),
+          'is_active' => 1
         ]);
       }
     }
 
     return TRUE;
   }
 
+  private function up1000_valueExists($value, $groupName) {
```
omarabuhussein: up1000_valueExists doesn't mean anything , a better name would probably be : 

up1000_IsOptionValueExist

omarabuhussein: also can you add a docblock for this method

hrui/CRM/HRUI/Upgrader/Steps/4702.php
```diff
@@ -36,10 +36,12 @@ private function up4702_setCustomCSSURL() {
...
-    civicrm_api3('OptionValue', 'create', [
+    civicrm_api3('OptionValue', 'get', [
       'option_group_id' => 'instant_messenger_service',
       'name' => 'Skype',
-      'weight' => 0,
+      'api.OptionGroup.create' => [
```
omarabuhussein: this is wrong , it should be like this : 

```php
  private function up4702_moveSkypeTop() {
    civicrm_api3('OptionValue', 'get', [
      'option_group_id' => 'instant_messenger_service',
      'name' => 'Skype',
      'api.OptionValue.create' => [
        'id' => $value.id'
        'weight' => 0,
      ],
    ]);
  }
```

org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1000.php
```diff
@@ -14,7 +14,7 @@ public function upgrade_1000() {
 
     // create required options
     foreach($relationships as $relationship) {
-      if (!$this->up1000_valueExists($relationship, 'relationship_with_employee_20150304120408')) {
+      if (!$this->up1000_isExistingOptionValue($relationship, 'relationship_with_employee_20150304120408')) {
```
omarabuhussein: I think up1000_isOptionValueExist make more sense and even shorter than up1000_isExistingOptionValue

PR #1895, opened 2017-05-29
PCHR-2180: Fix issue with Manager being able to view own leave requests

tunbola:
## Overview
Currently when adding a relationship for a contact, it is possible to add a Leave Approver relationship with between a contact and itself. For the case of a Leave manager, this would allow the manager view his/her own leave requests in the manage leave page. 

## Before
A contact can be assigned a Leave Approver relationship with itself.

## After
A contact cannot be assigned a Leave Approver relationship with itself.

relationship
---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
@@ -407,6 +407,24 @@ function hrleaveandabsences_hrcomments_selectWhereClause(&$conditions, $params) 
   $conditions = array_merge($conditions, $commentsWhereClause->get());
 }
 
+
+/**
+ * Implementation of the hook_civicrm_validateForm.
+ *
+ * @param string $formName
+ * @param array $fields
+ * @param array $files
+ * @param object $form
+ * @param array $errors
+ */
+function hrleaveandabsences_civicrm_validateForm($formName, &$fields, &$files, &$form, &$errors) {
+  if($formName == 'CRM_Contact_Form_Relationship') {
+    if(_hrleaveandabsences_contact_is_being_assigned_as_its_own_leave_approver($form, $fields)){
+      $errors['relationship_type_id'] = ts('You cannot assign a contact as its own leave approver');
+    }
+  }
+}
+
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -714,3 +714,25 @@ function _hrleaveandabsences_set_has_leave_approved_by_as_default_relationship_t
...
+ * @param string $formName
+ * @param array $fields
+ * @param array $files
+ * @param object $form
+ * @param array $errors
+ */
+function hrleaveandabsences_civicrm_validateForm($formName, &$fields, &$files, &$form, &$errors) {
+  if ($formName == 'CRM_Contact_Form_Relationship') {
+    //True If the contact is being assigned a relationship with itself
+    if($fields['related_contact_id'] == $form->_contactId) {
```
davialexandre: Instead of adding a comment, please extract this to a separate function with a better name describing what it does. My suggestion for the name is '_hrleaveandabsences_contact_is_being_assigned_as_its_own_leave_approver'.

This function will then look something like this:
```php
if (_hrleaveandabsences_contact_is_being_assigned_as_its_own_leave_approver($form)) {
  $errors['relationship_type_id'] = ts('You cannot assign a contact as its own leave approver');
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -714,3 +714,25 @@ function _hrleaveandabsences_set_has_leave_approved_by_as_default_relationship_t
...
+/**
+ * Implementation of the hook_civicrm_validateForm.
+ *
+ * @param string $formName
+ * @param array $fields
+ * @param array $files
+ * @param object $form
+ * @param array $errors
+ */
+function hrleaveandabsences_civicrm_validateForm($formName, &$fields, &$files, &$form, &$errors) {
```
davialexandre: Remember this file has 2 sections: one to add hooks implementation and another one for helpers method. This is a hook, but it's in the helpers section. Please move it to the appropriate location

PR #1897, opened 2017-05-30
PCHR-2274: Fix logic to compare leave request attributes

deb1990:
## Problem
In leave request popup if a comment is added and then deleted, the save button was staying enabled, which is wrong as it should be disabled.

## Solution
1. Previously while defining `initialLeaveRequestAttributes` we did not make a copy of it, and because of that when a comment got added to the request `initialLeaveRequestAttributes` was also getting updated because of call by reference. Now a clone of the object is used to define the `initialLeaveRequestAttributes`
`initialLeaveRequestAttributes = angular.copy(self.request.attributes());`

2. New function `hasRequestChanged` is created to compare the state of the leave request, here  `angular.equals` is used to ignore `$$hashkey` from `this.request.attributes()` which is not needed for object comparison.

3. Both 1 and 2 ensures the object comparison is not buggy as it was previously, so 
`canSubmit = canSubmit || this.request.comments.length !== initialCommentsLength;` is removed as it is not necessary anymore.

## Screencast

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,22 @@ define([
       };
 
       /**
+       * Checks if a leave request has been updated since opening the modal
+       *
+       * @return {Boolean}
+       */
+      this.hasRequestUpdated = function () {
```
AkA84: maybe `hasRequestChanged` is better worded

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,22 @@ define([
...
+        // angular copy is necessary to remove the $$hashkey property from
```
AkA84: I would put all this comments in the documentation, so they don't pollute the code
```js
/**
 * Checks...
 *
 * Angular copy is..
 * fileUploader...
 *
 * @return {Boolean}
 */
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,22 @@ define([
...
+        // all nested objects
+        var updatedRequestAttributes = angular.copy(this.request.attributes());
```
AkA84: where do the `$$hashkey` properties come from?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,22 @@ define([
...
+        // fileUploader property deleted because it will not be used
+        // in object comparison
+        delete updatedRequestAttributes.fileUploader;
+
+        return !_.isEqual(initialLeaveRequestAttributes, updatedRequestAttributes);
```
AkA84: do you think [`angular.equals`](https://docs.angularjs.org/api/ng/function/angular.equals) can solve the problem with `$$hashkey`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,22 @@ define([
...
+       *
+       * @return {Boolean}
+       */
+      this.hasRequestUpdated = function () {
+        // angular copy is necessary to remove the $$hashkey property from
+        // all nested objects
+        var updatedRequestAttributes = angular.copy(this.request.attributes());
+        // fileUploader property deleted because it will not be used
+        // in object comparison
+        delete updatedRequestAttributes.fileUploader;
```
AkA84: [`_.omit`](https://lodash.com/docs/4.17.4#omit) ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -656,13 +670,14 @@ define([
             initContact.call(self);
 
             if (self.isMode('edit')) {
-              initialLeaveRequestAttributes = self.request.attributes();
+              initialLeaveRequestAttributes = angular.copy(self.request.attributes());
+              // fileUploader property deleted because it will not be used
+              // in object comparison
+              delete initialLeaveRequestAttributes.fileUploader;
```
AkA84: you are duplicating logic here. keep this plain
```js
initialLeaveRequestAttributes = angular.copy(self.request.attributes());
```
and do the removal of properties you don't need to compare in the `hasRequestChanged` function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: It comes from angular, description added in code too.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: `initialLeaveRequestAttributes` is stored only once, so if I do the removal of property in `hasRequestChanged` then unnecessarily we have to remove the property every time while its enough to remove it only once as the object is not changing. Also `hasRequestChanged` is a scope function, so it gets called numerous times, it will not be efficient to remove the property there.

deb1990: Changed the code to use `_.omit`

deb1990:
> @deb1990 shouldn't there be a unit test? or don't you need to amend existing unit tests since you change the code a bit?

Test cases added now

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -1031,6 +1031,40 @@
           });
         });
 
+        describe('canSubmit', function () {
```
AkA84: `canSubmit()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -1031,6 +1031,40 @@
...
+            initTestController(directiveOptions);
+          });
+
+          it('does not allow to submit the leave request without changes', function () {
+            expect($ctrl.canSubmit()).toBe(false);
+          });
+
+          describe('when a comment is added', function() {
+            beforeEach(function() {
+              $ctrl.request.comments.push({
```
AkA84: for simplicity's sake, i wonder if this would work even if you did 
```js
$ctrl.request.comments.push(jasmine.any(Object));
```
the structure of the comment object is not important for this test, isnt'it?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -1031,6 +1031,40 @@
           });
         });
 
+        describe('canSubmit', function () {
+          beforeEach(function () {
+            var status = optionGroupMock.specificValue('hrleaveandabsences_leave_request_status', 'value', '3');
+            var leaveRequest = LeaveRequestInstance.init(mockData.findBy('status_id', status));
+            leaveRequest.contact_id = CRM.vars.leaveAndAbsences.contactId.toString();
```
AkA84: i really don't like the sequence of `var`s being interrupted by statements in between, it really breaks the flow of the code. Giving you two alternatives (my preference is on the first, it has one less variable declared)
```js
beforeEach(function () {
  var status = optionGroupMock.specificValue('hrleaveandabsences_leave_request_status', 'value', '3');
  var leaveRequest = LeaveRequestInstance.init(mockData.findBy('status_id', status));

  initTestController({
    contactId: leaveRequest.contact_id,
    leaveRequest: leaveRequest,
    userRole: 'staff'
  });
});
```

```js
beforeEach(function () {
  var status = optionGroupMock.specificValue('hrleaveandabsences_leave_request_status', 'value', '3');
  var leaveRequest = LeaveRequestInstance.init(mockData.findBy('status_id', status));
  var directiveOptions = {
    contactId: CRM.vars.leaveAndAbsences.contactId.toString(), // staff's contact id
    leaveRequest: leaveRequest,
    userRole: 'staff'
  };

  leaveRequest.contact_id = directiveOptions.contactId;

  initTestController(directiveOptions);
});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
AkA84: Couple of points

> hasRequestChanged is a scope function,

but why? the function doesn't seem to be called from the view, and it is also used only in the parent controller, so why isn't it a simple private function?

> initialLeaveRequestAttributes is stored only once, so if I do the removal of property in hasRequestChanged then unnecessarily we have to remove the property every time 

I understand your point, but the `_.omit` being called multiple times (if it happens) has really a negligible effect, compared to the positive effect of having a better-structure code.

The point here is that you have a logic (object comparison) that is not contained in a isolated, reusable function, but rather it's spread both inside and outside the said function. Case in point is the fact that you need to leave a comment in `this.initAfterContactSelection` to explain what you are doing
```js
this.initAfterContactSelection = function () {
  // fileUploader property deleted because it will not be used
  // in object comparison
  initialLeaveRequestAttributes = _.omit(angular.copy(self.request.attributes()), 'fileUploader');
}
```
compare to how much more clean it is when you encapsulate the whole comparison logic in the `hasRequestChanged` function
```js
this.initAfterContactSelection = function () {
  initialLeaveRequestAttributes = angular.copy(self.request.attributes());
};

this.hasRequestChanged = function () {
  return !angular.equals(
    _.omit(initialLeaveRequestAttributes, 'fileUploader'),
    _.omit(updatedRequestAttributes, 'fileUploader')
  );
};
```
so unless there is a technical reason for why you simply cannot do the above, please change the code like that

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -254,6 +252,23 @@ define([
       };
 
       /**
+       * Checks if a leave request has been changed since opening the modal
+       *
+       * Angular equals is necessary to ignore the $$hashkey(from angular) property
```
AkA84: This doesn't seem to happen anymore, so please remove it

deb1990: Changed it as discussed

PR #1900, opened 2017-05-31
PCHR-2249: Load the managees in manager leave page

deb1990:
## Problem
In Manager Report page, the filter for contacts shows all the contacts but it should only show the contacts managed by the logged in user.

## Solution
1. Instead of calling `Contact.all`, `Contact.leaveManagees` is used to load only the contacts managed by current user.
2. `leaveManagees()` has been added to the `Contact` model which calls the same function of `Contact API`.
3. For `department`, `level_type`, `location`, `region`, `contact` filters previously the whole objects was stored, changed it to only store the required value.
4. `id` is used in place `contact_id` for `Contact` objects because the `leaveManagees` does not return contact_id.
5. There was a bug before, if contact's were empty the LeaveRequests were showing up in the list, fixed it by adding the following code to `leave-request-api`
```javascript
var defer = $q.defer();

// if contact_id has an empty array for IN condition, there is no point making the
// call to the Leave Request API
// TODO Move to Base API
if (filters && filters.contact_id && filters.contact_id.IN && 	filters.contact_id.IN.length === 0) {
  defer.resolve({ list: [], total: 0, allIds: [] });
} else {
  defer.resolve(this.getAll('LeaveRequest', filters, pagination, sort, params, 'getFull', cache));
}

return defer.promise;
```
6. Added `vm.totalNoOfPages() === 0` condition to `vm.refresh()`
7. `job_contract_id` is used in place of `'id` in the chained API call to `HRJobContract`, previously it was wrong, the id returned by `HRJobRoles` is not same as id for `HRJobContract`.
8. `loadAllRequests()` renamed to more meaningful `loadManageesAndLeaves()`
9. `loadLeaveRequest()` renamed to more meaningful `loadLeaveRequests()`
10. Pagination buttons(First, Previous, Next , Last) are hidden in `manager-leave-requests.html` when necessary conditions are met 
manager_leave___hr16-4
manager_leave___hr16-4

11. Related unit tests updated.

## Refactoring
1. `Contact` model has been refactored to remove some unnecessary promise objects.

## Screenshot
manager_leave___hr16-4

## Comments
Styles for the UI Dropdown has been refactored in https://github.com/compucorp/civihr-employee-portal-theme/pull/165

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/models/contact.js
```diff
@@ -157,6 +157,16 @@ define([
...
           });
+        },
+
+        /**
+         * Finds all the contacts managed by the sent contact id
+         *
+         * @param {string} id - contact id
+         * @return {Promise} - Resolves with found contacts/API Errors
+         */
+        leaveManagees: function (id, params) {
```
AkA84: bit confused, if we didn't have this method on the model, how were we using `contactAPI.leaveManagees` then?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-requests.js
```diff
@@ -189,7 +189,7 @@ define([
      */
     vm.getUserNameByID = function (id) {
       var user = _.find(vm.filteredUsers, function (contact) {
-        return contact.contact_id === id;
+        return contact.id === id;
```
AkA84: in the PR description you have this

> In few places id is used in place contact_id because the leaveManagees does not return contact_id.

but, why? id is used in place of contact_id on what object? always `contact`? Why does it happen in "few places"? Did it work at all before?

org.civicrm.reqangular/src/common/models/contact.js
deb1990: We also have this in the instance, as it also makes sense to say `contact.leaveManagees()` and it will only load his managees. But in certain scenarios we also might need to get the managees when we only know the contact id. So added it to model.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-requests.js
deb1990: `id` is used when using `contact` object, and `contact_id` is used only for the `LeaveRequest`filter. The PR description is confusing, will fix it.

org.civicrm.reqangular/src/common/models/contact.js
```diff
@@ -59,14 +60,16 @@ define([
...
+      function injectContactIdsInFilters (filters, contactIds) {
+        filters = _(filters)
           .omit(groupFiltersKeys)
           .omit(jobRoleFiltersKeys)
           .assign({
             id: {in: _.intersection.apply(null, contactIds)}
           })
           .value();
+
+        return filters;
```
AkA84: why a separate `return` statement? Can't this simply be
```js
return _(filters)
  // ...
  .value();
```

org.civicrm.reqangular/src/common/models/contact.js
```diff
@@ -33,16 +34,16 @@ define([
...
-      function jobRoleContactids(filters) {
+      function jobRoleContactids (filters) {
         return JobRole.all(_.assign(filters, {
-          "api.HRJobContract.getsingle": {
-            "id": "$value.id"
+          'api.HRJobContract.getsingle': {
+            'id': '$value.job_contract_id'
           }
         }))
           .then(function (jobRoles) {
```
AkA84: fix the indentation

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-requests.js
```diff
@@ -356,6 +357,13 @@ define([
      * @return {Promise}
      */
     function loadLeaveRequest (type) {
```
AkA84: `loadLeaveRequests`, it's plural. Otherwise it looks like this method is used to load a specific leave request

org.civicrm.reqangular/test/models/contact_test.js
```diff
@@ -267,5 +268,26 @@ define([
...
+
+      it('calls leaveManagees function of contact API with same parameters', function () {
+        Contact.leaveManagees(contactID, params).then(function () {
+          expect(contactAPI.leaveManagees).toHaveBeenCalledWith(contactID, params);
+        });
+      });
+
+      it('returns contacts managed by the contact id', function () {
+        Contact.leaveManagees(contactID, params).then(function (data) {
+          expect(data).toEqual(contactAPIMock.mockedContacts().list);
```
AkA84: Doesn't `contactAPIMock.mockedContacts()` return *all* the mocked contacts? If yes, is this test even meaningful then?

org.civicrm.reqangular/src/common/models/contact.js
deb1990: Yes it was a stupid mistake.

org.civicrm.reqangular/test/models/contact_test.js
deb1990: Yes it returns all the contacts, here I am testing if that return is working fine, because any way the filtering logic is part of backend, and in `contactAPIMock` whatever value is set to be returned, needs to be checked here if the same is returned by the model.

PR #1902, opened 2017-05-31
PCHR-2240: WorkPattern::getValuesArray optimizations

davialexandre:
## Overview
The `WorkPattern::getValuesArray` method is used internally by a few important methods in L&A (For example, the` WorkPattern::getWeeks` and, indirectly, by the `WorkPatternCalendar::get` methods). This PR optimizes the implementation of this method, and consequentially also optimizes any of the methods that use it.

## Before
The method uses a chained API call to get a WorkPattern, together with its weeks and days. Chained API calls are just a way to make multiple API calls with a single request. Internally, the 3 calls will be executed separately (i.e. One call/query to get the Work Pattern, one call/query to get the WorkWeeks and one call/query to get the WorkDays).

Besides, API calls add some overhead, as each of call has to go through the whole API layer, which does a lot of things, including additional DB queries to check for custom groups and fields related to the entities being queries. We don't need any of this for this method.

## After
The method now uses a single query to get all the data, using the `CRM_Core_DAO::executeQuery()` method. Since now we're not using any API's, there won't be any additional queries for custom groups and fields, reducing drastically the number of DB queries triggered by the method.

## Comments
It was necessary to make a small change to the `getLeaveDaysForDate()` method. Since the DAO doesn't do any type of formatting to the returned data, just using `isset()` wasn't enough to check if the value was empty or not, so I had to change it and use the `empty()` function instead.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
@@ -2,6 +2,7 @@
 
 use CRM_HRCore_Date_BasicDatePeriod as BasicDatePeriod;
 use CRM_HRLeaveAndAbsences_BAO_WorkDay as WorkDay;
+use CRM_HRLeaveAndAbsences_BAO_WorkWeek as WorkWeek;
 use CRM_HRLeaveAndAbsences_BAO_WorkPattern as WorkPattern;
 use CRM_HRLeaveAndAbsences_Queue_PublicHolidayLeaveRequestUpdates as PublicHolidayLeaveRequestUpdatesQueue;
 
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -253,31 +254,94 @@ private static function getMaxWeight() {
...
+        wd.type,
+        wd.time_from,
+        wd.time_to,
+        wd.break,
+        wd.leave_days,
+        wd.number_of_hours,
+        wd.week_id
+      FROM {$workPatternTable} wp
+      LEFT JOIN {$workWeekTable} ww ON ww.pattern_id = wp.id
+      LEFT JOIN {$workDayTable} wd ON wd.week_id = ww.id
```
tunbola: Why don't we use Inner Joins here

tunbola:
In the PR overview section s/WorkPattern:getWeeks/WorkPattern::getWeeks
in the Before section s/The method a chained API/?, s/Internally, the 3 call/Internally, the 3 calls

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
davialexandre: Because, even though we do some validation on the form level to prevent the user of creating a WorkPattern without at least one Week and WorkDays, there's nothing on the database level to prevent it, which means we can have work patterns without the other child entities.

PR #1904, opened 2017-06-01
PCHR-2264 Sort request by start date

dhavalsharma:
## Overview
Now the requests are sorted by start date.

## Before
The requests were not sorted.

## After
The requests are sorted.

For manager
image

For staff
image

## Technical Details
Added sort option `'from_date DESC'` to `LeaveRequestAPI.all` calls in respective controllers for staff and manager. 

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js
```diff
@@ -26,6 +26,7 @@ define([
       all: function (filters, pagination, sort, params, cache) {
         $log.debug('LeaveRequestAPI.all');
 
+        sort = sort || 'from_date DESC';
```
AkA84: please pass this from the controller, do not put it as a default in the api

dhavalsharma: controller can still pass it and override the default behaviour here otherwise same will be used at multiple places

AkA84: i would keep the api service free of "business logic". I don't see any benefit of enforcing the sorting by from_date by default, let the controller ask for that

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/components/my-leave-report.js
```diff
@@ -395,7 +395,7 @@ define([
           to_date: {to: vm.selectedPeriod.end_date},
           request_type: 'toil',
           expired: true
-        })
+        }, null, 'from_date DESC')
```
AkA84: put this in a variable, there is no need for the duplication

PR #1905, opened 2017-06-01
PCHR-2235: Allow Public Holidays/Public Holiday Leave Requests With Past Dates To be Created

tunbola:
## Overview
Currently creating a Public Holiday with past dates is not allowed as it would involve creating Public Holiday Leave requests for employees for that date (If MTPHL is true) and this will impact on things like Entitlements, Brought Forward, Expired balance Changes and so on, However there are some valid reasons why this may be necessary, one of such is when migrating data from an old system to use the new L&A it may be necessary to create past public holidays for contacts, also when creating a contract for an employee and the contract is created after the dates for some public holidays have passed but those public holidays are supposed to be included in the contract.

## Before
The system does not allow creating Public Holidays in the past and also does not allow creating past public holidays within a contract period when creating a Contract for a contact.

## After
The system was  modified to allow creating Public Holidays in the past and also to allow creating past public holiday leave requests within a contract period when creating a Contract for a contact. After a Public Holiday Leave request with past dates is created, recalculation of Expired Balance Changes is done to keep the calculation in sync.

## Comments
Creating a Public Holiday Leave Requests will require Entitlement Recalculation to be done as this has an impact on Contacts Entitlement, Nothing is done in the PR about this since the Entitlement Calculation is to be done manually by the user who made the change.
- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
@@ -2408,6 +2408,11 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChangeTest extends BaseHeadlessTest
   public function testCalculateAmountForDateForAContactWithWorkPatternHavingMultipleWeeks() {
     $periodStartDate = new DateTime('2016-01-01');
 
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
+
     $contract = HRJobContractFabricator::fabricate(
       [ 'contact_id' => 1 ],
       [ 'period_start_date' => $periodStartDate->format('Y-m-d') ]
@@ -2459,6 +2464,11 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChangeTest extends BaseHeadlessTest
   public function testCalculateAmountForDateForAContactWithWorkPatternHavingOneWeek() {
     $periodStartDate = new DateTime('2016-01-01');
 
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
+
     $contract = HRJobContractFabricator::fabricate(
       [ 'contact_id' => 1 ],
       [ 'period_start_date' => $periodStartDate->format('Y-m-d') ]
@@ -2503,6 +2513,11 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChangeTest extends BaseHeadlessTest
 
   public function testCalculateAmountForDateForAContactUsingTheDefaultWorkPattern() {
     $periodStartDate = new DateTime('2016-01-01');
+
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
 
     $contract = HRJobContractFabricator::fabricate(
       [ 'contact_id' => 1 ],
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
@@ -181,7 +181,7 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreationTest exten
     $this->assertEquals(0, LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement));
   }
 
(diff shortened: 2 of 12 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -2408,6 +2408,11 @@ public function testRecalculateExpiredBalanceChangesWhenAbsenceTypeOfLeaveReques
   public function testCalculateAmountForDateForAContactWithWorkPatternHavingMultipleWeeks() {
     $periodStartDate = new DateTime('2016-01-01');
 
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
```
davialexandre: Why did you need to add the absence period on these tests? Is it because now the creation of Public Holidays will trigger the recalculation of expired records? If that is the case, what was the performance impact of this during the tests?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -531,4 +540,184 @@ public function testCreateLeaveRequestsForAllPublicHolidaysInTheFutureForDefault
...
+  public function testExpiredBalanceChangeIsRecalculatedForAllContactsWhenCreatingPublicHolidayWithPastDates() {
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-06-01'),
+      'end_date' => CRM_Utils_Date::processDate('+10 days')
+    ]);
+
+    $contact1 = ContactFabricator::fabricate();
+    $contact2 = ContactFabricator::fabricate();
+
+
```
davialexandre: please remove one of those 2 empty lines here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -531,4 +540,184 @@ public function testCreateLeaveRequestsForAllPublicHolidaysInTheFutureForDefault
...
+      'period_id' => $absencePeriod->id,
+      'type_id' => $this->absenceType->id,
+    ]);
+
+    HRJobContractFabricator::fabricate([
+      'contact_id' => $contact1['id']
+    ],
+    [
+      'period_start_date' => CRM_Utils_Date::processDate('2016-06-01'),
+    ]);
```
davialexandre: can we please make this a bit more concise by using this formatting?
```php
HRJobContractFabricator::fabricate(
  [ 'contact_id' => $contact1['id'] ],
  [ 'period_start_date' => CRM_Utils_Date::processDate('2016-06-01') ]
);
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -531,4 +540,184 @@ public function testCreateLeaveRequestsForAllPublicHolidaysInTheFutureForDefault
...
+
+    $expiredBalanceChange1 = $this->getBalanceChangeRecord($balanceChange1->id);
+    //The public holiday date (2016-11-03) is before the date the balance change expired so one will be deducted
+    //from $balanceChange1
+    //Public Holiday2 is in past but the date is not before an already expired balance change
+    //Public Holiday3 is in the future, so it does not affect the recalculation
+    $this->assertEquals(-2, $expiredBalanceChange1->amount);
+  }
+
+  private function getBalanceChangeRecord($balanceChangeID) {
```
davialexandre: Couldn't you have just used `LeaveBalanceChange::findById($id)` instead?

tunbola: Why didn't I think of that

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
tunbola: Yes absence period needs to be present when creating public holiday with past dates. See here [Here](https://github.com/civicrm/civihr/blob/c3fb4d1dca6a337b6f54d2337299792e6164c544/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php#L877-L877). There did not to be much performance difference as I ran tests both on my branch and also on the la-milestone-3 branch.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
davialexandre: well... my idea was to actually replace `getBalanceChangeRecord()` with `findById()`. Since now one method is just delegating the job to the other, there's really no reason for `getBalanceChangeRecord()` to exist

PR #1912, opened 2017-06-02
PCHR-2219: Remove the default API limit

omarabuhussein:
## Overview
CiviCRM API has by a default a limit of 25 results when doing a "get" action and when there is no limit specified. This could cause problems specially when some developers forget this fact so its better to remove the limit and whenever the developer need any limit on the results then he can just add the limit as a parameter to the API.

## Before
If no limit specified , all "get" API calls return only the first 25 results.

2017-06-02 19_30_26-civihr api _ hr16_3

2017-06-02 19_31_50-civihr api _ hr16_3

## After
If not limit specified , all the results appear and not just the first 25 :
 
2017-06-02 19_34_36-civihr api _ hr16_3

also if the limit specified  , it should still works fine as before : 

2017-06-02 19_31_50-civihr api _ hr16_3

## Technical Details

I've implemented [hook_civicrm_apiWrappers ](https://docs.civicrm.org/dev/en/master/hooks/hook_civicrm_apiWrappers/) hook inside hrcore extension and defined a wrapper class called CRM_HRCore_APIWrapper_DefaultLimitRemover  that take care of setting the limit to 0 if its not set.

---

- [x] Tests Pass

Some tests Are already broken on staging so they are broken on this branch too but the rest are passing without problem , Here are the failed tests on staging  : 

**Job roles Extension**

Tests: 20, Assertions: 41, Failures: 2.

**contactsummary Extension**

Fatal error: require_once(): Failed opening required 'CiviTest/CiviUnitTestCase.php' (include_path='.:/usr/share/php:/usr/share/pear') in /vagrant/hr16_6/sites/all/modules/civicrm/tools/extensions/civihr/contactsummary/tests/phpunit/CRM/Contactsummary/Utils/AbsencesTest.php on line 3

**hrui Extension**

PHP Fatal error:  Uncaught CiviCRM_API3_Exception: [2001: 'task_status' is not a valid option for field option_group_id

  thrown in /vagrant/hr16_6/sites/all/modules/civicrm/api/api.php on line 45

Fatal error: Uncaught CiviCRM_API3_Exception: [2001: 'task_status' is not a valid option for field option_group_id

  thrown in /vagrant/hr16_6/sites/all/modules/civicrm/api/api.php on line 45

vagrant@vagrant-ubuntu-trusty-64:/vagrant/hr16_6/sites/all/modules/civicrm/tools/extensions/civihr/hrui$ phpunit4 tests
/phpunit/CRM/

Installing hr166civi_bf66o schema
PHP Fatal error:  Uncaught CiviCRM_API3_Exception: [2001: 'task_status' is not a valid option for field option_group_id

  thrown in /vagrant/hr16_6/sites/all/modules/civicrm/api/api.php on line 45

Fatal error: Uncaught CiviCRM_API3_Exception: [2001: 'task_status' is not a valid option for field option_group_id

  thrown in /vagrant/hr16_6/sites/all/modules/civicrm/api/api.php on line 45
  
  
  
**Appraisals Extension**
  
FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
  
  
**Sample data Extension**

  FAILURES!
Tests: 29, Assertions: 90, Errors: 5.

So I will assume the tests on this branch are passed too since it is not the cause for these issues , but these test issues need to be addressed and fixed later.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
@@ -8,7 +8,7 @@ class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
   const NO_LIMIT_ON_RESULTS = 0;
 
   /**
-   * the wrapper contains a method that allows you to alter the parameters of the api request (including the action and the entity)
+   * {@inheritDoc}
    */
   public function fromApiInput($apiRequest) {
     $this->removeDefaultLimit($apiRequest);
@@ -17,7 +17,7 @@ class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
   }
 
   /**
-   * alter the result before returning it to the caller.
+   * {@inheritDoc}
    */
   public function toApiOutput($apiRequest, $result) {
     return $result;
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
@@ -5,7 +5,11 @@ class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
   /**
    * In civicrm API, 0 means there is no limit on the retrieved results
    */
-  const NO_LIMIT_ON_RESULTS = 0;
+  static private $DEFAULT_NO_LIMIT_VALUE = 0;
+
+  public function getDefaultNoLimitValue() {
+    return self::$DEFAULT_NO_LIMIT_VALUE;
+  }
 
   /**
    * {@inheritDoc}
@@ -30,7 +34,7 @@ class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
    */
   private function removeDefaultLimit(&$apiRequest) {
     if (empty($apiRequest['params']['options']['limit'])) {
-      $apiRequest['params']['options']['limit'] = self::NO_LIMIT_ON_RESULTS;
+      $apiRequest['params']['options']['limit'] = $this->getDefaultNoLimitValue();
     }
   }
 }
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
@@ -1,0 +1,36 @@
+<?php
+
+class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
+
+  /**
+   * In civicrm API, 0 means there is no limit on the retrieved results
+   */
+  const NO_LIMIT_ON_RESULTS = 0;
+
+  /**
+   * the wrapper contains a method that allows you to alter the parameters of the api request (including the action and the entity)
+   */
+  public function fromApiInput($apiRequest) {
+    $this->removeDefaultLimit($apiRequest);
+
+    return $apiRequest;
+  }
+
+  /**
+   * alter the result before returning it to the caller.
+   */
+  public function toApiOutput($apiRequest, $result) {
+    return $result;
+  }
+
+  /**
+   * Removes the default API limit if it's not set
+   *
+   * @param $apiRequest
+   */
+  private function removeDefaultLimit(&$apiRequest) {
+    if (empty($apiRequest['params']['options']['limit'])) {
+      $apiRequest['params']['options']['limit'] = self::NO_LIMIT_ON_RESULTS;
+    }
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcore/phpunit.xml.dist
+++ b/uk.co.compucorp.civicrm.hrcore/phpunit.xml.dist
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<phpunit backupGlobals="false" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" syntaxCheck="false" bootstrap="tests/phpunit/bootstrap.php">
+  <testsuites>
+    <testsuite name="My Test Suite">
+      <directory>./tests/phpunit</directory>
+    </testsuite>
+  </testsuites>
+  <filter>
+    <whitelist>
+      <directory suffix=".php">./</directory>
+    </whitelist>
+  </filter>
+  <listeners>
+    <listener class="Civi\Test\CiviTestListener">
+      <arguments/>
+    </listener>
+  </listeners>
+</phpunit>
--- a/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
+++ b/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
@@ -3,6 +3,8 @@
 use Civi\Test\HeadlessInterface;
 use Civi\Test\TransactionalInterface;
 
+use CRM_HRCore_APIWrapper_DefaultLimitRemover as DefaultLimitRemover;
+
 /**
  * Class CRM_HRCore_APIWrapper_DefaultLimitRemoverTest
  *
@@ -10,9 +12,10 @@ use Civi\Test\TransactionalInterface;
  */
 class CRM_HRCore_APIWrapper_DefaultLimitRemoverTest extends \PHPUnit_Framework_TestCase implements HeadlessInterface, TransactionalInterface {
 
-  private $defaultLimitRemoverObj;
-
-  private $testParameters = [];
+  /**
+   * @var DefaultLimitRemover
+   */
+  private $defaultLimitRemover;
 
   public function setUpHeadless() {
     return \Civi\Test::headless()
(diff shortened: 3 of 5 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
```
davialexandre: Can you please add a test to this class?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * In civicrm API, 0 means there is no limit on the retrieved results
+   */
+  const NO_LIMIT_ON_RESULTS = 0;
+
+  /**
+   * the wrapper contains a method that allows you to alter the parameters of the api request (including the action and the entity)
```
davialexandre: Please improve the formatting of all docblocks:
- Make sure the description always start with an uppercase letter
- For long paragraphs, please break it into multiple short lines, with around 80 chars
- Add `@params` and `@return` sections

omarabuhussein: I tend to leave default civi docblocks for such things , but sure .. I think it's better to reformat the default docblocks

uk.co.compucorp.civicrm.hrcore/phpunit.xml.dist
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<phpunit backupGlobals="false" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" syntaxCheck="false" bootstrap="tests/phpunit/bootstrap.php">
+  <testsuites>
+    <testsuite name="My Test Suite">
```
davialexandre: Rename the test suite to "HRCore Test Suite"

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/APIWrapper/DefaultLimitRemover.php
```diff
@@ -5,7 +5,11 @@ class CRM_HRCore_APIWrapper_DefaultLimitRemover implements API_Wrapper {
   /**
    * In civicrm API, 0 means there is no limit on the retrieved results
    */
-  const NO_LIMIT_ON_RESULTS = 0;
+  static private $DEFAULT_NO_LIMIT_VALUE = 0;
```
davialexandre: Since this is the default value for the Limit option, I think it makes more sense to name the variable as `DEFAULT_API_LIMIT_VALUE`.

Also, what's the reason for changing this from a constant to a variable? Did you change it only to make it private?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Civi\Test\TransactionalInterface;
+
+/**
+ * Class CRM_HRCore_APIWrapper_DefaultLimitRemoverTest
+ *
+ * @group headless
+ */
+class CRM_HRCore_APIWrapper_DefaultLimitRemoverTest extends \PHPUnit_Framework_TestCase implements HeadlessInterface, TransactionalInterface {
+
+  private $defaultLimitRemoverObj;
```
davialexandre: Cmon, we have annotations. There's no need to suffix the variable with Obj to indicate this is an object

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -0,0 +1,68 @@
...
+  private $testParameters = [];
+
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->installMe(__DIR__)
+      ->apply();
+  }
+
+  public function setUp() {
+    $this->defaultLimitRemoverObj = new CRM_HRCore_APIWrapper_DefaultLimitRemover();
```
davialexandre: You can alias the full class name to make it shorter and easier to read:

```php
use CRM_HRCore_APIWrapper_DefaultLimitRemover as DefaultLimitRemover;
...
$this->defaultLimitRemover = new DefaultLimitRemover()
```

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+    $this->testParameters['params'] = [
+      'action' => 'get',
+      'entity' => 'contact',
+      'id' => 1,
+    ];
+  }
+
+  public function testFromApiInputWithLimitSetInParameters() {
```
davialexandre: As a best practice, it's best that the test name describes how the class should behave. So, instead of `testFromApiInputWithLimitSetInParameters` a better name would be `testTheDefaultLimitShouldNotBeAppliedIfASpecificLimitHasBeenSet`

The same applies to the other test names

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public function setUp() {
+    $this->defaultLimitRemoverObj = new CRM_HRCore_APIWrapper_DefaultLimitRemover();
+
+    $this->testParameters['params'] = [
+      'action' => 'get',
+      'entity' => 'contact',
+      'id' => 1,
+    ];
```
davialexandre: Things like action, entity and id are not relevant to most of the tests, so there's no reason to have this on all of them. I think you can remove this variable altogether

uk.co.compucorp.civicrm.hrcore/phpunit.xml.dist
omarabuhussein: ops missed this

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
omarabuhussein: don't know , its too big name .. I feel like these big names lack readability even if it a test method

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -21,43 +29,37 @@ public function setUpHeadless() {
...
 
   public function setUp() {
-    $this->defaultLimitRemoverObj = new CRM_HRCore_APIWrapper_DefaultLimitRemover();
-
-    $this->testParameters['params'] = [
-      'action' => 'get',
-      'entity' => 'contact',
-      'id' => 1,
-    ];
+    $this->defaultLimitRemover = new DefaultLimitRemover();
```
davialexandre: Since you're not resetting `testParameters` on this method anymore, you should not share this variable among all the tests anymore (that is, you should not use as a class property). The reason is that one test should be isolated and not affected by changes made by other tests. One test might expect limit to always start as 0, but it will fail if it runs after another one that will change limit to 10.

So what needs to be done here is:
- Remove the `testParameters` property, as I mentioned on the previous comment
- On each test, create the the a variable with all the values it needs

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -21,43 +29,37 @@ public function setUpHeadless() {
...
     $this->assertEquals(
       $this->testParameters['params']['options']['limit'],
       $actualValue['params']['options']['limit']
     );
   }
 
-  public function testFromApiInputWithLimitNotSetInParameters() {
-    $actualValue =  $this->defaultLimitRemoverObj->fromApiInput($this->testParameters);
+  public function testTheDefaultNoLimitValueShouldBeAppliedIfNoLimitHasBeenSet() {
+    $actualValue =  $this->defaultLimitRemover->fromApiInput($this->testParameters);
```
davialexandre: So here is the problem I'm talking about. There's no way you can be sure of what `$this->testParameters` will look like. It might have been changed by another test and will make this test fail

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
```diff
@@ -21,43 +29,37 @@ public function setUpHeadless() {
...
 
     $this->assertEquals(
       $this->testParameters['params']['options']['limit'],
       $actualValue['params']['options']['limit']
     );
   }
 
-  public function testFromApiInputWithLimitNotSetInParameters() {
-    $actualValue =  $this->defaultLimitRemoverObj->fromApiInput($this->testParameters);
+  public function testTheDefaultNoLimitValueShouldBeAppliedIfNoLimitHasBeenSet() {
```
davialexandre: s/DefaultNoLimit/DefaultLimit/

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
davialexandre: the problem here is not the size of the name, but the syntax we're using. I personally prefer snake case for test names (`public function default_limit_should_not_be_applied_if_a_specific_limit_has_been_set`), but unfortunately, we're following civi's convention here. Maybe we can change this in the future

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/APIWrapper/DefaultLimitRemoverTest.php
omarabuhussein: I meant here "Default No Limit Value" which is 0  " .. but seems like its not very clear like that so I will change it

PR #1914, opened 2017-06-04
PCHR-2249: Remove chained API calls while getting count

deb1990:
## Overview
Remove chained API calls from base `API`while getting count of total number of records.

## Technical Details
1. In Base API's `getAll()` the same API call is made twice, 1st call is made only to get specific no of records according to pagination setup, and the 2nd call is made to get all the data to determine no of pages for `pagination`. The 2nd call does not need to have the chained API calls as it only cares about total no of records and not the actual data. So removing the chained API calls will improve performance.
```javascript
_.omit(params, function (__, key) { return _.startsWith(key, 'api.'); });
```
---

- [x] Tests Pass

### DIFF
--- a/org.civicrm.reqangular/src/common/services/api.js
+++ b/org.civicrm.reqangular/src/common/services/api.js
@@ -85,9 +85,8 @@ define([
           }.bind(this))(),
           (function () {
             var params = _.assign({}, filters, { 'return': 'id' });
-            params = _.omit(params, function (value, key) {
-              return key.startsWith('api.');
-            });
+            // Removing chained calls, they are not necessary for getting the count
+            params = _.omit(params, function (__, key) { return _.startsWith(key, 'api.'); });
 
             return this.sendGET(entity, action, params, cache);
           }.bind(this))()
--- a/org.civicrm.reqangular/src/common/services/api.js
+++ b/org.civicrm.reqangular/src/common/services/api.js
@@ -1,3 +1,4 @@
+/* eslint-env amd */
 define([
   'common/angular',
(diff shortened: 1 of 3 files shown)


org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -84,6 +85,9 @@ define([
           }.bind(this))(),
           (function () {
             var params = _.assign({}, filters, { 'return': 'id' });
+            params = _.omit(params, function (value, key) {
+              return key.startsWith('api.');
```
AkA84: `String.prototype.startsWith()` is not available on IE, please always check the compatibility (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)

Use [`_.startsWith`](https://lodash.com/docs/4.17.4#startsWith)

org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -84,6 +85,9 @@ define([
           }.bind(this))(),
           (function () {
             var params = _.assign({}, filters, { 'return': 'id' });
```
AkA84: Use the chaining feature of lodash (also add an inline comment to explain what's going on)
```js
(function () {
  // Removing chained calls, they are not necessary for getting the count
  var params = _({})
    .assign(filters, { 'return': 'id' })
    .omit(function (__, key) { return _.startsWith(key, 'api.'); })
    .value();

  return this.sendGET(entity, action, params, cache);
}.bind(this))()
```

PR #1917, opened 2017-06-05
PCHR-2240: Improvements to the response format of the WorkPattern.getCalendar API

davialexandre:
## Overview
The calendar tab of the Manager Leave tab has multiple performance issues. Some, are related to the amount of time it takes to load the calendar information for a big number of contacts. This PR introduces an improvements that reduces the time it takes to load the data form the server.

## Before
The `WorkPattern.getCalendar` only supported getting the calendar for a single contact per call. This means that, in order to display the calendar for 20 contacts, 20 API calls were necessary, meaning 20 http requests to the server and having both CiviCRM and Drupal being bootstrapped 20 times.

## After
The `WorkPattern.getCalendar` API now supports getting calendars for multiple contacts in a single call. This reduces the number of times Drupal + CiviCRM are bootstraped, reducing drastically the amount of time taken to start downloading the data from the server. The syntax to get the calendar for multiple contacts is:

```javascript
CRM.api3('WorkPattern', 'getcalendar', {
  "sequential": 1,
  "contact_id": {"IN":[1,2,3,4]},
  "period_id": ""
});
```
The response format was also changed to fit multiple calendars. The values array is now an array of calendars, where each item contains the contact_id and its related calendar.:

```javascript
{
  "is_error": 0,
  "count": 2,
  "values": [
    {
      "contact_id": 1,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        },
        {
          "date": "2017-01-02",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        }
      ]
    },
    {
      "contact_id": 2,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        },
        {
          "date": "2017-01-02",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        }
      ]
    }
  ]
}
```
The API still supports a single contact ID as the param (without the IN operator). The response format will be the same, where values will still be an array, but containing a single item.

## Frontend
Following is the list of frontend changes introduced in this PR

### Support multiple contacts in `WorkPatternAPI.getCalendar`
The `WorkPatternAPI.getCalendar` had been augmented to support multiple contact ids passed as an array. In case of multiple ids, the param passed to the API will be wrapped in a "IN" condition
```js
contact_id: _.isArray(contactId) ? { 'IN': contactId } : contactId,
```

### Support multiple contacts in `Calendar.get`
If multiple contacts are passed to `Calendar.get`, then the modal will return an array of `CalendarInstance`s. If only a contact has been passed, then it returns just one instance
```js
var list = data.values.map(function (contactCalendar) {
  return instance.init(contactCalendar);
});

return _.isArray(contactId) ? list : list[0];
```

### Changes to `CalendarInstance`
We were overriding the original `init` method to modify the attributes of the instance. Now we are using the correct method, `transformAttributes`  (which is called by `init` automatically). 

And about attributes, we are keeping the `contact_id` coming from the api, while dumping `calendar` and creating the `days` property instead

```js
transformAttributes: function (attributes) {
  // ...
  return _(attributes)
    .omit('calendar')
    .assign({ days: datesObj })
    .value();
},
```

#### Before
We were calling `Calendar.get` as many times as there were contacts
```js
return $q.all(vm.managedContacts.map(function (contact, index) {
  return Calendar.get(contact.id, vm.selectedPeriod.id)
    .then(function (calendar) {
      tempContactData[index].calendarData = vm._setCalendarProps(vm.managedContacts[index].id, calendar);
    });
}));
```

#### After
We are now passing an array containing all the contacts ids to `Calendar.get`
```js
tempContactData = _.clone(vm.managedContacts);

return Calendar.get(vm.managedContacts.map(function (contact) {
  return contact.id;
}), vm.selectedPeriod.id)
.then(function (calendars) {
  calendars.forEach(function (calendar) {
    var index = _.findIndex(tempContactData, function (contact) {
      return +contact.id === +calendar.contact_id;
    });

    tempContactData[index].calendarData = vm._setCalendarProps(calendar.contact_id, calendar);
  });
});
```
notice that because the contacts are stored in `tempContactData` by numeric indexes, we need to find first at which position the contact of each calendar is stored 

### Updated work pattern mocked data
The mocked data representing the response of `WorkPattern.getCalendar` had been updated to reflect the changes in the response's structure introduced in this PR.

Also, the module was returning an object whose methods you would have to call in order to get the data
```js
define(function () {
  var mockData = {
    daysData: { /* ... */ }
  }

  return {
    daysData: function () {
      return mockData.daysData;
    }
  }
});
```
but this to me is an unnecessary complication (what's the point of calling a method if all it's doing is just returning the data?), so i changed it to a more straightforward structure
```js
define(function () {
  return {
    getCalendar: { /* .. */}
  }
});
```

### Refactored tests and overall style
Some tests have been refactored, and i've fixed the style to follow `semistandard` of all the files i've been worked on

## Comments

There were two `PublicHolidayLeaveRequestCreation` tests failing due to the changes introduced by https://github.com/civicrm/civihr/pull/1905. They were fixed by this PR

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -70,44 +84,74 @@ public function testGetCalendarUsesTheWorkPatternCalendarService() {
...
-      'contact_id' => $contact['id']
+      'contact_id' => ['IN' => [$contact1['id'], $contact2['id']]]
     ])['values'];
 
+    $this->assertCount(2, $calendars);
+
+    // Since there's not guarantee about the order on which the calendars will
+    // be returned, we get all the contacts ID's from the result, put them in an
+    // array and then check if this array only contains the IDs we're expecting
+    $returnedContactsIDs = array_column($calendars, 'contact_id');
```
tunbola: We can actually sort this array and then
`$this->assertEquals($returnedContactsIDs, [$contact1['id'], $contact2['id']])'

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -70,44 +84,74 @@ public function testGetCalendarUsesTheWorkPatternCalendarService() {
...
-    $this->assertEquals('2015-01-08', $calendarDates[3]['date']);
-    $this->assertEquals($workDayTypes['working_day'], $calendarDates[3]['type']);
-    $this->assertEquals('2015-01-11', $calendarDates[6]['date']);
-    $this->assertEquals($workDayTypes['weekend'], $calendarDates[6]['type']);
-    $this->assertEquals('2015-01-16', $calendarDates[11]['date']);
-    $this->assertEquals($workDayTypes['working_day'], $calendarDates[11]['type']);
-    $this->assertEquals('2015-01-19', $calendarDates[14]['date']);
-    $this->assertEquals($workDayTypes['working_day'], $calendarDates[14]['type']);
+    // Since both contacts have the same work pattern, both calendars should be equal
+    foreach($calendars as $calendar) {
```
tunbola: Would be nice to test this with the two contacts using different work patterns. One contact can use the default work pattern and the other contact using a different work pattern.

davialexandre:
Keeping this PR open until we also have the FE changes necessary to handle the new API response format

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
```diff
@@ -10,18 +11,19 @@ define([
...
 
       /**
-       * This method returns the calendar for a specific period, as a list of days and their type
+       * This method returns the calendar for the given contact(s) and period,
+       * as a list of days and their type
        *
-       * @param {string} contactId The ID of the Contact
-       * @param {string} periodId The ID of the Absence Period
-       * @param {object} params
+       * @param {string/int/Array} contactId can be also an array for multiple contacts
```
deb1990: we should not say that its `int`, because now with semistandard in place we need to be sure whether it is `int` or `string`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
```diff
@@ -10,18 +11,19 @@ define([
...
+       * @param {string/int} periodId The ID of the Absence Period
```
deb1990: Same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
```diff
@@ -10,18 +11,19 @@ define([
...
+       * @param {object} params additional parameters
        * @return {Promise} Resolved with {Array} All calendar records
        */
       getCalendar: function (contactId, periodId, params) {
         $log.debug('WorkPatternAPI.getCalendar', contactId, periodId, params);
 
         return this.sendGET('WorkPattern', 'getcalendar',  _.assign({}, params, {
-          contact_id: contactId,
+          contact_id: _.isArray(contactId) ? { "IN": contactId } : contactId,
```
deb1990: We can always use `{ "IN": contactId }`, if it is a single value, we will get value for the same.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
```diff
@@ -1,53 +1,83 @@
...
+          workPatternPromise = WorkPatternAPI.getCalendar(dummyContactId, dummyPeriodId, additionalFilters);
+        });
 
-      it('calls the sendGET() method', function () {
-        expect(WorkPatternAPI.sendGET).toHaveBeenCalled();
-        expect(WorkPatternAPI.sendGET.calls.mostRecent().args[0]).toBe('WorkPattern');
-        expect(WorkPatternAPI.sendGET.calls.mostRecent().args[1]).toBe('getcalendar');
-      });
+        it('sends a GET request to the api', function () {
+          expect(WorkPatternAPI.sendGET).toHaveBeenCalled();
```
deb1990: This test is not necessary given that we are testing
```
expect(WorkPatternAPI.sendGET.calls.mostRecent().args[0]).toBe('WorkPattern');
expect(WorkPatternAPI.sendGET.calls.mostRecent().args[1]).toBe('getcalendar');
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/calendar-model.js
```diff
@@ -1,34 +1,39 @@
...
         /**
-         * This method returns the calendar for a specific period, as a list of days and their type
+         * This method returns the calendar(s) for the given contact(s) and period,
+         * as a list of days and their type
          *
-         * @param {string} contactId The ID of the Contact
-         * @param {string} periodId The ID of the Absence Period
-         * @param {object} params
-         * @return {Promise} Resolved with {Object} Calendar Instance
+         * @param {string/int/Array} contactId can be also an array for multiple contacts
```
deb1990: Should be just `string/Array` as I mentioned in my previous comment

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/calendar-model_test.js
```diff
@@ -25,29 +29,49 @@ define([
...
+      describe('when passing multiple contact ids', function () {
+        beforeEach(function () {
+          WorkPatternAPI.getCalendar.and.returnValue($q.resolve(mockData.daysData()));
+          CalendarPromise = Calendar.get([jasmine.any(String), jasmine.any(String)], jasmine.any(String));
+        });
+
+        it('returns multiple CalendarInstances', function () {
+          CalendarPromise.then(function (response) {
+            expect(_.isArray(response)).toBe(true);
+            expect('days' in response[0]).toBe(true);
```
deb1990: What if we change the no of responses from mocked data. We should loop on `response` here.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/calendar-instance_test.js
```diff
@@ -9,75 +9,52 @@ define([
...
-        key = Object.keys(CalendarInstance.days)[0];
-        date = moment(CalendarInstance.days[key].date).valueOf().toString();
+        key = Object.keys(instance.days)[0];
+        date = moment(instance.days[key].date).valueOf();
       });
 
-      it('dates arrays has been converted to an object with proper timestamp values', function () {
-        expect(key).toBe(date);
+      it('converts the dates array to an object with timestamps as keys', function () {
+        expect(+key).toBe(date);
```
deb1990: Why `+` key?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-calendar.js
```diff
@@ -129,12 +129,18 @@ define([
...
-        return Calendar.get(contact.id, vm.selectedPeriod.id)
-          .then(function (calendar) {
-            tempContactData[index].calendarData = vm._setCalendarProps(vm.managedContacts[index].id, calendar);
+      return Calendar.get(vm.managedContacts.map(function (contact) {
+        return contact.id;
+      }), vm.selectedPeriod.id)
+      .then(function (calendars) {
+        calendars.forEach(function (calendar) {
+          var index = _.findIndex(tempContactData, function (contact) {
+            return +contact.id === +calendar.contact_id;
```
deb1990: Why are we type casting to int, as we discussed with Davi the backend is supposed to return string always.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-calendar.js
```diff
@@ -129,12 +129,18 @@ define([
...
-      return $q.all(vm.managedContacts.map(function (contact, index) {
-        return Calendar.get(contact.id, vm.selectedPeriod.id)
-          .then(function (calendar) {
-            tempContactData[index].calendarData = vm._setCalendarProps(vm.managedContacts[index].id, calendar);
+      return Calendar.get(vm.managedContacts.map(function (contact) {
+        return contact.id;
+      }), vm.selectedPeriod.id)
+      .then(function (calendars) {
+        calendars.forEach(function (calendar) {
+          var index = _.findIndex(tempContactData, function (contact) {
```
deb1990: Can we index `tempContactData` or `vm.managedContacts` by id? then performance would increase.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/manager-leave/components/manager-leave-calendar_test.js
```diff
@@ -84,6 +85,32 @@
...
+         * in a separate file, so that mocks/data/work-pattern-data can inject it
+         * and dynamically generate the data by looping through all mocked contacts
+         */
+        spyOn(WorkPatternAPI, 'getCalendar').and.callFake(function () {
+          var mockedGetCalendarResponse = _.clone(workPatternMocked.getCalendar);
+          var singleMockedCalendarTemplate = mockedGetCalendarResponse.values[0];
+
+          return $q.resolve(_.assign(mockedGetCalendarResponse, {
+            values: ContactAPIMock.mockedContacts().list.map(function (contact) {
+              var clonedCalendar = _.clone(singleMockedCalendarTemplate);
```
deb1990: `return _.assign(_.clone(singleMockedCalendarTemplate), { contact_id: contact.id });` is better in this case

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-calendar.js
```diff
@@ -129,12 +130,18 @@ define([
     vm._loadCalendar = function () {
       tempContactData = _.clone(vm.managedContacts);
 
-      return $q.all(vm.managedContacts.map(function (contact, index) {
-        return Calendar.get(contact.id, vm.selectedPeriod.id)
-          .then(function (calendar) {
-            tempContactData[index].calendarData = vm._setCalendarProps(vm.managedContacts[index].id, calendar);
+      return Calendar.get(vm.managedContacts.map(function (contact) {
```
deb1990: Could you add some comments for this function, to make things more clear.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-calendar.js
deb1990: Also can we some how use IIFE to fix this problem? Maybe we use a loop instead and use IIFE, but for every contact we are searching though all contacts which is a performance problem.

deb1990:
@AkA84 After this changes My Leave calendar is still working right?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
AkA84: honestly, i don't know which is it, do you know? Also the fact it can be (potentially) either string or int it's not important here, since we are not making any `===` comparison

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
AkA84: see my previous reply

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
AkA84: True, but 
1. not using "IN" is more representative of the query we are doing (give me the calendar of *this* contact)
2. I have the suspect that "IN" is less performant on the backend that using a simple equality

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
AkA84: These are two distinct tests
1. Tests if a given GET request is called
2. Tests if the GET request has been done with specific parameters

The second test in fact takes for granted that the GET request happened `WorkPatternAPI.sendGET.calls.mostRecent()`.

I thought of leaving it like this:
```js
it(/* .. */, function () {
  expect(WorkPatternAPI.sendGET).toHaveBeenCalled();
  expect(WorkPatternAPI.sendGET.calls.mostRecent().args[0]).toBe('WorkPattern');
  expect(WorkPatternAPI.sendGET.calls.mostRecent().args[1]).toBe('getcalendar');
});
```
but at that point I preferred having the test suite being more expressive, hence the two tests

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/calendar-model.js
AkA84: same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/calendar-instance_test.js
AkA84: the `key` was a string, the `date` an integer, so it was necessary to call `toString()` on the `date`. I simply cast the `key` to be integer, less code to read now

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/manager-leave/components/manager-leave-calendar.js
AkA84: > Can we index tempContactData or vm.managedContacts by id? then performance would increase.
  
honestly i'm not even going to try doing that, this PR needs to be merged today and investigating how to improve how the contacts are stored and indexed and searched in the component would definitely make the PR miss the deadline

AkA84:
@deb1990 checked the my leave calendar, it works as expected

PR #1919, opened 2017-06-06
PCHR-2293: Manager Leave or My Leave pages don't fully load sometimes

igorpavlov:
## Overview

Sometimes when you navigate to or refresh Manager Leave or My Leave pages, they don't fully load. The web-page itself is loaded, including top navigation bar, page header and footer, but the page content is not. This doesn't always happen, sometimes page is functioning properly.

## Before

The Manager Leave page content is sometimes not loaded. Same thing happens for My Leave page.

new

## After

The Manager Leave page content is  **always** loaded, as well as My Leave page content.

new

## Technical Details

The bug appears only when Debugging mode is ON.

The bug happens because the config.js file in leave_absences/shared folder, which defines the path for ui-router, is loaded asynchronously relating to angular components. This is the behaviour of RequireJS.

screenshot-1

This PR makes sure the config file loads strictly before the components.

## Comments

- Code readability improved
- Same fix was applied to absence-tab.js

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js
```diff
@@ -9,14 +9,19 @@
...
-  require([
-    'leave-absences/shared/config',
-    'leave-absences/absence-tab/app'
-  ], function () {
-    document.dispatchEvent(typeof window.CustomEvent == "function" ? new CustomEvent('absenceTabReady') : (function () {
-      var e = document.createEvent('Event');
-      e.initEvent('absenceTabReady', true, true);
-      return e;
-    })());
+  require(['leave-absences/shared/config'], function(__){
```
AkA84: please keep this format
```js
require([
  'module1',
  'module2'
, function () {

});

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/absence-tab.js [hunk 2]
```diff
@@ -9,14 +9,19 @@
...
+    require(
+      [
+        'leave-absences/absence-tab/app'
```
AkA84: but if the problem is that `/app` has a dependency on `shared/config`, doesn't that mean that this can be easily solved by including it directly as a dependency when `/app` is defined?
```js
// leave-absences/absence-tab/app

define([
  'leave-absences/shared/config',
  // etc.
], function () {

});
```
this needs to be tested though, i don't remember now why shared/config was kept separate from /app (maybe something to do with tests setup?)

igorpavlov: @AkA84 Well, app.js asks for ui-router as a dep, and config.js redefines the path to the ui-router module. Since define works async as well, if we do:

image

Just tested and it now fails 100% of time, which is expected.

AkA84: right, makes sense. thanks!

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/manager-leave.js
```diff
@@ -9,12 +9,14 @@
...
       'leave-absences/manager-leave/app',
-      'leave-absences/shared/models/leave-request-model',      
+      'leave-absences/shared/models/leave-request-model',
     ],
-    function (__, angular) {
+    function (angular) {
       angular.bootstrap(document.querySelector('[data-leave-absences-manager-leave]'), ['manager-leave']);
     });
+  })
+
```
AkA84: empty line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/my-leave.js
```diff
@@ -9,11 +9,13 @@
...
+  require(['leave-absences/shared/config'], function(__){
+    require([
+      'leave-absences/my-leave/app'
     ],
-    function (__, angular) {
+    function (angular) {
       angular.bootstrap(document.querySelector('[data-leave-absences-my-leave]'), ['my-leave']);
     });
+  });
+
```
igorpavlov: Done. Also added a missing semicolon.

igorpavlov: And also removed trailing commas.

igorpavlov: And also ran through semistandard and fixed errors.

PR #1920, opened 2017-06-06
PCHR-2228: Hide opening and closing balance

deb1990:
## Problem
Opening Balance and Current Balance is shown for approved Leave Requests, which is wrong.

## Solution
1. Showing Opening Balance and Current Balance only if the leave request's status is `Awaiting Approval` or `More Information Required`.
2. `sharedSettings.statusNames` added to `scope` of `request-ctrl` as its needs to be used in the HTML.
3. If the user is `manager` then not hiding the Opening Balance and Current Balance as it is more complicated and pending a decision to be made.
4. `statusBeforeEdit` variable was used to store the original status of the request before opening the modal and any changes made by the manager in the status dropdown was directly stored on the request object, also for staff the status was changed to `awaiting_approval`. This approach was wrong, because the request object should contain the original value of the request and any changes should only be made before sending the API request.
Now before calling the API, `changeStatusBeforeSave()` is called which changes the status of the request based on the `role` and `mode`. If the API call fails for any reason the original status is set back 
`this.request.status_id = originalStatus;`
5. `newStatusOnSave` is created which acts as a model for the status dropdown in manager role.
6. `getStatusFromValue` changed from private to public as it is needs to be used in the UI
`<span class="col-xs-6">{{::$ctrl.getStatusFromValue($ctrl.request.status_id).label}}</span>`
7. `validateBeforeSubmit` has been created to abstract the validation logic before saving a leave request, the call to `this.request.isValid()` is include here and removed from `updateRequest` and `createRequest`.
8. Refactoring done to many other places, where `bind` is used in place of `self`, variable declaration corrected, promise chaining improved.

**Approved Request**
my_leave___hr16-4

**New Request**
my_leave___hr16-4

## Comments
It was not possible to remove calculation of `opening` and `closing` balance. Which is happening in few places and hugely coupled with the rest of code, to make the performance improve a bit, used `ng-if` in the HTML. 

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
@@ -43,7 +43,6 @@ define([
     $log.debug('app.run');
 
     $rootScope.pathTpl = sharedSettings.pathTpl;
-    $rootScope.sharedSettings = sharedSettings;
     $rootScope.settings = settings;
   }]);
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
@@ -43,6 +43,7 @@ define([
     $log.debug('app.run');
 
     $rootScope.pathTpl = sharedSettings.pathTpl;
+    $rootScope.sharedSettings = sharedSettings;
     $rootScope.settings = settings;
   }]);
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
@@ -35,16 +35,16 @@ define([
       this.absenceTypes = [];
       this.calendar = {};
       this.contactName = null;
-      this.submitting = false;
       this.errors = [];
       this.managedContacts = [];
-      this.requestDayTypes = [];
-      this.selectedAbsenceType = {};
-      this.sharedSettings = sharedSettings;
+      this.newStatusOnSave = null;
       this.period = {};
       this.postContactSelection = false; // flag to track if user is selected for enabling UI
+      this.requestDayTypes = [];
       this.requestStatuses = {};
-      this.statusBeforeEdit = {};
+      this.selectedAbsenceType = {};
+      this.statusNames = sharedSettings.statusNames;
+      this.submitting = false;
       this.supportedFileTypes = '';
       this.today = Date.now();
       this.balance = {
@@ -203,7 +203,7 @@ define([
         // check if manager has changed status
         if (this.isRole('manager') && this.requestStatuses) {
           // awaiting_approval will not be available in this.requestStatuses if manager has changed selection
-          canSubmit = canSubmit && !!getStatusFromValue.call(this, this.request.status_id);
+          canSubmit = canSubmit && !!this.getStatusFromValue(this.newStatusOnSave);
         }
 
         return canSubmit && !this.isMode('view');
@@ -267,17 +267,13 @@ define([
       };
 
       /**
-       * Checks if the leave request is of a specific status(sent as a param)
-       * If 'statusBeforeEdit' is available ->its an existing request, so search is done on this object
-       * If 'statusBeforeEdit' is not present -> its a new request, so search is done on this.request object
+       * Checks if the leave request has the given status
        *
        * @param {String} leaveStatus
        * @return {Boolean}
        */
       this.isLeaveStatus = function (leaveStatus) {
-        var status = this.statusBeforeEdit.name
-          ? this.statusBeforeEdit
-          : getStatusFromValue.call(this, this.request.status_id);
+        var status = this.getStatusFromValue(this.request.status_id);
 
         return status ? status.name === leaveStatus : false;
       };
@@ -354,28 +350,32 @@ define([
 
       /**
        * Submits the form, only if the leave request is valid, also emits event
-       * to listeners self leaverequest is created.
-       * Also, checks if its an update request from manager and accordingly updates leave request
+       * to notify event subscribers about the the save.
+       * Updates request based on role and mode
        */
       this.submit = function () {
+        var originalStatus = this.request.status_id;
+
         if (this.isMode('view') || this.submitting) {
           return;
         }
 
-        // current absence type (this.request.type_id) doesn't allow self
-        if (this.balance.closing < 0 && this.selectedAbsenceType.allow_overuse === '0') {
-          // show an error
-          this.errors = ['You are not allowed to apply leave in negative'];
-          return;
-        }
+        this.submitting = true;
+        changeStatusBeforeSave.call(this);
 
-        this.errors = [];
-
-        if (canViewOrEdit.call(this)) {
-          updateRequest.call(this);
-        } else {
-          createRequest.call(this);
-        }
+        validateBeforeSubmit.call(this)
+          .then(function () {
+            // TODO - Analyze why canViewOrEdit is used, as when in View mode, request should not be updated
+            return canViewOrEdit.call(this) ? updateRequest.call(this) : createRequest.call(this);
+          }.bind(this))
+          .catch(function (errors) {
+            // if there is an error, put back the original status
+            this.request.status_id = originalStatus;
+            errors && handleError.call(this, errors);
+          }.bind(this))
+          .finally(function () {
+            this.submitting = false;
+          }.bind(this));
       };
 
       /**
@@ -527,6 +527,18 @@ define([
 
           return this.isRole('manager') ? (canRemoveStatus || status.name === 'cancelled') : canRemoveStatus;
         }.bind(this));
+      };
+
+      /**
+       * Gets status object for given status value
+       *
+       * @param {String} value - value of the status
+       * @return {Object} option group of type status or undefined if not found
+       */
+      this.getStatusFromValue = function (value) {
+        return _.find(this.requestStatuses, function (status) {
+          return status.value === value;
+        });
       };
 
       /**
@@ -717,23 +729,15 @@ define([
       }
 
       /**
-       * Creates leaverequest
+       * Validates and creates the leave request
+       *
+       * @returns {Promise}
        */
       function createRequest () {
-        var self = this;
-        self.submitting = true;
-
-        self.request.isValid()
+        return this.request.create()
           .then(function () {
-            self.request.create()
-              .then(function (result) {
-                // refresh the list
-                postSubmit.call(self, 'LeaveRequest::new');
-                self.submitting = false;
-              })
-              .catch(handleError.bind(self));
-          })
-          .catch(handleError.bind(self));
+            postSubmit.call(this, 'LeaveRequest::new');
+          }.bind(this));
       }
 
       /**
@@ -743,6 +747,19 @@ define([
        */
       function canViewOrEdit () {
         return this.isMode('edit') || this.isMode('view');
+      }
+
+      /**
+       * Changes status of the leave request before saving it
+       * For staff the status_id should be always set to awaitingApproval before saving
+       * If manager has changed the status through dropdown, assign the same before calling API
+       */
+      function changeStatusBeforeSave () {
+        if (this.isRole('staff')) {
+          this.request.status_id = this.requestStatuses[sharedSettings.statusNames.awaitingApproval].value;
+        } else if (this.isRole('manager')) {
+          this.request.status_id = this.newStatusOnSave || this.request.status_id;
+        }
       }
 
       /**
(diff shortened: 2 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -266,6 +266,22 @@ define([
...
+       *
+       * @param {String} leaveStatus
+       * @return {Boolean}
+       */
+      this.isLeaveStatus = function (leaveStatus) {
+        var status = this.statusBeforeEdit.name
+          ? this.statusBeforeEdit
+          : getStatusFromValue.call(this, this.request.status_id);
+
+        return status ? status.name === leaveStatus : false;
```
AkA84: why is simply getting the leave request status is so complicated? it should be something like this
```js
this.isStatus = function (statusName) {
  var statusId = // using the statusName, get the id

  return this.request.status_id === statusId;
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/my-leave/app.js
```diff
@@ -43,6 +43,7 @@ define([
     $log.debug('app.run');
 
     $rootScope.pathTpl = sharedSettings.pathTpl;
+    $rootScope.sharedSettings = sharedSettings;
```
AkA84: i'm not sure this is the correct approach, although i understand why you did this. If we are to take the component-based approach, then it would make sense to reduce the number of "global" dependencies in each component as much as possible. To do that, you can simply inject `shared-settings` in each component and then store the `sharedSettings.statusNames` in the scope

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -65,7 +65,9 @@ <h4 class="modal-title">
               <crm-loading show="!$ctrl.loading.showBalanceChange">
                 <div ng-show="$ctrl.uiOptions.showBalance">
                   <hr/>
-                  <div class="row chr_leave-request-modal__form-group">
+                  <div
+                    class="row chr_leave-request-modal__form-group"
+                    ng-if="$ctrl.isRole('manager') || $ctrl.isLeaveStatus($root.sharedSettings.statusNames.awaitingApproval) || $ctrl.isLeaveStatus($root.sharedSettings.statusNames.moreInformationRequired)">
```
AkA84: you could make the method to accept an array to do OR checks
```html
ng-if="$ctrl.isRole('manager') || $ctrl.isLeaveStatus([$ctrl.statusNames.awaitingApproval, $ctrl.statusNames.moreInformationRequired])">
```

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
```diff
@@ -123,7 +125,10 @@ <h4 class="modal-title">
                       </tfoot>
                     </table>
                   </div>
-                  <div class="row" ng-class="{'text-danger': $ctrl.balance.closing < 0}">
+                  <div
+                    class="row"
+                    ng-class="{'text-danger': $ctrl.balance.closing < 0}"
```
AkA84: leave a space between the `{ }`

deb1990: We have not followed it upto now for HTML, so as discussed keeping it.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup.html
deb1990: I dont think its a good idea because then it will be exclusive for OR and to use with AND we need to call it separately, what we have now is better in my view

deb1990:
> It was not possible to remove calculation of opening and closing balance. Which is happening in few places and hugely coupled with the rest of code
Please explain why, give some more details

`closing` balance is calculated in `calculateBalanceChange`,
`opening` balance is calculated in `updateBalance`, `initAbsenceType`

and the code is very minimal and would not affect performance much, but if we add a flag to remove the calculation, in the end the overall performance would be same as before as we are including an extra check. I know this code should be refactored, but lets do it in the next ticket where we can analyse and refactor the logic. As a lot of time already has been spent refactoring the same.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -35,15 +35,16 @@ define([
...
       this.errors = [];
       this.managedContacts = [];
       this.requestDayTypes = [];
       this.selectedAbsenceType = {};
+      this.statusNames = sharedSettings.statusNames;
       this.period = {};
       this.postContactSelection = false; // flag to track if user is selected for enabling UI
       this.requestStatuses = {};
-      this.statusBeforeEdit = {};
+      this.managerDropDownStatus = null;
```
AkA84: Do not tie this directly to the manager. Call it `newStatusOnSave`. I'm not sure about this name either (it's not as expressive as i'd like), but the point is that the name you chose for the property does two things:

1. It ties it to the manager role. The admin can also change the status. Also, what if the staff will be given the ability to transition to some other status?
2. It ties to a specific element of the UI. What if you can change the status via something else than a dropdown?

Calling the property with a name that simply says what it is for, makes it more maintainable

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -35,15 +35,16 @@ define([
...
+      this.submitting = false;
```
AkA84: are you rearranging the properties in alphabetical order? if so, not all of them are in order

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -266,6 +267,18 @@ define([
       };
 
       /**
+       * Checks if the leave request is of a specific status(sent as a param)
```
AkA84: missing space before "("

AkA84: Checks if the leave request has the given status

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -341,24 +354,32 @@ define([
        * Also, checks if its an update request from manager and accordingly updates leave request
        */
       this.submit = function () {
```
AkA84: I think the whole method could be streamlined as such

```js
this.submit = function () {
  var originalStatus = this.request.status_id;

  if (this.isMode('view') || this.submitting) {
    return;
  }

  validate()
    .then(function () {
      this.submitting = true;

      changeStatusBeforeSave.call(this);
    })
    .then(function () {
      return /* updateRequest() or createRequest() */
    })
    .catch(function (errors) {
      this.request.status_id = originalStatus;

      errors && handleError.call(this, errors);
    })
    .finally(function () {
      this.submitting = false;
    }.bind(this))
};
```
This to me reads much better and it's easier to reason about. Following is the annotated code

```js
this.submit = function () {
  var originalStatus = this.request.status_id;

  // this was in `validateBeforeSubmit` but this is not validation logic
  // it simply a "gate" to prevent the accidental submit when the user is not supposed to
  if (this.isMode('view') || this.submitting) {
    return;
  }

  // this will contain both the current content of `validateBeforeSubmit` but also a call to `this.request.isValid()`
  // the objective is to encapsulate all the validation logic in just one method. The method will reject a promise if 
  // some validation will fail
  validate()
    .then(function () {
      // validation is successful, begin submitting
      this.submitting = true;

      // change the status
      changeStatusBeforeSave.call(this);
    })
    .then(function () {
      // either update or save
      return /* updateRequest() or createRequest() */
    })
    .catch(function (errors) {
      // the code will go here either from `validate()` or from the update/create request
     
      // in any case, put back the original status
      this.request.status_id = originalStatus;

      // if there are any errors, handle them
      errors && handleError.call(this, errors);
    })
    .finally(function () {
      // in any case, remove the submitting state
      this.submitting = false;
    }.bind(this))
};
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -341,24 +354,32 @@ define([
...
           return;
         }
 
         this.errors = [];
 
+        originalStatus = this.request.status_id;
+        changeStatusBeforeSave.call(this);
+
+        this.submitting = true;
         if (canViewOrEdit.call(this)) {
```
AkA84: is really `canViewOrEdit` the best way to tell if we need to update or edit?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -512,6 +534,18 @@ define([
       };
 
       /**
+       * Gets status object for given status value
+       *
+       * @param value of the status
```
AkA84: missing the {type}

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -699,23 +733,18 @@ define([
       }
 
       /**
-       * Creates leaverequest
+       * Validates and Creates leaverequest
```
AkA84: the format of this text is all wrong
`Validates and creates the leave request`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -728,6 +757,22 @@ define([
       }
 
       /**
+       * Changes status of the leave request before saving it
+       */
+      function changeStatusBeforeSave() {
+        if (this.isRole('staff')) {
+          // for staff the status_id should be always set to awaitingApproval before saving
```
AkA84: put this explanations in the function documentation, you are describing the "business logic" of the function

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -728,6 +757,22 @@ define([
...
+          this.request.status_id = this.requestStatuses[sharedSettings.statusNames.awaitingApproval].value;
+        } else if (this.isRole('manager')) {
+          // if manager has changed the status through dropdown,
+          // assign the same before calling API
+          if (this.managerDropDownStatus) {
```
AkA84: one liner
```js
this.request.status_id = this.managerDropDownStatus || this.request.status_id;
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: True its not and should be analyzed, but I do not want to spend more time on this ticket as the UI is working fine. Adding a TODO instead.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Not required anymore after comment is updated

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Fixed all

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Implemented the same but one change,
```javascript
this.submitting = true;
changeStatusBeforeSave.call(this);
```
is moved before calling `validateBeforeSubmit` as it involves API call, so submitting needs to be set, also before calling `isValid`API, the status with which the request will be saved needs to be set.

PR #1921, opened 2017-06-06
PCHR-2275: Modify Leave And Absence Alerts To Conform to What is in WireFrame

tunbola:
## Overview
When a  Leave request is to be created/updated and It fails validation, the error message is displayed in a red bar, at the bottom of the Leave Request modal. This PR modifies the error message sent from the back end for these validations to conform to that described in the L&A Wireframe document. 
Also currently the error code for the leave request validation exception that is thrown when a validation fails is sent to the Frontend. This PR changes this to send the Error message because it is user friendly and more readable.

## Before
The L&A leave request validation alerts does not conform to that described in the L&A Wireframe document.
The Back end sends error code for the validation exception to the Frontend.

## After
The L&A leave request validation alerts conforms to that described in the L&A Wireframe document.
Instead of the error code being sent to the Frontend, the Back end now sends the Error message, also the error message is sent after passing it through the translate function.

## Comments
There is no validation for one of the alerts described in the ticket:  _If the user requests the leave with end date >= contract end date please display the alert "This leave request is after your contract end date. Please modify dates of this request."_ 
This validation is no longer valid as [Things that affect my balance](https://compucorp.atlassian.net/wiki/display/PCHR/Things+that+affect+my+balance) says that a leave request can be created outside contract dates. Instead the previous validation on contracts was modified to throw an exception if a leave request overlapps multiple contracts with lapses in between any of the contract periods.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -412,7 +412,13 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       'end_date' => $toDate->format('Y-m-d'),
     ]);
 
-    if ($contract['count'] > 1) {
+    $contractForEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
+      'contact_id' => $params['contact_id'],
+      'start_date' => $toDate->format('Y-m-d'),
+      'end_date' => $toDate->format('Y-m-d'),
+    ]);
+
+    if ($contract['count'] > 1 || $contractForEndDate['count'] != 1) {
       throw new InvalidLeaveRequestException(
         'This leave request is after your contract end date. Please modify dates of this request',
         'leave_request_overlapping_multiple_contracts',
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -251,7 +251,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
 
     if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past) {
       throw new InvalidLeaveRequestException(
-        'You may only request TOIL for Overtime to be worked in the future. Please modify the date of this request',
+        'You may only request TOIL for overtime to be worked in the future. Please modify the date of this request',
         'leave_request_toil_cannot_be_requested_for_past_days',
         'from_date'
       );
@@ -406,19 +406,19 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $fromDate = new DateTime($params['from_date']);
     $toDate = new DateTime($params['to_date']);
 
-    $contract = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
+    $contractsContainingToOrFromDates = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
       'contact_id' => $params['contact_id'],
       'start_date' => $fromDate->format('Y-m-d'),
       'end_date' => $toDate->format('Y-m-d'),
     ]);
 
-    $contractForEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
+    $contractsContainingEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
       'contact_id' => $params['contact_id'],
       'start_date' => $toDate->format('Y-m-d'),
       'end_date' => $toDate->format('Y-m-d'),
     ]);
 
-    if ($contract['count'] > 1 || $contractForEndDate['count'] != 1) {
+    if ($contractsContainingToOrFromDates['count'] > 1 || $contractsContainingEndDate['count'] != 1) {
       throw new InvalidLeaveRequestException(
         'This leave request is after your contract end date. Please modify dates of this request',
         'leave_request_overlapping_multiple_contracts',
@@ -597,7 +597,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $intervalInDays = $interval->format("%a");
     $maxConsecutiveLeaveDays = $absenceType->max_consecutive_leave_days;
 
-    if (!empty($maxConsecutiveLeaveDays) && $intervalInDays > $absenceType->max_consecutive_leave_days) {
+    if (!empty($maxConsecutiveLeaveDays) && $intervalInDays > $maxConsecutiveLeaveDays) {
       throw new InvalidLeaveRequestException(
         'Only a maximum '. $maxConsecutiveLeaveDays .' days leave can be taken in one request. Please modify days of this request',
         'leave_request_days_greater_than_max_consecutive_days',
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -83,7 +83,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     self::validateNoOverlappingLeaveRequests($params);
     self::validateWorkingDay($params);
     self::validateBalanceChange($params);
-    self::validateLeaveDatesDoesNotOverlapContracts($params);
+    self::validateLeaveDatesDoesNotOverlapContractsWithLapses($params);
   }
 
   /**
@@ -395,35 +395,42 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   }
 
   /**
-   * This method validates that the leave request does not have dates in more than one contract period
+   * This method validates that the leave request does not overlapp
+   * contracts with lapses in between any of the contract periods.
    *
    * @param array $params
    *   The params array received by the create method
    *
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
    */
-  private static function validateLeaveDatesDoesNotOverlapContracts($params) {
+  private static function validateLeaveDatesDoesNotOverlapContractsWithLapses($params) {
     $fromDate = new DateTime($params['from_date']);
     $toDate = new DateTime($params['to_date']);
 
-    $contractsContainingToOrFromDates = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
+    $contractsOverlappingToAndFromDates = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
       'contact_id' => $params['contact_id'],
       'start_date' => $fromDate->format('Y-m-d'),
       'end_date' => $toDate->format('Y-m-d'),
     ]);
 
-    $contractsContainingEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
-      'contact_id' => $params['contact_id'],
-      'start_date' => $toDate->format('Y-m-d'),
-      'end_date' => $toDate->format('Y-m-d'),
-    ]);
+    if ($contractsOverlappingToAndFromDates['count'] > 1) {
+      $contractToCompare = reset($contractsOverlappingToAndFromDates['values']);
+      while($nextContract = next($contractsOverlappingToAndFromDates['values'])) {
+        $intervalInDays = self::getDateIntervalInDays(
+          new DateTime($contractToCompare['period_end_date']),
+          new DateTime($nextContract['period_start_date'])
+        );
 
-    if ($contractsContainingToOrFromDates['count'] > 1 || $contractsContainingEndDate['count'] != 1) {
-      throw new InvalidLeaveRequestException(
-        'This leave request is after your contract end date. Please modify dates of this request',
-        'leave_request_overlapping_multiple_contracts',
-        'from_date'
-      );
+        $contractToCompare =  $nextContract;
+
+        if($intervalInDays > 1) {
+          throw new InvalidLeaveRequestException(
+            'This leave request is after your contract end date. Please modify dates of this request',
+            'leave_request_overlapping_multiple_contracts',
+            'from_date'
+          );
+        }
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -251,7 +251,7 @@ private static function validateTOILPastDays($params) {
 
     if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past) {
       throw new InvalidLeaveRequestException(
-        'You cannot request TOIL for past days',
+        'You may only request TOIL for Overtime to be worked in the future. Please modify the date of this request',
```
mickadoo: I'm not sure if I'm reviewing the messages or if they're just copied and pasted, but not sure why overtime is capitalized here.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -296,9 +296,10 @@ private static function validateTOILToAccruedAmountIsValid($params) {
       }
     }
 
-    if ($totalProjectedToilForPeriod > $absenceType->max_leave_accrual && !$unlimitedAccrual) {
+    $maxLeaveAccrual = $absenceType->max_leave_accrual;
+    if ($totalProjectedToilForPeriod > $maxLeaveAccrual && !$unlimitedAccrual) {
       throw new InvalidLeaveRequestException(
-        'The TOIL amount plus all approved TOIL for current period is greater than the maximum for this Absence Type',
+        'The maximum amount of leave that you can accrue is '. $maxLeaveAccrual .' days. Please modify the dates of this request',
```
mickadoo: I quite like using `sprintf` for these kind of things. Even though it's longer I think it's clearer and easier to edit later (for example if adding another variable).

`$msg = 'there are ' . $numCats . ' cats';`
vs
`$msg = sprintf('there are %d cats', $numCats);`

But of course it's up to you.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -411,9 +412,15 @@ private static function validateLeaveDatesDoesNotOverlapContracts($params) {
...
+    $contractForEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
+      'contact_id' => $params['contact_id'],
+      'start_date' => $toDate->format('Y-m-d'),
+      'end_date' => $toDate->format('Y-m-d'),
+    ]);
+
+    if ($contract['count'] > 1 || $contractForEndDate['count'] != 1) {
       throw new InvalidLeaveRequestException(
-        'The Leave request dates must not have dates in more than one contract period',
+        'This leave request is after your contract end date. Please modify dates of this request',
```
mickadoo: What about the original exception case where more than one contract was found, is this still possible?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -411,9 +412,15 @@ private static function validateLeaveDatesDoesNotOverlapContracts($params) {
       'end_date' => $toDate->format('Y-m-d'),
     ]);
 
-    if ($contract['count'] > 1) {
+    $contractForEndDate = civicrm_api3('HRJobContract', 'getcontractswithdetailsinperiod', [
```
mickadoo: This variable name sounds like it's always a singular contract, it might be renamed something like `$contractsContainingToDate`. Also, is it possible that the $toDate is in the past and this call would return an inactive contract from the past, thus passing validation here?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -588,10 +595,11 @@ private static function validateLeaveDaysAgainstAbsenceTypeMaxConsecutiveLeaveDa
 
     $interval = $toDate->diff($fromDate);
     $intervalInDays = $interval->format("%a");
+    $maxConsecutiveLeaveDays = $absenceType->max_consecutive_leave_days;
 
-    if (!empty($absenceType->max_consecutive_leave_days) && $intervalInDays > $absenceType->max_consecutive_leave_days) {
+    if (!empty($maxConsecutiveLeaveDays) && $intervalInDays > $absenceType->max_consecutive_leave_days) {
```
mickadoo: The `$absenceType->max_consecutive_leave_days` has already been made a variable, so why not use it here too?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -712,15 +717,16 @@ public function testLeaveRequestShouldNotBeCreatedWhenAbsenceTypeIsNotActive() {
...
-   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
-   * @expectedExceptionMessage Leave Request days cannot be greater than maximum consecutive days for absence type
-   */
   public function testNumberOfDaysOfLeaveRequestShouldNotBeGreaterMaxConsecutiveLeaveDaysForAbsenceType() {
     $absenceType = AbsenceTypeFabricator::fabricate([
       'max_consecutive_leave_days' => 2
     ]);
 
+    $this->setExpectedException(
+      'CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException',
```
mickadoo: I like using the `class` keyword instead of strings for class names, like `CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException::class`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -2654,6 +2693,11 @@ public function testLeaveRequestIsValidShouldNotReturnErrorWhenValidationsPass()
       'end_date'   => CRM_Utils_Date::processDate('2016-12-31'),
     ]);
 
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $contactID],
+      ['period_start_date' => '2016-01-01']
+    );
+
```
mickadoo: Wow, this class is very long. I think we should impose some sort of limit in the future sometime. It could be split into separate test classes.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Well, Its copied and pasted :). But I did review before pasting. I'll change it to use lower case.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Yeah its still possible, but using the same error message for both scenarios.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Well it will always be a singular contract because the end date can only be in one contract. Whether the toDate is in the past or not, the contract containing it will be returned so it won't pass the validation

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: I agree, although this format is used in a lot of the tests so might take time to change it across board.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Well, since just one variable is involved for now. I'll just keep as is.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
mickadoo: If it is always singular then why check count?

It's not that clear what's happening when looking at this. It looks to me like it's checking for a contract that has a start and end date that are both on the final day of the leave request.

The exception message sounds like it's saying the contact has no contracts that contain the leave request period. 

Shouldn't the validation check if the contact has any contract that starts _before_ the leave request start and ends _after_ the leave request ends? I might be misinterpreting this, but it's not so clear when I look at it what's happening.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
mickadoo: Alright, don't you think it would be confusing for the user if they see

> This leave request is after your contract end date.

But the real problem is that the leave request spans two contracts?

tunbola: Well discussed with Mike and it he said it was Ok to consolidate the error messages because technically both scenarios mean the same thing

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
mickadoo: Talked to @tunbola about this, I was confused because of the behavior of the `getcontractswithdetailsinperiod` API call which I would expect returns contracts that start after the $startDate **and** end before the $endDate, but this seems not to be the case. It seems like it returns contracts that start after the $startDate **or** end before the $endDate. This is why a second call is required to validate that the $endDate is in fact inside a valid contract.

I mentioned that the same validation should be done on the $startDate, because as it is it's possible to have a contact with a gap in contracts who applies for a leave request that ends inside a valid contract, but starts where no contract exists.

|--- June ---|  no contract |--- August ---|

- Applies for leave request from 20 July to 2 August
- First API call returns second contract since end date is before its end date
- Second API call returns second contract since end date is before its end date
- Pass validation
- Big explosion with flames everywhere

It might be outside the scope of this, but he will discuss with @davialexandre. 

For now I think we should make it as clear as possible what's happening, maybe I'm just a bit stupid but I was very confused by this. I'll suggest changing `$contract` to `$contractsContainingToOrFromDates` or something similar and `$contractForEndDate` to `$contractsContainingEndDate` because even though ideally it should only be one, it's still a request that can return multiple contracts.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -395,35 +395,42 @@ private static function validateAbsencePeriod($params) {
   }
 
   /**
-   * This method validates that the leave request does not have dates in more than one contract period
+   * This method validates that the leave request does not overlapp
```
davialexandre: s/overlapp/overlap/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -395,35 +395,42 @@ private static function validateAbsencePeriod($params) {
...
+          new DateTime($nextContract['period_start_date'])
+        );
 
-    if ($contractsContainingToOrFromDates['count'] > 1 || $contractsContainingEndDate['count'] != 1) {
-      throw new InvalidLeaveRequestException(
-        'This leave request is after your contract end date. Please modify dates of this request',
-        'leave_request_overlapping_multiple_contracts',
-        'from_date'
-      );
+        $contractToCompare =  $nextContract;
```
davialexandre: there are two spaces after the `=` sign

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1690,8 +1690,8 @@ public function testLeaveRequestCanNotBeCreatedWhenTheDatesOverlapMoreThanOneCon
     $periodStartDate1 = date('2016-01-01');
     $periodEndDate1 = date('2016-06-30');
 
-    $periodStartDate2 = date('2016-07-01');
-    $periodEndDate2 = date('2016-12-31');
+    $periodStartDate2 = date('2016-07-02');
```
davialexandre: It doesn't make sense to use the date function like this. The first param is the date format, not an actual date. We should either use `CRM_Utils_Date::processDate()` (to make it consistent with the other dates) or just assign the date string to the variable.

Also applies to the other test

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1744,26 +1745,51 @@ public function testLeaveRequestCanNotBeCreatedWhenLeaveRequestDateIsGreaterThan
...
       'contact_id' => $periodEntitlement->contact_id,
       'status_id' => 1,
-      'from_date' => CRM_Utils_Date::processDate('2016-06-29'),
+      'from_date' => CRM_Utils_Date::processDate('2016-06-07'),
       'from_date_type' => $this->leaveRequestDayTypes['all_day']['value'],
-      'to_date' => CRM_Utils_Date::processDate('2016-07-01'),
+      'to_date' => CRM_Utils_Date::processDate('2016-06-16'),
       'to_date_type' => $this->leaveRequestDayTypes['all_day']['value'],
       'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE
     ]);
```
davialexandre: I wonder if we should also add new test to ensure we can create a leave request overlapping two contracts as long as there's no lapse between them

davialexandre:
@tunbola There a few things that need to be fixed/updated on the PR description:
- For Leave Requests, we're not using alerts to display the error messages. We display them in a red bar, at the bottom of the modal. I know the ticket is using alert to refer to an error message, but I think this is confusing as Alert usually refers to an Alert window
- I think the biggest change here is the fact that we're returning the actual error messages instead of error codes. This should be mentioned on the before and after sections and not on the technical details. The overview should mention this as well

PR #1923, opened 2017-06-06
PCHR-2186: Closing balance is not updated on contact change

igorpavlov:
## Overview

When a manager records a new absence, selects a contact and chooses start and end dates, the 
opening balance, change balance and closing balance are calculated. If manager changes a contact with a different opening balance (in the same modal), the opening balance is updated, but the closing balance is not.

## Before

Opening balance is updated on contact change, but closing balance is not.

new

## After

Both opening balance and closing balance are updated on contact change.

new

## Technical Details

*Initial solution*

In initAbsenceType() the opening balance was updated, but the closing balance was not. Fixing this triggered the tests to fail. This was due to the logic flaw in test. The tests checked that closing balance is always zero if no dates are selected - it does not make sense though, as the closing balance should be equal to opening balance. This test was being passed in the past because the closing balance was never updated and thus was equal to 0. The logic in the test was fixed as well.

*After PR review*

Rebased with https://github.com/civicrm/civihr/pull/1926 initAbsenceType was changed to setInitialAbsenceTypes, applied suggestions as per PR feedback.

## Comments

- An empty line added before function definition for consistency
- Lint run on toil-request-ctrl.js and leave-request-ctrl_test.js
- Added a missing test for closeAlert()

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
```diff
@@ -484,7 +484,7 @@
             });
 
             it('resets balance and types', function () {
-              expect($ctrl.balance.closing).toEqual(0);
+              expect($ctrl.balance.closing).toEqual($ctrl.balance.opening);
```
deb1990: Please add a comment in the code explaining why `$ctrl.balance.closing` would be same as `$ctrl.balance.opening`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -177,8 +177,7 @@ define([
           .then(function (balanceChange) {
             if (balanceChange) {
               self.balance.change = balanceChange;
-              // the change is negative so adding it will actually subtract it
-              self.balance.closing = self.balance.opening + self.balance.change.amount;
+              self.calculateClosingBalanceChange();
```
deb1990: https://github.com/civicrm/civihr/blob/bb675fc47a017056d17d1683b8ad6144af653181/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/toil-request-ctrl.js#L34-L34

`toil-request-ctrl`overrides the `calculateBalanceChange` function, so `calculateClosingBalanceChange` needs to be used there as well.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -934,6 +942,7 @@ define([
 
         // Init the `balance` object based on the first absence type
         this.balance.opening = this.selectedAbsenceType.remainder;
+        this.calculateClosingBalanceChange();
```
deb1990: Also why are we calculating opening or closing balance inside `initAbsenceType` function. it does not make sense. 
After looking at the code it seems that we are calculating because when `changeInNoOfDays` is called, we also need to reassign balance values. So instead of only calculating that, we are calling `initAbsenceType` function which again does not make sense. 

So 
1. `initAbsenceType` call should be removed from `changeInNoOfDays` and only the code to recalculate should be added inside.
2. `this.balance.opening = this.selectedAbsenceType.remainder;` should be removed from `initAbsenceType` and the calculation for both opening and closing balance should go to the new function which you created `calculateOpeningAndClosingBalance`
3. `calculateOpeningAndClosingBalance` should be called from `initAfterContactSelection`, `changeInNoOfDays` and `calculateBalanceChange`

I know all of this is not part of the ticket, but if we find some code which can be improved ,we always should refactor it.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: calculateClosingBalanceChange() was written in order to calculate closing balance based on opening balance and balance change. The initial thought was to have this function to use it in 2 places:
1) inside the calculateBalanceChange()
2) outside the calculateBalanceChange() when opening balance is changed and no other calculations needed to be done apart from calculating closing balance.

However, I removed calculateClosingBalanceChange() after all. I now think it would be safer to re-calculate all balances if you change the contact, because if we extend the logic in calculateBalanceChange() in the future there is a risk of having a similar bug again.

igorpavlov: calculateClosingBalanceChange removed, so no need to use it now.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -187,6 +186,14 @@ define([
       };
 
       /**
+       * Calculates and updates closing balance
+       */
+      this.calculateClosingBalanceChange = function () {
```
igorpavlov: calculateClosingBalanceChange removed

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrl_test.js
igorpavlov: done

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: Done, but with a _calculateClosingBalance()

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -140,6 +140,8 @@ define([
         this._reset();
         // reinitialize opening balance
         setInitialAbsenceTypes.call(this);
+
+        this._calculateOpeningAndClosingBalance();
```
deb1990: As I mentioned in my earlier comment, `setInitialAbsenceTypes` (previously named as `initAbsenceType`) should be removed from `changeInNoOfDays`. Because in the UI Single Day or Multiple Day is clicked the absence type should not be set again, only the balance should be calculated as the No of days changed.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -188,6 +189,15 @@ define([
       };
 
       /**
+       * Calculates and updates closing balance
```
deb1990: `Calculates and updates opening and closing balance`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -709,6 +719,8 @@ define([
             }
 
             self.postContactSelection = false;
+
```
deb1990: Here line break is not necessary.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -188,6 +189,15 @@ define([
       };
 
       /**
+       * Calculates and updates closing balance
+       */
+      this._calculateOpeningAndClosingBalance = function () {
```
deb1990: Please define the functions alphabetically , as it has `_` it should be down.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -709,6 +719,8 @@ define([
             }
 
             self.postContactSelection = false;
+
+            self._calculateOpeningAndClosingBalance();
```
deb1990: The position of calculating balance has changed, its not breaking anything right? (Double checking)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: The order looks good to me I think, however, your question triggered a thought that we want to use the whole calculateBalanceChange() and not just _calculateOpeningAndClosingBalance(). This is safer as the balance.change may also vary depending on a contact.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -227,6 +211,18 @@ define([
       };
 
       /**
+       * Validates and creates the leave request
+       *
+       * @returns {Promise}
+       */
+      function createRequest () {
```
AkA84: this should not be here, it's in the middle of the public methods, while instead it's a private function

PR #1924, opened 2017-06-06
PCHR-2154: Fix FTE Calculation on Import of Full Time Hours Type

MiyaNoctem:
## Overview
Importing contracts with 'Full Time' hours type was rendering blank FTE's, that were able to be fixed by editing the contract and saving with no modifications.

## Before
When importing contracts with Full Time Hours Type, Actual Hours amount is not and should not be given, as it should be calculated from the selected hours location instead.  However, when importing, hours amount was being left blank, causing FTE to be blank also.

## After
Added a check for hours type when calculating automatic values, so that when 'Full Time' hours type is selected for the contract being imported, the value of the number of hours is copied from the standard hours configured in the hours/location.

## Technical Details
Added a test that asserts when 'Full Time' hours type is used, the value of hours_amount is equal to the value of standard_hours.

---

- [x] Tests Pass

### DIFF
--- a/hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
+++ b/hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
@@ -265,6 +265,27 @@ class CRM_Hrjobcontract_Import_Parser_ApiTest extends CiviUnitTestCase implement
     $this->validateHourAutoFields($contact3['id'], $expected);
   }
 
+  public function testActualHoursIsDeducedForFullTimeHoursType() {
+    $contact1 = ContactFabricator::fabricate();
+    $contract = $this->buildContractInfo([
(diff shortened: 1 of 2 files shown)


hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -265,6 +265,27 @@ public function testFTEFieldsAreSetToZeroWhenImportingOnlyMandatoryFields() {
     $this->validateHourAutoFields($contact3['id'], $expected);
   }
 
+  public function testActualHoursIsDeducedForFullTimeHoursType() {
```
davialexandre: Just a comment: I think that the work "deduced" doesn't make sense here. For me, deduce is more about logical reasoning (when you deduce something based on some facts/details), which is different of what actually happens here, where you are actually copying the value from one field to others

PR #1925, opened 2017-06-07
PCHR-2215: Back button on Manage Leave Entitlements page doesn't redirect to the contract

igorpavlov:
## Overview

Once you create a contract you are redirected to Manage Leave Entitlements page. This page has a back button and it redirected to Contact Summary tab instead of Job Contract tab.

## Before

Back button redirected to Contact Summary tab.

new

## After

Back button redirects to Job Contract tab.

new

## Technical Details

Back button's href was changed from javascript:history.back() to {$smarty.get.returnUrl}. "returnUrl" is simply a get query param that we already have in the URL on the Manage Leave Entitlements page. We use $smarty instead of $_GET['returnUrl'] because $_GET will simply not work.

## Comments

Since this is a tpl file that was changed, there are no tests to pass.

---

- [ ] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.tpl
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.tpl
@@ -98,7 +98,7 @@
   </tbody>
 </table>
 <div class="crm-submit-buttons">
-  <a href="{$smarty.get.returnUrl}" class="button"><span>{ts}Back{/ts}</span></a>
+  <a href="{$returnUrl}" class="button"><span>{ts}Back{/ts}</span></a>
   {include file="CRM/common/formButtons.tpl" location="bottom"}
 </div>
 <div id="add-comment-dialog" title="{ts}Add/Edit comment{/ts}">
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.tpl
```diff
@@ -98,7 +98,7 @@
   </tbody>
 </table>
 <div class="crm-submit-buttons">
-  <a href="javascript:history.back()" class="button"><span>{ts}Back{/ts}</span></a>
+  <a href="{$smarty.get.returnUrl}" class="button"><span>{ts}Back{/ts}</span></a>
```
davialexandre: Ideally, it's a bad practice for templates to access the global scope. Any value required by the template must be passed in to it by the controller. In this case, the controller is `uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php`. To pass a variable to a template, you can use the `assign` method, inside the `buildQuickForm` method of the controller.

igorpavlov: Done!

PR #1926, opened 2017-06-07
PCHR-2227: TOIL not showing up in Record New Absence Popup

deb1990:
## Problem
The absence types which does allow to have Accrual Requests(TOIL) , are not visible in the Record New Absence popup's Dropdown.
my_leave___hr16-4

## Solution
1. in `leave-request-ctrl.js` `allow_accruals_request` was set to false, which means the popup will not display such absence types in the dropdown. It was wrong so removed it.

2. Code has been refactored, 

- `self` replaced by `bind`
- `initAbsenceType()` renamed to more meaningful `setInitialAbsenceTypes()`
- `canViewOrEdit()` removed as it is not required, and in most cases `!this.isMode('create')` is used instead of `canViewOrEdit()`, as both means same.

**Screenshot**
my_leave___hr16-4

## Comments
Still there is a issue when we first select TOIL and then click on Single Day, the dropdown values is being reset to Holiday. This is because `changeInNoOfDays()` function internally calls `setInitialAbsenceTypes()` which resets the absence type to the first value of the `absenceType` array. Now this wrong, because when `changeInNoOfDays()` is called only the balance calculation should be done, nothing else. As @igorpavlov 's PR (https://github.com/civicrm/civihr/pull/1923) also touched the same thing and same fix is required there too, I left the comment to fix this in his PR.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
@@ -133,13 +133,13 @@ define([
       };
 
       /**
-       * Change handler for change request type like multiple or single. It will
-       * reset dates, day types, change balance.
+       * Change handler when changing no. of days like Multiple Days or Single Day.
+       * It will reset dates, day types, change balance.
        */
       this.changeInNoOfDays = function () {
         this._reset();
         // reinitialize opening balance
-        initAbsenceType.call(this);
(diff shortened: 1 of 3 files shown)


igorpavlov:
- The problem should described in a non-technical language
- There should be a screenshot for both before and after the fix
- The technical solution should be in a separate section underneath before/after

deb1990:
> The problem should described in a non-technical language
There should be a screenshot for both before and after the fix
The technical solution should be in a separate section underneath before/after

Added screenshot for Before
For the rest of the stuff its fine as discussed in the L&A room

PR #1928, opened 2017-06-07
PCHR-2236: Dont display decimals for contracts entitlements

deb1990:
## Problem
While creating new contract, in the Leave tab, if the input boxes for Entitlement are clicked and then deselected, numbers are shown as floating point even though the number is an integer. Floating points should not be shown until the number is a float.
claudio_adams_jr____hr16-4

## Solution
Added `hrjc-number-float="true"` in the HTML which takes care of this scenario.

### DIFF
--- a/hrjobcontract/views/modalForm.html
+++ b/hrjobcontract/views/modalForm.html
@@ -588,6 +588,7 @@
                                name="leave-{{leaveEntry.leave_type}}"
                                class="form-control"
                                hrjc-number="2"
+                               hrjc-number-float="true"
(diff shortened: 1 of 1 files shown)


deb1990:
> Could you please re-record the after-fix GIF to show different inputs in all fields (floats and integers) to confirm that all fields are working properly now?

The input box is repeated using a `ng-repeat`so all the input boxes will behave in the same way. Is that fine?

igorpavlov:
Yep, true.

PR #1737, opened 2017-04-21
PCHR-2160: Add wrapper service to check permissions

dhavalsharma:
## Overview
This PR creates a new service to wrap check if currently logged in user is admin or not for now. 

In future other endpoints can be added to enhance the permissions check.

## Technical Details
Creates a new angular service to wrap calls to CRM.checkPerm.

## Comments
None

---

- [X ] Tests Pass

PR #1938, opened 2017-06-12
PCHR-2312: Improve checkPermission service

AkA84:
## Problem
The CheckPermissions service written in #1737 has two major flaws

1. The signature doesn't make sense. It has a single `canAdmin` method which, despite the very generic name (which in itself is very hard to understand), then proceeds to check if the currently logged in user has a very specific permission: to administer L&A
2. It has the "CiviHRLeaveAndAbsences: Administer Leave and Absences" permission hardcoded in the method. Meaning that the service must be "aware" of any L&A related permissions, breaking the separation of concerns principle: the L&A app should be the one concerned with being aware of what permissions are and are not present for L&A, not a generic service

## Solution
The service is now a simple function that encapsulates the `CRM.checkPerm` method and, given a permission or a list of permissions, returns a promise (in case in the future the permission check will be done asynchronously)
```js
checkPermission('permission the user has').then(function (outcome) {
  outcome; // true
});
```

If given multiple permissions, an "AND" logic is applied to the permission checks
```js
checkPermission(['permission the user has', 'permission the user does not have']).then(function (outcome) {
  outcome; // false
});
```

## Technical Details
Nothing particular, i renamed the file from `check-perm` to `check-permissions` to make the file name match with the service name

- [x] Tests Pass


org.civicrm.reqangular/test/services/check-permission_test.js
```diff
@@ -0,0 +1,78 @@
...
+        it('returns a promise', function () {
+          expect(promise).toEqual(jasmine.objectContaining({
+            then: jasmine.any(Function),
+            catch: jasmine.any(Function),
+            finally: jasmine.any(Function)
+          }));
+        });
+
+        it('utilizes the the `CRM.checkPerm` method', function () {
+          expect(CRM.checkPerm).toHaveBeenCalledWith(permissionToCheck);
```
AkA84: @bmnepali @igorpavlov @deb1990 an example of a unit test testing the actual implementation of a method. Usually done to put the emphasis not only on the *what*, but also on an important part of the *how* (which is considered so crucial to deserve to be explicitly tested. For example making sure that a Model service calls the expected API service)

AkA84:
closing this PR has the service should be added to CiviHR 1.6 and then synced in 1.7

PR #1940, opened 2017-06-12
PCHR-1687: Hotfix to fix the absolute URL problem of the Custom CSS URL

davialexandre:
## Overview

On https://github.com/civicrm/civihr/pull/1330, we started using an upgrader to set the Custom CSS URL instead of the old bash script. The problem however, is that with this change we ended up dropping the use of `[civicrm.root]` and started using an absolute URL. This can cause problems on sites using HTTPS and where the absolute URL was generated having the HTTP protocol hardcoded.

## Before
The upgrader method will set the Custom CSS URL with an absolute URL

## After
The upgrader method will set the Custom CSS using a relative path, including the `[civicrm.root]` placeholder.

---

- [ ] Tests Pass
No existing tests related to this


mickadoo:
Why are there 15 commits in this?

mickadoo:
@davecivicrm tested by first running all upgraders locally from staging branch. 

Inside `/civicrm/admin/setting/url?reset=1` the custom CSS url was: http://civihr.local/tools/extensions/org.civicrm.shoreditch/css/custom-civicrm.css

Then I checked out this branch and ran upgraders again. After upgrade the custom CSS url was:
[civicrm.root]/tools/extensions/org.civicrm.shoreditch/css/custom-civicrm.css

davialexandre:
@mickadoo This is a hotfix and it's based on master. These are all merge commits from master that don't exist on staging (from when we merge staging into master). It's ok to have them, since they don't introduce any code change. There's also this PR against master, which, as expected, doesn't include the merge commits https://github.com/civicrm/civihr/pull/1943

davialexandre:
Since @mickadoo tested this and it passed testing, I'm going to merge it

PR #1947, opened 2017-06-13
PCHR-2011: Work Pattern HTML+CSS

deb1990:
## Overview
As Part of this PR the Work Pattern's Tab sections is designed.

## After
santina_muller___hr16-4
Both Custom and Default work patterns are shown statically, but when the content will be dynamic either of the two will be displayed based on work pattern.

## Technical Details
1. HTML Markup Added
2. SCSS added( also in https://github.com/compucorp/org.civicrm.shoreditch/pull/16)

## Link To Mock
https://projects.invisionapp.com/share/WX71LBMPJ#/screens/161975305
https://projects.invisionapp.com/share/WX71LBMPJ#/screens/171887407


deb1990:
> Could you please either describe with text what exactly has been added in After section, or provide Before section with a screenshot. That would help to understand the changes made to UI quicker.

@igorpavlov I have added red border around what has been developed as part of this PR, is this fine?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/pages/_absence-tab-page.scss
```diff
@@ -7,4 +7,8 @@
   .tab-pane {
     padding: 0 !important;
```
igorpavlov: Is "!important" necessary? Does that mean that the bootstrap theme is applied afterwards and without important it will not work? Is there a way to avoid "!important" in this case?

uk.co.compucorp.civicrm.hrleaveandabsences/views/absence-tab/components/absence-tab-work-patterns.html
```diff
@@ -1 +1,56 @@
...
+  </tr>
+  </thead>
+  <tbody>
+  <tr>
+    <td>Compressed hours 40h - Mon to Thurs</td>
+  </tr>
+  </tbody>
+</table>
+
+<br><br>
```
igorpavlov: Is there a need to do <br> here? Is it possible to rather use margins in CSS instead?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/pages/_absence-tab-page.scss
deb1990: Yes by default, the tab-pane is supposed to have the padding, but for absence tab page the padding is necessary. As this requirement is only for absence tab, we should not change original theme, instead we should just override the theme for absence tab pages. 

Also `!important` is necessary here because otherwise it wont work because of specificity rules. And its not a problem because this style is only meant for absence tab and not going to be reused. Also if we want to avoid `!important` we would need to add more classes or IDs which would be overkill and is not really necessary.

uk.co.compucorp.civicrm.hrleaveandabsences/views/absence-tab/components/absence-tab-work-patterns.html
deb1990: As you can see from the mocks that at a time either Custom or Default work patterns will be shown, so when we make this template dynamic we wont need the `<br>` anymore. This is just temporary to keep some spacing between custom and Default work patterns block.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/pages/_absence-tab-page.scss
igorpavlov: OK, that's fair enough if we are sure the style won't be reused.

PR #1955, opened 2017-06-16
PCHR-2035: WorkPattern Modal HTML

deb1990:
## Overview
As part of this PR the HTML for Work Pattern Modal has been developed.
CSS changes are part of https://github.com/compucorp/org.civicrm.shoreditch/pull/19 & https://github.com/compucorp/civihr-employee-portal-theme/pull/173

## After
civihr_staff_compucorp_co_uk___hr17-1

## Technical Details
1. `absence-tab-custom-work-pattern-modal.html` created which contains markup for the tab.
2. `vm.openModal` is a temporary function which is used to open the modal in angular way, test cases not updated.

## Comments
Link to mock - https://projects.invisionapp.com/share/WX71LBMPJ#/screens/161975298

Should only be merged once, https://github.com/compucorp/org.civicrm.shoreditch/pull/19 is merged and 1.7 is synced.


uk.co.compucorp.civicrm.hrleaveandabsences/views/absence-tab/components/absence-tab-work-patterns.html
```diff
@@ -54,3 +54,73 @@
   </tr>
   </tbody>
 </table>
+
+
+<!-- Modal -->
+<div class="modal fade" id="workPatternModal" tabindex="-1" role="dialog">
```
AkA84: The modal is static at this stage, but even if it's static we should be able to open it by clicking a button in the  work pattern page, instead of having it in the tab's view file.

Please move this to a separate view file and add a simply method to open the modal in the [absence-work-patterns-tab](https://github.com/civicrm/civihr/blob/6e2eda4fce2234f19fcf5e59f470302922293977/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/components/absence-tab-work-patterns.js)

AkA84:
@deb1990 shouldn't you add a backstopjs test, this is a new UI component

deb1990:
> @deb1990 shouldn't you add a backstopjs test, this is a new UI component
First of all I agree that I forgot about it, but I think we should do it once the total implementation is done. Because function names etc can change.
I am using openModal to open the modal, which might be useful to write the selector, so we might  change it. Also I am not sure what other chnges can happen, as we have a ticket to complete the component, I believe its safe to do there.

PR #1956, opened 2017-06-16
PCHR-2027: Augment LeaveRequestAPI and LeaveRequestInstance with delete method

AkA84:
## Overview
This PR introduces the `delete()` method for the `LeaveRequestAPI` and amended the already existing one for `LeaveRequestInstance`

For the API, it calls the `LeaveRequest.delete` endpoint.
For the Instance, it calls the API by passing automatically its `id` property

## Technical Details
1. Refactored the `LeaveRequestAPI` (mainly rearranging the methods in alphabetical order
2. Updated the mocked LeaveRequestAPI and mocked data by adding the new method and the response that you would get from the real endpoint
3. The original `LeaveRequestInstance.delete()` method used the `update` endpoint, which was wrong. Hence the switch to the `delete` endpoint
4. Massively refactored the tests for both the api and the instance (giving more meaningful names to test, removing unnecessary tests, fixing badly set up tests, refactored the code in general)

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
```diff
@@ -304,6 +304,26 @@ define([
       });
     });
 
+    describe('delete()', function () {
+      var idToDelete = '123';
+
+      beforeEach(function () {
+        spyOn(LeaveRequestAPI, 'sendGET');
```
deb1990: When we test `API`, we should spy the GET call using `$httpBackend.whenGET` and return values from mocked data.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
```diff
@@ -605,119 +541,61 @@ define([
...
           });
         });
       });
 
       describe('deleteAttachment()', function () {
         beforeEach(function () {
-          promise.then(function () {
-            LeaveRequestInstance.deleteAttachment(attachments[0]);
-          });
+          $rootScope.$apply();
```
deb1990: Should not this be in `afterEach`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -227,14 +227,12 @@ define([
...
+         * Deletes the leave request
          *
-         * @return {Promise} Resolved with {Object} Deleted Leave request
+         * @return {Promise}
          */
         delete: function () {
-          this.is_deleted = true;
-
-          return LeaveRequestAPI.update(this.toAPI());
+          return LeaveRequestAPI.delete(this.id);
```
deb1990: Previously we had `is_deleted` key to know if a leave request is deleted, how do we know the same now? Because when we delete a leave request, front end also need to remove the same from the list. Right now the only option for that seems to be refreshing the whole list.But that does not sound very efficient. 

Also when the delete method is called the record is deleted from Backend, but still we are keeping the object for the same instance in the frontend. This again does not look right.

I do not have a proper answer to this, please let me know what plan do you have in mind to deal with this situation.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js
AkA84: only if we want to mock the returned data. In this case I don't care about what the server returns, I only want to check if a given method has been called and with what parameter. Hence `spyOn` is enough

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/instances/leave-request-instance_test.js
AkA84: no, this triggers the `loadAttachments()` in the `beforeEach` above. This would be the sequence
```
loadAttachments()
$apply() // triggers loadAttachments()
deleteAttachment()
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
AkA84: As discussed with @davialexandre , the `is_deleted` flag will not be exposed to the frontend, being an internal backend implementation (the backend automatically filters the soft-deleted leave requests when returning their list, for example).

So basically the frontend will not know that there is a soft-delete at all, it will just think that the requests get deleted, period.

So in the future (I can't see any current usage of the `is_deleted` flag in the whole L&A app), the angular app will simply remove the deleted request from the list, i guess via emitting an event, like
```js
// deleting the request
.then(function () {
  $rootScope.$emit('LeaveRequest::deleted', deletedRequest)
});

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js (hunk 1 above)
Comment (deb1990) [inline uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js]: How is it different from other methods like `all()` and others. Or do you mean we should not test any api calls which does not change the data before sending it back to the callee.

Comment (AkA84) [inline uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/leave-request-api_test.js]: Yep exactly. If you look at the tests that i simply removed, those were tests that were just checking the structure of the endpoint response. But it doesn't make sense to test for that, it's not the responsibility of the API service to define the structure of the response.
```js
// example
it('has the expected structure', function () {
  promise.then(response) {
    // the structure does not depend on the api service itself, what's the point of testing it?
    expect(response.foo).toBeDefine();
    expect(response.bar).toBeDefine();
  });
});
```
instead, if an api perform some transformation on the endpoint response before returning the value, then that is something worth testing
```js
it('has the expected structure', function () {
  promise.then(response) {
    // the structure does not depend on the api service itself, what's the point of testing this?
    expect(response.someCustomField).toBeDefined();
    expect(response.someCustomField).toBe(/* some expected value */);
  });
});
```

PR #1960, opened 2017-06-22
PCHR-2012: Augment WorkPattern API

deb1990:
## Overview
As part of this PR the following methods are added to WorkPattern API.
- `assignWorkPattern ()` - assigns a work pattern to a contact
- `get ()` - returns all the work patterns
- `workPatternsOf ()` - returns all the work patterns of a specific contact

## Technical Details
1. Mocked data for the newly created methods are added to `work-pattern-data.js`.

2. Usage of the functions

**assignWorkPattern()**
```javascript
WorkPatternAPI.assignWorkPattern(<contactId>, <workPatternID>, <effectiveDate>, <effectiveEndDate>, <changeReason>, <params>);
```

**get()**
```javascript
WorkPatternAPI.get(<params>);
```

**workPatternsOf()**
```javascript
WorkPatternAPI.workPatternsOf(< contactId >, <params>);
```

## Comments
1. `assignWorkPattern ()` - There is a change in implementation from Proof of concept, `contactId, workPatternID, effectiveDate, effectiveEndDate, changeReason` needs to be explicitly passed to this function by the callee as these parameters are directly related while assigning a work pattern.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
```diff
@@ -13,7 +13,43 @@ define([
...
+       *
+       * @param {string} contactId
+       * @param {string} workPatternID
+       * @param {string} effectiveDate
+       * @param {string} effectiveEndDate
+       * @param {string} changeReason
+       * @param {object} params - additional parameters
+       * @return {Promise} Resolved with {Array} All Work Patterns
+       */
+      assignWorkPattern: function (contactId, workPatternID, effectiveDate, effectiveEndDate, changeReason, params) {
```
igorpavlov: Despite of the fact, that the explanation is provided in the PR comment section, the implementation does not satisfy the requirements of the task provided in its description. There could be a reason  assignWorkPattern accepts only one parameter. This should to be double-checked with the task assigner.

deb1990: I dont see any particular reason of not having multiple parameters, because we do the same for other similar API calls. For example https://github.com/civicrm/civihr/blob/57a3415ee63faa8632c31a1fd2028a389679d54c/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js#L91-L91

@AkA84 could you look into this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
```diff
@@ -13,7 +13,43 @@ define([
...
+      },
+
+      /**
+       * Returns all the work patterns
+       *
+       * @param {object} params additional parameters
+       * @return {Promise} Resolved with {Array} All Work Patterns
+       */
+      get: function (params) {
+        return this.sendGET('WorkPattern', 'get', params || {})
```
igorpavlov: Please use _.assign here as well for consistency.

deb1990: `_.assign` is necessary for other API calls because `params`and the other parameters needs to be merged into 1 object, but in this case there is no other parameters as the function has a single param, so there is no need to use `_.assign`

igorpavlov: Agreed.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/work-pattern-api.js
davialexandre: @deb1990 Are you sure this link is correct? It points to a method that accepts only one parameter. I don't see how as an argument in favor of having multiple parameters.

I think that the problem here is that we have both specific params (workPatternID, effectiveDate, etc) and one generic `params` that can hold everything, which gives the impression that something like this is possible:

```javascript
WorkPatternAPI.assignWorkPattern(null, null, null, null, null, {
  'contact_id': 1,
  'pattern_id': 2,
  'effective_date': '2017-01-01',
 // ...
})
```

To avoid confusions like that, we should either have only specific params or a single generic one.

AkA84: @deb1990 i think in the end it's a matter of preferences. I personally like just one `param` parameter because when you call the method everything is very explicit and simple to understand, and also because you don't need to have a separate param for the "other" params (the optional ones)
```js
assignWorkPattern({
  contactId: 2,
  workPatternId: 3,
  effectiveDate: '2017-01-01',
  endDate: '2017-02-02',
  changeReason: 'foobar',
  someAdditionalParam: 'barbaz'
});
```
compare to
```js
assignWorkPattern(2, 3, '2017-01-01', '2017-02-02', 'foobar',  { someAdditionalParam: 'barbaz' });
```
in the second example, you have to remember exactly what each parameter means in the method's signature, which is not convenient nor expressive with methods with long signatures


Anyway although i prefer the first version, I'm fine with both. Just my 2 cents

deb1990: @AkA84 @davialexandre Yes wrong example sorry, here is the correct one https://github.com/civicrm/civihr/blob/57a3415ee63faa8632c31a1fd2028a389679d54c/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/apis/leave-request-api.js#L153

I think that specifying the unique parameters for this API explicitly is better because then its easier to just refer the API(JS Code) documentation to see which parameters I need to send to assign a work pattern, otherwise we would need to go to API explorer everytime and find out what the required parameters are, also we need to get their names.

PR #1963, opened 2017-06-23
PCHR-2134: Allow a LeaveRequest with half day to use all the remaining balance

davialexandre:
### Overview
In a scenario where you cannot use more than your remaining balance, and it also has half days (say 1.5), if you request a Leave Request with the same amount, the validation would fail, saying you cannot use more than the remaining balance

### Before

During the Leave Request validation, we pass the `$params` array to the `calculateBalanceChangeFromParams` method, which would then convert the `from_date_type` and `to_date_type` values to option values arrays (containing labels, values and names). These arrays would be passed to the `calculateBalanceChange` method, where the calculation actually happens. This method expects the date types as values though, not as arrays. Whenever it get the values in an unexpected format, it would consider it to be a full day. So, when `calculateBalanceChangeFromParams` passed a half day to it, it would be considered as a full day, resulting in the wrong balance change.

## After
Now `calculateBalanceChangeFromParams` doesn't convert the date types to arrays anymore, making `calculateBalanceChange` receive the data in the expected format and calculate the amount properly

## Technical Details
The tests never caught this before because this is a very specific scenario and most of them always use full days. A new test was added to cover this scenario.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -482,17 +482,12 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @return float
    */
   private static function calculateBalanceChangeFromCreateParams($params) {
-    $leaveRequestOptionsValue = self::getLeaveRequestDayTypeOptionsGroupByValue();
-
-    $fromDateType = $leaveRequestOptionsValue[$params['from_date_type']];
-    $toDateType = $leaveRequestOptionsValue[$params['to_date_type']];
-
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -482,17 +482,12 @@ private static function validateWorkingDayAndBalanceChange($params, $absenceType
    * @return float
    */
   private static function calculateBalanceChangeFromCreateParams($params) {
-    $leaveRequestOptionsValue = self::getLeaveRequestDayTypeOptionsGroupByValue();
```
omarabuhussein: you are not using getLeaveRequestDayTypeOptionsGroupByValue method any more , shouldn't you remove it ?

davialexandre: good catch!

PR #1965, opened 2017-06-26
PCHR-2004: Implement leave report component in absence tab

AkA84:
## Overview
This PR adds the leave report component to the contact summary page of any given contact, making possible for a CiviHR admin to see the leave requests of a staff member and perform actions on them

report-tab

The admin has access to a "delete" shortcut for each leave request

quick-delete

and he can also delete a request via the modal

delete-via-modal

### Renamed and moved location of `my-leave-report`
The `my-leave-report` component is now called `staff-leave-report` and has been moved in the _shared/_ folder

### Pushed L&A permissions also in the admin
Given that the component now uses the `checkPermission` service (see below), the L&A permissions need to be present in the global `CRM.permissions`, when one is browsing the absence tab. In order to do that, just as it's done for the SSP, the permissions are getting pushed in the property via the backend
```php
/**
 * Implementation of hook_civicrm_pageRun
 */
class CRM_HRLeaveAndAbsences_Page_AbsenceTab extends  CRM_Core_Page {
  // ...
  CRM_Core_Resources::singleton()->addPermissions([
    'access leave and absences',
    'administer leave and absences',
    'access leave and absences in ssp',
    'manage leave and absences in ssp',
  ]);
}
```

### Altered the action matrix if the user is an admin
The action matrix available for a leave request is different depending on whether a user is a staff member or an admin (an admin has access to the "delete" action, "edit" and "respond" verbs are swapped, etc).

The component now checks whether the current user is an admin via the `checkPermissions` service
```js
// Init block
(function init () {
  checkPermissions(sharedSettings.permissions.admin.administer)
  .then(function (isAdmin) {
    role = isAdmin ? 'admin' : role;
  })
```
and based on the `role` property, changes the output of the `actionsFor` method
```js
vm.actionsFor = function (leaveRequest) {
  if (role === 'admin') {
    // ...
  } else {
    // ...
  }

  return actions;
};
```
### Removed `canCancel()`
There was a `canCancel()` method on the scope, that was used, in the view, to alter the list of actions rendered on the page
```html
<li ng-repeat="action in report.actionsFor(request)">
  <a ng-if="action === 'cancel' && report.canCancel(request)" href
    ng-click="report.action(request, action)">
    {{action}}
  </a>
```
the method has now been removed and its logic integrated into the `actionsFor()` method, so that we can be sure that whatever list the latter returns, is the "final" list that doesn't need additional manipulation
```js
vm.actionsFor = function (leaveRequest) {
  // ...
  if (!canLeaveRequestBeCancelled(leaveRequest)) {
    actions = // ...
  }
  // ...
  return actions;
}
```
### "cancel" action is always allowed for admin users
The logic of the `canCancel()` method (see above) was flawed, as it was applied also for admin users. Admins instead can always cancel a leave request, regardless of any logic that normal users are subject to. This had been fixed
```js
function canLeaveRequestBeCancelled (leaveRequest) {
  // ...
  if (role === 'admin') {
    return true;
  }
}
```
### Refactored BackstopJS tests
The BackstopJS test for L&A had been refactored in several ways. From bundling all the scenarios in one JSON file, to having page objects extending the correct parents, to decoupling the page objects from they're own init-click event, etc

### Improve test performance
The L&A JS unit-test suite was getting pretty slow, getting ~14sec to run fully.

The reason for it was that we were compile the components' markup for each and every test, which obviously would take a toll on the performance
```js
function compileComponent () {
  var $scope = $rootScope.$new();
  var contactId = '202';

  component = angular.element('<absence-tab-entitlements contact-id="' + contactId + '"></absence-tab-entitlements>');
  $compile(component)($scope); // compile the markup
  $scope.$digest();

  controller = component.controller('absenceTabEntitlements');
}
```
This was not only expensive, but also pretty useless as we really didn't test the markup (except for a couple of tests that weren't really that useful anyway), but just the component's controller's logic.

Using the [`$componentController`](https://docs.angularjs.org/guide/component#unit-testing-component-controllers) service to instantiate just the component's controller speeds up things considerably as it allows us to skip the markup compile phase, and reduce the total run time to ~4s
```js
function compileComponent() {
  controller = $componentController('absenceTabEntitlements', null, { contactId: '202' });
  $rootScope.$digest();
}
```
#### Before
before

#### After
after

### Imported component's style in L&A style
The component style had been ported from the SSP theme (see https://github.com/compucorp/civihr-employee-portal-theme/pull/180) to the L&A style. Several tweaks and refactoring were performed, but in the end we're still talking about the components we introduced in this PR https://github.com/compucorp/civihr-employee-portal-theme/pull/119

### Implemented workaround for actions dropdown outside of Shoreditch namespace
The actions dropdown is attached to the `<body>` element to avoid overflowing issues if attached inside the table that it would belong to

overflow-issue

_(example of overflow issue)_

The problem of attaching it to the `<body>` though, is that the `<body>` doesn't have the Shoreditch namespace (`#bootstrap-theme`) applied to it, and as such any bootstrap-based elements within it would not get styled

no-style

_(example of the dropdown not getting styled when outside the namespace)_

Trying to attach the dropdown to the component's root (which _does_ have the namespace applied to it), would result in a offset miscalculation by the ui-bootstrap angular module

offset-miscalculation

_(example of offset miscalculation)_

So the only solution that could be found was to simply allow the dropdown styles to also be present outside of the `#bootstrap-theme` namespace. Given that this is only for the L&A extension's style, the workaround impact is minimal
```scss
// leaveandabsences.scss

// The bootstrap dropdown is attached to the <body> outside of the #bootstrap-theme
// namespace. Hence unfortunately the styles for the dropdowns need
// to be available out of the namespace as well
@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/mixins/ALL";
@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/_dropdowns.scss";
@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/overrides/style/_dropdowns.scss";
```

## Comments
I had intended to create a `<leave-request-actions></leave-request-actions>` component to encapsulate the markup and logic of the leave request actions list, but unfortunately due to time constraints I couldn't manage to do it. Next time.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-report.html
```diff
@@ -384,14 +387,17 @@
                                 <i class="fa fa-ellipsis-v"></i>
                               </a>
                               <ul class="dropdown-menu dropdown-menu-right" uib-dropdown-menu>
-                                <li ng-repeat="action in report.actionsFor(request)">
-                                  <a ng-if="action === 'cancel' && report.canCancel(request)" href
-                                    ng-click="report.action(request, action)">
+                                <li ng-repeat="action in ::report.actionsFor(request)"
+                                  ng-switch="['edit', 'respond', 'view'].indexOf(action) > -1">
+                                  <a href
+                                    ng-switch-when="true"
```
deb1990: Cant we just use ng-show and ng-show here?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/_variables.scss
```diff
@@ -0,0 +1,7 @@
+@import "SCSSROOT/org.civicrm.shoreditch/scss/bootstrap/overrides/variables";
+@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/variables";
+
+$chr-table-condensed-cell-padding: 8px 5px;
```
deb1990: I think these variables are used under the tables partials only, then we should move this to that partial only.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/mixins/_tables.scss
```diff
@@ -0,0 +1,97 @@
...
+
+        > td {
+          border-color: $crm-table-heading-border-color;
+        }
+      }
+    }
+  }
+}
+
+
```
deb1990: There should be a single line break.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/leaveandabsence.scss
```diff
@@ -1,9 +1,16 @@
...
+#bootstrap-theme {
+  @import "overrides/style/*";
+  @import "components/*";
+  @import "pages/*";
+}
+
+// The bootstrap dropdown is attached to the <body> outside of the #bootstrap-theme
+// namespace. Hence unfortunately the styles for the dropdowns need
+// to be available out of the namespace as well
+@import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/mixins/ALL";
```
deb1990: As we discussed before, instead of adding all the styles to the global namespace, cant we just create another namespace like `chr_dropdown`(or something else) and nest the styles?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.js
```diff
@@ -243,6 +215,31 @@ define([
...
+     *
+     * Based on following constants
+     * REQUEST_CANCELATION_NO = 1;
+     * REQUEST_CANCELATION_ALWAYS = 2;
+     * REQUEST_CANCELATION_IN_ADVANCE_OF_START_DATE = 3;
+     *
+     * @param  {LeaveRequestInstance} leaveRequest
+     * @return {Boolean}
+     */
+    function canLeaveRequestBeCancelled (leaveRequest) {
```
deb1990: Please keep alphabetical order.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.js
```diff
@@ -2,35 +2,37 @@
...
-    actionMatrix[sharedSettings.statusNames.cancelled] = ['view'];
-    actionMatrix[sharedSettings.statusNames.rejected] = ['view'];
+    actionMatrix[sharedSettings.statusNames.awaitingApproval] = ['edit', 'cancel', 'delete'];
+    actionMatrix[sharedSettings.statusNames.moreInformationRequired] = ['respond', 'cancel', 'delete'];
+    actionMatrix[sharedSettings.statusNames.approved] = ['view', 'cancel', 'delete'];
+    actionMatrix[sharedSettings.statusNames.cancelled] = ['view', 'delete'];
+    actionMatrix[sharedSettings.statusNames.rejected] = ['view', 'delete'];
+
     var requestSort = 'from_date DESC';
+    var role = 'staff';
```
AkA84: i've explained it in the technical details

uk.co.compucorp.civicrm.hrleaveandabsences/scss/leaveandabsence.scss
AkA84: I looked into it (https://angular-ui.github.io/bootstrap/#!#dropdown), you would have to add some configuration settings to append a custom class. Honestly it doesn't seem it's worth it

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.js
```diff
@@ -156,11 +159,17 @@ define([
 
     // Init block
     (function init () {
-      $q.all([
-        loadStatuses(),
-        loadAbsenceTypes(),
-        loadAbsencePeriods()
-      ])
+      checkPermissions(sharedSettings.permissions.admin.administer)
+      .then(function (isAdmin) {
```
AkA84: Can you show me how i would use that?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-report.html
AkA84: why? `ng-switch` seems a perfectly elegant solution to me in this case (and i think we use it in manager reports as well)

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -427,6 +427,20 @@ function hrleaveandabsences_civicrm_validateForm($formName, &$fields, &$files, &
   }
 }
 
+/**
+ * Implementation of hook_civicrm_pageRun
+ */
+function hrleaveandabsences_civicrm_pageRun(&$page) {
+  if ($page instanceof CRM_HRLeaveAndAbsences_Page_AbsenceTab) {
```
davialexandre: Instead of using the hook, you can use the page itself (in its `run` method) to push the permissions

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/_variables.scss
AkA84: This is a standard `_variables.scss` partial that any theme (local or global) has, also these variables are used across the board, in mixins, overriding styles, components. It's better to keep them here, i think

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.js
deb1990: I think you are trying to set the role, but we are already setting the role here https://github.com/civicrm/civihr/blob/d1c444dd03881aa21fd4af1b6d367d7ce808190c/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/absence-tab/components/absence-tab-container.js#L15-L15

Then we already know that we are in admin role using that rootscope variable. So I am not sure why checkPermission needed to be added here. And using the same rootScope variable we are deciding the role in other components, specially in leave-request-popup.

AkA84: Honestly I prefer using the `checkPermission` service , as it make the component independent of any outside properties/methods

PR #1966, opened 2017-06-26
PCHR-2345: Update labels of "1/2 AM" and "1/2 PM" option values

davialexandre:
## Overview
The labels were updated to:
- "1/2 AM":  "Half-day AM"
- "1/2 PM":  "Half-day PM"

## Before
captura de tela de 2017-06-26 12-17-38

## After
captura de tela de 2017-06-26 12-18-41

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -892,7 +892,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    *     '2 => [
    *       'id' => 2,
    *       'value' => '2',
-   *       'label' => '1/2 AM'
+   *       'label' => 'Half-day AM'
    *     ]
    *   ]
    */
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/201706261302.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_201706261302 {
+
+  /**
+   * Updates the labels of the "1/2 AM" and "1/2 PM" option values to "Hal-day AM"
```
MiyaNoctem: s/Hal-day/Half-day/g

MiyaNoctem:
I like the idea of using timestaps for the upgrades to avoid collisions among us, but did you think about this scenario? 

Say this ticket gets blocked, for whatever reason. Another dev creates upgrade 201707272000 tomorrow, and that ticket gets released.  Then this ticket gets unblocked, and merged.  This way, the upgrade for this ticket would never run.  We'd catch it in testing, but then again, it would require to use both a tester and a dev's time to catch it, while if it's just a conflict, either of them would immediately know what's going on.

Other than that, and a type-o, the code looks solid, but I'd like to know your opinion on the upgrade naming thing.

PR #1968, opened 2017-06-27
PCHR-2352: Better error handling duplicated ContactWorkPattern

davialexandre:
## Overview
The DB schema of the ContactWorkPattern has a unique key that prevents the creation of two records with the same contact_id and effective_date. If you use the API and try to create such a duplicate record, it will respond with a cryptic "DB Error: already exists message" and also the faulty SQL query, in case debug was enabled.

## Before
The response of the ContactWorkPattern.create API that would create a duplicate record would look like this:

```json
{
    "error_code": "already exists",
    "sql": "INSERT INTO civicrm_hrleaveandabsences_contact_work_pattern (contact_id , pattern_id , effective_date ) VALUES ( 204 ,  1 ,  20160607000000 )  [nativecode=1062 ** Duplicate entry '204-2016-06-07' for key 'unique_pattern_per_effective_date']",
    "tip": "add debug=1 to your API call to have more info about the error",
    "is_error": 1,
    "error_message": "DB Error: already exists",
    "debug_information": "INSERT INTO civicrm_hrleaveandabsences_contact_work_pattern (contact_id , pattern_id , effective_date ) VALUES ( 204 ,  1 ,  20160607000000 )  [nativecode=1062 ** Duplicate entry '204-2016-06-07' for key 'unique_pattern_per_effective_date']"
}
```

## After
The response for the same API call will look like:

```json
{
    "is_error": 1,
    "error_message": "This contact already have a Work Pattern with this effective date"
}
```

## Technical Details
In order to be able to return a meaningful message to the API, we need to catch the original error and throw a new one. This was done by adding a `validateParams()` method to the ContactWorkPattern BAO, which works exactly the same way as the other `validateParams()` method in BAOs like the AbsenceType or LeaveRequest. In the event of an error, it throws an `InvalidContactWorkPatternException`.

Besides the unique validation, I also added validation for mandatory fields. We also have schema level validation for them, but I added the `validateMandatory()` method to keep consistency and have higher level validations for this too.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/ContactWorkPatternTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/ContactWorkPatternTest.php
@@ -19,6 +19,39 @@ class CRM_HRLeaveAndAbsences_BAO_ContactWorkPatternTest extends BaseHeadlessTest
     CRM_Core_DAO::executeQuery('SET foreign_key_checks = 0;');
   }
 
+  /**
+   * @expectedException \CRM_HRLeaveAndAbsences_Exception_InvalidContactWorkPatternException
+   * @expectedExceptionMessage The contact_id field should not be empty
+   */
+  public function testItCannotBeCreatedWithoutAContactID() {
+    ContactWorkPattern::create([
+      'pattern_id' => 1,
+      'effective_date' => CRM_Utils_Date::processDate('2016-01-01'),
+    ]);
+  }
+
+  /**
+   * @expectedException \CRM_HRLeaveAndAbsences_Exception_InvalidContactWorkPatternException
+   * @expectedExceptionMessage The pattern_id field should not be empty
+   */
+  public function testItCannotBeCreatedWithoutAPatternID() {
+    ContactWorkPattern::create([
+      'contact_id' => 1,
+      'effective_date' => CRM_Utils_Date::processDate('2016-01-01'),
+    ]);
+  }
+
+  /**
+   * @expectedException \CRM_HRLeaveAndAbsences_Exception_InvalidContactWorkPatternException
+   * @expectedExceptionMessage The effective_date field should not be empty
+   */
+  public function testItCannotBeCreatedWithoutAnEffectiveDate() {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/ContactWorkPatternTest.php
```diff
@@ -36,13 +37,31 @@ public function testThereCannotBeTwoWorkPatternsForTheSameEmployeeWithTheSameEff
...
-    } catch(PEAR_Exception $e) {
-      $this->assertEquals('DB Error: already exists', $e->getMessage());
+    } catch(InvalidContactWorkPatternException $e) {
+      $this->assertEquals('This contact already have a Work Pattern with this effective date', $e->getMessage());
 
       return;
     }
 
-    $this->fail('Expected an DB error, but the contact work patternx was created successfully');
+    $this->fail('Expected an DB error, but the contact work pattern was created successfully');
```
tunbola: s/Expected an DB error/Expected a DB error

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/ContactWorkPatternTest.php
```diff
@@ -36,13 +37,31 @@ public function testThereCannotBeTwoWorkPatternsForTheSameEmployeeWithTheSameEff
...
+      'effective_date' => $effectiveDate,
+    ]);
+
+    // Update the pattern_id, but contact_id and effective_date are still the same
+    ContactWorkPattern::create([
+      'id' => $contactWorkPattern->id,
+      'contact_id' => 2,
+      'pattern_id' => 2,
+      'effective_date' => $effectiveDate,
+    ]);
```
tunbola: We should probably do an assert here, asserting the id and the pattern_id to be sure the update went through.

tunbola:
In the PR description s/"DB Erro: already exists message"/"DB Error: already exists message"

davialexandre:
Pushed it again to squash the PR feedback commits

PR #1970, opened 2017-06-28
PCHR-2150: Fix email button text

mickadoo:
## Overview
The button text in the email form was copied over from the "create user record" form and is wrong.

## Before
The button said "Create Records"

## After
The button says "Send"
 
## Comments
- I also did some minor change on the hrcore hook to bring attention to it. I added a [commit](https://github.com/civicrm/civihr/commit/acb77fc5c6d18918c8264d381d3356c4c1bb7fce) where I fixed the permission name but forgot to have it reviewed, so it would be good to also have this checked.
- Tests were failing for two reasons:
  * The feature to add a default limit for API calls added an API Wrapper and tests, but the API Wrapper did not call a require statement to load the Wrapper.php file. This worked when using the site through the browser because other API wrappers required this file, but not in testing where only this single wrapper was being tested.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -26,7 +26,10 @@ function hrcore_civicrm_config(&$config) {
  * @param array $tasks
  */
 function hrcore_civicrm_searchTasks($objectName, &$tasks) {
-  if ($objectName !== 'contact' || !CRM_Core_Permission::check('create users')) {
+  $isContact = $objectName === 'contact';
+  $canUse = CRM_Core_Permission::check('create users');
```
omarabuhussein: what about $canCreateUsers instead of $canUse .. it is more descriptive

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -69,4 +70,22 @@ protected function cleanup() {
     }
   }
 
+  private function loadRequiredData() {
```
davialexandre: This method name is too generic, which usually leads to a god method, where devs will use it to load any type of data. Can you give it a more specific name?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -23,7 +23,8 @@ class DrupalUserServiceTest extends CRM_HRCore_Test_BaseHeadlessTest {
   private $testContact;
 
   public function setUp() {
-    $this->testContact = ContactFabricator::fabricate();
+    CRM_Utils_API_ReloadOption::singleton();
+    $this->loadRequiredData();
```
davialexandre: Do we need to clear anything created by these methods on the `tearDown()` method? Or the transaction rollback after each test is enough to keep things clear?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -69,4 +70,18 @@ protected function cleanup() {
     }
   }
 
+  private function loadRequiredData() {
+    $this->testContact = ContactFabricator::fabricate();
+    $requiredActivityTypes = ['Create User Account', 'Send Onboarding Email'];
+
+    civicrm_api3('OptionGroup', 'create', ['name' => 'activity_type']);
```
omarabuhussein: **activity_type** is a core civicrm  option group so I think it is safe to assume it is already exist

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
davialexandre: In what situation this data won't be available on the database? I'm asking this because the tests pass on my local after a db schema reset.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -69,4 +70,18 @@ protected function cleanup() {
...
+
+    foreach ($requiredActivityTypes as $activityType) {
+      civicrm_api3('OptionValue', 'create', [
```
omarabuhussein: you need to ensure that these two activity types are not already created

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -26,7 +26,10 @@ function hrcore_civicrm_config(&$config) {
  * @param array $tasks
  */
 function hrcore_civicrm_searchTasks($objectName, &$tasks) {
-  if ($objectName !== 'contact' || !CRM_Core_Permission::check('create users')) {
+  $isContact = $objectName === 'contact';
```
davialexandre: The way this hook implementation is assuming that:
- hrcore only handles search tasks for contacts
- hrcore only handles search tasks that require the "create users" permission

Which is something that doesn't make sense. 

Another problem with it is that we have business logic inside the hook itself, which is a bad practice. Ideally, to avoid having a huge hook implementation in the future, we should put this logic inside a separate function and then call it from within the hook.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -23,7 +23,8 @@ class DrupalUserServiceTest extends CRM_HRCore_Test_BaseHeadlessTest {
   private $testContact;
 
   public function setUp() {
-    $this->testContact = ContactFabricator::fabricate();
+    CRM_Utils_API_ReloadOption::singleton();
```
omarabuhussein: I am not sure why you need this line

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
mickadoo: I was running tests for another extension that clear data from some tables. I don't think we should assume that anything will be in the database that's not added the the extension installer.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
mickadoo: Oops, it snook in when debugging the autoloading problem, good catch

uk.co.compucorp.civicrm.hrcore/hrcore.php
mickadoo: I would agree that this logic belongs somewhere else. I don't really like the method of adding a lot of functions inside hrcore.php though, and prefer to use classes. For this case it might be overkill though. 

I'll change the logic to remove the early return and add the search tasks inside an if statement.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
davialexandre: It actually depends on where this data comes from. It's safe if it comes from either:
- Any of these files https://github.com/civicrm/civicrm-core/blob/728bbd5bbad0b5269dcd0a4cb70384291d309e70/Civi/Test/Data.php#L20-L22 (These are loaded and the db schema is reset for the tests)
- Added by the extension (as you said)

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
mickadoo: The transaction rollback is enough to remove the data. I had problems where rollbacks failed before, but found that this was due to working with custom data. It's mentioned [here](https://wiki.civicrm.org/confluence/display/CRMDOC/Transaction+Reference) in the section "Special Topics: TRUNCATE and ALTER force immediate commit"

uk.co.compucorp.civicrm.hrcore/hrcore.php
davialexandre: What is the problem with adding functions to hrcore? I feel like it's still better than adding the logic directly to the hook itself.

In the future, if another dev needs to add a new search task, he'll find this hook check how you did things and will probably think that is fine to just add some other if or else and add the tasks to the list. Suddenly, this hook will have [more than 100 lines](https://github.com/civicrm/civihr/blob/staging/hrui/hrui.php#L227-L327) and no one will ever refactor it

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -23,8 +23,7 @@ class DrupalUserServiceTest extends CRM_HRCore_Test_BaseHeadlessTest {
   private $testContact;
 
   public function setUp() {
-    CRM_Utils_API_ReloadOption::singleton();
-    $this->loadRequiredData();
+    $this->createUserAndActivities();
     $this->registerCurrentLoggedInContactInSession($this->testContact['id']);
     $this->cleanup();
```
davialexandre: Why is the cleanup() inside the `setUp()` step? Usually, we clean things up on the `tearDown()`

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
```diff
@@ -69,4 +69,22 @@ protected function cleanup() {
     }
   }
 
+  private function createUserAndActivities() {
```
davialexandre: I see some problems here:
- Having an "And" in the method name, usually, means a method is doing 2 or more things. Which is exactly what is happening here. Creating the contact doesn't have anything to do with the activity types. Since it's just a single line of code, it can be called directly from the `setUp()` method. 
- The method doesn't really create Activities, but Activity Types
- Why do you need to manually create these Activitiy Types in the first place? Aren't they created by some upgrader in some extension?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
mickadoo: I added it originally when I was getting fatal errors and cleanup wasn't called. It shouldn't be needed so removing it.

uk.co.compucorp.civicrm.hrcore/hrcore.php
mickadoo: Sure, I agree, but if we just add functions to this file we'll end up with files like [this](https://github.com/civicrm/civihr/blob/c921837290c61e8c9b102dd3358862969e8c3b78/hrui/hrui.php) with almost 1000 lines. Neither result is optimal. That's why I'd prefer to move the logic to a separate class, so that's what I'll do.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
mickadoo: I was definitely getting errors that the option value didn't exist. It can happen if there are some changes in the database, for example I was doing some tests about what would happen if `activity_type` option group didn't exist. This test will fail then. But maybe this is a step too far, since we can't protect against manual database manipulation. I'll remove this function.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/SearchTask/SearchTaskAdderInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Determines whether or not the tasks should be added.
+   *
+   * @param $objectName
+   *
+   * @return bool
+   */
+  public static function shouldAdd($objectName);
+}
```
davialexandre: I don't see the point of this interface:
- You're not using actually using it anywhere (you're using the concrete class)
- It only has static methods, so you cannot do something like `__construct(SearchTaskAdderInterface $adder)` and call these methods (because calling static methods from an instance is a deprecated feature in PHP)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/SearchTask/ContactFormSearchTaskAdder.php
```diff
@@ -0,0 +1,35 @@
...
+
+use CRM_HRCore_Form_CreateUserRecordTaskForm as CreateUserRecordTaskForm;
+use CRM_HRCore_Form_SendInvitationEmailTaskForm as SendInvitationEmailTaskForm;
+use CRM_HRCore_SearchTask_SearchTaskAdderInterface as SearchTaskAdderInterface;
+
+class CRM_HRCore_SearchTask_ContactFormSearchTaskAdder implements SearchTaskAdderInterface {
+
+  /**
+   * @param $tasks
+   */
```
davialexandre: When the child methods have the same description as their parents, you can use `@inheritdoc`

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -26,18 +27,9 @@ function hrcore_civicrm_config(&$config) {
  * @param array $tasks
  */
 function hrcore_civicrm_searchTasks($objectName, &$tasks) {
-  if ($objectName !== 'contact' || !CRM_Core_Permission::check('create users')) {
-    return;
+  if (ContactFormSearchTaskAdder::shouldAdd($objectName)) {
+    ContactFormSearchTaskAdder::add($tasks);
   }
```
davialexandre: What about this?

```php
function hrcore_civicrm_searchTasks($objectName, &$tasks) {
  $adders = [
    new AdderOner(),
    new AdderTwo(),
    //....
  ];

  foreach($adders as $adder) {
     if($adder->shouldAdd($objectName)) {
        $adder->addTo($tasks); // Note that I changed the method name to addTo, otherwise, it might look like we're adding the tasks array to the adder
     }
  }
}
```

I don't like the idea of having to instantiate each adder, but I can't think of any other way to do that, given how the hooks work

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/SearchTask/SearchTaskAdderInterface.php
mickadoo: I don't see the point of it either for this case, but I got the impression that you wanted this to be very future proof for other developers coming along. If the interface is there the next dev is more likely to copy it, and in that case we'll have a consistent approach.

uk.co.compucorp.civicrm.hrcore/hrcore.php
mickadoo: I'm not crazy about instantiating those classes on every request either, I don't usually like static calls but I thought they were more suitable here since they shouldn't have any dependencies or internal state.

PR #1971, opened 2017-06-28
PCHR-2150: Validate email and username on creation

mickadoo:
## Overview
It is possible to create a user record with an valid email that does not match the drupal username validation rules, for example "foo+1@bar.com"

## Before
You can create a user record from the form that doesn't have a valid username. Attempts to update that user later from the Drupal admin area will fail.

## After
The username and email are validated before creation of user records. Validation failures are displayed with an explanatory message.

- [x] Tests Pass (you know it)

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
@@ -67,7 +67,7 @@ class CRM_HRCore_Service_DrupalUserService {
   }
 
   /**
-   * @param $name
+   * @param string $name
    *
    * @return bool
    */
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -67,6 +67,24 @@ public function sendActivationMail($contactID, $email) {
...
+   * @return bool
+   */
+  public function isValidUsername($name) {
+    return empty(user_validate_name($name));
+  }
+
+  /**
+   * @param $email
+   *
+   * @return bool
```
davialexandre: You're missing the param type on both these methods.

mickadoo:
@davialexandre the 2149 and 2150 branchs were merged to the epic branch since they passed review and we decided not to do separate testing on each epic sub-task. Both 2149 and 2150 will be released together according to Mike. Since then I've been treating them as similar since a lot of stuff applies to both tickets, but you're right - this PR is more specific to PCHR-2149 so I'll change the commits and branch name and create a new PR.  

Closing in favor of #1973

PR #1972, opened 2017-06-28
PCHR-2336: Create Absence Dates Report.

tunbola:
## Overview
We need to import data from the old Absence extension into the new Leave and Absence extension. The current Absence report in the Absence extension lists each leave request on a single line with a sum of the amount deducted for the leave requests rather than listing each leave dates on its own line with the amount deducted for each date. 
This new Absence Dates report is a modified version of the Absence Report that lists the leave dates on its own line, removes some un-needed columns also add some new columns that is necessary for data import into L&A extension.

The Absence Dates report has the following columns:

- Absence ID(The ID that identifies leave dates belonging to the same group)
- ContactID
- Absence Type
- Absence Date
- Qty(Amount Deducted for the date)
- Status
- Comment

And the report is filterable on ContactID, Status and Absence Type columns and  Sortable on Absence Date and Absence ID columns.
## Before
The current absence report does not provide sufficient data that can be imported into Leave and Absence extension.

## After
A new report named Absence Dates Report was created by cloning the Absence report and some modifications made. 
In the Absence extension, absence requests are stored in the `civicrm_activity` table and for each absence request, there is a parent activity. The current absence report fetches and displays the parent activity for each absence request. The parent activity also holds the absence type information while the child activity are of type `Absence`.

The following changes were incorporated into the Absence Dates report:

- The Absence dates report displays information for the child activities of type Absence which are the individual absence dates.
- The Absence type information is retrieved from the parent activity and attached to the corresponding child activities.
- Columns and filters not relevant to the Absence dates report and present in the Absence report were either hidden from display or removed and relevant ones added
- The contact ID of the absence request was added to the report and to the filters.
- Fixed a bug inherited from the Absence report that causes downloaded csv from the dashboard to have blank data.
- Some cleanup

See differences [HERE](https://www.diffchecker.com/kvhiV46t) between the Absence report and the Absence dates report.

absencedates

- [ ] Tests Pass
The changes made does not affect the tests on the old absence extension.

### DIFF
--- a/hrreport/CRM/HRReport/Form/Activity/HRAbsenceDates.mgd.php
+++ b/hrreport/CRM/HRReport/Form/Activity/HRAbsenceDates.mgd.php
@@ -1,0 +1,68 @@
+<?php
+// This file declares a managed database record of type "ReportTemplate".
+// The record will be automatically inserted, updated, or deleted from the
+// database as appropriate. For more details, see "hook_civicrm_managed" at:
+// http://wiki.civicrm.org/confluence/display/CRMDOC/Hook+Reference
+$isEnabled = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Extension', 'org.civicrm.hrabsence', 'is_active', 'full_name');
+$absenceOptionValue = civicrm_api3('OptionValue', 'get', [
+  'name' => 'Absence',
+  'option_group_id' => 'activity_type',
+  'sequential' => 1
+]);
+
+$activityTypeValue = $absenceOptionValue['values'][0]['value'];
+
+if ($isEnabled) {
+  $activityStatus = CRM_HRAbsence_BAO_HRAbsenceType::getActivityStatus();
+  $approvedStatus = array_search('Approved', $activityStatus);
+}
+else {
+  $activityStatus = CRM_Core_PseudoConstant::activityStatus();
+  $approvedStatus = array_search('Completed', $activityStatus);
+}
+
+return [
+  [
+    'name' => 'CiviHR Absence Dates Report Template',
+    'entity' => 'ReportTemplate',
+    'params' =>
+    [
+      'version' => 3,
+      'label' => 'CiviHR Absence Dates Report',
+      'description' => 'HR Report showing absence dates for each individual. ',
+      'class_name' => 'CRM_HRReport_Form_Activity_HRAbsenceDates',
+      'report_url' => 'civihr/absencedates',
+      'grouping' => 'Absence',
+      'component' => '',
+    ],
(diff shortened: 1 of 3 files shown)


hrreport/CRM/HRReport/Form/Activity/HRAbsenceDates.php
```diff
@@ -0,0 +1,640 @@
...
+                  'required' => TRUE,
+                ],
+            ],
+          'filters' =>
+            [
+              'activity_type_id' =>
+                [
+                  'title' => ts('Absence Type'),
+                  'operatorType' => CRM_Report_Form::OP_MULTISELECT,
+                  'options' => $this->absenceActivityType,
```
davialexandre: Why did you change this? On the absence report, it was a list of Absence Types (as the filter title says), but now it's just a fixed list with a single activity type

tunbola: The child activity don't hold the absence type information and all of them are of type `absence`.  The absence type information was added in the `alterDisplay` function after the query might have ran and the rows fetched. 
Since the parent activities were not part of the original query, it will be difficult to filter by the various absence types and I did not want to waste too much time on that. I can remove this filter because even if I ad back the absence types list, it won't work.
What do you suggest?

hrreport/CRM/HRReport/Form/Activity/HRAbsenceDates.mgd.php
```diff
@@ -0,0 +1,68 @@
...
+  [
+    'name' => 'CiviHR Absence Dates Report Template',
+    'entity' => 'ReportTemplate',
+    'params' =>
+    [
+      'version' => 3,
+      'label' => 'CiviHR Absence Dates Report',
+      'description' => 'HR Report showing absence dates for each individual. ',
+      'class_name' => 'CRM_HRReport_Form_Activity_HRAbsenceDates',
+      'report_url' => 'civihr/absencedates',
```
tunbola: good Idea

hrreport/CRM/HRReport/Form/Activity/HRAbsenceDates.php
davialexandre: then let's remove it for now. We can add it later in case we need it

tunbola: Sure.

tunbola: Removing the filter completely seems to break the report. So I simply set it to hidden.

PR #1980, opened 2017-07-03
PCHR-2340: Set 'Home' as Default Location and Primary Address

MiyaNoctem:
## Overview
When an administrator creates a new contact, addresses may have one of several different location types. The system will allow him to add several different addresses (one for each location type) and set one of them as primary. This conflicted with the logic implemented in the SSP, where it is required for the 'Home' address to be the primary address for the contact in order to be able to view and edit the contact's personal information.

## Before
The problem was SSP uses 'Home' location type to look for the address to be used on the dashboard and the My Details edit view, making the assumption this 'Home' address is also the contact's primary address. If the contact did not have the 'Home' address set as primary, but some other address, saving the address through the SSP effectively saved the information on the existing 'Home' address record, but no changes were made to the is_primary flag. This caused the information on the dashboard to remain empty, as the view is built by looking for addresses that are both 'Home' location AND primary.

## After
The requirement for the ticket specified to fix this, two things needed to happen:

1. Make sure on all existing installations and all new installations the default location type is 'Home'
2. Make sure that the address record saved from SSP is saved with Primary flag

Added an upgrader trait to hrcore extension that sets 'Home' location type as default and reserved, to prevent accidental modification or elimination of the record. It also updates all 'Home' location type addresses to be primary, for all contacts that have them.

Finally, I implemented _civicrm_pre hook on hrcore extension to check for when an address is being edited or saved. If the address' location type is 'Home', then it'll set the is_primary flag to true.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
@@ -29,12 +29,12 @@ trait CRM_HRCore_Upgrader_Steps_1004 {
 
   private function up1004_setAllHomeAddressesAsPrimary() {
     $homeLocation = civicrm_api3('LocationType', 'getsingle', [
-      'name' => "Home",
+      'name' => 'Home',
     ]);
 
     $contactsWithNonPrimaryHomeAddress = civicrm_api3('Address', 'get', [
       'sequential' => 1,
-      'return' => array('contact_id'),
+      'return' => ['contact_id'],
       'location_type_id' => $homeLocation['id'],
       'is_primary' => 0,
     ]);
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
@@ -1,0 +1,56 @@
+<?php
+
+trait CRM_HRCore_Upgrader_Steps_1004 {
+
+  /**
+   * Upgrader to set 'Home' as default location and set all 'Home' addresses as
+   * primary.
+   *
+   * @return bool
+   */
+  public function upgrade_1004() {
+    $this->up1004_setHomeLocationAsDefault();
+    $this->up1004_setAllHomeAddressesAsPrimary();
+
+    return true;
+  }
+
+  private function up1004_setHomeLocationAsDefault() {
+    civicrm_api3('LocationType', 'get', [
+      'sequential' => 1,
+      'name' => 'Home',
+      'api.LocationType.create' => [
+        'id' => '$value.id',
+        'is_default' => 1,
+        'is_reserved' => 1,
+      ],
+    ]);
+  }
+
+  private function up1004_setAllHomeAddressesAsPrimary() {
+    $homeLocation = civicrm_api3('LocationType', 'getsingle', [
+      'name' => "Home",
+    ]);
+
+    $contactsWithNonPrimaryHomeAddress = civicrm_api3('Address', 'get', [
+      'sequential' => 1,
+      'return' => array('contact_id'),
+      'location_type_id' => $homeLocation['id'],
+      'is_primary' => 0,
+    ]);
+
+    $contactIDs = [];
+    foreach ($contactsWithNonPrimaryHomeAddress['values'] as $currentContact) {
+      $contactIDs[] = $currentContact['contact_id'];
+    }
+
+    if (!empty($contactIDs)) {
+      $query = "
+        UPDATE civicrm_address
+        SET is_primary = CASE WHEN location_type_id = {$homeLocation['id']} THEN 1 ELSE 0 END
+        WHERE contact_id IN (" . implode(',', $contactIDs) . ")
+      ";
+      CRM_Core_DAO::executeQuery($query);
+    }
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
@@ -28,29 +28,11 @@ trait CRM_HRCore_Upgrader_Steps_1004 {
   }
 
   private function up1004_setAllHomeAddressesAsPrimary() {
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -134,3 +134,23 @@ function hrcore_civicrm_apiWrappers(&$wrappers, $apiRequest) {
...
+function hrcore_civicrm_pre($op, $objectName, $objectId, &$params) {
+  if ($objectName === 'Address' && ($op === 'edit' || $op === 'create')) {
+    $homeLocation = civicrm_api3('LocationType', 'getsingle', [
+      'name' => 'Home',
+    ]);
+
+    if ($params['location_type_id'] == $homeLocation['id']) {
+      $params['is_primary'] = 1;
+    }
+  }
```
davialexandre: We should avoid adding business logic directly into the hook.  One approach we've been trying to follow in other extensions is make make the hooks work like event dispatchers. It would like this:

```php
function hrcore_civicrm_pre($op, $objectName, $objectId, &$params) {
  $listeners = [
     new Listener1(),
     new Listener2(),
     new Listerner3(),
     ...
  ];

  foreach($listeners as $listener) {
    $listener->handle($op, $objectName, $objectId, &$params);
  }
}
```

As for the listener, you could have something like:
```php

class CRM_HRCore_Hook_Pre_PrimaryAddressSetter {
  public function handle($op, $objectName, $objectId, &$params) {
    if(!$this->shouldHandle($op, $objectName)) {
       return;
    }
    
    $homeLocation = civicrm_api3('LocationType', 'getsingle', [
      'name' => 'Home',
    ]);

    if ($params['location_type_id'] == $homeLocation['id']) {
      $params['is_primary'] = 1;
    }
  
  }

  private function shouldHandle() {
    // some logic here
  }
}

```

I know this might look like too much at this point, but it's a measure to prevent hooks to become a mess of if and else, like what we have in some of the other extensions

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  private function up1004_setHomeLocationAsDefault() {
+    civicrm_api3('LocationType', 'get', [
+      'sequential' => 1,
+      'name' => 'Home',
+      'api.LocationType.create' => [
+        'id' => '$value.id',
+        'is_default' => 1,
+        'is_reserved' => 1,
```
MiyaNoctem: It's an 'is_reserved' value.  I think it would be a good idea, so the value doesn't get changed or deleted if the SSP depends on it.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
```diff
@@ -0,0 +1,56 @@
...
+    foreach ($contactsWithNonPrimaryHomeAddress['values'] as $currentContact) {
+      $contactIDs[] = $currentContact['contact_id'];
+    }
+
+    if (!empty($contactIDs)) {
+      $query = "
+        UPDATE civicrm_address
+        SET is_primary = CASE WHEN location_type_id = {$homeLocation['id']} THEN 1 ELSE 0 END
+        WHERE contact_id IN (" . implode(',', $contactIDs) . ")
+      ";
```
MiyaNoctem: I could use the API, but I'd need two different calls, one to set 'Home' addresses as primary, the other one to set all the other addresses as not primary. I could also use a single call, kind of, by cycling through all contacts, making each call to set the is_primary flag to true and let the Address BAO handle the setting of all other addresses to false, but that would obviously take a lot more time - O(1) vs O(n).

On the other note, my idea was to set the is_primary flag to true for all contacts that had a 'Home' address that is not primary.  Only for those contacts should 'Home' addresses be set as primary, and other addresses be unset. To do that I need the contact's ID.

uk.co.compucorp.civicrm.hrcore/hrcore.php
MiyaNoctem: Nice approach, I'll keep using it from now on!

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
davialexandre: @MiyaNoctem I agree it's a good idea, but did you confirm with someone else if this is ok?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
davialexandre: @MiyaNoctem the point about using the API is that we should it as much as possible (when dealing with CiviCRM core entities) and we should resort to raw SQL queries only when there's no other option. I'm aware that using the API is not a performant option, but this upgrader will be executed only once on each client so I think it's fine.

Now, about setting and unsetting the is_primary value. You don't need to do that manually. If you use the API to set an Address as primary, it will automatically unset it for all the other addresses of the contact. So I think you can do everything with just:

```php
$result = civicrm_api3('Address', 'get', array(
  'sequential' => 1,
  'location_type_id' => "Home",
  'is_primary' => 0,
  'api.Address.create' => array('id' => "\$value.id", 'is_primary' => 1),
));
```

(I didn't really test this, so it might not work)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
MiyaNoctem: I asked Mike about this and suggested we share it with the rest of the CiviHR team, asking if we should make 'Home' reserved, or change the SSP to use primary address instead. The decision was to keep SSP hard-wired to the 'Home' address location type and make it reserved, as this was the original intention behind the functionality on the SSP (allow users to check and update their current home address).

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1004.php
MiyaNoctem: Done as suggested, using the API. I tested it and everything worked ok.

PR #1981, opened 2017-07-04
PCHR-2033: Calendar in Admin Dashboard

deb1990:
## Overview
As part of this PR the Calendar for the Admin Dashboard has been implemented.

## After
dashboard___hr17-1

Note: It was not possible to include a GIF because of massive performance issues, but I have checked the following is working
1. Show Single month
2. Show all months
3. Filtering
4. Updating the calendar when leave request is deleted.
5. Opening of Leave Request popup from Calendar.

## Technical Details
1. Permissions are added to `Dashboard.php`.
2.`loggedInUserId` is set as `contactId` in setting file.
3.`manager-leave-calendar` has been moved to shared folder, also the related test,html files are moved.
4. `checkPermissions` is used to determine whether the user is admin.
5. Some failing test cases in `leave-request-ctrl`has been fixed, which were failing due to timezone related issues.
5. `manager_calendar.scss` moved from SSP.

## Comments
1. Backstop Tests are not written as right now the performance is very bad and backstop is taking lot of time to take screenshot and even hangs sometimes.
2. Now `admin` can see all contacts in Manager Calendar(SSP), previously only the contacts with `has leave approved by` relationship were visible.
3. The performance is very bad while loading all the contacts, but this is a known issue and we have PCHR-2240 to fix that.

## Refactored leave-request-popup

As per our plan, the `leave-request-popup` has been refactored and a new component `leave-request-popup-comments-tab` has been created. All the code related to comment has been moved to this component and removed from `request-ctrl`.

---

- [x] Tests Pass

PR #1982, opened 2017-07-04
PCHR-2343: Display leave request statuses according to status matrix

reneolivo:
## Overview
This PR fixes the leave request statuses on the *Leave Request* modal that is shown when you edit leave requests on *Manager Leave*. The statuses dropdown should follow this matrix:

2238 manager sm

Meaning:

* When the request is new (*none* row) only *More Information Required* and Approved should be shown on the dropdown.
* When the request previous status was *Awaiting Approval*, then *More Information Required*, *Rejected*, *Approved*, and *Cancelled* should be shown.
* And so on...

The current behaviour doesn't take into account the previous status.

## How it was before

anim

## How it is now

anim

## Technical Details

Relevant test cases were added for the issue on *request-ctrl_test.js*. *request-ctrl.js* Needs refactoring to move away the new code into a different component. PR Review is needed before taking this step.

The current behaviour takes into account the manager privileges in order to return a list of status. That is missing from this PR. Further explanation is needed.

## Comments

All tests are passing except for the ones fixed on PR #1981

---

- [x] Tests Pass


igorpavlov:
1) Please amend the title of the PR to something like "PCHR-2343: Add cancel and awaiting approval options"
2) Please use Overview section to describe what does this PR do in non/semi-technical language and in your own words (do not copy from the ticket please).
3) Please avoid first-person phrasing (as per Technical Details). It would recommend to write something like "This PR adds...". Also the Technical Description should not contain things like "should be done" or "want to". I should only describe *how* this PR solves the problem. If you have any questions on the ticket, please feel free to use comments in JIRA.
4) Please run tests and if they are OK, tick the box (`[x]`) Tests Pass.

reneolivo:
is it ok if I force push to this branch to fix the commits?

davialexandre:
@reneolivo yes

tunbola:
@reneolivo, Sure you can, Since its your branch and you're the only one working on it at the moment.

reneolivo:
@igorpavlov I have updated the commit messages. Is there anything else I should look into?

Thanks for your patience everyone :)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request-ctrl_test.js
```diff
@@ -0,0 +1,158 @@
+/* eslint-env amd, jasmine */
```
deb1990: We should not create a test file for `request-ctrl.js`as it is a parent class and never used directly. All the tests for this class should go to `leave-request-ctrl_test`.

reneolivo: I'd recommend leaving the test in `request-ctrl_test.js` for two reasons:

1. The code that is being tested is on `request-ctrl.js`. It's easier to find the test code for that controller in a file with a similar name, even if it's a parent controller.
2. The code is shared between all child controllers. If for some reason `leave-request-ctrl_test.js` is to be removed, the test for this code will still work ensuring that `sick-request-ctrl.js` and `toil-request-ctrl.js` will work as expected.

Regardless, the better approach is to move all this code into a component so we don't clutter `request-ctrl.js` with extra business logic.

reneolivo:
@igorpavlov is everything in order now? Should I move the test code to `leave-request-ctrl_test.js` as suggested by @deb1990 ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request-ctrl_test.js
igorpavlov: Despite of the fact that I actually agree with @reneolivo in general, in that particular case we should move to `leave-request-ctrl_test` because the parent controller is indeed never used directly. These files were built a long time ago and I personally don't like how they are structured, but we don't want to mess with them at least in this particular PR.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -27,9 +27,11 @@ define([
     ) {
       $log.debug('RequestCtrl');
       var absenceTypesAndIds;
+      var availableStatusesMatrix = {};
       var initialLeaveRequestAttributes = {}; // used to compare the change in leaverequest in edit mode
       var role = '';
       var NO_ENTITLEMENT_ERROR = 'No entitlement';
+      var vm;
```
igorpavlov: In *this particular file* we have to use "this" and "self" instead of "vm" at the moment (I am not sure we want to mix vm and this, it is one or another, or you will need to refactor the whole file). Please define `this. requestStatuses = {};` at the start and all similar props, that would enhance the readability.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -248,22 +250,53 @@ define([
...
 
       /**
+       * Helper functions to get available statuses depending on the
+       * current request status value.
+       *
+       * @return {Array}
+       */
+      function getAvailableStatusesForCurrentStatus () {
+        var currentStatus = vm.getStatusFromValue(vm.request.status_id);
+        console.log(' ...', vm.request.status_id, vm.requestStatuses);
```
igorpavlov: Please remove console.log

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -539,10 +602,12 @@ define([
        */
       this._init = function () {
         var self = this;
+        vm = this;
```
igorpavlov: not needed.

reneolivo: @AkA84 is it ok to leave the *vm* var here or should I refactor? I'm using it for convenience saque somewhere else in the code.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1288,6 +1288,137 @@
...
+
+          $ctrl._init();
+          $ctrl.request = {
+            status_id: null
+          };
+          $ctrl.requestStatuses = status;
+        });
+
+        it('defines a getStatuses method', function () {
+          expect(typeof $ctrl.getStatuses).toBe('function');
```
deb1990: I dont think we need this test to be honest. Just below you are calling `$ctrl.getStatuses()`, and if it is not a function, we will get an error anyway.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1288,6 +1288,137 @@
...
+        });
+
+        it('returns an empty array if .request is not defined', function () {
+          delete $ctrl.request;
```
deb1990: The following structure should be followed
```javascript
 describe('when request is not defined', function() {
 	beforeEach(function(){
 	  delete $ctrl.request;
 	  //Or any other setup
 	});

 	it('returns an empty array', function () {
   // it block should contain only the expect sentence, or any other statement, which cannot be moved to beforeEach or AfterEach
	  var result = $ctrl.getStatuses();
	  expect(result).toEqual([]);
	  //or directly, I would prefer this approach
	  expect($ctrl.getStatuses()).toEqual([]);
	});
 })
```

Please apply this structure for all your unit tests.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1288,6 +1288,137 @@
...
+          $ctrl.requestStatuses = {};
+
+          var result = $ctrl.getStatuses();
+          expect(result).toEqual([]);
+        });
+
+        it('returns *More Information Required, Approve* when previous status was not defined', function () {
+          var result = $ctrl.getStatuses();
+
+          expectStatusToBeEqual(result, [
```
deb1990: Why do you need this `expectStatusToBeEqual` function, you can just do
`expect(result).toEqual([the array])`

deb1990:
@reneolivo As it involves a UI change, please provide GIF for before and after.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -248,22 +250,52 @@ define([
...
+
+      /**
+       * Helper function that returns an array of the statuses available
+       * for a specific status name.
+       *
+       * @return {Array}
+       */
+      function getAvailableStatusesForStatusName (statusName) {
+        return _.filter(vm.requestStatuses, function (nextStatus) {
+          return _.includes(
```
reneolivo: This line is 83 characters long. Not sure about your style guide, but this file seems to be following 80 characters as the limit.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -248,22 +250,52 @@ define([
...
+            availableStatusesMatrix[ statusName ],
+            nextStatus.name
```
reneolivo: I'll refactor this function since I think I can use `availableStatusesMatrix` directly. 

nextStatus.name is used here to see if the status is part of the statuses matrix which means it's a valid transition status. Now that I understand the code better I think it's safe to return availableStatuses directly or using a map.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
reneolivo: two reasons:

1 - `expect(result).toEqual([the array])` expects the order to be preserved. This function makes sure that all elements are equal, but doesn't care about the order.
2 - The error message is more friendly this way, it only gives the name of the statuses that were passed and the ones that were expected. Without this function it would print the whole content of the objects which would be harder to debug.

Should I remove this function then?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1288,6 +1288,137 @@
...
+        function expectStatusToBeEqual (array1, array2) {
+          expect(_.isArray(array1)).toBe(true, array1, 'is not an array');
+
+          expect(array1.length).toBe(
+            array2.length,
+            array1.map((e) => e.name), array2.map((e) => e.name)
+          );
+
+          array1.forEach(function (element) {
+            expect(array2.indexOf(element) >= 0).toBe(
```
reneolivo: This was only on the unit test since the runner browser is Chrome and it supports arrow functions by default.

I'll remove them though.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
igorpavlov: @reneolivo 83 should be fine.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -782,6 +787,30 @@ define([
...
+
+      /**
+       * Helper function that returns an array of the statuses available
+       * for a specific status name.
+       *
+       * @return {Array}
+       */
+      function getAvailableStatusesForStatusName (statusName) {
+        return _.map(availableStatusesMatrix[ statusName ], function (status) {
+          return this.requestStatuses[ status ];
```
igorpavlov: Please remove spaces `[ status ]` -> `[status]` as well as for *statusName* and other places.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1279,6 +1279,147 @@
...
+
+        describe('when previous status is *More Information Required*', function () {
+          beforeEach(function () {
+            $ctrl.request.status_id = status.more_information_required.value;
+          });
+
+          it('returns *Approve, More Information Required, Reject, Cancel* ', function () {
+            var result = $ctrl.getStatuses();
+
+            expect(result).toEqual([
```
igorpavlov: `expect($ctrl.getStatuses()).toEqual([`

The less code between *it* and *expect* - the better.

Please also check other places in this file for the similar thing.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1279,6 +1279,147 @@
...
+
+          status = _.indexBy(allStatuses, 'name');
+
+          initTestController({
+            contactId: CRM.vars.leaveAndAbsences.contactId,
+            userRole: 'admin'
+          });
+
+          $ctrl._init();
+          $ctrl.request = {
```
igorpavlov: Squash it please to one line.

```js
$ctrl.request = { status_id: null };


```

reneolivo:
Important note: I haven't added .min.js files to this PR! :warning:

deb1990:
@reneolivo Please add the min files before merging, and consider this to be approved

deb1990:
@reneolivo This PR contains so many commits and most of them are similar, so in this case you can squash all into 1 or 2 commit.

PR #1983, opened 2017-07-05
PCHR-2364: Implement CIVI Importer for Leave Request Imports.

tunbola:
## Overview
This PR implements Civi Importer in Leave and Absence extension to allow Leave/Absence records to be imported from the old absence extension and also from other sources in the future. 

## Before
Leave/Absence Records cannot be imported into L&A.

## After
Data from the Absence Dates report and data in similar format can now be imported into L&A to create Leave records. All present validations applicable to currently creating a leave request in L&A are applied when importing the data, errors are reported for rows that can not be imported and these can be downloaded by the user in order to make corrections. 
leaveimport

## Technical Details
Most of the files that make up the Civi Importer functionality implemented in L&A are copied from the Activity importer in core and modified to suit L&A. The modifications includes renaming some phrases and removing unneeded variables and minor cleanup. However the bulk of the technical work for the data import is done in the following classes
- The `CRM_HRLeaveAndAbsences_Import_Parser_BaseClass` class: This class sets the importable fields needed and also handles the validations for each row of data in the CSV to be imported.
- The `CRM_HRLeaveAndAbsences_Import_Helpers_DataImportTrait` is a helper Trait that handles creating leave requests, corresponding leave balance changes and comments.

## Comments
- The Leave Data Import functionality is accessible via `civicrm/admin/leaveandabsences/import` URL.
- When creating a leave request, the day types are defaulted to All Day for the sake of simplicity and because of the way the data is currently structured. This is because the half days and All day types in the old absence extension are not stored in db, it only determines the amount deducted.
- When creating Sickness requests, the `sickness_reason` required field is defaulted to 1 for now pending when [PCHR-2310](https://compucorp.atlassian.net/browse/PCHR-2310) will be done.
- The `toil_duration` required field is also set to a default value of 60 for now.
- A TOIL request is identified by the work ` (Credit)` in front of the relevant absence type in the CSV to indicate that it is of type TOIL. I need to conduct some investigations to see if the is_credit column can be added in the absence dates report as suggested [Here](https://compucorp.atlassian.net/wiki/pages/viewpage.action?pageId=120567511).
- Some activity related Css classes are still in the templates that may need to be swapped with relevant L&A CSS classes.

---

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+
+/**
+ * Class to parse activity csv files.
+ */
+class CRM_HRLeaveAndAbsences_Import_Parser_BaseClass extends CRM_HRLeaveAndAbsences_Import_Parser {
```
davialexandre: Seeing this now, I think that adding Class to the class name is a bit redundant. What about `CRM_HRLeaveAndAbsences_Import_Parser_BaseClass`?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+   * @param array $mapperKeys
+   */
+  public function __construct(&$mapperKeys) {
+    parent::__construct();
+    $this->_mapperKeys = &$mapperKeys;
+  }
+
+  /**
+   * Function of undocumented functionality required by the interface.
+   */
```
davialexandre: :ok_hand: :see_no_evil:

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+  protected function fini() {}
+
+  /**
+   * The initializer code, called before the processing.
+   */
+  public function init() {
+    $activityContact = CRM_Activity_BAO_ActivityContact::import();
+    $activityTarget['target_contact_id'] = $activityContact['contact_id'];
```
davialexandre: You're not using these variables anywhere else in this method. Do you really need them?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+   * Handle the values in mapField mode.
+   *
+   * @param array $values
+   *   The array of values belonging to this line.
+   *
+   * @return bool
+   */
+  public function mapField(&$values) {
+    return CRM_Import_Parser::VALID;
+  }
```
davialexandre: This looks weird. The method receives a value but doesn't do anything to it and always return the same thing. Please describe why in the method description

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+    $errorMessage = NULL;
+    $fields = self::getFields();
+
+    foreach($params as $key=>$value) {
+      if (isset($fields[$key]['type']) && $fields[$key]['type'] == CRM_Utils_Type::T_DATE) {
+
+        try {
+          new DateTime($value);
+        } catch (Exception $e) {
+          CRM_Contact_Import_Parser_Contact::addToErrorMsg('Invalid Date value for '.$fields[$key]['title'], $errorMessage);
```
davialexandre: you're importing a Leave Request but have to add the error to the Contact parser. Why? Why????????????????

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+   * @param array $params
+   *
+   * @return string
+   */
+  private function validateFieldTypes($params){
+    $errorMessage = NULL;
+    $fields = self::getFields();
+
+    foreach($params as $key=>$value) {
+      if (isset($fields[$key]['type']) && $fields[$key]['type'] == CRM_Utils_Type::T_DATE) {
```
davialexandre: To avoid all this $fields values manipulation, what about storing the type in a variable?

```php
$fieldType = empty($fields[$key]['value']) ? null : $fields[$key]['value'];
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * Validate that the Absence Type, Absence Status and any other option values in the csv row
+   * are valid option values.
+   *
+   * @param array $params
+   *
+   * @return string
+   */
+  private function validateOptionValues($params){
```
davialexandre: missing space before opening brace

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+   * specify the request type but rather the absence type is listed with '(Credit) in front
+   * to show that it is an accrual request. This is also simulated when creating the absence Types array to
+   * to match what is in the CSV.
+   *
+   * @return array
+   */
+  private function getAbsenceTypes() {
+    $absenceType = new AbsenceType();
+    $absenceType->is_active = 1;
+    $absenceType->find();
```
davialexandre: The AbsenceType BAO has a `getEnabledAbsenceTypes` method that does this same thing

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+
+    $absenceTypes = [];
+    while($absenceType->fetch()) {
+      $absenceTypes[$absenceType->title] = $absenceType->id;
+      if ($absenceType->allow_accruals_request) {
+        $absenceTypes[$absenceType->title. ' (Credit)'] = $absenceType->id;
+      }
+    }
```
davialexandre: So we match the absence types from the CSV with the ones in the Database using their names. Is that right? If so, it has to be documented somewhere

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
```diff
@@ -0,0 +1,184 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+use CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+
+trait CRM_HRLeaveAndAbsences_Import_Helpers_DataImportTrait {
```
davialexandre: You're only using this trait once. Do we really need it? What is the reason that made you decide to use a trait instead of simply adding the method to the class itself?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * Returns the date_type Leave request Option values.
+   *
+   * @return LeaveRequestCommentService
+   */
+  private function getLeaveRequestCommentService() {
+    if (empty($this->leaveRequestCommentService)) {
+      $this->leaveRequestCommentService = new CRM_HRLeaveAndAbsences_Service_LeaveRequestComment();
```
davialexandre: You can use an alias to shorten the class name here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
```diff
@@ -0,0 +1,184 @@
...
+      $absenceType->is_active = 1;
+      $absenceType->find();
+
+      $sicknessAbsenceTypes = [];
+      while($absenceType->fetch()) {
+        if ($absenceType->is_sick) {
+          $sicknessAbsenceTypes[$absenceType->title] = $absenceType->id;
+        }
+      }
+      $this->sicknessAbsenceTypes = $sicknessAbsenceTypes;
```
davialexandre: What about adding a new method in the AbsenceType BAO that returns the is_sick absence types?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
```diff
@@ -0,0 +1,323 @@
...
+   *
+   * @param int $onDuplicate
+   *   The code for what action to take on duplicates.
+   * @param array $values
+   *   The array of values belonging to this line.
+   *
+   * @return bool
+   *   the result of this processing
+   */
+  public function import($onDuplicate, &$values) {
```
davialexandre: Can you explain here, in the method docblock (or wherever you think it is better), how the importer handles the records in the CSV file. Especially how it deals with the fact that each row is a separate date of a leave request.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
tunbola: Wasn't too sure about that and the method won't likely be useful now except for the Data import.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
tunbola: The trait has three or so methods that helps create leave request, balance changes and comments. Didn't want to clutter the parser file with those methods and only want the validation methods only in the parser file. What do you suggest?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
tunbola: Yeah, thats correct. Do we document that in a Wiki?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
tunbola: You meant `CRM_HRLeaveAndAbsences_Import_Parser_Base` ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
tunbola: Yeah, Inherited the code from the core activity entity. I think the reason why it does nothing is because the no validation takes place during the mapping field phase of the import. Will add that to the doc block.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
tunbola: Found the method used in all the Importers in core and in Job roles and Job contracts extension. just checked the method itself and it doesn't do much. Will just implement it in the parser file directly.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
davialexandre: In this case, the method is not about reusability, but more about encapsulating the logic of what is a "Sick" absence type. If we decide to change how we mark an absence type as sick, we would have to change all the places where the is_sick field is used directly.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
davialexandre: yeah, that's what I meant :D

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/BaseClass.php
davialexandre: I don't think we need a Wiki for that. Just documenting it in the method itself, at the moment you're importing the file, should be enough

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
```diff
@@ -0,0 +1,351 @@
...
+      if ($fieldType && $fieldType == CRM_Utils_Type::T_DATE) {
+
+        try {
+          new DateTime($value);
+        } catch (Exception $e) {
+          CRM_Contact_Import_Parser_Contact::addToErrorMsg('Invalid Date value for '.$fields[$key]['title'], $errorMessage);
+        }
+      }
+
+      if ($fieldType && ($fieldType == CRM_Utils_Type::T_FLOAT || $fieldType == CRM_Utils_Type::T_INT)) {
```
davialexandre: Only `if ($fieldType == CRM_Utils_Type::T_FLOAT || $fieldType == CRM_Utils_Type::T_INT)` is enough, isn't? Same applies to the previous if

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
```diff
@@ -0,0 +1,351 @@
...
+   * that identifies leave dates belonging to the same parent leave request.
+   * Information such as the start date and end date of the leave request is replicated across all the individual
+   * leave request dates for the leave request.
+   * The Leave request is created from the first individual leave request date encountered;
+   * Comments, Leave Dates for all the leave request dates and balance change
+   * for that particular leave request date is also created.
+   * Balance changes only will be created for subsequent rows of leave request dates for the already created leave request.
+   *
+   * If there is an error while creating the leave request from the first leave request date encountered,
+   * All other leave request dates will not be considered and the error reported is reported for all the leave dates.
```
davialexandre: Thanks for adding the description. I see that each line has a different lenght. Can you please make sure each that each one have +- the same length?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Form/DataSource.php
```diff
@@ -0,0 +1,79 @@
...
+   * Process the uploaded file.
+   */
+  public function postProcess() {
+    $this->storeFormValues(array(
+      'onDuplicate',
+      'dateFormats',
+      'savedMapping',
+    ));
+
+    $this->submitFileForMapping('CRM_HRLeaveAndAbsences_Import_Parser_Base');
```
davialexandre: Using `CRM_HRLeaveAndAbsences_Import_Parser_Base::class` instead will help you in the future, if you need to rename this class again or find where it's used

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Form/Preview.php
```diff
@@ -0,0 +1,170 @@
...
+    $onDuplicate = $this->get('onDuplicate');
+
+    $mapper = $this->controller->exportValue('MapField', 'mapper');
+    $mapperKeys = array();
+
+    foreach ($mapper as $key => $value) {
+      $mapperKeys[$key] = $mapper[$key][0];
+    }
+
+    $parser = new CRM_HRLeaveAndAbsences_Import_Parser_Base($mapperKeys);
```
davialexandre: Use the alias here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Helpers/DataImportTrait.php
davialexandre: Except for the methods used to fetch option values, all the others are specific about the import process and I doubt they can be reused anywhere else. In this case, I think it's better to remove the trait and add the method to the parser.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
tunbola: There is a null field value injected to the fields array on int `['' => ['title' => ts('- do not import -')]]` that does not have any properties aside for the title, thats the reason for checking if it the field type exists first.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -526,6 +526,23 @@ public static function getEnabledAbsenceTypes()
...
+  public static function getEnabledSicknessAbsenceTypes() {
+    $sicknessAbsenceTypes = [];
+    $absenceTypes = self::getEnabledAbsenceTypes();
+    foreach($absenceTypes as $absenceType) {
+      if ($absenceType->is_sick) {
+        $sicknessAbsenceTypes[] = $absenceType;
+      }
+    }
+
+    return $sicknessAbsenceTypes;
```
davialexandre: what about a slightly more concise version:

```php
public static function getEnabledSicknessAbsenceTypes() {
  $enabledAbsenceTypes = self::getEnabledAbsenceTypes();
  return array_filter($enabledAbsenceTypes, function($absenceType) {
     return $absenceType->is_sick;
  });
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -730,4 +730,16 @@ public function testNonExpiredToilRequestsAreDeletedAndExpiredToilRequestsNotDel
     $date = date('Y-m-d', strtotime('+1 day'));
     $this->assertEquals($leaveRequestDate->date, $date);
   }
+
+  public function testGetEnabledSicknessAbsenceTypes() {
+    $absenceType1 = AbsenceTypeFabricator::fabricate(['is_sick' => 1]);
```
davialexandre: Please add another test to check that this method only returns active/enabled items

uk.co.compucorp.civicrm.hrleaveandabsences/xml/Menu/import.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<menu>
+  <item>
+    <path>civicrm/admin/leaveandabsences/import</path>
+    <page_callback>CRM_HRLeaveAndAbsences_Import_Controller</page_callback>
+    <title>Import Leave/Absence Requests</title>
+    <access_arguments>administer leave and absences</access_arguments>
+  </item>
+</menu>
```
davialexandre: I only noticed this now. Is there any reason that this is on its own xml instead of in `xml/Menu/hrleaveandabsences.xml`?

Also, we'll need to add a link to this page in the menu. Can you please check with Mike where would be the best place for it?

tunbola: I noticed that all the Importers in Core and in hrjobcontract and hrjobroles extension have the xml file for the import separately. I only followed the format. I already added the link [HERE](https://github.com/civicrm/civihr/pull/1983/files#diff-62db90d66196e016278482a8f7c81df3R493) but will check with Mike

davialexandre: @tunbola about the xml file: yes, please add it directly to `xml/Menu/hrleaveandabsences.xml`. Unless there's a technical reason, there's no point in having a separate file just for that.

about the menu: sorry, I missed that. Unfortunately, adding it to the hook is not enough. Since the extension has now been released (as a beta, to compucorp) we need to start using upgraders for this kind of stuff. Otherwise the menu won't be available on existing sites

PR #1989, opened 2017-07-10
PCHR-2240: Improve calendar performance

AkA84:
# Overview
This PR improves the performance, the UX and the UI of the leave calendar.

## Performance improvements
*Note: most of the gifs/screenshots here are taken from my local machine which has a very big set of example data (+140 contacts) to highlight the performance issues*

### API calls: number, size, loading time
*(Manager calendar)*

#### Before
api-calls-old

#### After
api-calls-new

### First load time
*(Manager calendar)*

#### Before
loading-time-old

The data (`WorkPattern` calendars, `LeaveRequest`s) of *the whole period* is loaded from the start

#### After
loading-time-new

The data of just the currently selected month is loaded from the start

### Loading other months
*(Staff calendar)*

#### Before
individual-month-loading-old

The months appear immediately, because, as said early, the data for the entire period is being fetched from the start, causing the very poor performance on page load

#### After
individual-month-loading-new

The data for each month is now being loaded individually, when the month is shown for the first time (if the month is hidden and then shown again, the data is not reloaded)

### Filtering contacts by showing only those with a leave request
*(Manager calendar)*

#### Before
only-with-requests-old
*(The gif doesn't get stuck, it's the actual loading time in real time)*

Not seen in the gif: the contacts were filtered period-wide. Meaning that if someone had a leave request in Feb 2017 but not in May 2017, that person would still be listed in May

#### After
only-with-requests-new

Not seen in the gif: the contacts are now filtered on a month-by-monnth basis. Meaning that if someone has a leave request in Feb 2017 but not in May 2017, that person will be listed in Feb but not in May

### When there are many users to list
*(Manager calendar)*

#### Before
pagination-old

All the users are displayed at once. Not only it makes it hard to use the page (especially if there are multiple months displayed), but put AngularJS under strain, as the performance of the framework drops if there are too many DOM elements it needs to keep track of

#### After
pagination-new

The users list is now paginated, showing 20 users per page

#### Before
create-request-old

The whole calendar is refreshed

#### After
create-request-new

The leave request is added on the fly

#### Before
update-request-old

The whole calendar is refreshed

#### After
update-request-new

The leave request is updated on the fly

### Current day
current-day-new

The current day is now highlighted

### Days outside current period
disabled-days-new

The calendar will show the full months even if some of the days are outside the current period. Those days are styled differently from the others, and the cursor show a "not allowed" icon

---

intermission

---

The rest of the PR description contains the technical documentation

# Performance improvements
## Use `api.OptionGroup` improvement
See https://github.com/civicrm/civihr/pull/2026

The one-request-per-call and the bundling capability had been used in the calendar's refactoring to minimize the number of requests for OptionValues

## Skip pagination call in `api.getAll()` if `pagination` is not passed
See https://github.com/civicrm/civihr/pull/2026

Given that in the calendar we are fetching *all* the `Contact`s and `LeaveRequest`s at once, this saves 2 unnecessary requests

## Reduce size of `WorkPattern.getcalendar` endpoint's response
The response of the `WorkPattern.getcalendar` was massive when multiple contacts were passed to it. This is an example of the size and loading time of the response with +100 contacts

getcalendar-call-old

There were two reasons for this:

### 1) Duplicated data
The response contained duplicated data in the form of the `type.name` and `type.label` fields, as those are always the same for any contact+type couple:

```js
{
  "is_error": 0,
  "count": 2,
  "values": [
    {
      "contact_id": 1,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        },
        // more days here ...
      ]
    },
    {
      "contact_id": 2,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": {
            "value": 3,
            "name": "weeked",
            "label": "Weekend"
          }
        },
        // more days here ...
      ]
    }
    // more contacs here ...
  ]
}
```
This had been addressed by simplifying the response as such:

```js
{
  "is_error": 0,
  "count": 2,
  "values": [
    {
      "contact_id": 1,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": 3
        },
        // more days here ...
      ]
    },
    {
      "contact_id": 2,
      "calendar": [
        {
          "date": "2017-01-01",
          "type": 3
        },
        // more days here ...
      ]
    }
    // more contacs here ...
  ]
}
```

The complication with the above change was that the `CalendarInstance` service was internally expecting to receive the `type.name` field from the backend, in order to use it in its `isXYZ()` methods:

```js
function checkDate (searchedDate, dayType) {
  var searchedDate = findTheDateInternall();

  // Assumes `type.name` is already present!
  return searchedDate ? searchedDate.type.name === dayType : false;
}

return ModelInstance.extend({
  isWorkingDay: function (date) {
    return checkDate(date, 'working_day');
  },
});
```
To solve the problem, the `CalendarInstance` has been update to asynchronously fetch the `OptionValues` to match the `type.value` to a `type.name`

```js
var dayTypesPromise;

function checkDate (searchedDate, dayType) {
  return loadDayTypes()
    .then(function (dayTypes) {
      var dayTypeOptionValue = _.find(dayTypes, function (dayType) {
        return dayType.name === typeName;
      })

      return searchedDate ? dayTypeOptionValue.value === searchedDate.type : false;
    });
}

function loadDayTypes () {
  // The promise is created only once and reused
  // so that it doesn't make an api call with each method call
  dayTypesPromise = dayTypesPromise || OptionGroup.valuesOf('hrleaveandabsences_work_day_type');

  return dayTypesPromise;
}
```
This added an additional complication: because of the on-the-fly data fetching, the `isXYZ` methods are now asynchronous, while before they were synchronous:

```js
// before
calendarInstance.isWeekend(aWeekend) // true

// after
calendarInstance.isWeekend(aWeekend) // a promise
  .then(function (response) {
    response; // true
  });
```
This meant having to refactor any call to those methods in the whole L&A codebase

### 2) Returning entire periods worth of data
The other reason for the response's size was that the endpoint accepted a period id as the only way to filter the data, meaning that the endpoint consumer was forced to get either 12 (or more) months worth of data, or nothing.

With the work done in https://github.com/civicrm/civihr/pull/2013, the `period_id` filter had been removed in favour of a `start_date` and `end_date` filters, making possible to get data for individual months.

Changes in the api and model services were made to match the new endpoint signature.

### End result
The size of the request is now much smaller and the time is greatly reduced

getcalendar-call-new

## Use `ContactInstance` directly
In the manager calendar we need to fetch the staff members that the currently logged in user manages, and we were doing that by first fetching the user's via the `Contact` model, and then calling `loadManagees()` on the `ContactInstance` returned by the model, for a total of two api calls:

```js
Contact.find(vm.contactId)          // 1st call
  .then(function (contact) {
    return contact.leaveManagees()  // 2nd call
  });
```
Given that we don't really need the current user's data, but really just an instance with his id to call `.leaveManagees()`, we can simply initialize a `CalendarInstance` directly:

```js
ContactInstance.init({ id: vm.contactId }).leaveManagees(); // 1 call in total
```

## Use `ng-if` instead of `ng-hide`
All the months were added to the DOM immediately on page load regardless of whether they were shown or not, making AngularJS struggle with trying to digest the data of all those DOM elements:

```html
<section ng-repeat="month in calendar.months">
  <!-- ng-hide does not remove from the DOM -->
  <ng-include ng-hide="ifNotVisible" src="manager-leave-calendar-month.html'"></ng-include>
</section>
```
This caused the page (and the browser) to be extremely slow and even grinding to a halt at times in case there were many contacts displayed

By using `ng-if` (which removes from the DOM if the condition is not met), we ensure that only the months that need to be shown to the user are present on the page, making life for AngularJS a bit easier

# `<leave-calendar[-x] />` components
Before this PR we actually had two calendar components: [`<staff-leave-calendar />`](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-calendar.js) and [`<manager-leave-calendar />`](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js). The two shared a common controller, [`CalendarCtrl`](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/calendar-ctrl.js), that they would extend with their custom methods.

```js
// staff-leave-calendar.js
components.component('staffLeaveCalendar', function controller () {
  var parentCtrl = $controller('CalendarCtrl');
  var vm = Object.create(parentCtrl);
});

// manager-leave-calendar.js
components.component('managerLeaveCalendar', function controller () {
  var parentCtrl = $controller('CalendarCtrl');
  var vm = Object.create(parentCtrl);
});

```

The problem with this set up is that it lead to duplication and many instances of "not the same, but quite", most notably:

### Templates
The two calendars had two [separate](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-calendar.html) [sets](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-calendar-month.html) [of](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/manager-leave-calendar.html) [templates](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/manager-leave-calendar-month.html) (each set had a calendar tpl and a month tpl). The differences were minimal, with the manager calendar showing a filters panel on top and the employees name on each line, while the staff calendar didn't. The rest was the same (overall page, month and day markup, legend, etc)

### View's data structure
The calendar's view's data structure between the two calendars was the same, but the way that data was stored differed.

#### The structure
In order to display the calendar correctly, it is necessary to get the `WorkPattern` calendar of each contact:

```js
Calendar.get([theContacts]), thePeriod); // old signature
```

The response is an array of objects, each object representing a contact's `WorkPattern` calendar, with a `days` property, which is a list of timestamp-indexed days:
```js
[
  {
    contact_id: '204',
    days: {
      <some-timestamp>: {
        date: '2017-01-01',
        type: 3 // The hrleaveandabsences_work_day_type OptionValue, see above
      },
      <some-timestamp>: {
        date: '2017-01-02',
        type: 2
      },
      // more days here...
    }
  }
  // more calendars here...
]
```

Once the calendars had this list of `CalendarInstance`s, they would loop through them and for each day they would attach a `UI` property to it, which would contain all the information necessary to correctly display their table-cell on the month, plus a `leaveRequest` property referencing a `LeaveRequestInstance` if there was one on that day

```js
  <some-timestamp>: {
    date: '2017-01-01',
    type: 3, // The hrleaveandabsences_work_day_type OptionValue, see above
    leaveRequest: {
      // leave request data..
    },
    UI: {
      isAM: true,
      isNonWorkingDay: false,
      isPM: false,
      isPublicHoliday: false,
      isRequested: true,
      isWeekend: false,
      styles: {
        backgroundColor: '#B32E2E',
        borderColor: '#B32E2E'
      }
    }
  }
]
```
This is how the calendar data was structured (everything in the `calendar.day` objects) in both calendars

#### How it was stored
This is where the calendars differed, mainly because the staff calendar had to manage only one contact, while the manager calendar had to deal with multiple contacts.

The staff calendar store a reference of each of the `calendar.day`s in a `data` property in each `month` object:

```js
// vm.months[0]; --> January

{
  data: [
    // this is the same object as above
    {
      date: '2017-01-01',
      type: 3,
      leaveRequest: // ...
      UI: // ...
    }
  ],
  loading: false,
  index: 0,
  year: 2017
}

```
The idea being that since there is only one contact's data to display, the data should be simply indexed by month.

The manager calendar instead first stored the months list in a `calendarData` property in each `contact` object, and then each month got its `calendar.day` objects stored like in the staff calendar:

```js
// vm.contacts[0]; --> One of the contacts to display in the calendar

{
  display_name: 'Mr. Foo Bar',
  calendarData: [
    // january
    {
      data: /* `calendar day`s here... */,
      loading: false,
      index: 0,
      year: 2017
    }
    // other months here...
  ],
  // more contact's data here
}
```
The idea being that since there are multiple contacts to display, it should be then indexed first by contact and then by month.

### Business logic
The two calendars had very similar (but not shared) business logic, a couple of notable examples:

* How to attach `UI` properties to day: [staff](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-calendar.js#L109) / [manager](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js#L281)
* How to delete a leave request: [staff](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-calendar.js#L151) / [manager](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js#L338)
* Hot to refresh the calendar: [staff](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-calendar.js#L46) / [manager](https://github.com/civicrm/civihr/blob/65d9147096729be1682d73324dce8d8d628b158e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js#L89)

---

Given all the above, in order to work on performance optimization and to have a maintainable solution, the two calendars had to be refactored and unified in a single, configurable, calendar. This in the end lead to the following:

## `<leave-calendar />`
```html
<leave-calendar contact-id="contactId" [role-override="forcedRole"]></leave-calendar>
```

The parent component. Its responsibilities are:
* Determine the role of the current user (or use the one given via the `role-override` binding)
* Inject a sub-controller based on the user's role (see next section)
* Load all the basic data (`Contact`s, `AbsencePeriod`s, `PublicHoliday`s, etc)
* Set the current `AbsencePeriod`
* Set the current month
* Pass some of the basic data to the `<leave-calendar-month />` components (see next component)
* Manage the injection of the `<leave-calendar-month />` components in the DOM
* Communicate to the `<leave-calendar-month />` which ones should be visible
* Manage the filters (period / months / contacts)
* Refresh the calendar

### Sub-controllers
Instead of having two calendar components, there is only one component that internally injects a sub-controller based on the user's role

```js
(function init () {
  setUserRole().then(injectSubController)
}())

function injectSubController () {
  subController = $controller('LeaveCalendar' + userRole + 'Controller').init(vm);
}
```

There are three sub-controllers: [LeaveCalendarStaffController](https://github.com/civicrm/civihr/blob/8686b8ce9218462723219ea081ddcd9d40faa33e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-staff.controller.js), [LeaveCalendarManagerController](https://github.com/civicrm/civihr/blob/8686b8ce9218462723219ea081ddcd9d40faa33e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-manager.controller.js), [LeaveCalendarAdminController](https://github.com/civicrm/civihr/blob/8686b8ce9218462723219ea081ddcd9d40faa33e/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-admin.controller.js)

The role of a sub-controller is to provide a specific list of contacts for the component to use (notice that now the list of contacts is normalized: it's always an array, even if the user is only a staff member), and also to tweaks some of the calendar settings if necessary (possible because the component passes its context to the sub-controller).

For example when the user is a manager he should see the contact names and the filters:

```js
// LeaveCalendarManagerController

return {
  init: function (_vm_) {
    vm.showContactName = true;
    vm.showFilters = true;

    return api();
  }
};
```

## `<leave-calendar-month />`
```html
<leave-calendar-month
  contacts="contacts"
  month="month"
  period="currentPeriod"
  show-contact-name="true"
  show-only-with-leave-requests="false"
  support-data="supportData">
</leave-calendar-month>
```
This is the component that does most of the work. Its responsibilities are:
* Load the `Calendar`s and `LeaveRequest` of the given contacts. The data will be limited to the month (so if the component represents January, it will load only the `Calendar`s and `LeaveRequest`s of January)
* Create the view's data structure
* Listen for the `<leave-calendar />` events to know when it should become visible
* Listen for a `LeaveRequest`'s create/update/delete events, so it can update the view on the fly
* Paginate the list of contacts if necessary

### Data structure
The data structure is now always the same, no matter what the user's role is:

#### Month
The outer "layer" of the data structure, it contains meta-information about the month itself, a flag to indicate if the data is still loading, and then the list of all the days of the month

```js
// vm.month

{
  index: 0,
  year: 2017,
  name: 'Jan',
  loading: true,
  days: // list of days
}

```

The list of days is created via MomentJS at the moment of the component's initialization:

```js
function buildMonthDaysStructure (dateMoment) {
  var pointerDay = dateMoment.clone().startOf('month');

  return _.map(_.times(dateMoment.daysInMonth()), function () {
    var day = // day structure

    pointerDay.add(1, 'day');

    return day;
  });
}
```

#### Day
Each day contains meta-information, whether if it's the current day or not, whether if it's enabled or not (it could be the day is outside the current period's date range), and most importantly it contains the data that each contact data has on the day:

```js
vm.month.days[2]

{
  date: '2017-01-02',
  name: 'Fri',
  index: 1,
  current: false,
  enabled: true,
  contactsData: // the data about the day for each of the contacts
}
```

#### Contact data
This is where all the information about a day for each contact is stored. A day for a contact could be a working day, for another a non-working day. A day could be taken as leave for a contact, as half-day leave for another, as a sickness day for another, and so on. This is where all this diversity of data is stored.

The data is indexed by contact id for easy access:

```js
// vm.month.days[2].contactsData[someContactId]

{
  isWeekend: false,
  isNonWorkingDay: false,
  isPublicHoliday: true,
  isAccruedTOIL: false,
  isRequested: true,
  isAM: false,
  isPM: true,
  leaveRequest: // reference to a leave request,
  styles: {
    backgroundColor: '#B32E2E',
    borderColor: '#B32E2E'
  },
}
```

### Communication between `<leave-calendar />` and `<leave-calendar-month />`
If you look in the `<leave-calendar>`'s template, you'll see that all the month components are actually injected in the DOM as soon as the `calendar.injectMonths` flag is set to `true`

```html
<!-- leave-calendar.html -->
<div ng-if="calendar.injectMonths">
  <leave-calendar-month ng-repeat="month in calendar.months track by $index"
    contacts="calendar.contacts"
    month="month"
    period="calendar.selectedPeriod"
    show-contact-name="calendar.showContactName"
    show-only-with-leave-requests="calendar.filters.userSettings.contacts_with_leaves"
    support-data="calendar.supportData">
  </leave-calendar-month>
</div>
```
 The reason for this is simple: the month components need to always be in the DOM in order to receive data from the parent component and to listen for the "show months" event. If the months were injected in the DOM only when they were supposed to be visible, they wouldn't be able to listen constantly for messages from `<leave-calendar />`

 What actually *doesn't* get injected in the DOM is the `<leave-calendar-month />`'s content, which is hidden under a `visible` flag:

 ```html
<!-- leave-month.html -->
 <section class="panel panel-default" ng-if="month.visible">
 </section>
 ```

 The `visible` flag is set to `true` when a month receives the "show month" from the parent component and see that it's in the list of months that should be visible. This way the component is always on the DOM, but not its markup, allowing AngularJS to manage as least as DOM elements as possible

The flag `calendar.injectMonths` is set to true by `<leave-calendar />` only when all the months of the currently selected period are injected in the DOM:

```js
(function init () {
  setUserRole()
  .then(injectSubController)
  .then(loadAllData)
  .then(injectAndShowMonths) // <- injects the months
  .then(function () {
    vm.loading.page = false;
  });
}());

function injectAndShowMonths () {
  waitUntilMonthsAreInjected().then(function () {
    sendShowMonthsSignal();
  });
}
```
The way components that all the months are injected, is by listening for a signal that each month sends as soon as it is initialized. Once the component gets as many signals as they are months in the currently selected period, it knows it can then show the months:
```js
function waitUntilMonthsAreInjected () {
  return $q(function (resolve) {
    var monthLoadedCounter = 0;
    var removeListener = $rootScope.$on('LeaveCalendar::monthInjected', function () {
      if (++monthLoadedCounter === vm.months.length) {
        removeListener();
        resolve();
      }
    });
  });
}
```

## `<leave-calendar-day />`
```html
<leave-calendar-day contact-data="someContactData"></leave-calendar-day>
```
It represents one of the table-cell in the calendar, it's only responsibilities is to correctly display the given day's contact's data

## `<leave-calendar-legend />`
```html
<leave-calendar-legend absence-types="absenceTypes"></leave-calendar-legend>
```
It's the calendar's legend, the only responsibilites to correctly display the various table-cell styles

# Misc improvements
1) Whenever possible the files had been renamed to comply with the AngularJS Style Guide
2) The mocked `WorkPattern` data is now using the mocked `Contact` data, as the whole mocked data should be as internally coherent as possible
```js
// work-pattern-data.js

return {
  getCalendar: {
    'is_error': 0,
    'version': 3,
    'count': 2,
    // assigning a mocked calendar to every mocked contact
    'values': ContactData.all.values.map(function (contact) {
      return // mocked calendar
    }
  }
}
```

# Comments
There are still many improvements that could be made, for example

1) All the components could be moved under a `leave-calendar` module, so to make a single bundle that apps can require via RequireJS and declare as dependency
2) The manager/admin calendar still could shave off some seconds of api calls by improving how it filters the managed contacts
3) The calendar filters could be a separate component (and also currently the filtering is done in the `LeaveCalendarManagerController`, while it should be done in the `<leave-calendar >` component)

* [x] Tests passed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js
```diff
@@ -316,39 +210,114 @@ define([
...
-            });
+            loadOptionValues(),
+            loadManagees()
+          ]);
         });
       });
+
       $rootScope.$on('LeaveRequest::updatedByManager', vm.refresh);
       $rootScope.$on('LeaveRequest::deleted', deleteLeaveRequest);
+    })();
```
igorpavlov: init();

function init () { ... }

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manager-leave-calendar.js
```diff
@@ -316,39 +210,114 @@ define([
...
+      });
+    }
+
+    /**
+     * Returns the filter object for contacts api
+     *
+     * @return {Object}
+     */
+    function prepareContactFilters () {
+      return {
```
igorpavlov: **var filters = vm.filters;**
...
**filters**.department ? **filters**.department.value : **undefined**

I think it is safer to send `undefined` as it means the filter should be completely _ignored_, but not that the filter's value is NULL

davialexandre:
@AkA84 the backend changes look good. I have questions though:

Are there any memory usage improvements after these changes? I believe that, since we're loading a lot less data now when the page loads, the memory consumption has drastically decreased. But what happens when the user decides to view every month and every contact at once? Also, if I understand things correctly, whenever a month is hidden from the calendar it will also be removed from the DOM, which I believe should free some memory, right?

Can you also give some more details about the reasons for the commit reverting the 1.6 commits? I'm not sure I understand where the 1.6 stuff came from.

Other than that, impressive work here. Good job!

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
```diff
@@ -53,8 +53,9 @@ define([
 
     describe('getCalendar()', function () {
       var workPatternPromise;
-      var dummyContactId = 1;
-      var dummyPeriodId = 2;
+      var contactId = 1;
```
igorpavlov: contactId in other code places is a String as far as I remember, so should it be a string here as well? Should we check for a consistency in the code, test and mocks?

Side note: we also use toEqual, which should be avoided I think since toBe is more secure.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
```diff
@@ -103,7 +105,7 @@ define([
       });
 
       describe('when multiple contact ids are passed', function () {
-        var multipleContacts = [dummyContactId, dummyContactId + 1];
+        var multipleContacts = [contactId, contactId + 1];
```
igorpavlov: If we a going to change contactId to string, then this should be 
`var multipleContacts = [contactId, (parseInt(contactId) + 1).toString()];`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-day.component.spec.js
```diff
@@ -0,0 +1,28 @@
...
+define([
+  'leave-absences/shared/components/leave-calendar-day.component'
+], function () {
+  'use strict';
+
+  describe('leaveCalendarDay', function () {
+    var $componentController, $log;
+
+    beforeEach(module('leave-absences.templates', 'leave-absences.components'));
+    beforeEach(inject(function (_$componentController_, _$log_) {
```
igorpavlov: Should there be a line break? I realise these are 2 function calls, but we separate jasmine calls with an empty line usually?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-legend.component.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  'mocks/data/absence-type-data',
+  'leave-absences/shared/components/leave-calendar-legend.component'
+], function (_, AbsenceTypeData) {
+  'use strict';
+
+  describe('leaveCalendarLegend', function () {
+    var $componentController, $log, controller, mockedAbsenceTypes;
+
+    beforeEach(module('leave-absences.templates', 'leave-absences.components'));
+    beforeEach(inject(function (_$componentController_, _$log_, _$rootScope_) {
```
igorpavlov: Same as above.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -0,0 +1,695 @@
+/* eslint-env amd, jasmine */
+
+(function (CRM) {
```
igorpavlov: Please unwrap it from CRM.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -0,0 +1,695 @@
...
+    'mocks/apis/work-pattern-api-mock',
+    'leave-absences/my-leave/app'
+  ], function (_, moment, helper, ContactData, AbsencePeriodData, AbsenceTypeData, LeaveRequestData, OptionGroupData, PublicHolidayData, WorkPatternData) {
+    'use strict';
+
+    describe('leaveCalendarMonth', function () {
+      var $componentController, $log, $provide, $q, $rootScope, Calendar,
+        LeaveRequest, OptionGroup, controller, daysInFebruary, february, leaveRequestInFebruary,
+        period2016, publicHolidays;
+      var currentContactId = CRM.vars.leaveAndAbsences.contactId;
```
igorpavlov: Please set it to '204' or smth instead of CRM.vars....

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -0,0 +1,695 @@
...
+      it('has the show-contact-name binding set to false by default', function () {
+        expect(controller.showContactName).toBe(false);
+      });
+
+      it('has the show-only-with-leave-request name binding set to false by default', function () {
+        expect(controller.showOnlyWithLeaveRequests).toBe(false);
+      });
+
+      describe('on init', function () {
+        it('does not load the contacts\' work pattern calendars', function () {
```
igorpavlov: Please avoid using ' because escaping does not look good. `does not load work pattern calendars of the contacts`. Same for other places in the code.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -0,0 +1,695 @@
...
+            dayTypes: OptionGroupData.getCollection('hrleaveandabsences_leave_request_day_type'),
+            leaveRequestStatuses: OptionGroupData.getCollection('hrleaveandabsences_leave_request_status'),
+            publicHolidays: publicHolidays
+          }
+        });
+
+        !!sendSignal && sendShowMonthsSignal();
+      }
+
+      function sendShowMonthsSignal (includeFebruary, forceReload) {
```
igorpavlov: Please add a description comment to function.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
```diff
@@ -0,0 +1,496 @@
+/* eslint-env amd, jasmine */
+
+(function (CRM) {
```
igorpavlov: Please unwrap from CRM

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
```diff
@@ -0,0 +1,496 @@
...
+  ], function (angular, moment, _, helper, absencePeriodData, absenceTypeData, optionGroupMock, publicHolidayData) {
+    'use strict';
+
+    describe('leaveCalendar', function () {
+      var $componentController, $controller, $controllerProvider, $log, $q,
+        $rootScope, $timeout, controller, $provide, AbsencePeriod, OptionGroup,
+        PublicHoliday, sharedSettings;
+      var mockedCheckPermissions = mockCheckPermissionService();
+      var currentContact = {
+        id: CRM.vars.leaveAndAbsences.contactId,
```
igorpavlov: Please `id: '204'` or smth.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
```diff
@@ -0,0 +1,496 @@
...
+
+          spyOn($log, 'debug');
+          spyOn($rootScope, '$emit').and.callThrough();
+          spyOn(AbsencePeriod, 'all');
+          spyOn(PublicHoliday, 'all').and.callThrough();
+          spyOn(OptionGroup, 'valuesOf').and.callFake(function (name) {
+            return OptionGroupAPIMock.valuesOf(name);
+          });
+
+          // Set 2016 as current period, because Calendar loads data only for the current period initially,
```
igorpavlov: Please use multiline comment styling.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
```diff
@@ -0,0 +1,496 @@
...
+          // and MockedData has 2016 dates
+          amend2016Period({ current: true });
+
+          compileComponent();
+        }
+      ]));
+
+      // The mocked "work pattern calendar" and "leave request" data is made for
```
igorpavlov: Same as above.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
```diff
@@ -0,0 +1,496 @@
...
+                'LeaveCalendar::showMonths',
+                jasmine.any(Array),
+                true
+              );
+            });
+          });
+        });
+      });
+
+      function amend2016Period (params) {
```
igorpavlov: Please add description comment to those functions (except compileComponent that should be at the very end?)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-admin.controller.spec.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-env amd, jasmine */
+
+(function (CRM) {
```
igorpavlov: Please unwrap from CRM

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-admin.controller.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  define([
+    'leave-absences/my-leave/app'
+  ], function () {
+    'use strict';
+
+    describe('LeaveCalendarAdminController', function () {
+      var $controller, $log, Contact, controller;
+      var contactId = CRM.vars.leaveAndAbsences.contactId;
```
igorpavlov: `= '204'` or smth

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-manager.controller.spec.js
```diff
@@ -0,0 +1,78 @@
+/* eslint-env amd, jasmine */
+
+(function (CRM) {
```
igorpavlov: Please unwrap from CRM

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-manager.controller.spec.js
```diff
@@ -0,0 +1,78 @@
...
+  define([
+    'leave-absences/my-leave/app'
+  ], function () {
+    'use strict';
+
+    describe('LeaveCalendarManagerController', function () {
+      var $controller, $log, $provide, $rootScope, Contact, ContactInstance, controller, realContactInstance;
+      var contactId = CRM.vars.leaveAndAbsences.contactId;
```
igorpavlov: `= '204'` or smth

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-manager.controller.spec.js
```diff
@@ -0,0 +1,78 @@
...
+        it('filters the contact using the filters selected by the user', function () {
+          expect(Object.keys(Contact.all.calls.mostRecent().args[0])).toContain('department');
+          expect(Object.keys(Contact.all.calls.mostRecent().args[0])).toContain('level_type');
+          expect(Object.keys(Contact.all.calls.mostRecent().args[0])).toContain('location');
+          expect(Object.keys(Contact.all.calls.mostRecent().args[0])).toContain('region');
+          expect(Object.keys(Contact.all.calls.mostRecent().args[0])).toContain('id');
+        });
+      });
+
+      function spyOnContactInstance () {
```
igorpavlov: Please add description comment.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-staff.controller.spec.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-env amd, jasmine */
+
+(function (CRM) {
```
igorpavlov: Please unwrap from CRM

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-staff.controller.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  define([
+    'leave-absences/my-leave/app'
+  ], function () {
+    'use strict';
+
+    describe('LeaveCalendarStaffController', function () {
+      var $controller, $log, Contact, controller;
+      var contactId = CRM.vars.leaveAndAbsences.contactId;
```
igorpavlov: `id = '204'` or smth

igorpavlov:
Please update the names of all touched tests to `.spec.js`

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -68,8 +53,23 @@ td.chr_leave-calendar__name {
...
 }
 
-//Public Holiday, Weekend and Non Working day needs to be specified before other styles
-//to give other styles more priority
+.chr_leave-calendar__item--disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+}
+
+// Public Holiday, Weekend and Non Working day needs to be specified before other styles
```
igorpavlov: Multiline comments styling please.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -68,8 +53,23 @@ td.chr_leave-calendar__name {
   }
 }
 
-//Public Holiday, Weekend and Non Working day needs to be specified before other styles
-//to give other styles more priority
+.chr_leave-calendar__item--disabled {
```
igorpavlov: I think we have `.chr_disabled{opacity:0.5;pointer-events:none}` - will that work? Or do you really need cursor: not-allowed; for UX?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -68,8 +53,23 @@ td.chr_leave-calendar__name {
...
+  cursor: not-allowed;
+  opacity: 0.5;
+}
+
+// Public Holiday, Weekend and Non Working day needs to be specified before other styles
+// to give other styles more priority
+.chr_leave-calendar__item--type--late {
+  background-color: #eca67e;
```
igorpavlov: Color should be stored in a variable?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -68,8 +53,23 @@ td.chr_leave-calendar__name {
...
+  border-color: #eca67e;
+}
+
+.chr_leave-calendar__item--type--non-working {
+  background-color: #41afcb !important;
```
igorpavlov: Same as above

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -5,48 +5,33 @@ $chr-leave-calendar-item-requested-border-height: 0.5em;
...
 }
 
-.chr_leave-calendar__items {
-
-  > td {
-    overflow: hidden;
-    height: $chr-leave-calendar-cell-height;
-    padding: 0 !important;
-  }
+.chr_leave-calendar__day--disabled {
```
igorpavlov: I think we have .chr_disabled{opacity:0.5;pointer-events:none} - will that work?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendarTest.php
```diff
@@ -57,9 +52,18 @@ public function testGetShouldUseTheDefaultWorkPatternIfTheContactHasNoActiveWork
       'effective_date' => CRM_Utils_Date::processDate('2015-02-05'),
     ]);
 
-    $calendar = new WorkPatternCalendarService($this->contact['id'], $absencePeriod, $this->jobContractService);
+    //15 days date interval.
```
igorpavlov: Add a space after //

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendarTest.php
```diff
@@ -412,11 +430,19 @@ public function testGetReturnsResultsForContactWithoutAContractAndFillsTheContac
       'effective_date' => CRM_Utils_Date::processDate('2016-03-07'),
     ]);
 
+    $startDate = '2016-03-03';
+    $endDate = '2016-03-09';
+    $datePeriod = new BasicDatePeriod($startDate, $endDate);
+
     //A contract is fabricated for the period between 2016-03-03 and 2016-03-09
```
igorpavlov: Add a space please after //

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendarTest.php
```diff
@@ -441,18 +467,20 @@ public function testGetReturnsResultsForContactWithoutAContractAndFillsTheContac
...
-      'start_date' => CRM_Utils_Date::processDate('2016-03-03'),
-      'end_date' => CRM_Utils_Date::processDate('2016-03-06'),
-    ]);
-
     WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => true]);
+    $startDate = '2016-03-03';
+    $endDate = '2016-03-06';
+    $datePeriod = new BasicDatePeriod($startDate, $endDate);
 
     //A contract is fabricated period between 2016-03-03 and 2016-03-06
```
igorpavlov: Please add a space after //

igorpavlov:
Please update the names of all touched tests to .spec.js

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+        contact: null,
+        contacts_with_leaves: false,
+        department: null,
+        level_type: null,
+        location: null,
+        region: null
+      }
+    };
+
+    vm.labelPeriod = labelPeriod;
```
AkA84: See [here](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS+Style+Guide#AngularJSStyleGuide-FunctionDeclarationstoHideImplementationDetails)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+    vm.refresh = refresh;
+
+    (function init () {
```
AkA84: See [here](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS+Style+Guide#AngularJSStyleGuide-ControllerActivationPromises) (my fault for changing things at the last minute)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+      })
+      .then(function () {
+        vm.loading.page = false;
+      });
+    }());
+
+    /**
+     * Creates a list of all the months in the currently selected period
+     */
+    function buildPeriodMonthsList () {
```
AkA84: Actually there was a different flaw here: if a period started on 20/3/2017 and ended on 2/4/2017, the months list would have been only "March" instead of "March, April"

I fixed this now

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+     *   regardless if it had already loaded it
+     */
+    function sendShowMonthsSignal (forceDataReload) {
+      var monthsToShow = !vm.selectedMonths.length
+        ? vm.months
+        : vm.months.filter(function (month) {
+          return _.includes(vm.selectedMonths, month.index);
+        });
+
+      $rootScope.$emit('LeaveCalendar::showMonths', monthsToShow, !!forceDataReload);
```
AkA84: That's the point, it's to cater in case the function is mis-called with the wrong parameters. I don't think that making sure that the param is passed boolean is a big deal

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    $log.debug('Component: leave-calendar-month');
+
+    var dataLoaded = false;
+    var calendars = {};
+    var leaveRequests = {};
+
+    var vm = this;
+    vm.currentPage = 0;
+    vm.pageSize = 20;
+    vm.visible = false;
```
AkA84: well it means that the month is visible 
```html
<section class="panel panel-default" ng-if="month.visible">
```
you have any suggestion on how to make it more meaningful?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    vm.showContactName = vm.showContactName ? !!vm.showContactName : false;
+    vm.showOnlyWithLeaveRequests = vm.showOnlyWithLeaveRequests ? !!vm.showOnlyWithLeaveRequests : false;
+
```
AkA84: Please check the [angular style guide](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS+Style+Guide)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    vm.contactsList = contactsList;
```
AkA84: Please check the [angular style guide](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS+Style+Guide)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+
+    (function init () {
```
AkA84: Please check the [angular style guide](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS+Style+Guide)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+          userRole = results[0] ? 'admin' : (results[1] ? 'manager' : 'staff');
+        });
+      }
+    }
+
+    /**
+     * Waits for the next digest cycle to make sure that all the data
+     * from the component had been transmitted to the child components
+     */
+    function waitForNextDigest () {
```
AkA84: I improved the logic now and got rid of the dependency on the angular digest cycle

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+      $rootScope.$emit('LeaveCalendar::monthInjected');
+    }());
+
+    /**
+     * Adds a leave request to the calendar
+     *
+     * @param {Object} event
+     * @param {LeaveRequestInstance} leaveRequest
+     */
+    function addLeaveRequest (event, leaveRequest) {
```
AkA84: Using `__` to indicate unused params (maybe I should add this to the styleguide.

It is common to use `_` in JS/Ruby (https://stackoverflow.com/questions/32197927/standard-conventions-for-indicating-a-function-argument-is-unused-in-javascript), but given we have `lodash` then the double underscore avoids conflicts

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    /**
+     * Decides whether sent date is a public holiday
+     *
+     * @param  {string} date
+     * @return {boolean}
+     */
+    function isPublicHoliday (date) {
+      return !!vm.supportData.publicHolidays[dateObjectWithFormat(date).valueOf()];
```
AkA84: Not the same thing, `_.includes` is to finding items in an array, here `vm.supportData.publicHolidays` is an object

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+
+      return $q.all([
+        loadMonthWorkPatternCalendars(),
+        loadMonthLeaveRequests()
+      ])
+      .then(function () {
+        return setMonthDaysContactData();
+      })
+      .then(function () {
+        dataLoaded = true;
```
AkA84: the first is a flag to see if the month's data is already loading, the second is a boolean used in the View (as indicated by the fact that it's on the scope)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    /**
+     * Loads the approved/pending leave requests of the month, limited
+     * to the calendar contacts. It then indexes the leave requests
+     *
+     * @return {Promise}
+     */
+    function loadMonthLeaveRequests () {
+      return LeaveRequest.all({
```
AkA84: why? i don't think it improves readibility in this case

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+      });
+    }
+
+    /**
+     * Loads the month's calendar of each contact
+     *
+     * @return {Promise}
+     */
+    function loadMonthWorkPatternCalendars () {
+      var monthStartDate = vm.month.days[0].date;
```
AkA84: why? i don't think it improves readibility in this case

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    /**
+     * Show the month and its data if it's included in the given list
+     *
+     * @param  {Object} event
+     * @param  {Array} monthsToShow
+     * @param  {Boolean} forceReload If true it forces the reload of the data
+     */
+    function showMonthIfInList (event, monthsToShow, forceReload) {
```
AkA84: see above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-manager.controller.js
```diff
@@ -0,0 +1,86 @@
...
+          return contacts.list;
+        });
+    }
+
+    /**
+     * Returns the filter object for contacts api
+     *
+     * @return {Object}
+     */
+    function prepareContactFilters () {
```
AkA84: We need to send `null` to the api in order to make it ignore the filter

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/calendar-model.js
```diff
@@ -20,14 +20,15 @@ define([
          * as a list of days and their type
          *
          * @param {string/int/Array} contactId can be also an array for multiple contacts
```
AkA84: Sorry, I'm afraid that changing either the implementation or the places where the method is called will cause a chain of changes that I just can't afford in this PR. It is a good point, but I won't risk increasing the scope of the PR

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
AkA84: > Side note: we also use toEqual, which should be avoided I think since toBe is more secure.

Can you point out the line

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-day.component.spec.js
AkA84: Honestly i'm still on the fence about this, i don't know why sometimes we have separate lines between the various beforeEach at the beginning

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
AkA84: Why? It's used here
```js
var currentContactId = CRM.vars.leaveAndAbsences.contactId;
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
AkA84: But why? We're mimicking the production environment, what's wrong with this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
AkA84: used double quotes instead

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
AkA84: same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
AkA84: same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
AkA84: I'm not sure i agree with this in the end. The multiline comment is usually used either to serve as some kind of header or to contain a *very long* and structured text. This is a simple comment that spans two line, that's all. It think it's better to use `//` as to mark the fact that it's a standard comment

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
AkA84: same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-calendar-admin.controller.spec.js
AkA84: Same as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
AkA84: This applies to all other instances of this comment

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
AkA84: This applies to all other instances of this comment

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
AkA84: No, we need something specific for the day cell

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
AkA84: Yep, i need the cursor changed as well

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
AkA84: Be kind and don't make me increase the scope of this PR even a little bit please

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
AkA84: Same begging as above

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: @AkA84 you are right, no need to change it.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -0,0 +1,351 @@
...
+    function setDefaultMonths () {
+      vm.selectedMonths = [moment().month()];
+    }
+
+    /**
+     * Sets the user's role based on his permissions
+     */
+    function setUserRole () {
+      if (vm.roleOverride) {
+        return $q.resolve().then(function () {
```
igorpavlov: Shame it doesn't work.

igorpavlov:
@AkA84 Alessandro, the fixes look good to me. I am happy to approve but after I have a chance to review how conflicts are resolved.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
igorpavlov: Sure.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: I think it will since there is some repetition. This is subjective though.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: I think it will since there is some repetition. This is subjective though.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-manager.controller.js
igorpavlov: @AkA84 OK, just FYI then - this might be handy to look at https://github.com/civicrm/civihr/pull/2023

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/calendar-model.js
igorpavlov: OK.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/apis/work-pattern-api_test.js
igorpavlov: Line 51 `expect(isolateScope.contactId).toEqual(+contactId);`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-day.component.spec.js
igorpavlov: OK, good to know.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
igorpavlov: I don't see a point of `CRM.vars.leaveAndAbsences.contactId;`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
igorpavlov: It should be assigned in globals.js, yes. But I don't see a point of use in tests.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
igorpavlov: OK

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar.component.spec.js
igorpavlov: OK. It just looks better I think, if we follow 80 chars rule (and we should).

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
igorpavlov: Got it.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
igorpavlov: Got it.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
igorpavlov: Sure. Apologies.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
igorpavlov: Apologies.

igorpavlov:
Merged with la-milestone-3, resolved conflicts, supported new Leave Request Popup service in the Calendar, added tests.

PR #1992, opened 2017-07-11
PCHR-2366: Fix entitlements comment popup

igorpavlov:
## Overview

There are 3 issues with Annual Entitlement comments:

- Author's name is incorrect
- Date/time is wrong as per different timezones
- The separator is wrong ("." instead of ":")

This PR fixed the issues above.

## Technical Details

1. Back-end fixed: the global `$user` variable was used only in this file in L&A. Its `$user->uid` property was *not Civi* user's ID, but *Drupal* user's ID. `$user->uid` was replaced with `CRM_Core_Session::getLoggedInContactID()` method which is broadly used in L&A.
2. Front-end fixed: now the date/time is parsed as UTC and represented as per the computer's timezone.

## Comments

1. Styles were fixed according to http://usejsdoc.org/tags-type.html and https://compucorp.atlassian.net/wiki/display/PCHR/Angular+1+Style+Guide+discussion

2. This solution will work if server's local timezone is UTC. This is an assumption we have to make and it was agreed.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -446,11 +440,7 @@ public function testSaveFromCalculationWillReplaceExistingLeavePeriodEntitlement
   }
 
   public function testSaveFromEntitlementCalculationCanSaveOverriddenValuesGreaterThanProposedEntitlement() {
-    // This mocks the logged in user so we can test
-    // the LeavePeriodEntitlement creation with a comment
-    global $user;
-    $user = new stdClass();
-    $user->uid = 1;
+    $this->registerCurrentLoggedInContactInSession(1);
```
tunbola: Took a look at why we actually need the `registerCurrentLoggedInContactInSession` function in this test, Its actually because of the comment. Also looking at the test title, it does not really need comments because it is testing another thing. I would suggest we remove registering and unregistering contact functions in this test and in the other test and pass comment as null to the `LeavePeriodEntitlement::saveFromCalculation` in both tests.
Then we need to add another test to verify that entitlements can  be saved properly with comments. In that test we will verify that the comment, comment_author_id and comment_date of the Entitlement are the correct values. Having this test will help prevent the bug you just fixed where the comment author was displaying wrongly.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -487,14 +477,12 @@ public function testSaveFromEntitlementCalculationCanSaveOverriddenValuesGreater
...
+    $this->unregisterCurrentLoggedInContactFromSession();
   }
 
   public function testSaveFromEntitlementCalculationCanSaveOverriddenValuesLessThanTheProposedEntitlement() {
-    // This mocks the logged in user so we can test
-    // the LeavePeriodEntitlement creation with a comment
-    global $user;
-    $user = new stdClass();
-    $user->uid = 1;
+    $this->registerCurrentLoggedInContactInSession(1);
```
tunbola: We can remove this just as I suggested above. Then we have a separate test to verify that entitlements are properly saved with comments.

tunbola:
@igorpavlov,  The `$user->uid` is actually the drupal user ID of the contact and not the session ID.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -519,6 +511,44 @@ public function testSaveFromEntitlementCalculationCanSaveOverriddenValuesLessTha
...
+
+    $periodEntitlement = LeavePeriodEntitlement::getPeriodEntitlementForContact(
+      $contact['id'],
+      $period->id,
+      $type->id
+    );
+
+    $this->assertNotNull($periodEntitlement);
+    $this->assertEquals($comment, $periodEntitlement->comment);
+    $this->assertNotNull($periodEntitlement->comment_date);
```
tunbola: Instead of just asserting that the comment_date is not NULL. You can do something like this [Here](https://github.com/civicrm/civihr/blob/d2b646d4d22e90ae978dc658d79929ec7df56b78/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Mail/Template/SicknessRequestNotificationTemplateTest.php#L85-L85). I.e you compare the comment date to the current time and check that the delta is within a certain range(in the link example, the delta is 10 secs). Both time will be in timestamp when comparing.

igorpavlov: Done.

igorpavlov:
Squashed PR Feedbacks.

PR #1993, opened 2017-07-11
PCHR-2239: Update record absence button

igorpavlov:
## Overview

This PR improves the UX of the Record New Absence buttons on the Manager Leave, My Leave and Admin Absence pages. The button on the Admin Dashboard remains the same.

## Before

new

## After

new

## Technical Details

The button dropdown is now a shared component, which means it can be injected anywhere in L&A. To use the component you will need:

```js
// Inject the component in the controller
define([
  'leave-absences/shared/components/leave-request-create-dropdown'
])
```

```html
<!-- Inject the component in the HTML -->
<leave-request-create-dropdown
  contact-id="vm.contactId">
</leave-request-create-dropdown>
```
Attributes:
- `contact-id` - _mandatory_, ID of the currently logged in user
- `selected-contact-id` - _optional_, ID of the pre-selected contact
- `btn-class` - _optional_, class of the master button, ex. "btn-primary"

Note: `uib-dropdown`, `uib-dropdown-toggle` and `uib-dropdown-menu` attributes were injected into component's HTML to make it work in SSP environment.

## Comments

1. Now all component implementations contain icons (confirmed with Mike)
2. There are some minor JavaScript styling improvements
3. Admin Dashboard is now added in the build

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-create-dropdown.html
```diff
@@ -0,0 +1,16 @@
+<div class="btn-group" uib-dropdown>
+  <button type="button" class="btn {{vm.btnClass}} dropdown-toggle" uib-dropdown-toggle>
+    <i class="fa fa-plus-circle" aria-hidden="true"></i> Record New Absence
+  </button>
+  <ul class="dropdown-menu" role="menu" uib-dropdown-menu>
+    <li ng-repeat="option in vm.leaveRequestOptions">
+      <a
+        href="#"
+        leave-request-popup leave-type="{{option.type}}"
```
reneolivo: since every other attribute is in its own line, `leave-type="{{option.type}}"` should be on its own as well

reneolivo: After seeing how the directive leave-request-popup is used elsewhere I think it does make sense to leave `leave-type` next to it.

igorpavlov: Done.

reneolivo:
I think this PR is really well done.

My only complaint would be with passing contactId as a parameter. This parameter is not used by the component itself but a child component. That child component should retrieve contactId from a service since it's the information about the current logged in user.

Anyways. I approve since the issue has more to do with the child component than this one.

igorpavlov:
> My only complaint would be with passing contactId as a parameter.

Correct. The child component (which is the Leave Request directive) requires the contactID, so we must pass it. As per the Leave Request directive itself, I see your point, but also sometimes passing the acting user could be helpful, especially if you need to act on someones behalf. So I personally don't mind passing the ID of the currently logged in user within that way.

PR #1996, opened 2017-07-12
PCHR-2396: Fix an upgrader error during the extension installation

davialexandre:
## Overview
Installing the extension on a new site will result in an error saying that `hrleaveandabsences_leave_request_day_type' is not a valid option for field option_group_id`. The error started after one of the changes introduced by https://github.com/civicrm/civihr/pull/1983, where we now run all the upgraders during the installation. One of them, tries to access the option group mentioned on the error, but it does this at a moment of the process where it's not available yet.

## Before
The installation process cannot be finished because of the mentioned error. Since the option group doesn't exists, the API call used to updated one of its option values fail and throw an Exception.

## After
The installation process completes without any problems. We now catch the thrown exception to avoid the error.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1000.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1000.php
@@ -36,11 +36,20 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1000 {
    * @param string $label
    */
   private function up1000_updateOptionValueLabel($name, $label) {
-    civicrm_api3('OptionValue', 'get', [
-      'sequential' => 1,
-      'option_group_id' => 'hrleaveandabsences_leave_request_day_type',
-      'name' => ['IN' => [$name]],
-      'api.OptionValue.create' => ['id' => '$value.id', 'label' => $label],
-    ]);
+    try {
+      civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
+        'option_group_id' => 'hrleaveandabsences_leave_request_day_type',
+        'name' => ['IN' => [$name]],
+        'api.OptionValue.create' => ['id' => '$value.id', 'label' => $label],
+      ]);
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1000.php
```diff
@@ -36,11 +36,20 @@ private function up1000_updateHalfDayPmLabel() {
...
    */
   private function up1000_updateOptionValueLabel($name, $label) {
-    civicrm_api3('OptionValue', 'get', [
-      'sequential' => 1,
-      'option_group_id' => 'hrleaveandabsences_leave_request_day_type',
-      'name' => ['IN' => [$name]],
-      'api.OptionValue.create' => ['id' => '$value.id', 'label' => $label],
-    ]);
+    try {
+      civicrm_api3('OptionValue', 'get', [
```
mickadoo: @davialexandre can you explain more in the description why this is required? I'm looking at #1998 for @tunbola but because of the order of method calls in the upgrader (first import XML, then run upgraders) I don't see how the upgrader could get here without the option group existing.

davialexandre: @mickadoo the short answer is: I don't know. Checking the order of the method calls was the first thing I did, and I agree that the option groups should be available at this point, but for some reason that I don't know why, they don't. I was able to reproduce this in 2 different sites. I didn't think it was worth to investigate further to found out what was happening.

The real problem here is that running the upgraders during the installation is something "wrong". We only do this to avoid duplicating installation code on the upgraders.

mickadoo: Well I'm not so comfortable with signing off on code that I don't understand. That said I don't want to hold up L&A development for this. It would be nice if there was an explanation, but if it's going to take too much time then I can approve @tunbola's PR.

PR #1998, opened 2017-07-13
PCHR 2310:  Add 'Other- Please add Comment' Sickness Reason Option value

tunbola:
## Overview
This PR adds the _Other- Please add Comment_ Sickness reason option value.

Also an issue whereby an error is thrown during installation on a fresh site where L&A has not been previously installed was fixed. See [PR](https://github.com/civicrm/civihr/pull/1996).

Also previously, the value of `1` was set as default sickness reason when importing sickness requests via the L&A Importer. This PR replaces that value with the `Other` sickness option value just created.

## After
the _Other- Please add Comment_ Sickness reason option value is now part of the sickness reason option group.

The PseudoConstant cache was flushed so that the values can be fetched from the db rather than returning the cached values. The error reported on [PR](https://github.com/civicrm/civihr/pull/1996) was due to the fact that the the call to fetch the L&A option groups was throwing errors because It was trying to look for them in the cache rather than fetching from db.

sickness reasons options - l_and_a 2017-07-13 12-13-55

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
@@ -1,0 +1,41 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1002 {
+
+  /**
+   * Creates the 'other' sickness reason option value.
+   *
+   * @return bool
+   */
+  public function upgrade_1002() {
+    try {
+      $result = civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
+        'option_group_id' => 'hrleaveandabsences_sickness_reason',
+        'name' => 'other',
+      ]);
+
+      if ($result['count'] == 0) {
+        civicrm_api3('OptionValue', 'create', [
+          'option_group_id' => 'hrleaveandabsences_sickness_reason',
+          'name' => 'other',
+          'label' => 'Other - Please leave a comment',
+          'value' => 12,
+          'weight' => 11,
+          'is_reserved' => 1,
+          'is_default' => 0,
+          'is_active' => 1
+        ]);
+      }
+    } catch(Exception $e) {
+      // We run all the upgraders during the extension installation, but, during
+      // this process, the hrleaveandabsences_sickness_reason option group
+      // is still not available and the get API call will fail and throw an exception.
+      // So, to avoid the installation process to stop, we simply catch the exception
+      // and don't do anything. The option group and values will be created just
+      // fine, based on the values set on xml/option_groups/sickness_reason_install.xml
+    }
+
+    return true;
+  }
+}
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Creates the 'other' sickness reason option value.
+   *
+   * @return bool
+   */
+  public function upgrade_1002() {
+    try {
+      $result = civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
```
mickadoo: minor, but you don't really need sequential here if just checking count. Indeed, you could just use the `getcount` method, like

```
$result = civicrm_api3('OptionValue', 'getcount', array(
  'name' => "other",
  'option_group_id' => "hrleaveandabsences_sickness_reason",
));
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
```diff
@@ -0,0 +1,41 @@
...
+          'label' => 'Other - Please leave a comment',
+          'value' => 12,
+          'weight' => 11,
+          'is_reserved' => 1,
+          'is_default' => 0,
+          'is_active' => 1
+        ]);
+      }
+    } catch(Exception $e) {
+      // We run all the upgraders during the extension installation, but, during
```
mickadoo: I'm not sure I agree with this logic. 

It seems like if step 1002 depends on an option group then it should have been installed either in the install hook or in one of the earlier upgraders. 

It's also a "catch all", so if there is some legitimate problem here, like one of the parameters is changed and it breaks the creation call, it will be ignored because of the try-catch.

What advantage do you get from adding it both to the xml and here?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
tunbola: Agree

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1002.php
tunbola: The one added to the xml only gets to run when L&A extension is installed. There are sites already where L&A is installed, so its not possible to uninstall and re-install to add the option value so thats the reason for adding it to the upgrader also. 
For a fresh install, since we run all upgraders during install, for some reason, Please check this [PR](https://github.com/civicrm/civihr/pull/1996), the option group is not available and an error is thrown, but the option values will install just fine because it will install via the xml file.

tunbola: @mickadoo , found the issue, Please check the PR overview and After section.

mickadoo: Interesting to see about the over-zealous caching. Glad you found the source of it :slightly_smiling_face:

PR #1999, opened 2017-07-13
PCHR-2395: Fix admin self leave request popup

igorpavlov:
## Overview

There are some issues with the Leave Request popup:

1. On My Leave page it looks and behaves the same as on Manager Leave or Admin Dashboard page. 
2. The popup flickers during loading.
3. The popup allows admins/managers to choose themselves in the dropdown when recording on behalf.

This PR fixes these issues.

## Before

new

## After

new

## Technical Details

Earlier in order to check if you record for yourself the system was checking if you are Staff. This is logically wrong, since Admin or Manager can also record for themselves via My Leave page, but they are not Staff.

1. A new flag was introduced: `this.isSelfRecord`. If this param is TRUE it indicates that you create/edit a request for yourself. For example:

```html
<a
  leave-request-popup
  contact-id="vm.contactId"
  is-self-record="true">
</a>
```

2. Currently logged in admin is removed from the their managees list.

3. The flickering is fixed by adding an additional condition to HTML.

## Comments

1. The Absence New Record dropdown now can also pass `this.isSelfRecord` through itself.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -1027,7 +1029,12 @@ define([
           // When in Admin Dashboard
           return Contact.all()
             .then(function (contacts) {
-              this.managedContacts = contacts.list;
+              var self = this;
+
+              this.managedContacts = _.remove(contacts.list, function (contact) {
+                // Removes the manager/admin from the list of managees
+                return contact.id !== self.directiveOptions.contactId;
+              });
```
reneolivo: `}.bind(this));` is being used above, it should be used here instead of `var self = this`, etc.

igorpavlov:
> I see spec files have been amended to include the isSelfRecord: true option, but are there any note worthy test cases for isSelfRecord: false?

Any similar places in tests that do not contain `isSelfRecord` test what happens if they are false. And since `isSelfRecord` is a binding, it only can be `true` of `false` (even if not passed at all).

```js
var directiveOptions = {
  contactId: leaveRequest.contact_id, // staff's contact id
  ...
};
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -1205,6 +1205,33 @@
...
+          it('is in create mode', function () {
+            expect($ctrl.isMode('create')).toBeTruthy();
+          });
+
+          it('has admin role', function () {
+            expect($ctrl.isRole('admin')).toBeTruthy();
+          });
+
+          it('loads exactly one contact', function () {
+            expect(_.find($ctrl.managedContacts, { 'id': adminId })).toBeUndefined();
```
reneolivo: if the statement is `it loads exactly one contact` shouldn't the test be something like `expect($ctrl.managedContacts.length).toBe(1)` or maybe the statement needs to change to something like `it does not include admin on contact list` ?

igorpavlov: Done.

PR #2006, opened 2017-07-14
PCHR-2385: hide expiry date field when request never expires

reneolivo:
## Overview

When an absence type is set to never expire on the admin and a new TOIL request is created, the expiry date field should not be displayed. This PR hides the expiry date field only when the particular request is set to never expire.

## Before

anim
This animation shows what happens when requests are set to expire, and what happens after they are set to never expire. Notice the "False" value. This is not the expected behaviour.

## After

anim
On this animation you can see that the Expiry Date field is hidden when requests are set to never expire.

## Technical Details

The absence type model file (`absence-type-model.js`) was modified to add an `canExpire` method. Using this method a boolean value is set when the toil request modal is loading. If the requests are set to expire it shows the expiration field, if not, it hides it (`leave-request-popup-details-tab.html`).

## Comments

This behaviour is the same for staff, mangers, and admins.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/toil-request-ctrl.js
```diff
@@ -52,6 +53,13 @@ define([
           return $q.resolve(vm.request.toil_expiry_date);
         }
 
+        // skips calculation of expiration date if request never expires
```
igorpavlov: I think multiline comment syntax would look better here.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/toil-request-ctrl.js
```diff
@@ -156,6 +164,9 @@ define([
 
         vm._init()
           .then(function () {
+            return initrequestCanExpire();
```
igorpavlov: initRequestCanExpire

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/absence-type-model.js
```diff
@@ -38,6 +40,26 @@ define([
...
+         * @param   {string} absenceTypeId
+         * @return  {Promise}
+         */
+        canExpire: function (absenceTypeId) {
+          return absenceTypeAPI.all({
+            accrual_expiration_unit: { 'IS NULL': 1 },
+            accrual_expiration_duration: { 'IS NULL': 1 },
+            allow_accruals_request: 1,
+            id: absenceTypeId,
+            sequential: 1
```
igorpavlov: I don't think you need to specify `sequential: 1`. Please look at this file `org.civicrm.reqangular/src/common/services/api.js` which contains sendGet function. Better double-check, but I really think it is not needed.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/absence-type-model.js
```diff
@@ -38,6 +40,26 @@ define([
...
+
+        /**
+         * Determines if the absence type can expire by querying if
+         * the expiration unit and duration are null.
+         *
+         * @param   {string} absenceTypeId
+         * @return  {Promise}
+         */
+        canExpire: function (absenceTypeId) {
+          return absenceTypeAPI.all({
```
igorpavlov: Do you need to return default fields, or may be just ID since you are only checking length in the end? Do you need to return all results or just limit to 1?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
```diff
@@ -86,6 +87,7 @@
         spyOn(AbsenceTypeAPI, 'all').and.callThrough();
         spyOn(AbsenceTypeAPI, 'calculateToilExpiryDate').and.callThrough();
         spyOn(TOILRequestInstance, 'init').and.callThrough();
+        spyOn(AbsenceType, 'canExpire').and.returnValue($q.resolve(true));
```
igorpavlov: Can we use callThrough here or is there a specific reason we cannot?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
```diff
@@ -348,6 +350,38 @@
...
+          describe('when request date changes', function () {
+            beforeEach(function () {
+              spyOn(AbsenceType, 'calculateToilExpiryDate');
+              $ctrl.request.to_date = new Date();
+              $ctrl.calculateToilExpiryDate();
+              $rootScope.$digest();
+            });
+
+            it('should not calculate the expiry date field', function () {
+              expect(AbsenceType.calculateToilExpiryDate)
```
igorpavlov: No new line needed here.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/models/absence-type-model.spec.js
```diff
@@ -79,5 +86,43 @@ define([
         });
       });
     });
+
+    describe('canExpire()', function () {
+      describe('passing api parameters', function () {
+        beforeEach(function () {
+          AbsenceType.canExpire(999);
```
igorpavlov: I think it would be better to move this constant in a variable just after describe.

igorpavlov:
Overview says "This PR fixes the following issues:", but there is only one issue described.

igorpavlov:
Please rebase with updated la-milestone-3 and fix conflicts. You may only need to run gulp.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
reneolivo: By default, the mock for `absenceTypeAPI.all` returns a list of records. The way `canExpire` works is to query the API using the absence type id and check if the expiration unit and duration are set to null. Since the mock returns a list of records canExpire will think that the query is true and the record never expires.

I tried to change the query so it returns the record only if the expiration unit and duration are not null, but it seems I can't make that kind of queries to the api. Or is there? I think it would make much more sense to change the query.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
reneolivo: if I don't add a new line the width of the code is going to be more than 80 chars, is that ok?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
reneolivo: Changed the underlying query using `{ 'IS NOT NULL': 1 }`. This allows to use `.and.callThrough()` here and the `canExpire` code makes more sense.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/toil-request-ctrl.spec.js
igorpavlov: Yes, we do not follow strict 80 chars rule. The key is to keep the code readable. In this case 1 line is OK.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/absence-type-model.js
```diff
@@ -38,6 +40,27 @@ define([
          */
         calculateToilExpiryDate: function (absenceTypeID, date, params) {
           return absenceTypeAPI.calculateToilExpiryDate(absenceTypeID, date, params);
+        },
+
+        /**
+         * Determines if the absence type can expire by querying if
+         * the expiration unit and duration are null.
```
igorpavlov: ... are not null.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup/leave-request-popup-details-tab.html
```diff
@@ -84,7 +84,7 @@
 </crm-loading>
 
 <!--Expiry Date Section - Only For Toil-->
-<div class="row" ng-show="$ctrl.isLeaveType('toil')">
+<div class="row" ng-show="$ctrl.isLeaveType('toil') && $ctrl.requestCanExpire">
```
igorpavlov: You have changed the `initRequestCanExpire` function so it returns the opposite value, but the check of `$ctrl.requestCanExpire` in the template still remains the same. Was here a bug in the previous commit?

reneolivo: no bug that I'm aware of. `initRequestCanExpire` has not changed (well, only the name), the one that change was the `canExpire` method, but the query is different. 

The filter is:
```
accrual_expiration_unit: { 'IS NOT NULL': 1 },
accrual_expiration_duration: { 'IS NOT NULL': 1 },
```

and the condition/result is: `return results.length > 0;` meaning: if the expiration unit and duration are not null, then it means that the record can expire.

reneolivo: The previous condition was :
```
accrual_expiration_unit: { 'IS NULL': 1 },
accrual_expiration_duration: { 'IS NULL': 1 },
```

and `return results.length === 0`

PR #2008, opened 2017-07-17
PCHR-2391: Fix error thrown in Sickness email template.

tunbola:
## Overview
When Importing Leave request data of type sickness, Smarty complains of an error on Line 205 of the sickness request email template although the sickness email still gets sent. 

## After
The error is a syntax error resulting from the php condition on that line not being properly formatted.
This PR properly formats the code and error is no more thrown when importing sickness requests.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hremails/email-templates/dist/sickness-request.civicrm.html
+++ b/uk.co.compucorp.civicrm.hremails/email-templates/dist/sickness-request.civicrm.html
@@ -202,7 +202,11 @@
                           {foreach from=$sicknessRequiredDocuments item=value key=id}
                             <div class="request-document" style="Margin-bottom: 10px; margin-bottom: 10px;">
                               <span class="request-document-required" style="Margin-right: 5px; margin-right: 5px; width: 25px;">
-                                <input type="checkbox" disabled {if="" in_array($id,="" $leaverequireddocuments)}checked{="" if}="" style="Margin: 0; margin: 0;">
+                                {if in_array($id, $leaveRequiredDocuments)}
+                                <input type="checkbox" disabled checked style="Margin: 0; margin: 0;">
+                                {else}
+                                <input type="checkbox" disabled style="Margin: 0; margin: 0;">
+                                {/if}
                               </span>
                               <span class="request-document-title" style="color: #4D5663;">{$value}</span>
                             </div>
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/managed_entities/EmailTemplate.mgd.php
```diff
@@ -752,7 +752,11 @@
                           {foreach from=$sicknessRequiredDocuments item=value key=id}
                             <div class="request-document" style="Margin-bottom: 10px; margin-bottom: 10px;">
                               <span class="request-document-required" style="Margin-right: 5px; margin-right: 5px; width: 25px;">
-                                <input type="checkbox" disabled {if="" in_array($id,="" $leaverequireddocuments)}checked{="" if}="" style="Margin: 0; margin: 0;">
+                                {if in_array($id, $leaveRequiredDocuments)}
+                                <input type="checkbox" disabled checked style="Margin: 0; margin: 0;">
+                                {else}
+                                <input type="checkbox" disabled style="Margin: 0; margin: 0;">
+                                {/if}
```
davialexandre: Is this enough for the template to be updated on existing sites?

tunbola: hmmm, I didn't think of that actually. Arrgh, I might need to find a way to add it to the upgrader then or do you have any suggestion in mind? Not sure how upgraders will work with managed entities.

uk.co.compucorp.civicrm.hremails/email-templates/dist/sickness-request.civicrm.html
```diff
@@ -202,7 +202,11 @@ <h2 class="callout-title" style="Margin: 0; Margin-bottom: 0; color: #4D5663; fo
                           {foreach from=$sicknessRequiredDocuments item=value key=id}
                             <div class="request-document" style="Margin-bottom: 10px; margin-bottom: 10px;">
                               <span class="request-document-required" style="Margin-right: 5px; margin-right: 5px; width: 25px;">
-                                <input type="checkbox" disabled {if="" in_array($id,="" $leaverequireddocuments)}checked{="" if}="" style="Margin: 0; margin: 0;">
```
igorpavlov: I see `$leaverequireddocuments` is lowercase for some reason, could it be a reason for the error?

The code provided has a lot of unnecessary repetition. Is there a reason why {if in_array($id, $leaveRequiredDocuments)}checked{/if} would not work?

tunbola: That's not the cause of the error. The main reason is that the php code `in_array($id,="" $leaverequireddocuments)` doesn't come out well after then template is built(see the double quotes `=""` in between). After some discussion with Alessandro and some tests, I discovered that if the logic is separated as it is now, the built template comes out fine.

uk.co.compucorp.civicrm.hrleaveandabsences/managed_entities/EmailTemplate.mgd.php
tunbola: Found out that clearing the cache for any site will update the managed entities and since we do this for client sites after an update, I guess we're good.

PR #2009, opened 2017-07-17
PCHR-2381: Only consider requests overlapping contracts for balance calculation

davialexandre:
## Overview
After deleting a contract, the Leave Requests and Public Holidays within its period are not displayed on the Absence Report anymore, but their balance changes are still counted in the balance calculation, resulting on an invalid value.

## Before
The Absence Report uses the `LeaveRequest.get` and` LeaveRequest.getFull` APIs to load the requests to be displayed. These APIs only return requests that overlap the contracts of a user. The balance information, however, is loaded using the `LeaveRequest.getBalanceChangeByAbsenceType` API, which has a slightly different logic that doesn't consider contracts for the balance calculation. If an admin deletes a contract, a call to `LeaveRequest.get` (or `getFull`) will return no requests for that contract period, but `LeaveRequest.getBalanceChangeByAbsenceType` will still return the number days deducted by leave requests during its period:

## After
The `LeaveRequest.getBalanceChangeByAbsenceType` API was updated so that it works the same way as `LeaveRequest.get` and `LeaveRequest.getFull` APIs and only count for requests overlapping contracts.

## Technical Details
Internally, `LeaveRequest.getBalanceChangeByAbsenceType` uses the `LeaveBalanceChange::getBalanceForEntitlement` and `LeaveBalanceChange::getLeaveRequestBalanceFor` methods. Both were updated to have their queries joining with the contract tables and filtering out requests not overlapping contracts. Due to this change, multiple tests had to be updated as now a contract is required in order to be able to get the balance information.

## Comments
In order to simplify the deletion logic, when a contract is deleted the respective leave requests ARE NOT deleted. We just keep them in the database and make sure that the APIs won't return them. That is why the `.get` and `.getFull` APIs work this way.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -5,6 +5,9 @@ use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate as LeaveRequestDate;
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern as ContactWorkPattern;
+use CRM_Hrjobcontract_BAO_HRJobContract as HRJobContract;
+use CRM_Hrjobcontract_BAO_HRJobContractRevision as HRJobContractRevision;
+use CRM_Hrjobcontract_BAO_HRJobDetails as HRJobDetails;
 
 class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsences_DAO_LeaveBalanceChange {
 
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -76,8 +86,29 @@ public static function getBalanceForEntitlement(LeavePeriodEntitlement $periodEn
       LEFT JOIN {$leaveRequestTable} leave_request 
              ON leave_request_date.leave_request_id = leave_request.id AND
                 leave_request.is_deleted = 0
+      LEFT JOIN {$contractTable} contract
```
tunbola: Can we also use Inner joins for this query just like the one in `getLeaveRequestBalanceForEntitlement` function.

tunbola:
PR description: s/LeaveRequest.getCalanceChangeByAbsenceType/LeaveRequest.getBalanceChangeByAbsenceType
s/LeaveBalanceChanged::getBalanceForEntitlement/LeaveBalanceChange::getBalanceForEntitlement

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
davialexandre: Not in this case. The problem here is that we have to join with balance changes that are either linked to a leave request date or to an entitlement. On `getLeaveRequestBalanceForEntitlement` we only care about those linked to the dates of a leave request, hence a INNER JOIN is possible

PR #2010, opened 2017-07-18
PCHR-2417: SSP BackstopJS tests

deb1990:
## Overview
As part of this PR, backstop js tests for SSP pages has been added

## Pending SSP Tests
1. `Pivot Table` tab of `People`and `Leave and Absence` of report section.
2. Popups which open when clicking on different buttons of Dashboard(These might not be necessary because we now have new version of L&A).

* I am not sure if I should spend more time on the above tests as I have already spent lot of time on this ticket.


AkA84:
@deb1990 would it be possible to also click on some of this result-driven page, so that you can see the style of the results table?

AkA84:
also, don't have a single SSP scenario, make a scenario for each page

uk.co.compucorp.civicrm.hrcore/backstop_data/scenarios/ssp.json
```diff
@@ -0,0 +1,32 @@
+{
+  "scenarios": [
+    {
+      "label": "SSP - HR Resources",
```
AkA84: create a scenario file for each page

PR #2011, opened 2017-07-18
PCHR-2264: Sort leave requests by date

reneolivo:
## Overview
On the My Leave Report section leave requests are not ordered by start date in ascending order. 

This PR arranges leave requests in ascending order using the start date. Expired requests are ordered by the date they expired.

## Before
anim
Notice how dates are sorted, but not in the ascending order. Expiry dates are not sorted at all.

## After
anim

## Technical Details

* The requests have been sorted on the controllers using the API parameter for results order.
* Since expired requests are flattened, lodash was used to sort them after they have been transformed.

## Comments

Ran BackstopJS. It complained about the leave request modal height, but that was due to the fact that the order of the request changed, the old one had a comment and the new one did not. As seen on this screen:

diff1

---

- [x] Tests Pass


igorpavlov:
- PR title: please make the first word *sort* capitalised.
- Overview: I think it is simpler to say "... are not sorted in descending order".
- Before: *Notice how the dates are sorted, but not in descending order.*
- Technical overview: the comment about BackstopJS goes to comments section. Technical overview just tells about the implementation in technical language. You may write something like "Angular orderBy was used to sort dates on front-end"

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-report.html
```diff
@@ -287,7 +287,7 @@
                           </tr>
                         </thead>
                         <tbody>
-                          <tr ng-repeat="entry in report.sections.expired.data track by $index">
+                          <tr ng-repeat="entry in report.sections.expired.data | orderBy:'+expiry_date' track by $index">
```
igorpavlov: Was the sorting by expiration date confirmed?

igorpavlov:
What was the main reason to choose Angular implementations and not:
a) call API with "order by" filter
b) sort inside the controller (eg. using lodash)
?

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/staff-leave-report.html
reneolivo: Sorting expired requests was not part of the original ticket, but it seemed like a similar issue that could come up in the future.

igorpavlov: OK, confirmed by Mike.

igorpavlov:
Apologies, the order is **ascending**, not descending.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.js
```diff
@@ -404,6 +404,7 @@ define([
         })
         .then(function (results) {
           vm.sections.expired.data = results.expiredBalanceChangesFlatten.concat(results.expiredTOILS);
+          vm.sections.expired.data = _.sortBy(vm.sections.expired.data, 'expiry_date');
```
igorpavlov: Is `expiry_date` property a string? If yes, then the sort is performed by comparing strings. Please make sure that `expiry_date` property can be used for sorting by string, otherwise you might want to use moment.js to convert it to timestamp first.

reneolivo: the expiry_date field is always a non empty string representation of a date using the `yyyy-mm-dd` format. So I think it's safe.

reneolivo:
@igorpavlov I have changed the sort method to the controller. It was on the view before for simplicity's sake, but it was not very difficult to reproduce it on the controller either.

PR #2012, opened 2017-07-18
PCHR-2380: Make requests list update after dropdown action

igorpavlov:
## Overview

In Manager Leave the requests table (list) was not updated after any action performed from the dropdown (except *Edit* action which triggers the full request popup).

Moreover, if you have a *status filter* selected (any but *All*) and perform a status transition of a request, then you will end up with a screen where you do not see status filter tabs, which means you cannot deselect the filter and remain stuck until refresh the page.

This PR fixes both issues described above.

## Before

bug

## After

fixed

## Technical Details

1. After the action from the dropdown is confirmed the controller is simply refreshed
2. If the filter selected is not "All" and the loaded requests list is empty, the controller reloads with the filter "All".

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/manage-leave-requests.spec.js
```diff
@@ -989,6 +1020,31 @@
         });
       });
 
+      /**
+       * Spyes on dialog.open() method and resolves it with the given value
```
reneolivo: *Spies typo?

igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.js
```diff
@@ -132,7 +130,10 @@ define([
         classConfirm: 'btn-' + map[action].btnClass,
         msg: map[action].msg,
         onConfirm: function () {
-          return leaveRequest[action]();
+          return leaveRequest[action]()
+            .then(function () {
+              vm.refresh();
```
reneolivo: I'm not 100% about this, but it seems `onConfirm` is waiting for a promise. Does `vm.refresh();` returns another promise? Should you return it as part of the chain? Or maybe you need to return the original result?
```
.then(function (actionResponse) {
  vm.refresh();
  return actionResponse;
});
```

Or maybe this does not really need to return anything at all.

igorpavlov: vm.refresh does not need to return a promise, it does not need to be returned either. `then` returns a promise, so the following chains are valid:

```js
return leaveRequest[action]()
return leaveRequest[action]().then()
return leaveRequest[action]().then().then() // etc
```

As per the result, the onConfirm function does not utilise results of passed promises. Please check this file `org.civicrm.reqangular/src/common/controllers/dialog/dialog-ctrl.js`.

PR #2013, opened 2017-07-19
PCHR-2240: Modify WorkPattern.getCalendar API to accept Start and End date parameters.

tunbola:
## Overview
In a bid to improve the performance on the Calendar page. This PR modifies the WorkPattern.getCalendar API to accept start and end date parameters that defines the range of calendar dates to return rather than accepting the Absence period ID for that purpose.

## Before
The WorkPattern.getCalendar API accepts the Absence period ID to determine the calendar dates to return.

## After
The WorkPattern.getCalendar API accepts start and end date parameters in place of  the Absence period ID to determine the calendar dates to return.

**Sample Request:**
```php
$result = civicrm_api3('WorkPattern', 'getcalendar', array(
  'sequential' => 1,
  'contact_id' => 204,
  'start_date' => "2017-05-01",
  'end_date' => "2017-05-05",
));
```
**Sample Response:** 

```php
{
  "is_error": 0,
  "version": 3,
  "count": 1,
  "id": 0,
  "values": [
    {
      "contact_id": 204,
      "calendar": [
        {
          "date": "2017-05-01",
          "type": "2"
        },
        {
          "date": "2017-05-02",
          "type": "2"
        },
        {
          "date": "2017-05-03",
          "type": "2"
        },
        {
          "date": "2017-05-04",
          "type": "2"
        },
        {
          "date": "2017-05-05",
          "type": "2"
        }
      ]
    }
  ]
}
```
## Comments
This PR also fixes broken tests caused by a change in the WorkPattern::getCalendar BAO function that makes the function to return the option value of the Workday Type rather than the array it was returning before.

- [X] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendarTest.php
```diff
@@ -67,24 +68,18 @@ public function testGetShouldUseTheDefaultWorkPatternIfTheContactHasNoActiveWork
     // middle, assuming all the other will be correct.
     $this->assertCount(15, $calendarDates);
     $this->assertEquals('2015-01-05', $calendarDates[0]['date']);
-    $this->assertEquals($workDayTypes['working_day'], $calendarDates[0]['type']);
+    $this->assertEquals($workDayTypes['working_day']['value'], $calendarDates[0]['type']);
```
davialexandre: The whole point of having this `getWorkDayTypeOptionsArray()` helper was to get a list of option values in the exact same form as the one returned by the calendar, which had a lot of more data than a simple `buildOptions()` would be able to return. With the change in the response format, the `type` property only contains the value, and now, I believe, `buildOptions()` can be used again. So here is what I think we should do:

- Replace all the calls to `getWorkDayTypeOptionsArray()` with `WorkDay::buildOptions('type_id', 'validate')`
- Remove all these `['value']` keys you've added
- Remove the `CRM_HRLeaveAndAbsences_WorkPatternHelpersTrait` completely, as it seems it's only used by the Work Pattern tests

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
```diff
@@ -36,8 +36,19 @@ class CRM_HRLeaveAndAbsences_Service_WorkPatternCalendar {
...
+   * @param int $contactID
+   * @param string $startDate
+   * @param string $endDate
+   * @param \CRM_HRLeaveAndAbsences_Service_JobContract $jobContractService
+   */
+  public function __construct($contactID, $startDate, $endDate, JobContractService $jobContractService) {
     $this->contactID = $contactID;
+    $absencePeriod = new AbsencePeriod();
+    $absencePeriod->start_date = CRM_Utils_Date::processDate($startDate, null, false, 'Y-m-d');
+    $absencePeriod->end_date = CRM_Utils_Date::processDate($endDate, null, false, 'Y-m-d');
```
tunbola: Actually what is needed is just an object that has the start_date and end_date property, even an stdClass should suffice. Not sure if it will be worth modifying the `CRM_HRCore_Date_BasicDatePeriod` class as it also does does things already in the constructor which we don't need, like modifying the end date.
Also another reason I opted for the current approach is because the Absence Period BAO function `adjustDatesToMatchPeriodDates` is used in the class.

davialexandre: > as it also does does things already in the constructor which we don't need, like modifying the end date.
You do you mean by this? Are you talking about the `modify('+1 day')` statement? This is required to "fix" the unusual behavior of PHP's default DatePeriod class. It doesn't actually change the given end date.

I still think we should use the `BasicDatePeriod` class. As I mentioned, having an incomplete entity is a problem. A huge one (I know there are other parts of the code where we use this, but it's something that should be avoided when we have alternatives). 

There's also the problem of the meaning. The calendar now gives you dates for a period, which might be:
- A fraction of an Absence Period
- Exactly the same as an Absence Period
- MORE than an Absence Period (would this would work?)

But the thing is, it's not an Absence Period anymore, but just a generic period of dates.

So here is what I would do:

- Move the `adjustDatesToMatchPeriodDates` method to the `BasicDatePeriod` class. The method is generic enough that this extraction should be very simple (also, a good signal that it should not belong to the AbsencePeriod BAO). Also, I think that instead of returning an array, this method should return a new `BasicDatePeriod` instance
- To get easy access to the `start` and `end` dates, you can add getters for the `BasicDatePeriod` class.
- Change the `WorkCalendar` constructor to accept a `BasicDatePeriod` instance instead of these typeless `startDate` and `endDate` params

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Date/BasicDatePeriodTest.php
```diff
@@ -54,6 +54,82 @@ public function testItDoesNotModifyTheGivenEndDate() {
...
+    $period = new BasicDatePeriod($startDate, $endDate);
+
+    $fromDate = '2016-10-05';
+    $toDate = '2016-10-11';
+
+    $adjustedPeriod = $period->adjustDatesToMatchPeriodDates($fromDate, $toDate);
+    //fromDate is less than the period start date therefore period start date will be returned
+    $this->assertEquals($period->getStartDate(), $adjustedPeriod->getStartDate());
+    $this->assertEquals($period->getEndDate(), $adjustedPeriod->getEndDate());
+
```
davialexandre: Please remove this empty line here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
```diff
@@ -20,9 +20,9 @@ class CRM_HRLeaveAndAbsences_Service_WorkPatternCalendar {
   private $contactID;
 
   /**
-   * @var \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod
+   * @var BasicDatePeriod
```
davialexandre: We can't use aliases like this in annotations :(

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
```diff
@@ -36,19 +36,26 @@ class CRM_HRLeaveAndAbsences_Service_WorkPatternCalendar {
...
 
-  public function __construct($contactID, AbsencePeriod $absencePeriod, JobContractService $jobContractService) {
+  /**
+   * CRM_HRLeaveAndAbsences_Service_WorkPatternCalendar constructor.
+   *
+   * @param int $contactID
+   * @param BasicDatePeriod $datePeriod
+   * @param \CRM_HRLeaveAndAbsences_Service_JobContract $jobContractService
+   */
+  public function __construct($contactID, $datePeriod, JobContractService $jobContractService) {
```
davialexandre: You forgot the type hint for the $datePeriod argument

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/WorkPatternTest.php
```diff
@@ -98,10 +91,16 @@ public function testGetCalendarCanReturnCalendarsForMultipleContacts() {
       'contact_id' => $contact2['id']
     ]);
 
-    $workDayTypes = $this->getWorkDayTypeOptionsArray();
+    $workDayTypes = array_flip(WorkDay::buildOptions('type', 'validate'));
+
+
```
davialexandre: Please remove one of these 2 empty lines here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/WorkPatternCalendar.php
tunbola: :(

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Date/BasicDatePeriod.php
```diff
@@ -47,6 +58,9 @@ public function __construct($start, $end) {
     } else {
       $end = new DateTime($end);
     }
+    $this->startDate = $start;
+    $this->endDate = clone $end;
```
tunbola: Tried this and it was affecting some other function that was expecting DateTime Objects. Since the startDate and endDate properties are private already, then these properties can't be modified.

PR #2014, opened 2017-07-19
PCHR-2309: Add notification service in reqangular

bmnepali:
## Overview
This PR creates a service called `notification` which is a wrapper  for `CRM.alert()`.

## Technical Details
1. Added a `notification` service as a wrapper fro `CRM.alert()`
```js
(function (CRM) {
  define([
    'common/modules/services'
  ], function (services) {
    'use strict';

    services.factory('notification', function () {
      return {

        /**
         * Uses CRM.alert to display alert notification
         *
         * @param {string} title
         * @param {string} body
         * @param {object} options
         */
        alert: function (title, body, options) {
          CRM.alert(body, title, 'alert', options);
        },
        ...
      }
    })
  });
}(CRM));
```

## Comments
- Added test cases for the `notification` service

- [x] Tests Pass


org.civicrm.reqangular/test/services/notification_test.js
```diff
@@ -0,0 +1,56 @@
...
+          unique: true,
+          expires: 0
+        };
+
+        spyOn(CRM, 'alert').and.returnValue({});
+      }));
+
+      describe('alert()', function () {
+        it('calls CRM.alert to display alert notification', function () {
+          notification.alert(title, body, options);
```
deb1990: Call the function in beforeEach, it is not supposed to have the setup for the test. Check other places too.

igorpavlov:
@bmnepali https://github.com/civicrm/civihr/pull/2014/files the dist/reqangular.min.js doesn't look like it was minified successfully. I checked out to your branch and ran gulp - it minified the file. Please make sure you wait for `Finished 'default'` when running gulp.

org.civicrm.reqangular/src/common/services/notification.js
```diff
@@ -0,0 +1,56 @@
...
+      return {
+
+        /**
+         * Uses CRM.alert to display alert notification
+         *
+         * @param {string} title
+         * @param {string} body
+         * @param {object} options
+         */
+        alert: function (title, body, options) {
```
igorpavlov: To me it looks like a lot of repetition in the file. Would it be better to do something like this instead?

```js
(function (CRM) {
  define([
    'common/modules/services'
  ], function (services) {
    'use strict';

    services.factory('notification', function () {
      var notificationFunctions = {};
      var notificationTypes = ['alert', 'success', 'info', 'error'];

      notificationTypes.forEach(function (notificationType) {
        notificationFunctions[notificationType] = function (title, body, options) {
          return CRM.alert(body, title, notificationType, options);
        }
      });

      return notificationFunctions;
    })
  });
}(CRM));
```

With lodash it could be done even more elegant, but I am not sure we want to depend on lodash just for this.

bmnepali:
@igorpavlov Thanks for your comment, i see your point but the PR was already merged as i did not notice your comment. So, the fix to your comment will be in the PR https://github.com/compucorp/civihr-tasks-assignments/pull/235.

igorpavlov:
No problem, thank you.

igorpavlov:
This PR doesn't work for me. Instead of CRM alert it throws native alert for some reason:

image

igorpavlov:
SSP doesn't have this jQuery("#crm-notification-container") object, so a native alert is called

```js
CRM.alert = function (text, title, type, options) {
  type = type || 'alert';
  title = title || '';
  options = options || {};
  if ($('#crm-notification-container').length) {
    var params = {
      text: text,
      title: title,
      type: type
    };
    // By default, don't expire errors and messages containing links
    var extra = {
      expires: (type == 'error' || text.indexOf('<a ') > -1) ? 0 : (text ? 10000 : 5000),
      unique: true
    };
    options = $.extend(extra, options);
    options.expires = options.expires === false ? 0 : parseInt(options.expires, 10);
    if (options.unique && options.unique !== '0') {
      $('#crm-notification-container .ui-notify-message').each(function () {
        if (title === $('h1', this).html() && text === $('.notify-content', this).html()) {
          $('.icon.ui-notify-close', this).click();
        }
      });
    }
    return $('#crm-notification-container').notify('create', params, options);
  }
  else {
    if (title.length) {
      text = title + "\n" + text;
    }
    alert(text);
    return null;
  }
};
```

PR #2016, opened 2017-07-20
PCHR-2423: Disabled Work Pattern Is Still Applied to Leave Request Amount Calculations and Contact Calendar

tunbola:
## Overview
When a Work Pattern is assigned to a Contact and the Work Pattern gets disabled, The disabled Work pattern is still applied to leave request deduction amount calculations and and is also used for the Contact's calendar  for the period when the Contact Work pattern is effective even though the Work pattern has been disabled.

## After
A disabled Work Pattern that is assigned to a Contact is no longer applied to Leave Request deduction amount calculations and not used for the Contact's calendar even though the Contact Work pattern is still effective, rather, the calculation is done based on the default Work pattern and also the calendar is displayed based on the default Work Pattern.

## Technical Details
The `ContactWorkPattern::getForDate` method was modified to return Contact Work patterns that is linked to an active Work Pattern.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
@@ -140,7 +140,7 @@ class CRM_HRLeaveAndAbsences_BAO_ContactWorkPattern extends CRM_HRLeaveAndAbsenc
 
   /**
    * Returns the ContactWorkPattern instance for the given contact and $date
-   * The function will only return the ContactWorkPattern instance that is
+   * This function will only return the ContactWorkPattern instance that is
    * linked to an active WorkPattern.
    *
    * @param int $contactID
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/ContactWorkPattern.php
```diff
@@ -152,7 +152,7 @@ public static function getForDate($contactID, DateTime $date) {
     $contactWorkPatternTable = self::getTableName();
     $workPatternTable = WorkPattern::getTableName();
 
-    $query = "SELECT * FROM {$contactWorkPatternTable} cwp
+    $query = "SELECT cwp.* FROM {$contactWorkPatternTable} cwp
```
davialexandre: Looks like `getForDate` and `getAllForPeriod` are doing almost the same thing. What if we change the former to only delegate the job to the latter? In that case, the "period" would be the date (i.e. the same value for both $start and $end)

PR #2019, opened 2017-07-21
PCHR-2164: Option group values for extensions

deb1990:
## Overview
Previously the supported file types for upload was hardcoded into the code, added support to get it from Option Groups.

## Technical Details
1. `file-mime-types` service has been created, it is used to map file extensions with mime types. `getMimeTypeFor()` method is present in the service for the same reason. **It has been moved to 1.6**

2. The `FileUploader` object has been moved out of `leave-request-instance.js` to `leave-request-popup-files-tab.js`. Similarly the uploading logic has been moved too. 
This is done because all the logic related to fetching Allowed File extensions should be in one place which is `leave-request-popup-files-tab.js`.

3. `allowedMimeTypes` has been deleted from `shared-settings` as it is moved to `file-mime-types` service.

4. `loadSupportedFileExtensions ` added in `leave-request-popup-files-tab` to fetch the allowed extensions and initialise the `file-uploader.`

5. Life cycle method `$onDestroy` added to `leave-request-popup-files-tab` to destroy all listeners.

6. While the `fileUploader` is defined in `leave-request-popup-files-tab`, the main object is defined in `request-ctrl` and passed via attributes. This is done because the `request-ctrl` needs access to the `fileUploader` object to determine whether the leave request has changed or not.

7. `setInitialAttributes()` is defined in `request-ctrl` which is the listener for `Request Updated` event, which broadcasted both from `leave-request-popup-files-tab` and `leave-request-popup-comments-tab`. It is used to update the `initialLeaveRequestAttributes` object which needs to get updated when comments and files are loaded.

8. `triggerChildComponentsSubmitAndWaitForResponse `added to `request-ctrl`, which broadcasts an event to start all child process. It also sends a callback function `doneCallback`, which decides if all the child components are finished and then it resolves the result.

9. Mock data added for `safe_file_extension` Option Group.

10. A loader has been added to the Files tab until the files and uploader is set. For same reason `filesLoaded` variable has been created.

11. Added Mock for `Drupal` in global.js to make the tests run successfully.

## Comments
Right now the `safe_file_extension` OptionGroup does not have the correct OptionValues, a backend fix is necessary for that.

## Additional Bug Fix
Found a bug in the leave popup of admin dashboard, where the uploader in files tab was not visible.
**Before**
dashboard___hr17-1

**After**
dashboard___hr17-1

**Fix**
Added `'leave-absences/shared/components/leave-request-popup-files-tab'` to `leave-absences/admin-dashboard/app.js`

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -85,10 +87,24 @@ define([
     };
 
     (function init () {
-      vm.supportedFileTypes = _.keys(sharedSettings.fileUploader.allowedMimeTypes);
+      loadSupportedFileTypes();
```
igorpavlov: This is now async, should not it be loaded via `$q` so we are sure that `vm.supportedFileTypes` are available when calling `vm.request.loadAttachments();` ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -120,18 +144,12 @@ define([
          * @return {object}
          */
         defaultCustomData: function () {
+          loadSupportedFileTypes.call(this);
```
igorpavlov: Same as above, can we be sure that async behaviour would not break anything?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -72,6 +72,30 @@ define([
...
+       * Load file extensions which are supported for upload
+       *
+       * @return {Promise}
+       */
+      function loadSupportedFileTypes () {
+        return OptionGroup.valuesOf('safe_file_extension')
+          .then(function (extensions) {
+            var allowedMimeTypes = {};
+            extensions.map(function (ext) {
+              allowedMimeTypes[ext.label] = sharedSettings.fileUploader.mimeTypesMap[ext.label];
```
igorpavlov: What happens if mimeTypesMap does not have any extension (.7z for example)? Shouldn't there be a some kind of fallback?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/leave-request-ctrls.spec.js
```diff
@@ -907,7 +883,7 @@
             var leaveRequest = LeaveRequestInstance.init(mockData.findBy('status_id', status));
 
             leaveRequest.contact_id = CRM.vars.leaveAndAbsences.contactId.toString();
-
+            leaveRequest.fileUploader.queue = [];
             initTestController({
```
igorpavlov: Add a line break here please.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/modules/shared-settings.spec.js
```diff
@@ -35,7 +35,7 @@ define([
       });
 
       it('contains the list of allowed mime types', function () {
-        expect(sharedSettings.fileUploader.allowedMimeTypes).toBeDefined();
+        expect(sharedSettings.fileUploader.mimeTypesMap).toBeDefined();
```
igorpavlov: Please compare `sharedSettings.fileUploader.mimeTypesMap` with some object instead of calling `toBeDefined`.

deb1990: If I want to compare it with the exact object I have to manually copy paste the whole object, which I dont think will bring any benefit because in the end it just a static object and I think its enough to keep it in the settings file only.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
deb1990: Yes it was possible in case the API throws an error. Changed the logic a bit, now "this.fileUploader" is set in the `finally`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
deb1990: `vm.request.loadAttachments()` has no relation with `supportedFileTypes`. `supportedFileTypes` is used to show the allowed extensions in the UI only.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
deb1990: That is why I have left a comment in the code saying "If other extensions are added in the OptionGroups, mime types need to be added here", so if we need to add more extensions we need to update this list here. I dont see any work around to this unfortunately.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
igorpavlov: Since the file types are now async, it is now possible to see the UI without file types to be loaded. In this case you will see an unfinished sentence and/or it may induce more issues in the future. I recommend to use a spinner on the whole block (which is inside a dashed border) until all data is loaded. This will guarantee a stable UI.

cc @AkA84

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/modules/shared-settings.spec.js
igorpavlov: Can we please test at least a structure of the object?

org.civicrm.reqangular/src/common/services/file-mime-types.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-env amd */
+
+define([
+  'common/lodash',
+  'common/modules/services'
+], function (_, services) {
```
AkA84: this should be a constant, not a service

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-comments-tab.js
```diff
@@ -141,6 +141,7 @@ define([
     function loadCommentsAndContactNames () {
       return vm.request.loadComments()
         .then(function () {
+          $rootScope.$broadcast('Request Updated');
```
AkA84: namespace the events, i guess something like `LeaveRequestModal::dataChanged` or something like that (the idea being that it's an event that is related to the modal only, and it's triggered when something changed)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -4,25 +4,30 @@ define([
...
   'common/moment',
   'leave-absences/shared/modules/components',
+  'common/services/file-upload',
+  'common/services/file-mime-types',
   'common/services/hr-settings'
 ], function (_, moment, components) {
   components.component('leaveRequestPopupFilesTab', {
     bindings: {
       canManage: '<',
+      fileUploader: '=',
```
AkA84: why not a one-way binding?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -4,25 +4,30 @@ define([
...
-    controller: ['$log', 'HR_settings', 'shared-settings', controller]
+    controller: ['$log', '$rootScope', '$q', 'HR_settings', 'shared-settings', 'OptionGroup', 'FileUpload', 'FileMimeTypes', controller]
   });
 
-  function controller ($log, HRSettings, sharedSettings) {
+  function controller ($log, $rootScope, $q, HRSettings, sharedSettings, OptionGroup, FileUpload, FileMimeTypes) {
     $log.debug('Component: leave-request-popup-files-tab');
 
     var vm = Object.create(this);
+    var events = [];
```
AkA84: first `events`, then `vm`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -70,8 +75,10 @@ define([
       var filesWithSoftDelete = _.filter(vm.request.files, function (file) {
         return file.toBeDeleted;
       });
+      var queueLength = (vm.fileUploader && vm.fileUploader.queue)
```
AkA84: put this two counts into separate functions `filesMarkedForDeletion` and `fileUploaderQueue`, makes for better and more readable code
```js
vm.getFilesCount = function () {
  var filesToDelete = filesMarkedForDeletion();
  var queue = fileUploaderQueue();

  return vm.request.files.length + queue.length - filesToDelete.length; 
}

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js [hunk 11]
```diff
@@ -84,11 +91,76 @@ define([
       return !attachment.attachment_id || vm.canManage;
     };
 
+    /**
+     * Gets called when the component is destroyed
+     */
+    vm.$onDestroy = function () {
```
AkA84:
```js
vm.$onDestroy = unsubscribeFromEvents();

function unsubscribeFromEvents () {
  events.forEach(event);
}

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js [hunk 12]
```diff
@@ -84,11 +91,76 @@ define([
...
+      // destroy all the event
+      events.map(function (event) {
```
AkA84: `map` is used to return an array, you're not doing that. use `forEach`

AkA84: also, these are not events, but listeners. Please rename to `listeners`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +91,76 @@ define([
...
+        loadSupportedFileTypes(),
+        vm.request.loadAttachments()
+      ])
+      .then(function () {
+        $rootScope.$broadcast('Request Updated');
+      })
+      .finally(function () {
+        vm.filesLoaded = true;
+      });
+      events.push($rootScope.$on('uploadFiles: start', uploadAttachments));
```
AkA84: create a `initListeners()` function and put it at the top of the stack

```js
(function init () {
  initListeners();

  $q.all() //..
}())

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js [hunk 14]
```diff
@@ -84,11 +91,76 @@ define([
...
+     * @return {Promise}
+     */
+    function loadSupportedFileTypes () {
+      var allowedMimeTypes;
+
+      return OptionGroup.valuesOf('safe_file_extension')
+        .then(function (extensions) {
+          allowedMimeTypes = {};
+
+          vm.supportedFileTypes = extensions.map(function (ext) {
```
AkA84: you are mixing two things here in the same `map`

1) You are creating the `vm.supportedFileTypes`
2) You are populating the `allowedMimeTypes`

The fact that both are happening the in the `map` that you use to define `vm.supportedFileTypes` is very confusing. Please do the two thing separately

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +91,76 @@ define([
...
+            allowedMimeTypes[ext.label] = FileMimeTypes.getMimeTypeFor(ext.label);
+            return ext.label;
+          });
+        })
+        .finally(function () {
```
AkA84:
```js
.finally(initFileUploader)
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +91,76 @@ define([
...
+      if (vm.fileUploader.queue && vm.fileUploader.queue.length > 0) {
+        vm.fileUploader.uploadAll({ entityID: vm.request.id })
+        .then(function () {
+          $rootScope.$broadcast('uploadFiles: success');
+        })
+        .catch(function () {
+          $rootScope.$broadcast('uploadFiles: error');
+        });
+      } else {
+        $rootScope.$broadcast('uploadFiles: success');
```
AkA84: the logic here is weird, why are we sending a success signal even if nothing happened?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +91,76 @@ define([
...
+    }
+
+    /**
+     * Upload attachment in file uploder's queue, fires an event when done
+     */
+    function uploadAttachments () {
+      if (vm.fileUploader.queue && vm.fileUploader.queue.length > 0) {
+        vm.fileUploader.uploadAll({ entityID: vm.request.id })
+        .then(function () {
+          $rootScope.$broadcast('uploadFiles: success');
```
AkA84: namespace the events

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -708,6 +708,9 @@ define([
       function createRequest () {
         return this.request.create()
           .then(function () {
```
AkA84:
```js
.then(uploadAttachment)

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js [hunk 19]
```diff
@@ -1176,6 +1186,30 @@ define([
...
 
       /**
+       * Fire and event to start Uploading attachment and resolves it according to the result
+       *
+       * @returns {Promise}
+       */
+      function uploadAttachment () {
+        var deferred = $q.defer();
+
+        var successEvent = $rootScope.$on('uploadFiles: success', function () {
```
AkA84: i think this is overcomplicate. You can simply set the listeners in the `init` of the controller, rather than creating and destroying them on the fly

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-comments-tab.js
deb1990: Done `LeaveRequestModal::requestObjectUpdated`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
deb1990: Because we are updating the object in files-tab which is used in request-ctrl. So it should be 2 way.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
deb1990: As discussed, it needs to be present.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
deb1990: Logic has changed, now we have `initiateChildProcess` method. I think its better to keep them here as we have other related variables enclosed, which otherwise needs to be in the global scope.

org.civicrm.reqangular/src/common/services/file-mime-types.js
deb1990: As discussed, kept it service but made the functions promise based.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
```diff
@@ -915,6 +934,33 @@ define([
       }
 
       /**
+       * Fire an event to start child processes which needs to be done after leave request is saved.
+       * Waits for the response before resolving the promise
+       *
+       * @returns {Promise}
+       */
+      function initChildProcess () {
```
AkA84: Ok I was thinking of a way to simplify the process and though of this: what if the parent component, when it triggers the "submit started" event, passes a callback that the child component will then use to signal when it's own process is done and eventually pass an error to it? This way we would have one event only ("submit", versus "submit", "childProcess::success" and "childProcess::error") and I think the logic would be simpler

Here is a POC
```js
function createRequest () {
  return this.request.create()
    .then(triggerChildComponentsSubmitAndWaitForResponse)
    .then(continueSubmit);
}

function triggerChildComponentsSubmitAndWaitForResponse () {
  var deferred = $q.defer();
  var responses = 0;
  var errors = [];

  if (childComponentsCount > 0) {
    $rootScope.$broadcast('LeaveRequestModal::submit', doneCallback);
  } else {
    deferred.resolve();
  }

  function doneCallback (error) {
    error && errors.push(error);

    if (++responses === childComponentsCount) {
      unsubscribeFromEvents();

      errorResponses.length > 0 ? deferred.reject(errors) : deferred.resolve();
    }
  }

  return deferred.promise;
}
```
I think if this works it would make things easier to follow and manage

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +89,121 @@ define([
...
+    function fileUploaderQueue () {
+      return (vm.fileUploader && vm.fileUploader.queue)
+        ? vm.fileUploader.queue : [];
+    }
+
+    /**
+     * Initializes all the listeners
+     */
+    function initListeners () {
+      listeners.push($rootScope.$on('LeaveRequestModal::childProcess::start', uploadAttachments));
```
deb1990: I think the current name is also meaningful. Because we want the childComponent to start processing so its okay to say `LeaveRequestModal::childComponent::start`, and the child processing logically can start at any time but only after submit. So keeping the current name is future proof. Also the components reply with `LeaveRequestModal::childComponent::success/error` which is also inline with the current naming convention.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -84,11 +89,121 @@ define([
...
+      });
+    }
+
+    /**
+     * Upload attachment in file uploder's queue, fires an event when done
+     */
+    function uploadAttachments () {
+      if (vm.fileUploader.queue && vm.fileUploader.queue.length > 0) {
+        vm.fileUploader.uploadAll({ entityID: vm.request.id })
+        .then(function () {
```
deb1990: It seems correct to me, 
```
vm.bluhBluh()
.then()
.catch()
```

org.civicrm.reqangular/src/common/services/file-mime-types.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-env amd */
```
deb1990: As discussed, not making any style guide fixes in this PR.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -4,25 +4,31 @@ define([
   'common/lodash',
```
deb1990: As discussed, not making any style guide fixes in this PR.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-files-tab.js
```diff
@@ -67,11 +73,10 @@ define([
      * @return {Number} of files
      */
     vm.getFilesCount = function () {
```
deb1990: As discussed, not making any style guide fixes in this PR.

org.civicrm.reqangular/test/services/file-mime-types_test.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-env amd, jasmine */
+
+define([
+  'common/angular',
+  'common/angularMocks',
```
deb1990: `angularMocks`is required otherwise `module` not defined error is showing up.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-comments-tab.js
```diff
@@ -16,10 +16,10 @@ define([
       return sharedSettings.sharedPathTpl + 'directives/leave-request-popup/leave-request-popup-comments-tab.html';
```
deb1990: As discussed, not making any style guide fixes in this PR.

PR #2020, opened 2017-07-21
PCHR-2428: Fix Issue with being able to disable the default Work Pattern.

tunbola:
## Overview
The default Work pattern is the work pattern used by contacts without an effective Contact Work pattern, therefore it must be enabled always and should not be disabled.

## Before
The default Work Pattern can be disabled from the Work Pattern listing page. Also a new Work Pattern can be created as default and inactive.

## After
The default Work Pattern cannot be disabled from the Work Pattern listing page and also a new Work Pattern cannot be created as default and inactive.

## Technical Details
A `validateParams` method was created in the Work Pattern BAO patterned after the one in LeaveRequest BAO and some other BAO's in L&A. A new `validateDefaultWorkPattern` method was created to handle the validations for this task.

## Comments
A validation method that ensures that there is at least one active Work Pattern available was removed as it was no longer necessary because with the new changes introduced, the default Work Pattern will always be active.

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
@@ -51,16 +51,14 @@ class CRM_HRLeaveAndAbsences_BAO_WorkPattern extends CRM_HRLeaveAndAbsences_DAO_
     $entityName = 'WorkPattern';
     $hook = empty($params['id']) ? 'create' : 'edit';
 
+    self::validateParams($params);
+
     if(isset($params['is_default']) && $params['is_default']) {
       self::unsetDefaultWorkPatterns();
     }
 
     if($hook == 'create') {
       $params['weight'] = self::getMaxWeight() + 1;
-    }
-
-    if($hook == 'edit' && isset($params['is_active']) && !$params['is_active']) {
-      self::checkIfPatternCanBeDisabled($params['id']);
     }
 
     CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -104,35 +102,60 @@ public static function del($id)
...
+  public static function validateDefaultWorkPattern($params) {
+    $isToBeSetAsDefault = isset($params['is_default']) && $params['is_default'];
+    $isToBeDisabled = isset($params['is_active']) && !$params['is_active'];
+    $isUpdate = !empty($params['id']);
 
-    if($total == 0) {
+    if (!$isUpdate && $isToBeSetAsDefault && $isToBeDisabled) {
       throw new CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException(
-        "You cannot disable a Work Pattern if it's the last one"
+        "You cannot create a new Work Pattern as default and set disabled"
```
davialexandre: Use single quotes here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -104,35 +102,60 @@ public static function del($id)
...
       );
     }
+
+    if($isUpdate && $isToBeDisabled) {
+      $isDefaultWorkPattern = self::isDefault($params['id']);
+
+      if ($isDefaultWorkPattern) {
+        throw new CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException(
+          "You cannot disable the default Work Pattern"
```
davialexandre: Use single quotes here too

PR #2021, opened 2017-07-24
PCHR-2432: Fix date types not loading when editing a leave request

reneolivo:
## Overview
This PR fixes an issue when opening an existing leave request the date types fields stay in a loading state that never completes.

## Before
anim

## After
anim

## Technical Details

The error was introduced after moving code on the `request-ctrl.js`'s `updateAbsencePeriodDatesTypes` method into `loadAbsencePeriodDatesTypes` to separate initializing vs updating values. The specific issue was failing to set the loading `fromDayTypes` and `toDayTypes` variables to false after the initialization phase was completed.

Tests were added to the initialization phase to make sure the loading variables are set to false after requests are completed.

The error handling on `updateAbsencePeriodDatesTypes` was refactored in order to take advantage of the already defined `handleError` function.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request-ctrl.js
@@ -139,7 +139,6 @@ define([
           return $q.resolve();
         }
 
-        self.errors = [];
         self.loading.showBalanceChange = true;
         return LeaveRequest.calculateBalanceChange(getParamsForBalanceChange.call(self))
           .then(function (balanceChange) {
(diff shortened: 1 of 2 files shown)


deb1990:
@reneolivo In the code you have added error handling. But why we were having an error in the first place? Also please add an explanation in the PR about why the error was happening.

reneolivo:
The error handling was already there, I just changed it so it uses the `handleError` function on `request-ctrl.js` as to not repeat code, and because that one handles errors in a better way (checks if the error is already an array, sets loading variables to false, etc.).

Going through the chain of events of `updateAbsencePeriodDatesTypes` the only place where an error can occur is on the method `updateBalance` but that one already handles errors. I could remove the error handling from `updateAbsencePeriodDatesTypes` then.

On the other hand, `loadAbsencePeriodDatesTypes` can fail when resolving `_loadAbsenceTypes` and `_loadCalendar`. These methods should have their own error handling so any other block calling these methods doesn't have to add error handling.

Should I go ahead and move the error handling to `_loadAbsenceTypes` and `_loadCalendar` and remove them from `loadAbsencePeriodDatesTypes` and `updateAbsencePeriodDatesTypes`?

PR #2024, opened 2017-07-25
PCHR-2437: Fix error When importing Leave Request with some Date Formats

tunbola:
## Overview
When Importing Leave Requests and the date format in the CSV is modified to another format apart from `Y-m-d` e.g `d/m/Y` , There are some inconsistencies in the date validation function and an error is reported for some of the Leave requests saying that the dates are Invalid.

## Before
The core PHP DateTime class is used to validate dates in the CSV to be imported but for one reason or the other, a parse error is thrown for some dates e.g `20/09/2016` cannot be parsed while `03/11/2016` can be parsed by DateTime. `Caught exception: DateTime::__construct(): Failed to parse time string (20/09/2016)`

## After
This PR switches to using the Date Format radio button field on the Upload Data page so as to allow the user uploading the CSV to specify the date format in the CSV. If the date format does not match what was chosen for the field, an Invalid Date Format Error is reported.
A core Civi function `CRM_Utils_Date::formatDate` is used to format the date based on the chosen format.
import leave-absence requests - l_and_a 2017-07-25 17-39-32

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
@@ -64,6 +64,12 @@ class CRM_HRLeaveAndAbsences_Import_Parser_Base extends CRM_HRLeaveAndAbsences_I
   private $sicknessReasons;
 
   /**
+   * @var string
+   *  Date Format type chosen on Leave Request Upload Data page
+   */
+  private $dateFormatType;
+
+  /**
    * Class constructor.
    *
    * @param array $mapperKeys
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
```diff
@@ -94,6 +100,8 @@ public function init() {
     $this->setActiveFields($this->_mapperKeys);
     $this->absenceTypes = $this->getAbsenceTypes();
     $this->absenceStatuses = $this->getAbsenceStatuses();
+    $session = CRM_Core_Session::singleton();
+    $this->dateFormatType = $session->get('dateTypes');
```
davialexandre: Just to confirm, will `$session->get('dateTypes')` return a single value, despite the fact that it is using the plural form?

tunbola: Yes, It returns a single value. Don't know why Civi chose to name it in plural.

davialexandre: ¯\_(ツ)_/¯

PR #1952, opened 2017-06-15
PCHR-2008: Add annual entitlements component

igorpavlov:
## Overview

The Annual Entitlements table for Entitlements Tab (as seen here https://github.com/civicrm/civihr/pull/1948) needs now to be populated with real data.

## After

Annual Entitlements table becomes a separate Angular component that fetches the real data and renders it into the view.

new

## Technical Details

1. Angular component for Annual Entitlements was created
2. The component is now implemented in Entitlements Tab
3. FontAwesome is used for icons
4. CRM.alert is utilised to show comments

The component can be used in the following way:

_Don't forget to pass the the Contact ID_
```html
<annual-entitlements contact-id="' + contactId + '"></annual-entitlements>
```

## Comments

- If there are no annual entitlements, instead of the table you see "None found." caption

- ⚠️ If the browser width is ~1200px, the part of the table is out of bounds. This is a problem with a container and not with a component itself, please refer to https://github.com/civicrm/civihr/pull/1959 and https://github.com/civicrm/civihr/pull/1948.

- There is no support for footer in CRM.alert at the moment. At the moment the footer is constructed via rich HTML directly via body text. Appropriate TODO comment has been left in the code.

- If you click "Edit Entitlements" and then "Back" you will be redirected to a correct upper-level tab "Absences" but wrong lower-level tab "Report" instead of "Entitlements". This happens due to the lack of routing support in the Absence tab.
- As per @AkA84 suggestion the following file was wrapped because of global CRM var usage: https://github.com/civicrm/civihr/pull/1952/commits/72af2add0b3fa911c6442ab7082ebf2706f28eb0#diff-01793dce4cd1e0d27c128255e5bc8fce *js/angular/src/leave-absences/absence-tab/components/annual-entitlements.js*
- Functions added to some test files in order to return mocks in objects without promises
- Backstop tests written. *data-mode* attributes were added in *views/absence-tab/components/absence-tab-container.html* file. they are needed in order to allow backstop to open particular tabs and wait for them.
- The TODO comment for 'common.mocks' injection in *js/angular/src/leave-absences/absence-tab/app.js* was left in the code
- The tables' column width as per @deb1990 suggestion cannot be done with pure CSS

---

- [x] Tests Pass

PR #2025, opened 2017-07-26
PCHR-2394: Handle errors in confirmation popup

igorpavlov:
## Overview

When there is an exception (handled by server, but not front-end), the Leave Request Actions dialog stays on a screen forever and shows a spinning animation. Also, the status of the leave request is changed in front-end and once page is refreshed - is reverted back.

This PR fixes these issues.

## Technical Details

The promise chain (from API service to the LeaveRequestActions component) was populated with catching exceptions. New Notification service was injected https://github.com/civicrm/civihr/pull/2014 and used to show notifications if the dialog promise is rejected.

## Comments

Additionally:

1. Annual Entitlements now utilise new Notification service as well
2. Absence Tab components are renamed according with the Style Guide.
3. Annual Entitlements show comment icon as well as the icon the the comment are blue.

screen shot 2017-08-10 at 11 06 09

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/absence-tab/components/annual-entitlements.component.spec.js
```diff
@@ -118,6 +119,17 @@ define([
       });
     });
 
+    describe('when user wants to see a comment to an entitlement', function () {
```
reneolivo: what are we testing here? why is notification.error called when the user wants to see a comment?

igorpavlov: Yeah, good point. Unfortunately I have to use `.error`. The notification service API provides methods: `success`, `error`, `alert` and `info`. Each method shows the connected icon, which has its own colour, for example, `error` shows **red** icon. And that is exactly the icon as per mockups in the original PR (https://github.com/civicrm/civihr/pull/1952). 

So, answering your question: we test _if user wants to see a comment, then a notification with a red icon should spawn_

igorpavlov: The only way to avoid this confusion is to wrap the notification call into a separate function, but in this case the test will lose its purpose, since I want to be sure that a **notification** is shown, not a wrapper to a notification is called.

igorpavlov: Added a comment in tests.

reneolivo: The test is confusing because it implies that an error will be displayed, even if it's not. `notification.error` is being used because it has the same aesthetics as the message that is to be displayed to the user, but we should make a difference even if it's only code-wise.

A `notification.notice` should be added and used instead. This will require refactoring the notifications service a bit.

igorpavlov: As per discussion with @AkA84 we decided just to change the colour of icon which will also change notification.error call to notification.info

reneolivo:
All good.

PR #2026, opened 2017-07-27
PCHR-2240: `api` and `api.optionGroup` performance improvements

AkA84:
## Overview
This PR contains performance improvements to the `api` and `api.optionGroup` services, which were then used in the L&A calendar performance optimization here https://github.com/civicrm/civihr/pull/1989

Also it contains minor improvements to mocked data

### Use join in `api.optionGroup.valuesOf()`
When calling `OptionGroup.valuesOf(nameOrListOfOptionGroupNames)`, the code would proceed to make two separate calls: one to fetch the id(s) of the group(s), the other to fetch the actual values:

```js
// simplified code

function idsOf(groupNames) {
  return this.sendGET('OptionGroup', 'get', {
    name: { 'IN': groupNames }
  });
}

return api.extend({
  valuesOf: function (groupNames, params) {
    return idsOf()
      .then(function (groupIds) {
        return this.sendGET('OptionValue', 'get', {
          option_group_id: { 'IN': groupIds }
        });
      });
  }
});
```
The CiviCRM apiv3 allows us to perform joins on certain entities, and one of those entities are `OptionGroup` and `OptionValue`. With a single joined call, the above had been simplified as follows:

```js
// simplified code

return api.extend({
  valuesOf: function (groupNames, params) {
    return this.sendGET('OptionValue', 'get', {
      'option_group_id.name': { 'IN': multiple ? groupNames : [groupNames] },
    });
  }
});
```
This way if before it took 6 requests to fetch the values of 3 OptionGroups, it now takes simply 3 direct requests.

With the ability to require multiple OptionGroups' values in the same call, you can even fetch all the group's values in just 1 request:

```js
OptionGroup.valuesOf(['group1', 'group2', 'group2']) // 1 request
  .then(function (result) {
    result.group1 // group1 data
    result.group2 // group2 data
    result.group3 // group3 data
  })
```

### Skip pagination call in `api.getAll()` if `pagination` is not passed
The `api.getAll()` method is used to fetch a list of some entities and bundle it with the `total` parameter, that allows to paginate the list

```js
api.getAll('Contacts', { group: 1 }, { page: 2, size: 10 })
  .then(function (result) {
    result.list; // 2nd page, with 10 contacts
    result.total; // 150 contacts in total
  });
```
The value of the `total` property is obtained by making a second call alongside the one that returns the list that we need to use. This second call requires the `id` field of *all* the records

```js
return $q.all([
  (function () {
    // first call to get the records we want
    if (pagination) {
      params.options.offset = (pagination.page - 1) * pagination.size;
      params.options.limit = pagination.size;
    }

    return this.sendGET(entity, action, params);
  }.bind(this))(),
  (function () {
    // second call for pagination

    // `params` does not contain the pagination parameters here
    return this.sendGET(entity, action, params, );
  }.bind(this))()
]).then(function (results) {
  return {
    list: results[0].values,
    total: results[1].count
  };
});
```
The problem was that even when we didn't pass the `pagination` parameter, this second call would happen anyway.

```js
// still making 2 calls even if we already want to get all the contacts
api.getAll('Contacts', { group: 1 });
```
This had now been fixed, and the second call happen only if the `pagination` parameter is passed

```js
return $q.all([
  (function () {
    // first call
  }.bind(this))(),
  (function () {
    if (!pagination) {
      return $q.resolve();
    }

    // second call
  }.bind(this))()
]).then(function (results) {
  var currentList = results[0];
  var allList = pagination ? results[1] : currentList;

  return {
    list: currentList.values,
    total: allList.count
  };
});
```

### Separate mocked `Contact` data from mocked `Contact` api
The mocked `Contact` data was in the same file of the mocked api, making it impossible to require and inject the data only. The two had now been separated

- [x] Tests Pass


igorpavlov:
In the description please change this snippet:

```js
return $q.all([
  (function () {
    // first call
  }.bind(this))(),
  (function () {
    // second call

    if (!pagination) {
      return $q.resolve();
    }
  }.bind(this))()
]).then(function (results) {
  var currentList = results[0];
  var allList = pagination ? results[1] : currentList;

  return {
    list: currentList.values,
    total: allList.count
  };
});
```
to 
```js
return $q.all([
  (function () {
    // first call
  }.bind(this))(),
  (function () {
    if (!pagination) {
      return $q.resolve();
    }

    // second call
  }.bind(this))()
]).then(function (results) {
  var currentList = results[0];
  var allList = pagination ? results[1] : currentList;

  return {
    list: currentList.values,
    total: allList.count
  };
});
```
(swap second call and return $q.resolve)

org.civicrm.reqangular/src/common/services/api/option-group.js
```diff
@@ -1,95 +1,82 @@
...
-                data.values.forEach(function (value) {
-                    idsByName[value.name] = value.id;
-                });
+    /**
+     * Groups OptionValues by OptionGroup name
+     *
+     * @param {array} optionValues
+     * @return {object} an object formed as such:
+     *  {
+     *    group_name_1: [ { value_1 }, { ... }, { value_n } ],
```
igorpavlov: `{ value_1 }` looks like ES6 syntax to me, is it `{ value_1: value_1 }` or [value_1 ,... , value_n] ?

org.civicrm.reqangular/src/common/services/api/option-group.js
```diff
@@ -1,95 +1,82 @@
...
-             */
-            valuesOf: function (groupNames, params) {
-                var multiple = _.isArray(groupNames);
+      /**
+       * Returns the values of the option groups with the given names
+       *
+       * @param {string/array} groupNames
+       *   If the value is an array of names, the method will group
+       *   the values by option group names
+       * @param {object} optional parameters for the query
```
igorpavlov: `@param {object} params - optional parameters for the query`

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
```diff
@@ -1,232 +1,93 @@
...
-                                return _.includes(filters[key].IN, contact[key]);
-                            } else {
-                                return contact[key] === filters[key];
-                            }
-                        });
-                    });
-                }
+        if (pagination) {
+          start = (pagination.page - 1) * pagination.size;
+          end = start + pagination.size;
```
igorpavlov: This line is not needed.

org.civicrm.reqangular/src/common/services/api/option-group.js
```diff
@@ -1,95 +1,82 @@
...
-                            if (multiple) {
-                                return valuesByGroup(data.values, groupIds);
-                            } else {
-                                return data.values;
-                            }
-                        });
-                    }.bind(this));
-            }
+        return this.sendGET('OptionValue', 'get', _.assign({
+          'option_group_id.name': { 'IN': multiple ? groupNames : [groupNames] },
```
mickadoo: You can check yourself, but when using the PHP API this explicit join isn't required, `option_group_id` accepts _either_ integers for the group IDs or strings for the group names, example from API explorer:

```
  "option_group_id.name": {"IN":["account_relationship","activity_contacts"]}
}).done(function(result) {
  // do something
});
```

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
```diff
@@ -1,232 +1,93 @@
...
-            all: function (filters, pagination, value) {
-                var list, start, end;
+        list = value || ContactData.all.values;
 
-                list = value || this.mockedContacts().list;
+        if (filters) {
+          list = list.filter(function (contact) {
+            return Object.keys(filters).every(function (key) {
+              if (key === 'display_name') {
+                return (new RegExp(filters[key], 'i')).test(contact[key]);
```
igorpavlov: Please escape filters[key] so symbols such as `*.^$` etc are treated as normal characters.

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
```diff
@@ -1,232 +1,93 @@
...
-                        }
-                    ]
-                };
-            }
-        }
+        return {
+          total: contacts.length,
+          list: contacts.map(function (mockedContact) {
+            return mockedContact;
+          })
```
igorpavlov: Should this include the list of IDs as well?

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
```diff
@@ -1,232 +1,93 @@
...
-            /**
-             * Adds a spy on every method for testing purposes
-             */
-            spyOnMethods: function () {
-                _.functions(this).forEach(function (method) {
-                    spyOn(this, method).and.callThrough();
-                }.bind(this));
-            },
+      /**
+       * # DRAFT #
```
igorpavlov: Is it still a draft?

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
AkA84: As discussed, not necessary for something used only in tests

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
AkA84: Actually this method needs to go, i added a `@todo` to the method description. I just didn't do it in this PR cause it would add a bit more complications that I didn't need it for this ticket

org.civicrm.reqangular/test/mocks/services/api/contact-mock.js
AkA84: see above

PR #2027, opened 2017-07-28
PCHR-2435: Add session service

reneolivo:
## Overview
This PR adds a way to determine the identity of the current logged in user in the platform.

## Technical Details
* A hook was added to hrcore so every page on the Admin portal provides the session variable.

* A helpers folder was added to tests. When you add a file using the convention `name.helper.js` it will be automatically included before tests run. A `globals.helper.js` file was added which defines the `Drupal` global variable in a known state so it can be used for testing purposes.

* A session service was added on reqangular. The Session service provides the current logged in contact id for now, but could provide more data about the session in the future. 

This service has a `get` method that returns a promise. This is to future proof the method since a session endpoint will be added to the API. 

This is a sample usage of Session in a controller:

```js
define([
  'common/modules/controllers',
  'common/services/session'
], function (controllers) {
  controllers.controller('sampleController', ['Session', function (Session) {
    var loggedInContactId;
    var vm = this;

    vm.$onInit = function () {
      Session.get().then(function (value) {
          loggedInContactId = value.contact_id;
      });
    });
  }]);
});
```

## Comments

This PR minifies `org.civicrm.reqangular/dist/reqangular.min.js` which was accidentally left unminified in PR #2014 

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -156,3 +156,15 @@ function hrcore_civicrm_pre($op, $objectName, $objectId, &$params) {
...
+ * Implements hrcore_civicrm_pageRun. This hook adds session variable to
+ * CRM.vars object.
+ *
+ * @link https://docs.civicrm.org/dev/en/master/hooks/hook_civicrm_pageRun/
+ */
+ function hrcore_civicrm_pageRun($page) {
+    CRM_Core_Resources::singleton()->addVars('session', [
+       'contact_id' => CRM_Core_Session::getLoggedInContactID()
+    ]);
+ }
```
davialexandre: This hook can actually be used to do a lot more than only add variables to the CRM.vars object. Here are some suggestions:
- Remove the "This hook adds session variable to CRM.vars object" sentence from the docblock
- Extract the `CRM_Core_Resources` call to a separate function that will have the single responsibility of adding the vars. I suggest naming it `_hrcore_set_js_session_vars` (The js is to avoid any confusion with the PHP session), but feel free to use any other name that you might find it's better
- Call this new function from hrcore_civicrm_pageRun

igorpavlov:
Please amend the code snippet:
1. Typo "fucntion".
2. Remove `vm.message = '';` and `vm.addMessage = ...` 
3. Change `'chat'` to `'sampleController'`
4. `var contactId;` should be `var loggedInContactId;`

org.civicrm.reqangular/src/common/services/session.js
```diff
@@ -0,0 +1,42 @@
...
+          ? CRM.vars.session.contact_id
+          : Drupal.settings.currentCiviCRMUserId
+      };
+
+      if (!session.contact_id) {
+        throw new Error('Session Error: *Logged In Contact Id* is not defined');
+      }
+
+      /**
+       * This method returns the sessions as a promise. Even though the sessions
```
igorpavlov: Please change the comment to something like this to emphasise the plans on the API.

```js
/**
 * Returns the session data of the currently logged in user (contact)
 *
 * @TODO Even though the session values are taken from the global CRM and Drupal variables, 
 * this should be amended in the future once a Session endpoint in the API is ready.
 * 
 * @return {Promise} resolves with {Object}
 */
```

org.civicrm.reqangular/src/common/services/session.js
```diff
@@ -0,0 +1,42 @@
...
+    'use strict';
+
+    services.service('Session', ['$q', function ($q) {
+      /**
+       * `CRM.vars.session.logged_in_contact_id` is only defined on
+       * the Admin Portal. `Drupal.settings.currentCiviCRMUserId` is defined
+       * everywhere, but it comes from the SSP. Admin Portal should not rely
+       * on definitions from SSP.
+       */
+      var session = {
```
igorpavlov: var session = {};

igorpavlov:
@reneolivo sorry, I don't know why this `var session = {}` has been submitted. Could you revert back, it was actually fine.

PR #2030, opened 2017-07-28
PCHR-2462: Fix background view resetting when recording a leave request

igorpavlov:
## Overview

When you decide to record a new leave request and select a type to record, the background view is refreshed. For example, if you are on a Manager Leave page, go to Calendar tab and then decide to record a leave request, once you open a leave request modal the view will reset to a default Report tab.

This PR fixes this issue.

## Technical Details

The issue happened because of `<a>` tag which invokes the leave request modal directive contained `href="#"`, that refreshed URL with the following structure `https://example.com/page#route` to `https://example.com/page#` which refreshed the view due to the Angular routing.

## Comments

This PR also renames the component from `leave-request-create-dropdown.js` to `leave-request-record-actions.component.js` according to the style guide.

---

- [x] Tests Pass


reneolivo:
Can the component name specify what type of component it is? Leave requests can be recorded in multiple ways (a simple link, multiple buttons, radio selection, etc). In this case the component is a dropdown and it should be added to the name. It could be *record-leave-request-dropdown.component*

igorpavlov:
@reneolivo `In this case the component is a dropdown and it should be added to the name.` - that was exactly the point of the refactoring. The component name should describe a business logic and explain **what it does**, not **how it does it**. However, I agree, we could have other components which also invoke recording leave requests. Why would you have 2 components for the same business logic? This would mean we provide different UIs for same things - this doesn't sound UX friendly.

reneolivo:
I agree with naming components after their business logic, but this component does not record a leave request as soon as you click it, it redirects you to the modal component that will allow you to do so.

Also, UX wise you could have multiple interfaces to do the same thing. In Google Slides I can insert a text box in at least 4 different ways:

* On the menu, by clicking Insert > Textbox
* There's a Textbox button on the toolbar
* I can right click and paste text (which would create a Textbox)
* I can press CTRL + V

PR #2032, opened 2017-07-31
PCHR-2435: Fix session mock

reneolivo:
## Overview
This PR fixes issues with the mock for the session service. These issues are explained in the technical details.

## Technical Details
* The session mock was named session instead of session-mock.
* The session mock was not included in the build.mocks.js file, which makes it available to other modules.
* The Drupal variable is referenced from the global window object. If Drupal is not defined, it won't break modules referencing this file.

## Comments
Merged this branch into a temporary one in la-milestone-3 to make sure that the session-mock works as inteded.

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/services/session.js
```diff
@@ -42,4 +42,4 @@
       };
     }]);
   });
-})(CRM, Drupal);
+})(CRM, window.Drupal);
```
deb1990: In the description it says `The Drupal variable is referenced from the global window object. If Drupal is not defined, it won't break modules referencing this file.` but the code used is `Drupal.settings.currentCiviCRMUserId`. So if Drupal is not defined, we would get error right?

deb1990: The service returns `contact_id `, I think it should be named as `loggedInUserId`, other wise we are going to get same confusion again.

reneolivo: Before you would get an error when including the module, just including it, not even using it.

Now you won't get an error if you include it, but yes, you will get an error if you try to use the module.

An error should be displayed if the logged in contact id is not defined, but this type of error will be confusing, so yes, let me refactor the bit inside.

reneolivo: This was a suggestion from Davi, a session object only has information about the currently logged in user so it would be redundant to add that prefix.

Also, this service will communicate with a session api in the future and the value will probably come as `contact_id`. We could return whatever name convention we want though. Should I leave it as it is or rename it?

deb1990: Okay then I think it is fine. But we should use camel casing for anything related to javascript i.e. `contactID` or `contactId`

org.civicrm.reqangular/src/common/services/session.js
```diff
@@ -1,12 +1,15 @@
...
+/* global CRM */
 
 (function (CRM, Drupal) {
   define([
     'common/modules/services'
   ], function (services) {
     'use strict';
 
+    // Default Drupal value in case it's not defined
+    if (typeof Drupal === 'undefined') Drupal = { settings: {} };
```
deb1990: if block should always be used with curly braces, but in this case we can simply do
`Drupal = Drupal || { settings: {} };`

org.civicrm.reqangular/src/common/services/session.js
deb1990: After this change I think we need change in all places where `session.contact_id` is used. right?

reneolivo: yes, but it's being used only at one place that is still not merged to la-milestone-3 so there's no worry (for this case)

PR #2033, opened 2017-08-01
PCHR-2362: Fix leaving comments in different timezones

igorpavlov:
## Overview

This PR fixes the issue with leaving comments to Leave Requests in different timezones.

### Before

Local Date/time was generated on client side and then was sent to server. On the server the timezone was not saved, so the information about timezone the message was originally submitted from is lost. Moreover, you could easily make up the date/time of your comments.

### After

Now the date/time is not sent from the front-end, instead, it is set on the server as UTC.

Note:

- If front-end does not send an explicit submission date, the server will set the date/time as a comment submission date as per its local timezone
- We make an assumption that server has UTC as a server local timezone

Then dates, once retrieved from the database, parsed as a per the local timezone on client-side. 

Instead of time you see "Just now" caption for freshly created comments since their submission date will be defined only once saved on a server. 

Moreover, now comments are submitted sequentially to preserve their order. A new $q service decorator was used (introduced in the following PR https://github.com/civicrm/civihr/pull/2049).

## Comments

1. Additionally, comments rich text box is improved

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-comments-tab.component.spec.js
```diff
@@ -85,10 +86,10 @@ define([
 
       it('adds comment with proper values', function () {
         expect(controller.request.comments[0]).toEqual({
-          contact_id: staffId,
+          contact_id: contactId,
           created_at: jasmine.any(String),
```
reneolivo: This line should probably be removed or even test that it's not passed at all since this value should be generated on the server.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup/leave-request-popup-comments-tab.html
```diff
@@ -1,19 +1,14 @@
...
     ng-hide="commentsCtrl.getActiveComments().length === 0">
     <div
       class="chr_leave-request-comment"
-      ng-repeat="comment in commentsCtrl.getActiveComments() | orderBy: commentsCtrl.orderComment : true">
+      ng-repeat="comment in commentsCtrl.getActiveComments() | orderBy: commentsCtrl.id : true">
       <div class="chr_leave-request-comment__meta">
         <span class="chr_leave-request-comment__author">{{commentsCtrl.getCommentorName(comment.contact_id)}}</span>
         <span>-</span>
-        <span class="chr_leave-request-comment__date">{{commentsCtrl.formatDateTime(comment.created_at)}}</span>
+        <span class="chr_leave-request-comment__date">{{comment.comment_id ? commentsCtrl.formatDateTime(comment.created_at) : "Just now"}}</span>
```
reneolivo: it's better to add this to the controller so it can be tested. Something like this:

`{{ commentsCtrl.getCommentDate(comment); }}`

```js
vm.getCommentDate = function (comment) {
  comment.comment_id
  ? formatDateTime(comment.created_at)
  : 'Just now';
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
```diff
@@ -77,23 +77,24 @@ define([
...
-            promises.push(LeaveRequestAPI.deleteComment(comment.comment_id));
-          }
-        });
+
+        // Pushing a chain of API calls to create new comments sequentially
+        promises.push($q.sequence(this.comments.filter(function (comment) {
+          return !comment.comment_id;
+        }).map(function (comment) {
+          return function () {
+            return LeaveRequestAPI.saveComment(leaveRequestId, comment);
```
reneolivo: previously the comments array would be updated with the new comment information (which would have the *comment_id* property). This probably was done in case the modal would stay open after saving it and the user would make changes and save again. In that scenario the comments would not be saved twice. It seems now they do, am I right?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-comments-tab.component.spec.js
igorpavlov: I removed the line, but after some short investigation I don't think it is actually needed to test if it is *not* passed. It makes sense to test if something is passed, but I am not quite sure testing opposite is necessary in this case.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/directives/leave-request-popup/leave-request-popup-comments-tab.html
igorpavlov: While I agree with you, I would personally prefer to leave "Just now" in the view.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/models/instances/leave-request-instance.js
igorpavlov: @AkA84 asked me to remove this line `self.comments[index] = commentData;` since the modal is closed immediately after save. But as per discussion we need to make sure the dups are not created if chain breaks. So will amend.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-comments-tab.component.spec.js
igorpavlov: It is used to show the comment immediately in the UI without saving. If you remove contact_id - it will render an empty text span.

reneolivo: I meant *created_at* not *contact_id*, sorry for the confusion

igorpavlov: So, it doesn't affect the order of just created comments so we can successfully remove it!

PR #2035, opened 2017-08-01
PCHR-2170:  Fix issues with TOIL Days Remaining calculation being incorrect

tunbola:
## Overview
Currently on the Manage Entitlement page, we don't display a column that shows the TOIL accrued by the contact for the previous period, This causes some confusion in the remaining days column calculation. 

## Before
For example with the current default sample data in L&A, the contact took 1 leave day for the TOIL absence Type in 2016, If the entitlement for TOIL in 2016 was overridden to be 10, When being redirected to calculate entitlements for 2017 Period.
Remaining Days = Previous Year Entitlement - Days Taken.
Expected Remaining Days: 10 - 1 = 9
What is actually seen is 10 - 1 = 10

manage leave entitlements for period -2017- - 01-01-2017 to 31-12-2017 - l_and_a 2017-08-01 13-23-37

## After
The reason for 10 - 1 = 10 being observed for the Remaining days in the before section is that the contact accrued an additional TOIL of 1 day during the period 2016. So the calculation for the remaining days column is actually correct but the user needs more visibility. A new column Prev Year Accrued TOIL was added to display the information that is already being accounted for in the calculation for Remaining Days column.
So the new formula should be
Remaining Days = (Previous Year Entitlement + Prev. Year Accrued TOIL) - Days Taken.
(10 + 1 ) - 1 = 10

manage leave entitlements for period -2017- - 01-01-2017 to 31-12-2017 - l_and_a 2017-08-01 13-31-05

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
@@ -2274,7 +2274,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChangeTest extends BaseHeadlessTest
     $this->assertEquals(3, $totalBalanceChange);
   }
 
-  public function testGetTotalApprovedToilForPeriod() {
+  public function testGetTotalApprovedToilForPeriodShouldOnlyAccountForApprovedRequests() {
     $contactID = 1;
     $absenceTypeID = 1;
     $leaveRequestStatuses = array_flip(LeaveRequest::buildOptions('status_id', 'validate'));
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -2274,6 +2274,58 @@ public function testGetTotalTOILBalanceChangeForContactWithinAGivenPeriodAndWith
     $this->assertEquals(3, $totalBalanceChange);
   }
 
+  public function testGetTotalApprovedToilForPeriod() {
```
davialexandre: Please always try to give tests more meaningful and descriptive names. I better name here would be `testGetTotalApprovedToilForPeriodShouldOnlyAccountForApprovedRequests` (or something like that)

PR #2036, opened 2017-08-01
PCHR-2299: Allow non active pay scales

mickadoo:
## Overview
When saving a pay scale the "is_active" checkbox is ignored if unchecked. This means that you cannot set a payscale to inactive.

## Before
When saving a payscale with "is active" unchecked it is always saved as active.

## After
The checkbox is not ignored and it is possible to save a checkbox as inactive.

## Technical Details
The cause was that checkbox values are not send from the client in the $_POST request unless they are checked. The solution is to set a default value if not included in the request body.

## Comments
Since this is a small file I did some cleanup, replacing the old array syntax and trying to reduce the long lines to under 80 characters.

---

- [x] Tests Pass

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Form/PayScale.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Form/PayScale.php
@@ -1,30 +1,34 @@
 <?php
 
+use CRM_Hrjobcontract_Page_JobContractTab as JobContractTab;
+use CRM_Hrjobcontract_SelectValues as SelectValues;
+use CRM_Core_Session as Session;
+
 /*
-  +--------------------------------------------------------------------+
-  | CiviHR version 1.4                                                 |
-  +--------------------------------------------------------------------+
-  | Copyright CiviCRM LLC (c) 2004-2014                                |
-  +--------------------------------------------------------------------+
-  | This file is a part of CiviCRM.                                    |
-  |                                                                    |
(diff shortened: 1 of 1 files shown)


hrjobcontract/CRM/Hrjobcontract/Form/PayScale.php
```diff
@@ -133,35 +153,62 @@ static function formRule($fields, $files, $self) {
...
+      'pay_scale',
+      'currency',
+      'amount',
+      'pay_frequency',
+      'is_active'
+    ];
+
+    // set defaults
+    foreach ($properties as $property) {
+      if (!array_key_exists($property, $params)) {
```
tunbola: Should we bother about setting all the other form values to a default if its not present? The formRule already pretty much ensures that all the other fields will be set so this is not necessary. Besides 0 might not be a right default value for some of the fields.
I think what is pretty much needed is just 
`$params['is_active'] = 'CRM_Utils_Array::value('is_active', $params, 0)`;

tunbola:
Nice work cleaning up!

hrjobcontract/CRM/Hrjobcontract/Form/PayScale.php
mickadoo: Good point, I removed the `getParams` method and most of the defaults

PR #2037, opened 2017-08-01
PCHR-2308: Change error message displayed when trying to apply for leave for absence type that doesn't allow negative

tunbola:
## Overview
This PR changes the error message displayed when applying leave for more than the balance available for an Absence type that does not allow to go into negative balance.

## Before
When applying for leave and the absence type does not allow to go into negative balance the error message displayed is _You are not allowed to apply leave in negative_.

screen shot 2017-06-09 at 17 19 33

## After
When applying for leave and the absence type does not allow to go into negative balance the error message displayed is _You cannot make a request for this leave type at this time as this would leave you with a negative balance._

my leave - l_and_a 2017-08-01 17-55-24

- [X] Tests Pass


igorpavlov:
The commit title is too long, please amend according to best practices (<50 chars)

image

PR #2038, opened 2017-08-02
PCHR-2285: Open request details on listed request click

igorpavlov:
⚠️ This branch must wait for sync from https://github.com/civicrm/civihr/pull/2039 and then be merged immediately

## Overview

This PR adds an ability to view/respond/edit a leave request from Manager Leave or Admin Dashboard Leave Requests pages by simply clicking on a request *row* in the table.

## Technical Details

1. The table now has a `.table-hover` class (according to the Bootstrap StyleGuide)
2. Each row has a `.pointer` class
3. Actions (dropdown) component has instructions to stop event propagation on click events (do not fire clicks on parent component, i.e. table row).

### Refactoring

This PR has a refactoring of Leave Request directive - it now utilises session and does **not** need `contact-id` to be passed.

Example 1 - When creating a request
```html
<!-- Do not pass contact-id to the directive anymore -->
<a href
  leave-request-popup
  leave-type="toil"
  selected-contact-id="203"
  is-self-record="true">
  Create new request
</a>
```
Example 2 - When editing a request
```html
<!-- Do not pass contact-id to the directive anymore -->
<a href
  leave-request-popup
  leave-request="request"
  is-self-record="false">
  Edit request
</a>
```
 `contact-id` passing was removed from several view templates.

### Session Model Support

Session is refactored in https://github.com/civicrm/civihr/pull/2039

Session is used in 2 files now (the .spec one uses a mock).

image

Since Session is now a Model, the file paths and dependencies were changed accordingly.

## Comments

1. _Note: any view templates connected with calendars has **not** been yet touched due to the potential clashing with the following PR (https://github.com/civicrm/civihr/pull/1989). `contact-id` passing should be removed later on._
2. Tests are amended to keep the same logic and checks, but without considering `contact-id` to be passed.
3. JS styling improvements were done.

---

- [x] Tests Pass

PR #2039, opened 2017-08-02
PCHR-2285: Turn Session into Model and refactor it

igorpavlov:
⚠️ This branch must be synced with 1.7 only when https://github.com/civicrm/civihr/pull/2038 is ready for new changes

## Overview

Session service now turns into Model since it will represent Session CiviHR Entity soon.

## Technical Details

1. Session moved into _models_ folder and named accordingly
2. Session now has its instance (along with basic tests)
3. Model tests **logic were not touched**, only dependencies injections were amended

## Comments

It was suggested that checked that 1.6 does not yet use Session anywhere.

image

---

- [x] Tests Pass


igorpavlov:
I closed it accidentally.

org.civicrm.reqangular/src/common/models/instances/session-instance.instance.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env amd */
+
+define([
+  'common/modules/models-instances',
+  'common/models/instances/instance'
+], function (instances) {
+  'use strict';
+
+  instances.factory('SessionInstance', ['ModelInstance',
```
AkA84: do we really need an instance in this case? the model simply returns a "static" object with some data in it

org.civicrm.reqangular/src/common/models/session.model.js
```diff
@@ -0,0 +1,50 @@
...
+  define([
+    'common/modules/models',
+    'common/models/model',
+    'common/models/instances/session-instance.instance'
+  ], function (models) {
+    'use strict';
+
+    models.factory('Session', [
+      '$log', '$q', 'Model', 'SessionInstance',
+      function ($log, $q, Model, SessionInstance) {
```
AkA84: you are not using the instance here

org.civicrm.reqangular/src/common/models/instances/session-instance.instance.js
igorpavlov: I made a placeholder. It is useless now, but in the past we had placeholders (ex. Absence-tab)

org.civicrm.reqangular/src/common/models/session.model.js
igorpavlov: Will remove.

igorpavlov:
Last commit includes built files and amends build files, no need to review.

PR #2029, opened 2017-07-28
PCHR-2379: Optimize notification service and fix pre-commit issues

bmnepali:
## Overview
As per @igorpavlov comment on the optimization of the notification service, this PR adds the optimization along with   some  pre-commit warnings.

## Technical Details
1. Optimized notification service for even ore compact version as follows:
```js
/* eslint-env amd */

(function (CRM) {
  define([
    'common/lodash',
    'common/modules/services'
  ], function (_, services) {
    'use strict';

    services.factory('notificationService', function () {
      return _(['alert', 'success', 'info', 'error'])
        .map(function (type) {
          return [type, function (title, body, options) {
            return CRM.alert(body, title, type, options);
          }];
        })
        .zipObject()
        .value();
    });
  });
}(CRM));
```
2. Adopting support for new angular js style guide and naming conversion for test files:
- Made changes to recognize `*.spec.js` as test spec file  replacing `*_test.js` files in `gulpfile.js` and `karma.conf.js` files.
- Renamed `*_test.js` spec files to `.spec.js`
- Fix pre-commit issues for all the spec files

- [x] Tests Pass

PR #2040, opened 2017-08-03
PCHR-2378: Add file.service.js to open file in new tab + test cases

bmnepali:
## Overview
This PR adds a new service which will be responsible to open documents like pdf, images (gif, jpg) will be opened in new tab whereas other documents like .docx, .zip  will be downloaded directly as these can't be viewed in browser. 

Note: Due to IE and Edge browsers handling the blob url as security threat [link](https://stackoverflow.com/questions/20654531/ie-frees-blob-urls), the attached files are downloaded directly.

## Technical Details
1. Create a new service file as `org.civicrm.reqangular/src/common/services/file.service.js` which will contain the following code.
```js
/* globals URL */
/* eslint-env amd */

define([
  'common/modules/services'
], function (services) {
  'use strict';

 services.factory('fileService', ['$http', '$window', '$q', '$log', function ($http, $window, $q, $log) {
    $log.debug('Service: fileService');

    return {

      /**
       * Gets the file from server, creates its blob and opens it in new browser tab.
       * @param  {object} file
       * @return {object}
       */
      openFile: function (file) {
        var deferred = $q.defer();

        $http.get(file.url, {responseType: 'arraybuffer'})
          .success(function (data) {
            var fileBlob = new Blob([data], { type: file.fileType });

            $window.open(URL.createObjectURL(fileBlob), '_blank');
            deferred.resolve(fileBlob);
          });

        return deferred.promise;
      }
    };
  }]);
});
```
The function `openFile` gets the file from server for the given url and creates its **blob** and opens in the new tab.

2. Added test cases for the newly created service in file `org.civicrm.reqangular/test/services/file.service_test.js`. 

After reopened due to issues with IE.
1. Add fix for IE and Edge using `$window.navigator.msSaveOrOpenBlob`.
```js
openFile: function (file) {
        return $http.get(file.url, { responseType: 'arraybuffer' })
          .success(function (data) {
            var fileBlob = new Blob([data], { type: file.fileType });

            // IE Edge and 11+ fix
            if ($window.navigator.msSaveOrOpenBlob) {
              $window.navigator.msSaveOrOpenBlob(fileBlob, file.name);
            } else {
              $window.open(URL.createObjectURL(fileBlob), '_blank');
            }
          });
      }
```
So, the files in IE and Edge will get downloaded instead to be opened in new tab:
2378-later-min

## Comments
The file naming convention for tests is still old. It's because the files have been already renamed in #2029. So, this PR needs to be rebased once #2029 is merged.

- [x] Tests Pass


igorpavlov:
Please improve the PR overview, here is an example:

This PR adds a new service which will be responsible for opening documents like pdf and images in a new tab where these documents can also be downloaded.

(or I didn't quite understand the overview)

bmnepali:
@igorpavlov  Update the PR overview.

org.civicrm.reqangular/src/common/services/file.service.js
```diff
@@ -0,0 +1,34 @@
...
+  'use strict';
+
+  services.factory('fileService', ['$http', '$window', '$q', '$log', function ($http, $window, $q, $log) {
+    $log.debug('Service: fileService');
+
+    return {
+
+      /**
+       * Gets the file from server, creates its blob and opens it in new browser tab.
+       * @param  {object} file
```
igorpavlov: Add a line with * please

org.civicrm.reqangular/src/common/services/file.service.js
```diff
@@ -0,0 +1,34 @@
...
+       * @return {object}
+       */
+      openFile: function (file) {
+        var deferred = $q.defer();
```
igorpavlov: Do you need $q here at all? $http returns promise https://docs.angularjs.org/api/ng/service/$http#get

org.civicrm.reqangular/test/services/file.service_test.js
```diff
@@ -0,0 +1,48 @@
...
+          'fileType': 'image/png'
+        };
+
+        $httpBackend.whenGET().respond(['binaryFile']);
+      }));
+
+      describe('openFile()', function () {
+        beforeEach(function () {
+          spyOn($window, 'open').and.callThrough();
+          blob = new Blob(['binaryFile'], { type: file.fileType });
```
igorpavlov: Line break please

org.civicrm.reqangular/src/common/bundles/services.js
```diff
@@ -6,7 +6,8 @@ define([
   'common/services/file-mime-types',
   'common/services/hr-settings',
   'common/services/notification',
-  'common/services/pub-sub'
+  'common/services/pub-sub',
+  'common/services/file.service'
```
igorpavlov: Wrong order.

org.civicrm.reqangular/src/common/services/file.service.js
bmnepali: done! Thanks for pointing it out.

org.civicrm.reqangular/src/common/services/file.service.js
```diff
@@ -6,28 +6,24 @@ define([
...
     $log.debug('Service: fileService');
 
     return {
 
       /**
        * Gets the file from server, creates its blob and opens it in new browser tab.
-       * @param  {object} file
+       *
+       * @param {object} file
        * @return {object}
```
igorpavlov: It returns Promise now

igorpavlov: image

igorpavlov:
image

org.civicrm.reqangular/src/common/services/file.service.js
bmnepali: Ops! will fix!

bmnepali:
@igorpavlov I guess yes! As I don't have that much change for the `fileService`.

PR #2044, opened 2017-08-04
PCHR-2482: Import Contract Entitlements for L&A

tunbola:
## Overview
Contract Entitlements are imported along with the contract information for a contact via the job Contract Importer. However, the contract entitlements to be imported are still pointing to the absence/leave types of the old Absence extension. Also the contract entitlements for all the leave types are imported as a comma separated value in the `Contract Leave Amount` column of the CSV which is not a very intuitive way to import the entitlements.
## Before
The contract entitlements are pointing to the old absence extension leave types and the contract entitlements for all the leave types are imported as a comma separated values in the `Contract Leave Amount` column.

## After
The contract entitlements are pointing to L&A absence/leave types and the contract entitlements for all the leave types are imported using a column for each of the leave/absence types.
_Sample CSV is attached to the [ticket](https://compucorp.atlassian.net/browse/PCHR-2482)_.

importentitlement

## Technical Details
- The active L&A absence types are fetched and added as part of the importable fields for the Job contract importer. 
- The Importer will only import Leave amounts for the absence Types columns present in CSV and mapped in the Importer Upload page since they are not required fields. For example If Sick and TOIL absence Types are present as columns in the CSV and only Sick column  is mapped, then only the leave amount for the Sick Absence Type is imported. If none of the Absence Type columns are present or If they are present and not mapped, then leave amounts are not imported for any Absence type.
- Also  when importing contractual entitlements the setting to add public holiday to entitlement or not is gotten from the absence type `add_public_holiday_to_entitlement` property and added appropriately.
## Comments
Note that
 [] Tests Pass

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobLeave.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobLeave.php
@@ -9,22 +9,9 @@ class CRM_Hrjobcontract_Import_EntityHandler_HRJobLeave extends CRM_Hrjobcontrac
   }
 
   public function handle(array $params, CRM_Hrjobcontract_DAO_HRJobContractRevision $contractRevision, array &$previousRevision) {
-    $importExportUtility = ImportExportUtility::singleton();
-    $leaveTypes = $importExportUtility->getLeaveTypes();
-    $leaveData = [];
-
-    foreach($leaveTypes as $leaveType) {
-      $key = filter_var($leaveType['title'], FILTER_SANITIZE_STRING);
-      if(!empty($params[$key])){
-        $leaveData[] = [
-          'leave_type' => $leaveType['id'],
-          'leave_amount' => "{$params[$leaveType['title']]}",
-          'add_public_holidays' => $leaveType['add_public_holiday_to_entitlement'],
-          "jobcontract_revision_id" => $contractRevision->id,
-          "jobcontract_id" => $contractRevision->jobcontract_id,
-        ];
-      }
-    }
+    $leaveData = $this->prepareLeaveData(
+      $params, $contractRevision->jobcontract_id, $contractRevision->id
+    );
 
     return civicrm_api3('HRJobLeave', 'replace', [
       'sequential' => 1,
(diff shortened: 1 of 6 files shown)


hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php
```diff
@@ -679,20 +657,30 @@ private function validateField($key, $value)  {
...
+          }
+        }
     }
 
     return array('value'=>$convertedValue, 'error_message'=>$errorMessage);
   }
 
   /**
-   * Checks if given field requires returntype to be 'value', 'label' or 'id' 
+   * Checks if given field requires returntype to be 'value', 'label' or 'id'
```
davialexandre: s/returntype/return type/

hrjobcontract/CRM/Hrjobcontract/SelectValues.php
```diff
@@ -242,21 +242,21 @@ public static function buildHourLocations()
...
-    ));
+      'is_active' => 1
+    ]);
     $result = $result['values'];
-    $options = array();
+    $options = [];
     foreach ($result as $item) {
-      $label = $item['title'];
-      $options[] =  array( 'id'=>$item['id'], 'label'=> $label);
+      $options[] =  ['id'=>$item['id'], 'label'=> $item['title']];
```
davialexandre: Please adjust the formatting here adding spaces between the operators

hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobLeave.php
```diff
@@ -22,33 +33,4 @@ public function handle(array $params, CRM_Hrjobcontract_DAO_HRJobContractRevisio
       'jobcontract_revision_id' => $contractRevision->id,
     ])['values'][0];
   }
-
```
tunbola: No reason really. Just thought the handle function can accommodate the logic without the need for an extra function

hrjobcontract/CRM/Hrjobcontract/Import/EntityHandler/HRJobLeave.php
```diff
@@ -1,19 +1,30 @@
 <?php
 
+use CRM_Hrjobcontract_ExportImportValuesConverter as ImportExportUtility;
```
tunbola: I actually don't like the name `ExportImportValuesConverter`, I believe the class is more of an Utility class but didn't want to change the name and have to do that in many places. Thats why I aliased it to that name because I believe it makes more sense and its more readable.

PR #2046, opened 2017-08-07
PCHR-2483: Hide badge when count is 0

deb1990:
## Overview
As part of this PR, the notification badge loader has been removed and the badge is made to be hidden when the count is 0.

## Before
2483b

## After

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-notification-badge.html
```diff
@@ -1,5 +1,4 @@
-<span class="chr_leave_notification badge badge-primary">
-  <i ng-if="badge.loading.count" class="fa fa-refresh fa-spin"></i>
-  <span ng-if="!badge.loading.count">{{badge.count}}</span>
+<span ng-if="!badge.loading.count && badge.count > 0" class="chr_leave_notification badge badge-primary">
```
AkA84: does it make sense to have `badge.loading` at all now? just show it if the count is > 0

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-notification-badge.html
```diff
@@ -1,5 +1,4 @@
...
+  <span>{{badge.count}}</span>
```
AkA84: also do you need a nested `<span>` now that you don't have the icon anymore?

PR #2048, opened 2017-08-07
PCHR-1830: Hide multiple language options on localization form

mickadoo:
## Overview
CiviHR does not work with multiple language sites. This is because the Drupal views module will try to create views using hardcoded column names. These column names are changed by enabling a multiple language site, for example `civicrm_option_value.label` becomes `civicrm_option_value.label_en_US`

Enabling multiple language and clearing the cache will cause a total site failure because the required views columns will be missing when rebuilding the views. In order to avoid this problem (which happened on the demo site) we will hide the option to enable multiple language.

## Before
You could see "Multiple Languages Support" under Administer > Localization > Languages, Currency, Locations. Enabling it would kill your site.

image

## After
"Multiple Languages Support" is hidden from the form on Administer > Localization > Languages, Currency, Locations.

image

## Technical Details
- I'm not 100% happy with the method of hiding the description in the form, but I didn't want to override the template. I've asked on Mattermost whether it would make sense to allow disabling this section in the template at form level, but have yet to receive a reply.

---

- [] Tests Pass

Tests are broken because of errors in the installer and other extensions. After talking to @davialexandre we decided to ignore these for now and fix them later.

### DIFF
--- a/hrui/hrui.php
+++ b/hrui/hrui.php
@@ -103,6 +103,10 @@ function hrui_civicrm_buildForm($formName, &$form) {
   if ($formName === 'CRM_Admin_Form_Options' && $form->elementExists('value')) {
     $form->removeElement('value');
   }
+
+  if ($form instanceof CRM_Admin_Form_Setting_Localization) {
+    $form->removeElement('makeMultilingual');
+  }
 }
 
 /**
@@ -701,6 +705,16 @@ function hrui_civicrm_alterContent( &$content, $context, $tplName, &$object ) {
      });
    </script>";
   }
+
+  if ($tplName === 'CRM/Admin/Form/Setting/Localization.tpl') {
+    // hide multiple language description
+    $content = str_replace('<h3>Multiple Languages Support</h3>', '', $content);
+    $content .= sprintf(
+      '<script>%s</script>',
+      'CRM.$(".crm-localization-form-block-description").hide();'
+    );
+  }
+
 }
 
 /**


tunbola:
Code looks good with the route taken. Will approve since the broken tests will be fixed in another ticket.

hrui/tests/phpunit/CRM/HRUI/HelperTest.php
```diff
@@ -14,21 +14,13 @@ class CRM_HRUI_HelperTest extends \PHPUnit_Framework_TestCase implements Headles
...
-
   /**
    * @return CiviEnvBuilder
    */
   public function setUpHeadless() {
-    return \Civi\Test::headless()
-      ->install($this->requiredExtensions)
-      ->installMe(__DIR__)
-      ->apply();
+    CRM_HRCase_Upgrader::createRelationshipTypes();
```
tunbola: I don't think this a good approach, i.e calling a function in an upgrader class directly. If the HRUI tests need this to run, perhaps it will be good to have a helper function create the relationship types, the helper function could be in HRUI itself or in HRCase.

hrui/tests/phpunit/bootstrap.php
```diff
@@ -4,6 +4,14 @@
 ini_set('safe_mode', 0);
 eval(cv('php:boot --level=classloader', 'phpcode'));
 
+// Add other extensions to include path to allow use of CRM/* classes
+$civiHRRoot = realpath(__DIR__ . '/../../../');
+$extensionDirs = glob($civiHRRoot . '/*', GLOB_ONLYDIR);
+foreach ($extensionDirs as $extensionDir) {
+  $include_path = $extensionDir . PATH_SEPARATOR . get_include_path( );
```
tunbola: I don't think we need to include all the extension directories If we find a better way to solve the dependency on relationship types.

mickadoo: We will still need to include the hrcore directory so that I can use the fabricator classes. I think it might be nice to include this in all bootstrap files, instead of having to manually modify the `include_path` each time you want to reference a class from another extension in civihr.

hrui/tests/phpunit/CRM/HRUI/HelperTest.php
mickadoo: That makes sense, I just didn't want to duplicate the code, but it's a good point that calling specific upgrader functions is a bit messy. I'll copy the code to the test case.

hrui/tests/phpunit/bootstrap.php
mickadoo: I've tried installing hrcore as you suggested in chat but ran into more errors regarding failed validation on custom CSS URL, errors in raw SQL queries in upgraders and generally a whole lot of stuff that has nothing at all to do with this test. This is pretty frustrating, how an unrelated change in an installer will break this a test in this extension. I'd prefer to avoid trying to fix every problem with the other extensions in this ticket, so unless you have an alternative maybe we can leave it like this.

PR #2049, opened 2017-08-07
PCHR-2362: Add sequential promise execution decorator

igorpavlov:
## Overview

This PR allows to decorate the $q service with a method for sequential task execution, for example, sequential API calls.

## Technical Details

This PR adds a decorator to $q library that allows to execute sequential promises.

Include the decorator and decorate the $q service:
```js
/* eslint-env amd */

define([
  'common/angular',
  'common/decorators/q/q-sequence',
  'common/modules/my-module'
], function (angular, $qSequence) {
  'use strict';

  return angular.module('myModule', [
    'my.module'
  ]).config(['$provide', function ($provide) {
    $provide.decorator('$q', $qSequence);
  }]);
});
```

And finally use the `$q.sequence` method:
```js
$q.sequence([
  function () { return Promise1 },
  function (valueFromPromise1) { return Promise2 },
  function (valueFromPromise2) { return Promise3 },
  // ...
]).then(function (valueFromPromise3) {
  // do something
}).catch(function (error) {
  // handle error
});
```

## Comments
This PR also:
- Fixes a name of a Session model test file
- Improves comments in gulp file
- Removes infinite loop during `gulp` command
- Configures instances to utilise the new $q sequence decorator

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/decorators/q/q-sequence.js
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @param {Array} promises collection of promises
+     * @return {Promise}
+     */
+    function executePromisesInSequence (promises) {
+      var sequence = promises.reduce(function (promiseChain, promise) {
+        return promiseChain.then(promise);
+      }, $delegate.resolve());
+
+      return $delegate.all(sequence);
```
reneolivo: Doesn't `promises.reduce` return a single promise value instead of an array? I don't think $delegate.all is needed here (at least in this scenario)

igorpavlov: https://caolan.github.io/async/docs.html#series

igorpavlov:
@reneolivo 
1. The intention was to write a function similar to https://caolan.github.io/async/docs.html#series and not https://caolan.github.io/async/docs.html#waterfall

reneolivo:
If it's supposed to work like *series* it's fine, but the $deferred.all is not needed.

igorpavlov:
Please move SSP backstop to SSP folder. Also could you please rename backstop to leave-balance-report.js?

PR #2050, opened 2017-08-07
PCHR-2134: Prevent sample data extension from creating duplicate option values and other fixes

omarabuhussein:
## Overview
Sample data extension is creating option values with hardcoded (values) which may allow having two option values belong to the same option group with the same (value). 

## Before
1- Go to Administer > Customize Data.. > Activity Types and ensure that the **Passport** CiviDocument is present
2- In the Database update the value for this activity type to 150
3- Install sample data extension : 
```php
drush cvapi Extension.install deb
ug=1 keys="uk.co.compucorp.civicrm.hrsampledata"
```
4- Go to the tasks and assignments page and try to create a new assignment with the joining timeline
5- Creation of an assignment that uses the Passport CiviDocument will fail since there will be two activity types with ID = 150.

## After
1- Go to Administer > Customize Data.. > Activity Types and ensure that the **Passport** CiviDocument is present
2- In the Database update the value for this activity type to 150
3- Install sample data extension : 
```php
drush cvapi Extension.install deb
ug=1 keys="uk.co.compucorp.civicrm.hrsampledata"
```
4- Go to the tasks and assignments page and try to create a new assignment with the joining timeline
5- Creation of an assignment that uses the Passport CiviDocument will succeed.

## Technical Details
The fix was achieved by : 
1- Removing value column from civicrm_option_value.csv .
2- replacing any used value in other CSV files with the (option value) name instead (this was done for **Relationship With Employee** field only since it was the only field using the value directly).
3- Updating any Importer class that expect a value to expect a name instead by doing name to value conversation inside the importer class (This was also done to **Relationship With Employee** for the same reason above).

## Comments

While working on this, I faced many issues with hrsample data extension and I fixed them all in this PR, these issues are  : 

- Pension Provider was changed in other PR long ago to use contact ID instead of option values but unfortunately it was overlooked and no one noticed the failed tests in this extension due to this change, so I changed the  CSV files by adding a new pension provider contact and replacing the old option values used with this new contact.

- **Main** location type was a Civi core address and it was removed in another PR from civihr installations and there was a single contact in this extension that was using this location type so I changed it to use **Home** location type instead since it doesn't matter what the location type is for a sample contact.

- The ID of the (Cost Center) option value for Job roles should be stored instead of its value so I changed that so the ID is used instead. ( Though this need to be changed later so this field use value as other fields instead of the ID) 
- (Annual Leave) absence type title was used instead of its name (Annual_Leave) which is also fixed.

## Fixed and changed things since Tunbola review on  **Aug 8**

- When installing the extension through UI, the installation breaks since it removes the current logged in user associated contact which cause errors when civicrm tries to log data into civicrm_logs table, this was fixed by keeping any contact record with attached user account (using civicrm UFMatch API). And to reassoisated the default 3 users (civihr_staff, civihr_admin and civihr_manager) I had to remap those contacts Inside UFMatch user to point to the new created contacts with emails (jake@sccs.org, adam@sccs.org and zoe@sccs.org).

- Because the extension takes some time to be installed (due to the number of sample data get created), I increased the timeout on runtime while installing the extension to 5 minutes so users who are going to install the extension via UI don't need to worry about passing the configured php timeout.

- When I created the extension at the first time, I assumed that some data created by some extension are still exists as well as some data created by CiviCRM (like **Home** location type and some other option values). I removed this assumption from the current code so now the extensions creates any data it needs.

- [x] Tests Pass


uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -183,4 +220,26 @@ private function updateCMSUserEmail($cmsID, $newEmail) {
...
+   * @param $newEmail
+   */
+  private function updateMappedContactID($cmsID, $newEmail) {
+    $newEmailData = civicrm_api3('Email', 'get', [
+      'sequential' => 1,
+      'email' => $newEmail,
+      'limit' => '1',
+    ]);
+    if (!empty($newEmailData['id'])) {
+      $newContactID = $newEmailData['id'];
```
tunbola: This would not be the contact ID but the email ID. We need to get the new `contact_id` from `$newEmailData['values']['contact_id']`;

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -32,12 +32,48 @@ public function uninstall() {
...
     $this->cleanCustomFieldsValues();
     $this->cleanSampleDataValues();
   }
 
   /**
+   * Cleaning contacts table and other
+   * related ones
+   */
+  private function cleanContactTables() {
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 0");
```
tunbola: We can use single quotes here since its not storing a variable.

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -32,12 +32,48 @@ public function uninstall() {
...
+    $contactsWithAttachedUsers = $this->getContactsWithAttachedUsers();
+    $contactsWithAttachedUsers = implode(',', $contactsWithAttachedUsers);
+    $contactsDeleteQuery = "DELETE FROM civicrm_contact WHERE id != 1 OR id NOT IN ({$contactsWithAttachedUsers})";
+    CRM_Core_DAO::executeQuery($contactsDeleteQuery);
+    $emailDeleteQuery = "DELETE FROM civicrm_email WHERE contact_id != 1 OR contact_id NOT IN ({$contactsWithAttachedUsers})";
+    CRM_Core_DAO::executeQuery($emailDeleteQuery);
+    CRM_Core_DAO::executeQuery("SET foreign_key_checks = 1");
```
tunbola: single quotes here too.

uk.co.compucorp.civicrm.hrsampledata/sql/uninstall.sql
```diff
@@ -5,13 +5,6 @@ TRUNCATE TABLE civicrm_address;
 TRUNCATE TABLE civicrm_case;
 TRUNCATE TABLE civicrm_case_activity;
 TRUNCATE TABLE civicrm_case_contact;
-
-DELETE FROM civicrm_contact WHERE id !=1;
-
-DELETE FROM civicrm_email WHERE contact_id !=1;
-
-DELETE FROM civicrm_uf_match;
-
```
tunbola: Any reason why we don't need to cleanup the data in these tables anymore?

uk.co.compucorp.civicrm.hrsampledata/tests/phpunit/CRM/HRSampleData/Cleaner/OptionValueTest.php
```diff
@@ -20,67 +20,82 @@ public function setUp() {
...
+    $optionValue = $this->apiGet(
       'OptionValue',
-      ['name' => $optionValue['name'], 'option_group_id' => $this->testOptionGroup['name']]
+      ['name' => $testOptionValue['name'], 'option_group_id' => $this->testOptionGroup['name']]
     );
+    $this->assertEquals($testOptionValue['name'], $optionValue['name']);
+  }
 
-    $this->assertEquals($optionValue['name'], $testOptionValue['name']);
+  public function testProcessWithOptionGroupTitle() {
```
tunbola: The test title should probably be changed to reflect that delete_on_uninstall is on.

uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php
```diff
@@ -14,13 +14,17 @@ public function __construct($extensionName, $extensionDir) {
   }
 
   public function install() {
+    set_time_limit(0);
```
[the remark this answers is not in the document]
omarabuhussein: This extension creates a lot of sample data (via API) so there is no really a way to do make it fast unless we are using direct SQL queries which is not Upgrade-proof solution and could cause many problems later.

uk.co.compucorp.civicrm.hrsampledata/sql/uninstall.sql
omarabuhussein: They are cleaned here https://github.com/civicrm/civihr/blob/7eec0c7e9e9d5a2c52ad8837f19e188b01fedc99/uk.co.compucorp.civicrm.hrsampledata/CRM/HRSampleData/Upgrader.php#L46

 because there are some contacts attached to current logged in users when installing the extension via UI, it had to be done in this way.

PR #2053, opened 2017-08-09
PCHR-2353: Add the ContactHrJobRoles.get API

davialexandre:
## Overview
This PR Adds the `ContactHrJobRoles.get` API. Its main purpose is to allow getting the Job Roles of specific contacts without having to pass through the intermediary HRJobContract API.

A call to this API will return a list of fields from the Job Roles table (not all of them) + a contact_id, which is a field from the Job Contract table.

This API also addresses an issue related to permission. Without it, a user would need access to both HRJobRoles and HRJobContracts APIs, which both might have confidential data. With the new API, we can keep those APIs hidden behind more restrictive permissions and expose only ContactHrJobRoles, which, by the way, doesn't return any type of confidential data.

## Before
To get something like "All the contacts that work in the IT department", an API call like this would be necessary:

```php
civicrm_api3('HrJobRoles', 'get', array(
  'sequential' => 1,
  'department' => "IT",
  'api.HRJobContract.get' => array('id' => "\$value.job_contract_id"),
));
```

Which would give us a response like:

```json
{

    "is_error": 0,
    "version": 3,
    "count": 1,
    "id": 2,
    "values": [
        {
            "id": "2",
            "job_contract_id": "4",
            "title": "fdasfdasfdsafdsa",
            "hours": "0",
            "department": "IT",
            "level_type": "Junior Manager",
            "cost_center": "|",
            "cost_center_val_type": "|",
            "percent_pay_cost_center": "|",
            "amount_pay_cost_center": "|",
            "funder": "|",
            "funder_val_type": "|",
            "percent_pay_funder": "|",
            "amount_pay_funder": "|",
            "location": "Home",
            "start_date": "2016-11-01 00:00:00",
            "api.HRJobContract.get": {
                "is_error": 0,
                "version": 3,
                "count": 1,
                "id": 4,
                "values": [
                    {
                        "id": "4",
                        "contact_id": "9",
                        "is_primary": "1",
                        "deleted": "0",
                        "is_current": "1"
                    }
                ]
            }
        }
    ]
}
```

This presents a few problems:
- At least 2 API calls are necessary (they are chained on the example)
- You need permissions to the two APIs used
- As you can see, the HRJobRoles API has a lot of confidential information, like "funder", "amount funded", etc

## After
You can get the same information as before with a single API call:

```php
civicrm_api3('ContactHrJobRoles', 'get', array(
  'sequential' => 1,
  'department' => "IT"
));
```

Which will return something like:

```json
{

    "is_error": 0,
    "version": 3,
    "count": 1,
    "id": 2,
    "values": [
        {
            "id": "2",
            "title": "fdasfdasfdsafdsa",
            "department": "IT",
            "level_type": "Junior Manager",
            "location": "Home",
            "contact_id": "9"
        }
    ]
}
```

This is how it addresses the aforementioned problems:
- Only one API call is necessary. Internally, the number of queries was also optimized and only one query is required
- Permissions to access the HrJobRoles and HRJobContract APIs aren't necessary anymore. All you need is permission to access ContactHrJobRoles. The permission required to it is: `access AJAX API`
- It includes only a very minimal set of the Job Roles and the Job Contract fields. In fact, the example contains ALL the fields that the API can return. Basically, it only returns the option groups, which are usually useful for filtering + the contact_id

Also, since now we don't need a chained API call, the response format is much simpler and easier to parse.

## Technical Details

`ContactHrJobRoles` is not a real entity. Even though, we do have a `CRM_Hrjobroles_BAO_ContactHrJobRoles`. The reasons is that, internally, the classes that build the db query expect a BAO/DAO class and call some certain methods on them. One of these methods is `fields()`, which was overridden on the BAO to only return a minimum set of `HrJobRoles` fields and add the `contact_id` field, which originally belongs to the `HRJobContract` entity. Since `ContactHrJobRoles` is basically a Job Roles with less fields, the former extends the latter.

The query itself is done by `CRM_Hrjobroles_API_Query_ContactHrJobRolesSelect` class. This wraps a specialized version of the core's `Api3SelectQuery` class and builds a query joining the Job Roles table with the Job Contracts. This specialized version of `Api3SelectQuery` is the `CRM_Hrjobroles_API_Query_VirtualEntitySelectQuery` class. It's specialized in the way that it allows an API to return fields that don't originally belong to the entity being queried (In this case, the contact_id).

Finally, the `hook_civicrm_alterAPIPermissions` hook was used to make sure the required permissions for this API is `access AJAX API`.

---

- [ ] Tests Pass
The `CRM_Hrjobroles_BAO_HrJobRolesTest::testGetCurrentDepartmentsList` and `CRM_Hrjobroles_BAO_HrJobRolesTest::testGetCompactCurrentDepartmentsList` test are both failing and were like that before this PR. They will be fixed later.

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -50,23 +50,23 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
    */
   public function installCostCentreTypes() {
     try{
-      $result = civicrm_api3('OptionGroup', 'create', array(
+      $result = civicrm_api3('OptionGroup', 'create', [
           'sequential' => 1,
           'name' => "cost_centres",
           'title' => "Cost Centres",
           'is_active' => 1
-      ));
+      ]);
 
       $id = $result['id'];
 
       $val = 'Other';
-      civicrm_api3('OptionValue', 'create', array(
+      civicrm_api3('OptionValue', 'create', [
         'sequential' => 1,
         'option_group_id' => $id,
         'label' => $val,
         'value' => $val,
         'name' => $val,
-      ));
+      ]);
 
     } catch(Exception $e){
       // OptionGroup already exists
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -49,17 +49,6 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
    * Creates new Option Group for Cost Centres
    */
   public function installCostCentreTypes() {
-
-    function save($val, $key, $id){
-      civicrm_api3('OptionValue', 'create', array(
-          'sequential' => 1,
-          'option_group_id' => $id,
-          'label' => $val,
-          'value' => $val,
-          'name' => $val,
-      ));
-    }
-
     try{
       $result = civicrm_api3('OptionGroup', 'create', array(
           'sequential' => 1,
@@ -70,11 +59,14 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
 
       $id = $result['id'];
 
-      $options = array(
-          'Other' => 'Other'
-      );
-
-      array_walk($options, 'save', $id);
+      $val = 'Other';
+      civicrm_api3('OptionValue', 'create', array(
+        'sequential' => 1,
+        'option_group_id' => $id,
+        'label' => $val,
+        'value' => $val,
+        'name' => $val,
+      ));
 
     } catch(Exception $e){
       // OptionGroup already exists
--- a/com.civicrm.hrjobroles/api/v3/ContactHrJobRoles.php
+++ b/com.civicrm.hrjobroles/api/v3/ContactHrJobRoles.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
+ * ContactJobRole.get API
+ *
+ * @param array $params
+ * @return array API result descriptor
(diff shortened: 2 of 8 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -70,11 +59,14 @@ function save($val, $key, $id){
 
       $id = $result['id'];
 
-      $options = array(
-          'Other' => 'Other'
-      );
-
-      array_walk($options, 'save', $id);
+      $val = 'Other';
+      civicrm_api3('OptionValue', 'create', array(
```
tunbola: Please use short array syntax here.

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
```diff
@@ -0,0 +1,106 @@
...
+    $contactID = 1;
+    $this->registerCurrentLoggedInContactInSession($contactID);
+
+    CRM_Core_Config::singleton()->userPermissionClass->permissions = [];
+
+    $this->setExpectedApiPermissionException('access AJAX API');
+    civicrm_api3($this->entity, $this->action, ['check_permissions' => true]);
+  }
+
+  public function testTheGetActionReturnsOnlyTheAllowedFields() {
```
tunbola: What about a test that adds the fields to return to the API and some of the passed fields should be the fields that are not supposed to be returned.

com.civicrm.hrjobroles/api/v3/ContactHrJobRoles.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+/**
+ * ContactJobRole.get API
+ *
+ * @param array $params
+ * @return array API result descriptor
```
tunbola: Please add a line in between the param, return and throw doc statements

PR #2057, opened 2017-08-11
PCHR-2377: Fix issue with popup being blocked in chrome

bmnepali:
## Overview
Once the issue with opening attachment in document modal in new tab was fixed, In chrome by default the popup setting being in blocked mode, the attachment cannot be opened in new tab or downloaded directly. This PR fixes this issue and now attachment can be opened or downloaded as expected.

## Before
- Due to the popup status in chrome being in blocked mode, new tab cannot be opened.
screen shot 2017-08-10 at 10 18 00 pm

## After
- Now the attachment can be opened in new tab or downloaded even if the popup mode is in blocked by default.
final-min 1

## Technical Details
By taking reference from this [thread](https://stackoverflow.com/questions/9514698/bypass-popup-blocker-on-window-open-when-jquery-event-preventdefault-is-set) in stackOverflow, when `$window.open()`  is called form inside of the asynchronous call, the chrome blocks the popup in chrome.  

So the `openFile` function in `/common/services/file.service.js` has been updated as follows so that the `$window.open()` does not get called from inside the asynchronous call.

```js
openFile: function(file) {
  // To bypass popup being blocked in chrome
  // Ref: https://stackoverflow.com/questions/9514698 
  var newTab = $window.open('', '_blank');

  return $http.get(file.url, {
      responseType: 'arraybuffer'
    })
    .success(function(data) {
      var fileBlob = new Blob([data], {
        type: file.fileType
      });

      // IE Edge and 11+ fix
      if ($window.navigator.msSaveOrOpenBlob) {
        $window.navigator.msSaveOrOpenBlob(fileBlob, file.name);
      } else {
        newTab.location = URL.createObjectURL(fileBlob);
      }
    });
};
```

- [x] Tests Pass


igorpavlov:
PR description:
- typo `now attachment . can be opened`
- the Before screenshot is low res, should be hight res like the one in the After section
- typo `if the opup mode`
- in the snippet you can use this link instead https://stackoverflow.com/questions/9514698 + fix the indentation please

org.civicrm.reqangular/src/common/services/file.service.js
```diff
@@ -18,6 +18,10 @@ define([
        * @return {promise}
        */
       openFile: function (file) {
+        // To bypass popup being blocked in chrome
+        // Ref: https://stackoverflow.com/questions/9514698/bypass-popup-blocker-on-window-open-when-jquery-event-preventdefault-is-set
```
igorpavlov: Please shorten the link to SO (remove /bypass-...)

igorpavlov:
I wonder what happens if the file is **not** being instantly downloaded.

1) You have to download a heavy file or/and your connection is slow
2) You click on the download button (or perform any other action)
3) You see just a **blank page** until the file starts being downloaded

Ideally, it should open a page with at least a message "Please wait for your file download to start..." or smth.

Please let me know your thoughts. That is *not* a change request, but a call for a discussion.

bmnepali:
@igorpavlov I have implemented your suggestions above. And about your question on downloading large files, I have already added code to display spinner until the files gets downloaded in  file "tasks-assignments/controllers/modal/modal-document.controller.js" T&A as below 
```js
/**
 * Open the given file and display the file in new tab
 * @param {object} file
 */
 function viewFile (file) {
   $rootScope.$broadcast('ct-spinner-show');
   fileService.openFile(file).then(function (blobFile) {
     $rootScope.$broadcast('ct-spinner-hide');
   });
}
```
 So the spinner will be displayed, in case there are large files to be downloaded.

bmnepali:
Closing this PR, as this issue will be fixed in the respective ticket branch `PCHR-2378-open-files-to-preview-in-browser-from-documents-modal`.

PR #2064, opened 2017-08-14
PCHR-2287: Dont hide status filter

deb1990:
## Overview
As part of this PR the following feature has been developed for the Manage Request page.
1. When the left side status filters are clicked, only the table data reloads. Instead of reloading the filters too.

## Before
2287b

## After
2287a

## Technical Details
1. `vm.loading`has been changed to `{ content: true, page: true, table: true }`.
 **page** - This loader is used for the whole page
 **content** - This loader is used for the filter and table section excluding the header and sub header.
 **table** - This loader is used for the table section where the leave requests are shown

2. New parameter `dataType` has been added to `refresh()`, `loadManageesAndLeaves()` and newly created `loadData()` to decide which data(table or filter) to load.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -254,36 +254,43 @@ define([
     }
 
     /**
-     * Loads the managees and calls loadLeaveRequests()
+     * Loads the leave requests for the given data type
      *
+     * @param {String} dataType - which data to load, table or filter or both
```
AkA84: the name is misleading, here you are really talking about sections (table, filters, etc). Data type hints that there are acttualy different types of data related to leave request

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -254,36 +254,43 @@ define([
...
      * @return {Promise}
      */
-    function loadManageesAndLeaves () {
-      vm.loading.content = true;
+    function loadData (dataType) {
+      if (dataType) {
+        return loadLeaveRequests(dataType);
+      } else {
```
AkA84: no need for this `else`

AkA84: you could even do something like this
```js
function loadData (dataType) {
  return $q.all(['table', 'filters'].filter(function (section) {
    return !!dataType ? section === dataType : true;
  }).map(function (section) {
    return loadLeaveRequests(section)
  });
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -254,36 +254,43 @@ define([
...
     /**
-     * Loads the managees and calls loadLeaveRequests()
+     * Loads the leave requests for the given data type
      *
+     * @param {String} dataType - which data to load, table or filter or both
      * @return {Promise}
      */
-    function loadManageesAndLeaves () {
-      vm.loading.content = true;
+    function loadData (dataType) {
```
AkA84: why are you calling this function `loadData` if all it's doing is to load leave requests? Just call it something like `loadLeaveRequestsForSection` or something similar (the name is just a draft)

AkA84: and actually, after taking a look at the code a bit more, why do you need this function at all? It's gonna confused things more. You definitely need to rename this to something like `loadLeaveRequestsXYZ`, but then you already have a `loadLeaveRequests` function that also accepts a "section" param! There is some duplication here and it's definitely not needed.

Since you are calling `loadData` only inside `loadManageesAndLeaves`, why don't you simply put this logic in that method and get rid of this? Granted, it would put back a bit too much "meat" in the method, but i think it's better than how things are now in this pr

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -254,36 +254,43 @@ define([
...
+      }
+    }
 
+    /**
+     * Loads the managees and calls loadLeaveRequests()
+     *
+     * @param {String} dataType - which data to load, table or filter
+     * @return {Promise}
+     */
+    function loadManageesAndLeaves (dataType) {
```
AkA84: Why do you need to bundle them together? Do you *always* need to fetch first the contacts and then the leave requests?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -307,19 +314,24 @@ define([
      */
     function loadLeaveRequests (type) {
```
AkA84: `type` should be something like `section`. We are using the wrong terminology here.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
```diff
@@ -503,7 +516,7 @@ define([
      */
     function refreshWithFilter (status) {
       vm.filters.leaveRequest.leaveStatus = status;
-      vm.refresh();
+      vm.refresh(null, false, 'table');
```
AkA84: give it `1` as the page, that's what you're doing anyway

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
deb1990: changed to `section`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
deb1990: I have put the logic on the `loadManageesAndLeaves`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
deb1990: I have put the logic on the `loadManageesAndLeaves`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/manage-leave-requests.component.js
deb1990: We do not always need to fetch contacts. But it was done to simplify the refresh logic, right now we have a single refresh function which always gets called whenever we want to do any kind of refresh. Normally its not a good idea because we will be calling the api unnecessarily, but in our case as the API is cached so we never call the same api twice for same filters. But it helps to simplify the code a lot.

PR #2066, opened 2017-08-16
PCHR-2514: Implement AngularJS Style Guide and JS Semi-Standard across Job Roles app

AkA84:
# Overview
This PR fixes several regression/unfixed issues in Job Roles, and also makes the angular app and its file to follow both the AngularJS Style Guide and the JS Semi-Standard Style

### New role panel's style was not correct
For some reason when i was working on #1725 I did not fully apply the CSS classes on the `add_new_role.html` partial like I did for the `job_role_panel.html`

#### Before
new-role-tab-before

#### After
new-role-tab-after

### Switching type of Funder/Cost Centre would not switch the amount field
This was caused by the merge of the `amnesty-styling` branch in the Shoreditch theme (see https://github.com/compucorp/org.civicrm.shoreditch/pull/24).

In particular the issue was caused by [js/enable-select2.js](https://github.com/civicrm/org.civicrm.shoreditch/blob/master/js/enable-select2.js), which would try to activate the select2 plugin on any normal `<select>` tag it finds.

Applying the `.no-select2` class to any `[editable-select]` element solved the issue (for more information about the class, check the Shoreditch PR linked above)

#### Before
type-switch-before

#### After
type-switch-after

### Cost Centres would not get saved
The cause is the same as before, fix is the same

#### Before
cost-centre-save-before

#### After
cost-centre-save-after

### Adding multiple rows of Funder/Cost Centre would cause an Angular error in the "add role" panel
This was caused by the missing `.ui-select-match` and `.ui-select-choices` classes on the `<ui-select>` elements contained in each row. Those classes are necessary for the `ui-select` directive to work correctly

#### Before
new-row-error-before

#### After
new-row-error-after

### The page would throw a JS error on load
sort-name-error

This was because the single job roles templates were being injected in the DOM before the data was fully loaded
```html
<crm-loading show="!jobroles.loading">
  <ul class="list-unstyled">
    <li ng-repeat="job_roles_data in jobroles.presentJobRoles"
      ng-include src="jobroles.jobRolePanelUrl">
    </li>
  </ul>
</crm-loading>
```
(`cm-loading` internally uses `ng-show`, not `ng-if`)

Making sure that the templates were added to the DOM only after the data was loaded fixed the problem
```html
<crm-loading show="!jobroles.loading">
  <ul class="list-unstyled" ng-if="!jobroles.loading">
    <li ng-repeat="job_roles_data in jobroles.presentJobRoles"
      ng-include src="jobroles.jobRolePanelUrl">
    </li>
  </ul>
</crm-loading>
```

### Folders structure and file names
The folders structure and the file names have been changed

#### Before
files-before-2

#### After
files-after

Notice that `app.js` is now inside the `modules/` folder and it's called `job-roles.module.js`

### Sub-modules
The content `app.js` (now `job-roles.module.js`) file containing the main `hrjobroles` module had been split into several sub-modules to enhance the separation of concern

#### Before
```js
angular.module('hrjobroles', [
  'ngAnimate',
  'ngSanitize',
  'ngResource',
  'common.angularDate',
  'common.services',
  'common.directives',
  'compuNgRoute',
  'xeditable',
  'angular.filter',
  'ui.bootstrap',
  'ui.select',
  'hrjobroles.controllers',
  'hrjobroles.directives',
  'hrjobroles.filters',
  'hrjobroles.services'
])
.constant('settings', {
  // ...
})
.config(['settings','$routeProvider','$resourceProvider','$httpProvider','$logProvider',
  function(settings, $routeProvider, $resourceProvider, $httpProvider, $logProvider){
  // ...
  }])
.run(['settings','$rootScope','$q', '$log', 'editableOptions',
  function(settings, $rootScope, $q, $log, editableOptions){
    // ...
  }
]);
```

#### After
```js
angular.module('hrjobroles', [
  'hrjobroles.core',
  'hrjobroles.config',
  'hrjobroles.run',
  'hrjobroles.constants',
  'hrjobroles.controllers',
  'hrjobroles.filters',
  'hrjobroles.services'
]);
```

#### Before
The way that an actual component (a service, a directive, an angular component, a controller, etc) was listed as a dependency in the app was the following way:

1. List its AMD module in the main module file
    ```js
    // app.js
    define([
      'job-roles/controllers/hr-job-roles-controller',
    ], function () {
    }
    ```
2. Have a file where the component's module is created
    ```js
    // controllers.js
    define([
      'common/angular'
    ], function (angular) {
      'use strict';

      return angular.module('hrjobroles.controllers', []);
    });
    ```
3. Inject the module AMD module in the component's file, and register the component on it
    ```js
    // hr-job-roles-controller.js
    define([
      'job-roles/controllers/controllers',
    ], function (controllers) {
      'use strict';

      controllers.controller('HRJobRolesController', [/* .. */, function (/* .. */) {
        // ...
      }]);
    });
    ```

This had several drawbacks:
1. The main module file would contain a very long list of all the individual components that were required by the app
2. The module file didn't have much going on
3. The component file had this inject-module-and-register-on-it thing going on that at times was difficult to wrap your head around

#### After
The above had been changed in the following way

1. In the component file, simply inject the actual modules that are required in the component implementation. Then define the `__name` (more on it later) of the component and return it
    ```js
    // job-roles.controller.js
    define([
      'common/lodash',
      'common/moment'
    ], function (angular, _, moment) {
      'use strict';

      JobRolesController.__name = 'JobRolesController';
      JobRolesController.$inject = [/* ... */];

      function JobRolesController (/* ... */) {
      }

      return JobRolesController;
    });
    ```
2. In the component's module file, inject the component and register it using its `__name` property
    ```js
    // job-roles.controllers.js
    define([
      'job-roles/controllers/job-roles.controller',
    ], function (JobRolesController) {
      'use strict';

      return angular.module('hrjobroles.controllers', [])
        .controller(JobRolesController.__name, JobRolesController)
    });
    ```
3. In the main module file, simply require the component's module
    ```js
    define([
      'job-roles/modules/job-roles.controllers',
    ], function (angular) {

    });
    ```

This has several advantages
1. The component's file doesn't need to inject the module anymore. All you have in the file is the component's definition, which is then returned
2. The component's module file now contains a manifest of all the components that are registered on it, and by using the `__name` property the component is still the "owner" of its name (meaning, in order to know with what name a component is registered, you can simply look it up in its own file, you don't need to go to the module file to find it)
3. The main module file at this point simply requires the sub-modules and nothing else, meaning that there is no more confusing list of all the components in one place, but instead each component is neatly listed in its own sub-module

This is the main module before and after:
```js
// before
define([
  'common/ui-select',
  'common/services/angular-date/date-format',
  'common/services/dom-event-trigger',
  'common/modules/routers/compu-ng-route',
  'common/modules/directives',
  'common/directives/angular-date/date-input',
  'job-roles/vendor/angular-editable',
  'job-roles/vendor/angular-filter',
  'job-roles/controllers/controllers',
  'job-roles/directives/directives',
  'job-roles/filters/filters',
  'job-roles/services/services',
  'job-roles/controllers/hr-job-roles-controller',
  'job-roles/controllers/modal-dialog',
  'job-roles/services/hr-job-roles-service',
  'job-roles/services/date-validation',
  'job-roles/services/hr-job-roles-service-filter',
  'job-roles/directives/example',
], function () {
  'use strict';

  angular.module('hrjobroles', [
    'ngAnimate',
    'ngSanitize',
    'ngResource',
    'common.angularDate',
    'common.services',
    'common.directives',
    'compuNgRoute',
    'xeditable',
    'angular.filter',
    'ui.bootstrap',
    'ui.select',
    'hrjobroles.controllers',
    'hrjobroles.directives',
    'hrjobroles.filters',
    'hrjobroles.services'
  ])
});
```
```js
// after
define([
  'common/angular',
  'job-roles/modules/job-roles.config',
  'job-roles/modules/job-roles.constants',
  'job-roles/modules/job-roles.controllers',
  'job-roles/modules/job-roles.core',
  'job-roles/modules/job-roles.filters',
  'job-roles/modules/job-roles.run',
  'job-roles/modules/job-roles.services'
], function (angular) {
  'use strict';

  angular.module('hrjobroles', [
    'hrjobroles.core',
    'hrjobroles.config',
    'hrjobroles.run',
    'hrjobroles.constants',
    'hrjobroles.controllers',
    'hrjobroles.filters',
    'hrjobroles.services'
  ]);
});
```

This is the `hrjobroles.controllers` module before and after
```js
// before
define([
  'common/angular'
], function (angular) {
  'use strict';

  return angular.module('hrjobroles.controllers', []);
});
```
```js
// after
define([
  'common/angular',
  'job-roles/controllers/job-roles.controller',
  'job-roles/controllers/modal-dialog.controller'
], function (angular, JobRolesController, ModalDialogController) {
  'use strict';

  return angular.module('hrjobroles.controllers', [])
    .controller(JobRolesController.__name, JobRolesController)
    .controller(ModalDialogController.__name, ModalDialogController);
});
```
---

- [x] Tests Pass


com.civicrm.hrjobroles/js/src/job-roles/filters/get-active-values.filter.js
```diff
@@ -1,13 +1,17 @@
...
-], function(_,filters) {
+  'common/lodash'
+], function (_) {
   'use strict';
 
-  filters.filter('getActiveValues', ['$log', function($log) {
+  getActiveValues.__name = 'getActiveValues';
+  getActiveValues.$inject = ['$log'];
+
+  function getActiveValues ($log) {
```
reneolivo: Since filters are functions that can be used somewhere else, it should be good to explain what the filter function does and why.

com.civicrm.hrjobroles/views/include/add_new_role.html
```diff
@@ -169,169 +169,177 @@
...
-                      ng-model="funder.funder_id">
-                      <ui-select-match prevent-animations allow-clear placeholder="Select a funder" class="ui-select-match">
-                        {{$select.selected.sort_name}}
-                      </ui-select-match>
-                      <ui-select-choices
+          <div class="table-tab">
+            <table class="table">
+              <thead>
+                <tr>
+                  <th style="width:40%">Funder</th>
```
reneolivo: Just a comment: multiple sources state that the width attribute is not part of the HTML5 and using inline styles and classes are the recommended way to solve this issue. So this is perfectly fine.

[Source](http://help.simplytestable.com/errors/html-validation/the-x-attribute-on-the-y-element-is-obsolete-use-css-instead/the-width-attribute-on-the-td-element-is-obsolete-use-css-instead/)

reneolivo:
It would be good to squash some of these commits. They could be split in module renaming, linting and style guide, and fixing the actual issue.

com.civicrm.hrjobroles/js/test/services/filters.service.spec.js
```diff
@@ -0,0 +1,189 @@
+/* eslint-env amd, jasmine */
+
+define([
+  'common/angular',
+  'mocks/data/job-role.data',
+  'common/angularMocks',
+  'job-roles/modules/job-roles.module'
+], function (angular, Mock) {
```
reneolivo: `Mock` gives the impression that this is the Mock module, it's better to rename this variable to `mockJobRoles` or similar.

reneolivo:
Fixing previous issues with the files is outside of the scope of this PR so they should be addressed when the specific file is modified. So all is good here.

PR #2071, opened 2017-08-17
PCHR-2493: Disable hrvisa and hrident extensions

mickadoo:
## Overview
Since the functionality offered by the hrvisa and hrident extensions is planned to be replaced by the improved documents feature these extensions are not required going forward. However, as they are potentially used by existing clients we will not remove them from the code.

## Before
For new installations the hrvisa and hrident extensions are enabled.

## After
The hrvisa and hrident extensions will not be enabled for new installations. References to code that depends on these extensions will be removed or altered.

## Technical Details
This PR changes the installation script, but also adds a conditional for anywhere outside the hrvisa and hrident extensions that their code is used, or anywhere that uses custom groups, fields, activities or anything that is normally created by these extensions.

---

- [ ] Tests Pass
hrvisa tests are broken on staging :cry: 
hrident has only a single skipped test :crying_cat_face:

### DIFF
--- a/hrui/CRM/HRUI/Upgrader/Steps/4701.php
+++ b/hrui/CRM/HRUI/Upgrader/Steps/4701.php
@@ -46,6 +46,12 @@ trait CRM_HRUI_Upgrader_Steps_4701 {
     $identTableName = $this->up4701_getIdentTableName();
     $identFieldName = $this->up4701_getIdentFieldName();
 
+    $isEnabled = _hrui_is_extension_enabled('org.civicrm.hrident');
+
+    if (!$isEnabled) {
+      return TRUE;
+    }
+
     $query = "
       UPDATE {$inlineCustomGroup['table_name']}, $identTableName
          SET {$niSSNField['column_name']} = $identFieldName
--- a/hrui/CRM/HRUI/Upgrader/Steps/4701.php
+++ b/hrui/CRM/HRUI/Upgrader/Steps/4701.php
@@ -43,14 +43,14 @@ trait CRM_HRUI_Upgrader_Steps_4701 {
     $createResult = civicrm_api3('CustomField', 'create', $fieldData);
     $niSSNField = array_shift($createResult['values']);
 
-    $identTableName = $this->up4701_getIdentTableName();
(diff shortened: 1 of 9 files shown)


hrui/CRM/HRUI/Upgrader/Steps/4701.php
```diff
@@ -46,6 +46,12 @@ public function upgrade_4701() {
     $identTableName = $this->up4701_getIdentTableName();
     $identFieldName = $this->up4701_getIdentFieldName();
 
+    $isEnabled = _hrui_is_extension_enabled('org.civicrm.hrident');
```
tunbola: Why not add this before the `$identTableName` variable is assigned?

hrui/hrui.php
```diff
@@ -1034,3 +1031,14 @@ function _hrui_createDeveloperMenu(&$menu) {
...
+
+function _hrui_is_extension_enabled($key) {
+  $isEnabled = CRM_Core_DAO::getFieldValue(
+    'CRM_Core_DAO_Extension',
+    $key,
+    'is_active',
+    'full_name'
+  );
+
+  return  !empty($isEnabled) ? true : false;
```
tunbola: Seems there's an extra space that needs to be removed here before `!empty`.

tunbola:
Looks good @mickadoo, Just wondering how you tracked all the references to the two extensions and hope there won't be some left over somewhere that was missed?

hrui/CRM/HRUI/Upgrader/Steps/4701.php
mickadoo: Good idea

mickadoo:
@tunbola I also hope this. It will need to be tested well. I made a list of all the custom groups and fields that are created by the extensions and just searched for them. Then I checked for any activity types that are created and searched for them, and checked for all usages of public functions from the classes in these extensions. It's not such a clean way to do it, but I don't know any better way to find alll usages of anything related to these extensions. If you do please let me know :-)

tunbola:
I guess its fine then. Its needs to be well tested like you said.

PR #2075, opened 2017-08-17
PCHR-2482: Update Contract entitlement for Contact with existing Contract.

tunbola:
## Overview
The Job contract import was recently [modified](https://compucorp.atlassian.net/browse/PCHR-2482) to allow contract entitlements to be imported with each Absence type in its own column and the amount for the absence type for that contract can be entered in the CSV.  However contract entitlement information are imported along with the contract information for a contact and as such cannot work for the scenario where the contact has an existing contract and only entitlements need to be updated.

## Before
Contract entitlement can not be updated for a contact having an existing contract.

## After
Entitlements can now be updated for a contact having an existing contract. Find sample CSV attached to the [ticket](https://compucorp.atlassian.net/browse/PCHR-2482).
A new import mode `Update Current Contract Entitlements` was added on the Job contract import page.

updateentitlement.

## Technical Details
- A new parser class `CRM_Hrjobcontract_Import_Parser_EntitlementUpdate` was added to provide fields and handle the data to be imported in  `Update Current Contract Entitlements` mode.
- The Imported entitlements is imported for the latest revision of the current contract of the contact.
- The entitlements imported replaces whatever entitlement was stored against the contract revision before.
- If contract entitlements are not imported for all the enabled absence types, a default entitlement of 0 is added for the absence type missing in the CSV.

 - [X] Tests Pass


hrjobcontract/CRM/Hrjobcontract/Import/Form/Summary.php
```diff
@@ -1,4 +1,12 @@
 <?php
 class CRM_Hrjobcontract_Import_Form_Summary extends CRM_Hrjobcontract_Import_Form_SummaryBaseClass {
-  protected $_parser = 'CRM_Hrjobcontract_Import_Parser_Api';
-}
\ No newline at end of file
+
+  protected function getParser() {
```
davialexandre: I can see 2 problems with this `getParser()` method:

1. In all the places where this method is called, you instantiate the parser right after the call. Something like:
```php
$parser = $this->getParser();
$parser = new $parser($some, $params);
``` 
If the whole idea is to build a parser instance, wouldn't it make more sense for the `getParser()` to return that instead of the class name?

2. This `getParser()` method is duplicated in three different classes and it would be better to avoid that. Here are some suggestions:
- Create a `ParserFactory` class, with a method like `create($inputMode, $other, $params, $to, $constructor)`, which will return an instance of the correct parser for the given inputMode
- Move this method to a trait and use it in these classes

hrjobcontract/CRM/Hrjobcontract/Import/Parser/EntitlementUpdate.php
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Sets the fields for the contract entitlement import.
+   */
+  public function setFields() {
+    $this->_fields = array_merge(['do_not_import' => ['title' => ts('- do not import -')]], $this->getFields());
+  }
+
+  /**
+   * handle the values in preview mode
```
davialexandre: Please start the sentence with a uppercase letter

hrjobcontract/CRM/Hrjobcontract/Import/Parser/EntitlementUpdate.php
```diff
@@ -0,0 +1,226 @@
...
+   * @var array
+   *   Params for the current entity being prepared for the api.
+   */
+  protected $_params = [];
+
+  /**
+   * Sets the fields for the contract entitlement import.
+   */
+  public function setFields() {
+    $this->_fields = array_merge(['do_not_import' => ['title' => ts('- do not import -')]], $this->getFields());
```
davialexandre: Please break this in multiple lines to make it easier to read it

hrjobcontract/CRM/Hrjobcontract/Import/Parser/EntitlementUpdate.php
```diff
@@ -0,0 +1,226 @@
...
+   *
+   * @return bool
+   *   the result of this processing
+   */
+  public function import($onDuplicate, &$values) {
+    $response = $this->summary($values);
+    if ($response != CRM_Import_Parser::VALID) {
+      return $response;
+    }
+    $params = &$this->getActiveFieldParams();
```
davialexandre: Is the & operator really necessary here?

PR #2076, opened 2017-08-18
PCHR 2521: Leave balance tab pagination

reneolivo:
## Overview
This PR adds pagination capabilities to the leave balance tab.

## Before
anim
Pagination links do nothing when clicked.

## After
anim
* Only displaying 10 records to showcase how the pagination works.
* When filters are updated, the report always goes to the first page.

## Technical Details
* `uib-pagination` was used for pagination.
* `leave-balance-tab` has a pagination object that is managed by uib-pagination and is passed to the leave balance report model to retrieve a specific page and amount of records.
* a new `loadReportCurrentPage` public method was added. This method is fired from the view when the page changes, and it's also fired when the filters are updated. In this last case, it will load the first page of the report.

## Comments
* Contact links were removed since they won't allow viewing the contact details for now.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
```diff
@@ -22,16 +22,18 @@ define([
...
     LeaveBalanceReport, notification, Session) {
     var filters = {};
-    var pageSize = 50;
     var vm = this;
 
     vm.absencePeriods = [];
     vm.absenceTypes = [];
     vm.loading = { component: true, report: true };
     vm.loggedInContactId = null;
+    vm.pagination = { page: 1, size: 50 };
```
deb1990: Here the size is set to 50 but in the screenshot it is showing 10 at a time. Is the screenshot not updated or there is something wrong with the logic.

deb1990:
@reneolivo The commit name `PCHR-2478: add leave-balance-tab-filter component` doe not make sense as there is no component created for the same. Could you squash and reword the same please.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
reneolivo: I made a comment, on the screenshot I'm only displaying 10 records to showcase pagination ( because I only have 26 records available :'( )

deb1990: Okay no problem :)

PR #2078, opened 2017-08-21
PCHR-2501: Allow Case Types with "&" in activity types

mickadoo:
## Overview
This is a recognized, and fixed issue in CiviCRM core, issue [CRM-16360](https://issues.civicrm.org/jira/browse/CRM-16360). Assignment types that use ampersands in the activity type names will not be interpreted correctly, since `&` is a reserved character. The definition will be saved to the database, but loading it on edit will cause the standard timeline tab to not be displayed, and any attempt to save on this screen will result in an empty definition.

## Before
After saving a case type with & in the definition, attempts to edit it would not show a standard timeline. Further saving would delete most of the definition from the database.

## After
The fix was applied to the overridden file. You can save case types with `&` in the definition and they will be correctly encoded before saving to the database.

## Comments
I copied and pasted the entirety of https://github.com/civicrm/civicrm-core/blob/4.7.18/CRM/Case/BAO/CaseType.php with the exception of the line to [schedule reconciliation](https://github.com/civicrm/civicrm-core/blob/4.7.22/CRM/Case/BAO/CaseType.php#L79), which is still an open issue for us related to the saving of managed entities.

---

- [x] Tests Pass

hrcase just had a single (skipped) test.

### DIFF
--- a/hrcase/CRM/Case/BAO/CaseType.php
+++ b/hrcase/CRM/Case/BAO/CaseType.php
@@ -1,9 +1,9 @@
 <?php
 /*
  +--------------------------------------------------------------------+
- | CiviCRM version 4.5                                                |
+ | CiviCRM version 4.7                                                |
  +--------------------------------------------------------------------+
- | Copyright CiviCRM LLC (c) 2004-2014                                |
+ | Copyright CiviCRM LLC (c) 2004-2017                                |
  +--------------------------------------------------------------------+
  | This file is a part of CiviCRM.                                    |
  |                                                                    |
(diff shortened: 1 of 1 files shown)


omarabuhussein:
two things : 

1- I think CiviHR is now upgraded to work with CiviCRM 4.7.22, so you may need to upgrade this overridden file to that version instead of CiviCRM 4.7.18 version.

2- I know you are not the one who override this page originally, but can you add it to this page https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/100532509/Overridden+CiviCRM+files .

mickadoo:
@omarabuhussein you're right - it should be v4.7.22, luckily there were no changes to this file from 4.7.18 to 4.7.22 so I'll leave it as it is.

I added this file to the list of overridden files.

mickadoo:
Closing as removing the override as part of PCHR-1928 solves this issue.

PR #2079, opened 2017-08-21
PCHR-2505: Use decimals for Absence Type fields that store date amounts

davialexandre:
## Overview
It was not possible to save an Absence Type with a decimal value for the default entitlement (20.5, for example). When trying to do that, the user would get a "The value should be a positive number" error.

## Before
The error happened because the AbsenceType form was using the `positiveInteger` validation for this field, which fails if the number if not an integer or if it's an integer but not positive.

Besides the validation, there was also a problem with the field type in the database. It was an `unsigned int` and it was not capable of storing the decimal part of the value.

## After
The `positiveInteger` validation isn't used anymore. A custom validation for positive decimals was added and it checks if the number is either a positive integer or decimal and, if it's a decimal, that it has only 1 digit.

The database field type was updated to be a decimal.

## Technical Details
- An upgrader was added to change the fields type on existing sites
- The AbsenceType.get API returns decimal numbers as strings instead of float. This was causing the default entitlement to always be displayed with the decimal digits on the Job Contract modal, even when it didn't have any (for example, when the default entitlement is 10). To fix this, the `UtilsService.getAbsenceTypes()` methods was updated to convert the `default_entitlement` values to floats, right after they are fetched.
- Some tests were updated to use decimal values for the `default_entitlement` in order to also add some coverage for this scenario

## Comments
Besides the `default_entitlement` field, there were other fields of the AbsenceType entity that are used to store date amounts but didn't support decimal values. All of them were updated to use the new decimal validation and their types in the db were updated accordingly.

---

- [x] Tests Pass
The L&A phpunit tests are all passing. The hrjobcontract angular tests are failing, but the failure is not related to the changes in this PR and they will be fixed by a later PR.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -186,9 +186,28 @@ public static function getAvailableColors() {
    * @return array An array containing the values
    */
   public static function getValuesArray($id) {
-    $result = civicrm_api3('AbsenceType', 'get', array('id' => $id));
+    $result = civicrm_api3('AbsenceType', 'get', ['id' => $id]);
```
tunbola: Any reason why we are not using the BAO to retrieve the Absence Type details here. We can cast the object to an array after fetching. 
Just thinking it will reduce unnecessary API overhead.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -186,9 +186,28 @@ public static function getAvailableColors() {
...
+      'max_consecutive_leave_days',
+      'max_leave_accrual',
+      'max_number_of_days_to_carry_forward'
+    ];
+    foreach($decimalFields as $decimalField) {
+      if(!empty($absenceType[$decimalField])) {
+        // The API return these values as strings, exactly the way they are stored
+        // in the BD (0.00), so we need this to:
+        // 1. Make sure the numbers will always be displayed with only 1 decimal digit
+        // 2. Make sure the number without decimal digitls (10.00, for example) will
```
tunbola: s/digitls/digits

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
davialexandre: No reason to be honest. I believe that, when I first created it, I based myself in some other form that was using the same strategy.

Since this is being used only by the form, I think it's fine to leave it as it is.

PR #2080, opened 2017-08-21
PCHR-2353: Add Contact Job Role and amend Contact entities

igorpavlov:
## Overview

In the Leave Calendar managers cannot use filters. This PR fixes this issue.

## Technical Details

The issue's reason is the lack of default permissions provided to managers.
1) The new entity ContactJobRole was created to provide a consistent endpoint for contact job roles. It utilises ContactHRJobRole API provided by https://github.com/civicrm/civihr/pull/2053
2) This entity is now used in Contact model instead of Job Role

---

- [x] Tests Pass


org.civicrm.reqangular/test/models/contact.spec.js
```diff
@@ -220,14 +224,14 @@ define([
...
-                .finally(done) && $rootScope.$digest();
+        .finally(done) && $rootScope.$digest();
       });
 
       it('returns an instance of the model', function (done) {
         Contact.find(targetId).then(function (contact) {
           expect(ContactInstanceMock.isInstance(contact)).toBe(true);
         })
-                .finally(done) && $rootScope.$digest();
+        .finally(done) && $rootScope.$digest();
```
igorpavlov: Done.

PR #2084, opened 2017-08-22
PCHR-2527: Make calendar show requests touching different months

igorpavlov:
## Overview

Currently the Leave Calendar does not reflect leave requests that have start and end dates in different months.

## Technical Details

Previously, only requests that had the both start and end dates from the specified month are shown.

```js
from_date in January AND to_date in January
```

Now, all requests that have at least one day from the specified months are shown.

```js
from_date in January OR to_date in January
```

## Comments

1. The additional check is needed since in this PR the Calendar Month component now gets requests that might touch days from other months. These days should not be considered since they are not going to be rendered.

```js
while (pointerDate.isSameOrBefore(toDate)) {
  // Ensure that pointerDate is in same month/year that component represents
  if (pointerDate.month() === vm.month.index && pointerDate.year() === vm.month.year) {
    days.push(_.find(vm.month.days, function (day) {
      return day.date === pointerDate.format('YYYY-MM-DD');
    }));
  }

  pointerDate.add(1, 'day');
}
```

2. Please see this PR https://github.com/civicrm/civicrm-core/pull/9844 for the reference to the OR operator support in CiviCRM.

⚠️ This PR can only be merged after https://github.com/civicrm/civihr/pull/2085 is merged and synced.

---

- [x] Tests Pass

PR #2085, opened 2017-08-22
PCHR-2527: Allow custom options to be passed to API calls

igorpavlov:
## Overview

Currently the `api.getAll()` function ignores custom options passed as a filter. This PR allows to use custom options in the `getAll()` function.

## Technical Details

```js
// The OR expression can now be used, for example
api.getAll(entity, {
  options: { or: [['field1', 'field2', 'field3']] }
});
```
## Comments

This PR is needed to accomplish https://github.com/civicrm/civihr/pull/2084.

Please see this PR https://github.com/civicrm/civicrm-core/pull/9844 for the reference to the OR operator support in CiviCRM.

---

- [x] Tests Pass


org.civicrm.reqangular/test/services/api.spec.js
```diff
@@ -185,11 +193,59 @@ define([
...
+          spyOn(api, 'sendGET').and.returnValue(returnValue);
+          $rootScope.$digest();
+        });
+
+        describe('when no custom options are passed', function () {
+          beforeEach(function () {
+            api.getAll(entity, {});
+          });
+
+          it('default options are still used in the API call', function () {
```
deb1990: The description of the test should make a sentence with `it`.
Example: it - "still uses the default options in the API call".
Please check other instances.

PR #2087, opened 2017-08-22
PCHR-2503: Fix issues with being unable to export Contract Entitlements

tunbola:
## Overview
Currently, trying to export contract leave entitlements when exporting a contact information does not work, the resulting CSV downloaded contains a blank column or inaccurate data for the Contract Leave Amount column.
Also since the import format for Contract entitlements has been changed in CiviHR 1.7, there is need also to allow Contract entitlements to be exported in this format to allow clients migrating entitlements data from 1.6 to 1.7 to do so seamlessly.

## Before
Exporting the contract leave entitlement data is broken and does not give accurate data. The export format is not compatible for import into a CiviHR 1.7 site

## After
Exporting the contract leave entitlement now gives accurate data and it also exports data in a format that is compatible for import into 1.7 in addition to the format compatible with 1.6.

exportentitlement

## Technical Details
- The [function](https://github.com/civicrm/civihr/blob/d1a60dd47bc4eab84e4a5475947b2302027963f7/hrjobcontract/CRM/Hrjobcontract/ExportImportValuesConverter.php#L254-L254) responsible for formatting the `Contract Leave Amount` column was expecting data in the form `Leave Type ID 1: Entitlement Amount1, Leave Type ID 2 : Entitlement Amount2` but was currently receiving data in the form `Entitlement Amount1` with each Leave entitlement amount in its own row rather than the whole contract entitlement being concatenated on a single row for the contact. 
The query responsible for fetching the associated exportable columns for the Job contract extension was modified to GROUP concatenate the Leave Type ID with the Leave entitlement amount in the format expected by the `leave_amount` export formatter whenever the Contract Entitlements are to be exported.
- The `hrjobcontract_leave_leave_amount` [type](https://github.com/civicrm/civihr/blob/7b1567bb5f9135c8574b707967409a6ee7614c91/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php#L177-L177) was modified to a string. This type is used by Civi to determine the column type and length when creating a temporary table to store the results of the export. Was getting a `Data too long for column 'hrjobcontract_leave_leave_amount` error before changing the type to String. Technically the type is supposed to be  a string because when importing Contract entitlements, the contract entitlements are imported as comma separated pairs of Leave type/Entitlement amount.
- After fixing the contract entitlements to export data in the format `Leave Type Name 1: Entitlement Amount1, Leave Type Name 2 : Entitlement Amount2`, the overall results was intercepted to also return these entitlements as actual columns with accompanying data in a [format](https://github.com/civicrm/civihr/pull/2044) supported in CiviHR 1.7.

## Comments
- This solution has also been tested to work on a 1.7 site, tested by locally syncing  1.6 with 1.7.

- [X] Tests Pass

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
@@ -174,7 +174,7 @@ class CRM_Hrjobcontract_DAO_HRJobLeave extends CRM_Hrjobcontract_DAO_Base
               ) ,
               'hrjobcontract_leave_leave_amount' => array(
                 'name' => 'leave_amount',
-                'type' => CRM_Utils_Type::T_STRING,
+                'type' => CRM_Utils_Type::T_INT,
                 'title' => ts('Contract Leave Amount') ,
                 'import' => true,
                 'where' => 'civicrm_hrjobcontract_leave.leave_amount',
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
@@ -174,7 +174,7 @@ class CRM_Hrjobcontract_DAO_HRJobLeave extends CRM_Hrjobcontract_DAO_Base
               ) ,
               'hrjobcontract_leave_leave_amount' => array(
                 'name' => 'leave_amount',
-                'type' => CRM_Utils_Type::T_INT,
+                'type' => CRM_Utils_Type::T_STRING,
                 'title' => ts('Contract Leave Amount') ,
                 'import' => true,
                 'where' => 'civicrm_hrjobcontract_leave.leave_amount',
--- a/hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
@@ -69,6 +69,9 @@ class CRM_Hrjobcontract_Export_Converter {
               $contractEntitlementsArray[trim($leaveType)] = trim($entitlement);
             }
             foreach ($leaveTypes as $type => $typeId) {
+              //The exported file will always have columns for all the absence type, so we need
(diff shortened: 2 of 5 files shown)


hrjobcontract/CRM/Hrjobcontract/DAO/HRJobLeave.php
```diff
@@ -174,7 +174,7 @@ static function &fields()
               ) ,
               'hrjobcontract_leave_leave_amount' => array(
                 'name' => 'leave_amount',
-                'type' => CRM_Utils_Type::T_INT,
+                'type' => CRM_Utils_Type::T_STRING,
```
davialexandre: I saw the explanation about this on the PR description, but I'm not sure of why it has to be a string and not a decimal.

Regardless of the type, if you change the DAO you have to:
- Upgrade the docblock related to this field
- Update the install SQL file
- Add an upgrader to update existing sites

hrjobcontract/CRM/Hrjobcontract/Export/Converter.php
```diff
@@ -43,6 +60,18 @@ public static function writeCSVFromTable($exportTempTable, $headerRows, $sqlColu
...
+          if ($column === 'hrjobcontract_leave_leave_amount') {
+            $contractEntitlementsArray = [];
+            $contractEntitlements = explode(',', $row[$column]);
+            foreach ($contractEntitlements as $contractEntitlement) {
+              list($leaveType, $entitlement) = explode(':', $contractEntitlement);
+              $contractEntitlementsArray[trim($leaveType)] = trim($entitlement);
+            }
+            foreach ($leaveTypes as $type => $typeId) {
+              $row[$type] = !empty($contractEntitlementsArray[$type]) ? $contractEntitlementsArray[$type] : '';
+            }
```
davialexandre: If I got this right, the exported file will always have columns for all the absence types, because some contacts might have entitlement for an absence type and others don't. Is this correct? If so, I think it would be nice to add a comment here explaining this

tunbola: Yes that's correct.

PR #2091, opened 2017-08-24
PCHR-2537: Change Accrued TOIL logic for calendar

deb1990:
## Overview
This ticket fixes the following bug,
When a Overtime leave request is created from `my-leave-calendar` page and the popup is closed, the calendar shows the newly created request in BLUE(which is wrong) and once refreshed, the same leave request is shown as AT(not blue)(which is correct)

## Before
calendar bug

## Technical Details
1. The following logic is used to determine if a request is accrued TOIL, `leaveRequest.request_type === 'toil'`

---

- [x] Tests Pass


deb1990:
Its not required anymore, closing

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -473,9 +473,9 @@
             });
           });
 
-          describe('when the balance change of the leave request is positive', function () {
```
reneolivo: what happens to other request types with a positive balance? are we deleting that functionality?

deb1990: The TOIL `request_type` is a group, and any absence type which has a `allow_accruals_request` set to true, false into this category. So any request type with positive balance should already have `allow_accruals_request` set to true, so this will work too. cc @davialexandre

davialexandre: That's correct @deb1990. Currently, only toil requests can have a positive balance_change

PR #2081, opened 2017-08-22
PCHR-2513: Refactor leave popup

deb1990:
## Overview
As part of this PR, a new component has been developed for the details tab of the leave popup.

### `leave-request-popup-details-tab` component
It abstracts the functionality of the details tab.
#### Markup
```html
 <leave-request-popup-details-tab
  request="" 
  is-mode="" 
  period="" 
  absence-types=""
  is-leave-status="" 
  selected-absence-type="" 
  balance=""
  absence-periods="" 
  is-role="" 
  leave-type=""
  check-submit-conditions="" >
</leave-request-popup-details-tab>
```

#### Attributes
1. `request`: {Object} The request object.
2. `is-mode`: {Function} 
3. `period`: {Object} absence period for the current date, it is set in the parent controller and required by both parent and this component. 
4. `absence-types`: {Object} it is set in the parent controller and required by both parent and this component. 
5. `is-leave-status`:  {Function} 
6. `selected-absence-type`: {Object} it is set in the parent controller and required by both parent and this component. 
7. `balance`: {Object} it is required by both parent and this component. 
8. `absence-periods`: {Object} it is set in the parent controller and required by both parent and this component. 
9. `is-role`: {Function}  
10. `leave-type`: {Object} Passed from parent.
11. `check-submit-conditions`: {Function} This function is required by the `request.controller` to 
determine if the leave request can be submitted, so it is defined in `leave`, `sick` or `toil` child controllers so that the `request.controller` can access it.

### Change in architecture
**Before**
1. `request.controller` was used to manage the Header, Footer and the Details tab of the leave popup. 
2. `leave-request.controller`, `sick-request.controller`, `toil-request.controller` were child controllers of the `request.controller`.
3. `leave-popup.service`used to decide which child controller to use for the popup, based on leave type.

**Now**
1. `request.controller`now manages only the header and footer section.
2. `leave-request-popup-details-tab` component manages the Details tab.
3. `leave-request.controller`, `sick-request.controller`, `toil-request.controller` are child controllers of the `leave-request-popup-details-tab` component.
4. `leave-popup.service` always uses `request.controller`.

### Other Details
1. The markup for the body of the popup has been moved to `leave-request-popup-body.html`, which increased the readability in the `leave-request-popup.html`
```html
<ng-include
  src="sharedPathTpl + 'directives/leave-request-popup/leave-request-popup-header.html'">
</ng-include>

<crm-loading show="!$ctrl.loading.absenceTypes">
  <ng-include
[...]

PR #2092, opened 2017-08-24
PCHR-2513: Refactor leave popup

deb1990:
Original PR, https://github.com/civicrm/civihr/pull/2081 was reverted because the ticket needs to be tested separately

### Update
New commits added to fix the issues reported are
https://github.com/civicrm/civihr/pull/2092/commits/49089344d2c3bc3c2890045732a1cd88491f094c

### Technical Details
1. `isSelfRecord` added to the bindings of the details tab component, it was missed before.
2. New variable `staffMemberSelectionComplete` is defined on the `request.controller`, which is used to load all child tabs. Having this variable makes sure that the child components are injected only when all necessary data is loaded. 
Also this is used to hide all fields which should not be shown before the contact is selected. Eg: 'Set status to' field
3. In `leave-request-popup-details-tab.component.js` the code is changed from `calculateBalanceChange()` to `vm.calculateBalanceChange()`, so that the inherited function(in TOIL Controller) is called.
3. Markup is changed to show "Please select the Staff Member above", before the contact is selected by the manager.

select contact


igorpavlov:
Could you please explain why `$rootScope.$on('LeaveRequestPopup::ContactSelectionComplete', afterContactSelection);` was deleted?

deb1990:
>Could you please explain why $rootScope.$on('LeaveRequestPopup::ContactSelectionComplete', afterContactSelection); was deleted?

It was added to make the child component aware, when all the necessary data has been loaded in the parent. But this was not working properly, because of `tick` related issues, i.e. even when some of the data was loaded on the parent, the same variable was not getting updated on the child instantly. As it was waiting for a tick to get updated. That is the reason the child injection has been delayed, and it is now injected only when all the necessary data has been loaded.

PR #2093, opened 2017-08-24
PCHR-2450: Contacts to display on admin calendar

igorpavlov:
## Overview

Currently the **admin** Leave Calendar shows all contacts in the system. This is unnecessary since there are contacts who neither have a contract in the selected absence period nor leave requests. In other words, there are currently contacts with zero data.

This PR reduces contacts shown to admin to only those who either have a contract in the selected absence period *or* have leave requests.

ℹ️ Please note that the calendar only shows "Approved", "Admin Approved" or "Waiting for Approval" statuses.

ℹ️ The calendar was tested in both Manager Leave and Admin sections.

## Before

Calendar defaults to showing all contacts and completely ignores the selected Absence Period.

## After

Calendar defaults to showing only contacts that either have leave requests or have contracts for the absence period.

## Technical Details

1. Admin Calendar Sub Controller now loads all Contracts along with all Contacts and caches Contracts:
```js
//... Loads Contracts now as well:
function api () {
  return {
    loadContacts: function () {
      return Contact.all()
        //...
        .then(function () {
          return $q.all([
            loadContacts(),
            getContactIdsToReduceTo()
          ]);
        })
        //...
    }
  };
}
//... Then create a list of Contact IDs to reduce the final list of Contacts in the Calendar to:
function getContactIdsToReduceTo () {
  return loadContracts()
  .then(function (contracts) {
    var contractsInAbsencePeriod = contracts.filter(function (contract) {
      var details = contract.info.details;
      var isContractInAbsencePeriod = (
        (!details.period_start_date ||
        moment(details.period_start_date).diff(moment(vm.selectedPeriod.end_date)) <= 0) &&
        (!details.period_end_date ||
        moment(details.period_end_date).diff(moment(vm.selectedPeriod.start_date)) > 0)
      );

      return isContractInAbsencePeriod;
    });

    vm.contactIdsToReduceTo = _.uniq(contractsInAbsencePeriod.map(function (contract) {
      return contract.contact_id;
    }));
  });
}
```
2. The list of Contact IDs generated above is passed to the Main Calendar Controller and then passed to Month Calendar Controller. Also if this list exists in the Main Controller it forces the data to reload.
```js
//... This can be either null or an array
vm.contactIdsToReduceTo = null;
//...
```
```html
<leave-calendar-month ng-repeat="month in calendar.months track by $index"
  ...
  contact-ids-to-reduce-to="calendar.contactIdsToReduceTo"
  ...>
</leave-calendar-month>
```
3. The Month Calendar Controller loads Leave Requests and has the list of Contact IDs passed, so it now has all the data to show the right Contacts.

```js
function reduceContacts () {
  if (vm.contactIdsToReduceTo) { // <-- ONLY IF THIS PROP EXISTS, OTHERWISE - LEAVE IT AS IS
    vm.contacts = vm.contacts.filter(function (contact) {
      return (_.includes(vm.contactIdsToReduceTo, contact.contact_id) ||
        _.find(leaveRequests, function (leaveRequest) {
          return leaveRequest.contact_id === contact.contact_id;
        }));
    });
  }

  return $q.resolve();
}
```

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-calendar.html
```diff
@@ -88,6 +88,7 @@
       <div ng-if="calendar.injectMonths">
         <leave-calendar-month ng-repeat="month in calendar.months track by $index"
           contacts="calendar.contacts"
+          contact-ids-to-reduce-to="calendar.contactIdsToReduceTo"
```
reneolivo: `contacts` attribute should pass the filtered contacts. This should not be a responsibility of `leave-calendar-month`. `contact-ids-to-reduce-to` should be removed and the filtering logic should be moved to parent controller.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-admin.controller.js
```diff
@@ -60,6 +70,46 @@ define([
...
+     * to the currently selected absence period
+     *
+     * @return {Promise}
+     */
+    function getContactIdsToReduceTo () {
+      return loadContracts()
+      .then(function (contracts) {
+        var contractsInAbsencePeriod = contracts.filter(function (contract) {
+          var details = contract.info.details;
+          var isContractInAbsencePeriod = (
```
reneolivo: instead of:
```js
var isContractInAbsencePeriod = (
  (!details.period_start_date ||
  moment(details.period_start_date).diff(moment(vm.selectedPeriod.end_date)) <= 0) &&
  (!details.period_end_date ||
  moment(details.period_end_date).diff(moment(vm.selectedPeriod.start_date)) > 0)
);
```

you could try:
```js
var isContractInAbsencePeriod = (
  moment(details.period_start_date).isSameOrBefore(vm.selectedPeriod.end_date) &&
  moment(details.period_end_date).isAfter(vm.selectedPeriod.start_date)
);
```
(And why > 0 instead of >= 0?)

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-calendar.html
igorpavlov: Unfortunately, leave-calendar-month loads Leave Requests. The new logic for additional filtering of Contacts is *Contacts have either Leave Requests OR Contracts*. Leave requests are loaded in the `leave-calendar-month` while Contracts **have to be loaded** in `admin-leave-calendar` since this logic relates to the Admin only. We had this discussion with @AkA84 and it seems that this is the only proper way to go.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/leave-calendar-admin.controller.js
igorpavlov: Unfortunately, Contract may **not have** `end_date`.

image

And moment will return `false`, while I it needs to return `true` since if there is no defined end date - the contract automatically falls into the selected absence period.

image

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-calendar.html
reneolivo: After having a call with @igorpavlov this is the best solution we can provide so far without blocking the ticket for too long, and this feature is critical. But there needs to be refactoring done because the logic to display contacts is so confusing that it needs to be explained in a call, it can't be inferred by simply looking at the components.

Here are some of my suggestions for a future ticket:

* Stop using controllers as services:
```js
   // leave-calendar.component.js #131
    function injectSubController () {
      subController = $controller('LeaveCalendar' + _.capitalize(userRole) + 'Controller').init(vm);
    }
```

Instead of using controllers for each role, these should be services (ex: `adminLeaveCalendarContactsService`). 

* A map **value** can be created to retrieve the proper service. This:
```js
// note: the services names are examples
module.value('leaveCalendarContactsByRole', function (adminService, managerService, staffService) {
  return { admin: adminService, manager: managerService, staff: staffService });
});

//and use it like so:
contactService = leaveCalendarContactsByRole['admin'];
```

Instead of:
```js
subController = $controller('LeaveCalendar' + _.capitalize(userRole) + 'Controller').init(vm);
```

* the services should return the contacts that are to be displayed on the calendar. The component should not reduce any value or apply any filters to the contacts to display. This will make the component unaware of any business logic of the contacts to display depending on the role, and just focuses on displaying the contacts it was passed as an attribute or retrieved from a service.

PR #2094, opened 2017-08-24
PCHR-2480: Fix bug in unassigned filter when contact has multiple relationships

tunbola:
## Overview
The unassigned filter on the Leave and Absence Dashboard page does not work as it should. For example leave requests for a contact with an active leave approver relationship also having another relationship will show up when the unassigned tab is clicked. The unassigned tab is only supposed to show leave requests for contacts without an active leave approver relationship.

## Before
The unassigned filter when clicked will return leave request results for a contact having an active leave approver relationship and other relationships.

## After
The unassigned filter when clicked returns only results contacts without an active leave approver relationship.

## Technical Details
- The query responsible for fetching results for `LeaveRequest.get` and `LeaveRequest.getFull` Api's joins directly to the `relationship` table via the leave request `contact_id`, This will make a Join to all relationships that the contact has, For the case of when unassigned is true, the query does not factor in the scenario where a contact already has leave manager relationship and the contact is returned with the results because it has other relationships that are not leave manager relationships.
- The query is modified to filter off contacts that already have active leave manager relationship and the results returned.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
@@ -103,22 +103,13 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect {
       $conditions[] = "NOT (" . implode(' AND ', $this->hasActiveLeaveManagerCondition()) . ") 
                        OR (r.is_active IS NULL AND rt.is_active IS NULL)";
 
-      $contactID = false;
-      $contactPassedAsArray = !empty($this->params['contact_id']['IN']);
-      $contactPassedAsID = isset($this->params['contact_id']) && is_numeric($this->params['contact_id']);
 
-      if($contactPassedAsArray) {
-        $contactID = $this->params['contact_id']['IN'];
-      }
-
-      if($contactPassedAsID) {
-        $contactID = [$this->params['contact_id']];
-      }
 
       $query = "a.contact_id NOT IN(SELECT contact_id_a FROM civicrm_relationship r LEFT JOIN
                        civicrm_relationship_type rt ON rt.id = r.relationship_type_id WHERE
                        ". implode(' AND ', $this->hasActiveLeaveManagerCondition());
 
+      $contactID = $this->getContactIdFromParams();
       if ($contactID) {
         $query .= " AND r.contact_id_a IN(" . implode(',', $contactID) . "))";
       }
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -102,6 +102,31 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
...
+      $contactPassedAsArray = !empty($this->params['contact_id']['IN']);
+      $contactPassedAsID = isset($this->params['contact_id']) && is_numeric($this->params['contact_id']);
+
+      if($contactPassedAsArray) {
+        $contactID = $this->params['contact_id']['IN'];
+      }
+
+      if($contactPassedAsID) {
+        $contactID = [$this->params['contact_id']];
+      }
```
davialexandre: Please extract this logic to get the contact ID from the params to a separate method

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -102,6 +102,31 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
...
+
+      $query = "a.contact_id NOT IN(SELECT contact_id_a FROM civicrm_relationship r LEFT JOIN
+                       civicrm_relationship_type rt ON rt.id = r.relationship_type_id WHERE
+                       ". implode(' AND ', $this->hasActiveLeaveManagerCondition());
```
davialexandre: This query already has another subquery and it would be good if we could avoid this new one. Have you tried to see if this was possible?

tunbola: Tried to think of a better way but this seemed to be the way out in this case. Do you have some suggestions?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -102,6 +102,22 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
     if($hasUnassignedAsTrue) {
       $conditions[] = "NOT (" . implode(' AND ', $this->hasActiveLeaveManagerCondition()) . ") 
                        OR (r.is_active IS NULL AND rt.is_active IS NULL)";
+
+
+
```
davialexandre: please leave only one empty line here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveRequestSelect.php
```diff
@@ -312,4 +328,26 @@ private function hasActiveLeaveManagerCondition() {
...
   }
+
+  /**
+   * Gets the contactID from the params array and
+   * returns it as an array.
+   *
+   * @return array|bool
+   */
+  private function getContactIdFromParams() {
+    $contactID = false;
```
davialexandre: why not an empty array or null?

PR #2099, opened 2017-08-29
PCHR 2471: Fix opening balance for open requests

reneolivo:
## Overview
When trying to edit a leave request, it will display the opening balance as the remaining balance - the leave request change which leads to wrong closing balances. Example:

1. The original Opening Balance was 10.
2. The Balance Change was 2.
3. The new Balance is 8 after the request is saved.
4. When editing the request the Opening Balance will be 8 and the Closing Balance is 6. The correct values should be 10 and 8 respectively.

This PR fixes this issue as well as refactoring component initialization on test file.

## Before
anim
Balance is 9, opening balance is 9, change is 6, closing balance is 3.

## After
anim
Balance is 9, opening balance is 15, change is 6, and closing balance is 9.

## Technical Details
Only when editing, an original opening balance object is created:
```js
originalOpeningBalance = {
   absenceTypeId: vm.request.type_id,
   value: vm.selectedAbsenceType.remainder - vm.request.balance_change
};
```
Since balance_change is negative, the value will hold the original balance before the request was created. Using the previous example this would be 9 - -6 = 9 + 6 = 15.

When calculating the new opening and closing balances, if the selected absence types is the original one, it will use the original balance:
```js
if (originalOpeningBalance &&
originalOpeningBalance.absenceTypeId === vm.selectedAbsenceType.id) {
   vm.balance.opening = originalOpeningBalance.value;
} else {
   vm.balance.opening = vm.selectedAbsenceType.remainder;
}
```

## Comments
Components initialization for this test was refactored. Instead of calling `compileComponent` with 6 arguments, now you call it by passing 1 object argument. The benefits of this approach are two:

1. No need to remember the order of the parameters as long as you remember the names.
2. Defaults were set in place to ease the component initialization.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -314,6 +316,21 @@ define([
     }
 
     /**
+     * Initialize the original opening balance when in edit mode. This allows
+     * to display the opening balance before the request was created. The
+     * formula is absence type reminder + balance change (balance change is a
+     * negative number so it needs to be subtracted).
+     */
+    function initOriginalOpeningBalance () {
+      if (vm.isMode('edit')) {
```
deb1990: Original Opening Balance, should be saved for approved requests in edit mode only, right? Then you are not checking if it is approved request only.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -464,7 +481,12 @@ define([
      * Calculates and updates opening and closing balances
      */
     function _calculateOpeningAndClosingBalance () {
-      vm.balance.opening = vm.selectedAbsenceType.remainder;
+      if (originalOpeningBalance &&
+        originalOpeningBalance.absenceTypeId === vm.selectedAbsenceType.id) {
```
deb1990: Please leave a comment in the code, explaining why checking the type is necessary.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
reneolivo: Sorry, my bad, it's not just for open leave request, it's for any leave request that has already been saved.

deb1990: I think the requirement for the ticket is applicable for approved requests only, because only when a request is approved, the balance gets updated in the backend. Doing the same for non approved requests, will be wrong, is not it?
Suppose you have entitlement for 20 days, and a leave request for 10 days has been created, but till it is approved, the `selectedAbsenceType.remainder` still remains 20. So if you show 20+10 =30, it will be wrong.

reneolivo: in that case the remainder will be 10. I double checked, and even if the request is in an open state the remainder stores the balance as if the request is approved.

reneolivo: Here's the *after* for a request in a open state:
anim

PR #2101, opened 2017-08-30
PCHR-2149: Merge Actions into One

mickadoo:
## Overview
We had planned for two separate bulk action, one to create user accounts and one to send invitation emails. This was unified into a single action "Create User Account(s)"

## Before
There were two forms, one for creating user records, the other for sending invitation emails. You could not assign default roles to users.

## After
There is a single bulk action for creating user records, which includes the option to add roles to selected users.

## Comments

- It is mandatory to select at least one role
- Bootstrap class are now used to style the form
- Previously the specs demanded new activity types be created when account was created or email send, but this was removed from the requirements.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -25,8 +26,7 @@ public function buildQuickForm() {
   public function preProcess() {
     parent::preProcess();
 
-    $haveNoAccount = $this->getContactsWithoutAttribute('uf_id');
-    $haveAccount = array_diff_key($this->contactDetails, $haveNoAccount);
+    $haveAccount = $this->getContactsWithAccount();
```
MiyaNoctem: Is there a reason to create this variable? If not, for consistency, can you just assign the value for the form by calling the getContactsWithAccount() method?

```php
$this->assign('contactsWithAccount', $this->getContactsWithAccount());
```

mickadoo: You're right, it was used before but it doesn't need to be separate here.

PR #2103, opened 2017-08-30
PCHR-2558: Add change_balance parameter to LeaveRequest.create API.

tunbola:
## Overview
When a user edits an already created leave and changes any other thing apart from the date, the user can decide to either use the balance change as of when the leave request was created or recalculate if the balance change is now different. This PR implements the Backend changes needed to achieve this.

## Before
There is no way for the `LeaveRequest.create` API to know whether the leave balance for an already created request should be retained or recalculated. 

## After
The `LeaveRequest.create` API was modified to accept a new parameter called `change_balance`. This parameter when present and true will allow the leave request balance changes to be updated using the current work pattern for the contact. If not present or if present and false means that the leave request balance will remain untouched.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -41,13 +41,14 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     }
     unset($params['is_deleted']);
 
+    $datesChanged = self::datesChanged($params);
     $instance = new self();
     $instance->copyValues($params);
 
     $transaction = new CRM_Core_Transaction();
     try {
       $instance->save();
-      $instance->saveDates();
+      $instance->saveDates($datesChanged);
       $transaction->commit();
     } catch(Exception $e) {
       $transaction->rollback();
@@ -752,10 +753,21 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   }
 
   /**
-   * Creates and saves LeaveRequestDates for this LeaveRequest
+   * Creates and saves LeaveRequestDates for this LeaveRequest.
+   *
+   * If its an update and the dates has not changed when comparing
+   * the values in the db to the values about to be updated, i.e
+   * dateChanged = false, the previous dates are not deleted and
+   * re-created but left as is.
+   *
+   * @param bool|null $datesChanged
    */
-  private function saveDates() {
-    $this->deleteDatesAndBalanceChanges();
+  private function saveDates($datesChanged = null) {
+    if($datesChanged === false) {
+      return;
+    }
+
+    $this->deleteDates();
 
     $datePeriod = new BasicDatePeriod($this->from_date, $this->to_date);
 
@@ -768,10 +780,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   }
 
   /**
-   * Deletes all the dates and balance changes related to this LeaveRequest
+   * Deletes all the dates related to this LeaveRequest
    */
-  private function deleteDatesAndBalanceChanges() {
-    LeaveBalanceChange::deleteAllForLeaveRequest($this);
+  private function deleteDates() {
     LeaveRequestDate::deleteDatesForLeaveRequest($this->id);
   }
 
@@ -1073,5 +1084,29 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $period = AbsencePeriod::getPeriodContainingDates($fromDate, $toDate);
 
     return [$absenceType, $period];
+  }
+
+  /**
+   * Checks if the from_date or to_date of a leave request has changed by comparing the
+   * date values to be updated to the current values in the database
+   *
+   * @param array $params
+   *
+   * @return bool|null
+   *   Returns null for when a leave request is newly
+   *   created.
+   */
+  private static function datesChanged($params) {
+    if(!empty($params['id'])) {
+      $leaveRequest = self::findById($params['id']);
+      $fromDate = new DateTime($params['from_date']);
+      $toDate = new DateTime($params['to_date']);
+      $leaveRequestFromDate = new DateTime($leaveRequest->from_date);
+      $leaveRequestToDate = new DateTime($leaveRequest->to_date);
+
+      return $leaveRequestFromDate != $fromDate || $leaveRequestToDate != $toDate;
+    }
+
+    return null;
   }
 }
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -364,4 +367,126 @@ private function getDefaultParams($params = []) {
     ];
     return array_merge($defaultParams, $params);
   }
+
+  public function testBalanceChangeIsUpdatedForAlreadyCreatedLeaveRequestWhenBalanceHasChangedAndChangeBalanceParameterIsTrueAndLeaveRequestDatesDidNotChange() {
```
davialexandre: Wow, this is quite a long method name :D 
What about this slightly shorter version?

`testBalanceChangeIsUpdatedForAnExistingLeaveRequestWhenChangeBalanceParameterIsTrueAndDatesDidNotChange`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -364,4 +367,126 @@ private function getDefaultParams($params = []) {
...
+    $leaveRequest = $this->getleaveRequestService()->create(
+      $params,
+      false
+    );
+
+    $balance = LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest);
+    $this->assertEquals($newBalance, $balance);
+  }
+
+  public function testBalanceChangeIsNotUpdatedForAlreadyCreatedLeaveRequestWhenChangeBalanceParameterIsFalseAndLeaveRequestDatesDidNotChange() {
```
davialexandre: Perhaps

`testBalanceChangeIsNotUpdatedForAnExistingLeaveRequestWhenChangeBalanceParameterIsFalseAndDatesDidNotChange`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -768,10 +780,9 @@ private function saveDates() {
   }
 
   /**
-   * Deletes all the dates and balance changes related to this LeaveRequest
+   * Deletes all the dates related to this LeaveRequest
    */
-  private function deleteDatesAndBalanceChanges() {
-    LeaveBalanceChange::deleteAllForLeaveRequest($this);
+  private function deleteDates() {
     LeaveRequestDate::deleteDatesForLeaveRequest($this->id);
```
davialexandre: Can you explain why we're not deleting the balance changes here anymore? If we delete the dates but not their balance changes, we'll end up with a lot of orphan records in the database

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -364,4 +367,126 @@ private function getDefaultParams($params = []) {
...
+    $params['change_balance'] = 0;
+
+    $leaveRequest = $this->getleaveRequestService()->create(
+      $params,
+      false
+    );
+
+    $balance = LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest);
+    $this->assertEquals($previousBalance, $balance);
+  }
```
davialexandre: What about tests for:
- When the dates changed and change_balance is 0
- When the dates changed and change_balance is 1

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: I moved the function to https://github.com/civicrm/civihr/blob/ca0b8ee912f946e2186d8426469f6a56e864594b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChange.php#L15-L15.
Think of this scenario: The user modifies the status of an already created leave request and change_balance = 0, in the leave request service, we don't recalculate balance change but meanwhile during the leave request update in the BAO, the balance change would have been deleted here and the user ends up with Zero balance change for the request.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2873,4 +2873,36 @@ public function testUpdatingToilThrowsExceptionWhenUpdatingApprovedToilWithDates
...
+    ];
+    $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation($params);
+
+    $dates = $leaveRequest->getDates();
+    $beforeDatesID = [];
+    foreach($dates as $date) {
+      $beforeDatesID[] = $date->id;
+    }
+
+    //update leave request without changing the dates
```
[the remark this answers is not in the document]
tunbola: True that.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
davialexandre: In that scenario, the dates won't have changed, which means the `deleteDates` (or `deleteDatesAndBalanceChanges`) method won't be called and the balance changes shouldn't be touched as well.

Calling `LeaveBalanceChange::deleteAllForLeaveRequest` AFTER the leave request dates have been deleted will simply not delete anything, because the balance changes are linked to the request dates, not to the leave request itself. If you delete the dates first, you delete the link between the leave request and the balance changes and there's no way to know which balance changes belong to the request.

tunbola: You're right. Thanks. Will revert. Wondering what I was thinking. (shrug)

tunbola: Just thought of another scenario why I added it now. A scenario where the user modifies an already created leave request without changing the dates and decides to change the balance change i.e change_balance = 1. In this case, the dates and balance changes will not be deleted and a new balance change will be added to the already existing one causing an increase in balance change. For this case, calling `LeaveBalanceChange::deleteAllForLeaveRequest($leaveRequest);` will successfully delete the old leave requests balances because the dates will still exist and prevent the bug.

tunbola: Not sure what to do. Maybe I should revert to delete leave request dates and balance changes so we don't have orphan records in the case when the date changes in addition to adding the method in  https://github.com/civicrm/civihr/blob/ca0b8ee912f946e2186d8426469f6a56e864594b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveBalanceChange.php#L15-L15

davialexandre: Yeah, sounds like a good plan

PR #2102, opened 2017-08-30
PCHR-2536: Add unique leave request links in emails

igorpavlov:
## Overview

This PR utilises https://github.com/civicrm/civihr/pull/2051 and makes links in leave requests email notifications lead directly to Leave Request Modal with a correct leave request opened.

## Technical Details

Links are now populated with a proper GET query containing leave request ID.

```php
$queryString = '?leave-request-id=' . $this->leaveRequest->id;
$leaveUrl = CRM_Utils_System::url('my-leave#/my-leave/report' .
  $queryString, [], true);
```

## Comments

This PR also fixes the issue with showing balance. The problem was happening because `LeaveRequest.find()` method utilised `get` API3 method which lacked **balance_change** and **dates** properties to be returned. Now `getFull` method is used for the `.find()` method, which is also already used in `.all()` method.

---

- [x] Tests Pass

PR #2104, opened 2017-08-30
PCHR-2529: Improve Leave Calendar Pagination

igorpavlov:
## Overview

Current pagination looks like it paginates through dates, which is confusing. This PR moves pagination elements so they clearly represent a pagination through contacts instead.

## Before

ℹ️ The limit of contacts was **temporarily** changed from 20 to 2 in order to shoot GIFs. This PR does **not** change the limit.

### SSP
**Desktop**
sspd
**Mobile**
sspm

### Admin
**Desktop**
admind
**Mobile**
*Note: admin does not support mobile yet (viewport scale is not set yet), so the screencast was taken by simply shrinking the browser viewport to a reasonably thin size*
adminm

## Technical Details

1. The "previous" pagination item was just moved to the *Employee Name* header while "next" pagination item was transformed into a table - that allowed to re-use Bootstrap and write **less** custom CSS.

```html
<table class="table table-bordered chr_leave-calendar__pagination-next" ng-if="month.contactsList().length > month.pageSize">
  <thead>
  <tr>
    <th>
      <div class="row chr_leave-calendar__month-container">
        <div class="col-xs-4 col-md-3 col-lg-2 chr_leave-calendar__pagination">
          <button class="btn-link" ng-disabled="month.currentPage >= month.contactsList().length / month.pageSize - 1" ng-click="month.currentPage = (month.currentPage + 1)">
            <i class="fa fa-lg fa-chevron-down"></i>
          </button>
        </div>
      </div>
    </th>
  </tr>
  </thead>
</table>
```

## Comments

1. This PR fixes heights of table headers and the bottom margin of the Month component.
2. This PR also includes built JavaScript files forgotten to be built in the previous PR (https://github.com/civicrm/civihr/pull/2102).
3. There was an issue with BackstopJS, the property for the name of the viewport was wrong, it was changed from "name" to "label".

---

- [x] Tests Pass


mi-kon:
@jamienovick do you think we should centre the arrows within the cell rather than centre it under the column header?

igorpavlov:
@mi-kon @jamienovick Please see 3 alternatives:

a1
a2
a3

mi-kon:
Personally my choice is #2. Or #1 if we can make arrow and column header on the same line. 
In #3  arrows look more like sorting options than pagination. 

Waiting for @jamienovick 's feedback

igorpavlov:
@mi-kon let me try to make the header title and the arrow on the same line

igorpavlov:
@mi-kon @jamienovick Another alternative version:

image

mi-kon:
It does look nicer to me! my only concern is that arrows can still be confused with sorting options 
@jamienovick ?

jamienovick:
Whilst I prefer the look of option 3+4 I agree with Mike that its probably still possible to confuse with sort options. I think option 2 therefor which removes the column label is the best. Its a given that the list of names are staff names and this clearly will scroll that list so I think we should go with that one... @mi-kon ? Thanks Igor great job to show the options.

PR #2105, opened 2017-08-30
PCHR-2495: Add the LeavePeriodEntitlement.getLeaveBalances API

davialexandre:
## Overview
This PR adds the `LeavePeriodEntitlement.getLeaveBalances` API. This new API aggregates data that is already available via other APIs like `LeavePeriodEntitlement.getRemainder` and `LeavePeriodEntitlement.getEntitlement`, but it does this with a much better performance and reduced number of queries, especially because it can be used to get information for multiple contacts with a single request.

## Before

To get the same information provided by the new API, one would need a chained API call similar to this one:

```javascript
CRM.api3('Contact', 'getleavemanagees', {
  "sequential": 1,
  "managed_by": 205,
  "api.LeavePeriodEntitlement.getentitlement": {"contact_id":"$value.id","period_id":2},
  "api.LeavePeriodEntitlement.getremainder": {"contact_id":"$value.id","period_id":2,"include_future":1}
})
```

As you can see, we have three chained API in that request. Even worse, the two inner call are executed once for each result returned by the parent call to `Contact.getLeaveManagess`. In a scenario where `getLeaveManagees` would return 100 contacts, we would end up with 201 API calls, not to mention the internal SQL queries executed by each of these API calls.

Besides that, the format in which the data is returned isn't easy to loop through, as the chained API call results in a deeply nested data structure. Something like this:

```json
{
            "contact_id": "4",
            "display_name": "civihr_staff@compucorp.co.uk",
            "hrjobroles_id": "1",
            "id": "1",
            "contact_is_deleted": "0",
            "api.LeavePeriodEntitlement.getentitlement": {
                "is_error": 0,
                "version": 3,
                "count": 3,
                "values": [
                    {
                        "id": "7",
                        "entitlement": 28
                    },
                    {
                        "id": "8",
                        "entitlement": 0
                    },
                    {
                        "id": "9",
                        "entitlement": 0
                    }
                ]
            },
            "api.LeavePeriodEntitlement.getremainder": {
                "is_error": 0,
                "version": 3,
                "count": 3,
                "values": [
                    {
                        "id": "7",
                        "remainder": {
                            "current": 14.5,
                            "future": 14.5
                        }
                    },
                    {
                        "id": "8",
                        "remainder": {
                            "current": 0,
                            "future": 0
                        }
                    },
                    {
                        "id": "9",
                        "remainder": {
                            "current": 0,
                            "future": 0
                        }
                    }
                ]
            }
        },
```

## After

The new API requires a single API call. Here is an example of how to do the same thing as the previous example of a chained call:

```javascript
CRM.api3('LeavePeriodEntitlement', 'getleavebalances', {
  "sequential": 1,
  "period_id": 2,
  "managed_by": 205
})
```

The API response is also simpler and easier to loop through:

```json
{
    "is_error": 0,
    "version": 3,
    "count": 4,
    "values": {
        "4": {
            "contact_id": "4",
            "contact_display_name": "civihr_staff@compucorp.co.uk",
            "absence_types": [
                {
                    "id": 1,
                    "entitlement": "28.00",
                    "used": 13.5,
                    "balance": 14.5,
                    "requested": 0
                },
                {
                    "id": 2,
                    "entitlement": "0.00",
                    "used": 0,
                    "balance": 0,
                    "requested": 0
                },
                {
                    "id": 3,
                    "entitlement": "0.00",
                    "used": 0,
                    "balance": 0,
                    "requested": 0
                }
            ]
        }
   }
}
```
A few things to note:
- All the balances are aggregate together inside the `absence_types` property, instead of separatedely in each chain response.
- We now have a `used` property for each absence type. On a chained call, we would have to obtain that value manually, by subtracting the balance/remainder from the entitlement.

This API also supports the `is_count` option, which will return the number of contacts matching the given criteria. This is how it can be used:

```javascript
CRM.api3('LeavePeriodEntitlement', 'getleavebalances', {
  "sequential": 1,
  "period_id": 2,
  "options": {"is_count":1}
})
```
And this is how its response looks like:

```json
{
    "is_error": 0,
    "result": 5
}
```
Result is the number of contacts that would be returned on a non is count query.

## Technical Details
Three new methods were created to make it possible for the API get the data with the minimum number of SQL queries as possible. The methods are:
- `LeavePeriodEntitlement::getEntitlementsForContacts`
- `LeaveBalanceChange::getBalanceForContacts`
- `LeaveBalanceChange::getOpenLeaveRequestBalanceForContacts`

The most important thing about them is that they all accept a list of Contact IDs and return the entitlement/balance for all the received contacts with a single Query. Those classes already have methods to get the same information, but they were built to get it for a single contact.

After the new methods were created, I tried to refactor the existing methods to use the new ones, but it turned out to be a too complex task and I gave up. The main difficulty is that the existing methods work based on a LeavePeriodEntitlement instance instead of a contact ID. Another difficulty is that some of the existing methods (`LeaveBalanceChange::getBalanceForEntitlement` for example) have a lot of filters (like filter by status, by public holidays, by future balance etc) that were really not necessary for this API. In order to use the new methods to do the work of the old ones, I would have to implement all of those filters in them, which would increase the scope of the PR.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
@@ -1,0 +1,91 @@
+<?php
+
+use CRM_Contact_BAO_Contact as Contact;
+use CRM_Contact_BAO_Relationship as Relationship;
+use CRM_Contact_BAO_RelationshipType as RelationshipType;
+use CRM_Hrjobcontract_BAO_HRJobContract as HRJobContract;
+use CRM_Hrjobcontract_BAO_HRJobDetails as HRJobDetails;
+use CRM_Hrjobcontract_BAO_HRJobContractRevision as HRJobContractRevision;
+use CRM_HRLeaveAndAbsences_API_Query_Select as SelectQuery;
+
+class CRM_HRLeaveAndAbsences_API_Query_LeaveBalancesSelect {
+
+  use CRM_HRLeaveAndAbsences_Service_SettingsManagerTrait;
+
+  public function __construct($params) {
+    $this->params = $params;
+    $this->buildCustomQuery();
+  }
+
+  private function buildCustomQuery() {
+    $customQuery = CRM_Utils_SQL_Select::from(Contact::getTableName() . ' as a');
+
+    $this->addJoins($customQuery);
+    $this->addWhere($customQuery);
+
+    $this->query = new SelectQuery('Contact', $this->params);
+    $this->query->merge($customQuery);
+  }
+
+  private function addJoins(CRM_Utils_SQL_Select $query) {
+    $joins = [
+      'INNER JOIN ' . HRJobContract::getTableName() . ' jc ON a.id = jc.contact_id',
+      'INNER JOIN ' . HRJobContractRevision::getTableName() . ' jcr ON jcr.id = (SELECT id
+                    FROM ' . HRJobContractRevision::getTableName() . ' jcr2
+                    WHERE
+                    jcr2.jobcontract_id = jc.id
+                    ORDER BY jcr2.effective_date DESC
+                    LIMIT 1)',
+      'INNER JOIN ' . HRJobDetails::getTableName() . ' jd ON jd.jobcontract_revision_id = jcr.details_revision_id'
+    ];
+
+    if(!empty($this->params['managed_by'])) {
+      $joins[] = 'LEFT JOIN ' . Relationship::getTableName() . ' r ON r.contact_id_a = a.id';
+      $joins[] = 'LEFT JOIN ' . RelationshipType::getTableName() . ' rt ON rt.id = r.relationship_type_id';
+    }
+
+    $query->join(null, $joins);
+  }
+
+  private function addWhere($customQuery) {
+    $conditions = [
+      'a.is_deleted = 0',
+      'jc.deleted = 0',
+    ];
+
+    if(!empty($this->params['managed_by'])) {
+      $managerID = (int)$this->params['managed_by'];
+
+      if(!isset($this->params['unassigned'])) {
+        $activeLeaveManagerCondition = $this->hasActiveLeaveManagerCondition();
+        $activeLeaveManagerCondition[] = "r.contact_id_b = {$managerID}";
+        $conditions = array_merge($conditions, $activeLeaveManagerCondition);
+      }
+    }
+
+
+    $customQuery->where($conditions);
+  }
+
+  public function run() {
+    $results = $this->query->run();
+
+    return $results;
+  }
+
+
+  private function hasActiveLeaveManagerCondition() {
+    $today =  '"' . date('Y-m-d') . '"';
+    $leaveApproverRelationshipTypes = $this->getLeaveApproverRelationshipsTypesForWhereIn();
+
+    $conditions = [];
+    $conditions[] = 'rt.is_active = 1';
+    $conditions[] = 'rt.id IN(' . implode(',', $leaveApproverRelationshipTypes) . ')';
+    $conditions[] = 'r.is_active = 1';
+    $conditions[] = "(r.start_date IS NULL OR r.start_date <= {$today})";
+    $conditions[] = "(r.end_date IS NULL OR r.end_date >= {$today})";
+
+    return $conditions;
+  }
+
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
@@ -86,7 +86,7 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveBalancesSelect {
       'INNER JOIN ' . HRJobDetails::getTableName() . ' jd ON jd.jobcontract_revision_id = jcr.details_revision_id'
     ];
 
-    if($this->getManagerID()) {
+    if($this->shouldFilterByManager()) {
       $joins[] = 'LEFT JOIN ' . Relationship::getTableName() . ' r ON r.contact_id_a = a.id';
       $joins[] = 'LEFT JOIN ' . RelationshipType::getTableName() . ' rt ON rt.id = r.relationship_type_id';
     }
@@ -115,12 +115,8 @@ class CRM_HRLeaveAndAbsences_API_Query_LeaveBalancesSelect {
       )"
     ];
 
-    $managerID = $this->getManagerID();
-    if($managerID) {
-      $activeLeaveManagerCondition = $this->hasActiveLeaveManagerCondition();
-      $activeLeaveManagerCondition[] = "r.contact_id_b = {$managerID}";
-      $conditions = array_merge($conditions, $activeLeaveManagerCondition);
-    }
+    $activeLeaveManagerCondition = $this->hasActiveLeaveManagerCondition();
+    $conditions = array_merge($conditions, $activeLeaveManagerCondition);
 
     $customQuery->where($conditions);
   }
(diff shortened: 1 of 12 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
```diff
@@ -0,0 +1,293 @@
...
+
+    $conditions = [
+      'a.is_deleted = 0',
+      'jc.deleted = 0',
+      "(
+        (jd.period_end_date IS NOT NULL AND jd.period_start_date <= '{$absencePeriod->end_date}' AND jd.period_end_date >= '{$absencePeriod->start_date}')
+          OR
+        (jd.period_end_date IS NULL AND 
+          (
+            (jd.period_start_date >= '{$absencePeriod->start_date}' AND jd.period_start_date <= '{$absencePeriod->end_date}')
```
tunbola: Do we really need this condition? Once the jd.period_end_date IS NULL, I'm thinking the only criteria left to be fulfilled is that the jd.period_start_date is less than the Absence period end date.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -241,4 +241,116 @@ function civicrm_api3_leave_period_entitlement_getentitlement($params) {
...
+ * @return void
+ */
+function _civicrm_api3_leave_period_entitlement_getleavebalances_spec(&$spec) {
+  $spec['managed_by'] = [
+    'name' => 'managed_by',
+    'title' => 'Managed By',
+    'description' => 'Include only Leave Balances for contacts managed by the contact with the given ID',
+    'type' => CRM_Utils_Type::T_INT,
+    'api.required' => 0,
+    'FKClassName'  => 'CRM_Contact_DAO_Contact',
```
tunbola: Please adjust appropriately space between operand and operator in this array.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -241,4 +241,116 @@ function civicrm_api3_leave_period_entitlement_getentitlement($params) {
...
+  ];
+
+  $spec['period_id'] = [
+    'name' => 'period_id',
+    'title' => 'Absence Period ID',
+    'description' => 'Include only Balances from Leave Requests taken during the Period with the given ID',
+    'type' => CRM_Utils_Type::T_INT,
+    'api.required' => 1,
+    'FKClassName'  => 'CRM_HRLeaveAndAbsences_BAO_AbsencePeriod',
+    'FKApiName'    => 'AbsencePeriod',
```
tunbola: adjust space between operand and operator.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -241,4 +241,116 @@ function civicrm_api3_leave_period_entitlement_getentitlement($params) {
...
+  ];
+
+  $spec['contact_id'] = [
+    'name' => 'contact_id',
+    'title' => 'Contact ID',
+    'description' => 'Include only Leave Balances for contacts with the given ID',
+    'type' => CRM_Utils_Type::T_INT,
+    'api.required' => 0,
+    'FKClassName'  => 'CRM_Contact_DAO_Contact',
+    'FKApiName'    => 'Contact',
```
tunbola: adjust space between operand and operator.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlement.php
```diff
@@ -241,4 +241,116 @@ function civicrm_api3_leave_period_entitlement_getentitlement($params) {
...
+  ];
+
+  $spec['type_id'] = [
+    'name' => 'type_id',
+    'type' => CRM_Utils_Type::T_INT,
+    'title' => 'Absence Type ID',
+    'description' => 'Include only Balances for the Given Absence Type',
+    'api.required' => 0,
+    'FKClassName'  => 'CRM_HRLeaveAndAbsences_BAO_AbsenceType',
+    'FKApiName'    => 'AbsenceType',
```
tunbola: adjust space between operand and operator.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2873,4 +2873,48 @@ public function testUpdatingToilThrowsExceptionWhenUpdatingApprovedToilWithDates
     $params['to_date'] = CRM_Utils_Date::processDate('2016-01-08');
     LeaveRequestFabricator::fabricate($params, true);
   }
+
+  public function testCanReturnTheListOfAllStatuses() {
+    $leaveStatuses = array_flip(LeaveRequest::buildOptions('status_id', 'validate'));
+
+    $statuses = LeaveRequest::getStatuses();
+
```
tunbola: Why not `$this->assertEquals($leaveStatuses, $status)` ?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/LeaveBalancesSelect.php
```diff
@@ -227,15 +223,28 @@ private function getLeaveBalances($contacts) {
...
+      $conditions[] = 'r.is_active = 1';
+      $conditions[] = "(r.start_date IS NULL OR r.start_date <= {$today})";
+      $conditions[] = "(r.end_date IS NULL OR r.end_date >= {$today})";
+
+      if(!$this->leaveManagerService->currentUserIsAdmin()) {
+        $managerID = (int)CRM_Core_Session::getLoggedInContactID();
+        $conditions[] = "r.contact_id_b = {$managerID}";
+      }
+
+      if(!empty($this->params['managed_by'])) {
```
tunbola: Thinking  of a scenario where the current user is a manager and managed_by is present which has a value other that the current logged in user ID. Looks like both IF conditions will be met and we have ` r.contact_id_b = {$currentUserID} AND r.contact_id_b = {$managerID}"`, although no results will be returned. 
Wondering why the previous if/else condition is not used again so as to avoid having this condition `$conditions[] = "r.contact_id_b = {$managerID}";` twice in such scenario?

davialexandre: The previous version of this had a single `r.contact_id_b = {$managerID}` where the $managerID was given by:

```php
$managerID = null;
if (!currentUserIsAdmin()) {
  $managerID = getLoggedInContactID();
} elseif(!empty($params['managed_by'])) {
  $managerID = $params['managed_by'];
}
```

With that, this is what would happen:
- An admin tries to use the API without passing a manager ID: No join with the relationship tables and all the contacts could be returned
- An admin tries to use the API passing a manager ID: A join with the relationship tables was added and the value passed to the managed_by param was used on the r.contact_id_b condition, meaning that only contacts managed by the given manager would be returned.
- A manager tries to use the API without passing a manager ID: A join with the relationship tables was added and the ID of the current logged in contact would be used as the managerID, meaning that only contacts managed by the current user would be returned
- A manager tries to use the API passing a manager_id of another manager: A join with the relationship tables was added, but the given managed_by value is discarded, as we always use the ID of the current user when they are not admins. This means that the managees of the current user would be returned

On the last scenario above, the correct behavior would be an empty response instead of a list of contacts managed by the current user. That's why we need two separate conditions.

tunbola: Got it.

PR #2107, opened 2017-08-31
PCHR 2543: Migrate admin screens to bootstrap

reneolivo:
## Overview
This PR updates the Absece Types, Absence Periods, Public Holidays, and General Settings screens to use Bootstrap markup.

This PR needs [https://github.com/compucorp/org.civicrm.shoreditch/pull/31](https://github.com/compucorp/org.civicrm.shoreditch/pull/31) before merging.

### Before
anim
### After
anim
Fixed size inputs have a proper width.

### Before
anim
### After
anim

### Before
anim
### After
anim

### Before
anim
### After
anim

## Technical Details
* **leaveandabsence.css** style was added to the screens in order to override some of the styles:
```PHP
```
**hrleaveandabsences.css** was not used because it's a plain CSS file and the SaSS @extend method was needed in order to properly style buttons and use theme color variables.
* Indentation was fixed for some files. Were using 4 spaces, changed them to 2.
* BackstopJS scenarios were added for all screens. Since there was no need for user input, no scripts were added.

## Comments
* Viewport scale for the admin is not defined.

---

- [x] Tests Pass


igorpavlov:
Please fix the title of the PR: 

> PCHR-2543: Migrate admin screens to bootstrap

igorpavlov:
ℹ️ Please note, the following suggestions should be confirmed at first.

## Absence Types

1. The bar on the top looks like **info** for me rather than a **warning**. Wouldn't it be better to make it blue instead of yellow?
2. The selector looks a bit broken where the arrow is.

## Absence Periods + Public Holidays

1. The chevron icon on buttons is a bit off, shouldn't it be vertically centered?

## ALL
1. The right padding from the main block is bigger than left one. Should be equal?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_buttons.scss
```diff
@@ -0,0 +1,14 @@
+.crm-button .crm-i {
+  top: .6em;
+  left: .6em;
+}
+
+.crm-button-type-next {
+  @include button-outline-variant(white, $brand-primary);
+  .crm-i { display: none; }
```
igorpavlov: Please add line-break before `.crm-i`

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_buttons.scss
```diff
@@ -0,0 +1,14 @@
+.crm-button .crm-i {
+  top: .6em;
+  left: .6em;
+}
+
+.crm-button-type-next {
+  @include button-outline-variant(white, $brand-primary);
```
igorpavlov: Please add a line-break before `@include`

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_buttons.scss
```diff
@@ -0,0 +1,14 @@
...
+  .crm-i { display: none; }
+}
+
+.crm-button-type-cancel {
+  border-color: $gray-light;
+  input { color: $text-color !important; }
```
igorpavlov: Please add a line break before `input`

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_radio-buttons.scss
```diff
@@ -0,0 +1,11 @@
+.crm-form-radio + label {
+  &:before {
```
igorpavlov: Please add a line break before `&:before`

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_buttons.scss
igorpavlov: Please break this into 3 lines. I think CSS is easier to read when everything is consistent. JS might be another story though.

```css
input {
  color: $text-color !important;
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_inputs.scss
```diff
@@ -0,0 +1,5 @@
+// removes fixed with for inputs added by CiviCRM:
```
igorpavlov: Did you mean `width`?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_inputs.scss
```diff
@@ -0,0 +1,5 @@
...
+input, textarea, select {
+  width: auto !important;
```
igorpavlov: How can we make sure this doesn't break anything in other places?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_buttons.scss
reneolivo: before @include ? like this:

```scss
.crm-button-type-next {

  @include button-outline-variant(white, $brand-primary);
}
```

igorpavlov: Yep. No line-break needed if you list styles, but if you include smth or nest, then a line-break is needed. Like here:

image

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_inputs.scss
reneolivo: These inputs for CiviCRM are always 230px in length. This breaks two things:
1. inputs can't adapt to their parent width.
2. can't have smaller inputs, all of them are 230px wide.

Why this code shouldn't affect other places:

This is wrapped inside of the bootstrap theme id like this:
```scss
#bootstrap-theme input, #bootstrap-theme textarea, #bootstrap-theme select {
  width: auto !important;
}
```
So that means it won't affect anything outside bootstrap.

`width: auto` is the default behaviour for bootstrap inputs: 
https://github.com/twbs/bootstrap-sass/blob/5d6b2ebba0c2a5885ce2f0e01e9218db3d3b5e47/assets/stylesheets/bootstrap/_forms.scss#L503

If an input needs to be of a specific length the size attr can be used. 

If an input needs to be of a specific pixel width then yes, this might affect those types of inputs. I did a quick look around the app and nothing seems to be breaking.

If this is a concern we could wrap this for the admin by adding an extra class.

igorpavlov: Awesome. Thank you for the explanation.

reneolivo:
### New changes:

Alerts are set to notifications instead of warnings:

screenshot 2017-08-31 12 07 19

select2 dropdowns have the right width and padding:
screenshot 2017-08-31 12 18 53

Absence Periods, Absence Types, and Public Holidays "Add" button has the chevron icon aligned correctly:
screenshot 2017-08-31 12 22 14

> The right padding from the main block is bigger than left one. Should be equal?

They have the same width, it looks weird because of the way I captured the screen.... sorry.
screenshot 2017-08-31 12 28 58

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_inputs.scss
igorpavlov: To document the decision: it was decided that this is the only way to go since CRM sets 230px !important.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_select2.scss
```diff
@@ -1,4 +1,10 @@
 // removes extra padding on select2 ul element added by bootstrap:
-.select2-container ul.select2-choices {
+.select2-container.crm-select2 ul.select2-choices {
   padding-left: 0;
 }
+
+// This fixes the arrow dropdown width by removing 4 padding and and adding 8 width:
+.select2-container.crm-select2 .select2-choice .select2-arrow {
+  padding: 0;
```
igorpavlov: Do you need to flush all padding or may you just flush just padding-right?

igorpavlov: Do you need to flush all padding or may you just flush just padding-right?

igorpavlov:
Could you please take a screenshot of a fixed chevron in buttons?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/overrides/style/_select2.scss
reneolivo: All paddings need to be removed. If I remove padding-right it will look as if it were aligned to the right.

reneolivo:
@igorpavlov I had added a screenshot:

before:
screenshot 2017-08-31 12 46 01


After
https://user-images.githubusercontent.com/1642119/29933979-fcacf1d6-8e25-11e7-8e4a-5e6ea142d8ec.png

PR #2110, opened 2017-09-01
PCHR-2553: Fix editing own request by manager

igorpavlov:
## Overview

Currently managers cannot edit their own requests. This PR fixes this issue.

## Technical Details

*File js/angular/src/leave-absences/shared/controllers/request.controller.js*

The issue was in the `setRoles()` function - it ignored the `vm.isSelfRecord` value. `isSelfRecord` is essential for **My Leave** page since on this page all users with any permissions make/edit leave requests for **themselves**.

This means simply adding a force check on the `isSelfRecord` solves the issue:
```js
function initRoles () {
  role = 'staff';

  if (vm.isSelfRecord) {
    return; // <-- force the role to be "staff" in case of a self record
  }

  // Then if not a self record, set roles basing on permissions.
  // (ex. on Manager Leave page the functionality is different to Managers and Admins)
  return checkPermissions(sharedSettings.permissions.admin.administer).then...

}
```

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -498,16 +498,20 @@ define([
     function initRoles () {
       role = 'staff';
 
+      if (vm.isSelfRecord) {
```
deb1990: We should not really change the `role` based on `isSelfRecord`. Because they are mutually exclusive, in future we might have a requirement where the my leave page needs to have different functionality based on the role(staff or manager). So we should keep the role as it is, but add conditions based on `isSelfRecord` to achieve the same.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request.controller.spec.js
```diff
@@ -828,6 +828,38 @@
         });
       });
 
+      describe('user edits their own leave request popup', function () {
+        var leaveRequest;
+
+        ['staff', 'manager', 'admin'].forEach(function (permissionsRole) {
+          testRoleForSelfRecord(permissionsRole);
+        })
```
deb1990: Semicolon missing? Is the linter working for you?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
igorpavlov: Sure. I am following the similar logic as per Leave Calendar:

```js
bindings: {
  contactId: '<',
  roleOverride: '@?'
}
```
```js
function setUserRole () {
  if (vm.roleOverride) {
    return $q.resolve().then(function () {
      userRole = vm.roleOverride;
    });
  } else {
    return $q.all([
      checkPermissions(sharedSettings.permissions.admin.administer),
      checkPermissions(sharedSettings.permissions.ssp.manage)
    ])
    .then(function (results) {
      userRole = results[0] ? 'admin' : (results[1] ? 'manager' : 'staff');
    });
  }
}
```
And here is the config file for the **my-leave** module:
```js
.state('my-leave.calendar', {
  url: '/calendar',
  template: '<leave-calendar contact-id="myleave.contactId" role-override="staff"></leave-calendar>'
});
```
I actually wanted initially to override the role basing on the `$rootScope.section`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -498,7 +498,7 @@ define([
     function initRoles () {
       role = 'staff';
 
-      if (vm.isSelfRecord) {
+      if ($rootScope.section === 'my-leave') {
```
reneolivo: is it possible to see my leave requests in a different section? can they be seen on manager leave? contact page?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request.controller.spec.js
```diff
@@ -376,7 +376,7 @@
 
                 leaveRequest.contact_id = CRM.vars.leaveAndAbsences.contactId.toString();
 
-                initTestController({ isSelfRecord: true, leaveRequest: leaveRequest });
+                initTestController({ leaveRequest: leaveRequest, isSelfRecord: true });
```
reneolivo: why did you remove `isSelfRecord: true` from other tests but left this one?

igorpavlov: The implementation of `isSelfRecord` was changed and now is used only in one place in the code:

```js
function changeStatusBeforeSave () {
  if (vm.isSelfRecord) {
    vm.request.status_id = vm.requestStatuses[sharedSettings.statusNames.awaitingApproval].value;
  } else if (vm.canManage) {
    vm.request.status_id = vm.newStatusOnSave || vm.request.status_id;
  }
}
```
and exactly this place is tested: 

```js
it('status changes to waiting approval before calling API', function () {
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
igorpavlov: Yes, this component is used in Absence Tab:

image

PR #2113, opened 2017-09-04
PCHR-2556: Contact.getLeaveManagees API to support the unassigned parameter.

tunbola:
## Overview
Currently the Contact.getLeaveManagees API only supports one mandatory parameter: `managed_by` which allows the API to returned contacts managed by the contact ID passed in the managed_by parameter. To make the API more flexible, the unassigned parameter is introduced in this PR to allow the API to return contacts that do not have an active Leave Approver relationship.

## Before
The Contact.getLeaveManageesAPI does not support the unassigned parameter.

## After
The Contact.getLeaveManageesAPI now supports the unassigned parameter. 
These are the rules that guide the new unassigned parameter in relation to the existing managed_by parameter:
- If the `unassigned` parameter isn't present in the request, then `managed_by` must be required.
- If `unassigned` is false, then `managed_by` must be required.
- If `unassigned` is true, then `managed_by` must not be passed.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/ContactSelect.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/ContactSelect.php
@@ -51,6 +51,7 @@ class CRM_HRLeaveAndAbsences_API_Query_ContactSelect {
 
     $this->addJoins($customQuery);
     $this->addWhere($customQuery);
+    $this->addGroupBy($customQuery);
     $this->filterReturnFields();
 
     $this->query = new SelectQuery('Contact', $this->params);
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -190,4 +187,78 @@ public function testGetLeaveManageesReturnsEmptyResultsWhenStaffMemberIsTryingTo
...
+    $this->callAPI('Contact', 'getleavemanagees', []);
+  }
+
+  public function testGetLeaveManageesReturnsOnlyContactsWithoutActiveLeaveApproverRelationshipWhenUnassignedTrue() {
+    $contact1 = $this->contact1;
+    $contact2 = $this->contact2;
+    $contact3 = ContactFabricator::fabricate();
+    $manager = $this->manager;
+
+    $relationshipType = RelationshipTypeFabricator::fabricate();
```
davialexandre: Please add a comment here noting why $contact3 won't be returned, even though this contact has a relationship with a manager

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -190,4 +187,78 @@ public function testGetLeaveManageesReturnsEmptyResultsWhenStaffMemberIsTryingTo
...
+    $this->assertEquals(3, $result['count']);
+    $this->assertEquals($result['values'][$contact2['id']]['id'], $contact2['id']);
+    $this->assertEquals($result['values'][$contact3['id']]['id'], $contact3['id']);
+    $this->assertEquals($result['values'][$manager['id']]['id'], $manager['id']);
+  }
+
+  public function testGetLeaveManageesReturnsNoInformationForContactWithActiveLeaveManagerAndOtherRelationshipWhenUnassignedIsTrue() {
+    $contact1 = $this->contact1;
+    $contact2 = $this->contact2;
+    $manager = $this->manager;
```
davialexandre: I think we don't need these local copies here too

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -190,4 +187,78 @@ public function testGetLeaveManageesReturnsEmptyResultsWhenStaffMemberIsTryingTo
...
+    RelationshipFabricator::fabricate([
+      'contact_id_a' => $contact1['id'],
+      'contact_id_b' => $manager['id'],
+      'relationship_type_id' => $relationshipType['id']
+    ]);
+
+    $result = $this->callAPI('Contact', 'getleavemanagees', ['unassigned' => true]);
+
+    //contact2 and and manager will be returned because they are without active leave
+    //approver relationdip.
```
davialexandre: s/relationdip/relationship/

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -190,4 +187,78 @@ public function testGetLeaveManageesReturnsEmptyResultsWhenStaffMemberIsTryingTo
...
+    //type leave approver.
+    RelationshipFabricator::fabricate([
+      'contact_id_a' => $contact1['id'],
+      'contact_id_b' => $manager['id'],
+      'relationship_type_id' => $relationshipType['id']
+    ]);
+
+    $result = $this->callAPI('Contact', 'getleavemanagees', ['unassigned' => true]);
+
+    //contact2 and and manager will be returned because they are without active leave
```
davialexandre: s/and and/and/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/ContactSelect.php
```diff
@@ -126,5 +150,25 @@ private function filterReturnFields() {
...
+
+    $conditions = [];
+    $conditions[] = 'rt.is_active = 1';
+    $conditions[] = 'rt.id IN(' . implode(',', $leaveApproverRelationshipTypes) . ')';
+    $conditions[] = 'r.is_active = 1';
+    $conditions[] = "(r.start_date IS NULL OR r.start_date <= '$today')";
+    $conditions[] = "(r.end_date IS NULL OR r.end_date >= '$today')";
+
+    return $conditions;
+  }
```
davialexandre: With this PR, this will be the third class with this same method (the others are `LeaveBalanceSelect` and `LeaveRequestSelect`). Perhaps we should see if we can refactor and extract this logic to somewhere more reusable?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -190,4 +187,78 @@ public function testGetLeaveManageesReturnsEmptyResultsWhenStaffMemberIsTryingTo
...
+    //approver relationdip.
+    $this->assertEquals(2, $result['count']);
+    $this->assertEquals($result['values'][$contact2['id']]['id'], $contact2['id']);
+    $this->assertEquals($result['values'][$manager['id']]['id'], $manager['id']);
+  }
 }
```
davialexandre: What about adding tests for when you're logged in as admin and using unassigned = true?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/ContactTest.php
```diff
@@ -23,6 +26,8 @@ class api_v3_ContactTest extends BaseHeadlessTest {
 
   public function setUp() {
     CRM_Core_DAO::executeQuery('SET foreign_key_checks = 0;');
+    $tableName = Contact::getTableName();
+    CRM_Core_DAO::executeQuery("DELETE FROM {$tableName}");
```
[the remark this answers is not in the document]
tunbola: Well, I had some weird issues with the test results, turned out my test db had some default contacts already in the database that counted among the unassigned contacts.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Query/ContactSelect.php
tunbola: What about a `RelationshipHelper` trait having a 'getLeaveApproverCondition` method?

davialexandre: It would solve the specific problem of having this method duplicated between these three classes, but we still have duplicated joins on these queries. For me, it seems that all of these are queries that support the `managed_by` param. That is, queries that return information of something managed by someone. I wonder if it would make sense to have a base `ManagedBySelect` class to centralize this logic and then these other 3 would extend it. Do you think it makes sense?

tunbola: Sounds good but not sure how complicated it's going to be and if its worth the time.

PR #2114, opened 2017-09-04
PCHR-2542: Add the type field to LeaveRequestDate

davialexandre:
## Overview
When we calculate the balance changes for a LeaveRequest, the `LeaveRequest::calculateBalanceChange()` method returns the number of days to be deducted for each date of the request together with the date type (half day, non-working day, weekend, public holiday etc). However,  this information isn't stored anywhere, making it impossible, once the LeaveRequest has been saved, to retrieve what were the date types at the moment when they were calculated.

## Before
The date type returned by the `LeaveRequest::calculateBalanceChange()` wasn't stored in the database.

## After
A new `type` field was added to the `LeaveRequestDate` entity. It's optional, and its value is null at the moment the Leave Request Dates are created. On a subsequent step in the process of saving a Leave Request, when we calculate and save its balance changes,  we get the types from the balance change calculation and update the respective date with it.

## Technical Details
- To add the new field, the DAOs had to be updated. However, the processes of updating it ended up causing changes to many other DAOs. The reason is that with the version of CiviCRM we're using at the moment, DAOs now have a new `indices` method and GenCode.php added it to other DAOs.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1004.php
```diff
@@ -0,0 +1,24 @@
...
+   * Adds the type field to the Leave Request Date table
+   *
+   * @return bool
+   */
+  public function upgrade_1004() {
+    $leaveRequestDateTable = LeaveRequestDate::getTableName();
+    if(!SchemaHandler::checkIfFieldExists($leaveRequestDateTable, 'type')) {
+      CRM_Core_DAO::executeQuery("
+        ALTER TABLE {$leaveRequestDateTable} 
+        ADD type VARCHAR(522) NULL
```
tunbola: The `VARCHAR` length here should be `512` I think.

PR #2120, opened 2017-09-06
PCHR-2544: Add the LeaveRequest.getBreakdown API

davialexandre:
## Overview
When opening an existing Leave Request, we should show its breakdown (the list of each of its dates, together with a number of days deducted for each of them). 

This is information is stored in the database, but scattered in a few tables. This new API should be used to expose an easy way to fetch it while avoiding to expose too many data structure details.

## Before
There's no easy way to get the breakdown of a Leave Request

## After
The `LeaveRequest.getBreakdown` is available to be used to fetch the breakdown of a Leave Request.
It has a mandatory `leave_request_id` param, that should be used to give the ID of the Leave Request to which we want to get the breakdown to.

The breakdown of a Leave Request is just a list of its dates together with the number of days deduct for it. Here is a sample response:

```json
{
    "is_error": 0,
    "version": 3,
    "count": 8,
    "values": [
        {
            "id": "99",
            "type": "1",
            "label": "All Day",
            "date": "2017-06-09",
            "amount": "-1.00"
        },
        {
            "id": "100",
            "type": "4",
            "label": "Weekend",
            "date": "2017-06-10",
            "amount": "0.00"
        },
        {
            "id": "101",
            "type": "4",
            "label": "Weekend",
            "date": "2017-06-11",
            "amount": "0.00"
        },
        {
            "id": "102",
            "type": "1",
            "label": "All Day",
            "date": "2017-06-12",
            "amount": "-1.00"
        }
    ]
}

```

The response will be empty either when trying to get the breakdown of a non-existing Leave Request or of a Leave Request to which you don't have access to (for example, a staff member trying to access the request of another staff member or a manager trying to access the request of another manager managee).

## Technical Details
It's possible to get all the information for the breakdown with a single SQL, but I opted to not do that. The reason is that the information comes from 2 different tables, so it would be hard to decide to each of the BAOs this query should belong to, so I decided to do this in 2 separate steps. First we get the dates of a Leave Request, then next we pass those dates to the new `LeaveBalanceChange::getForLeaveRequestDates()` method, which will return the Balance Changes linked to the given dates. Finally, inside the `LeaveRequestService.getBreakdown()` we merge the all the data and create the breakdown array.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -1266,4 +1266,31 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
 
     return $balances;
   }
+
+  /**
+   * Returns a list of LeaveBalanceChange instances linked to the given
+   * LeaveRequestDate instances.
+   *
+   * The results are indexed by the LeaveRequestDates ID.
+   *
+   * @param CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate[]
+   *
+   * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
+   */
+  public static function getForLeaveRequestDates($dates) {
+    $balanceChanges = [];
+
+    $datesIDs = CRM_Utils_Array::collect('id', $dates);
+
+    $balanceChange = new self();
+    $balanceChange->source_type = self::SOURCE_LEAVE_REQUEST_DAY;
+    $balanceChange->whereAdd('source_id IN (' . implode(',', $datesIDs) . ')');
+    $balanceChange->find();
+
+    while($balanceChange->fetch()) {
+      $balanceChanges[$balanceChange->source_id] = clone $balanceChange;
+    }
+
+    return $balanceChanges;
+  }
 }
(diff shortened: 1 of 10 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -1266,4 +1266,31 @@ public static function getOpenLeaveRequestBalanceForContacts($contactIDs, $absen
...
+   * The results are indexed by the LeaveRequestDates ID.
+   *
+   * @param CRM_HRLeaveAndAbsences_BAO_LeaveRequestDate[]
+   *
+   * @return CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange[]
+   */
+  public static function getForLeaveRequestDates($dates) {
+    $balanceChanges = [];
+
+    $datesIDs = CRM_Utils_Array::collect('id', $dates);
```
tunbola: Never knew Civi had this function!

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -3749,6 +3749,162 @@ public function testDeleteAttachmentShouldThrowAnExceptionIfAttachmentIDIsMissin
...
+    $this->setPermissions(['access AJAX API']);
+
+    $result = civicrm_api3('LeaveRequest', 'getBreakdown', [
+      'leave_request_id' => $leaveRequest->id,
+      'check_permissions' => true,
+    ]);
+    $this->assertCount(3, $result['values']);
+  }
+
+  public function testGetBreakdownShouldNotReturnIfAnAdminTriesToAccessTheBreakdownOfAnyLeaveRequest() {
```
tunbola: This should probably be `testGetBreakdownShouldNotReturnEmptyIfAnAdminTriesToAccessTheBreakdownOfAnyLeaveRequest` or `testGetBreakdownReturnsResultsIfAnAdminTriesToAccessTheBreakdownOfAnyLeaveRequest`

tunbola:
In PR description s/LeaveRequestServive.getBreakdown()/LeaveRequestService.getBreakdown()

PR #2124, opened 2017-09-08
PCHR-2546: Migrate Leave and Absences' import screen to Bootstrap.

reneolivo:
## Overview
This PR migrates the Leave and Absences' import screen to Bootstrap.

This PR is linked to https://github.com/compucorp/org.civicrm.shoreditch/pull/33

### Step 1
screen1

### Step 2
screen2

### Step 3
screen3

### Step 4
screen4

### Step 1
screen1

### Step 2

screen2

### Step 3
screen3

### Step 4
screen4

## Technical Details

* Changed CiviCRM HTML elements to their Bootstrap counterparts for our files.
* For CiviCRM components and files that can't be edited (the Date Format, for example) the changes were done using CSS. These CSS changes are on the  `scss/pages/_admin-leave-and-absences-import-form.scss` file.
* Added Backstop Scenarios for each page of the import form. A single page object, `page-objects/leave-absence-import.js`, is used for convenience to go from one page of the form to the next. The `stepsUrls` map is used to a store a unique identifier for each of the steps URLs. The `submitAndWaitForStep` clicks on the "next step" button and waits for the page change to the specific step URL:

```js
casper.thenClick('.crm-leave-and-balance-import .validate')
.then(function () {
  return casper.waitForUrl(urlRegExp);
});
```
* the file `/uploads/leave-and-absences-import-data.csv` is needed to complete the first step of the form. This is a sample import data.

## Comments
Some table rows had the error class which was not doing anything (ie: they were rendered as normal). Styles were added to display these rows as Bootstrap's danger in case this was a requirement for the original screen.

Had to fix a typo. Misspelled absence for some scenario names.

---

- [x] Tests Pass


igorpavlov:
1. "Upload data, match fields etc" should be a button group as per Styleguide
2. Selector arrow looks broken (I thought it was fixed in earlier PRs?)
3. Tables do not follow the Styleguide
4. Can we remove these shadows on selectors and input fields globally?

cc @mi-kon

mi-kon:
@AkA84 can you verify feedback from Igor please?

reneolivo:
@igorpavlov 

> 1. "Upload data, match fields etc" should be a button group as per Styleguide

Those are the wizard buttons generated by CiviCRM using the following code: 
`{include file="CRM/common/WizardHeader.tpl"}`

The output is the following:
```html
<div id="wizard-steps">
  <ul class="wizard-bar">
    <li class="current-step">»1. Upload Data</li>
    <li class="future-step"> 2. Match Fields</li>
    <li class="future-step"> 3. Preview</li>
    <li class="future-step"> 4. Summary</li>
  </ul>
</div>
<h2>Upload Data (step 1 of 4)</h2>
```

As you can see this doesn't output bootstrap elements and there's no easy way yo change the source. Should I go ahead and change the style using CSS?

> 2. Selector arrow looks broken (I thought it was fixed in earlier PRs?)

Well spotted. We indeed fixes this arrow, but only for selects that had the `.crm-select2` class (which this one does not). I will try to come up with a fix for both cases since apparently the same solution can't be applied in this case.

> 3. Tables do not follow the Styleguide

Missed adding .table class. I thought this table was not accesible, but it is.

> 4. Can we remove these shadows on selectors and input fields globally?

"globally" when inside #bootstrap-theme or even outside that wrapper? (will wait for confirmation first).

reneolivo:
# New Changes: 

* Implemented new Wizard component. The CSS/HTML specs for this Wizard can be found [here](https://github.com/civicrm/org.civicrm.shoreditch/pull/5).
* Made all input texts look like bootstrap form controllers.
* Fixed new issue with drop down arrows (this is a different issue altogether).

## After:

### Step1
wizard-1

### Step 2
wizard-2

### Step 3
wizard-3

### Step 4
wizard-4

### Technical details

* A new **WizardHeader.tpl** file was added. The location is at `civihr/hrui/templates/CRM/HRUI/common/WizardHeader.tpl` and can be used like this: `{include file="CRM/HRUI/common/WizardHeader.tpl"}`. This will generate the wizard header with all the steps and titles. The idea is that this template could be used in all extensions.

  This may not be a 100% replacement for WizardHeader.tpl ([source](https://github.com/civicrm/civicrm-core/blob/master/templates/CRM/common/WizardHeader.tpl)) since that one has more functionality that could not be texted on this implementation.

* Style changes are explained at their own pull requests.

hrui/templates/CRM/HRUI/common/WizardHeader.tpl
```diff
@@ -0,0 +1,16 @@
+{if count( $wizard.steps ) > 1}
+  <div class="panel panel-default crm_wizard__title">
+    <div class="panel-body">
+      <ul class="nav nav-pills">
+        {section name=step loop=$wizard.steps}
+        <li class="{if $wizard.currentStepNumber == $wizard.steps[step].stepNumber}active{/if}">
```
igorpavlov: Is it possible to use {section name=stepN loop=$wizard.steps} `{assign var="step" value=$wizard.steps[stepN]}` and then use `step` instead of `$wizard.steps[step]`?

PR #2125, opened 2017-09-08
PCHR-2570: Update balance

deb1990:
## Overview
As part of this PR, the following bug has been fixed.
If user changes the type of existing leave request, the balance does not change and is shown for the previous type

## Before
2570b

## Technical Details
1. An event `LeaveRequestPopup::updateBalance` is fired from `request.controller` which is subscribed by `leave-request-popup-details-tab.component`, and when the event is fired, it calls `vm.updateBalance` method

---

- [x] Tests Pass


igorpavlov:
Whilst this solution is totally fine, I am thinking we could also try and just make `updatebalance` accessible to main controller via `vm.updatebalance` and call it directly. I am just thinking that using event listeners between such a closely bound to each other controllers is a bit overkill. This a note for future and not a change request.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -437,6 +449,16 @@ define([
     }
 
     /**
+     * Gets called when the component is destroyed
+     */
+    function unsubscribeFromEvents () {
+      // destroy all the event
```
igorpavlov: 1) A comment to a function should describe what a function does and not from where is it called. What if we decide to call this function from another place - that would mean we must then change the comment, which does not sound right.
2) The `//` comment becomes unnecessary if we amend the function description
```js
/**
  * Destroys all event listeners
  */
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -437,6 +449,16 @@ define([
...
+      _.forEach(listeners, function (listener) {
+        listener();
```
igorpavlov: Does it destroy the listener?

deb1990: Yes :) Same is used in other components.

deb1990:
> Whilst this solution is totally fine, I am thinking we could also try and just make updatebalance accessible to main controller via vm.updatebalance and call it directly. I am just thinking that using event listeners between such a closely bound to each other controllers is a bit overkill. This a note for future and not a change request.

I think, exposing child controllers functions into parent's, is not the best solution because we are violating the decoupled architecture. Note: This was done for `vm.checkSubmitConditions` method(not recommended), because using events would have made the flow asynchronous. 
So usage of events is the best possible solution to keep the system decoupled, until we find a better architecture like using redux.

PR #2126, opened 2017-09-11
PCHR-2164: Upload error even after removing file

deb1990:
## Overview
The PR fixes the following bug,
if user is not able to upload a file, removes it and adds another one, an error about first file still appears. Though the second file is added to request even if user closed it without saving

## Before
2164b

## Technical Details
1. Previously in `file-upload.js`, the `uploadAll` function was always returning the promise of a same deferred object, That is why once the promise is rejected, all the subsequent promises were rejected. Changed the code to create new instance of deferred object for every `uploadAll` function call.

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/services/file-upload.js
```diff
@@ -68,7 +68,7 @@ define([
          */
         uploader: function (customSettings) {
           var uploader, oldUploadAll;
-          var deferred = $q.defer();
+          var deferred;
```
AkA84: you don't need to define `deferred` here, since you're using it only in the closure
```js
return function (additionalFormData) {
  var deferred = $q.defer();
```

deb1990: It is used outside the uploadAll function, 
https://github.com/civicrm/civihr/blob/aa765c04777f5b992c93f0c7dbfe2f41d551f315/org.civicrm.reqangular/src/common/services/file-upload.js#L84-L84
https://github.com/civicrm/civihr/blob/aa765c04777f5b992c93f0c7dbfe2f41d551f315/org.civicrm.reqangular/src/common/services/file-upload.js#L87

PR #2130, opened 2017-09-12
PCHR-2508: Manager Report Styling Bug fixes

deb1990:
## Overview
This PR fixes some styling bugs related to Manager Report page.
1. Space between + and New has been increased
2. Height of Leave Requests dropdown and New button has been increased to 40px.
3. Alignment of arrow and Filter text has been corrected.

## Before
manager_leave___hr17-2

## After
manager_leave___hr17-2


igorpavlov:
Just a thought: Some of these fixes look like they might be global fixes. As an example the chevron and the Filter caption.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave.scss
```diff
@@ -2,12 +2,29 @@
   margin-right: 5px;
   position: relative;
   top: 1px;
-  width: calc(100% - 75px);
+  width: calc(100% - 80px);
+
+  select {
+    height: 40px;
```
igorpavlov: 40px is used in a couple of places. Should it go to a variable?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave.scss
```diff
@@ -2,12 +2,29 @@
...
   .dropdown-menu {
     left: auto;
     right: 0;
   }
+
+  .dropdown-toggle {
+    height: 40px;
+  }
+
+  i.fa-plus {
```
igorpavlov: What if we add another i.fa-plus somewhere or change the icon? Please add either unique class and assign a style to it or specify the unique parent container for that `i`. Please do not use use `i.fa-plus`

deb1990: I just checked, that the `margin-right`needs to be applied for not only the `fa-plus`, but also for the `fa-plus-circle`. So instead I am using `i.fa` now.

deb1990:
> Just a thought: Some of these fixes look like they might be global fixes. As an example the chevron and the Filter caption.

The chevron and the Filter caption is used in some other places of L&A, but there is no global fix for this, as we need to fix them individually.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave.scss
```diff
@@ -1,16 +1,18 @@
+$leave-input-height: 40px;
```
igorpavlov: Can we use a more specific name for this variable? Not all inputs are affected by this style, but only some of them. Here are some suggestions: 
- `$leave-emphasised-input-height`
- `$leave-bigger-input-height`
- `$leave-taller-input-height`

PR #2131, opened 2017-09-12
PCHR-2509: Responsive manager calendar

deb1990:
## Overview
As part of the PR the following bugs are fixed
1. Title text for the months should be 16px
2. In the legend , texts with 3 lines, should start from same left margin.

## Before
manager_leave___hr17-2

## After
manager_leave___hr17-2
---


igorpavlov:
Is the size of the header dictated by the Styleguide?
image

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -164,21 +164,27 @@ $chr-leave-calendar-legend-size:                  20px;
   }
 
   > .chr_leave-calendar__item {
-    float: left;
     font-size: 8px;
     height: $chr-leave-calendar-legend-size;
     margin-right: 10px;
+    position: absolute;
```
igorpavlov: Could you please explain what it fixes and how?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
```diff
@@ -164,21 +164,27 @@ $chr-leave-calendar-legend-size:                  20px;
...
   font-size: $font-size-small;
+  margin-left: 26px;
 }
 
 
 .chr_leave-calendar__month-header {
   background: $panel-default-heading-bg;
+
+  .panel-title {
+    font-size: 16px;
```
igorpavlov: Please see a comment to the PR about the Styleguide.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
deb1990: It fixes the legend, where the 3rd line used to be shifted to the left. But after removing `float: left;`, that does not happen anymore. But still to keep the text adjacent to the Icon, it is necessary to add `position: absolute;`, so that its layout is calculated separately, and it does not interfere with the text's position. Does that make sense?

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_leave-calendar.scss
deb1990: No its different, because Shoreditch and SSP are 2 completely separate themes. In SSP 18px is the default font size for panel-title. But in this particular case it needs to be 16px(confirmed by Jon).

PR #2134, opened 2017-09-12
PCHR-2563: Fix bug when validating Balance Change on Leave Request Import

tunbola:
## Overview
Currently, when Importing Leave Requests the validations for balance change is buggy because it does use the amount to be deducted as it is in the CSV for the validation, rather it uses the amount calculated from the contact's work pattern for the dates to be imported for. 
This PR fixes this issue.

## Before
When importing Leave Requests, the leave request balance change amount used for validating balance change is the one gotten from the `LeaveRequest::calculateBalanceChange` method rather than the `total_qty` field in the CSV.

## After
- The second parameter passed to `LeaveRequest::create` was changed from `$validate` to `$validationMode` where the available validation Modes are: VALIDATIONS_ON, VALIDATIONS_OFF and IMPORT_VALIDATION.
- The validationMode is passed as  IMPORT_VALIDATION when creating LeaveRequests on import, this allows some Validations to be bypassed when Leave Requests are imported.
- The validations for Balance Change is now done in the `CRM_HRLeaveAndAbsences_Import_Parser_Base` class.

## Comments
Also there were no tests for the import functionality before. This PR adds some basic tests for the import functionality.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
@@ -632,7 +632,7 @@ class CRM_HRLeaveAndAbsences_Import_Parser_Base extends CRM_HRLeaveAndAbsences_I
 
     if(!$absenceType->allow_overuse && $leaveRequestBalance > $currentBalance) {
       throw new InvalidLeaveRequestException(
-        'There are only '. $currentBalance .' days leave available. This request cannot be made or approved',
+        'There are only '. $currentBalance .' days leave available. This request cannot be imported',
         'leave_request_balance_change_greater_than_remaining_balance',
         'type_id'
       );
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
@@ -3,7 +3,10 @@
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 use CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange as LeaveBalanceChange;
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
+use CRM_HRLeaveAndAbsences_BAO_LeavePeriodEntitlement as LeavePeriodEntitlement;
 use CRM_HRLeaveAndAbsences_Service_LeaveRequestComment as LeaveRequestCommentService;
+use CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException as InvalidLeaveRequestException;
 
 /**
  * Class to parse activity csv files.
@@ -459,8 +462,13 @@ class CRM_HRLeaveAndAbsences_Import_Parser_Base extends CRM_HRLeaveAndAbsences_I
 
   /**
    * Creates a leave request from the params array.
+   * Also it validates the balance change and entitlement
+   * before making the call to LeaveRequest::create with
+   * the validation mode being of type IMPORT.
    *
    * @param array $params
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
    *
    * @return LeaveRequest
    */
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Import/Parser/Base.php
```diff
@@ -563,4 +573,81 @@ private function getAbsenceStatuses() {
...
+      );
+    }
+
+    $leaveRequestBalance = $params['total_qty'];
+    $currentBalance = $leavePeriodEntitlement->getBalance();
+    $absenceType = AbsenceType::findById($params['type_id']);
+
+    if(!$absenceType->allow_overuse && $leaveRequestBalance > $currentBalance) {
+      throw new InvalidLeaveRequestException(
+        'There are only '. $currentBalance .' days leave available. This request cannot be made or approved',
```
davialexandre: This message will be included in the errors report after import, right? If that's true, I think that "This request cannot be made or approved" won't make sense, as I'm not trying to make or approve anything, but to import some data.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Import/Parser/BaseTest.php
```diff
@@ -0,0 +1,288 @@
...
+   */
+  public function testValidateLeaveParamsThrowsExceptionWhenContactHasNoPeriodEntitlementForTheAbsenceType() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date'   => CRM_Utils_Date::processDate('2016-12-31'),
+    ]);
+
+    $contactId = 1;
+    $row1 = [
+      'contact_id' => $contactId,
```
davialexandre: You're not using this $contactId variable to anything else other than this, so I think you can drop it and just pass 1 here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Import/Parser/BaseTest.php
```diff
@@ -0,0 +1,288 @@
...
+      'start_date' => '2016-01-01',
+      'end_date' => '2016-01-01',
+      'total_qty' => 3,
+      'status_id' => 1,
+      'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE,
+    ];
+
+    $fields = array_keys($row1);
+    $importObject = $this->getImportObject($fields);
+    $importObject->validateLeaveParams($row1);
```
davialexandre: Perhaps you should use `setExpectedException()` here the same way as you did for the next test?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Import/Parser/BaseTest.php
```diff
@@ -0,0 +1,288 @@
...
+      'absence_date' => '2016-01-01',
+      'qty' => 4,
+      'start_date' => '2016-01-01',
+      'end_date' => '2016-01-01',
+      'total_qty' => 4,
+      'status_id' => 1,
+      'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE,
+    ];
+
+    $this->setExpectedException(
```
davialexandre: What about moving this to right before the `validateLeaveParams()` call to make it clear that we're expecting this method to throw the exception?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Import/Parser/BaseTest.php
```diff
@@ -0,0 +1,288 @@
...
+
+    $fields = array_keys($row1);
+    $importObject = $this->getImportObject($fields);
+    //Since the request type is TOIL, balance change validation will not
+    //be taken into account.
+    foreach(LeaveRequest::getCancelledStatuses() as $status) {
+      $row1['status_id'] = $status;
+      $importObject->validateLeaveParams($row1);
+    }
+  }
```
davialexandre: Instead of doing validations for the `validateLeaveParams()` params method, wouldn't it be better to test if the LeaveRequest can or cannot be imported on those scenarios and keep the validation method private?

tunbola: If we test the using the import method of the `CRM_HRLeaveAndAbsences_Import_Parser_Base` class, the exceptions are already caught there and all it would return is a `CRM_Import_Parser::ERROR` error. Would it be fine to just test for that?

davialexandre: Yeah. One of the problems of testing only the `validateLeaveParams()` method is that there's no way to guarantee that this method will actually be called when you run the import. So I would say it's more important to test that we'll get an error trying to import an invalid LeaveRequest than to test the validation in isolation.

PR #2137, opened 2017-09-14
PCHR-2584: Support Admin logic in L&A Balance Tab

igorpavlov:
## Overview

This PR adds support to Admin logic: admins see contacts who have either Leave Requests or Job Contracts in the given period.

## Before

Admin doesn't see any contacts (since the Manager's logic is applied instead).

image

## After

Admin sees correct list of contacts.

image

## Technical Details

The user role defines how we call .getLeaveBalances() API:
```js
var params = {
  period_id: filters.absence_period,
  type_id: filters.absence_type
};

// managed_by is used for manager, but if omitted - it will return correct contacts for admin
(userRole === 'manager') && (params.managed_by = vm.loggedInContactId);

return LeaveBalanceReport.all(params, vm.pagination)
  .then(function (response) { /*...*/ }
```

## Comments

This PR also adds a caption if there are not Leave Balances found.

image

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
```diff
@@ -101,15 +106,17 @@ define([
      * @return {Promise}
      */
     function loadReportCurrentPage () {
+      var params = {
+        period_id: filters.absence_period,
+        type_id: filters.absence_type
+      };
+
+      (userRole === 'manager') && (params.managed_by = vm.loggedInContactId);
```
reneolivo: I know this is simple JS logic, but it's too clever. Keep it simple, use an if:
```js
if (userRole === 'manager') {
  params.managed_by = vm.loggedInContactId;
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
```diff
@@ -144,6 +151,18 @@ define([
     }
 
     /**
+     * Sets the user's role based on their permissions
+     *
+     * @return {Promise}
+     */
+    function setUserRole () {
```
reneolivo: why not add this logic to the filters component? The filter component should pass all parameters necessary to filter the request. This component should worry about retrieving and displaying the records only.

reneolivo: Remember that the filter component is not just a for visual filters.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
igorpavlov: @AkA84 loves this syntax. I started liking it too. I will change, np.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
igorpavlov: > Remember that the filter component is not just a for visual filters.

I am not quite sure it is. It produces filters and then passes it to the master component. It is the master component that calls API via `loadReportCurrentPage()` basing on these filters. 
That's actually why I decided to delete `managed_by` from `filters = {}` since I realised it is not a filter, but a property that is needed to call the API.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
igorpavlov: @reneolivo funnily enough, this change doesn't affect .min files!

igorpavlov:
@reneolivo I made the changes we discussed today. However, this time I cannot avoid renaming filters since we want to transparently pass them without any mapping and unfortunately they have to match the property at the API endpoint then.

PR #2138, opened 2017-09-14
(WIP) PCHR-2560: Onboarding email

mickadoo:
## Overview
For new users invited to the system by either the bulk "Create User Accounts" action from the advanced search, or by clicking "Create User Account" from the user profile page, the welcome email should match the new format. This is a Drupal system email and the email content will be modified in civihr_employee_portal, but the email template will be generated using Foundation in the hremails extension.

## Before
The "account activated" system email is generated by Drupal and uses the Drupal template.

image

## After
The "account activated" system email uses the template from hremails. In this PR the only change is the new template and partials.

image

## Notes

- The email content is not final yet so this PR is still WIP

---

- [ ] Tests Pass
There are no tests in the hremails extension.


mickadoo:
Closed in favor of #2214

PR #2139, opened 2017-09-14
PCHR-2503: Make Contract ID compulsory when importing Contract Entitlements

tunbola:
## Overview
When importing Contract entitlements for Contacts with existing Contract, Only the Contact ID field is required and entitlements are imported for the latest revision of the Contact's current Contact. However since Entitlement Calculation needs to be done after the import, there is a flaw in this process as the Entitlement calculation considers all active Contracts within a Period for the contact, so it is necessary to update entitlements for all the Contact's Contracts.
Also a message informing the user of the need to go to the Entitlement calculation after a successful Contract entitlement update is added in this PR.

## Before
- When importing Contract entitlements for a Contact, the Contact ID field is required
- There is no message prompting the user to go to the entitlement calculation page after successfully importing Contract entitlements.

## After
- When importing Contract entitlements for a Contact, the Contract ID field is now required.
- There is now a  message prompting the user to go to the entitlement calculation page after successfully importing Contract entitlements.
import job contracts - civihr 1 7 demo 2017-09-14 18-13-22

---

- [X] Tests Pass

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Import/Form/SummaryBaseClass.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Import/Form/SummaryBaseClass.php
@@ -33,6 +33,8 @@
  *
  */
 
+use CRM_Hrjobcontract_Import_Parser as Import_Parser;
+
 /**
  * This class summarizes the import results
  */
@@ -53,6 +55,7 @@ class CRM_Hrjobcontract_Import_Form_SummaryBaseClass extends CRM_Import_Form_Sum
   public function preProcess() {
     // set the error message path to display
     $errorFile = $this->assign('errorFile', $this->get('errorFile'));
+    $this->_importMode = $this->get('importMode');
 
     $totalRowCount = $this->get('totalRowCount');
     $relatedCount = $this->get('relatedCount');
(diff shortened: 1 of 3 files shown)


hrjobcontract/CRM/Hrjobcontract/Import/Form/SummaryBaseClass.php
```diff
@@ -104,5 +107,12 @@ public function preProcess() {
     foreach ($properties as $property) {
       $this->assign($property, $this->get($property));
     }
+
+    $isEntitlementUpdate = $this->_importMode == Import_Parser::UPDATE_ENTITLEMENTS;
+    $entitlementPageUrl = CRM_Utils_System::url(
+      'civicrm/admin/leaveandabsences/periods/manage_entitlements'
```
davialexandre: Without passing any params, the system will calculate the entitlement for ALL contacts during the current period. This URL accepts a list of contacts to only do the calculation for them. Do you think it would be possible keep a list of the contacts whose contracts were imported and then pass their IDs to the URL?

tunbola: Did some investigations and it will be possible to do that but because of the data limitations of GET request, would leave as is, as agreed.

PR #2146, opened 2017-09-15
PCHR-2596: Fix Absence Tab on 13 inch screen

igorpavlov:
## Overview

Currently right hand side area in Absence Tab is completely unaccessible on 13 inch screen. This PR fixes this issue.

## Before

Calculation:

screen shot 2017-09-18 at 10 20 57

## After

Calculations:

screen shot 2017-09-18 at 10 09 21

## Technical Details

The following rule was applied to make paddings consistent and ensure the component is never wider than its container.

```css
.absence-tab-page {

  >.container-fluid {
    width: 83vw;
    padding-right: calc(8px + 1vw);
    max-width: 1156px;
  }

  ...
}
```

- `padding-right` is used instead of `margin-right` because `margin: 0 auto` is already used
- `padding-right: calc(8px + 1vw)` is used instead of `width: calc(84vw - 8px)` because in case of `width` the container will not be aligned to the center.


uk.co.compucorp.civicrm.hrleaveandabsences/scss/pages/_absence-tab-page.scss
```diff
@@ -1,5 +1,19 @@
...
+  /**
+   * This is a temporary fix.
+   * It provides a consistency to paddings and at the same time
+   * preserves the same maximum width for larger screens.
+   * @TODO this fix should be removed once all other
+   * screens on admin have consistent layout
+  */
+  >.container-fluid {
+    width: 99%;
+    padding-right: calc(8px + 1%);
```
reneolivo: I can see that this is working but shouldn't width be something like `width: calc(99% - 8px)` ?

igorpavlov: @reneolivo I updated the PR description and it should now answer your question. I now also use viewport width instead of % since it give even higher precision.

PR #2147, opened 2017-09-15
PCHR-2545: Fix work pattern screens

reneolivo:
## Overview
This PR fixes issues that happened after migrating the Work Patterns screens to Bootstrap. Some of the issues are:

* Not using font awesome icons on the work pattern list (for the default work pattern checked icon).
* Form title was not a panel heading.
* Form inputs and selects didn't look like Shoreditch inputs.

## Before
anim

## After
anim

## Technical Details
* moved all `.work-pattern-week` rules to a single block.
* added the `.crm-form-block` class to the panel, which makes inputs and select look like `.form-control` elements.
* removed all input rules since they now are included in the `.form-control` rule. Except for `min-width` since this is 230px by default. Set it to 0 so the inputs would adapt inside the table. Same for `select` elements.
* Moved `h1` form element inside a `.panel-heading` element.
* removed this line of block:
  ```html
  <style>		
    .btn-slide ul.panel { display: none !important; }		
  </style>
  ```
  since it was previously added for testing and it was not removed.
* BackstopJS tests for Work Pattern screens were run successfully. 

---

- [x] Tests Pass


igorpavlov:
Could you please provide screenshots for mobile view since "moved all .work-pattern-week rules to a single block"?

reneolivo:
@igorpavlov 

iPad portrait
anim

iPad landscape
anim

The only noticeable glitch is the Weekend label overlapping the dropdown icon.

PR #2149, opened 2017-09-18
PCHR-2538: Fix issue with not being able to update an already Approved Leave when Entitlement Balance is Zero

tunbola:
## Overview
When trying to update an already approved leave request, e.g trying to change the status or add a comment, or upload a file and the Entitlement Balance is Zero, an error: ` 'There are only 0 days leave available. This request cannot be made or approved` is returned to the manager/user trying to update the Leave.

## Before
Trying to update an already approved leave request e.g trying to change the status or add a comment, or upload a file returns an error when the Entitlement Balance is Zero.
- The reason for the error returned is that when validating an already approved leave, the already approved leave is considered as a fresh leave request not minding the fact that its balance change has previously been deducted from the entitlement Balance.
- Since the approved leave to be updated is treated as a fresh leave request, once the leave balance is greater than the entitlement balance not only when it is Zero, an error is thrown.

## After
- An already approved leave request is no longer treated as a fresh leave request, Instead leave request ID is passed to the `getBalance` method to exclude the balance of that leave request from the entitlement balance calculation.
- In the case when the dates has changed and the new leave balance is now greater than the old leave balance, an error is thrown inasmuch as the leave balance is greater than the entitlement balance.
- In the case when the Work Pattern has changed and the change_balance parameter is true and the new leave balance is now greater than the old leave balance, an error is thrown inasmuch as the leave balance is greater than the entitlement balance.
- The change_balance parameter is taken into consideration during leave request validation. 

## Comments
- A broken test for Leave Import was also fixed as part of this PR
---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -475,15 +475,16 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       );
     }
 
-    $currentBalance = $leavePeriodEntitlement->getBalance();
-
+    $requestsToExcludeFromBalance = [];
     if (!empty($params['id'])) {
       $oldLeaveRequest = self::findById($params['id']);
 
-      if ($oldLeaveRequest && $oldLeaveRequest->id && self::isAlreadyApproved($oldLeaveRequest)) {
-        $currentBalance = $leavePeriodEntitlement->getBalance([$oldLeaveRequest->id]);
+      if (self::isAlreadyApproved($oldLeaveRequest)) {
+        $requestsToExcludeFromBalance[] = $oldLeaveRequest->id;
       }
     }
+
+    $currentBalance = $leavePeriodEntitlement->getBalance($requestsToExcludeFromBalance);
 
     if(!$absenceType->allow_overuse && $leaveRequestBalance > $currentBalance) {
       throw new InvalidLeaveRequestException(
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -477,6 +477,20 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
 
     $currentBalance = $leavePeriodEntitlement->getBalance();
 
+    if (!empty($params['id'])) {
+      $oldLeaveRequest = self::findById($params['id']);
+
+      if ($oldLeaveRequest && $oldLeaveRequest->id && self::isAlreadyApproved($oldLeaveRequest)) {
+        $periodContainingOldLeaveDates = self::getPeriodContainingDates($oldLeaveRequest);
+        $isSamePeriodWithOldLeave = $periodContainingOldLeaveDates->id == $period->id;
+
+        if ($isSamePeriodWithOldLeave) {
+          $oldLeaveRequestBalance = abs(LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($oldLeaveRequest));
+          $currentBalance += $oldLeaveRequestBalance;
+        }
+      }
+    }
+
     if(!$absenceType->allow_overuse && $leaveRequestBalance > $currentBalance) {
       throw new InvalidLeaveRequestException(
         'There are only '. $currentBalance .' days leave available. This request cannot be made or approved',
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -466,7 +466,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       return;
     }
 
-    $leaveRequestBalance = self::calculateBalanceChangeFromCreateParams($params);
+    $leaveRequestBalance = self::getLeaveRequestBalance($params);
     if ($leaveRequestBalance == 0) {
       throw new InvalidLeaveRequestException(
         'Leave Request must have at least one working day to be created',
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -477,6 +477,20 @@ private static function validateEntitlementAndWorkingDayAndBalanceChange($params
...
+        $periodContainingOldLeaveDates = self::getPeriodContainingDates($oldLeaveRequest);
+        $isSamePeriodWithOldLeave = $periodContainingOldLeaveDates->id == $period->id;
+
+        if ($isSamePeriodWithOldLeave) {
+          $oldLeaveRequestBalance = abs(LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($oldLeaveRequest));
+          $currentBalance += $oldLeaveRequestBalance;
+        }
+      }
+    }
+
```
davialexandre: This is basically the same problem from PCHR-2471. I feel that a better approach would be to change LeavePeriodEntitlement.getBalance() to also, optionally, accept the ID of a Leave Request to be excluded from the balance. 

The `$leavePeriodEntitlement` instance is only related to the period of containing the from_date and to_date in $params, right? If that's true, then:
- If the dates didn't change, we'll get the balance containing all the approved leave requests, except the one with the given ID, and then the `$leaveRequestBalance > $currentBalance` will work
- If the dates changed, but are still in the same period, the same logic above will work
- If the dates changed and are now in a different period, `$leavePeriodEntitlement` will refer to this new period and will try to exclude the balance changes for the given leave request, but, on the DB, the leave request still belongs to the old period and will not be included, i.e. the logic will still work

There's another problem here though. Currently, we're always using the `calculateBalanceChangeFromCreateParams()` method to get the balance change for the Leave Request. If there are changes to the work pattern(s) used during the request period, it can cause the balance to be different than what it was when the request was approved. This should be consistent with the new `change_balance` logic. If that param is present and true, then we calculate the balance change according to the new data, otherwise, we consider the value we have in the DB

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -477,6 +477,14 @@ private static function validateEntitlementAndWorkingDayAndBalanceChange($params
...
     $currentBalance = $leavePeriodEntitlement->getBalance();
 
+    if (!empty($params['id'])) {
+      $oldLeaveRequest = self::findById($params['id']);
+
+      if ($oldLeaveRequest && $oldLeaveRequest->id && self::isAlreadyApproved($oldLeaveRequest)) {
+        $currentBalance = $leavePeriodEntitlement->getBalance([$oldLeaveRequest->id]);
+      }
+    }
+
```
davialexandre: What about this?

```php
    $requestsToExcludeFromBalance = [];
    if (!empty($params['id'])) {
      $oldLeaveRequest = self::findById($params['id']);

      if ($oldLeaveRequest && $oldLeaveRequest->id && self::isAlreadyApproved($oldLeaveRequest)) {
         $requestsToExcludeFromBalance[] = $oldLeaveRequest->id;
      }
    }
    $currentBalance = $leavePeriodEntitlement->getBalance();
```

It's a small optimization but with that we avoid calling getBalance() twice when the leave request has already been approved

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -477,6 +477,14 @@ private static function validateEntitlementAndWorkingDayAndBalanceChange($params
 
     $currentBalance = $leavePeriodEntitlement->getBalance();
 
+    if (!empty($params['id'])) {
+      $oldLeaveRequest = self::findById($params['id']);
+
+      if ($oldLeaveRequest && $oldLeaveRequest->id && self::isAlreadyApproved($oldLeaveRequest)) {
```
davialexandre: The `findById()` method throws an error if it can't find a record with the given ID. So it either stops there or returns a valid object, meaning there's no need to check if the `$oldLeaveRequest` variable is null

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -3264,4 +3265,469 @@ public function testLeaveRequestBeCreatedWhenLeaveRequestHasNoWorkingDayAndValid
...
+
+    $params['id'] = $leaveRequest->id;
+    $params['status'] = 3;
+
+    //Update Leave Request status
+    $leaveRequest = LeaveRequest::create($params);
+    $this->assertNotNull($leaveRequest->id);
+  }
+
+  public function testAlreadyApprovedLeaveRequestCanNotBeUpdatedWhenDatesChangeAndLeaveBalanceGreaterPreviousLeaveBalanceAndEntitlementBalanceIsZero() {
```
davialexandre: Perhaps a slightly shorter name:
```
testApprovedRequestCanNotBeUpdatedWhenCurrentBalanceIsZeroAndDatesChangeAndBalanceChangeIsGreaterThanPreviousBalanceChange
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -3264,4 +3265,469 @@ public function testLeaveRequestBeCreatedWhenLeaveRequestHasNoWorkingDayAndValid
...
+    //Entitlement balance is Zero. The three leave days have been deducted.
+    $this->assertEquals(0, $periodEntitlement->getBalance());
+
+    //Update leave request and add one more day to the request so that
+    //It will now create a balance change of -4 as opposed to -3 previously.
+    $params['id'] = $leaveRequest->id;
+    $params['from_date'] = CRM_Utils_Date::processDate('2016-11-14');
+    $params['to_date'] = CRM_Utils_Date::processDate('2016-11-17');
+
+    //Since the dates as changed, the request is treated as a fresh request as if it is just being requested
```
davialexandre: s/as/are/

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -3264,4 +3265,469 @@ public function testLeaveRequestBeCreatedWhenLeaveRequestHasNoWorkingDayAndValid
...
+    ];
+
+    $leaveRequest = LeaveRequestFabricator::fabricate($params, true);
+
+    $this->assertNotNull($leaveRequest->id);
+    //Entitlement balance is Zero. The three leave days have been deducted.
+    $this->assertEquals(0, $periodEntitlement1->getBalance());
+
+    //Update Leave Request dates with dates being in the second period.
+    //The balance deducted when leave request was made will not be added to the entitlement balance
```
davialexandre: Since we have two entitlements in this test, it's better to clarify to which of them you referring to in this sentence

PR #2150, opened 2017-09-18
PCHR-2608: Don't install the L&A sample data automatically

davialexandre:
## Overview
Currently, the auto_install.sql in L&A contains some sample data that gets installed whenever the extension is also installed.

This data was added to make it easier for developers and testers to quickly have something to play with right after installing the extension and save some time by avoiding having to create contracts, calculate entitlements, create leave requests etc.

However, creating the sample data can cause some problems:
- When running the phpunit tests, civi resets the schema and installs the extension, which results in the test db getting populate with the sample data and making some tests to fail
- For client sites we should not create the sample data, so we have to manually remove it from the auto_install.sql prior to the extension installation

This PR is a way to still keep it easier to import the sample data to the system but avoid the mentioned problems.

## Before
The sample data was automatically created when the L&A extension was installed

## After
The sample data isn't created when the extension is installed.

Now, there's bash script that can be used to import/clear the sample data. Inside the L&A folder, there's now a bin folder which contains a `sample_data.sh` script.

To import the sample data to a site, you just have to run:

```
./bin/sample_data.sh import
```

And to clear the sample data from the site, run:

```
./bin/sample_data.sh clear
```

**Note**: for both commands, it's assumed that you're running them from the L&A root folder

## Technical Details
- The `auto_install.sql` and `auto_uninstall.sql` are not only responsible for handling the extension schema and some default data necessary for the extension to work
- The sample data is now handled by the `sample_data.sql` and `sample_data_cleanup.sql` files, inside of the `sql` folder
- The sample data scripts were improved to avoid having hardcoded IDs for the staff and manager users. They now get the ID based on the contact display name and store them in variables that are used for all the other statements
- The cleanup script was updated to better removed the data related to contracts. It now only removes the data related to the civihr_staff user, which should decrease the risk of removing contracts from other users, created before the sample data was imported
- `drush` and `cv` are both required to run the `sample_data.sh` script. If they are not present, the command will fail. The server where the test sites run has this command and the devs local environment should have them as well, so it's ok to have these requirements.
---

- [ ] Tests Pass
There's one failing test regarding the Import that will be fixed by another PR


uk.co.compucorp.civicrm.hrleaveandabsences/sql/sample_data_cleanup.sql
```diff
@@ -0,0 +1,48 @@
...
+  LEFT JOIN civicrm_hrjobcontract_role role
+    ON role.jobcontract_revision_id = revision.id
+WHERE contract.contact_id = @staffID;
+
+TRUNCATE civicrm_hrleaveandabsences_absence_period;
+TRUNCATE civicrm_hrleaveandabsences_public_holiday;
+TRUNCATE civicrm_hrleaveandabsences_leave_period_entitlement;
+TRUNCATE civicrm_hrleaveandabsences_leave_request;
+TRUNCATE civicrm_hrleaveandabsences_leave_request_date;
+TRUNCATE civicrm_hrleaveandabsences_leave_balance_change;
```
tunbola: Thinking maybe we should delete only the data previously inserted into these tables by the sample data sql.

Joining leave_request to leave_request_date to balance_change table where the leave_request contact_id is @staffID can help get rid of leave request, leave request dates and their balance changes for the sample staff contact. Joining period_entitlement to balance changes too can help get rid of entitlements for the contact and the entitlement balance changes. For public holidays also, seems the ID's are sequential.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_uninstall.sql
```diff
@@ -12,17 +12,3 @@ DROP TABLE IF EXISTS `civicrm_hrleaveandabsences_work_week`;
 DROP TABLE IF EXISTS `civicrm_hrleaveandabsences_work_pattern`;
 DROP TABLE IF EXISTS `civicrm_hrleaveandabsences_absence_period`;
 DROP TABLE IF EXISTS `civicrm_hrleaveandabsences_public_holiday`;
-
--- Deletes sample data for contracts (for l&a, the drop tables already take care of the job)
-SET foreign_key_checks = 0;
```
tunbola: I think there are some references to old tables in this file that we should remove too. e.g `civicrm_hrleaveandabsences_sickness_request`, `civicrm_hrleaveandabsences_sickness_request`

uk.co.compucorp.civicrm.hrleaveandabsences/sql/sample_data_cleanup.sql
davialexandre: I considered that, but I thought it was not worth. The main problem here are the Absence Periods. All the sample data depend on having the 2016 and the 2017 periods. That's why it creates those periods. So we have two options:
- Either delete the Absence Periods while clearing the sample data. In this case, there's no reason to keep all the other data, as there will be no period to contain them
- Or keep the Absence Period and delete the data. In this case, we have to consider two scenarios:
  - What happens if the user changes the absence period that we left in the DB?
  - We need to make the import script smarter to check if the period exists and only add them if they don't. To be honest, it's not only a matter of checking if they exist but also check if the dates are correct etc.

As you can see, it's a bit more complex than simply deleting the requests belonging to the staff. That's why I gave up and kept this working as it was before.

Note that this is different for contracts because the contract extension could have been installed before L&A and we don't want to delete data that was there previously

tunbola: Ok then. Inasmuch as sample data will not be installed on a real client site, I guess its fine then.

PR #2152, opened 2017-09-19
PCHR-2594: Lock document scroll while a modal is open

igorpavlov:
## Overview

This PR adds a decorator to existing Angular Bootstrap Modal service to prevent user scrolling the body while a modal is opened.

## Before

It is possible to scroll through the document while a modal is open.

## After

It is **not** possible to scroll through the document while a modal is open. The document scroll is then released once the modal is closed or dismissed.

## Technical Details

A decorator was created for Angular Bootstrap Modal which enables this behaviour automatically wherever Modal service is used.

```js
function open () { // Extends .open()
  // These styles are needed to completely lock scroll (including iOS)
  var lockScrollStyling = ';overflow: hidden;height: 100%;width: 100%;';
  // Both HTML and BODY need to inherit the styles mentioned above
  var elements = [
    { node: $document[0].body },
    { node: $document[0].getElementsByTagName('html')[0] }
  ];

  // Applying styles
  elements.forEach(function (element) {
    element.style = element.node.getAttribute('style');
    element.node.setAttribute('style', element.style + lockScrollStyling);
  });

  // Calling original (not overridden) Modal .open() function passing all original arguments
  modalInstance = openFunction.apply(this, arguments);

    // Reverting styles on modal close
    elements.forEach(function (element) {
      element.node.setAttribute('style', element.style);
    });
  });

  // returning same object as in original .open() method
  return modalInstance;
}
```

## Comments

Note: 100% height/width is needed for iOS.

This solution does not utilise CSS classes like:
```css
body.lock-scroll, html.lock-scroll {
  height: 100%;
  width: 100%;
  overflow: hidden;
}
```
since it makes it reliable on CSS styling 

⚠️ Tests for modal closing do not currently work and were temporarily suppressed.

---

- [x] Tests Pass


org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-env amd */
+
+define([], function () {
```
AkA84:
```js
define(function () {
```

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Decorates uibModal to set prevent body from scrolling while a modal is open
+     * and revert to the original styles values when it is closed
+     *
+     * @param  {any}
+     * @return {Object} Modal Instance
+     */
+    function open () {
+      var lockScrollStyling = ';overflow: hidden;height: 100%;width: 100%;';
```
AkA84:
```js
 var lockScrollStyle = 'overflow: hidden; height: 100%; width: 100%;';
```
The `;` should not be part of the variable content, just add it when you concatenate the strings

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
...
+      var modalInstance;
```
AkA84: put vars that are declared but not assigned at the top of the list
```js
var modalInstance;
var lockScrollStyle = // ...
var elements = // ...
```

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
...
+      var elements = [
```
AkA84: Define the whole object structure from the start, don't add `style` only later, it makes the code less readable

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
...
+        { node: $document[0].body },
+        { node: $document[0].getElementsByTagName('html')[0] }
```
AkA84: you sure you need the html tag as well? Doesn't bootstrap deal with the body only?

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,44 @@
...
+      ];
+
+      elements.forEach(function (element) {
+        element.style = element.node.getAttribute('style');
```
AkA84: call the property `originalStyle` or something like that, it makes its purpose more evident

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
+/* eslint-env amd, jasmine */
+
+define([
+  'common/lodash',
+  'common/angularMocks',
+  'common/angular',
+  'common/decorators/ui-bootstrap/uib-modal.decorator',
+  'common/angularBootstrap'
+], function (_, angularMocks, angular, $uibModalDecorator, $document) {
```
AkA84: First you list the modules that you need to access in the callback, then the others. There is no use for `angularMocks`, so it shouldn't be there

AkA84: I don't think `$document` should be there at all

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    var $uibModal, $provide, $document;
+
+    beforeEach(module('ui.bootstrap'));
+
+    beforeEach(function () {
+      module(function (_$provide_) {
+        $provide = _$provide_;
+      });
+      inject(function () {
+        $provide.decorator('$uibModal', $uibModalDecorator);
```
AkA84: Can't you do

```js
module(function ($provide) {
  $provide.decorator('$uibModal', $uibModalDecorator);
});
```

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+      });
+    });
+
+    beforeEach(inject(function (_$uibModal_, _$document_) {
```
AkA84: alphabetical order for the params

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+      $uibModal = _$uibModal_;
+      $document = _$document_;
+    }));
+
+    describe('init', function () {
+      var arg1 = { template: '<component></component>' };
+      var arg2 = 'another_argument';
+      var elements = {};
+      var doc, originalStyles;
```
AkA84: first unassigned vars, then the assigned ones

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+      beforeEach(function () {
+        doc = $document[0];
```
AkA84: 1) if you are using `doc` only in the `beforeEach` callback, then you don't need it declared it in the outer function
2) I don't think this variable is needed at all, just use `$document[0]` directly

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+        elements.body = doc.body;
+        elements.html = doc.getElementsByTagName('html')[0];
+
+        if (!originalStyles) {
```
AkA84: why do you need to check if the var is already assigned?

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+          originalStyles = {
```
AkA84: shouldn't you mirror what you do in the decorator, having the original style directly as a property of the element? This looks inconsistent

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,83 @@
...
+            html: _.clone(elements.html.style)
+          };
+        }
+      });
+
+      describe('when modal is opened', function () {
+        var modalInstance;
+
+        beforeEach(function () {
+          spyOn($uibModal, 'open').and.callThrough();
```
AkA84: you don't need this

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
igorpavlov: That adds code duplication, but OK.

igorpavlov: Done.

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
igorpavlov: Exactly, Bootstrap only deals with `<body>`. As per https://stackoverflow.com/questions/3047337/does-overflowhidden-applied-to-body-work-on-iphone-safari (please see the answer with 57 upvotes). The `position: fixed` is an overkill though.

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
igorpavlov: AngularMocks define the `module` function.

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
igorpavlov: I need to assign in only **once**. I tried beforeAll, but it is executed **after** nested beforeEach.

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
igorpavlov: I don't think the implementation must be exactly the same as tests. Test is doing exactly what it is designed for and the implementation produces the same results but differently. If we are testing the implementation in the exactly same way we have a risk of writing wrong tests (if we wrote wrong implementation).

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,50 @@
...
+    function open () {
+      var modalInstance;
+      var lockScrollStyle = 'overflow: hidden;height: 100%;width: 100%;';
+      var elements = [
+        {
+          node: $document[0].body,
+          originalStyle: $document[0].body.getAttribute('style')
+        },
+        {
+          node: document.getElementsByTagName('html')[0],
```
deb1990: Lets use `$document` here too.

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Decorates uibModal to set prevent body from scrolling while a modal is open
+     * and revert to the original styles values when it is closed
+     *
+     * @param  {any}
+     * @return {Object} Modal Instance
+     */
+    function open () {
+      var modalInstance;
+      var lockScrollStyle = 'overflow: hidden;height: 100%;width: 100%;';
```
deb1990: Add a space after `;`, 
`'overflow: hidden; height: 100%; width: 100%;'`

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      var originalStyles;
+      var arg1 = { template: '<component></component>' };
+      var arg2 = 'another_argument';
+      var elements = {};
+
+      beforeEach(function () {
+        elements.body = $document[0].body;
+        elements.html = $document[0].getElementsByTagName('html')[0];
+
+        if (!originalStyles) {
```
deb1990: Why do we need to check this condition?

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-env amd */
+
+define(function () {
+  'use strict';
+
+  return ['$delegate', '$document', function ($delegate, $document) {
+    var openFunction = $delegate.open;
```
deb1990: Name it something like `originalOpenFunction`, so that it is distinguishable from `open`.

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
igorpavlov: Why? This style is appended to the end of the current style.

deb1990: Sorry I meant space after `:`

org.civicrm.reqangular/test/decorators/ui-bootstrap/uib-modal.decorator.spec.js
igorpavlov: I need to assign in only once. I tried beforeAll, but it is executed after nested beforeEach.

PR #2153, opened 2017-09-19
PCHR-2547: Fix Manage Entitlement form

reneolivo:
:warning: this PR needs #2151 to me merged first.

## Overview
This PR fixes the inputs on the Manage Entitlements form. It also fixes the panel footers of several admin screens.

## Before
anim

## After
anim
* The footer has 20px all around instead of 15px top/bottom and 20px left/right.
* The inputs are form-control inputs and have the right width.

## Technical Details
* The size attribute for the proposed entitlement fields was set to 4 characters. This gives it a specific width:
```php
$this->add(
        'text',
        $proposedEntitlementFieldName,
        '',
        [
          'class' => 'overridden-proposed-entitlement',
          'maxlength' => 4,
          'size' => 4
        ]
);
```

* `.crm-form-block` class has been added to the entitlements panel to make inputs display as form controllers.
* Some `.panel-body` or `.row` elements were meant to be `.panel-footer`. All such elements on the admin screens that have been migrated to bootstrap have been changed so they now are panel footers instead.
* CSS rules on `_admin-leave-and-absences-import-form.scss` and `_admin-leave-and-absences-list-base.scss` have been removed because they were made obsolete by [this Shoreditch PR](https://github.com/compucorp/org.civicrm.shoreditch/pull/45).

## Comments
BackstopJS tests were run on admin screens.

---

- [x] Tests Pass


reneolivo:
There's another change:

* Removing `.button` class from "Add new ....."  buttons on list pages.
* Added the class `.btn-icon` to the font awesome icons, which gives a 5px margin between the icon and the button's text.

### Before
download 38
```html
<a href="{crmURL q="action=add&reset=1"}" class="button btn btn-primary">
  <i class="fa fa-plus"></i>
  <span>{ts}Add new entitlement period{/ts}</span>
</a>
```


### After
download 37
```html
<a href="{crmURL q="action=add&reset=1"}" class="btn btn-primary">
  <i class="fa fa-plus btn-icon"></i>
  <span>{ts}Add Work Pattern{/ts}</span>
</a>
```

PR #2170, opened 2017-09-22
PCHR-2535: Disallow transitions to non-cancelling statuses with a negative closing balance

igorpavlov:
## Overview

At the moment is it possible to:

1) Set Leave Request status to More information required with a negative closing balance by Manager
2) Set Leave Request status to Awaiting for approval with a negative closing balance by Staff
3) Approve a Leave Request with a negative closing balance by manager

This PR fixes this issue.

## Technical Details

The problem was in the function that calculates the balance change for **existing** leave requests:

```php
public static function getTotalBalanceChangeForLeaveRequest(LeaveRequest $leaveRequest, $expiredOnly = false) {
  ...
  return $balance; // for Leave Requests using current TOIL balance this value was negative, though the actual calculation was correct
}
```
This lead to the following condition to pass while it should have been failing:

```php
//                                        v That expression will always be false
if(!$absenceType->allow_overuse && $leaveRequestBalance > $currentBalance) {
    'There are only X days leave available...'
  );
}
```
This issue was fixed by simply returning the original (not absolute) value in the balance calculation function for **new** requests:

```php
private static function calculateBalanceChangeFromCreateParams($params) {
  // ...
  return $leaveRequestBalance['amount']; // <-- remove abs()
}
```

## Comments

This PR also fixes trailing spaces.

This PR also fixes the existing test, which ignored the calculation of balance change. The test was still passing, however, if we remove the implementation this test actually checks, it will throw another error (no working days specified), not the error we expect to be thrown (negative closing balance).

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -88,32 +88,32 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     $query = "
       SELECT SUM(leave_balance_change.amount) balance
       FROM {$balanceChangeTable} leave_balance_change
-      LEFT JOIN {$leaveRequestDateTable} leave_request_date 
-             ON leave_balance_change.source_id = leave_request_date.id AND 
+      LEFT JOIN {$leaveRequestDateTable} leave_request_date
+             ON leave_balance_change.source_id = leave_request_date.id AND
                 leave_balance_change.source_type = '". self::SOURCE_LEAVE_REQUEST_DAY ."'
-      LEFT JOIN {$leaveRequestTable} leave_request 
+      LEFT JOIN {$leaveRequestTable} leave_request
              ON leave_request_date.leave_request_id = leave_request.id AND
                 leave_request.is_deleted = 0
-      LEFT JOIN {$contractTable} contract 
+      LEFT JOIN {$contractTable} contract
              ON leave_request.contact_id = contract.contact_id
-      LEFT JOIN {$contractRevisionTable} contract_revision 
+      LEFT JOIN {$contractRevisionTable} contract_revision
             ON contract_revision.id = (
               SELECT id FROM {$contractRevisionTable} contract_revision2
               WHERE contract_revision2.jobcontract_id = contract.id
               ORDER BY contract_revision2.effective_date DESC
               LIMIT 1
             )
-      LEFT JOIN {$contractDetailsTable} contract_details 
+      LEFT JOIN {$contractDetailsTable} contract_details
              ON contract_revision.details_revision_id = contract_details.jobcontract_revision_id
-      
+
       WHERE ((
               $whereLeaveRequestDates AND
               contract.deleted = 0 AND
-              ( 
+              (
                 leave_request.from_date <= contract_details.period_end_date OR
                 contract_details.period_end_date IS NULL
               )  AND
-              ( 
+              (
                 leave_request.to_date >= contract_details.period_start_date OR
                 (leave_request.to_date IS NULL AND leave_request.from_date >= contract_details.period_start_date)
               )  AND
@@ -124,7 +124,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
             )
             OR
             (
-              leave_balance_change.source_id = {$periodEntitlement->id} AND 
+              leave_balance_change.source_id = {$periodEntitlement->id} AND
               leave_balance_change.source_type = '" . self::SOURCE_ENTITLEMENT . "'
             ))
     ";
@@ -272,34 +272,34 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     $query = "
       SELECT SUM(leave_balance_change.amount) balance
       FROM {$balanceChangeTable} leave_balance_change
-      INNER JOIN {$leaveRequestDateTable} leave_request_date 
-              ON leave_balance_change.source_id = leave_request_date.id AND 
+      INNER JOIN {$leaveRequestDateTable} leave_request_date
+              ON leave_balance_change.source_id = leave_request_date.id AND
                  leave_balance_change.source_type = '" . self::SOURCE_LEAVE_REQUEST_DAY . "'
-      INNER JOIN {$leaveRequestTable} leave_request 
+      INNER JOIN {$leaveRequestTable} leave_request
               ON leave_request_date.leave_request_id = leave_request.id AND
                  leave_request.is_deleted = 0
-      INNER JOIN {$contractTable} contract 
+      INNER JOIN {$contractTable} contract
              ON leave_request.contact_id = contract.contact_id
-      INNER JOIN {$contractRevisionTable} contract_revision 
+      INNER JOIN {$contractRevisionTable} contract_revision
             ON contract_revision.id = (
               SELECT id FROM {$contractRevisionTable} contract_revision2
               WHERE contract_revision2.jobcontract_id = contract.id
               ORDER BY contract_revision2.effective_date DESC
               LIMIT 1
             )
-      INNER JOIN {$contractDetailsTable} contract_details 
+      INNER JOIN {$contractDetailsTable} contract_details
              ON contract_revision.details_revision_id = contract_details.jobcontract_revision_id
-             
+
       WHERE {$whereLeaveRequestDates} AND
             contract.deleted = 0 AND
-            ( 
+            (
               leave_request.from_date <= contract_details.period_end_date OR
               contract_details.period_end_date IS NULL
             )  AND
-            ( 
+            (
               leave_request.to_date >= contract_details.period_start_date OR
               (leave_request.to_date IS NULL AND leave_request.from_date >= contract_details.period_start_date)
-            ) AND  
+            ) AND
             leave_balance_change.expired_balance_change_id IS NULL AND
             leave_request.type_id = {$periodEntitlement->type_id} AND
             leave_request.contact_id = {$periodEntitlement->contact_id}
@@ -348,7 +348,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     $query = "
       SELECT bc.*
       FROM {$balanceChangeTable} bc
-      INNER JOIN {$leaveRequestDateTable} lrd 
+      INNER JOIN {$leaveRequestDateTable} lrd
         ON bc.source_id = lrd.id AND bc.source_type = %1
       INNER JOIN {$leaveRequestTable} lr
         ON lrd.leave_request_id = lr.id
@@ -697,16 +697,16 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
           leave_request_date.date,
           leave_request.type_id,
           balance_change.amount
-        FROM {$leaveRequestDateTable} leave_request_date 
-        INNER JOIN {$leaveBalanceChangeTable} balance_change 
+        FROM {$leaveRequestDateTable} leave_request_date
+        INNER JOIN {$leaveBalanceChangeTable} balance_change
             ON balance_change.source_id = leave_request_date.id AND balance_change.source_type = %1
         INNER JOIN {$leaveRequestTable} leave_request
-            ON leave_request_date.leave_request_id = leave_request.id AND 
+            ON leave_request_date.leave_request_id = leave_request.id AND
                (leave_request.request_type IN(%2, %3, %9) AND leave_request.is_deleted = 0)
-        WHERE ({$wherePeriodEntitlementDates}) AND 
+        WHERE ({$wherePeriodEntitlementDates}) AND
               (leave_request_date.date BETWEEN %4 AND %5) AND
-              (leave_request.status_id IN (%6, %10)) AND 
-              (leave_request.contact_id = %7) AND 
+              (leave_request.status_id IN (%6, %10)) AND
+              (leave_request.contact_id = %7) AND
               (leave_request.type_id = %8)
       ";
 
@@ -764,7 +764,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
     $absencePeriodTable     = AbsencePeriod::getTableName();
 
     $query = "
-      SELECT 
+      SELECT
         balance_to_expire.*,
         coalesce(absence_period.start_date, leave_request.from_date) as start_date,
         coalesce(leave_request.contact_id, period_entitlement.contact_id) as contact_id,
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -405,7 +405,7 @@ public static function getTotalBalanceChangeForLeaveRequest(LeaveRequest $leaveR
       }
     }
 
-    return $balance;
+    return abs($balance);
```
tunbola: I think some tests usually require this function to return a negative balance and not positive.
And really the result should be negative because the leave requests cause a negative balance, if you need it in positive it should be in the function where this is returned.

igorpavlov: `getLeaveRequestBalance` returns `calculateBalanceChangeFromCreateParams()` or `getTotalBalanceChangeForLeaveRequest()`.

`calculateBalanceChangeFromCreateParams` returns **abs()**
but 
`getTotalBalanceChangeForLeaveRequest` doesn't.

Is that expected?

tunbola: Yeah its expected. If you check the code `calculateBalanceChangeFromCreateParams` returns an absolute value for the results returned by the `LeaveRequest::calculateBalanceChange` function, You need to wrap the value returned from  `getTotalBalanceChangeForLeaveRequest` in `abs` also as we need the leave balance in positive to compare it against the current entitlement balance.

igorpavlov: Yep, did exactly what you said ^^. Was not sure that's the best solution, but since you suggest the same, I am not much more confident. Thank you.

davialexandre:
@igorpavlov can you try changing the condition [here](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php#L489)

to:

```php
!$absenceType->allow_overuse && ($currentBalance + $leaveRequestBalance) < 0
```

And see it that will fix the problem?

tunbola:
In that case, @davi, IF we change to what you suggested, we need to change `calculateBalanceChangeFromCreateParams` to not return an absolute value anymore.
I actually like this better.

igorpavlov:
@Davi thank you, I will certainly try. However, I have a slight concern, since `$leaveRequestBalance =self::getLeaveRequestBalance($params)` and `getLeaveRequestBalance()` can return negative values as well as positive depending on what is inside params.

davialexandre:
@tunbola good point! And now I understand why we had this `$leaveRequestBalance > $currentBalance in the first place`

So here is an explanation of what happened and probably why we never spotted this problem before:
- On #2149 we changed that validation a little bit. Before that, we used to get the Leave Request balance with the `calculateBalanceChangeFromCreateParams` method, which always return a positive value.
- After #2149 we now use the `getLeaveRequestBalance()` method, but it is a bit inconsistent. For existing Leave Requests it will return a negative value (and the validation will fail) and for new Leave Request it will return a positive value (making the validation pass).

So what @igorpavlov solution was trying to do was to somehow reintroduce the behavior we had before #2149. Unfortunately, it caused some tests to fail.

What this validation is trying to do anyway is to check if the current balance + the leave request balance (that can be either positive or negative) will result in a negative balance. So that's I suggested the condition change. That plus what @tunbola suggested should fix the issue @igorpavlov

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,219 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+      'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE
+    ];
+
+    $this->setExpectedException(
+      'CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException',
+      'There are only '. $entitlementBalance .' days leave available. This request cannot be made or approved'
+    );
+
+    // Create new request by staff with Awaiting Approval status
+    $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation($leaveRequestData, true);
```
davialexandre: It would be better to move this statement before the `setExpectedException()` one, just to make sure that the exception will be thrown by LeaveRequest::create().

This applies to the other tests

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,219 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+    // Approve leave request by manager
+    $updatedLeaveRequestData = $leaveRequestData;
+    $updatedLeaveRequestData['id'] = $leaveRequest->id;
+    $updatedLeaveRequestData['status_id'] = $leaveRequestStatuses['approved'];
+    $updatedLeaveRequest = LeaveRequest::create($updatedLeaveRequestData);
+
+    // Test that the status has not been changed
+    $this->assertEquals($updatedLeaveRequest->status_id, $leaveRequestStatuses['awaiting_approval']);
```
davialexandre: The test stops when it gets the exception it is expecting. This assertion will never be reached.

This applies to the other tests

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,219 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
     }
   }
 
+  public function testManagerCanNotApproveLeaveRequestIfBalanceChangeIsMoreThanEntitlementBalanceWhenAllowOveruseFalse() {
```
davialexandre: The tests you added all mention a manager or a staff, but I don't think this is really necessary. The validation you're trying to test here doesn't care about the user role. At this point, we should only worry about the balances. 

What we need to test here is that an open leave request cannot be updated if its balance change will make the balance negative). In this case an open leave request can be either:
- An Approved Leave Request that was updated to either Awaiting approval or More Information Required
- An More Information Required that was updated to Awaiting Approval or vice versa

igorpavlov: @davialexandre so you mean it should be possible (for example) transit between More Information Required and Awaiting Approval even if the closing balance is negative?

igorpavlov:
@Davi, final check:

@Tunbola wrote:

>In that case, @DAVI, IF we change to what you suggested, we need to change calculateBalanceChangeFromCreateParams to not return an absolute value anymore.
I actually like this better.

Wouldn't that affect the front-end?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: Sure. Just wondering, is such an approach allowed? It allows to test multiple transitions in one test:
```js
$statusTransitions = [
  ['approved', 'more_information_required'],
  ['awaiting_approval', 'approved']
];

foreach ($statusTransitions as $statusTransition) {
  $statusFrom = $statusTransition[0];
  $statusTo = $statusTransition[1];
  $exceptionMessage = '';

  // Balance change is 1 since this is a single day leave request
  $leaveRequestData = [
    'contact_id' => $periodEntitlement->contact_id,
    'status_id' => $statusFrom,
    'from_date' => $requestDate,
    'from_date_type' => $requestDateType,
    'to_date' => $requestDate,
    'to_date_type' => $requestDateType,
  ];

  // Create a new leave request
  $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation($leaveRequestData, true);

  // Attempt to change leave request status
  $updatedLeaveRequestData = $leaveRequestData;
  $updatedLeaveRequestData['id'] = $leaveRequest->id;
  $updatedLeaveRequestData['status_id'] = $statusTo;

  // Attempt to transit the leave request status
  try {
    $updatedLeaveRequest = LeaveRequest::create($updatedLeaveRequestData);
  } catch (Exception $e) {
    $exceptionMessage = $e->getMessage();
  }

  $this->assertEquals($exceptionMessage, 'There 2 are only '. $entitlementBalance .' days leave available. This request cannot be made or approved');
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $periodEntitlement->contact_id],
+      ['period_start_date' => $periodStartDate]
+    );
+
+    $workPattern = WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => 1]);
+    ContactWorkPatternFabricator::fabricate([
+      'contact_id' => $contactId,
+      'pattern_id' => $workPattern->id
+    ]);
```
davialexandre: Since you've fabricated the Work Pattern as the default one, it's not necessary to assign it to the contact using a `ContactWorkPattern`. Default Work Patterns are used whenever a contact doesn't have any patterns.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+
+    $leaveRequestStatuses = array_flip(LeaveRequest::buildOptions('status_id', 'validate'));
```
davialexandre: This:
```php
array_flip(LeaveRequest::buildOptions('status_id', 'validate'));
```

Can be replaced with:

```php
LeaveRequest::getStatuses()
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+
+    // Balance change is 1 since this is a single day leave request
+    $leaveRequestData = [
+      'type_id' => $absenceType->id,
+      'contact_id' => $periodEntitlement->contact_id,
+      'status_id' => $leaveRequestStatuses['approved'],
+      'from_date' => $requestDate,
+      'from_date_type' => $requestDateType,
+      'to_date' => $requestDate,
+      'to_date_type' => $requestDateType,
```
davialexandre: The Leave Request fabricator has some default values for the `from_date_type` and `to_date_type` so you don't need to specify them here.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+    ]);
+
+    $leaveRequestStatuses = array_flip(LeaveRequest::buildOptions('status_id', 'validate'));
+
+    // Balance change is 1 since this is a single day leave request
+    $leaveRequestData = [
+      'type_id' => $absenceType->id,
+      'contact_id' => $periodEntitlement->contact_id,
+      'status_id' => $leaveRequestStatuses['approved'],
+      'from_date' => $requestDate,
```
davialexandre: You're not using the `$requestDate` variable anywhere else in this code, so there's no need to keep this in a separate variable. Please change this to:

```php
'from_date' => CRM_Utils_Date::processDate('2016-06-14'),
```

Please also do the same to `to_date`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+  public function testLeaveRequestCannotBeUpdatedIfBalanceChangeIsMoreThanEntitlementBalanceWhenAllowOveruseFalse() {
+    $contactId = 1;
+    $periodStartDate = CRM_Utils_Date::processDate('2016-01-01');
+    $periodEndDate = CRM_Utils_Date::processDate('2016-12-31');
+    $requestDate = CRM_Utils_Date::processDate('2016-06-14');
+    $requestDateType = $this->leaveRequestDayTypes['all_day']['value'];
+
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => $periodStartDate,
+      'end_date'   => $periodEndDate,
```
davialexandre: You're not using these variables anywhere else. You can just pass in the dates here instead.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -910,6 +910,71 @@ public function testManagerCanCancelOrRejectLeaveRequestEvenIfBalanceIsGreaterTh
...
+    $periodEntitlement = LeavePeriodEntitlementFabricator::fabricate([
+      'type_id' => $absenceType->id,
+      'contact_id' => $contactId,
+      'period_id' => $absencePeriod->id
+    ]);
+
+    // Current balance must be less than the balance change
+    $entitlementBalance = 0;
+    $this->createLeaveBalanceChange($periodEntitlement->id, $entitlementBalance);
+    $periodStartDate = $absencePeriod->start_date;
```
davialexandre: There's no need for this assignment here. You can just pass the `$absencePeriod->start_date` value to the Job Contract fabricator

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: I don't even seem to need to fabricate a work pattern in this case.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: I use this var twice and I think it provides an understanding that this is a single day request:

``php
...
'from_date' => $requestDate,
'from_date_type' => $requestDateType,
'to_date' => $requestDate,
'to_date_type' => $requestDateType,
...
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: I get an exception: 'The from_date_type field should not be empty'

igorpavlov: Investigating...

igorpavlov:
Just squashed my PR Feedbacks into one.

PR #2171, opened 2017-09-22
PCHR-2601: Fix issue with TOIL not being Accrued properly when TOIL dates do not change and Toil to Accrue changes.

tunbola:
## Overview
This [#2103](https://github.com/civicrm/civihr/pull/2103) made some changes to the way Balance changes are updated. Basically, If the dates of a leave request do not change and the `change_balance` parameter is false or not present, the balance change does not get updated. However it fails to factor in TOIL requests because the balance changes are created more differently than for a Leave Request.

## Before
- Balance changes for a TOIL request is equivalent to the value of the `toil_to_accrue` parameter and does not depend on the TOIL request dates. The validation for skipping a balance change update when the dates of a leave request does not change was causing issues in Accruing TOIL especially when the TOIL dates does not change but the Toil to Accrue changes.

## After
- The logic for skipping a balance change update was updated to factor in a TOIL request and its peculiar way of creating its own balance changes.

---

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
@@ -268,9 +268,7 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequest {
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest|NULL
    */
   protected function createRequestWithBalanceChanges($params) {
-    $updateBalanceChange = !empty($params['change_balance']);
-    $skipBalanceChangeUpdate = !empty($params['id']) &&
-      !$this->datesChanged($params) && !$updateBalanceChange;
+    $skipBalanceChangeUpdate = $this->skipBalanceChangeUpdate($params);
     $leaveRequest = LeaveRequest::create($params, LeaveRequest::VALIDATIONS_OFF);
 
     if(!$skipBalanceChangeUpdate) {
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -347,4 +345,22 @@ public function getBreakdown($leaveRequestID) {
...
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function skipBalanceChangeUpdate($params) {
+    $isToilRequest = $params['request_type'] == LeaveRequest::REQUEST_TYPE_TOIL;
+    $datesDidNotChange = $this->datesChanged($params) === false;
+    $leaveRequestDatesNotChanged = !$isToilRequest && $datesDidNotChange;
+    $toilDatesAndToAccrueNotChanged = LeaveRequest::toilToAccrueChanged($params) === false && $datesDidNotChange;
```
davialexandre: Do we really need to care about the dates here?

tunbola: Yes, because if we don't, If the dates did change, the previous dates would have been deleted and new dates re-created Here: https://github.com/civicrm/civihr/blob/de275bfda6e3ce241c288436efe818dffbfd32b9/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php#L806-L806. If the TOIL to accrue did not change, that means the balance change update will be skipped leaving behind new dates with no balance change record.

PR #2173, opened 2017-09-25
PCHR-2664: Add validation for unique Work Pattern labels

tunbola:
## Overview
When trying to create a Work Pattern with same label as one already existing in the database, An db error is shown on the screen.

## Before
localhost-8954-civicrm-admin-leaveandabsences-work_patterns 2017-09-25 16-21-53 1
## After
- When trying to create a Work Pattern with same label as one already existing in the database, rather than a Db error, A user friendly error is shown on the Work Pattern listing page
workpattern

- [X] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
@@ -156,18 +156,21 @@ class CRM_HRLeaveAndAbsences_BAO_WorkPattern extends CRM_HRLeaveAndAbsences_DAO_
    */
   private static function validateWorkPatternTitle($params) {
     $label = CRM_Utils_Array::value('label', $params);
-    $workPattern = new self();
-    $workPattern->label = $label;
-    $workPattern->find(true);
-
-    if (!$workPattern->id) {
+    if (!$label) {
       return;
     }
 
-    $isCreate = empty($params['id']);
-    $isSeparateUpdate = !empty($params['id']) && $workPattern->id != $params['id'];
+    $workPattern = new self();
+    $workPattern->label = $label;
 
-    if ($isCreate || $isSeparateUpdate) {
+    if (!empty($params['id'])) {
+      $id = (int) $params['id'];
+      $workPattern->whereAdd("id <> $id");
+    }
+
+    $workPattern->find(true);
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -145,6 +147,34 @@ public static function validateDefaultWorkPattern($params) {
...
+    }
+
+    $isCreate = empty($params['id']);
+    $isSeparateUpdate = !empty($params['id']) && $workPattern->id != $params['id'];
+
+    if ($isCreate || $isSeparateUpdate) {
+      throw new InvalidWorkPatternException(
+        'Work Pattern with same label already exists!'
+      );
+    }
```
davialexandre: What about this?

```php
    $label = CRM_Utils_Array::value('label', $params);
    if (!$label) {
        return;
    }

    $workPattern = new self();
    $workPattern->label = $label;

    if (!empty($params['id'])) {
       $id = (int) $params['id'];
       $workPattern->whereAdd('id <> $id');
    }

    $workPattern->find(true);

    if ($workPattern->id) {
      throw new InvalidWorkPatternException(
        'Work Pattern with same label already exists!'
      );
    }
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/WorkPatternTest.php
```diff
@@ -698,4 +689,49 @@ public function testCannotDisableTheDefaultWorkPattern() {
...
+    $workPattern = WorkPatternFabricator::fabricate($params);
+
+    //update the work pattern
+    $params['id'] = $workPattern->id;
+    $params['description'] = 'This is a cool Work Pattern';
+
+    try{
+      $workPattern = WorkPatternFabricator::fabricate($params);
+      $this->assertEquals($workPattern->description, $params['description']);
+    }catch(CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException $e) {
```
davialexandre: missing space before the try and catch curly braces

davialexandre:
@tunbola the PR title is a bit too long. What about something like "Add validation for unique Work Pattern labels"?

PR #2178, opened 2017-09-26
PCHR-2545: Fix Work Pattern issues

reneolivo:
## Overview
This PR fixes the following issues on the Work Pattern admin form:

* It changes the wording of the validation messages from `Please inform the [field name]` to `Please fill in the [field name]`.
* Stopped select inputs from turning into select2 components by applying the fix introduced in this PR:
https://github.com/compucorp/org.civicrm.shoreditch/pull/52

### Before
download 51

## After
download 50

## Select2
Please check https://github.com/compucorp/org.civicrm.shoreditch/pull/52 for before and after screens of the fix for this issue.

## Technical Details
* Validation messages were updated in these files:
  - `CRM/HRLeaveAndAbsences/Form/WorkPattern.php`
  - `tests/phpunit/WebTest/WorkPattern/FormTest.php`

  The last one is a unit test that expects these messages to appear when the validation fails.

* The `no-select2` class was added only to the `Form/WorkPattern/Calendar.tpl` file so it doesn't affect other selects outside of the calendar form.

---

- [x] Tests Pass


igorpavlov:
I think comments section is not needed. PHP tests passed for me

image

PR #2180, opened 2017-09-26
PCHR-2545: Fix dropdown button issue

reneolivo:
## Overview
This PR fixes two issues regarding Dropdown menus  (`.btn-slide` and/or`.crm-hover-button`):

1. Dropdown menus would stop working after enabling or disabling the record they were part. This happens on list pages. This PR fixes that issue.
2. Dropdown options would not fire any events when clicked (ie: actions such as *set as default* would not work on dropdown buttons)

### Before
anim
The dropdown menu works until the record is disabled, then it doesn't work any longer.

### After
anim

### Technical Details
The issue happened because when a record is enable or disabled, the whole table row is removed and replaced by a different one which doesn't have the events attached that can open the modal again.

The issue was fixed by adding a delegated click event, since the `attrchange` plugin does not support event delegation. `attrchange` is still used for closing the panel since CiviCRM is the handler for the closing event by adding or removing the `.btn-slide-active` class to the element.

### Before
anim

### After
anim
(the Delete action works the same)

### Technical details
This issue was happening because the action events are added as delegated events ([source](https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/js/crm/hrleaveandabsences.js#L40)) and the delegated root element was not the body element, which is the parent of the popup panel when it's in an open state. A mapping function was added that would forward click events to their source element ([source](https://github.com/civicrm/civihr/pull/2180/files#diff-5aa2eb87442af5119577aea46f98fadcR84)).

Why not just display the popup modal instead of cloning it and attaching it to the body? This file introduced this fix since dropdown menus were being hidden by the `overflow: hidden` rule of the parent container.

## Comments
This file is not covered by any unit test framework. Not sure how to approach testing aside from doing manual tests.

---

- [ ] Tests Pass


igorpavlov:
Is that a regression issue or is it something that always exist?

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
 (function ($) {
   'use strict';
 
-  // .btn-slide is a continer of popup and a button triggering it
-  $('.btn-slide').attrchange(function (e) {
-    var $body = $('body');
+  $('body').on('click', '.btn-slide', function (e) {
     var $button = $(this);
-    var $closestEntity, $popup, buttonOffset;
+    var $body = $('body');
```
igorpavlov: I think this can go outside this function and then you can also do `$body.on('click'...`

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
...
-    if ($button.hasClass('btn-slide-active')) {
+    /**
+     * Listens to mouse events happening outside the of the panel in order to
+     * close it. The way to detected mouse out is by listening to the
+     * .btn-slide-active class change, which is added and removed by CiviCRM.
+     * the data *attrchange-is-on* is set to true in order to avoid adding
+     * multiple listeners to the same element.
+     */
+    function listenToMouseOutEvent () {
+      // AttrChange event is already listened, skip:
```
igorpavlov:
```js
// If AttrChange event is already listened, then skip:
```

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
...
+      if ($button.data('attrchange-is-on')) {
+        return;
+      }
+
+      $button.data('attrchange-is-on', true);
+
```
igorpavlov: Please remove this empty line

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
...
+      $button.attrchange(function () {
+        // Button is already open, skip:
```
igorpavlov:
```js
// If button is already open, then skip:
```

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
...
+        if ($button.hasClass('btn-slide-active')) {
+          return;
+        }
+
+        closePopupPanel();
+      });
+    }
+
+    /**
```
igorpavlov: I think it would be easier to read if you break the sentence like that:
```js
/**
 * Opens the popup panel
 * and adds the .civihr-popup-open class to the body
 */
```

Please note the change from "pop" to "popup" as well (for consistency)

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -25,17 +72,10 @@ Prevent Popups to overflow tables.
       buttonOffset = $button.offset();
 
       $popup
-        .css('top', parseInt(buttonOffset.top, 10) + $button.outerHeight())
-        .css('left', parseInt(buttonOffset.left, 10) - ($popup.width() - $button.outerWidth()));
+        .css('top', +buttonOffset.top + $button.outerHeight())
+        .css('left', +buttonOffset.left - $popup.width() - $button.outerWidth());
```
igorpavlov: I can notice brackets in the previous version of the code and the operation here is subtraction:
```js
($popup.width() - $button.outerWidth())
```
But in the current change these brackets were removed. I think that actually breaks the formula

image

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
 (function ($) {
   'use strict';
 
-  // .btn-slide is a continer of popup and a button triggering it
-  $('.btn-slide').attrchange(function (e) {
-    var $body = $('body');
+  $('body').on('click', '.btn-slide', function (e) {
```
igorpavlov: Can we use `event` or `ev` instead of `e` since `e` can also mean "error"?

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -4,14 +4,61 @@ Prevent Popups to overflow tables.
...
+
+    e.preventDefault();
+
+    (function init () {
+      openPopupPanel();
+      listenToMouseOutEvent();
+    })();
+
+    /**
+     * Closes the popup panel and removes the .civihr-popup-open class from the
```
igorpavlov: I think it would be easier to read if you break the sentence like that:
```js
/**
 * Closes the popup panel
 * removes the .civihr-popup-open class from the body
 */
```

hrui/js/src/civihr-popup/civihr-popup.js
reneolivo: yes, that's a good observation, though this variable is not being used, so I'll remove it.

reneolivo:
@igorpavlov the popup not showing up was a regression issue that started happening after the Bootstrap migration.

For the click issue I ran `git bisect` on a previous commit before the Bootstrap migration on both CiviHR and Shoreditch and the issue was present in a different way: it would only work after disabling/enabling the whole row for the first time.

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -5,15 +5,13 @@ Prevent Popups to overflow tables.
 */
 (function ($) {
   'use strict';
+  var $body = $('body');
 
-  $('body').on('click', '.btn-slide', function (e) {
+  $('body').on('click', '.btn-slide', function () {
```
igorpavlov: Just `$body.on...` would be enough

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -5,15 +5,13 @@ Prevent Popups to overflow tables.
...
     var $button = $(this);
-    var $body = $('body');
     var $closestEntity = $button.closest('.crm-entity')[0];
     var $popup = $button.children('ul.panel');
 
-    // e.preventDefault();
```
igorpavlov: Not sure how I didn't notice that this was commented, probably was looking in a different commit. I wonder if we should call e.stopPropagation() in case buttons appear inside other blocks with click events?

hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -42,10 +40,10 @@ Prevent Popups to overflow tables.
...
 
-      $clone
-        .css('top', +buttonOffset.top + $button.outerHeight())
-        .css('left', +buttonOffset.left - $clone.width() - $button.outerWidth());
-
+      $clone.css({
+        left: +buttonOffset.left - ($clone.width() - $button.outerWidth()),
+        top: +buttonOffset.top + $button.outerHeight()
+      });
       $clone.show();
```
igorpavlov: No empty line needed. There are also other places in the code where there is an empty line between function calls, could you check please (like here https://github.com/civicrm/civihr/pull/2180/commits/a88eb67df7dd48cd34e94a4daaaf05ecba99a3b7#diff-5aa2eb87442af5119577aea46f98fadcR98)?

PR #2179, opened 2017-09-26
PCHR-2663: Show original balance change in the Leave Request Modal

igorpavlov:
## Overview

At the moment, when the work pattern has been changed and it affects a current Leave Request, the Leave Request Report shows original balance change (before work pattern change), but the Leave Request Modal shows the new re-calculated breakdown. Instead, the Leave Request Modal should show the original balance change breakdown, until re-saved with new dates.

## Technical Details

1. In Leave Request API a new method was created to match the endpoint as per https://github.com/civicrm/civihr/pull/2120

```js
getBalanceChangeBreakdown: function (leaveRequestId) { ... }
```

2. In Leave Request Instance a new `getBalanceChangeBreakdown()` method was created which returns the balance change breakdown with the same structure as the `calculatebalancechange` API endpoint.

3. In Leave Request Details component `calculateBalanceChange` function now accepts a boolean parameter that affects the way the balance breakdown is calculated: original one is retrieved or re-calculated one.

## Comments

1) Renamed all models and APIs from "-api.js" and "-model.js" to ".api.js" and ".model.js"
2) Ran semistandard over Models and API files to fix lint issues and indentation

---

- [x] Tests Pass

PR #2183, opened 2017-09-28
PCHR-2681: Notify leave approver about balance recalculation

igorpavlov:
## Overview

At the moment if there is an existing leave request and the current work patten has changed affecting that leave request, saving such leave request by admin / manager without changing its dates will not update the balance change of this leave request. Please read the description to this PR https://github.com/civicrm/civihr/pull/2179 for more info.

This PR, however:

1) Notifies admin / manager if the currently displayed balance change has been changed due to the work pattern change and needs to be updated.
2) Forces the balance change to be recalculated when admin or manager save a request

## Technical Details

In master request controller (request.controller.js) the submit() function now has new promises in the chain:

```js
function submit () {
  // ...
  return vm.request.isValid()
    .then(checkIfBalanceChangeHasChanged) // <-- Checks if balance change has changed
    .then(decideIfBalanceChangeNeedsAForceRecalculation) // <-- Forces recalculation
    .then(function () {
      // ... create/update
    }) // ...
```
`checkIfBalanceChangeHasChanged()` compares the original balance change and the calculated one and calls a dialog if they are not equal.

## Comments

This PR also cleans up the touched controller file.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -177,6 +178,25 @@ define([
...
 
     /**
+     * Checks if the balance change has changed since this leave request was saved
+     *
+     * @return {Promise}
+     */
+    function checkIfBalanceChangeHasChanged () {
+      if (!vm.isMode('edit') || vm.isRole('staff')) { return; }
+
+      return LeaveRequest.calculateBalanceChange(_.pick(vm.request,
```
reneolivo: Please move `LeaveRequest.calculateBalanceChange` to the request instance ([around here](https://github.com/civicrm/civihr/blob/338fd5f1261227b766ef6ac8483289d59121250a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/leave-request.instance.js#L230)).

The code should be updated to something like `vm.request.calculateBalanceChange().then(...)`.

The leave request popup tab component should also be updated too [around here](https://github.com/civicrm/civihr/blob/338fd5f1261227b766ef6ac8483289d59121250a/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js#L151).

The reason for this is to make it easier to calculate the balance change for the request without the need to know which parameters to pass.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request.controller.spec.js
```diff
@@ -450,24 +460,67 @@
             controller.checkSubmitConditions.and.returnValue(true);
 
             controller.submit();
-            $scope.$apply();
           });
 
-          it('allows user to submit', function () {
-            expect(controller.canSubmit()).toBeTruthy();
-          });
+          describe('if balance change has not been changed', function () {
```
reneolivo: has not been *updated*

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request.controller.spec.js
```diff
@@ -450,24 +460,67 @@
...
-            expect(LeaveRequestAPI.isValid).toHaveBeenCalled();
-            expect(LeaveRequestAPI.update).toHaveBeenCalled();
+            it('sends update event', function () {
+              expect($rootScope.$emit).toHaveBeenCalledWith('LeaveRequest::updatedByManager', controller.request);
+            });
           });
 
-          it('sends update event', function () {
-            expect($rootScope.$emit).toHaveBeenCalledWith('LeaveRequest::updatedByManager', controller.request);
+          describe('if balance change has been changed', function () {
```
reneolivo: has been *updated*

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
igorpavlov: Awesome idea.

igorpavlov: I `grep`ped calculateBalanceChange and it seems the model method was only used in request.controller.js and request-details-tab.component.js - so all good!

PR #2184, opened 2017-09-28
PCHR-2609: Update main menu css

AkA84:
## Overview
This PR contains work related to the top main menu restyling done in https://github.com/compucorp/org.civicrm.shoreditch/pull/53, specifically work done in the overhaul of the CiviHR logo.

### `.chr_logo` component
The component amends the currently used custom [icomoon](icomoon.io) font to provide a more reusable and customizable logo

#### Basic version
logo-basic

```html
<div style="font-size: 25px; color: orange;">
  <span class="chr_logo"></span>
</div>
```
The simplest version of the logo contains just the icon, and takes the font size and color from the parent component

#### Modifier: with title
logo-basic-title

```html
<div style="font-size: 25px; color: orange;">
  <span class="chr_logo chr_logo--full"><i></i></span>
</div>
```
The "CiviHR" title is displayed near the icon (notice the `<i>` element, necessary for the title to appear)

#### Modifier: with default colors
logo-default-color

```html
<div style="font-size: 25px; color: orange;">
  <span class="chr_logo chr_logo--full chr_logo--default-color"><i></i></span>
</div>
```
It uses the official default colors, regardless of the font color of the parent element

#### Modifier: with default size
logo-default-size

```html
<div style="font-size: 25px; color: orange;">
  <span class="chr_logo chr_logo--full chr_logo--default-size"><i></i></span>
</div>
```
It uses the default size, regardless of the font size of the parent element

### `.___outside-namespace` marker
The entire `org.civicrm.bootstrapcivihr` css (`civihr.css`) is being namespace via gulp (the `#bootstrap-theme` id is pre-pended to every selector via postCSS in the post-processed css).

The problem with that that the `.chr_logo` component (and any style related to it, see the [`_common.scss`](https://github.com/civicrm/civihr/blob/5526c3eec1a0894c262608acbbec961c9af4ff72/org.civicrm.bootstrapcivihr/scss/pages/_common.scss) partial) are implemented in parts of the page outside of the namespace.

To solve the problem, a special class/marker was introduced: `.___outside-namespace`

The class is used as a wrapper to the style that we *do not* want to be namespaced:
```scss
.___outside-namespace {
  .crm-link-home {
    position: relative;
    top: 2px;

    .chr_logo {
      font-size: 17px;
    }
  }
}
```
The `sass` gulp task will then recognize the marker and remove it
```js
var outsideNamespaceRegExp = /^\.___outside-namespace/;
// ...
gulp.task('sass', ['sass:sync'], function () {
  // ...
  .pipe(transformSelectors(removeOutsideNamespaceMarker, { splitOnCommas: true }))
});

// ...
function removeOutsideNamespaceMarker (selector) {
  return outsideNamespaceRegExp.test(selector) ? selector.replace(outsideNamespaceRegExp, '') : selector;
}
```

resuling in the style being outside of the namespace:
```css
.crm-link-home {
  position: relative;
  top: 2px;
}

.crm-link-home .chr_logo {
    font-size: 17px;
}
```
### Consolidated logo scss partials in bootstrapcivihr
The hrui extensions had the style and fonts for the logo. All of it was removed, as the logo style and fonts are now in bootstrapcivihr

## Comments
There is a duplication of both the namespace (`#bootstrap-namespace`) and the outside-namespace marker (`.___outside-namespace`) in the scss files and in the gulpfile. Basically they have to be defined twice.

We should find a way to import SASS variables in the gulpfile, so that we can defined those variables only once


org.civicrm.bootstrapcivihr/scss/components/_logo.scss
```diff
@@ -0,0 +1,80 @@
...
+    url('../fonts/icomoon.woff?tozcjz') format('woff'),
+    url('../fonts/icomoon.svg?tozcjz#icomoon') format('svg');
+  font-weight: normal;
+  font-style: normal;
+}
+
+#{$outsideNamespaceMarker} {
+  .chr_logo {
+    display: inline-block;
+    font-family: 'icomoon' !important;
```
reneolivo: why `!important`? is there a chance that this font would be overwritten by a parent rule?

deb1990:
@AkA84 Looks good to me, this will be really usefull.

igorpavlov:
> We should find a way to import SASS variables in the gulpfile, so that we can defined those variables only once

https://www.npmjs.com/package/gulp-sass-variables ?

org.civicrm.bootstrapcivihr/scss/components/_logo.scss
```diff
@@ -0,0 +1,80 @@
...
+      bottom: .2em;
+      color: inherit;
+      content: "\e902";
+      font-size: .7em;
+      font-style: normal;
+      margin-left: 2.4em;
+      position: relative;
+    }
+  }
+}
```
reneolivo: Missing line at the end of the file

org.civicrm.bootstrapcivihr/scss/pages/_common.scss
```diff
@@ -0,0 +1,10 @@
+#{$outsideNamespaceMarker} {
+  .crm-link-home {
+    position: relative;
+    top: 2px;
+
+    .chr_logo {
+      font-size: 17px;
+    }
+  }
+}
```
reneolivo: Missing line at the end of the file

hrui/css/hrui.css
```diff
@@ -2,50 +2,8 @@
...
-
-.icon-logo-full:before {
-  color: #727f8a;
-  font-size: 22px;
-}
-
-body:not(.toolbar-drawer) {
-  padding-top: 30px;
+body {
+  padding-top: 60px !important;
```
reneolivo: Why `!important` ? I see a commit [PCHR-2609: Enforce new padding to body in all circumstances](https://github.com/civicrm/civihr/pull/2184/commits/1aa20f0df979ee240a29d551de161b26cc4e7434), but what are the circumstances that would overwrite the padding-top of the page? Can't they be fixed?

org.civicrm.bootstrapcivihr/scss/components/_logo.scss
AkA84: This comes from the css generated by icomoon, they suggest keeping the `!important`

hrui/css/hrui.css
AkA84: Drupal applies a top padding to the body via javascript, that's why we have to override it via css with an `!important`

PR #2187, opened 2017-10-03
PCHR-2695: Add calculation_unit to Absence Type Table

tunbola:
## Overview
This PR is a follow up to [#2185](https://github.com/civicrm/civihr/pull/2185). It adds the calculation_unit column to the Absence type table and the column references option values for the the previously created `hrleaveandabsences_absence_type_calculation_unit` option group. Also upgraders to add the calculation_unit column and also the calculation_unit option group are added for sites that already have L&A installed.

## Before
- The calculation_unit column was not present in the Absence Type table.

## After
- The calculation unit_column was added to the Absence Type table with the system created absence types having the days option as the default calculation unit.
- An upgrader was added to create the calculation_unit column for sites that already have L&A installed and previously created Absence types are updated to have the days option as the default calculation unit.
- An upgrader was added to create the calculation_unit option group for sites that already have L&A installed.

---

This change does not affect any test.
- [ ] Tests Pass


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1005.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1005 {
```
davialexandre: Instead of duplicating the data we have on the xml file and creating the option values manually, I think it would be better to simply import the xml file here. Please have a look on https://github.com/civicrm/civihr/blob/eee9c831c95cfa04fe2eec5c0321345424e5c1be/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader.php#L33-L46 and check if we can do something similar to what is done during the installation

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1006.php
```diff
@@ -0,0 +1,31 @@
...
+
+    if(!SchemaHandler::checkIfFieldExists($absenceTypeTable, 'calculation_unit')) {
+      CRM_Core_DAO::executeQuery("
+        ALTER TABLE {$absenceTypeTable} 
+        ADD calculation_unit varchar(512) NOT NULL   
+        COMMENT 'One of the values of the Absence type calculation units option group'");
+
+      CRM_Core_DAO::executeQuery("
+        UPDATE {$absenceTypeTable}
+        SET calculation_unit = 1 WHERE calculation_unit = ''");
```
davialexandre: a small readability improvement: We already have the option group available at this point, so, instead of hardcoding the days value here, how about something like this:

```
$daysUnitValue = get_value_from_option_group();
...
SET calculation_unit = {$daysUnitValue}
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1005.php
tunbola: Good suggestion.

PR #2188, opened 2017-10-03
PCHR-2613: Customize quicksearch field

AkA84:
## Overview
This PR introduces customizations to the quicksearch field in the top right corner of a CiviHR site

## Before
before-menu

The field was just a static field, with a "Contacts" placeholder

## After
after-menu

The field is now completely collapsed until the user hovers overs it, and stays expanded if there is an ongoing search. The placeholder text is changed to "Quick Search"


hrui/js/src/hrui.js
```diff
@@ -171,4 +173,25 @@
...
+   * The class is removed only when the element
+   * loses the hover AND it is empty (= there is no ongoing search)
+   */
+  function addCustomClassToQuickSearchField () {
+    $('#crm-qsearch').hover(
+      function () {
+        $(this).addClass('search-ongoing');
+      },
+      function () {
+        if (!$('#sort_name_navigation').val().length &&
```
deb1990: Creating some variables here would make the code more readable.
```js
var onGoingSearchResultItems = $('.crm-quickSearchField:visible', '#root-menu-div').length;
// same goes for $('#sort_name_navigation').val().length

File: hrui/js/src/hrui.js [hunk 2]
```diff
@@ -171,4 +173,25 @@
...
   });
+
+  /**
+   * Adds a custom class to the quicksearch field
+   * so that custom behaviour can be applied to it
+   *
+   * The class is removed only when the element
+   * loses the hover AND it is empty (= there is no ongoing search)
+   */
+  function addCustomClassToQuickSearchField () {
```
deb1990: can we name it `toggleHoverClassofQuickSearchField`?

hrui/js/src/hrui.js
```diff
@@ -91,6 +91,21 @@
   }
 
   /**
+   * Changes the placeholder text of the quicksearch field
+   */
+  function changeQuickSearchFieldPlaceholder () {
+    $('#crm-qsearch .ui-autocomplete-input').attr('placeholder', 'Quick Search');
```
deb1990: Can we do it from the markup?

hrui/js/src/hrui.js
```diff
@@ -1,11 +1,115 @@
+/* global Ps */
+
 // Copyright CiviCRM LLC 2013. See http://civicrm.org/licensing
 (function ($, _) {
+  $('.CRM_HRRecruitment_Form_Application').addClass('crm-form-block');
```
deb1990: This file is a mess right now, with no proper structure to the code. Can we clean this up a bit.
```js
(function init () {
  $(document).on('crmLoad', crmOnLoad);
  $(document).ready(onReady);
})();

//Keep all the private functions below
// Also can we break crmOnLoad into smaller chunks

hrui/js/src/hrui.js
AkA84: no, the markup is core markup, we can't change it

igorpavlov:
Has this solution been tested on mobile devices with touch events? If you need help testing on iOS please let me know.

igorpavlov:
The code looks fine.

igorpavlov:
It works like this on touch devices:
1) you tap - it expands
2) you tap again - it focuses
3) you enter the text with a virtual keyboard, hit virtual enter and it searches
4) when you tap anywhere but on the search object, it collapses

cc @AkA84

PR #2197, opened 2017-10-06
PCHR-2696: Changes to the Absence Type Form for Leave in Hours Requirements

tunbola:
## Overview
This PR makes some changes to the Absence Type Form by adding the `Calculate Leave in field` which allows the user to select if Leave for the Absence Type is to be calculated in Hours or Days. It also implements some Form logic for the new changes.

## Before
- The `Calculate Leave in` field was not present in the Absence Type form

## After
- The `Calculate Leave in field` was added to the Absence Type form. This field is a select field that uses the option values in the `absence_type_calculation_unit` option group as its select options.
- For an Absence Type that is linked to an existing leave request, the `Calculate Leave in` field cannot be changed.
- If Leave is in hours, Public Holiday cannot be added to the entitlement.
- References to days in the form labels were removed

absencetypeform2

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -304,7 +304,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
 
       if($absenceTypeHasBeenUsed) {
         throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
-          'The Calculation unit cannot be change because the Absence Type is In Use!'
+          'The Calculation unit cannot be changed because the Absence Type is In Use!'
         );
       }
     }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -2,6 +2,7 @@
 
 use CRM_HRLeaveAndAbsences_Queue_PublicHolidayLeaveRequestUpdates as PublicHolidayLeaveRequestUpdatesQueue;
 use CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException as InvalidAbsenceTypeException;
+use CRM_HRLeaveAndAbsences_Service_AbsenceType as AbsenceTypeService;
 
 class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_AbsenceType {
 
@@ -244,6 +245,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
     self::validateAbsenceTypeTitle($params);
     self::validateTOIL($params);
     self::validateCarryForward($params);
+    self::validateCalculationUnit($params);
   }
 
   /**
@@ -273,6 +275,38 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
       throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
         'You cannot add public holiday to entitlement when Absence Type calculation unit is in Hours'
       );
+    }
+  }
+
+  /**
+   * Validates that the calculation_unit column cannot be changed once the
+   * Absence Type is in use.
+   *
+   * @param array $params
+   *  The params array received by the create method
+   *
+   * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException
+   */
+  private static function validateCalculationUnit($params) {
+    if(empty($params['id'])) {
+      return;
+    }
+
+    if(empty($params['calculation_unit'])) {
+      return;
+    }
+
+    $oldValue = self::getFieldValue(self::class, $params['id'], 'calculation_unit');
+    $calculationUnitChanged = $oldValue != $params['calculation_unit'];
+    if($calculationUnitChanged) {
+      $absenceTypeService = new AbsenceTypeService();
+      $absenceTypeHasBeenUsed = $absenceTypeService->absenceTypeHasEverBeenUsed($params['id']);
+
+      if($absenceTypeHasBeenUsed) {
+        throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
+          'The Calculation unit cannot be change because the Absence Type is In Use!'
+        );
+      }
     }
   }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -252,6 +252,8 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    * There can be only one AbsenceType where this field is true. So this
    * method checks if one such type already exists and throws an error if that
    * is the case.
+   * It also checks to ensure that public holiday cannot be added to entitlement
+   * if the absence type leave is to be calculated in hours.
    *
    * @param array $params
    *  The params array received by the create method
@@ -262,6 +264,14 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
     if(!self::fieldIsUnique('add_public_holiday_to_entitlement', 1, $params)) {
       throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
         'There is already one Absence Type where public holidays should be added to it'
+      );
+    }
+
+    $calculationUnitOptions = array_flip(self::buildOptions('calculation_unit', 'validate'));
+    $calculationUnitIsInHours = $params['calculation_unit'] == $calculationUnitOptions['hours'];
+    if($params['add_public_holiday_to_entitlement'] == 1 && $calculationUnitIsInHours) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
+        'You cannot add public holiday to entitlement when Absence Type calculation unit is in Hours'
       );
     }
   }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
@@ -207,7 +207,7 @@ class CRM_HRLeaveAndAbsences_Form_AbsenceType extends CRM_Core_Form {
     $this->add(
       'checkbox',
       'allow_accruals_request',
-      ts('Allow staff to request to accrue additional days leave of this type during the period')
+      ts('Allow staff to request to accrue additional leave of this type during the period')
     );
     $this->add(
       'text',
@@ -243,7 +243,7 @@ class CRM_HRLeaveAndAbsences_Form_AbsenceType extends CRM_Core_Form {
     $this->add(
       'text',
       'max_number_of_days_to_carry_forward',
-      ts('Maximum number of days that can be carried forward to a new period? (Leave blank for unlimited)'),
+      ts('Maximum amount of leave that can be carried forward to a new period? (Leave blank for unlimited)'),
       $this->getDAOFieldAttributes('max_number_of_days_to_carry_forward')
     );
     $this->add(
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
@@ -1,6 +1,7 @@
 <?php
 
 use CRM_HRLeaveAndAbsences_Service_AbsenceType as AbsenceTypeService;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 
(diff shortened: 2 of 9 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -264,6 +268,46 @@ private static function validateAddPublicHolidayToEntitlement($params) {
...
+
+    $oldValue = self::getFieldValue(self::class, $params['id'], 'calculation_unit');
+    $calculationUnitChanged = $oldValue != $params['calculation_unit'];
+    if($calculationUnitChanged) {
+      $absenceTypeService = new AbsenceTypeService();
+      $absenceTypeHasBeenUsed = $absenceTypeService->absenceTypeHasEverBeenUsed($params['id']);
+
+      if($absenceTypeHasBeenUsed) {
+        throw new CRM_HRLeaveAndAbsences_Exception_InvalidAbsenceTypeException(
+          'The Calculation unit cannot be change because the Absence Type is In Use!'
```
davialexandre: s/change/changed/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
```diff
@@ -140,6 +143,15 @@ private function addBasicDetailsFields() {
         ['disabled' => 'disabled']
       );
     }
+    $this->addSelect(
+      'calculation_unit',
+      [
+        'options' => $this->getCalculationUnitOptions(),
+        'option_url' => NULL,
```
davialexandre: Is it really necessary to set the option_url to null?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
```diff
@@ -140,6 +143,15 @@ private function addBasicDetailsFields() {
...
+        'label' => 'Calculate Leave in'
```
davialexandre: please wrap the label with a `ts()`

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -270,11 +274,58 @@
...
+                  var prev_calculation_unit_val = $('.absence-calculation-unit select option:selected').val();
+                  $('.absence-calculation-unit select').on('change', function(e) {
+                    var add_public_holiday_entitlement = $('.public-holiday-option input[type=radio]:checked').val();
+                    var hours_unit_value = getHoursUnitValue();
+                    if(this.value == hours_unit_value && add_public_holiday_entitlement == 1) {
+                      $('.absence-calculation-unit select').select2('val', prev_calculation_unit_val);
+                      showPublicHolidayEntitlementErrorAlert();
+                    }
+                    else{
+                      prev_calculation_unit_val = $('.absence-calculation-unit select option:selected').val();
```
davialexandre: I don't understand why the else is necessary. Would you please give more details?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -270,11 +274,58 @@
...
+                  setEntitlementLabelText();
+                  var prev_calculation_unit_val = $('.absence-calculation-unit select option:selected').val();
+                  $('.absence-calculation-unit select').on('change', function(e) {
+                    var add_public_holiday_entitlement = $('.public-holiday-option input[type=radio]:checked').val();
+                    var hours_unit_value = getHoursUnitValue();
+                    if(this.value == hours_unit_value && add_public_holiday_entitlement == 1) {
+                      $('.absence-calculation-unit select').select2('val', prev_calculation_unit_val);
+                      showPublicHolidayEntitlementErrorAlert();
+                    }
+                    else{
```
davialexandre: Missing space before the {

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -270,11 +274,58 @@
...
+                    }
+                    setEntitlementLabelText();
+                  });
+                }
+
+                function setEntitlementLabelText() {
+                  $('.entitlement-label').text($('.absence-calculation-unit select option:selected').text());
+                }
+
+                function initPublicHolidayControls() {
```
davialexandre: Instead of all this complex logic, what about simply disabling the public holiday checkbox when the calculation unit is Hours?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -270,11 +274,58 @@
...
+                  return hours_unit_value;
+                }
+
                   $(document).ready(function() {
                       initToilControls();
                       initCarryForwardControls();
                       initColorPicker();
                       initDeleteButton();
+                      initCalculationUnitControls()
+                      initPublicHolidayControls()
```
davialexandre: It looks like that some of the old code here is still using 4 spaces instead of 2 for indentation. Can you please fix that as well?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -793,4 +798,50 @@ public function testExceptionIsThrownWhenUpdatingAnAbsenceTypeWithTitleOfAnother
...
+      'calculation_unit' => $this->calculationUnitOptions['hours']
+    ]);
+
+    LeaveRequestFabricator::fabricateWithoutValidation([
+      'type_id' => $absenceType->id,
+      'contact_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-01-02'),
+      'to_date' => CRM_Utils_Date::processDate('2016-01-02'),
+    ]);
+
```
davialexandre: How about using the setExpectedException() method here instead of the annotation?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -648,11 +648,14 @@ private function createAbsencePeriod($startDate, $endDate) {
   }
 
   private function createAbsenceType() {
+    $calculationUnitOptions = array_flip(AbsenceType::buildOptions('calculation_unit', 'validate'));
```
davialexandre: Can't we replace this method with the Absence Type fabricator?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/NotificationReceiverTest.php
```diff
@@ -41,11 +41,13 @@ public function testShouldDeleteReceiversForAbsenceType()
 
   private function instantiateAbsenceType()
   {
+    $calculationUnitOptions = array_flip(AbsenceType::buildOptions('calculation_unit', 'validate'));
     $this->absenceType = AbsenceType::create([
```
davialexandre: And perhaps this can also be replaced by the fabricator?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/EntitlementCalculatorTest.php
```diff
@@ -51,11 +51,13 @@ public function testCanOnlyReturnCalculationsForEnabledAbsenceTypes()
   }
 
   private function createBasicType($params = array()) {
+    $calculationUnitOptions = array_flip(AbsenceType::buildOptions('calculation_unit', 'validate'));
```
davialexandre: And replace this one too?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
tunbola: Well, If it's not set to Null, it adds a link to edit/create option values for the calculation_unit  in front of the Calculate Leave in field.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
tunbola: The `prev_calculation_unit_val` variable needs to be updated anytime the `Calculate Leave ` select field changes, even if its an allowed/valid change. This is to ensure that when the change is not valid (e.g selecting Hours option when add public holiday entitlement is yes), The field can always be set back to what it was before the user made the change.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
tunbola: There are different scenarios to this, E.g user  selects `Days` as the calculation unit value, then he selects `add public holiday entitlement` as yes. At this point we will not disable the field because its a valid change, Then the user is now able to select The Hours option value, which is not allowed.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
davialexandre: Why don't we have the same problem with other select fields in this form, like this one: https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php#L216-L219

tunbola: Thats because `accrual_expiration_unit` does not reference an option group.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
davialexandre: Sorry, I think I wasn't clear. What I meant was:
- User selects `Days` as the calculation unit
- Next, selects `add public holiday to entitlement`
- Next changes her mind and select `Hours` as the calculation unit
- We automatically disable the `add public holiday to entitlement` checkbox and reset value to its default one
- User then decides to go back to `Days` again
- We enable the field once more

I think this is simpler because it's just a matter of enabling and disabling fields and we won't need to handle the click on the public holiday field. However, the user won't be informed of anything, but perhaps that's ok. Especially because the specification doesn't mention anything about this message. Can you please check with Mike if we should have the message or not?

davialexandre: Note that for this to work, we need to make sure to also disable the field right after opening form to edit an Absence Type which the calculation unit is hours

mi-kon:
@tunbola 

Can I suggest the following:

1. Disable form elements which cannot be changed due to validation rules. E.g. Calculate leave in or public holidays, using disabled attribute 
2. Display (?) icon after such field labels and on click open default civicrm notification window in top right corner with explanation text (as you do this now on click)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -648,12 +649,7 @@ private function createAbsencePeriod($startDate, $endDate) {
   }
 
   private function createAbsenceType() {
-    return AbsenceType::create([
-      'title' => 'Type ' . microtime(),
-      'color' => '#000000',
-      'default_entitlement' => 20,
-      'allow_request_cancelation' => 1,
-    ]);
+    return AbsenceTypeFabricator::fabricate();
```
davialexandre: Sorry if I wasn't clear. My suggestion was to completely replace the helper method (`createAbsenceType()`) with the fabricator. Can you please do that here and in the other classes?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -128,154 +132,200 @@
...
+                  });
+                }
 
-                      });
+                function calculationUnitRelatedControls() {
+                  publicHolidayToggle();
+                  setEntitlementLabelText();
+                  toggleEntitlementHelpIcon();
+                }
+                function setEntitlementLabelText() {
```
davialexandre: Please add an empty line between these two functions

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -128,154 +132,200 @@
...
-                        {/literal}
-                        {/if}
-                        {literal}
+                  $('.absence-calculation-unit select').on('change', function(e) {
+                    calculationUnitRelatedControls();
+                  });
+                }
 
-                      });
+                function calculationUnitRelatedControls() {
```
davialexandre: Since what this method does is just toggling some fields, how about renaming it to `toggleCalculationUnitRelatedControls`?

PR #2200, opened 2017-10-09
PCHR-2745: Run JS tests with Jenkins

davialexandre:
## Overview

This PR updates the Jenkinsfile to successfully run the JS unit tests in Jenkins.

## Before

Jenkins wasn't able to run the JS unit tests successfully. One of the reasons was that Karma was set to use Chrome as the browser to run the test in, but the server doesn't have a graphical environment available where the browser window can be opened. 

Besides the browser issue, in order to be able to interpret the tests results, Jenkins requires them to to be available as a JUnit XML report. The current Karma config didn't allow that.

## After

The browser issue was fixed by using Chrome in headless mode, which doesn't require a graphical environment. This was achieved by adding this Custom Launcher to the `karma.config.js` file of each extension containing unit tests:

```js
ChromeHeadless: {
        base: 'Chrome',
        flags: [
          '--headless',
          '--disable-gpu',
          // Without a remote debugging port, Google Chrome exits immediately.
          '--remote-debugging-port=9222'
        ]
```

Next, `ChromeHeadless` was set as the browser to run the tests in.

For the test output format, a few steps had to be taken:

1. The `karma-junit-reporter` and `yargs` packages were added as a dev dependencies to all the extensions with JS unit tests
2. In the `karma.conf.js` file of each extension, we set the junitReporter to output the tests results to a specific file inside the extension folder
3. We then use the `yargs` lib inside the `gulpfile.js` of each extension, to read a `--reporters` cli argument and pass any given reporter to karma. This is necessary to allow Jenkins to ask for the `junit` reporter to be used, whereas devs can still use the `progress` during development. If no argument is passed, `progress` will be used by default
4. Update the `Jenkinsfile` to use the new reporter argument. It now uses this command to run the tests for each extension: `gulp test --reporters junit,progress` (Progress is only used so that devs can see a better output while checking the command response on the Jenkins interface).
5. The `Jenkinsfile` was also updated so that the output file created by the Junit Reporter is moved to the Jenkins Workspace folder, where it can finally read all the XML files and analyze the results.

Finally, the way the JS dependencies are installed changed a little bit. First, to decrease the installation time, we now use `yarn` instead of `npm`. Second, we don't install them in parallel anymore. Initially, it was thought that installing them this way would save us time, but the result was actually the opposite, as having multiple installation processes running at the same time increases the number of Disk IO operations and slow things down. Because of that, and also because we don't have that many extensions with JS tests, the pipeline configuration was changed to install the packages in sequence.

## Comments

A few JS tests were failing and @reneolivo fixed them all. For more details about what were those tests and how he fixed them, please check his comment in this PR.

Unfortunately, the build is still failing. The reason is that lots of PHP unit tests for L&A fail on the first run, but pass on the subsequent runs. We still don't know what is causing this, but it will be fixed by a separate PR.


reneolivo:
### Description

Some of the JS tests were breaking. This PR fixes them. 

### Test fixes

Here is a list of extensions and fixes:

- com.civicrm.hrjobroles - was breaking
- uk.co.compucorp.contactaccessrights  - was ok
- org.civicrm.contactsummary  - was breaking
- org.civicrm.hrjobcontract  - was ok
- org.civicrm.hrui - doesn't have any unit tests
- uk.co.compucorp.civicrm.hrleaveandabsences  - has no tests on this branch

#### Job Roles

It was testing for dates that use strings and Date objects, but using `new Date()` makes the date go further than the job contract end, which is 2016-05-05 (don't want to imagine this test was breaking since then).

The solution was to use `new Date(contract.end_date)` so it has a Date object that is not greater than the contract end date.

#### Contact Summary

Linting was applied to the files that were fixed, but did not follow the style guide as it would have taken much longer and would require moving a lot of files around.

The `contact-summary.templates` module was added to all service tests. This fixes the loading of template files. The file required was `views/mainTemplate.html` and it was required by the main `app.js` file.

The contact test was missing the mock for the `ContractService` and it was trying to make real requests.

The contract service was missing the `api_HRJobHour_get` and `api_HRJobPay_get` properties. Not sure if these are still valid properties returned by the endpoint, but they are expected by the service.

The whole LeaveService tests were ignored by using `xdescribe` (exclude) . The whole test suit is not following the structure of the service any more, the controller is using .getCurrent(), the test is using .get(), .get() calls .getData() which is not even defined in the LeaveService.

Short answer, The LeaveService tests were excluded for the following reason:
1. The tests were not following the service structure any more.
2. The LeaveService doesn't exist in 1.7 ([/contactsummary/js/src/contact-summary/services](https://github.com/civicrm/civihr/tree/1.7-wip/contactsummary/js/src/contact-summary/services))
3. This is probably going to be replaced by the Leave Widget which is underway.

### Comments

Please feel free to squash the commits if you feel like it. I just left them like that to make it easier to check the changes.

Jenkinsfile
```diff
@@ -296,21 +289,28 @@ def listCivihrExtensions() {
       name: 'Job Roles',
```
AkA84: pretty ugly that we still don't have a global list of extensions..

contactsummary/package.json
```diff
@@ -1,6 +1,12 @@
 {
   "name": "contact-summary",
   "version": "0.0.1",
+  "semistandard": {
```
AkA84: this is not needed https://github.com/civicrm/civihr/blob/staging/package.json

AkA84:
@reneolivo  the test files are still using the old naming convention, _test.js instead of .spec.js. Can you please update them and update gulpfile as well

reneolivo:
### New changes
Contact Summary Tests now follow the style guide. `_test.js` files were changed to `.spec.js` files. Mock files had the `.mock.js` suffix appended. Gulp and Karma files were updated in order to follow the new naming convention.

PR #2202, opened 2017-10-09
PCHR-2719: Import components between extensions

reneolivo:
## Overview
This PR allows importing JS code from other extensions when building the minified files using RequireJS. This will help import the Leave Widget component from the L&A extension into the Contact Summary.

It also enables source maps for minified files.

### Build Process
The `build.js` file now specifies the extension path it wants to include in the following way:
```js
({
  paths: {
    'leave-absences': '%uk.co.compucorp.civicrm.hrleaveandabsences%/js/angular/src/leave-absences'
  }
})
```
The `gulp requirejs-bundle` was modified so it creates a temporary `build.tmp.js` file, by using `buildFileManager` which is a collection of methods to read extensions from the build file, create the temporary file, and return the watch path for each extension.

The manager can be initialized using `buildFileManager()` which would return a collection of helper methods. This approach was chosen so the manager only runs for the methods that it's needed. 

When the manager starts it will read the build.js file and extract the extensions needed:

```js
function initRequiredExtensions() {
  buildFileContent = fs.readFileSync('./js/build.js', 'utf8'); // stores the content of the build file
  paths = buildFileContent.match(/((%{([\w._-]+)})([^'"]*))/g) || []; // finds all extensions marked with the %extension-name% pattern. it also find the path to the files (`[^'"]*`, the path is similar to this: `/js/angular/src`)
}
```

This function creates a list of required extensions, the alias name, path, and the extension full path.

To get the full path of the extension the command `cv('path -x ' + extensionName)` was used (from the [civicrm-cv](https://github.com/civicrm/cv-nodejs) package). This is used in the `getExtensionObject` which returns the extension name, and full path:

```js
 // parts is an array with the values for the alias name (leave-absences) and the extension name (uk.co.compucorp.civicrm.hrleaveandabsences).
function getExtensionObject (parts) {
    var name = getTrimmedWords(parts[1]); // need to clean white spaces, etc.
    var path = cv('path -x ' + name); // returns the extension full path

    return {
      name: name,
      path: path[0].value
    };
  }
```

The `createTempFile` function creates the temp build.js that replaces the extension names with their full path:

```js
function createTempFile () {
    var tempBuildFileContent = buildFileContent; // copies the build file contents

    requiredExtensions.forEach(function (record) {
      var extensionRegExp = new RegExp(record.alias.path, 'g');

      // for each required extension, replace the extension name with its full path.
      // alias.path is the original build file extension path, alias.fullPath is the full path with the extension
      // name replaced.
      tempBuildFileContent = tempBuildFileContent.replace(
        extensionRegExp, record.alias.fullPath);
    });

    fs.writeFileSync('./js/build.tmp.js', tempBuildFileContent, 'utf8');
  }
```

A `getWatchPaths` function was added in order to return a list of extension paths that can be watched in order to rebuild the dist file when the extension files change. This is helpful if you want to tweak the extension files and see the results in Contact Summary (for this PR case).

```js
function getWatchPaths () {
    // returns a list of extensions and the files to watch:
    return requiredExtensions.map(function (record) {
      return record.alias.fullPath + '/**/**.js';
    });
  }
```

And this is how it can be used inside of the gulpfile:
```js
gulp.task('watch', function () {
  var sourcePath = 'js/src/**/*.js';
  var watchPaths = buildFileManager().getWatchPaths();
  watchPaths.push(sourcePath);

  // watches the contact summary files, as well as extension files:
  gulp.watch(watchPaths, ['requirejs']).on('change', function (file) {
    try { test.for(file.path); } catch (ex) { test.all(); }
  });
});
```

### Moving Leave Widget files
The Leave Widget files were moved to the L&A extension in order to limit the number of files to be imported by the Contact Summary.

The `contactsummary.php` file was modified in order to add the `leaveAbsences` variable which is required by the `shared-settings` module in the L&A extension.

### Source maps

Sourcemap options were added to the `build.js` file of each extension. The build options are these:
```js
generateSourceMaps: true,
useSourceUrl: true,
```

#### Before
map-error-1

#### After
screenshot from 2017-10-09 17-49-42
**Note**: This is a sample error to demonstrate how the source map works.

Not the most helpful of error messages, but at least it gives you the file path and line.

### Removing Debug Mode support

Debug Mode support has been removed from all extensions. PHP files with the following pattern have been amended so they only load minified files.

#### Before

```php
CRM_Core_Resources::singleton()->addScriptFile(
  'com.civicrm.hrjobroles',
  CRM_Core_Config::singleton()->debug
    ? 'js/src/job-roles.js'
    : 'js/dist/job-roles.min.js',
  1010
);
```

#### After

```php
CRM_Core_Resources::singleton()->addScriptFile(
  'com.civicrm.hrjobroles',
  'js/dist/job-roles.min.js',
  1010
);
```

### Contact Summary Tests

After importing code from the L&A extension, the contact summary tests stopped working because the L&A files could not be found by Requirejs and Karma. To make them run again the following must have been done:

* A new `leave-absences` path was added to the requirejs configuration inside the `test-main.js` file. The path ends up like this: `/base/tools/extensions/civihr/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences`
* In the karma configuration, a new path was added for L&A extension files:
`{ pattern: civihrPath + 'uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/**/*.js', included: false }`
* Added a new CRM.vars field ` leaveAndAbsences` in the `globals.js` file. This variable is needed by the shared settings.

There were a few tests failing for the contact, contactDetails, and contract services. These tests were fixed, and lint issues were amended, but upgrading them to follow the style guide was skipped as this would have required changing all services and their references, which is outside of the scope of this PR.

## Tests
- com.civicrm.hrjobroles :white_check_mark: all good
- uk.co.compucorp.contactaccessrights :white_check_mark: all good
- org.civicrm.contactsummary :white_check_mark: all good
- org.civicrm.hrjobcontract :white_check_mark: all good
- org.civicrm.hrui :no_good_man: doesn't have any unit tests.
- uk.co.compucorp.civicrm.hrleaveandabsences :white_check_mark: all good

## Comments
This PR is being sent to **PCHR-2654-leave-widget**.


contactsummary/contactsummary.php
```diff
@@ -127,9 +127,14 @@ function contactsummary_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
...
 function contactsummary_civicrm_pageRun($page) {
   if ($page instanceof CRM_Contact_Page_View_Summary) {
-    CRM_Core_Resources::singleton()->addSetting(array(
-      'tabSettings' => array('active' => CRM_Utils_Request::retrieve('selectedChild', 'String', $this, FALSE, 'contactsummary')),
-    ));
+    CRM_Core_Resources::singleton()
+      ->addSetting(array(
+        'tabSettings' => array('active' => CRM_Utils_Request::retrieve('selectedChild', 'String', $this, FALSE, 'contactsummary')),
+      ))
+      ->addVars('leaveAndAbsences', [
```
AkA84: Not sure if it's even possible to avoid this, but this hard coupling between the extensions is pretty bad

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,153 @@ var exec = require('child_process').exec;
...
-    exec('r.js -o js/build.js', function (err, stdout, stderr) {
-        err && err.code && console.log(stdout);
+gulp.task('requirejs-bundle', ['create-full-path-build-file'], function (done) {
+  exec('r.js -o js/build.tmp.js', function (err, stdout, stderr) {
+    err && err.code && console.log(stdout);
+    done();
+  });
+});
+
+gulp.task('create-full-path-build-file', function (done) {
```
AkA84: why is this needed to be a public task? Shouldn't it be a simple private function?

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,153 @@ var exec = require('child_process').exec;
...
 
-gulp.task('sass-sync', function(){
+gulp.task('sass-sync', function () {
   civicrmScssRoot.updateSync();
 });
 
-gulp.task('requirejs-bundle', function (done) {
-    exec('r.js -o js/build.js', function (err, stdout, stderr) {
-        err && err.code && console.log(stdout);
+gulp.task('requirejs-bundle', ['create-full-path-build-file'], function (done) {
```
AkA84: rename it as `requirejs`, like in https://github.com/civicrm/civihr/blob/la-milestone-3/uk.co.compucorp.civicrm.hrleaveandabsences/gulpfile.js

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,153 @@ var exec = require('child_process').exec;
...
+
+gulp.task('create-full-path-build-file', function (done) {
+  fs.readFile('./js/build.js', 'utf8', function (error, buildFile) {
+    if (error) {
+      throw error;
+    }
+
+    var extensions = buildFile.match(/%([A-z0-9.\-_]+)%/g) || [];
+    // return unique extensions:
+    extensions = extensions.filter(function (ext, index) {
```
AkA84: use lodash https://lodash.com/docs#uniq

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,153 @@ var exec = require('child_process').exec;
...
+      return extensions.indexOf(ext) === index;
+    });
+
+    var promises = extensions.map(function (extension) {
+      var cleanExtensionName = extension.match(/%([A-z0-9.\-_]+)%/)[1];
```
AkA84: you have the same regex pattern repeated, put it in a variable

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,153 @@ var exec = require('child_process').exec;
...
+      return cv('path -x ' + cleanExtensionName)
+      .then(function (path) {
+        return {
+          extension: extension,
+          path: path[0].value
+        };
+      });
+    });
+
+    Promise.all(promises).then(function (results) {
```
AkA84:
```js
Promise.all(extensionNames(buildFile).map(function (name) {
  // return promises
}))
.then(function (results) {

});

function extensionNames (buildFile) {
  return _(buildFile.match(/%([A-z0-9.\-_]+)%/g) || [])
    .uniq()
    .map(function (notCleanExtensionsName) {
      // clean the name
    })
    .value();
}
```

contactsummary/gulpfile.js
AkA84: this whole task/function does too many things, please split it into separate smaller functions

contactsummary/js/src/contact-summary/app.js
```diff
@@ -2,22 +2,23 @@
...
-  'contact-summary/components/leave-widget/leave-widget.component',
   'contact-summary/directives/donutChart'
 ], function (angular) {
   var app = angular.module('contactsummary', [
     'ngRoute',
     'ngResource',
     'ui.bootstrap',
     'common.services',
+    'leave-absences.components',
+    'leave-absences.settings',
```
AkA84: i think contacsummary shouldn't know anything about ` leave-absences.settings`, this looks like an exposure of internal details of L&A.  `leave-absences.components`, should be the one defining as a dependency `leave-absences.settings`

contactsummary/package.json
```diff
@@ -18,5 +18,8 @@
     "path": "^0.12.7",
     "protractor": "1.4.0",
     "requirejs": "^2.1.22"
+  },
+  "dependencies": {
```
AkA84: this should be a devDependency

contactsummary/package.json
```diff
@@ -18,5 +18,8 @@
...
+    "civicrm-cv": "^0.1.2"
```
AkA84: please specify a minor range, not a major one

contactsummary/scss/contactsummary.scss
```diff
@@ -2,6 +2,7 @@
 @import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/variables";
 
 #bootstrap-theme {
+  @import "SCSSROOT/uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/leave-widget";
```
AkA84: just because of this i would be in favour for contactsummary to be the owner of the widget. I mean, havng also an hard coupling between css is yet another bad thing

contactsummary/contactsummary.php
reneolivo: Agreed. Some of these need to go into an API endpoint (logged in contact id, upload token, etc.)

davialexandre: @AkA84 @reneolivo Just reminding that we have a ticket for such API: PCHR-2465. We definitely should implement it and use it instead

davialexandre:
@AkA84 @reneolivo I didn't fully review the code, but I'm curious: Why does the Leave Widget belong to the L&A extension instead of the Contact Summary extension? The way I see it, the Widget should be part of Contact Summary and it should use the APIs provided by L&A.

contactsummary/scss/contactsummary.scss
reneolivo: The hard coupling would have to come from somewhere with that approach as well. If the owner is the Contact Summary then we'd need to import a lot of services from L&A (AbsencePeriod, AbsenceType, Entitlement, LeaveRequest, and WorkPattern).

reneolivo:
@davialexandre, @AkA84 I moved the Leave Widget to L&A because it needs to import a lot of services from L&A. I'm trying to avoid the Contact Summary to know a lot about L&A. It would need to know so much, that any change to the underlying Models or APIs might break it, and with our current test workflow we would not know.

By importing the component we make sure it works as intended without knowing any details about the models inside of L&A.

contactsummary/contactsummary.php
reneolivo: That's really nice @davialexandre!!! The only issue is that the shared settings is used to load template files:

Component:
```js
templateUrl: ['shared-settings', function (sharedSettings) {
  return sharedSettings.sharedPathTpl + 'components/leave-widget/leave-widget.html';
}]
```

Shared Settings:
```js
sharedPathTpl: CRM.vars.leaveAndAbsences.baseURL + '/views/shared/',
```

And templateUrl nor template don't seem to work with promises. (was checking). Maybe there could be a workaround, but this needs more investigation.

AkA84:
@reneolivo not sure i like the placeholder format, it's kind of unusual, what about
```js
'leave-absences': '%{uk.co.compucorp.civicrm.hrleaveandabsences}/js/angular/src/leave-absences'
```

reneolivo:
> @reneolivo not sure i like the placeholder format, it's kind of unusual, what about

@AkA84 that's fine. I was just trying to avoid using `{extension-name}` because that can easily be confused with a json property regex-wise, but `%{extension-name}` is fine.

AkA84:
@reneolivo i'm not sure i understand how you arranged for the gulpfile to perform the string replacement. As far as I understood it, it should be something simple as

1) Find a `/%{[\w._-]})/` placeholder, which is an extension name
2) With `cv` find the path of the extension
3) Replace the placeholder with the extension

But in the gulpfile that you write there is a lot more going on, including some "alias" (which has a path? which is also different from the extension path?) stuff. 

Are we sure we're not overcomplicating things here?

deb1990:
@reneolivo Looks good to me.

reneolivo:
@AkA84 that was done in order to grab the details for each extension first and then you can do one of two things: to create the build.tmp.js file or to return a list of extension files to watch.

So this is the workflow:

1. read the build.js file for each `%{extension-name}`.
2. create an object for each extension with the alias name (just for reference), path (the original build.js path for this alias), and the full path (which is the path we are interested in).

And then one of these two:
* create the build.tmp.js file
* return a list of files from the extensions to watch over them.

bmnepali:
@reneolivo 👍 Looks good to me.

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,220 @@ var exec = require('child_process').exec;
...
+    var extensions = [];
+    var extensionPathRegExp = /((%{([\w._-]+)})(.*))/g;
+
+    // linter doesn't like assignments as condition.
+    while (true) {
+      var matches = extensionPathRegExp.exec(buildFileContent);
+
+      if (!matches) break;
+
+      var extension = {
```
AkA84: Don't use an object as a temporary variable, it's kinda confusing
```js
var name, placeholder, pathToFiles, pathToExt;
var matches = extensionPathRegExp.exec(buildFileContent);

if (!matches) {
	break;
}

name = matches[3],
placeholder = matches[2],
pathToFiles = matches[4]
pathToExt = getExtensionPathByName(name);

extensions.push({
  placeholder: placeholder,
  path: pathToExt,
  fullPath: path.join(pathToExt, pathToFiles)
});

return extensions;
```

igorpavlov:
@reneolivo 
1) not sure you need `.*`
```js
paths = buildFileContent.match(/.*%([\w._-]+)%.*/g) || []; // finds all extensions marked with the %extension-name% pattern.
```
2) shouldn't dot be escaped `[\w._-]`, it might now match every symbol because of the dot

contactsummary/js/karma.conf.js
```diff
@@ -29,6 +29,9 @@ module.exports = function (config) {
       // all the common/ dependencies
       civihrPath + 'org.civicrm.reqangular/dist/reqangular.min.js',
 
+      // Extensions files
```
AkA84: "external" extensions

contactsummary/js/test/globals.js
```diff
@@ -1,12 +1,15 @@
...
-    CRM.vars = {
-        contactsummary: {
-            baseURL: '/base/tools/extensions/civihr/contactsummary'
-        }
-    };
+  CRM.vars = {
+    contactsummary: {
+      baseURL: '/base/tools/extensions/civihr/contactsummary'
+    },
+    leaveAndAbsences: {}
```
AkA84: is this needed at all?

davialexandre:
@reneolivo I can see by the PR description that all the tests for hrjobroles and contactsummary are passing, which is different than the current situation we have on staging, as you can see here: https://jenkins.compucorp.co.uk/blue/organizations/jenkins/civihr/detail/PR-2200/9/tests

Did you fix those failing tests as part of this PR or were they simply passing already?

com.civicrm.hrjobroles/js/build.js
```diff
@@ -5,6 +5,8 @@
   out: 'dist/job-roles.min.js',
   name: 'job-roles',
   skipModuleInsertion: true,
+  generateSourceMaps: true,
+  useSourceUrl: true,
```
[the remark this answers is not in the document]
reneolivo: It creates a separate a source map file instead of embedding it inside the minified file

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,220 @@ var exec = require('child_process').exec;
...
+
+  /**
+   * Returns watch paths for JS files that are part of the extensions specified
+   * in the build file.
+   *
+   * @return {Array}
+   */
+  function getExtensionWatchPaths () {
+    var buildFileContent = getBuidFileContent();
+    var requiredExtensions = getExtensionsFromBuildContent(buildFileContent);
```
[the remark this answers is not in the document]
reneolivo: I was trying to avoid that since this is not necessary for sass tasks, etc. I could do as you say, or I could return a constructor function, something like this:

```js
function buildFileManager () {
  var externalExtensions;

  this.createTempFile = createTempFile;
  this.getExtensionWatchPaths = getExtensionWatchPaths;

  (function init () {
    initExternalExtensions()
  })();

  // definitions
}
```

and use it like this:

```js
new buildFileManager().createTempFile()
```

contactsummary/js/test/globals.js
reneolivo: unfortunately yes. the shared-settings module of L&A needs to find this variable in the global scope (CRM.vars.leaveAndAbsences)

reneolivo:
@davialexandre I only fixed the errors on la-milestone-3, I see some of the error messages are familiar, but not all of them.
`Expected 'Error' to be true.` 
`Unexpected request: GET ...........contactsummary/views/mainTemplate.html`

deb1990:
@reneolivo Should not theses changes should go to 1.6?

contactsummary/gulpfile.js
```diff
@@ -9,110 +9,220 @@ var exec = require('child_process').exec;
...
+   * - placeholder: The extension placeholder. Ex: %{uk.co.compucorp.extension-name}
+   * - path: The path to the extension. Ex: /root/path/to/ext/
+   * - fullPath: The path + the sub path specified in the build file. Ex:
+   *   /root/path/to/ext/js/angular/src/shared'
+   */
+  function getExtensionsFromBuildContent (buildFileContent) {
+    var extensions = [];
+    var extensionPathRegExp = /((%{([\w._-]+)})(.*))/g;
+
+    // linter doesn't like assignments as condition.
```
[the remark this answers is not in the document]
reneolivo:
```js
while(variable = functionCall()) {

}
```
throws "Expected a conditional expression and instead saw an assignment."

it seems this fixes the issue:
`while((variable = functionCall()))`
:\

reneolivo:
@reneolivo please remember to merge this when merging to a base branch:
https://github.com/compucorp/civihr-tasks-assignments/pull/269

PR #2211, opened 2017-10-13
PCHR-2749: Close quick search field on outside click + refactoring

AkA84:
## Overview
This PR improves the behaviour of the collapsible quick search field,

## Before
The field does not get collapsed when the user clicks outside of it

before

## After
Now the field gets collapsed with an outside click

after

## Comments
This happens only if the field is empty. Otherwise the field remains visible (as it's the expected behaviour)


hrui/js/src/hrui.js
```diff
@@ -140,8 +140,23 @@
...
-    toggleCustomClassToQuickSearchField();
     changeQuickSearchFieldPlaceholder();
+    manageCustomClassOfQuickSearchField();
+  }
+
+  /**
+   * Checks if the user has clicked outside of the quick search field
+   * by analyzing the given click target
+   *
+   * @param {object} target
```
igorpavlov: Could it be better to state `@param {Element} target` instead?

igorpavlov:
@AkA84 Any particular reason why not smth like that?:
```js
$elem.blur(function () {
   if ($elem.val() === '') $elem.hide();
});
```

hrui/js/src/hrui.js
```diff
@@ -140,8 +140,23 @@
...
+   * @return {boolean}
+   */
+  function hasUserClickedOutsideQuickSearchField (target) {
+    $target = $(target);
```
igorpavlov: Should be `var $target = ...`

I wonder why semistandard didn't yell? Or you can mutate `target = ...` to save some memory.

hrui/js/src/hrui.js
```diff
@@ -140,8 +140,23 @@
...
+
+    return !$target.is('#crm-qsearch') &&
+      !$target.is('#root-menu-div') &&
+      !$target.parents('#crm-qsearch, #root-menu-div').length;
```
igorpavlov: What about `!$target.closest('#crm-qsearch, #root-menu-div').length;` ?

hrui/js/src/hrui.js
```diff
@@ -160,6 +175,15 @@
   }
 
   /**
+   * Checks if the quick search field itself has any current value
```
igorpavlov: Checks if the quick search field currently has any value

hrui/js/src/hrui.js
```diff
@@ -160,6 +175,15 @@
...
+   *
+   * @return {boolean}
+   */
+  function isQuickSearchOnGoing () {
+    return $('#sort_name_navigation').val().length;
```
igorpavlov: what if `$('#sort_name_navigation').val() === '        '` don't we need to trim()?

this is not a change request, up to you.

hrui/js/src/hrui.js
```diff
@@ -197,6 +232,22 @@
   }
 
   /**
+   * Removes the given custom class when the user clicks
+   * outside the quick search field (if there is no ongoing search)
+   *
+   * @param {string} customClass
+   */
+  function removeCustomClassOnOutsideClick (customClass) {
+    $('body').on('click', function (event) {
```
igorpavlov: $(document) would be safer in case `body` has margins.

hrui/js/src/hrui.js
```diff
@@ -197,6 +232,22 @@
...
+      $target = $(event.target);
```
igorpavlov: Same as above. It should yell $target is not defined.

hrui/js/src/hrui.js
AkA84: wow, that's bad. I also forgot to enable the git hook in my local env, will do now

hrui/js/src/hrui.js
AkA84: changed it to
```js
return !!$('#sort_name_navigation').val().trim();
```

PR #2214, opened 2017-10-16
PCHR-2560: Onboarding email

mickadoo:
## Overview
The onboarding epic requires two new email templates; one for the system invitation mail and the other for the password reset mail.

## Before

Both emails used the default Drupal mail.

#### System Invite
image

#### Password reset
image

## After

Two new templates exist created with Foundation

#### System Invite
image

#### Password reset
image

## Comments
This PR just generates the templates, they are used inside civihr_employee_portal since it's a Drupal mail that needs to be replaced.


AkA84:
@mickadoo it seems that there is extra padding at the bottom, right after the action button. Anything that we can do something about?

uk.co.compucorp.civicrm.hremails/email-templates/src/layouts/civihr-default.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
AkA84: what's this https://github.com/civicrm/civihr/blob/staging/uk.co.compucorp.civicrm.hremails/email-templates/src/layouts/default.html then? Could you check? I can't recall now

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-account-activated.html
```diff
@@ -0,0 +1,37 @@
+---
+layout: civihr-default
+title:  Welcome to CiviHR!
+---
+
+<container>
+
```
AkA84: no need to keep these empty lines

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-account-activated.html
```diff
@@ -0,0 +1,37 @@
...
+  <row>
+    <columns bgcolor="#ffffff">
+      <spacer></spacer>
+      <h2>Please create a password</h2>
+      <spacer></spacer>
+      Hello, and welcome to CiviHR, the open-source HR solution for non-profit organisations.
```
AkA84: shouldn't the text be wrapped in a `<p>` *or whatever equivalent custom tag foundation gives us) instead of having to add a spacer every time?

AkA84:
also @mickadoo you have two commits with the same title, either you change the title or you squash them, thanks

uk.co.compucorp.civicrm.hremails/email-templates/src/layouts/civihr-default.html
mickadoo: I think this might be because some templates only existed in the 1.7 branch which I didn't have available. If there is already a default then it doesn't make sense to have civihr-default too so I'll see if I can use default.html instead.

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-account-activated.html
mickadoo: I was experimenting with `<p>` when I found the problem with Inbox. The end result is just whatever seemed to be looking good at the time, but I'll try using `<p>`. If the default spacing with `<p>` isn't enough do you think it would be better to add a class to that `<p>` and adjust the spacing with CSS or use `<p>` in combination with `<spacer>`?

mickadoo:
@AkA84 haven't tested with Litmus as I don't have account details. I can ping Jamie about it.

A few things you mentioned I'll need to go back and change, I'll hopefully have it ready for re-review soon.

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-account-activated.html
AkA84: @mickadoo better to use `<p>` only, with maybe a class. Having to remember to use `<p>` + `<spacer>` every time is just annoying

mickadoo: This shouldn't be an issue now since the markup is coming from Drupal

mickadoo:
@AkA84 I've fixed the padding, replaced email content with a placeholder, squashed / renamed some commits and removed the unnecessary extra template (so we only have default.html).

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-account-activated.html
```diff
@@ -0,0 +1,17 @@
+---
+layout: default
+subject:  Welcome to CiviHR!
+---
+
+<container>
+
```
AkA84: remove this new line

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-password-reset.html
```diff
@@ -0,0 +1,18 @@
+---
+layout: default
+subject:  Password Reset
+---
+
+<container>
+
```
AkA84: remove this new line

uk.co.compucorp.civicrm.hremails/email-templates/src/pages/drupal-password-reset.html
```diff
@@ -0,0 +1,18 @@
...
+  <row>
+    <columns bgcolor="#ffffff">
+      {$body}
+      <button href="{$resetLink}" class="expanded alert end">
+        Reset Password
+      </button>
+    </columns>
+  </row>
+
```
AkA84: remove this new line

mickadoo:
Autoclosed when 2339 branch was merged and couldn't reopen it, so leaving closed in favor of #2223

PR #2216, opened 2017-10-17
PCHR 2764: Create LeavePeriodEntitlementLog Entity

tunbola:
## Overview
The  LeavePeriodEntitlementLog Entity will be used to log changes made when changes are made to a Leave Period entitlement on the Entitlement calculation page.

## Before
The LeavePeriodEntitlement Log Entity does not exist.

## After
The LeavePeriodEntitlement Log Entity was created.
Table name:  **civicrm_hrleaveandabsences_leave_period_entitlement_log**

Fields
- id: type=int, required=true
- entitlement_id: type=int, required=true
- entitlement_amount: DECIMAL(20,2), required=true
- comment: type:TEXT
- author_id: type:int,FK to Contact. The contact that represents the user who made a change to this entitlement, required=true
- created_date: type:DATETIME. The date and time this entitlement was updated

## Technical Details
- Since this Entity is basically used to keep a log of changes, updates are dis-allowed for the Entity.


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlementLog.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param array $params
+ *
+ * @return array API result descriptor
+ *
+ * @throws API_Exception
+ */
+function civicrm_api3_leave_period_entitlement_log_create($params) {
+  return _civicrm_api3_basic_create(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+}
```
davialexandre: Since people should not be able to create or update log entries using the API, I think it would be better to completely remove this endpoint

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeavePeriodEntitlementLog.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param array $params
+ *
+ * @return array API result descriptor
+ *
+ * @throws API_Exception
+ */
+function civicrm_api3_leave_period_entitlement_log_delete($params) {
+  return _civicrm_api3_basic_delete(_civicrm_api3_get_BAO(__FUNCTION__), $params);
+}
```
davialexandre: Perhaps we should remove this endpoint as well

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementLog.php
```diff
@@ -0,0 +1,90 @@
...
+   * Create a new LeavePeriodEntitlementLog based on array-data
+   *
+   * @param array $params
+   *
+   * @return CRM_HRLeaveAndAbsences_DAO_LeavePeriodEntitlementLog|NULL
+   */
+  public static function create($params) {
+    $entityName = 'LeavePeriodEntitlementLog';
+    $hook = empty($params['id']) ? 'create' : 'edit';
+
```
davialexandre: I think it would be better to prevent the update here instead using the `validateUpdatesNotAllowed` method. One of the reasons is that, if the update won't really happen, there's reason to trigger the pre hook. If we throw an exception right here, the hook will never be triggered.

PR #2223, opened 2017-10-17
PCHR-2560: Onboarding email

mickadoo:
## Overview
The onboarding epic requires two new email templates; one for the system invitation mail and the other for the password reset mail.

## Before

Both emails used the default Drupal mail.

#### System Invite
image

#### Password reset
image

## After

Two new templates exist created with Foundation

#### System Invite
image

#### Password reset
image

## Comments
This PR just generates the templates, they are used inside civihr_employee_portal since it's a Drupal mail that needs to be replaced.


mickadoo:
Merging as #2214 was approved but auto closing of PR could not be reopened even after changing base branch.

PR #2225, opened 2017-10-18
PCHR-2789: Fix L&A tests in Jenkins

davialexandre:
## Overview
This PR fixes several errors while running the L&A tests with Jenkins. There were different types of errors, but almost all of them happened only the first time the tests were executed on a brand new site (which is exactly what Jenkins does). The only exception was with a set of tests around Leave Request attachments, that failed because the test site didn't have the attachments patch applied.

## Before
This is the situation before the fixes:

captura de tela de 2017-10-23 09-38-53

## After

All the tests are passing:

captura de tela de 2017-10-23 10-17-42

Please check the next section for details on the problems and the solutions that were applied.

## Technical Details

As previously mentioned, there were different types of errors. Here is a list of them, and the solutions that were applied:

#### Attachments tests

There are a number of tests in this extension that requires changes in CiviCRM core related to attachments. I've sent PRs to core for these, but they've not been merged yet, as they still require some work. In the meantime, we've been using a patch to apply those changes in our sites. The site created by Jenkins didn't have this patch and the tests were failing. 

**Solution**: Update the Jenkins file with commands to apply the patch. With this change, now that staging is CiviHR 1.7, I've also updated it to use the hr17 build type, rather than hr16.

#### Tests failing because of data providers using option groups

The `LeaveRequestHelpers` trait, provided some methods to quickly get a list of option values for Leave Request statuses. Some tests were using these methods as PHPUnit data providers. Unfortunately, in order to cache things, PHPUnit calls the dataProviders way before actually running the tests, at a point where the CiviCRM environment isn't properly set and the Option Groups haven't been stored in the database yet. This resulted in Data Providers returning invalid values (`null`) that made the tests fail.

Note: This only happens when we run the extension the first time. On a second run, since the environment is the same (that is, the environment set by the `setUpHeadless()` method), civi won't reset the database and the Option Values will be there, making the data providers work just fine.

**Solution**: I've updated the trait methods to load data from the Option Group XMLs instead of the database. I've also checked all the tests that were using those methods to make sure if they're really necessary. Where it wasn't, I replace the helper methods calls to a call for `LeaveRequest::getStatuses()` instead.

#### Tests failing because of a missing `hrleaveandabsences.settings_manager` service

The extension defines a service that encapsulates some logic to return a different Settings Manager during the tests (an in-memory one, to speed things up). This service was exposed using the CiviCRM container, which is normally heavily cached. The problem happened because, when we run the tests, Civi loads the cached container created during the tests execution of the previous extension, which didn't include the necessary service and the tests would fail.

Note: This also only happens the first time we run the extension tests. The next time, the container cache would have been updated (when the L&A extension is installed during the `setupHeadless()` call) and the tests will pass.

**Solution**: The `CIVICRM_CONTAINER_CACHE` constant was defined to `never` when runnint the L&A tests. This way, we always force it to start from a clean state.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
@@ -7,7 +7,6 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
   /**
    * @var array|null
    *   Caches the list of option values for the type field.
-   *   When it is null, it means the values were not loaded yet
    */
   private static $workDayTypes;
 
@@ -54,7 +53,7 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
    * @return string
    */
   private static function getTypeValue($optionName) {
-    if(is_null(self::$workDayTypes)) {
+    if(empty(self::$workDayTypes)) {
       self::$workDayTypes = array_flip(self::buildOptions('type', 'validate'));
 
       // The option value values are stored as strings on the database, so we
@@ -204,6 +203,7 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
     if(!in_array($type, $validTypes)) {
       throw new InvalidWorkDayException(
         'Invalid Work Day Type'
+        //'Invalid Work Day Type: ' . $type . 'The valid types are: '.implode(', ', $validTypes)
       );
     }
   }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
@@ -203,7 +203,6 @@ class CRM_HRLeaveAndAbsences_BAO_WorkDay extends CRM_HRLeaveAndAbsences_DAO_Work
     if(!in_array($type, $validTypes)) {
       throw new InvalidWorkDayException(
         'Invalid Work Day Type'
-        //'Invalid Work Day Type: ' . $type . 'The valid types are: '.implode(', ', $validTypes)
       );
     }
   }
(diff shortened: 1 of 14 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
```diff
@@ -204,6 +203,7 @@ private static function validateWorkDayType($params) {
     if(!in_array($type, $validTypes)) {
       throw new InvalidWorkDayException(
         'Invalid Work Day Type'
+        //'Invalid Work Day Type: ' . $type . 'The valid types are: '.implode(', ', $validTypes)
```
tunbola: We don't need this line right?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/helpers/OptionGroupHelpersTrait.php
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /**
+   * This methods parses the given Option Group XML and returns a list of
+   * Option Values contained in it.
+   *
+   * Important note: It assumes that the XML represents a single Option Group,
+   * and that all the Option Values in belong to the same Group.
+   *
+   * @param $xml
```
tunbola: Missing param type.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/bootstrap.php
```diff
@@ -1,5 +1,5 @@
 <?php
-
+define('CIVICRM_CONTAINER_CACHE', 'never');
```
tunbola: We should probably have the define statements grouped together. Maybe move the `define('CIVICRM_MAILER_SPOOL', 1);` up also to after this line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -38,7 +38,7 @@ public function setUp() {
     $this->registerCurrentLoggedInContactInSession($this->leaveContact);
     CRM_Core_Config::singleton()->userPermissionClass->permissions = [];
 
-    $this->leaveRequestStatuses = $this->getLeaveRequestStatuses();
+    $this->leaveRequestStatuses = LeaveRequest::getStatuses();
```
tunbola: Shouldn't this test class be using the LeaveRequestHelpersTrait.php so that the option values are loaded from the xml file?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkDay.php
davialexandre: Nice catch!

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
davialexandre: It turns out that this was only really used by a single method, so instead of using the class property, I've change it to load the statuses inside the test

PR #2226, opened 2017-10-19
PCHR-2720: Leave Widget sick leave balance and heatmap

reneolivo:
## Overview
This PR adds real data for the Leave Widget in these two areas:
1. Sick Leave balances. A list of sick types and their total balance.
2. Sickness heatmap. A week heatmap with the days where most sick leaves were taken.

**Note**: This PR is targeted for the feature branch `PCHR-2654-leave-widget`.

## Before
download 86

## After
anim
**Note:** the difference between the values from the absence type balances and the heatmap is due to the non-working days (weekends, non working days, public holidays, etc)

### Absence Types Taken component

This is a generic component that would display a list of absence types, their balances, and a week heatmap for that balance.

In the case for the sickness section, the component was used in this manner:

```html
<leave-widget-absence-types-amount-taken
  title="Sickness"
  absence-period="leaveWidget.currentAbsencePeriod"
  absence-types="leaveWidget.sicknessAbsenceTypes"
  contact-id="leaveWidget.contactId"
  leave-request-statuses="leaveWidget.leaveRequestStatuses"
></leave-widget-absence-types-amount-taken>
```

The bindings are:
* **title**: The title that will be displayed at the start of the component and above the heatmap.
* **absence-period**: The absence period to use for consulting the balance.
* **absence-types**: Only balances for these absence types will be consulted. In this case we want to consult the balance for any sickness absence types. ([here](https://github.com/civicrm/civihr/pull/2226/files#diff-d127156ec6d73b695fee22489c6a8618R120) is where sickness absence types are selected)
* **contact-id**: Only requests for this contact will be taken into consideration for the balance.
* **leave-request-statuses**: Only leave request with these statuses are loaded. In this case we only count the balance for approved, admin approved, awaiting for approval, or more information required statuses. ([here](https://github.com/civicrm/civihr/pull/2226/files#diff-d127156ec6d73b695fee22489c6a8618R131) is where statuses are loaded)

Leave requests are loaded in order to calculate the balance for each absence type and to use the leave request dates to construct the heatmap. The code to load the leave requests is this:

```js
return LeaveRequest.all({
  contact_id: vm.contactId,
  from_date: { '>=': vm.currentAbsencePeriod.start_date },
  to_date: { '<=': vm.currentAbsencePeriod.end_date },
  status_id: { IN: statusIds },
  type_id: { IN: absenceTypeIds }
})
.then(function (response) {
  vm.leaveRequests = response.list;

  mapAbsenceTypesBalance();
});
```

### Sick leave balances

This code maps each absence type with the total balance change from each leave request of that type:
```js
function mapAbsenceTypesBalance () {
  vm.absenceTypes = vm.absenceTypes.map(function (absenceType) {
    var balance;

    balance = vm.leaveRequests.filter(function (request) {
      return +request.type_id === +absenceType.id;
    })
    .reduce(function (balance, request) {
      return balance + request.balance_change;
    }, 0);
    balance = Math.abs(balance);

    return _.assign({ balance: balance }, absenceType);
  });
}
```
`Math.abs` is used for two reasons:
1. Balance changes are negative, so this display the final balance as positive.
2. TOIL requests have positive balance changes (this component is not only for sick types).

### Leave Requests Heatmap component

A leave requests heatmap component was created. This component takes a list of leave requests and it will map their balances to a heatmap. This is how the component is used:

```html
<leave-requests-heatmap
  leave-requests="absenceTypesTaken.leaveRequests"
></leave-requests-heatmap>
```

When the leave requests are bound to the component, a call to the `mapLeaveRequestsToHeatmapValues` function is made. This takes care to map the leave request balances to a heatmap object grouped by week days:
 
```js
function mapLeaveRequestsToHeatmapValues () {
  vm.heatmapValues = {};

  vm.leaveRequests.reduce(function (dates, request) {
    return dates.concat(request.dates);
  }, [])
```
this flattens the request dates into a single array.

```js
  arrayOfDates.forEach(function (date) {
    // 0 = Monday, 6 = Sunday:
    var dayOfTheWeek = moment(date.date).isoWeekday() - 1;

    if (!vm.heatmapValues[dayOfTheWeek]) {
      vm.heatmapValues[dayOfTheWeek] = 0;
    }

    vm.heatmapValues[dayOfTheWeek]++;
  });
}
```
This maps each date to a day of the week, adding 1 balance change per date. 

`isoWeekday` returns a number from 1 to 7, 1 being Monday, and Sunday being 7. For the purpose of the heatmap, Monday needs to be 0, and Sunday 6.

### Heatmap component
A Heatmap component was created in reqangular for general use. The component can be used as this:

```html
<heatmap values="heatMapValues"></heatmap>
```
The `heatpMapValues` is an object that maps days of the weeks (0 = Monday, 1 = Tuesday, 2 = Wednesday, etc.) to their values, which can be any type of number. If a value of `false` is provided, the day will appear as disabled (gray background).

```js
vm.weekHeatMapValues = {
  0: 1, // Monday: 1
  1: 2, // Tuesday: 2
  2: 1, // Wednesday: 1
  3: 10, // Thursday: 10
  4: 6, // Friday: 6
  5: false, // Saturday: disabled
  6: false // Sunday: disabled
};
```

This will try to categorize each value into one of these 4 possibilities:
* **low** for values equal to 0, or very close to the lowest value. In this case 1.
* **medium** for values close to the average. In this case the average is `(1 + 2 + 1 + 10 + 6) / 5 = 4`.
* **high** for values close to the highest value. In this case 10.
* **disabed** for values strictly equal to `false`.

A heat range object is created in order to find the closest values:
```js
var heatRanges = [
  { label: 'low', value: 1 },
  { label: 'medium', value: 4 },
  { label: 'high', value: 10 }
];
```
And to actually find the closest heat value:
```js
var heat = heatRanges.reduce(function (previous, current) {
  return Math.abs(current.value - value) < Math.abs(previous.value - value)
    ? current
    : previous;
});
```
This compares the value (1, 2, 1, 10, 6) against the ranges values (low vs medium, medium vs high) and returns the one that is closest.

The result is stored in `vm.days` with the following structure:

```js
vm.days = [
  { value: 0, heat: 'low', label: 'Monday', shortLabel: 'M' },
  { value: 2, heat: 'low', label: 'Tuesday', shortLabel: 'T' },
  { value: 2, heat: 'low', label: 'Wednesday', shortLabel: 'W' },
  { value: 10, heat: 'high', label: 'Thursday', shortLabel: 'T' },
  { value: 6, heat: 'medium', label: 'Friday', shortLabel: 'F' },
  { value: 0, heat: 'disabled', label: 'Saturday', shortLabel: 'S' },
  { value: 0, heat: 'disabled', label: 'Sunday', shortLabel: 'S' }
];
```

### Controller on changes

A `ControllerOnChanges` helper was added to the L&A tests. The purpose of this helper is to mock `$onChanges` calls ([documentation](https://docs.angularjs.org/guide/component)).

The helper can be used as follows:

```js
ControllerOnChanges.setupController(controllerInstance);

ControllerOnChanges.mockChange('leaveRequests', leaveRequestsArray);
```

This will trigger a call to the controller's $onChanges method similar to this:

```js
controllerInstance.leaveRequest = leaveRequestsArray;

controllerInstance.$onChanges({
  leaveRequest: {
    currentValue: leaveRequestsArray,
    previousValue: aReferenceToThePreviousValue
  }
});
```

This helper is used for both the [leave-widget-balance.component.js](https://github.com/civicrm/civihr/blob/f05161cde174c7939a7a6a46ca656d8983c12029/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-widget/leave-widget-balance.component.spec.js) and [leave-widget-heatmap.component.spec.js](https://github.com/civicrm/civihr/blob/f05161cde174c7939a7a6a46ca656d8983c12029/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-widget/leave-widget-heatmap.component.spec.js) spec files, but it's encouraged to be used for all tests.


reneolivo:
## Updates

### Component renaming
Some components, and their tests and templates, were renamed after discussing them with @AkA84.

* `leave-widget-heatmap` ---> `leave-widget-absence-types-amount-taken`
* `leave-widget-balances` ---> `leave-widget-absence-types-available-balance.component`
* `week-heatmap` ---> `heatmap`

The code in `leave-widget-absence-types-amount-taken` was split. The absence types list remains in the component, but the logic to map leave requests to heatmap values was moved to a new component dubbed `leave-requests-heatmap`. This last component would map the leave requests values and them pass them to the `heatmap` component.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+        vm.leaveRequestStatuses && vm.leaveRequestStatuses.length;
+    }
+
+    /**
+     * Loads the status ids for absence types and leave requests for the contact.
+     * After loading the dependencies, it emits a child is ready event.
+     *
+     * @return {Promise}
+     */
+    function loadDependencies () {
```
AkA84: this function is

1. misleading in its name. It says that it loads the dependencies, but it also is in charge of emitting an event. It does too ,any things
2. Not really needed, you can simply do
    ```js
    function $onChanges () {
      loadLeaveRequests().then(function () {
        $scope.$emit('LeaveWidget::childIsReady');
      });
    }
    ```
    making `$onChanges` much easier to read

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+      return loadLeaveRequests()
+        .finally(function () {
```
AkA84: why `finally`? you want to emit the event even if there was an error?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+     * Loads all leave requests for the contact, in the current period, of the
+     * allowed statuses, and of the specific absence types.
+     *
+     * @return {Promise}
+     */
+    function loadLeaveRequests () {
+      var absenceTypeIds = vm.absenceTypes.map(function (absenceType) {
+        return absenceType.id;
+      });
+      var statusIds = vm.leaveRequestStatuses.map(function (status) {
```
AkA84: put these two in separate functions

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+        from_date: { '>=': vm.absencePeriod.start_date },
+        to_date: { '<=': vm.absencePeriod.end_date },
+        status_id: { IN: statusIds },
+        type_id: { IN: absenceTypeIds }
+      })
+      .then(function (response) {
+        vm.leaveRequests = response.list;
+
+        mapAbsenceTypesBalance();
+      });
```
AkA84:
```js
.then(function (response) {
  vm.leaveRequests = response.list;
})
.then(mapAbsenceTypesBalance);
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Finds and stores the balance for each absence type.
```
AkA84: document the use of `Math.abs`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+
+        balance = vm.leaveRequests.filter(function (request) {
+          return +request.type_id === +absenceType.id;
+        })
+        .reduce(function (balance, request) {
+          return balance + request.balance_change;
+        }, 0);
+        balance = Math.abs(balance);
+
+        return _.assign({ balance: balance }, absenceType);
```
AkA84:
```js
return _.assign({ balance: Math.abs(balance) }, absenceType);
```
?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
```diff
@@ -1,13 +1,29 @@
 /* eslint-env amd */
 
 define([
+  'common/angular',
   'common/lodash',
+  'common/bundles/components',
+  'common/modules/directives',
+  'common/modules/models',
```
AkA84: wait, this can't be right

1. You should not require the whole bundle in any circumstance. The bundle is used only for building reqangular.min.js, nothing else
2. You should not require the modules of `common/` either. You should just require what you actually need (a service, a component, a directive, etc) which in turn will require internally its own module

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
```diff
@@ -1,13 +1,29 @@
...
+  'common/filters/time-unit-applier.filter',
+  'leave-absences/shared/modules/shared-settings',
   'leave-absences/shared/modules/components',
   'leave-absences/shared/models/absence-period.model',
   'leave-absences/shared/models/absence-type.model',
-  './leave-widget-balance.component'
-], function (_, components) {
-  components.component('leaveWidget', {
+  './leave-widget-absence-types-amount-taken.component',
```
AkA84: why the `.`

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,94 @@
...
+    controllerAs: 'heatmap',
+    template: ['$templateCache', function ($templateCache) {
+      return $templateCache.get('components/heatmap.html');
+    }]
+  });
+
+  function heatmapController () {
+    var dayLabels = [
+      { label: 'Monday', shortLabel: 'M' },
+      { label: 'Tuesday', shortLabel: 'T' },
```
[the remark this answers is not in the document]
reneolivo: heatmap.html:
```html
<li ... title="{{day.label}}: {{day.value}}"> <-- here
    {{day.shortLabel}}
</li>
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,94 @@
...
+        { label: 'medium', value: average },
+        { label: 'high', value: _.max(arrayValues) }
+      ];
+
+      vm.days = dayLabels.map(function (dayLabel, day) {
+        var heat;
+        var value = vm.values[day];
+
+        if (value === false) {
+          return _.assign({ value: 0, heat: 'disabled' }, dayLabel);
```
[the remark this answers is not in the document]
reneolivo: adding it to the ranges can be dangerous, since values can be include `0`, it would display those values as disabled, and that's not the intention.

I'll try to refactor this code to make it more clear.

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        value = value || 0;
```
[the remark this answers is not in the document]
reneolivo: If an object like this is passed:

```js
{
  1: 100,
  2: 200,
  3: 300
}
```

it fills the missing day values with 0. This scenario is covered in the tests. When this code is split it will be more clear. Or else I'll add a comment.

org.civicrm.reqangular/test/components/heatmap.component.spec.js
```diff
@@ -0,0 +1,108 @@
...
+define([
+  'common/angular',
+  'common/lodash',
+  'common/components/heatmap.component'
+], function (angular, _) {
+  describe('Heatmap component', function () {
+    var $componentController, cmp;
+
+    beforeEach(function () {
+      module('common.components');
```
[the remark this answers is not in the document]
reneolivo: I usually do it like that, but other tests in reqangular follow this format. I'll double check, but I even think the format we normally use does not work (again, need to double check).

org.civicrm.reqangular/test/components/heatmap.component.spec.js
```diff
@@ -0,0 +1,108 @@
...
+          { value: 2, heat: 'low' },
+          { value: 2, heat: 'low' },
+          { value: 10, heat: 'high' },
+          { value: 6, heat: 'medium' },
+          { value: 0, heat: 'disabled' },
+          { value: 0, heat: 'disabled' }
+        ]);
+
+        cmp.values = values;
+        cmp.$onChanges({ values: { currentValue: values } });
```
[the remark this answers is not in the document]
reneolivo: `$onChanges` is triggered automatically when the app is running, but not when running tests. That's the reason for the helper function `controllerOnChanges` in L&A. 

Since this was a single component, I didn't make a generic `controllerOnChanges` that could be used in any extension. If the need to stub these calls we might consider using [angular-stub-changes](https://github.com/abyx/angular-stub-changes).

More info:
http://www.codelord.net/2017/01/17/testing-components-with-%24onchanges-using-angular-stub-changes/

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-requests-heatmap.component.js
```diff
@@ -0,0 +1,57 @@
...
+      if (changes.leaveRequests) {
+        mapLeaveRequestsToHeatmapValues();
+      }
+    }
+
+    /**
+     * Stores the total leave balance for each day of the week.
+     */
+    function mapLeaveRequestsToHeatmapValues () {
+      vm.heatmapValues = {};
```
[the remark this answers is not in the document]
reneolivo: I need to add `vm.heatmapValues = {}` to the component definition (before on init). But this line needs to stay because it clears previous heat map values in case the leave request bindings are updated.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-requests-heatmap.component.js
```diff
@@ -0,0 +1,57 @@
...
+
+      vm.leaveRequests.reduce(function (dates, request) {
+        return dates.concat(request.dates);
+      }, [])
+      .forEach(function (date) {
+        // 0 = Monday, 6 = Sunday:
+        var dayOfTheWeek = moment(date.date).isoWeekday() - 1;
```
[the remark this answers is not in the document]
reneolivo: returning Monday as 0? No, `moment().weekDay()` would return 0 for Sunday and 1 for Monday, `moment().isoWeekDay()` would return  7 for Sunday and 1 for Monday.

Potentially, `moment().weekDay()` could return 0 for Monday, but I'd need to change the locale, something I rather not do since that would affect the whole app.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
+/* eslint-env amd */
+
+define([
+  'common/lodash',
+  'common/moment',
+  'leave-absences/shared/modules/components',
+  'leave-absences/shared/components/leave-requests-heatmap.component',
+  'leave-absences/shared/models/leave-request.model'
```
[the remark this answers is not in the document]
reneolivo: what's included in the main file (leave-widget.component.js) are the modules for components and models. If I want to include a specific model/component I must specify it (can't just use `<leave-requests-heatmap></leave-request-heatmap>` for example).

This is good because it doesn't include ALL components or models, just the ones needed by the leave widget.

Also, I'd avoid importing the components or models (dependency) in the main file because if we remove the component that needs the dependency we risk including it even if it's not needed.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -0,0 +1,122 @@
...
+     */
+    (function init () {
+      $scope.$emit('LeaveWidget::childIsLoading');
+    })();
+
+    /**
+     * Implements the $onChanges method for angular controllers. When bindings
+     * are ready for use, it loads all dependencies.
+     */
+    function $onChanges () {
```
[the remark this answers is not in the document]
reneolivo: That approach is good when I want something to happen when ONE binding is updated, not so much for multiple bindings since they might not be passed at the same time (promises, etc.)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
reneolivo: yes, otherwise we risk blocking the whole leave widget. If one of the components doesn't fire the child is ready event, the leave widget will keep waiting for it forever.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
reneolivo: It clones the absence type and adds the balance property. Should I add a comment to that or use another approach?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
reneolivo: I think this part was giving me issues, I'll give it a try and let you know what the specific issue was (if any)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
reneolivo: requirejs allows you to include relative paths, which I'm using because it makes the leave widget more portable (if it needs to change places this line of code doesn't need to change).

I'm using `./component-name` instead of `component-name`, which would also work, to make it more explicit that the required component is in the same folder.

org.civicrm.reqangular/test/components/heatmap.component.spec.js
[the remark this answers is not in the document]
AkA84: if it doesn't work it would be quite weird and should be investigated

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
AkA84: no i mean, do `Math.abs` directly inside `_.assign`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
AkA84: if it was, then something was definitely wrong. The above should not be like that, no exception

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
AkA84: got it, but please stick with the convention that we have for now (= use `leave-absences/`)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-requests-heatmap.component.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    beforeEach(function () {
+      ctrl = $componentController('leaveRequestsHeatmap');
+      controllerOnChanges.setupController(ctrl);
+    });
+
+    it('must be defined', function () {
+      expect(ctrl).toBeDefined();
+    });
+
+    describe('when leve requests are bound', function () {
```
AkA84: typo

AkA84: also, when the leave requests are passed is more descriptive

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-requests-heatmap.component.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+          if (!expectedHeatMap[dayOfTheWeek]) {
+            expectedHeatMap[dayOfTheWeek] = 0;
+          }
+
+          expectedHeatMap[dayOfTheWeek]++;
+        });
+      });
+
+      it('transform leave requests into heatmap values', function () {
```
AkA84: transforms

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-widget/leave-widget-absence-types-amount-taken.html
```diff
@@ -0,0 +1,31 @@
...
+        <td class="leave-widget__sickness-type">
+          {{absenceType.title}}
+        </td>
+        <td class="leave-widget__sickness-balance">
+          {{absenceType.balance | timeUnitApplier:'days'}}
+        </td>
+      </tr>
+    </tbody>
+  </table>
+
```
AkA84: do not leave these empty lines in html files

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-widget/leave-widget.html
```diff
@@ -4,47 +4,21 @@ <h2 class="panel-title">Leave</h2>
...
-      <leave-widget-balance
+      <leave-widget-absence-types-available-balance
+        absence-period="leaveWidget.currentAbsencePeriod"
         absence-types="leaveWidget.absenceTypes"
         contact-id="leaveWidget.contactId"
-        current-absence-period="leaveWidget.currentAbsencePeriod"
-      ></leave-widget-balance>
+      ></leave-widget-absence-types-available-balance>
 
       <hr />
```
AkA84: remove these empty lines

mi-kon:
to reiterate @reneolivo - heatmap is very low priority. If fixing PR particularly for that bit takes too much time, i'd rather skip this bit for first release of leave widget

reneolivo:
@mi-kon most of the feedback is related to the heatmap yes, I'll timebox it to 1hr. If the fixes can't be completed I'll remove the heatmap from the component and leave the code so it can be fixed at another time.

Regardless, this will be the last thing I'll check for the whole PR.

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,94 @@
...
+     * heat range equivalent.
+     *
+     * The values are categorized into *low*, *medium*, *high*, and *disabled*
+     * (in case the value is a strict *false*). For the *low* and *high* range
+     * the minimum and maxim values are used. For the *medium* range the average
+     * for all the values is selected.
+     *
+     * A value of *false* is treated as *disabled*.
+     */
+    function mapValuesToDaysAndHeatRanges () {
```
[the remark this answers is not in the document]
reneolivo: I split the code, but the calls to the functions I left them inside this function instead of `$onChanges`, in case someone calls the function directly it executes all the functions it needs to run properly.

org.civicrm.reqangular/test/components/heatmap.component.spec.js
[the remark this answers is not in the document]
reneolivo: Angular Mocks needs to be included: `'common/angularMocks'`, it was missing.

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+     * heat level for the value.
+     *
+     * @param {Object[]} heatLevelRanges - An array of heat levels and their values.
+     * @param {Number} heatValue - The value to compare against heat levels.
+     *
+     * @param {String}
+     */
+    function getHeatLevelForValue (heatLevelRanges, heatValue) {
+      var heatRange = heatLevelRanges.reduce(function (previous, current) {
+        return Math.abs(current.value - heatValue) < Math.abs(previous.value - heatValue)
```
AkA84: Put `current.value - heatValue` and `previous.value - heatValue` in two variables with descriptive names, so the comparison is more comprehensible

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+      var heatValues = getCleanHeatValues();
+
+      average = Math.floor(_.sum(heatValues) / heatValues.length);
+      // Minimum is the lowest value of the 7 days, or 0 if less than 7 days:
+      minimum = heatValues.length < 7 ? 0 : _.min(heatValues);
+
+      return [
+        { level: 'low', value: minimum },
+        { level: 'medium', value: average },
+        { level: 'high', value: _.max(heatValues) }
```
AkA84: put the calculation for `minimum`, `low` and `high` in three separate functions, that you can then document, something like
```js
function lowHeatRange(values) {}
function mediumHeatRange(values) {}
function highHeatRange(values) {}
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+        mapValuesToHeatmapDays();
+      }
+    }
+
+    /**
+     * Returns a clean copy of the heatmap values that are not false.
+     *
+     * @return {Array}
+     */
+    function getCleanHeatValues () {
```
AkA84: the name is not descriptive, it's not clear what "clean" means
```js
filterOutFalseHeatValues()
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Maps a list of week day values into the days for the heatmap.
+     *
+     * The values are categorized into *low*, *medium*, *high*, and *disabled*.
+     * A value of *false* is treated as *disabled*.
+     */
+    function mapValuesToHeatmapDays () {
+      var heatLevelRanges = getHeatLevelRanges();
+
+      resetHeatmapDays();
```
AkA84: `mapValuesToHeatmapDays` is doing too many things. If you want to reset the values everytime they change, then do it in `$onChanges`
```js
function $onChanges (changes) {
  if (changes.values) {
    resetHeatmapDays();
    mapValuesToHeatmapDays();
  }
}
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Resets the heatmap days to their default values. This is done for each
+     * day of the week.
+     * _.range(1, 8) returns an array with the numbers from 1 to 7.
+     * .isoWeekday(1) will return Monday, 7 will return Sunday.
+     */
+    function resetHeatmapDays () {
+      vm.heatmapDays = _.range(1, 8).map(function (dayNumber) {
+        var dayLabel = moment().isoWeekday(dayNumber).format('dddd');
```
AkA84:
```js
vm.heatmapDays = _.range(1, 8).map(function (dayIndex) {
  var day = moment().isoWeekday(dayIndex);
  
  return {
    nameLong: day.format('dddd'),
    nameShort:  day.format('d'),
    heatLevel: 'low',
    heatValue: 0,
  }
}
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+
+        return {
```
AkA84: also what about grouping the properties to make the object more expressive
```js
{
  name: {
    short,
    long
  },
  heat: {
    level,
    value
  }
}

File: org.civicrm.reqangular/src/common/components/heatmap.component.js [hunk 29]
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Returns the heat low, medium, and high ranges for the heatmap.
+     * For the *low* and *high* range the minimum and maximum values are used.
+     * For the *medium* range the average for all the values is selected.
+     *
+     * @return {Object[]}
+     */
+    function getHeatLevelRanges () {
```
AkA84: `calculateHeatLevelRanges` is a better name

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+
+      return heatRange.level;
+    }
+
+    /**
+     * Returns the heat low, medium, and high ranges for the heatmap.
+     * For the *low* and *high* range the minimum and maximum values are used.
+     * For the *medium* range the average for all the values is selected.
+     *
+     * @return {Object[]}
```
AkA84: we don't use this syntax, just put `{Array}`. Please stick to conventions

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Maps a list of week day values into the days for the heatmap.
+     *
+     * The values are categorized into *low*, *medium*, *high*, and *disabled*.
+     * A value of *false* is treated as *disabled*.
+     */
+    function mapValuesToHeatmapDays () {
+      var heatLevelRanges = getHeatLevelRanges();
```
AkA84: Same thing with `resetHeatmapDays`, `mapValuesToHeatmapDays` is doing too many things:

1. It's resetting the heatmap days
2. It's calculating the heat level ranges
3. It's mapping the values to the heatmap days

It should do only one thing, the mapping (as the name says). The other stuff should happen when the values change

```js
function $onChanges (changes) {
  if (changes.values) {
    resetHeatmapDays();
    calculateHeatLevelRanges();
    mapValuesToHeatmapDays();
  }
}
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+      _.forEach(vm.values, function (heatValue, isoWeekDayIndex) {
+        var heatmapDay = vm.heatmapDays[isoWeekDayIndex - 1];
+
+        if (heatValue === false) {
+          heatmapDay.heatLevel = 'disabled';
+          return;
+        }
+
+        heatmapDay.heatValue = heatValue;
+        heatmapDay.heatLevel = getHeatLevelForValue(heatLevelRanges, heatValue);
```
AkA84: put `heatLevelRanges` in a controller's private variable, so that you don't have to pass it to the function (it will make more sense also when you move `calculateHeatLevelRanges` to `$onChanges`)

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -0,0 +1,131 @@
...
+      var heatLevelRanges = getHeatLevelRanges();
+
+      resetHeatmapDays();
+
+      _.forEach(vm.values, function (heatValue, isoWeekDayIndex) {
+        var heatmapDay = vm.heatmapDays[isoWeekDayIndex - 1];
+
+        if (heatValue === false) {
+          heatmapDay.heatLevel = 'disabled';
+          return;
```
AkA84: this `return` is pretty ugly here, this reads better
```js
if (heatValue === false) {
  heatmapDay.heatLevel = 'disabled';
} else {
  heatmapDay.heatValue = heatValue;
  heatmapDay.heatLevel = getHeatLevelForValue(heatLevelRanges, heatValue);
}
```

org.civicrm.reqangular/src/common/components/heatmap.component.js
AkA84: `calculateHeatLevelForValue` is more descriptive, cause that's what the function does

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-requests-heatmap.component.js
```diff
@@ -0,0 +1,69 @@
...
+
+        vm.heatmapValues[dayOfTheWeek]++;
+      });
+    }
+
+    /**
+     * Returns a single array of dates, extracted from each leave request
+     * dates.
+     *
+     * @return {String[]}
```
AkA84: again, don't use this syntax. It's {Array}

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-requests-heatmap.component.js
```diff
@@ -0,0 +1,69 @@
...
+     */
+    function datesOfLeaveRequests () {
```
AkA84: `leaveRequestsDates`

PR #2227, opened 2017-10-19
PCHR-2619: Generate user sub menu in hrcore

AkA84:
## Overview
This PR adds a new Civi Page, `CRM_HRCore_Page_UserMenu`, to the `uk.co.compucorp.civicrm.hrcore` extension which returns the user sub-menu that will be used both in the SSP and in the CiviHR admin.

### Path classes
The user submenu has 3 paths that can have different values depending on the current CMS

* Default user image
* Edit account link
* Logout link

For abstraction purposes, these differences had been encapsulated in `CRM_HRCore_CMSData_Paths` classes, one for each CMS (Drupal, Joomla, WordPress), whose job is to return the aforementioned paths. The classes adhere to `CRM_HRCore_CMSData_PathsInterface` interface to enforce consistency: 
```php
interface CRM_HRCore_CMSData_PathsInterface {
  public function getDefaultImagePath();
  public function getEditAccountPath();
  public function getLogoutPath();
}
```

The `CRM_HRCore_Page_UserMenu` page class instantiates a "path" class via a `CRM_HRCore_CMSData_PathsFactory` factory, to get the correct class based on the current CMS

```php
// UserMenu.php
private function instantiateCmsPaths() {
  $cmsName = CRM_Core_Config::singleton()->userFramework;

  $this->cmsPaths = PathsFactory::create($cmsName, $this->contactData);
}
```
the factory in turns returns the instance only after checking that the given CMS name is valid:
```php
// PathsFactory.php
public static function create($cmsName, $contactData) {
  $allowedCms = ['Drupal', 'WordPress', 'Joomla'];

  if (!in_array($cmsName, $allowedCms)) {
    throw new Exception("CMS \"{$cmsName}\" not recognized", 1);
  }

  $className = "CRM_HRCore_CMSData_Paths${cmsName}";

  return new $className($contactData);
}
```
### Template
The template of the menu is taken from [page.tpl.php](https://github.com/compucorp/civihr-employee-portal-theme/blob/staging/civihr_default_theme/templates/page/page.tpl.php#L81-L98)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+    $this->assign('username', $this->contactData['display_name']);
+    $this->assign('image', $this->getUserImagePath());
+    $this->assign('editLink', $this->cmsPaths->getEditAccountPath());
+    $this->assign('logoutLink', $this->cmsPaths->getLogoutPath());
+
+    return parent::run();
+  }
+
+  /**
+   * Gets the currently logged in contact's data, including the
```
omarabuhussein: typo : including the her user --> should be --> including the user . would be better to specify the returned array format in @return array (ideally an object should be returned but it is not possible here)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+   * her user id in the CMS
+   *
+   * @return array
+   */
+  private function getContactData() {
```
omarabuhussein: as davi said before, this method is called get but it actually set data, you should rename it to setContactData or just make it return data and keep the name as is, I suggest to rename it to setContactData

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+    $rawContactData = civicrm_api('Contact', 'getsingle', array(
+      'version' => 3,
+      'return' => array('id', 'display_name', 'image_URL'),
+      'id' => CRM_Core_Session::getLoggedInContactID(),
+      'api.User.getsingle' => array('contact_id' => "\$value.contact_id")
+    ));
+
+    $this->contactData = $this->normalizeContactDataAPIResponse($rawContactData);
```
omarabuhussein: this method is better to be used near to were it is defined, so I would rather define normalizeContactDataAPIResponse directly below getContactData to avoid jumping while reading the code.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Gets the currently logged in contact's data, including the
+   * her user id in the CMS
+   *
+   * @return array
+   */
+  private function getContactData() {
+    $rawContactData = civicrm_api('Contact', 'getsingle', array(
```
omarabuhussein: getsingle will through an exception if it found nothing, so I would either catch the exception if it happen, but this will add make the code more nested so I would choose get instead of getsingle and add limit => 1 (but the output will be a bit different)

also please use [] instead of array().

also why not to use civicrm_api3 instead of passing the version.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+      'version' => 3,
+      'return' => array('id', 'display_name', 'image_URL'),
+      'id' => CRM_Core_Session::getLoggedInContactID(),
+      'api.User.getsingle' => array('contact_id' => "\$value.contact_id")
```
omarabuhussein: same apply to getsingle in the chained API call, I would use get and limit => 1 instead.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+use CRM_HRCore_CMSData_PathsFactory as PathsFactory;
```
omarabuhussein: CMSPathsFactory is more clear name, I think.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/PathsDrupal.php
```diff
@@ -0,0 +1,27 @@
+<?php
+  class CRM_HRCore_CMSData_PathsDrupal implements CRM_HRCore_CMSData_PathsInterface {
+    private $contactData;
+    private $paths = [
+      'defaultImage' =>  '/%{base}/images/profile-default.png',
```
omarabuhussein: also there is no benefits from having $paths as class property, it is not something shared between the class methods so I would rather remove it and and define each path as a separate constant instead.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  /**
+   * Instantiates the paths class of the current CMS
+   */
+  private function instantiateCmsPaths() {
+    $cmsName = CRM_Core_Config::singleton()->userFramework;
+
+    $this->cmsPaths = PathsFactory::create($cmsName, $this->contactData);
```
davialexandre: This method uses `$this->contactData` which is only set when you call `getContactData()`, so you have a hidden dependency here. Instead of accessing the property directly, use the method. If you're concerned about making an API call everytime you call if, you can implement it like this:

```php
public function getContactData() {
    if(!empty($this->contactData)) {
       $this->contactData = getDataFromAPI();
    }

    return $this->contactData;
}
```

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
...
+class CRM_HRCore_Page_UserMenu extends CRM_Core_Page {
+
+  private $contactData;
+  private $cmsPaths;
+
+  public function run() {
+    $this->getContactData();
+    $this->instantiateCmsPaths();
+
+    $this->assign('username', $this->contactData['display_name']);
```
davialexandre: you should use the getter here instead of accessing the class property directly

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+use CRM_HRCore_CMSData_PathsFactory as PathsFactory;
+
+class CRM_HRCore_Page_UserMenu extends CRM_Core_Page {
+
+  private $contactData;
+  private $cmsPaths;
```
davialexandre: Please don't forget to document these properties too

omarabuhussein:
@AkA84  I forget to mention this, can you add unit tests to CRM_HRCore_CMSData_PathsDrupal  and CRM_HRCore_CMSData_PathsFactory  classes too

uk.co.compucorp.civicrm.hrcore/xml/Menu/hrcore.xml
```diff
@@ -36,4 +36,10 @@
     <title>Create User Account</title>
     <access_arguments>create users</access_arguments>
   </item>
+  <item>
```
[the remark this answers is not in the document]
AkA84: @omarabuhussein this page will indeed be called via ajax in the CiviHR admin, as the user submenu will be injected via JS using the hrui extension, so that we don't have to touch the core files

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/CMSPathsFactory.php
```diff
@@ -0,0 +1,26 @@
...
+   *
+   * @param string $cmsName
+   * @param array $contactData
+   *
+   * @return object
+   *
+   * @throws Exception if the CMS is not recognized
+   */
+  public static function create($cmsName, $contactData) {
+    $allowedCms = ['Drupal', 'WordPress', 'Joomla'];
```
davialexandre: Since you kept only the Drupal class, it's better to remove the references to other CMSes in here

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/PathsDrupal.php
```diff
@@ -3,22 +3,26 @@
...
-   * @var array
+   * @const string
    */
-  private $contactData;
+  const DEFAULT_IMAGE_PATH = '/%{base}/images/profile-default.png';
+
+  /**
+   * @const string
+   */
+  const EDIT_PATH = '/user/%{userId}/edit';
```
omarabuhussein: EDIT_USER_PATH is more clear name

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/PathsDrupal.php
```diff
@@ -3,22 +3,26 @@
 class CRM_HRCore_CMSData_PathsDrupal implements CRM_HRCore_CMSData_PathsInterface {
 
   /**
-   * The contact data used to build the paths
-   *
-   * @var array
+   * @const string
    */
-  private $contactData;
+  const DEFAULT_IMAGE_PATH = '/%{base}/images/profile-default.png';
```
omarabuhussein: DEFAULT_USER_IMAGE_PATH  is more clear name

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+use CRM_HRCore_CMSData_CMSPathsFactory as CMSPathsFactory;
```
omarabuhussein: sorry I think I wasn't clear, but what I meant here https://github.com/civicrm/civihr/pull/2227#discussion_r145736349 

is just to rename the alias to contain CMS so "CMSPathsFactory" instead of "PathsFactory" as you did here but not changing the actual class name from "CRM_HRCore_CMSData_PathsFactory" to "CRM_HRCore_CMSData_CMSPathsFactory " since CMS is repeated twice here.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/Paths/DrupalTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+  public function testItReturnsTheDrupalLogoutLink() {
+    $this->assertEquals($this->drupalPaths->getLogoutPath(), '/user/logout');
+  }
+}
+
+/**
+ * Mock of the original drupal function
+ *
+ * @return string}
```
omarabuhussein: typo : remove } from string}

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/PathsFactoryTest.php
```diff
@@ -0,0 +1,36 @@
...
+    $contactData = [];
+    $pathsClass = CMSPathsFactory::create('Drupal', $contactData);
+
+    $this->assertTrue($pathsClass instanceof CRM_HRCore_CMSData_PathsInterface);
+  }
+
+  public function testItThrowsAnExceptionIfGivenAnUnrecognizedCMSName() {
+    $cmsName = 'UnrecognizedCMS';
+
+    $this->setExpectedException('Exception', "CMS \"{$cmsName}\" not recognized");
```
omarabuhussein: this method is  removed in newer phpunit versions, you need to use annotation instead : https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.expectedException

davialexandre: It wasn't removed it was just renamed to `expectedException()`: https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

But yeah, I this specific case, an annotation would work better. Usually, we use `setExpectedException` in there is more than one place in the code that can throw the exception and we want to make sure the test will catch the right one

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/PathsFactoryTest.php
```diff
@@ -0,0 +1,36 @@
...
+    return \Civi\Test::headless()
+      ->installMe(__DIR__)
+      ->apply();
+  }
+
+  public function testItReturnsAClassWithImplementingTheExpectedInterface() {
+    $contactData = [];
+    $pathsClass = CMSPathsFactory::create('Drupal', $contactData);
+
+    $this->assertTrue($pathsClass instanceof CRM_HRCore_CMSData_PathsInterface);
```
davialexandre: Please use this instead: 
```php
$this->assertInstanceOf(CRM_HRCore_CMSData_PathsInterface::class, $pathsClass);
```

PR #2234, opened 2017-10-23
PCHR-2813: Fix PHPUnit test errors

davialexandre:
This PR fixes several issues with running the PHP unit tests in Jenkins. Following is a description of each of the problems and their respective solutions.

### Ignored PHPUnit errors

Problem: Sometimes, due to unexpected issues (like missing classes or a syntax error in one of the tests, for example) the PHPUnit command might fail without even starting to run the tests. However, the command used to run the tests was working in a way that it would always return a success code, regardless of any error, which would make Jenkins mark the build as successful.

Solution: Change the command used by Jenkins. Basically, it was changed from `phpunit4 || true` to simply `phpunit4`

### HRReport errors

Problem: the phpunit4 command was failing with the following error:

```
PHP Fatal error: Call to undefined function _hrjob_phpunit_populateDB() in /opt/buildkit/build/hr17-dev_PR-2225/sites/all/modules/civicrm/tools/extensions/civihr/hrreport/tests/phpunit/CRM/HRReport/Form/Contact/HRDetailTest.php on line 62
```

Solution: All the tests in this extension were removed. These are some pretty old tests, that were failing because they reference some functions from the hrjob extension, which we don't use anymore. Besides that, the tests were not relevant anymore, as they test some old reports that don't work anymore and will be removed in the future.

### HRUI errors

Problem: The phpunit4 command was failing with the following error:

```
aught CiviCRM_API3_Exception: [2001: 'task_status' is not a valid option for field option_group_id
thrown in /opt/buildkit/build/hr17-dev_PR-2225/sites/all/modules/civicrm/api/api.php on line 45
```

This extension had a very bizarre dependency tree. It had a single test, for a helper class that was used to get a list of managers. A manager was defined by the "Line Manager is" relationship type, which was defined by the `HRCase` extension. Unfortunately, `HRCase` depends on T&A. Finally, in addition to all that, the test was also using fabricators from other extensions. This all together was making it necessary to install lots of extensions to test a single method. To make things even worse, the test was failing because an Option Group created by one of these extensions could be found in the database.

Solution: To reduce the dependency tree complexity and also as one small step towards having a single CiviHR extension, I did two things:
- Made HRCore the owner of the "Liner Manager is" relationship type. The extension will now be responsible to create it on new sites and delete it from existing ones, if necessary
-  Moved the helper to HRCore. It's now a service called `CRM_HRCore_Service_Manager`

### HRVisa errors

Problem: Tests are still showing information about deprecated methods (`CRM_Core_OptionGroup::getValue()`). Also, there are some notices about undefined variables

Solution: Replace all the calls to the deprecated method with call to helper methods that would return the same information. This also fixed the notices.

### HR Core

Problem: Tests were failing with the following error:

```
CiviCRM_API3_Exception: API (TASettings, get) does not exist (join the API team and implement it!) in civicrm_api3() (line 45 of /opt/buildkit/build/hr17-dev_PR-2225/sites/all/modules/civicrm/api/api.php).
```

[Here](https://github.com/civicrm/civihr/commit/001c2d37a7125bb72ee49307144c3a458db16a0f#diff-66261b0096e4a29163b58f8e2c2a966eR8) the PHPUnit bootstrap file of that extension was updated to bootstrap Drupal during the tests. The reason was that some new tests were added to test classes that wrap some Drupal functions and, when they were executed, these functions would be called and the test would fail, because said functions weren't available.

However, when Drupal gets bootstraped, the civihr_employee_portal module does a lot of initialization, including setting some variables with settings from T&A. To do this, it uses some of the T&A APIs. Unforturnately, in the test enviroment, T&A isn't installed and the API call will fail, throwing an error that will finally prevent the tests to run.

Solution: Mocks for the drupal functions were created. They are all inside the `drupal_function_mocks.php` file, that is included only once the tests are being executed. Note that this is not a perfect solution. Not all the Drupal functions can be mocked as easily as those and it might happen that difference tests will need mocks for the same function, but with a different behavior. This was simple and quick solution to allow the tests to be run in Jenkins, but in the future we'll probably need to revisit this and work on a stronger solution.

### Others fixes and improvements

- There's a small fix to the L&A sample data. The Brough Forward for the 2017 period had an expiry date in 2016. It was updated to 2017
- Except for hrrecruitment, which has tests still using the `CiviUnitTestCase` class, all the other extensions have a `BaseHeadlessTest` class that all of its tests extend from. The main reason for this is to make sure all of them use the same `setUpHeadless()` method, which avoids the db schema being reset unnecessarily. This make the tests run a bit faster (According to Jenkins ~1.5 min faster)
- Add more granular tests suites to all the extensions. This was copied from the L&A extension. All the extensions with PHPUnit tests now have 3 test suites:  `Full` (which includes all the tests), `Unit Tests` (which includes only the unit tests. That its, those inside phpunit/api and phpunit/CRM) and `Web Tests` (which includes the Web/Selenium Tests. That is, those inside phpunit/WebTests). Jenkins was updated to only run the `Unit Tests` test suite by default. The reason for this is that, currently, our Web Tests are a mess and Jenkins is not prepared to run them.
- Add a command to cleanup test reports from previous builds. It turns out that Jenkins wasn't doing this automatically. On a scenario where we build the same PR multiple times, I was possible that reports from a previous build would affect the results of the most recent one (for example, I removed the HRReport tests, but they were still being displayed on the final test report)


hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+        return $this->callAPISuccessGetValue($entity, $params, $checkAgainst);
+
+      case 'getsingle':
+        return $this->callAPISuccessGetSingle($entity, $params, $checkAgainst);
+
+      case 'getcount':
+        return $this->callAPISuccessGetCount($entity, $params, $checkAgainst);
+    }
+    $result = civicrm_api3($entity, $action, $params);
+    $this->assertAPISuccess($result, "Failure in api call for $entity $action");
```
tunbola: Please leave a line before the return statement.

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+    $result = civicrm_api3($entity, 'getvalue', $params);
+    if ($type) {
+      if ($type == 'integer') {
+        // api seems to return integers as strings
+        $this->assertTrue(is_numeric($result), "expected a numeric value but got " . print_r($result, 1));
+      }
+      else {
+        $this->assertType($type, $result, "returned result should have been of type $type but was ");
+      }
+    }
```
tunbola: An empty line before the return.

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+      'debug' => 1,
+    ];
+    $result = civicrm_api3($entity, 'getsingle', $params);
+    if (!is_array($result) || !empty($result['is_error']) || isset($result['values'])) {
+      throw new Exception('Invalid getsingle result' . print_r($result, TRUE));
+    }
+    if ($checkAgainst) {
+      // @todo - have gone with the fn that unsets id? should we check id?
+      $this->checkArrayEquals($result, $checkAgainst);
+    }
```
tunbola: Empty line before return.

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+    $params += [
+      'debug' => 1,
+    ];
+    $result = civicrm_api3($entity, 'getcount', $params);
+    if (!is_int($result) || !empty($result['is_error']) || isset($result['values'])) {
+      throw new Exception('Invalid getcount result : ' . print_r($result, TRUE) . " type :" . gettype($result));
+    }
+    if (is_int($count)) {
+      $this->assertEquals($count, $result, "incorrect count returned from $entity getcount");
+    }
```
tunbola: Empty line before return.

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+   *   - array
+   *   - object
+   *
+   * @throws Exception
+   * @return array|int
+   */
+  public function callAPISuccessGetSingle($entity, $params, $checkAgainst = NULL) {
+    $params += [
+      'debug' => 1,
+    ];
```
tunbola: Why not merge this with the params as done in `callAPISuccess` for consistency.

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+   * @param array $checkAgainst
+   *   Array to compare result against.
+   *   - boolean
+   *   - integer
+   *   - double
+   *   - string
+   *   - array
+   *   - object
+   *
+   * @throws Exception
```
tunbola: Please leave a line before `@return`

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+    return $result;
+  }
+
+  /**
+   * This function exists to wrap api getValue function & check the result
+   * so we can ensure they succeed & throw exceptions without litterering the test with checks
+   * There is a type check in this
+   * @param string $entity
+   * @param array $params
+   * @param null $count
```
tunbola: A line here and also before `@return`
Also param type for count should be `null|int`

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+   *   for getsingle the array is compared against an array passed in - the id is not compared (for
+   *   better or worse )
+   *
+   * @return array|int
+   */
+  public function callAPISuccess($entity, $action, $params, $checkAgainst = NULL) {
+    $params = array_merge(['debug' => 1], $params);
+    switch (strtolower($action)) {
+      case 'getvalue':
+        return $this->callAPISuccessGetValue($entity, $params, $checkAgainst);
```
tunbola: If these functions are public methods that can be called directly, Why do we have them in the `callAPISuccess` function also ?  I initially thought they were private methods and would be exposed via `callAPISuccess`

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+      'debug' => 1,
+    ];
+    $result = civicrm_api3($entity, 'getvalue', $params);
+    if ($type) {
+      if ($type == 'integer') {
+        // api seems to return integers as strings
+        $this->assertTrue(is_numeric($result), "expected a numeric value but got " . print_r($result, 1));
+      }
+      else {
+        $this->assertType($type, $result, "returned result should have been of type $type but was ");
```
tunbola: The statement doesn't seem complete. Seems we need to append the type of the result to the message.
Also why do we need a separate logic for `integer`, wont this statement catch it?

hrjobcontract/CRM/Hrjobcontract/Test/BaseHeadlessTest.php
```diff
@@ -0,0 +1,164 @@
...
+    $params += [
+      'debug' => 1,
+    ];
+    $result = civicrm_api3($entity, 'getsingle', $params);
+    if (!is_array($result) || !empty($result['is_error']) || isset($result['values'])) {
+      throw new Exception('Invalid getsingle result' . print_r($result, TRUE));
+    }
+    if ($checkAgainst) {
+      // @todo - have gone with the fn that unsets id? should we check id?
+      $this->checkArrayEquals($result, $checkAgainst);
```
tunbola: Interesting. Is this a Phpunit native function? Can't seem to find it.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -622,8 +608,8 @@ private function validateResult($contactID, $entity = NULL)  {
   }
 
   private function createTestContact($params)  {
-    $contactID = $this->individualCreate($params);
-    return $contactID;
+    $contact = ContactFabricator::fabricate($params);
```
tunbola: Empty line before return.

hrjobcontract/tests/phpunit/CRM/Hrjobcontract/Import/Parser/ApiTest.php
```diff
@@ -583,47 +583,48 @@ private function runImport($params)  {
...
+          $jobHealth = civicrm_api3($entity, 'getsingle', ['jobcontract_revision_id' => $revisionID]);
+          $this->assertEquals($this->_insurancePlanTypes[0]['value'], $jobHealth['plan_type']);
+          $this->assertEquals($this->_insurancePlanTypes[1]['value'], $jobHealth['plan_type_life_insurance']);
           break;
 
         default:
-          $this->callAPISuccessGetSingle($entity, array('jobcontract_revision_id'=>$revisionID));
+          $result = civicrm_api3($entity, 'getsingle', ['jobcontract_revision_id' => $revisionID]);
+          $this->assertNotEmpty($result);
+          $this->assertFalse(array_key_exists('is_error', $result));
```
tunbola: Why not `assertArrayNotHasKey()`.

PR #2235, opened 2017-10-23
PCHR-2620: Implement user menu in CiviHR admin

AkA84:
## Overview
This PR implements the user menu generated in the `uk.co.compucorp.civicrm.hrcore` extension (see #2227) in the main menu of the CiviHR admin

## Before
no-user-menu-in-admin

## After
user-menu-in-admin

### Injecting the markup via ajax
To avoid changing the core files, the user menu is injected via ajax by the `org.civicrm.hrui` extension. The path is the one defined in #2227

```js
function addUserMenuToMainMenu () {
  $.ajax('/civicrm/hrcore/usermenu?snippet=4', {
    dataType: 'html',
    success: injectUserMenuInAMainMenuWrapper
  });
}
```
The `snippet=4` parameter is necessary for a `CRM_Core_Page` to return the page without any wrapping html, that would be otherwise added by default. As for why the value of the parameter needs to be 4, i have no idea...I got the value by reverse-engineering the core code, see [here](https://github.com/civicrm/civicrm-core/blob/master/templates/CRM/common/snippet.tpl#L30-L35)

### Amending the outer main menu markup
The original main menu is not contained in any parent element, it's directly placed between the page element and the dropdow menus container
```html
<div id="page"></div>
<ul id="civicrm-menu"></ul>
<div id="root-menu-div"></div>
```
This needed to be modified in order to correctly laying out the user menu near the main menu and to apply custom styles to the whole block, so `org.civicrm.hrui` , after fetching the menu via ajax, modifies the markup so that `#civicrm-menu` is contained inside a `#civihr-menu` element, which contains the user menu as well
```html
<div id="page"></div>
<div id="civihr-menu">
  <ul id="civicrm-menu"></ul>
  <div class="chr_user-menu"></div>
</div>
<div id="root-menu-div"></div>
``` 

### Porting user menu and profile card styles from the SSP
The styles of the two component contained in the user menu markup, the [user menu](https://github.com/compucorp/civihr-employee-portal-theme/blob/staging/civihr_default_theme/scss/components/_header.scss#L179) itself and the ["profile card"](https://github.com/compucorp/civihr-employee-portal-theme/blob/staging/civihr_default_theme/scss/components/_profile-card.scss) (the photo + username component) had been ported over from the SSP to `org.civicrm.bootstrapcivihr`, since now they should belong to CiviHR (given that the markup comes from `uk.co.compucorp.civicrm.hrcore`).

The original styles in SSP will be deleted in a future PR

### Adding a `overrides/crm` folder
A new `overrides/crm` folder has been added, where any style related to original civicrm elements that need to be customized in CiviHR is placed (as opposed to `overrides/style` where bootstrap elements are overridden):
```
scss/
└── overrides/
       ├── crm/
       ├──  ├── _civicrm-menu.scss
       ├──  └── _link-home.scss
       ├── style/
       └── variables.scss
```


uk.co.compucorp.civicrm.hrcore/templates/CRM/HRCore/Page/UserMenu.tpl
```diff
@@ -1,28 +1,30 @@
-<div class="chr_header__user-menu">
-  <div class="chr_header__user-menu__data">
-    <span class="chr_header__user-menu__name">{$username}</span>
+<div class="chr_user-menu">
```
deb1990: 1.Why is this changed? Cant we use previous values? 
2. And if we are changing the name, we should remove any existing styles for those classes.  Could you check please.

AkA84: 1. it's a more generic name, that fits both the SSP and the CiviHR admin (the same component will be used in both places). `chr_header` is a component in the SSP, but it doesn't exist in the admin, hence the removal
2. As I wrote in the PR description, I will be removing the styles referring the old class name in a future PR
    > The original styles in SSP will be deleted in a future PR

PR #2236, opened 2017-10-23
PCHR-2621: Implement user menu in SSP

AkA84:
## Overview
This PR applies amendments to the solution first introduced by #2227 , so that the same user menu markup can also be used in the SSP

Related PRs: 
* https://github.com/compucorp/civihr-employee-portal/pull/370
* https://github.com/compucorp/civihr-employee-portal-theme/pull/250

### `CRM_HRCore_UserMenuMarkup`
Instead of the `CRM_HRCore_Page_UserMenu` page class, the owner of the user menu markup (and the one responsible for processing it with the correct data) is now a new class `CRM_HRCore_UserMenuMarkup `

The class uses an internal instance of Smarty to process its own template, which is placed in a location that can be found by parsing the name of the class itself (just like it works for the standard CiviCRM pages). The implementation and the markup are still the same as they were before.
```php
public function __construct() {
  $this->smarty = CRM_Core_Smarty::singleton();
  $this->setTemplatePath();
}
```
All that the `CRM_HRCore_Page_UserMenu` class needs to do, is to simply use the class and put the given markup in the placeholder of an empty template
```php
class CRM_HRCore_Page_UserMenu extends CRM_Core_Page {
  public function run() {
    $this->assign('userMenuMarkup', (new CRM_HRCore_UserMenuMarkup())->getMarkup());
  }
}
```
```html
<!-- UserMenu.tpl -->
{$userMenuMarkup}
```

### Using the "outside namespace marker" outside individual partials
The individual partials were using `#{$outsideNamespaceMarker} ` directly within them, but that would have made importing those same partials in the SSP theme awkward (the SSP theme doesn't use that marker at all, so we should have had to define an empty variable to still successfully compile the SASS files)

```scss
// _civihr-menu.scss
#{$outsideNamespaceMarker} {
  #civihr-menu {
  }
}
```

So now the marker is used in the main `civihr.scss` file only, leaving the individual partials without it and safe to import somewhere else
```scss
// civihr.scss
#{$outsideNamespaceMarker} {
  @import "components/outside-namespace/*";
  @import "overrides/crm/*";
}
```
```scss
// _civihr-menu.scss
#civihr-menu {
}
```


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -1,102 +1,13 @@
 <?php
 
-use CRM_HRCore_CMSData_PathsFactory as CMSPathsFactory;
-
 class CRM_HRCore_Page_UserMenu extends CRM_Core_Page {
```
davialexandre: Just to confirm, we still need this page because on the admin you'll load the markup using Ajax, right?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
```diff
@@ -1,102 +1,13 @@
...
    * {@inheritdoc}
    */
   public function run() {
-    $this->instantiateCmsPaths();
-
-    $this->assign('username', $this->contactData()['display_name']);
-    $this->assign('image', $this->getUserImagePath());
-    $this->assign('editLink', $this->cmsPaths->getEditAccountPath());
-    $this->assign('logoutLink', $this->cmsPaths->getLogoutPath());
+    $this->assign('userMenuMarkup', (new CRM_HRCore_UserMenuMarkup())->getMarkup());
```
davialexandre: In PHP, you can have a string representation of a class. That is, if you "echo" an object, PHP will print this string representation. This is done by implementing the `__toString()` method in a class. It means that, if you rename this `getMarkup()` method to `__toString()`, you can change this line to simply:

```php
$this->assign('userMenuMarkup', new CRM_HRCore_UserMenuMarkup());
```

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
AkA84: Yep

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Page/UserMenu.php
AkA84: will it be understandable what's going on though? Is it common to do it?

davialexandre: I think it's common in cases we objects represent UI elements. As for understandability, I think it's fine as this is not a hidden secret of PHP or something, but it might be better to ask other devs about this. Or, if you think it's too much, then it's ok to leave it as it is. I really have no strong opinions on this

AkA84: if it's ok with you i'd leave it as is, especially cause we could have other non-backend devs working on it

org.civicrm.bootstrapcivihr/scss/overrides/_variables.scss
```diff
@@ -9,13 +9,14 @@ $outsideNamespaceMarker: '.___outside-namespace';
...
-$user-menu-color: $gray-darker;
-$user-menu-line-height: 60px;
-$user-menu-minify-breakpoint: 1250px;
-$user-menu-bg-hover: $gray-lighter;
-$user-menu-dropdown-bg: $crm-white;
-$user-menu-dropdown-link-color: $gray-darker;
-$user-menu-dropdown-link-bg-hover: $gray-lighter;
+$chr-user-menu-color: $gray-darker;
+$chr-user-menu-line-height: 60px;
+$chr-user-menu-minify-breakpoint: 1250px;
```
reneolivo: You don't need to specify this variable since you are already importing the SSP theme and it's defined there.

AkA84: not sure what you mean, i'm not importing the ssp theme here at all

PR #2238, opened 2017-10-24
PCHR-1849: Contract does not use currency settings from CiviCRM

bmnepali:
## Overview
Currently, the amount in pay tab fields in Job Contract form is not formatted based on decimal and thousand separators defined in settings. This PR adds the functionality to format and calculate the amounts in the respective fields in the job contract modal.

## Before
* Note: Defined separators ins settings:
- decimal=> " , "
- thousand => " ' "

1. When added the amount in pay tab fields with different characters as separators and tried to save the contract, the error displays saying the format is not valid.
oct-31-2017 20-30-04-iloveimg-compressed

## After
1. When the user inputs the value in respective field, the amount is automatically formatted using the defined decimal and thousand separators for the settings. As a result the validation passes and contract can be saved.
oct-31-2017 15-33-36-min

2. After reopened fix:
1849-final-min

#### A: Changes in reqangular repo:
1. Divided  the `Setting` call into mock, api , model and instance.
- `/test/mocks/services/api/settings-mock.js`
- `/src/common/services/api/settings.api.js`
- `/src/common/models/settings.model.js`
- `/src/common/models/instances/settings-instance.js`

2. Create a service called `FormatCurrencyService` in `reqangular` repo, which will be responsible to get  both the formatted and unformatted amount. Ref: `common/services/format-currency.service.js`.
```js
define([
  'common/modules/services',
  'common/services/api',
  'common/models/settings.model'
], function (services) {
  'use strict';

  services.factory('FormatCurrencyService', FormatCurrencyService);
  FormatCurrencyService.$inject = ['$log', '$q', 'Settings'];
  function FormatCurrencyService ($log, $q, Settings) {
    $log.debug('Service: FormatCurrencyService');

    return {
      format: format
    };

    /**
    * Adds defined thousand and decimal separators in the given amount
    *
    * @param {string} amount
    * @return {promise}
    */
    function format (amount) {
      return Settings.fetchSeparators().then(function (separators) {
        return {
          formatted: addSeparators(amount, separators),
          parsed: removeSeparators(amount, separators)
        };
      });
    }
    ...
  }
});
```
##### Inside `FormatCurrencyService`:
1.1 Create a function `prepareRegexForSeparator()` to generate the valid regex expression for  removing all non numeric characters except defined decimal separator in special cases.:
```js
function prepareRegexForSeparator (decimalSeparator) {
      var escape = ['\\', ']'];

      _.includes(escape, decimalSeparator) && (decimalSeparator = '\\' + decimalSeparator);

      return '[^0-9' + decimalSeparator + ']';
}
```

1.2 Create a function `replaceDecimalSeparator` which will accept the string or number and it will be converted into the amount with respective decimal defined in settings.
In summary: It will convert the calculable amount to the amount with correct decimal separator.
```js
/**
     * Replaces the decimal separator "." by defined decimal separator
     *
     * Add given amount is floating point value with decimal separator as "." ,
     * we need to change "." into defined decimal separator for proper formatting
     *
     * @param {String|Number} amount floating point number  (eg. 123213.98 or 123213.98)
     * @param {String} decimalSeparator defined decimal separator
     * @returns {String}
     */
    function replaceDecimalSeparator (amount, decimalSeparator) {
      amount = amount.toString();

      if (+amount && amount.indexOf('.') !== -1) {
        amount = amount.replace('.', decimalSeparator);
      }

      return amount;
    }
```

And  `addThousandSeparator()` function which will be responsible for adding thousand separators to the given amount and returning the final formatted value with 2 numbers after decimal separator.
```js
    /**
     * Add thousand separators to the given amount.
     *
     * @param {String} amount
     * @return {String} amount separated with correct separators
     */
    function addThousandSeparator (amount, separators) {
      var amountAfterDecimal;
      var amountBeforeDecimal;
      var decimalIndex;
      var expression;

      expression = prepareRegexForSeparator(separators.decimal);
      // Remove all characters except decimal separator
      amount = amount.toString().replace(new RegExp(expression, 'g'), '');

      // Find the index of first ocuring decimal separator
      decimalIndex = amount.indexOf(separators.decimal);

      if (decimalIndex !== -1) {
        // Hold the whole number before decimal separator
        amountBeforeDecimal = +(amount.substring(0, decimalIndex));
        // Remove all non numeric characters form values after decimal separator
        amountAfterDecimal = amount.substring(decimalIndex + 1).replace(/[^0-9]/g, '');
      } else {
        amountBeforeDecimal = +(amount.replace(/[^0-9]/g, ''));
      }

      // Add thousand separators to the value before decimal separator
      // converting it to string
      amountBeforeDecimal = amountBeforeDecimal.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1' + separators.thousand);

      // if amountAfterDecimal exists, concat amountBeforeDecimal and amountAfterDecimal by separator
      // keeping only two values after decimal separator
      return (decimalIndex !== -1) ? (amountBeforeDecimal.toString() + separators.decimal + amountAfterDecimal.substring(0, 2)) : amountBeforeDecimal;
}
```
**What it does?**
- Takes a ** `string` or `integer` or `floating point` value  and defined separators in settings.
- If the input value is `floating point` value, the decimal separator is converted to that defined in settings.
- Then it separates the given amount into amount before and after decimal separator.  
- Then add thousand separators to the amount before decimal separator and removes any character in the amount after decimal separator.
- Finally returns the formatted amount concatenating the before and the after amount based on wether the decimal separator is present in input value.

**What do we need it?**
- We need the amount to be saved with separators defined in settings and the amount to be displayed in pay fields. 

#### Important Functions:
1.3 Create a function to remove non numerical characters except the characters defined for decimal separator and making the value calculable by replacing the decimal separator by `.` and keeping the 2 values after the decimal separator.
```js
function removeSeparators (amount, separators) {
      var expression = configureExpression(separators.decimal);
      var strippedAmount = amount.toString().replace(new RegExp(expression, 'g'), '');
      var amountWithDecimal = strippedAmount.replace(separators.decimal, '.');

      return parseFloat(amountWithDecimal).toFixed(2);
}
```
**What it does?**
- Takes a amount containing separators, then **it removes all the separators except the decimal separator defined in settings.**
- Then **it replaces the defined decimal separator with “.” with 2 values after decimal to make it calculable and used in calculations.** 
_Because for example if **268'462'765,89** is a formatted amount and when we want this value in calculation. We can't use it as it's a string and it needs to be converted to `268462765.89`_

**What do we need it?**
- We need the value for calculation in format (732648, 498141.89) as the formatted numbers with separators will not be calculable numbers.

#### B. Changes in hrjobcontract repo:
1. Create following variables to be used in calculation in pay form in jobcontract modal in `controllers/form/form-pay.js`.
```js
      var calcBenefitsPerCycleNet = 0;
      var calcPayAnnualizedEst = 0;
      var calcPayPerCycleGross = 0;
```

2. Add the `formatPayAmount` function to format the pay amount in pay amount field.
```js
function formatPayAmount () {
        FormatCurrencyService.format(entityPay.pay_amount).then(function (amount) {
          entityPay.pay_amount = amount.formatted;
        });
 }
```

3.  Use `calcPayAnnualizedEst` variable to hold the numerical value for calculation and `entityPay.pay_annualized_est` to hold the formatted value.
```js
$scope.calcAnnualPayEst = function () {
        if (+entityPay.is_paid) {
          FormatCurrencyService.format(entityPay.pay_amount).then(function (amount) {
            calcPayAnnualizedEst = ((+amount.parsed) * (workPerYear[entityPay.pay_unit] || 0)).toFixed(2);
            entityPay.pay_annualized_est = amount.formatted;
          });
        }
};
```

4. Use `calcPayAnnualizedEst` variable to hold the numerical value for calculation.
```js
$scope.calcBenefitsPerCycle = function () {
    if (+entityPay.is_paid) {
       ...
       for (i; i < len; i++) {
         if (+entityPay.annual_benefits[i].type === 2) {
           entityPay.annual_benefits[i].amount_abs = (entityPay.annual_benefits[i].amount_pct / 100 * calcPayAnnualizedEst).toFixed(2);
         }
         ...
      }
      ...
   }
};
```

5.  Use `calcBenefitsPerCycleNet` variable to hold the numerical value for calculation and `entityPay.benefits_per_cycle_net` to hold the formatted value.
```js
$scope.calcBenefitsPerCycleNet = function () {
        if (+entityPay.is_paid) {
          calcBenefitsPerCycleNet = ($scope.benefits_per_cycle - $scope.deductions_per_cycle);

          FormatCurrencyService.format(calcBenefitsPerCycleNet).then(function (amount) {
            entityPay.benefits_per_cycle_net = amount.formatted;
          });
        }
 };
```

6. Use `calcPayAnnualizedEst` variable to be used in calcualtion.
```js
$scope.calcDeductionsPerCycle = function () {
    if (+entityPay.is_paid) {
      ...
      for (i; i < len; i++) {
         if (+entityPay.annual_deductions[i].type === 2) {
            entityPay.annual_deductions[i].amount_abs = (entityPay.annual_deductions[i].amount_pct / 100 * calcPayAnnualizedEst).toFixed(2);
         }
         ...
      }
      ...
    }
};
```

7. Use `calcPayPerCycleGross` variable to hold the numerical value for calculation and `entityPay.pay_per_cycle_gross` to hold the formatted value.
```js
$scope.calcPayPerCycleGross = function () {
        if (+entityPay.is_paid) {
          calcPayPerCycleGross = (calcPayAnnualizedEst / getCycles()).toFixed(2);

          FormatCurrencyService.format(calcPayPerCycleGross).then(function (amount) {
            entityPay.pay_per_cycle_gross = amount.formatted;
          });
        }
};
```

8. Use `calcPayPerCycleNet` variable to hold the numerical value for calculation and `entityPay.pay_per_cycle_net` to hold the formatted value.
```js
$scope.calcPayPerCycleNet = function () {
        if (+entityPay.is_paid) {
          var calcPayPerCycleNet = (+calcPayPerCycleGross + +calcBenefitsPerCycleNet).toFixed(2);
          FormatCurrencyService.format(calcPayPerCycleNet).then(function (amount) {
            entityPay.pay_per_cycle_net = amount.formatted;
          });
        }
};
```

9. Remove the `hrjc-number` directive form for fields for pay in contract modal template `views/modalForm.html` as its no more needed.
```html
-  <input id="{{prefix}}pay-amount" ... hrjc-number ng-model="entity.pay.pay_amount" ...>
+  <input id="{{prefix}}pay-amount" ... ng-model="entity.pay.pay_amount" ...>
```

#### After reopened
Fix for FE issues:
1. Return amount with decimalseparator and default decimal value as "0<decimal separator in settings>00" if amount is does not contain decimal points while formatting in currencyFormatService.js in reqangular:
```js
/**
     * Add thousand separators to the given amount.
     *
     * @param {String} amount
     * @return {String}
     */
    function addThousandSeparator (amount, separators) {
      .
      return (decimalIndex !== -1)
        ? (amountBeforeDecimal.toString() + separators.decimal + amountAfterDecimal.substring(0, 2))
        : amountBeforeDecimal + separators.decimal + '00';
    }
```
And if the amount is not valid return the default value with decimal separator:
```js
if(!amount) {
   return '0'+ separators.decimal + '00';
}
```

2. Add "ng-blur" on pay amount field and call a function to format and make related calculations:
```html
<input ng-model="entity.pay.pay_amount" ng-blur="onPayAmountChange()"/>
```
```js
// handler
onPayAmountChange() {
   formatPayAmount();
   // call other functions to make calulations
}
```
Note: Previously, formatAmount was called for every keystroke in pay amount field.

3. The Gross Pay Per Cycle and Net Pay per Cycle was not working because the value `entity.pay.pay_annualized_est` does not get changed when we change value of pay unit. Hence, the calculation never occur. To fix this add `calcPayAnnualizedEst` in to scope which gets updated on changing the pay Unit value. And then calculate the Gross pay Per Cycle and Net Pay Per Cycle once `$scope.calcPayAnnualizedEst` is updated.
```js
...
$scope.calcPayAnnualizedEst = 0;
...
```
Update watcher:
```js
$scope.$watch('calcPayAnnualizedEst', function () {
    $scope.calcPayPerCycleGross();
    $scope.calcBenefitsPerCycle();
    $scope.calcDeductionsPerCycle();
});
```
4. Get the default value (formatted based on currency settings in civicrm) :
```js
$scope.setDefaults = function () {
    ...
    FormatCurrencyService.format('0')
       .then(function (amount) {
           entityPay.pay_amount = amount.formatted;
        });
};
```

### For QA:
1. Check that, in Job contract  modal in pay tab when an amount is entered, it gets formatted automatically using the settings decimal and thousand separator.
2. After the amount is entered, the job contract  gets saved successfully.
3. Verify that  the while editing the job contract, previously added values appear as formatted.
4. Check that the contact can be saved after editing the extisting data.

## Tests:
✅ Jasmine Tests - passed
⏹  CSS distributive files - not needed to run
⏹ Backstop Tests - n/a


igorpavlov:
@Buddha please add GIFs instead of static screenshots.

igorpavlov:
@bmnepali Since this PR is about decimal and thousands delimiters I have a couple of questions:
1) which of delimiters are supported from the list on this page? https://en.wikipedia.org/wiki/Decimal_mark
2) As far as I understand we need to (1) parse any format to a float to save it to DB (2) format it back in a default (or preset) format. Parsing may be simply achieved with regular expressions:

```js
// This is for parse
replace(/(^([\d]{,3})|([\d]{3}))[list_of_thousand_delimiters]/g, "$1"); // remove all thousand delimiters
![image](https://user-images.githubusercontent.com/3973243/31952904-a177c424-b8d9-11e7-8548-7107e248f910.png)
replace(/[list_of_decimal_delimiters]([\d]{2})$/, ".$1") // change any decimal delimiter to a point
![image](https://user-images.githubusercontent.com/3973243/31952934-b840febe-b8d9-11e7-94d3-0a045e612dcc.png)
```
Would that work?

Formatting the number back to currency should be even more straightforward.

AkA84:
@bmnepali you have this commit "PCHR-1849: Include FormatCurrencyService + semistandard issues fix"

Do not put ever this things together, either you fix the style, or you implement the changes.

AkA84:
> Use calcPayAnnualizedEst variable to hold the numerical value for calculation and entityPay.benefits_per_cycle to hold the formatted value.

I think this is a copy past error, there is no trace of `calcPayAnnualizedEst` in the snippet below

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+    describe("when ' and , are thousand and decimal separators respectively", function () {
+      beforeEach(function () {
+        $httpBackend.whenGET(/entity=Setting/).respond(setingsMock.settings);
+      });
+      describe('addSeperators()', function () {
+        describe('when the amount is without decimal', function () {
+          it('formats the given amount using defined decimal and thousands separator', function () {
+            FormatCurrencyService.addSeperators(6000000000).then(function (result) {
+              expect(result).toEqual("6'000'000'000");
```
igorpavlov: This is a test for Switzerland?

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -0,0 +1,116 @@
+/* eslint-env amd, jasmine */
```
igorpavlov: Should we add more edge cases like 1.01, "1.101", 0, -1, -0, "xxx", "100.000.00" etc.

hrjobcontract/js/src/job-contract/app.js
```diff
@@ -1,8 +1,12 @@
-define([
```
igorpavlov: Please do not mix auto semistandard changes with manual changes. Not for now, but for future.

hrjobcontract/js/src/job-contract/controllers/form/form-pay.js
```diff
@@ -38,6 +41,18 @@ define([
       $scope.benefits_per_cycle_net = 0;
       $scope.deductions_per_cycle = (0).toFixed(2);
 
+      // Format the pay amount to be displayed in the pay amount field
```
igorpavlov: This is a function, shouldn't then comment be decorated as per functions?

hrjobcontract/js/src/job-contract/controllers/form/form-pay.js
```diff
@@ -79,7 +94,14 @@ define([
 
       $scope.calcAnnualPayEst = function () {
         if (+entityPay.is_paid) {
-          entityPay.pay_annualized_est = (entityPay.pay_amount * workPerYear[entityPay.pay_unit] || 0).toFixed(2);
+          FormatCurrencyService.removeCharacters(entityPay.pay_amount).then(function (calculabeAmount) {
+            calcPayAnnualizedEst = ((+calculabeAmount) * (workPerYear[entityPay.pay_unit] || 0)).toFixed(2);
+            return calcPayAnnualizedEst;
```
igorpavlov: Line break is needed

hrjobcontract/js/src/job-contract/controllers/form/form-pay.js
```diff
@@ -1,175 +1,216 @@
...
+      entityPay.annual_deductions = entityPay.annual_deductions || [];
+
+      $scope.collapseBenDed = !entityPay.annual_benefits.length && !entityPay.annual_deductions.length;
+      $scope.benefits_per_cycle = (0).toFixed(2);
+      $scope.benefits_per_cycle_net = 0;
+      $scope.deductions_per_cycle = (0).toFixed(2);
+
+      // Format the pay amount to be displayed in the pay amount field
+      // as per the settings
+      $scope.formatPayAmount = function () {
```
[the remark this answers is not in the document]
bmnepali: I have used in the view to test, later limited its use in controller only. I will make it private.

hrjobcontract/views/modalForm.html
```diff
@@ -429,7 +429,7 @@ <h4 class="modal-title">{{copy.title}}</h4>
                             ng-if="benefit.type == '2'"
                             ng-model="benefit.amount_pct"
                             ng-disabled="isDisabled"></td>
-                          <td><input type="text" class="form-control" hrjc-number ng-model="benefit.amount_abs" ng-disabled="isDisabled || benefit.type == '2'"></td>
```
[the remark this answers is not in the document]
bmnepali: The `hrjc-number` directive was used for the two things:
1. To determine the number of values to keep after decimal separator
3. Round figure the given value to nearest value

I have removes it, as the amounts in the field are now `string` with different separators. And the directive does not work with strings.

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,86 @@
...
+
+define([
+  'common/modules/services',
+  'common/services/api'
+], function (services) {
+  'use strict';
+
+  services.factory('FormatCurrencyService', FormatCurrencyService);
+
+  FormatCurrencyService.$inject = ['api', '$q', '$log'];
```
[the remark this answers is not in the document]
bmnepali: got it. Will fix.

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,86 @@
...
+
+  function FormatCurrencyService (api, $q, $log) {
+    $log.debug('Service: FormatCurrencyService');
+
+    var promise;
+
+    return {
+      addSeperators: addSeperators,
```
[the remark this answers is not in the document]
bmnepali: ops! will fix... thanks for catching!

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Removes characters other than decimal separator
+     * Converts the formatted amount to numeric value (making calulable) by
+     * removing the thousand separators and replacing the decimal separator by " ."
+     *
+     * @param  {string|Integer} amount
+     * @return {integer}
+     */
+    function removeCharacters (amount) {
```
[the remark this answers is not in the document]
bmnepali: Right! sounds much better, will fix.

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -0,0 +1,116 @@
...
+            FormatCurrencyService.removeCharacters("799'600'000,00").then(function (result) {
+              expect(result).toEqual('799600000.00');
+              expect(+result).toEqual(jasmine.any(Number));
+            });
+          });
+        });
+      });
+    });
+
+    describe('when # and @ are thousand and decimal separators respectively', function () {
```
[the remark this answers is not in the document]
bmnepali: Makes sense, will remove it.

hrjobcontract/js/src/job-contract/controllers/form/form-pay.js
```diff
@@ -1,175 +1,216 @@
...
+        FormatCurrencyService.removeCharacters(entityPay.pay_amount).then(function (calculableAmount) {
+          return calculableAmount;
+        }).then(function (calculableAmount) {
+          FormatCurrencyService.addSeperators(calculableAmount).then(function (formattedAmount) {
```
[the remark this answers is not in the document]
bmnepali: Here:
1. `removeCharacters()` function will take the formatted amount and convert it to calculable numerical value.
2. `addSeperators()` function take the numerical value and adds the separators.

I think they should remain separate. But let me check if they can be put together.

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Gets defined settings data form backend
+     *
+     * @param  {Object} params
+     *
+     * @return {Promise}
+     */
+    function getSettings () {
+      return api.sendGET('Setting', 'get').then(function (data) {
```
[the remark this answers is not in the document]
bmnepali: Will discuss with other dev and will implement it.

org.civicrm.reqangular/test/services/format-currency.spec.js
bmnepali: Re-written test cases.

org.civicrm.reqangular/test/services/format-currency.spec.js
bmnepali: NO! it was just to test if it formats the amount correctly.

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    afterEach(function () {
+      $httpBackend.flush();
+      $rootScope.$apply();
+    });
+
+    describe("when ' and , are thousand and decimal separators respectively", function () {
+      beforeEach(function () {
+        $httpBackend.whenGET(/entity=Setting/).respond(setingsMock.settings);
+      });
+      describe('addSeperators()', function () {
```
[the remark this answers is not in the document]
bmnepali: Fixed!

org.civicrm.reqangular/src/common/models/instances/settings-instance.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env amd */
+
+define([
+  'common/modules/models-instances',
+  'common/models/instances/instance'
+], function (instances) {
+  'use strict';
+
+  instances.factory('SettingsInstance', ['ModelInstance',
+    function (ModelInstance) {
```
igorpavlov: This indentation is a bit difficult to read imho, could you please try that instead:

```js
instances.factory('SettingsInstance', [
  'ModelInstance',
  function (ModelInstance) {
    return ModelInstance.extend({});
  }
]);
```

org.civicrm.reqangular/src/common/models/settings.model.js
```diff
@@ -0,0 +1,54 @@
...
+
+  models.factory('Settings', [
+    '$log', 'Model', 'api.settings', 'SettingsInstance',
+    function ($log, Model, SettingsAPI, instance) {
+      $log.debug('Settings');
+
+      var promise;
+
+      return Model.extend({
+
```
igorpavlov: No line break needed

org.civicrm.reqangular/src/common/models/settings.model.js
```diff
@@ -0,0 +1,54 @@
...
+        /**
+         * Calls the get() method of the Settings API, and returns a
```
igorpavlov:
```js
/**
 * Returns settings
 *
 * @param {Object} params
 * @return {Promise} resolved with an array of Settings Instances for each setting
 */
```

org.civicrm.reqangular/src/common/models/settings.model.js
```diff
@@ -0,0 +1,54 @@
...
+            });
+        },
+
+        /**
+         * Fetched the thousand and decimal seperators form backend
+         *
+         * @return {promise}
+         */
+        fetchSeparators: function () {
+          promise = promise || this.get();
```
igorpavlov: Why is this needed?

org.civicrm.reqangular/src/common/services/api/settings.api.js
```diff
@@ -0,0 +1,33 @@
...
+  'common/modules/apis',
+  'common/services/api'
+], function (_, apis) {
+  'use strict';
+
+  apis.factory('api.settings', ['$log', 'api', function ($log, api) {
+    $log.debug('api.settings');
+
+    return api.extend({
+
```
igorpavlov: No line break needed

org.civicrm.reqangular/src/common/services/api/settings.api.js
```diff
@@ -0,0 +1,33 @@
...
+      /**
+       * Returns all settings.
+       *
+       * @param  {Object} params  matches the api endpoint params (ex. contact_id)
+       * @return {Promise}
```
igorpavlov: Please add what the promise resolves with

org.civicrm.reqangular/src/common/services/api/settings.api.js
```diff
@@ -0,0 +1,33 @@
...
+       */
+      get: function (params) {
+        $log.debug('api.settings.get');
+
+        params = params || {};
+
+        return api.sendGET('Setting', 'get', params, false)
+        .then(function (data) {
```
igorpavlov: Please add 2 spaces indentation here (move .then block 2 spaces to the right).

org.civicrm.reqangular/src/common/services/api/settings.api.js
```diff
@@ -0,0 +1,33 @@
...
+
+  apis.factory('api.settings', ['$log', 'api', function ($log, api) {
+    $log.debug('api.settings');
+
+    return api.extend({
+
+      /**
+       * Returns all settings.
+       *
+       * @param  {Object} params  matches the api endpoint params (ex. contact_id)
```
igorpavlov: You do not need two spaces after `params`

org.civicrm.reqangular/test/mocks/services/api/settings-mock.js
```diff
@@ -0,0 +1,37 @@
...
+define([
+  'common/lodash',
+  'common/mocks/module',
+  'common/mocks/data/settings.data'
+], function (_, mocks, SettingsData) {
+  'use strict';
+
+  mocks.factory('api.settings.mock', ['$q', function ($q) {
+    return {
+      get: function (params) {
```
igorpavlov: please add a func description

org.civicrm.reqangular/test/mocks/services/api/settings-mock.js
```diff
@@ -0,0 +1,37 @@
...
+        return $q(function (resolve, reject) {
+          resolve(SettingsData.get.values);
+        });
+      },
+
```
igorpavlov: i am really not sure these line breaks needed. @AkA84 ?

org.civicrm.reqangular/test/models/instances/settings-instance.spec.js
```diff
@@ -0,0 +1,25 @@
...
+  'common/models/instances/settings-instance'
+], function (_) {
+  'use strict';
+
+  describe('SettingsInstance', function () {
+    var SettingsInstance, ModelInstance;
+
+    beforeEach(module('common.models.instances'));
+
+    beforeEach(inject(['SettingsInstance', 'ModelInstance',
```
igorpavlov: This indentation might be easier to read. Up to you.

```js
beforeEach(inject(['SettingsInstance', 'ModelInstance',
  function (_SettingsInstance_, _ModelInstance_) {
    SettingsInstance = _SettingsInstance_;
    ModelInstance = _ModelInstance_;
  }
]));
```

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -2,85 +2,123 @@
...
-        var withThousandSeparator = amount.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1' + separators.thousand);
-        var formattedCurrency = withThousandSeparator.replace('.', separators.decimal);
-
-        return (+amount) ? formattedCurrency : ('0' + separators.decimal + '00');
+    * Adds defined thousand and decimal separators in the given amount
+    *
+    * @param {string} amount
+    * @return {promise}
+    */
+    function format (amount) {
```
igorpavlov: Why `format` and not `addSeparators`?

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -2,85 +2,123 @@
...
+      return Settings.fetchSeparators().then(function (separators) {
+        return {
+          formatted: addSeparators(amount, separators),
+          unformatted: removeSeparators(amount, separators)
```
igorpavlov: I would use the word `original` instead of unformatted since unformatted might mean "formatted and then reverted back"

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -2,85 +2,123 @@
...
+     * @return {string}
      */
-    function getSettings () {
-      return api.sendGET('Setting', 'get').then(function (data) {
-        return data.values[0];
-      });
+    function configureExpression (decimalSeparator) {
+      var expression;
+
+      // Handle reserved regex characters that might be used as separators
```
igorpavlov: I think it could be a bit more condenced

```js
var escape = ['\\', ']'];

_.includes(escape, decimalSeparator) && (decimalSeparator = '\\' + decimalSeparator);

return '[^0-9' + decimalSeparator + ']';
```

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -2,85 +2,123 @@
...
+      }
+
+      return expression;
     }
 
     /**
-     * Removes characters other than decimal separator
-     * Converts the formatted amount to numeric value (making calulable) by
-     * removing the thousand separators and replacing the decimal separator by " ."
+     * Formats the given amount with thousand separators
```
igorpavlov: the name of the func is `removeCharacters`, so I guess it does not format. It removes characters! :) Please amend the description to make it a bit more specific.

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -2,85 +2,123 @@
...
+      var amountBeforeDecimal;
+      var decimalIndex;
+      var expression;
 
-        return amountWithDecimal;
-      });
+      // Handling special case:
+      // If we are given to format the amount with decimal as "." ,
+      // we need to change "." into defined decimal separator for proper formatting
+      if (+amount && (amount.toString()).indexOf('.') !== -1) {
```
igorpavlov: I wonder what if I pass '0.0'

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -1,113 +1,81 @@
...
+          $rootScope.$apply();
         });
 
-        describe('when the amount is with decimal', function () {
-          it('removes the non numerical characters except decimal character "." making it calulable', function () {
-            FormatCurrencyService.removeCharacters('799#600#000@00').then(function (result) {
-              expect(result).toEqual('799600000.00');
-              expect(+result).toEqual(jasmine.any(Number));
-            });
+        it('formats' + test[2] + ' to ' + test[3] + ' by ' + test[1] + ' and ' + test[0] + ' as decimal and thousand separators', function () {
```
igorpavlov: please add a space after `formats`

org.civicrm.reqangular/src/common/models/settings.model.js
```diff
@@ -0,0 +1,54 @@
...
+         * SettingsInstance for each setting.
+         *
+         * @param  {Object} params
+         * @return {Promise}
+         */
+        get: function (params) {
+          return SettingsAPI.get(params)
+            .then(function (settings) {
+              return settings.map(function (setting) {
+                return instance.init(setting, true);
```
AkA84: we are creating a `SettingInstance` for each setting? why? i thought that an instance would represent the whole group of settings, what's the benefit of having an instance for each individual setting instead?

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,124 @@
...
+
+    return {
+      format: format
+    };
+
+    /**
+    * Adds defined thousand and decimal separators in the given amount
+    *
+    * @param {string} amount
+    * @return {promise}
```
AkA84: `{Promise}`

org.civicrm.reqangular/src/common/services/format-currency.service.js
AkA84: cause it's not adding separators *every time*, it's simply ensuring that the given amount is correctly formatted, and if it isn't, it only then adds the separators

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,124 @@
...
+    */
+    function format (amount) {
```
AkA84: problem is that now that the method returns an object that provides both the `formatted` and `unformatted` (is that even correct to call it that way?) amount, then the method has a kind of incorrect name

org.civicrm.reqangular/src/common/services/format-currency.service.js
AkA84: yeah, something like that

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,124 @@
...
+      return amountWithDecimal;
+    }
+
+    /**
+     * Configure expression for special characters like \ and ]
+     *
+     * @param  {string} decimalSeparator
+     * @return {string}
+     */
+    function configureExpression (decimalSeparator) {
```
AkA84: this function has a very vague name. What expression? for what? the name doesn't say

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Formats the given amount with thousand separators
+     *
+     * @param  {string|integer} amount
+     * @param  {object} separators
+     * @return {string}
+     */
+    function addSeparators (amount, separators) {
```
AkA84: i'm not reviewing the whole function, but i can tell at a glance that it definitely does too much. Please split it into smaller functions

org.civicrm.reqangular/src/common/models/instances/settings-instance.js
bmnepali: looks better! will implement.

org.civicrm.reqangular/src/common/models/settings.model.js
bmnepali: > `promise = promise || this.get();` 
This is used to cache the promise fetching separators  else get the separators for backend.

org.civicrm.reqangular/src/common/services/api/settings.api.js
bmnepali: sure! added!

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: @igorpavlov Actually, the resulted value for `unformatted` is a value with decimal separator as "." With sole purpose of making the formatted value to be used in calculations. So, i think calling it `original` is not suitable either. I prefer to call `calculable` or `asNumber`. What do you say?

cc @AkA84

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: Sorry! `* Formats the given amount with thousand separators` comment is for `addSeparators()` function not for `removeCharacters()` function.

org.civicrm.reqangular/test/mocks/services/api/settings-mock.js
bmnepali: sorry! i will.

org.civicrm.reqangular/test/services/format-currency.spec.js
bmnepali: Added! good catch.

org.civicrm.reqangular/src/common/models/settings.model.js
bmnepali: Actually, `SettingsAPI.get(...)` returns the array of setting object results as:
```js
[
   {
      propertyOne: '',
      propertyTwo: '',
      ....
   }
]
```
So, it will represent the whole group of settings not the individual settings property.

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: Well, here is what actually happens:
1.  `formatted`:  Will hold the value with added thousand and decimal separators.
2.  `unformatted`: Will hold the value that is ready for calculation.

So, whatever the given amount, for both cases above we are formatting. So, i think its appropriate to call `format`. Does it makes sense?

cc @igorpavlov

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: Agreed! 
We should better call it `prepareRegexExpression` or `expressionWithEscapedCharacters` what do you suggest?

cc @igorpavlov

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: ok! will break it down. Thanks!

org.civicrm.reqangular/src/common/models/settings.model.js
igorpavlov: @bmnepali I see. Could you please doublecheck that it doesn't call API second time you call it?

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: Fair enough @AkA84 !

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: @bmnepali  `parsed` ?

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: @bmnepali please make sure the descriptions are correct then

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: @bmnepali any feedback on that?

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: well, to be honest all need to be capitalised since they are all classes. even the {String}.

image

cc @bmnepali @AkA84

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: As I suggested above, `unformatted` might become `parsed`. 

1. `parsed`: Will hold the value that is ready for calculation.
2. `formatted`: Will hold the value with added thousand and decimal separators.

cc @AkA84

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: `prepareRegexForSeparator` ? cc @bmnepali @AkA84

org.civicrm.reqangular/src/common/services/format-currency.service.js
igorpavlov: @bmnepali If you are gonna split the func please move inline comments to func descriptions

cc @AkA84

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: yeah! I have already added a test case for this case. it returns '0,0' (where "," is a decimal separator. And final result  after adding  separator will be '0'.

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: ok sure!

org.civicrm.reqangular/src/common/models/settings.model.js
bmnepali: yeah! let me check if i can add any test for it.

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: `prepareRegexForSeparator ` looks better.

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: `parsed` as well looks fine.

org.civicrm.reqangular/src/common/models/settings.model.js
```diff
@@ -16,15 +16,14 @@ define([
...
-         * SettingsInstance for each setting.
+         * Returns settings
          *
-         * @param  {Object} params
-         * @return {Promise}
+         * @param {Object} params
+         * @return {Promise} resolved with an array of Settings Instances
          */
         get: function (params) {
+          $log.debug('First call!!!');
```
igorpavlov: Is that your debugging log left? If yes, please remove it. If it is intended, please amend the caption inside the log to make it look more professional.

org.civicrm.reqangular/src/common/services/api/settings.api.js
```diff
@@ -11,22 +11,21 @@ define([
...
     return api.extend({
-
       /**
-       * Returns all settings.
+       * Returns all settings
        *
-       * @param  {Object} params  matches the api endpoint params (ex. contact_id)
-       * @return {Promise}
+       * @param  {Object} params matches the api endpoint params (ex. contact_id)
+       * @return {Promise} resolved settings data from backend as array of objects
```
igorpavlov: resolve**s**

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -21,84 +22,96 @@ define([
...
+     * Converts the formatted amount to numeric value (making calulable)
+     * by removing the thousand separators and replacing the decimal separator by " ."
+     * and keeping 2 values after decimal separator.
      *
-     * @param  {string|Integer} amount
-     * @param  {object} separators
-     * @return {integer}
+     * @param  {String|Number} amount
+     * @param  {Object} separators
+     * @return {String}
```
igorpavlov: Nope, the func actually returns Number.

org.civicrm.reqangular/src/common/services/format-currency.service.js
```diff
@@ -21,84 +22,96 @@ define([
...
-      // except defined decimal separator in the string
-      expression = configureExpression(separators.decimal);
+      return amount;
+    }
+
+    /**
+     * Add thousand separators to the given amount.
+     *
+     * @param {String} amount
+     * @return {String} amount separated with correct separators
```
igorpavlov: In that particular case `@return {String}` would be enough since the description is enough.

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -30,21 +30,22 @@ define([
...
       // Common cases
+      [',', '*', 0.0, '0'],
       [',', '*', 12345675, '12,345,675'],
       [',', '*', 123456.75, '123,456*75'],
       [',', '*', '123456.7.5', '12,345,675'],
       [',', '*', '12345675', '12,345,675'],
       [',', '*', '12,345,675', '12,345,675'],
       [',', '*', '00,456,75*76', '45,675*76'],
       [',', '*', '12345675*76', '12,345,675*76'],
-      [',', '*', '12,34,56*75*76', '123,456*7576'],
```
igorpavlov: Why has this been amended? Was the expectation wrong or was the test failing?

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -30,21 +30,22 @@ define([
...
+      [',', '*', '12,34,56*75*76', '123,456*75'],
       [',', '*', '123 456 75*76', '12,345,675*76'],
 
       // Edge cases
       ['?', '^', 'abcd!89', '89'],
       ['?', '!', '00000!89', '0!89'],
       ["'", ',', '1234567,76', "1'234'567,76"],
-      ['*', '^', '1234567^76', '1*234*567^76'],
```
igorpavlov: Why has this been changed?

org.civicrm.reqangular/test/services/format-currency.spec.js
bmnepali: The code was now amended to format the amount with 2 numbers after decimal separator. So, it's been changed.

org.civicrm.reqangular/test/services/format-currency.spec.js
bmnepali: Seems I changed the value during fixing tests, but no worries its still valid.

org.civicrm.reqangular/src/common/models/settings.model.js
bmnepali: removed!

org.civicrm.reqangular/src/common/services/format-currency.service.js
bmnepali: Fixed!

org.civicrm.reqangular/test/services/format-currency.spec.js
igorpavlov: @bmnepali it might be then needed to 

- doublecheck that we have either whole numbers OR fixed(2) numbers
- understand what is the operation for exceeding decimals: floor, round or ceil
- add more edge cases for 3 digits: 1234.123    1234.128      1234.1

bmnepali: Ok! its confirmed with @davialexandre:
> If you try to insert/update a value with more than 2 decimal digits, MySQL will simply ignore the additional ones and will only save the first two

peek 2017-11-01 13-11

So, if the user inputs the value like 123,345.845 it will be converted into 123,345.84, simply ignoring the additional numbers by only keeping the first two numbers after decimal.

igorpavlov:
The last commit name has a spelling mistake `PCHR-1849: Implemants PR review feedbacks` - it should be "implem**e**nts"

org.civicrm.reqangular/test/services/format-currency.spec.js
```diff
@@ -30,27 +30,32 @@ define([
     var testMap = [
       // [thousandSeparator, decimalSeparator, inputAmount, outputAmount]
       // Common cases
+      [',', '*', 0.0, '0'],
+      [',', '*', 'xxxxx', '0'],
+      [',', '*', '0.8889', '0*88'],
+      [',', '*', '-0.8889', '0*88'],
```
igorpavlov: Is this confirmed that negative currency goes positive?

bmnepali: yes! confirmed with @mike

bmnepali:
Fixed mere conflicts by rebase.

bmnepali:
Fixed merge conflicts by rebase.

davialexandre:
Closing this as it has gone stale

PR #2240, opened 2017-10-24
PCHR-2625: Add sub menu to Leave menu item

AkA84:
## Overview
This PR adds an upgrader to the L&A extension in order to remove the direct link of the "Leave" menu item in the admin, and to add instead a sub menu to it

## Before
The "Leave" menu item is a simple directly link to /civicrm/leaveandabsences/dashboard

## After
The "Leave" menu item is now a toggle for a sub menu, containing items pointing to the individual sections of the L&A admin

after


mickadoo:
I'm not sure what the approach is, I've seen both done, but if we're updating the menu items here I wonder if they should be updated in the hrleaveabsenses XML that contains the menu items. It might be good to have a consistent approach on this.

PR #2233, opened 2017-10-23
PCHR-2654: Leave Widget

reneolivo:
## Overview
This PR introduces the Leave Widget to the Contact Summary page.

This PR is composed of the following PRs:

* #2192 - Leave Widget HTML & CSS
* #2202 - Import components between extensions (only into Contact Summary for now)
* #2206 - Leave Widget Available Balance
* #2226 - Leave Widget Absence Types Amount Taken (Sickness leave taken) and Heatmap
* #2229 - Leave Widget Next Leave
* #2241 - Leave Widget help texts, support for small screens, and fixed an issue (see PR for more information)

## How it looks
anim

## Technical Details
* This PR will render debug mode obsolete. All JS changes need to be compiled from now on.
* A test that was not needed and marked as `xit` was removed in this commit: https://github.com/civicrm/civihr/pull/2233/commits/b06bfe00d9cf9b1f31bcbfa0357bcec210e58bc7

## Smock testing
This PR makes considerable changes to all extension because of the removal of Debug Mode support. Smock tests were conducted for Contact Summary and Leave & Absences (create leave requests, editing, consulting them, approving them). Everything seems in order.

For the remaining extensions, clicking on each section didn't display any errors on the console, but QA will help discover any issue before merging to staging.

PR #2241, opened 2017-10-24
PCHR-2654: Leave Widget responsive + help texts

reneolivo:
## Overview
This PR adds the following fixes to the Leave Widget:
* Makes the Leave Widget, and the Contact Summary as a whole, to be small screen friendly.
* Displays help text messages.
* Fixes the Sick leaves taken.

This PR is targeted to the feature branch `PCHR-2654-leave-widget` which is currently under review in PR #2233 

### Before
anim
The widget hides at different width sizes.

### After
anim

### Technical Details
Just replacing the contact summary `.container` html class for `.container-fluid` was enough.

## Displaying help texts
A new `help-text` directive was created. The directive provides a small information icon, and when clicked, a help message is displayed.

The directive works like this:

```html
<help-text title="Please help me">
  <p>Help text message</p>
  <ul>
    <li>Any</li>
    <li>Other</li>
    <li>Content</li>
  </ul>
</help-text>
```

The title attribute is optional. It will display `Help` for the title if this attribute is not provided.

### Displaying leave widget help text
anim
The texts were taken from the [Leave Widget Wiki](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/124846367/Leave+widget)

#### Help Text Directive
A reqangular directive was created. The help text content is transcluded into the directive and passed to the controller by doing the following:

```js
function helpTextLink ($scope, $element, $attrs) {
    $scope.getHelpText = getHelpText;

    function getHelpText () {
      return $element.find('.help-text').html();
    }
}
```

A function is used instead of passing the help text right away in case the HTML content changes (because of bindings, etc.).

The controller displays the message by using the notification service as follows:

```js
function displayHelpText () {
    var helpText = $scope.getHelpText();

    notificationService.info($scope.title || defaultTitle, helpText);
}
```

#### Leave Widget help texts
Both `leave-widget-absence-types-amount-taken.component` and `leave-widget-absence-types-available-balance.component` needed to display help text messages. 

The implementation on `leave-widget-absence-types-available-balance.component` is straight forward.

Since `leave-widget-absence-types-amount-taken.component` is a generic component that would display the amount taken for any absence types list (in this case Sickness, but it could be anything), then the word "Sickness" was replaced for `{{absenceTypesTaken.title | lowercase}}` and "sick leaves" was replaced for `{{absenceTypesTaken.leaveName | lowercase}}` (which defaults to "leaves"). These two values are passed through the leave widget:

```html
<leave-widget-absence-types-amount-taken
  title="Sickness" <--- here
  leave-name="sick leaves" <--- here
  absence-period="leaveWidget.currentAbsencePeriod"
  absence-types="leaveWidget.sicknessAbsenceTypes"
  contact-id="leaveWidget.contactId"
  leave-request-statuses="leaveWidget.leaveRequestStatuses"
></leave-widget-absence-types-amount-taken>
```

## Fixing Absence Types Amount Taken

By mistake, a `vm.absenceTypes = []` was added to the `leave-widget-absence-types-amount-taken.component`, this would overwrite the binding value for absence types, which would then execute a wrong query statement ( `type_id: { IN: [] }` ). 

The variable definition, and the test for it were removed.


igorpavlov:
@Rene the right padding on the GIF doesn't look equal to the left one, if this intentional?

org.civicrm.reqangular/src/common/components/heatmap.component.js
```diff
@@ -5,7 +5,7 @@ define([
   'common/moment',
   'common/modules/components'
 ], function (_, moment, components) {
-  return components.component('heatmap', {
```
igorpavlov: Could you please explain why this was removed?

org.civicrm.reqangular/src/common/templates/help-text.html
```diff
@@ -0,0 +1,5 @@
+<button class="info-button" ng-click="helpText.displayHelpText()">
+  <i class="fa fa-info-circle"></i>
+
```
igorpavlov: Not sure this empty line is needed

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget.component.js
```diff
@@ -118,6 +119,7 @@ define([
         vm.sicknessAbsenceTypes = types.filter(function (type) {
           return +type.is_sick;
         });
+        console.log('nein, nein, nein', types, vm.sicknessAbsenceTypes);
```
igorpavlov: yah, yah, yah!

org.civicrm.reqangular/src/common/components/heatmap.component.js
reneolivo: just the return, not the component definition. It was removed because returning the component is not standard. It also blocks you from doing things like this:

```js
return components.component({
   // ...
});

controller.$inject = ['$scope']; // <--- won't work because of the return.

function controller ($scope) {
  // ...
}
```

reneolivo:
## Large Desktop Support
Large desktop support has also been added by setting up a max width for the container.

### Before
download 93

### After
download 92

anim

### Technical description

The `container-fluid` now has `max-width: $container-large-desktop`. This variable comes from Bootstrap and it's equal to 1156px.

PR #2243, opened 2017-10-25
PCHR-2627: Use developer permission on developer menu and paths

AkA84:
This PR adds a new permission checks on the Developer's menu items and paths. Those can now be accessed only by users with the "access CiviCRM" and "access CiviCRM developer menu and tools" permissions

### DIFF
--- a/hrui/hrui.php
+++ b/hrui/hrui.php
@@ -590,6 +590,17 @@ function hrui_civicrm_managed(&$entities) {
 }
 
 /**
+ * Implements hook_civicrm_alterMenu().
+ *
+ * @param Array $items List of http routes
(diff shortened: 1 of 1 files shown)


hrui/hrui.php
```diff
@@ -590,6 +590,17 @@ function hrui_civicrm_managed(&$entities) {
 }
 
 /**
+ * Implements hook_civicrm_alterMenu().
+ *
+ * @param Array $items List of http routes
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_navigationMenu
```
davialexandre: I think the correct link here is this one: https://docs.civicrm.org/dev/en/latest/hooks/hook_civicrm_alterMenu/

PR #2244, opened 2017-10-25
PCHR-2716: Create custom Timepicker directive for Leave in Hours

igorpavlov:
## Overview

A custom Timepicker selector is created that supports basic time selection and can be configured on the fly with the following options:
- Time from (time, ex. '10:30')
- Time to (time, ex. '18:20')
- Interval (in minutes, ex. '15')
- Default option (ex. 'Please select time')

## After

Note: this GIF was made in a temporary template for demonstration purposes only

You may notice from the GIF above how the interval changes on the fly, this is also made for demonstration purposes.

## Technical Details

The directive can be used in the following way:
```js
<select
  timepicker-select // necessary to execute a directive
  timepicker-select-time-from="'9:30'" // optional, defaults to '00:00'
  timepicker-select-time-to="'14:10'" // optional, defaults to '23:59'
  timepicker-select-interval="report.interval" // optional, defaults to '1'
  timepicker-select-default-option="'Select time'" // optional, if omitted, there will be no default option
></select>
```

## Comments

This PR also includes a fix for the unit calculation filter which failed to provide an expected value for a particular edge case.

----

✅ Jasmine Tests - passed
✅ Manual Tests - passed
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - n/a
⏹ CSS distributive files - n/a


reneolivo:
Pick and select are synonymous, remove one from the directive name please

igorpavlov:
@reneolivo 
1) the case was 20.75 which rendered 20h instead of 21h (20 is an example)
2) select means `<select>` in this case, but I think you are right

org.civicrm.reqangular/src/common/directives/timepicker-select.directive.js
```diff
@@ -0,0 +1,79 @@
...
+  function timepickerSelectController ($scope) {
+    var vm = this;
+
+    vm.defaultOption = $scope.timepickerSelectDefaultOption;
+    vm.options = [];
+
+    /**
+     * Builds options for the selector
+     */
+    function buildOptions () {
```
reneolivo: using moment:

```js
function buildOptions () {
  var interval = +$scope.timepickerSelectInterval || 1;
  var timeFrom = moment.duration($scope.timepickerSelectTimeFrom || '00:00');
  var timeTo = moment.duration($scope.timepickerSelectTimeTo || '23:59');

  vm.options = [];

  // @Todo: validate timeFrom < timeTo

  while (timeFrom.asMinutes() <= timeTo.asMinutes()) {
    var time = moment.utc(timeFrom.asMilliseconds());
    vm.options.push(time.format('HH:mm'));
    timeFrom.add(interval, 'minutes');
  }
}
```

reneolivo: Pros: no need for `calculateLimit` and `addTrailingZero`.

org.civicrm.reqangular/src/common/templates/timepicker-select.html
```diff
@@ -0,0 +1,2 @@
+<option ng-if="selector.defaultOption" value="">{{ selector.defaultOption }}</option>
```
reneolivo: Can we call it placeholder? Default Option makes me think that is the time that will be automatically selected.

org.civicrm.reqangular/test/directives/timepicker-select.directive.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+
+    function buildDirective (options) {
+      var element, scopeKey;
+      var attributes = '';
+      var $scope = $rootScope.$new();
+
+      options = options || {};
+
+      _.each(options, function (optionValue, optionKey) {
```
reneolivo: can we replace this bit and do something like this instead:

```js
element = angular.element( '<select timepicker-select></select>');

_.each(options, function (optionValue, optionKey) {
  element.attr(optionKey, '"' + optionValue + '"');
});
```
Notice the `'"'` which would make the attributes a string instead of a scope binding. (`attr="'value'"` vs `attr="scope.property"`)

org.civicrm.reqangular/src/common/directives/timepicker-select.directive.js
```diff
@@ -0,0 +1,79 @@
...
+
+define([
+  'common/modules/directives'
+], function (directives) {
+  'use strict';
+
+  directives.directive('timepickerSelect', ['$templateCache', function ($templateCache) {
+    return {
+      scope: {
+        timepickerSelectDefaultOption: '<',
```
reneolivo: rename to placeholder and make it a `@` binding instead:

`<select timepicker-select timepicker-select-placeholder="Please select a time..."></select>`

org.civicrm.reqangular/src/common/directives/timepicker-select.directive.js
```diff
@@ -0,0 +1,79 @@
...
+    function calculateLimit (time) {
+      time = time.split(':');
+
+      return (+time[0]) * 60 + (+time[1]);
+    }
+
+    /**
+     * Adds a trailing zero if a number is less than 10
+     */
+    function addTrailingZero (number) {
```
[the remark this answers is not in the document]
igorpavlov: Removed that func.

org.civicrm.reqangular/test/directives/timepicker-select.directive.spec.js
igorpavlov: That would work only with "@" bindings. for "<" you need to at first assign props in scope then pass these props to a directive.

org.civicrm.reqangular/src/common/directives/timepicker-select.directive.js
```diff
@@ -25,49 +26,27 @@ define([
   function timepickerSelectController ($scope) {
     var vm = this;
 
-    vm.defaultOption = $scope.timepickerSelectDefaultOption;
+    vm.placeholder = $scope.timepickerSelectPlaceholder;
```
reneolivo: Instead of doing this, please use:

```js
scope: {
  placeholder: '@timepickerSelectPlaceholder?'
}
```

**Sidenote:** I think we are missing the `?` modified for several of the bindings, which makes the binding optional. Can you please confirm?

reneolivo: Why the proposed solution? If the binding changes, the placeholder will change too (devs can use this: `<select ... time-picker-select-placeholder="{{ placeholder }}"></select>`)

org.civicrm.reqangular/src/common/directives/timepicker-select.directive.js
```diff
@@ -0,0 +1,58 @@
...
+        timepickerSelectInterval: '<'
+      },
+      restrict: 'A',
+      controllerAs: 'selector',
+      controller: ['$scope', timepickerSelectController],
+      template: $templateCache.get('timepicker-select.html')
+    };
+  }]);
+
+  timepickerSelectController.$inject = ['$scope', 'notificationService'];
```
reneolivo: please remove `notificationService`

PR #2255, opened 2017-10-30
PCHR-2638: Word replacement, Case(s) -> Workflow(s)

AkA84:
This PR changes the original word replacement of "Cases(s) -> Assignment(s)" to Cases(s) -> Workflow(s)" and does a bit of a clean up of the word replacement file


davialexandre:
@AkA84 How does this work for existing sites? I saw that you've updated the replacement names, which is what Civi uses to identify and control if they have been updated or not. Did you check if it is smart enough to catch this update or if it will simply create new replacements (and keep the old ones)?

AkA84:
@davialexandre by clearing the cache the replacements are updated automatically

PR #2257, opened 2017-10-31
PCHR-2843: Fix Leave Calendar for disabled absence types

igorpavlov:
## Overview

This PR fixes the Leave Calendar when there are leave requests created with an Absence Type that was disabled later on.

## Technical Details

The implementation is similar to https://github.com/civicrm/civihr/pull/2256, only those leave requests are loaded that belong to existing Absence Types.

```js
function loadMonthLeaveRequests () {
  return LeaveRequest.all({
    // ... existing filters
    type_id: { IN: _.pluck(vm.supportData.absenceTypes, 'id') }
  }, /* ... */)
  .then(function (leaveRequestsData) {
    // ...
  });
}
```

## Comments

This PR also refactors Leave Calendar Month component's tests a bit.

-------

✅ Jasmine Tests - passed
✅ Manual Tests - passed
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -142,18 +152,22 @@
               }));
             });
 
+            it('loads leave requests for *enabled* absence types only', function () {
```
deb1990: it should be "for `*active*`"

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
```diff
@@ -109,31 +109,41 @@
           });
 
           describe("contacts' work pattern calendar", function () {
+            var calendarCallRecentsArgs;
+
+            beforeEach(function () {
+              calendarCallRecentsArgs = Calendar.get.calls.mostRecent().args;
```
deb1990: How is this even working? You are storing the args even before the function is called.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
igorpavlov: Not really, it would be inconsistent with what we have in DB, API and the interface:

image

deb1990: The API accepts `is_active` parameter, so already it is not consistent as we can see. Its fine, please ignore.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-calendar-month.component.spec.js
igorpavlov: I do not think so, there is a function call `sendShowMonthsSignal` in the parent `describe` that triggers the API call. `sendShowMonthsSignal` sends a signal and then digests the rootscope.

deb1990: Okay, I missed that part, Apologies.

PR #2259, opened 2017-10-31
PCHR-2624: Remove quicksearch related code

AkA84:
## Overview
This PR makes several adjustments to support the changes in [this](https://github.com/compucorp/org.civicrm.shoreditch/pull/63) Shoreditch PR

### Moved quicksearch field script from hrui to shoreditch
All the code in hrui.js about changes to the quicksearch field behaviour had been moved to Shoreditch, as those changes are related to the look-and-feel of the menu, which is of course defined by the theme and not by hrui

### Improved jquery selector to avoid Safari issue
This selector for some reason caused a JS expression error in Safari
```js
$('.menumain:not(.crm-Self_Service_Portal')
```
Changing it to the follow fixed the issue
```js
$('.menumain').not('.crm-Self_Service_Portal')
```

### Amended the markup of the home menu item
The home menu item, defined in `BAO/Navigation.php`, had been changed to match the markup defined in https://github.com/compucorp/org.civicrm.shoreditch/pull/63


hrui/CRM/Core/BAO/Navigation.php
```diff
@@ -675,7 +675,9 @@ public static function createNavigation($contactID) {
 
       $hideLabel = ts('Hide Menu');
       $prepandString = "
-        <li class='menumain crm-link-home'><a href='$homeURL'>$homeIcon $homeLabel</a>";
+        <li class='menumain crm-link-home'><a href='$homeURL'>
+          <span class=\"menumain-icon\">$homeIcon</span>
```
igorpavlov: Why do you need to use `\"` sometimes and sometimes `'`? Could we simply use `'` for JS and `"` for HTML?

hrui/js/src/hrui.js
```diff
@@ -294,38 +229,14 @@
   function toggleActiveClassOnHoverOnAnyMainMenuItem () {
     var className = 'activetarget';
 
-    $('.menumain:not(.crm-Self_Service_Portal').hover(function () {
+    $('.menumain').not('.crm-Self_Service_Portal').hover(function () {
```
igorpavlov: Doesn't :not() work? Is it just for readability purposes?

AkA84: please read the PR description, i've explained it there

PR #1995, opened 2017-07-12
PCHR-2358: Update help links

bmnepali:
## Overview
Update the footer details, and create new "Help" and "Developer" menu in CiviCRM and hide the support menu.

## Before
Support is not hidden and Help and Developer menu are absent.
screen shot 2017-07-12 at 8 00 05 pm

Footer has more info with links:
screen shot 2017-07-12 at 8 12 28 pm

## After
Support is hidden and Help and Developer menu are present.
after

Footer has less links and information:
screen shot 2017-07-12 at 8 13 11 pm

## Technical Details
Modify "/civihr/hrui/hrui.php" file to Hide "Suppot" menu and "Help" and "developer" Menu

1. Disable "Support menu"
```php
/**
 * Changes to some core menu items
 *
 */
function _hrui_coreMenuChanges(&$params) {
  ...
  $supportNavId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'Support', 'id', 'name');

  unset($params[$supportNavId]);
  ...
}
```
2.Add help and Developer menu using _navigationMenu hook.
```php
function hrui_civicrm_navigationMenu(&$menu) {
  ...
  _createHelpMenu($menu);
  _createDeveloperMenu($menu);
}

/**
 * Creates Help Menu in navigation bar
 *
 * @param Array $menu List of available menu items
 */
function _hrui_createHelpMenu(&$menu) {
  _hrui_civix_insert_navigation_menu($menu, '', [
    'name' => ts('Help'),
    'permission' => 'access CiviCRM',
  ]);

  _hrui_civix_insert_navigation_menu($menu, 'Help', [
    'name' => ts('User Guide'),
    'url' => 'http://civihr-documentation.readthedocs.io/en/latest/',
    'target' => '_blank',
    'permission' => 'access CiviCRM'
  ]);

  _hrui_civix_insert_navigation_menu($menu, 'Help', [
    'name' => ts('CiviHR website'),
    'url' => 'https://www.civihr.org/',
    'target' => '_blank',
    'permission' => 'access CiviCRM'
  ]);

  _hrui_civix_insert_navigation_menu($menu, 'Help', [
    'name' => ts('Get support'),
    'url' => 'https://www.civihr.org/support',
    'target' => '_blank',
    'permission' => 'access CiviCRM'
  ]);
}

/**
 * Creates Developer Menu in navigation bar
 *
 * @param Array $menu List of available menu items
 */
function _hrui_createDeveloperMenu(&$menu) {
[...]

PR #1351, opened 2017-01-12
PCHR-1685: Update footer info

sleekvick:
# Old Footer

The old footer didn't have a logo and a link to https://civihr.org

civihr_version

# Iconmoon for logo

The logo at the bottom is from the iconmoon font, in the css the icon-logo class is used to reference to icon.

# Update footer info

The footer now contains the logo and a link to the https://civihr.org.

## Admin Footer
screen shot 2017-01-16 at 5 32 35 pm

## Dashboard footer

screen shot 2017-01-16 at 5 23 08 pm

PR #2264, opened 2017-11-03
PCHR-2832: Compatibility fixes for CiviCRM 4.7.27

davialexandre:
## Overview
This PR includes fixes to make CiviHR compatible with CiviCRM 4.7.27. It mainly comprised of two types of fixes:

- Update the Jenkinsfile so that Jenkins will create the test site using CiviCRM 4.7.27 rather than 4.7.22
- Make sure that all the overridden files are in sync with their original version in CiviCRM 4.7.27 (That is, they have everything that version + our customizations)

## Before

Jenkins would run the tests on a site using CiviCRM 4.7.22.

The overridden files in CiviHR were VERY outdated and based on different versions of CiviCRM.

These are the overridden files:

- hrui/CRM/Core/BAO/Navigation.php
- hrui/templates/CRM/common/footer.tpl
- hrjobcontract/CRM/Export/BAO/Export.php
- hrjobcontract/CRM/Hrjobcontract/Dedupe/Merger.php
- hrjobcontract/CRM/Hrjobcontract/Form/Merge.php
- hrjobcontract/templates/CRM/Hrjobcontract/Form/Merge.tpl
- hrjobcontract/templates/CRM/Hrjobcontract/Form/Merge.hlp

## After

Jenkins run the tests on a site using CiviCRM 4.7.27.

All the overridden files are now in sync with CiviCRM 4.7.27

## Technical Details

One of the reasons we override the Navigation.php file is to replace the CiviCRM logo with the CiviHR one in the main navigation menu. In the past, this was done by adding an image (using the `img` tag) to the menu markup and our override would simply change the URL of the image being added. However, this was changed in core. The markup is now a `span` tag with a specific class, which is them styled to have a background image as the logo. To conform to this change, I remove the customization that would change the image url and added some CSS to change the logo in hrui.css:

```css
#civicrm-menu .crm-logo-sm {
  background: url('../../assets/civihr.png') !important;
  background-size: 20px 20px !important;
  height: 20px !important;
  width: 20px !important;
}
```

The `!important` is necessary as otherwise the styles in civicrm-core will have precedence.

## Comments

For Jenkins, I also had to update the attachments patch, used for L&A. This attachment is currently kept on a gist, so it's not included as part of the PR. The changes to it were minimal. In fact, the code itself didn't change at all. The only difference is the position where the changes to Civi/Core/Container.php will be applied in the file, as it can be seen by this diff:

https://www.diffchecker.com/nJlwX3b2

Most of the overridden files were quite outdated. Some of them haven't been updated in almost one year. In addition to that, some of them have gone through some refactoring in core. For these reasons, the diffs in this PR will show a considerable number of changes. To make the review process easier, here are some links where you can compare the overridden file (after the sync) with its relative version in civicrm-core:

**Diff Navigation.php**
https://www.diffchecker.com/rOH7wsHS

**Export.php**
https://www.diffchecker.com/UP8HbpRV

**footer.tpl**
https://www.diffchecker.com/p2h7ey7N

**Merger.php**
https://www.diffchecker.com/SvKR430j

**Merge.php**
https://www.diffchecker.com/iC1vXeoC

**Merge.tpl**
https://www.diffchecker.com/k3ZxArM1

The Merge.hlp file didn't really have any customization in it, so a diff isn't necessary.

These files have been overridden by different reasons and in different PRs. Here is a list of these PRs, where you can find some more details about what they do:

**Merger.php, Merge.php, Merge.tpl, Merge.hlp**
Originally overridden in https://github.com/civicrm/civihr/pull/900.
Later updated in https://github.com/civicrm/civihr/pull/942 to make it compatible with CiviCRM 4.7.6

**footer.tpl**
Originally overridden in https://github.com/civicrm/civihr/pull/1203 to display the CiviHR version.
Next, it was updated to open links on new tabs and include the CiviHR logo: https://github.com/civicrm/civihr/pull/1351
Finally, it was updated to change some links https://github.com/civicrm/civihr/pull/1995

**Navigation.php**
Originally, it was part of the hrnavigation extension, where it was overridden to rename some link labels. This was done as part of PCHR-383, but, unfortunately, we didn't use Pull Requests back then.
Later, hrnavigation was merged into hrui: https://github.com/civicrm/civihr/pull/1149
Next, we customized the file further to change some more links: https://github.com/civicrm/civihr/pull/1212
And synced it with CiviCRM 4.7.25 at the beginning of 2017: https://github.com/civicrm/civihr/pull/1383

**Export.php** 
It was originally overridden to fix how the job contract end reason was being exported: https://github.com/civicrm/civihr/pull/1211
Later, it was updated by https://github.com/civicrm/civihr/pull/2087 and https://github.com/civicrm/civihr/pull/2123 to change the exported file to include enough information about leave entitlements, so that they could be imported in CiviHR 1.7


mickadoo:
In description: `s/The overridden files in CiviCRM/The overridden files in CiviHR`

Jenkinsfile
```diff
@@ -51,10 +51,10 @@ pipeline {
           def buildBranch = params.CIVIHR_BRANCH != '' ? params.CIVIHR_BRANCH : env.CHANGE_TARGET != null ? env.CHANGE_TARGET : env.BRANCH_NAME != null ? env.BRANCH_NAME : 'staging'
 
           // Build site with CV Buildkit
-          sh "civibuild create ${params.CIVIHR_BUILDNAME} --type hr17 --civi-ver 4.7.22 --hr-ver ${buildBranch} --url $WEBURL --admin-pass $ADMIN_PASS"
+          sh "civibuild create ${params.CIVIHR_BUILDNAME} --type hr17 --civi-ver 4.7.27 --hr-ver ${buildBranch} --url $WEBURL --admin-pass $ADMIN_PASS"
           sh """
             cd $DRUPAL_MODULES_ROOT/civicrm
-            wget https://gist.githubusercontent.com/davialexandre/deafc6d4929fbf58b97105e54bfb300f/raw/2ec70492a8d5ce419337cc58097c1dfd3e2d4df6/attachments.patch
+            wget -O attachments.patch https://gist.githubusercontent.com/davialexandre/199b3ebb2c69f43c07dde0f51fb02c8b/raw/0f11edad8049c6edddd7f865c801ecba5fa4c052/attachments-4.7.27.patch
```
mickadoo: Judging by the diff in the patch the only change was the line affected, is that right?

hrui/css/hrui.css
```diff
@@ -84,3 +84,10 @@ body:not(.toolbar-drawer) {
   margin-right: 1%;
 }
 
+#civicrm-menu .crm-logo-sm {
```
mickadoo: Where can I find this? I can see the `#civicrm-menu` but not the `.crm-logo-sm` class

mickadoo:
Thanks for preparing the diffs, makes it a lot easier to check the changes. Even so, it's a bit hard to understand what they're all doing because it's changes from a good few different PRs rolled into one :sweat_smile: 

I left some questions, not exactly requests for changes, more just requests for clarification so I'll leave the review status as open for now.

hrui/css/hrui.css
davialexandre: The markup is added by the Navigation.php file here: https://github.com/civicrm/civihr/pull/2264/commits/3c614e59b6f4441a24f0d524225821d504dc3929#diff-ca7c014b54483fa044ce8589d3f7ce51R570

It's then added as part of the menu, as it can be seen here:

captura de tela de 2017-11-08 09-38-00

Jenkinsfile
davialexandre: Yup, that's right. I've updated the ticket description with more information about this and a diff between the patches: https://www.diffchecker.com/nJlwX3b2

davialexandre:
@mickadoo thanks for the review. I've amended the ticket description to include to include some more information about where and why the overridden files were created/changed.

hrui/css/hrui.css
mickadoo: My bad, hadn't checked out the right branch in CiviHR. It's there now. I tried removing the `!important` and it still seems to work. Is it required? It might be an option to make the selector more specific either. I don't know much about CSS but I have the idea that using `!important` should be the last option.

davialexandre: I had some problems when I tried it without the `!important`, but I wouldn't be surprised if I did something wrong. But, to be honest, I  wouldn't care too much about this. In this release, we'll also have the menu updates (https://github.com/civicrm/civihr/pull/2262) which will completely change the menu and override this. Given that, I think it's ok to keep it as is.

PR #2265, opened 2017-11-03
PCHR-2525: UI-Select Related Bug Fixes

deb1990:
## Problem 1
Nothing happens when clicked on Dropdown Icon of the Multi Select.

## Before
before

## Technical details
1. As part of this PR, styles related to `ui-select` has been removed, as it is taken care from shoreditch(https://github.com/compucorp/org.civicrm.shoreditch/pull/65).

2. A new decorator `ui-select-focus.decorator.js` has been added, which focuses and opens the dropdown when clicked in the multi ui-select, but it suppresses the event when clicked on the `.ui-select-match` or `.ui-select-choices` .

## Problem 2
Also contacts are shown in Alphabetical order in 'Staff Member' dropdown of the Manager calendar.

## Before
manager_leave___hr17-4

## After
manager_leave___hr17-4

-------

✅ Jasmine Tests - passed
✅ JS distributive files - added
✅ Manual Tests - passed
✅  Backstop Tests - passed
⏹ PHP Unit Tests - not needed
✅  CSS distributive files - added


uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_manager_calendar.scss
```diff
@@ -30,10 +30,6 @@
 
     .ui-select-container {
```
AkA84: do we still need this?

deb1990: Yes we need it, it has the `height` property.

PR #2266, opened 2017-11-03
PCHR-2868: Move username in the dropdown menu

AkA84:
## Overview
This PR moves the username from being always visible in the user menu, to being visible only as part of the dropdown menu. This frees up some real estate in the top main menu of both the SSP and the admin

### SSP
ssp-before

### Admin
admin-before

### SSP
ssp-after

### Admin
admin-after


uk.co.compucorp.civicrm.hrcore/templates/CRM/HRCore/UserMenuMarkup.tpl
```diff
@@ -10,6 +9,9 @@
   </div>
   <nav class="chr_user-menu__dropdown">
     <ul>
+      <li class="chr_user-menu__dropdown__username">
+        <span>signed in as <strong>{$username}</strong></span>
```
igorpavlov: This instantly caught my attention when I looked at GIFs, I think "signed" should be capitalised.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_user-menu.scss
```diff
@@ -15,18 +15,13 @@
 }
 
 .chr_user-menu__data {
-  > .chr_user-menu__name {
-    float: left;
-  }
-
   > .chr_profile-card {
```
igorpavlov: Don't we add an empty line before nesting in order not to confuse with the properties?

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_user-menu.scss
```diff
@@ -44,12 +39,19 @@
     margin: 0;
   }
 
+  li {
+    line-height: 50px;
+    white-space: nowrap;
+
+    > a,
+    > span {
+      padding: 0 30px 0 20px;
```
igorpavlov: I think this should be a var, you use it twice in this file. 

P.S. I would recommend to use the suffix smth like "-gap" and not "-padding" so you can use it later with, for example, margins.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_user-menu.scss
```diff
@@ -62,3 +64,15 @@
...
+
+.chr_user-menu__dropdown__username {
+  color: $chr-user-menu-dropdown-username-color;
+
+  &:after {
+    border-top: 1px solid $gray-darker;
+    content: '';
+    display: block;
+    height: 0;
+    margin: 0 30px 0 20px;
```
igorpavlov: Please read above about the var.

PR #2250, opened 2017-10-26
PCHR-2716: Create time amount picker directive

igorpavlov:
## Overview

This PR add a custom Amount Time Picker (please do not confuse with a Timepicker Select directive). This directive is needed for Leave in Hours and may be reused in a lot of different places including Leave Request Modal, Job Contract leave tab etc.

## After

Note: this GIF was made in a temporary template for demonstration purposes only

## Technical Details

The directive can be used in the following way:

```js
<time-amount-picker
  time-amount-picker-value='vm.hours' // bind with a value
  time-amount-picker-min-amount='9.5' // optional, can be a variable, defaults to 0
  time-amount-picker-max-amount='18' // optional, can be a variable, defaults to 24
  time-amount-picker-interval='15' // optional, can be a variable, defaults to 1
></time-amount-picker>
```
`vm.hours` will be updated with an amount of hours (for example 11 or 15.75) once selectors changed.

----

✅ Jasmine Tests - passed
✅ JS distributive files - included
✅ Manual Tests - passed
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - n/a
⏹ CSS distributive files - n/a

PR #2269, opened 2017-11-06
PCHR-2716: Improve Time Amount Picker Directive

igorpavlov:
## Overview

The directive created in https://github.com/civicrm/civihr/pull/2250 lacks the following features:
- Re-parse incoming value when it is changed
- Support the case when minimum and maximum amounts are both 0
- Support disabling selectors

This PR adds these features.

### Support the case when minimum and maximum amounts are both 0
```js
$scope.maxAmount || 24; // will return 24 if 0
!isNaN(+$scope.maxAmount) ? +$scope.maxAmount : 24; // will return 0 if 0
```

### Support disabling selectors
```js
disabled: '<timeAmountPickerDisabled' // now supports another param
```
```html
<time-amount-picker time-amount-picker-disabled='true'></time-amount-picker>
<!-- or -->
<time-amount-picker time-amount-picker-disabled='vm.variable'></time-amount-picker>
```
### Re-parse incoming value when it is changed

```js
$scope.$watch('value', function () { // watches incoming value
  parseCurrentValue();
});
```

----------

✅ Manual Tests - passed
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - not needed


org.civicrm.reqangular/src/common/directives/time-amount-picker.directive.js
```diff
@@ -96,17 +97,21 @@ define([
...
       }
-
-      $scope.value = +vm.selectedHours + vm.selectedMinutes / 60;
     }
 
     /**
      * Starts watching inbound attributes for changes
      */
     function watchTimeAmountPickerOptions () {
+      $scope.$watch('disabled', function (disabled) {
```
reneolivo: This $scope.$watch is not needed since vm.disabled will be automatically updated

igorpavlov: @reneolivo vm.disabled will not be automatically updates since scope binding are not bound to the controller.

PR #2270, opened 2017-11-07
PCHR-2842: Endless Calendar Spinner fix

deb1990:
## Overview
As part of this PR, the endless spinner bug, when a period with more than 12 months is selected, is fixed.

## Before
before

## After
after

## Problem
The logic to wait for all the months to be destroyed was wrong. If a period with 12 months is selected, and then a period with 24 months is selected, we were waiting for `vm.months.length` number of elements to get destroyed, which is 24, but previously only 12 months existed.

Because of this flaw, the `makeSureMonthsAreNotInjected` promise was never getting resolved, and hence the bug came into place.

## Solution
- Moved the `makeSureMonthsAreNotInjected()` function call before the months are recalculated.


deb1990:
@AkA84 Fixed, as per your suggestion. Also I removed the first commit with previous solution to keep the history clean.

PR #2274, opened 2017-11-08
PHCR-2647: Merge PHP content

AkA84:
## Overview
_A brief description of the pull request. Try to keep it non-technical. Please mark your pull request with the appropriate [label(s)](https://github.com/civicrm/civihr/blob/staging/CONTRIBUTING.md#label-types)_

## Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

## After
_What has been changed. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

## Technical Details
_If the PR introduces noteworthy technical changes, please describe them here. Provide code snippets if necessary_

## Comments
_Anything else you would like the reviewer to note_


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Listener/AbstractListener.php
```diff
@@ -0,0 +1,25 @@
...
+
+    public function __construct(&$object) {
+      $this->object = &$object;
+    }
+
+    protected function canHandle() {
+      return $this->object instanceof $this->objectClass;
+    }
+
+    protected function isExtensionEnabled($key) {
```
omarabuhussein: isExtensionEnabled does not fit into this class, it has nothing to do with it,  I understand that you added it to avoid code duplication  but I think the effect of adding it to this class is worse that the effect of duplicating the code. 

I would say let's duplicate it for now until we find a more appropriate place to put such methods.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Listener/AbstractListener.php
```diff
@@ -0,0 +1,25 @@
+<?php
+  abstract class CRM_HRCore_Listener_AbstractListener {
+
+    protected $objectClass;
+    protected $object;
+
+    public function __construct(&$object) {
+      $this->object = &$object;
```
omarabuhussein: not all civicrm hooks pass what you call $object, I think it is more reasonable if you pass such variables at the onHookLisnter level, for example : 


```php
public function onBuildForm(&$object) {
}
```

and remove the $object property

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Listener/Form/Admin/Extensions.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+class CRM_HRCore_Listener_Form_Admin_Extensions extends CRM_HRCore_Listener_AbstractListener {
```
omarabuhussein: so here too, instead of CRM_HRCore_Listener_Form_Admin_Extensions use CRM_HRCore_HookListener_Form_Admin_Extensions

AkA84:
@davialexandre  @omarabuhussein 

So I have a solution now that I think is a decent improvement, but now I feel I have two solutions at once

This is the class infrastructure I have currently

files


* a base hook listener (`BaseListener`) it basically contains common methods. There is no real "pattern" to those methods, I just use it as a bucket for common stuff. Not pretty, I just wanted it work
* event-based hook listeners (`OnAlterMenu`, `OnConfig`, etc). They all have a `handle` method which expects whatever params where passed to the original hook. Originally I had thought of a single common event listener (basically what is now `BaseListener`) which would take care of everything not page or form related, but it was becoming colossal very quickly, so I split it into smaller classes
* a base object-based hook listener (`ObjectBasedListener`). The idea is that this is the parent class of all the form and page listeners, which are categorised as object-based because they all need an object representing whether the form or the page to function
* the page/form hook listeners (`Form_Contact`, `Page_CaseDashboard`). They inherit from `ObjectBasedListener` and offer different event-based methods (`onAlterContent()`, `onBuilForm()`, etc), which runs logic tied to the specific form/page

In `hrcore.php` I experimented with dynamically call the event-based hook listeners to make the code more concise and cool-looking (of course) and it seems to work pretty well

```php

function hrcore_civicrm_config(&$config) {
  _hrcore_civix_civicrm_config($config);
}

function hrcore_civicrm_install() {
  _hrcore_call_hook_listener();
}

function hrcore_civicrm_uninstall() {
  _hrcore_call_hook_listener();
}

function _hrcore_call_hook_listener() {
  $caller = debug_backtrace(FALSE, 2)[1];
  $hookName = substr($caller['function'], strrpos($caller['function'], '_') + 1);
  $listenerClass = 'CRM_HRCore_HookListener_EventBased_On' . ucfirst($hookName);

  // necessary for install, uninstall, enable, and disable hooks
  require_once 'CRM/HRCore/HookListener/BaseListener.php';
  require_once str_replace('_', '/', $listenerClass . '.php');

  call_user_func_array([new $listenerClass(), 'handle'], $caller['args']);
}
```

but here's the dilemma though (maybe i'm worrying too much about it, but anyway...). This is how the `buildForm` hook implementation looks like

```php
function hrcore_civicrm_buildForm($formName, &$form) {
  $listeners = [
    new ContactFormHookListener($form),
    new ExtensionsFormAdminHookListener($form),
    new OptionsFormAdminHookListener($form),
    new LocalizationFormAdminHookListener($form),
  ];

  foreach ($listeners as $listener) {
    $listener->onBuildForm();
  }
}
```

In here i'm using the form listeners, but to make things consistent with the way things have evolved, shouldn't it look more like this?

```php
function hrcore_civicrm_buildForm($formName, &$form) {
  _hrcore_call_hook_listener()
}

class CRM_HRCore_HookListener_EventBased_OnBuildForm {

  public function handle(&$config) {
    $listeners = [
      new ContactFormHookListener($form),
      new ExtensionsFormAdminHookListener($form),
      new OptionsFormAdminHookListener($form),
      new LocalizationFormAdminHookListener($form),
    ];

    foreach ($listeners as $listener) {
      $listener->onBuildForm();
    }
  }
}
```
but if that's the case (and i believe it is), then does it look ok to have those page/form listeners not being...listening at all (as they are used internally by the event-based listeners)? Should they be renamed, restructured in some different way?

davialexandre:
@AkA84 in Symfony (Civi uses Symfony internally for even dispatching) we have two ways of[ listening to events](http://symfony.com/doc/2.8/event_dispatcher.html): Listeners and Subscribers. The former is pretty much what you have in your Event based folder (A single class with an onSomething() method). The latter is similar to your object-based model, as it is just a single class that can listen to multiple events (usually on separate method to each event. So, in terms of organization, I think things are ok.

However, in terms of implementation, I think we can/must do better. The first thing I don't like much is the dynamic call to hooks relying on the backtrace. The backtrace is actually non-necessary at all. We could simply do something like this:

```php
function hrcore_civicrm_uninstall() {
  _hrcore_call_hook_listener('uninstall');
}

function _hrcore_call_hook_listener($hookName) {
  $listenerClass = 'CRM_HRCore_HookListener_EventBased_On' . ucfirst($hookName);
  call_user_func_array([new $listenerClass(), 'handle'], $caller['args']);
}
```

Note that the require_once for the BaseListener should not be inside the hook. Actually, you should not need it at all, as Civi autoloader should be able to load it automatically. 

Another option would be to take advantage of the features added by https://github.com/civicrm/civicrm-core/pull/9949. With it, you can do something like this:

```php
function hrcore_civicrm_config() {
  Civi::service('dispatcher')->addListener('hook_civicrm_uninstall', [new CRM_HRCore_Event_OnUninstallListener, 'onUninstall']);
}
```

And yeah, you'll will need at least one hook. But it's ok if it's nothing more than a list of hooks.

The sameway, I think we can fully embrace the Symfony way of doing things here and use Subscribers for the page objects. This would probably require some changes to the classes though:

- First, they would need to extend the `Symfony\Component\EventDispatcher\EventSubscriberInterface;` interface and implement the `getSubscribedEvents()` (equivalent to the `canHandle()` you have at the moment. Here's a non function example of that

```php
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
class CRM_HRCore_Event_ContactPageSubscriber implements EventSubscriberInterface {
  public function getSubscribedEvents() {
      return [
         'hook_civicrm_buildForm' => 'onBuildForm',
         'hook_civicrm_pageRun' => 'onPageRun'
      ]
  }

  /**
   * We need all the params here though, as Civi will directly call this method
   */
  public function onBuildForm($all, $the, $params, $unfortunatelly) {
  }

  /**
   * We need all the params here though, as Civi will directly call this method
   */
  public function onPageRun($all, $the, $params, $unfortunatelly) {
  }
}
```

Next, the same way as you would do to an Event Listener, you can add the Subscriber in using the config hook:

```php
function hrcore_civicrm_config() {
  Civi::service('dispatcher')->addSubscriber(new CRM_HRCore_Event_ContactPageSubscriber());
}
```

Note that I never really used events like this in Civi, so there might be things that won't work (maybe the subscriber thing) and others that will require some tweaks.

Also note that, with this approach, I think we can change things in terms of organization a little bit. In the examples, I've put everything under the  `CRM_HRCore_Event` folder/namespace, and then I used the `Subscriber` and `Listener` suffixes to differentiate things.

omarabuhussein:
Hmmm, I have some notes : 
1- tbh I don't like how _hrcore_call_hook_listener() is written, it feel hacky to me, all hooks including the current should stick to how you wrote it original : 

```php
function hrcore_civicrm_buildForm($formName, &$form) {
  $listeners = [
    new ContactFormHookListener($form),
    new ExtensionsFormAdminHookListener($form),
    new OptionsFormAdminHookListener($form),
    new LocalizationFormAdminHookListener($form),
  ];

  foreach ($listeners as $listener) {
    $listener->onBuildForm();
  }
}
```
in my opinion.

2- BaseListener doesn't add anything here, this is wrong from of inheritance (or I would say misuse), use composition instead.

3- CiviCRM doc already has categorization for hooks : https://docs.civicrm.org/dev/en/master/hooks/ 
maybe it is better to sticket to it instead of inventing a new one ?

davialexandre:
Closing this as we decided we'll first discuss the extensions organization before we proceed with any extension merging task

PR #2277, opened 2017-11-09
PCHR-2878: TOIL validation for Leave in Hours

tunbola:
## Overview
When creating a TOIL request, the amount to be accrued is validated against the values of the toil amounts option group. This was causing an a validation error when requesting TOIL in hours and the number of hours does not match the option values in the toil amounts option group because for TOIL in hours, the user can enter a custom amount in hours for TOIL to be accrued

## Before
Only TOIL in hours that match the values of the TOIL amount option group can be requested
toilbefore

## After
The user can request TOIL for a custom amount in hours that is not part of the values of the TOIL amount option group.
toilafter

## Technical Details
- I had to add the specification for the `toil_to_accrue` param for the LeaveRequest.create API. The API field was bound to the toil_amount option group and any value passed in the parameter that is not part of the toil amount option group throws the error `CiviCRM_API3_Exception: 'value' is not a valid option for field toil_to_accrue`. The specification added allows any parameter that is float to be passed, since the BAO already validates the Toil amount to accrue for Leave in days

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -87,12 +87,12 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     self::validateRequestType($params);
     self::validateTOILFieldsBasedOnRequestType($params);
     self::validateSicknessFieldsBasedOnRequestType($params);
-    self::validateLeaveRequestFieldsBasedOnAbsenceTypeCalculationUnit($params);
+    list($absenceType, $absencePeriod) = self::loadCommonObjects($params);
+
+    self::validateLeaveRequestFieldsBasedOnAbsenceTypeCalculationUnit($params, $absenceType);
     self::validateStartDateNotGreaterThanEndDate($params);
     self::validateNoOverlappingLeaveRequests($params);
     self::validateLeaveDatesDoesNotOverlapContractsWithLapses($params);
-
-    list($absenceType, $absencePeriod) = self::loadCommonObjects($params);
     self::validateAbsenceTypeIsActiveAndValid($params, $absenceType);
     self::validateTOILRequest($params, $absenceType, $absencePeriod);
     self::validateLeaveDaysAgainstAbsenceTypeMaxConsecutiveLeaveDays($params, $absenceType);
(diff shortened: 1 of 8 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -13,6 +13,14 @@ function _civicrm_api3_leave_request_create_spec(&$spec) {
...
     'api.required' => 0,
   ];
+
+  $spec['toil_to_accrue'] = [
+    'name' => 'toil_to_accrue',
+    'title' => 'TOIL to accrue amount',
+    'description' => 'Amount of TOIL to accrue for the request',
+    'type' => CRM_Utils_Type::T_FLOAT,
+    'api.required' => 0,
+  ];
```
davialexandre: Instead of duplicating all this information that is coming from the LeaveRequest DAO, perhaps we can just unset what we don't need. Can you please check if something would work?

```php
unset($spec['toil_to_accrue']['pseudoconstant']);
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -286,22 +286,30 @@ private static function validateTOILRequest($params, $absenceType, $absencePerio
     if($params['request_type'] !== self::REQUEST_TYPE_TOIL) {
       return;
     }
-
-    self::validateTOILToAccrueIsAValidOptionValue($params);
+    $isCalculationUnitInHours = AbsenceType::isCalculationUnitInHours($params['type_id']);
```
davialexandre: This static method always bothered me a bit and seeing how it is being used here gave me an idea:

- Make it an instance method, instead of static. In terms of performance, it will be the same, since currently the method already creates an AbsenceType instance anyway. The difference is that now this instantiation will happen outside the method.
- Note that this `validateTOILRequest` method receives an AbsenceType instance. We can use that, and pass it to the other toil validation methods, the same way as we are currently doing for `validateTOILPastDays()`

I quickly checked other places that are also using this static method and I think they can easily be updated to use an instance method instead. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2462,6 +2462,38 @@ public function testLeaveRequestCanNotBeCreatedWhenRequestTypeIsToilAndToilToAcc
...
+    //Max accrual of 10 hours
+    $maxLeaveAccrual = 10;
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'allow_accruals_request' => true,
+      'calculation_unit' => 2,
+      'max_leave_accrual' => $maxLeaveAccrual,
+    ]);
+
+    $this->setExpectedException(
+      'CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException',
```
davialexandre: Please use `CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException::class` instead

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Good idea

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
tunbola: This works!

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/LeaveRequest.php
```diff
@@ -14,13 +14,7 @@ function _civicrm_api3_leave_request_create_spec(&$spec) {
...
   ];
 
-  $spec['toil_to_accrue'] = [
-    'name' => 'toil_to_accrue',
-    'title' => 'TOIL to accrue amount',
-    'description' => 'Amount of TOIL to accrue for the request',
-    'type' => CRM_Utils_Type::T_FLOAT,
-    'api.required' => 0,
-  ];
+  unset($spec['toil_to_accrue']['pseudoconstant']);
```
davialexandre: Please leave a comment here explaining why this was necessary

PR #2281, opened 2017-11-10
PCHR-2654: Next Leave fixes

reneolivo:
## Overview

* The Next Leave section now ignores leave requests that have disabled absence types.
* The next leave request now uses Moment dates instead of strings to display date information.

#### Before
download

#### After
download

**Note:** the Holiday absence type was disabled for both cases.

#### Technical Details
The leave request fetched can only be of the absence types provided by the leave widget component. This is an excerpt of what the request looks like in the next leave component:

```js
LeaveRequest.all({
  contact_id: vm.contactId,
  from_date: { '>=': today },
  request_type: 'leave',
  status_id: { IN: getStatusIds() },
  type_id: { IN: getAbsenceTypeIds() }, // <---- here
  options: { limit: 1, sort: 'from_date DESC' }
}, null, null, null, false)
```

Since the absence types provided by the leave widget are active ones only, disabled leave requests are not returned.

As a side note, `_.pluck` was used to return the ids and values from the absence types and statuses isntead of using `array.map` since pluck  can be more straightforward:

```
function getAbsenceTypeIds () {
  return _.pluck(vm.absenceTypes, 'id');
}
```

#### Before
download

#### After
download

#### Technical details
AngularJS' date filter doesn't support strings that contain time portions. To fix this, the strings were converted to Moment dates:

```js
function convertStringsToMomentDates () {
  if (vm.nextLeaveRequest.from_date) {
    vm.nextLeaveRequest.from_date = moment(vm.nextLeaveRequest.from_date);
  }

  if (vm.nextLeaveRequest.to_date) {
    vm.nextLeaveRequest.to_date = moment(vm.nextLeaveRequest.to_date);
  }
}
```

And displayed in the following manner:

```html
<div class="date-range-calendar__to" ng-if="nextLeave.nextLeaveRequest.to_date && !nextLeave.nextLeaveRequest.to_date.isSame(nextLeave.nextLeaveRequest.from_date, 'day')">
  <span class="date-range-calendar__day">{{nextLeave.nextLeaveRequest.to_date.format('D')}}</span>
  <span class="date-range-calendar__month">{{nextLeave.nextLeaveRequest.to_date.format('MMM')}}</span>
  <span class="date-range-calendar__year">{{nextLeave.nextLeaveRequest.to_date.format('YYYY')}}</span>
  <span class="date-range-calendar__time text-muted">
    ({{nextLeave.dayTypes[nextLeave.nextLeaveRequest.to_date_type].label}})
  </span>
</div>
```

It would be good to add this change to the `leave-request.instance` file, but it would potentially break a lot of functionality through the apps so this approach was not taken for now.


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-next-leave.component.js
```diff
@@ -79,6 +80,20 @@ define([
...
+     */
+    function convertStringsToMomentDates () {
+      if (vm.nextLeaveRequest.from_date) {
+        vm.nextLeaveRequest.from_date = moment(vm.nextLeaveRequest.from_date);
+      }
+
+      if (vm.nextLeaveRequest.to_date) {
+        vm.nextLeaveRequest.to_date = moment(vm.nextLeaveRequest.to_date);
+      }
+    }
```
bmnepali: Just wondering if we do like this in this case:
```js
function convertStringsToMomentDates () {
   vm.nextLeaveRequest.from_date = vm.nextLeaveRequest.from_date && moment(vm.nextLeaveRequest.from_date);
   vm.nextLeaveRequest.to_date = vm.nextLeaveRequest.to_date && moment(vm.nextLeaveRequest.to_date);
}
```

PR #2283, opened 2017-11-10
PCHR-2683: Hide unentitled absence types

reneolivo:
## Overview
This PR hides unentitled absence types from the leave widget balances section. These can happen when the absence type was created after the user entitlements were configured.

## Before
download

## After
download

## Technical Details
The problem arose because absence types were being matched against entitlements, even if the user didn't have any entitlements for that absence type.

To fix this, the structure for the absence type entitlements was changed so it holds the entitlements and their matching absence type:

```js
function mapAbsenceTypesWithTheirEntitlements (entitlements) {
  return vm.absenceTypes.map(function (absenceType) {
    var entitlement = _.find(entitlements, function (entitlement) {
      return +absenceType.id === +entitlement.type_id;
    });

    return {
      absenceType: absenceType,
      entitlement: entitlement
    };
  });
}
```

And then absence types with no entitlements are filtered out:

```js
function filterOutUnentitledAbsenceTypes (absenceTypeEntitlements) {
  return absenceTypeEntitlements.filter(function (absenceTypeEntitlement) {
    return absenceTypeEntitlement.absenceType && absenceTypeEntitlement.entitlement;
  });
}
```

**Question:** why not iterate over entitlements instead of absence types?
Because absence types can be disabled, so entitlements for disabled absence types would be displayed. This solution fixes both scenarios.

#### Side fix:

Next entitlements spec file was fixed so it tests for moment dates included in the next leave request instead of testing for strings.

```js
LeaveRequest.all({
  contact_id: contactId,
  from_date: { '>=': moment().format(sharedSettings.serverDateFormat) },
  request_type: 'leave',
  status_id: { IN: leaveRequestStatusIds },
  options: { limit: 1, sort: 'from_date DESC' }
}, null, null, null, false)
.then(function (response) {
  expectedNextLeave = _.clone(response.list[0]);
  expectedNextLeave['type_id.title'] = getAbsenceTypeTitleFor(expectedNextLeave);
  expectedNextLeave.balance_change = Math.abs(expectedNextLeave.balance_change);
  expectedNextLeave.from_date = moment(expectedNextLeave.from_date); // here
  expectedNextLeave.to_date = moment(expectedNextLeave.to_date); // and here
  expectedRequestStatus = getExpectedRequestStatus(expectedNextLeave);
});
```


reneolivo:
This PR has been outdated by recent changes.

PR #2284, opened 2017-11-10
PCHR-2832: Fix Job Roles and Contracts export issues

davialexandre:
## Overview

After updating CiviCRM to 4.7.27, it became impossible to export contacts with fields from Job Roles and Contracts. Whenever such field was present in the export mapping, the operation would end in a 500 error.

## Before

Trying to export contacts with information/fields from Job Roles and Contracts was resulting in a 500 error.

## After

All the fields from Job Roles and Contracts can be selected to be exported and the CSV file is created and downloaded successfully.

## Technical Details

On https://github.com/civicrm/civicrm-core/pull/9727 some new metadata has been added to the DAOs fields. Each field now has the entity name and the bao/dao class. After that, these new metadata started being used in different parts of CiviCRM and one of them is the export. For that process, as it can be seen here https://github.com/civicrm/civicrm-core/blob/0606198bdac72ea958414e8560ab26550de4f6fd/CRM/Contact/BAO/Query.php#L5954, Civi uses the `bao` property to get information about pseudconstants (Option values). This made that any attempt to export Job Roles or Contract
fields using pseudconstants impossible, as it would result in a Fatal Error.

To fix this, I've added the missing metadata to all the DAOs of these two extensions. Unfortunately, I have to do this manually. The Job Contract extension doesn't have the necessary schema XMLs to build and update the DAOs. On the other hand, the Job Roles extension does have these files, however, one of its fields creates an FK to one of the Job Contracts tables, meaning that, in order to update things for Job Roles we would need the files from Job Contracts. This missing files will be fixed later, but for now the updates need to happen manually.

## Comments

As part of this PR there are a couple more minor updates. The first one is the refactoring of the `build-dao.sh` script to only require a single param (the extension key) to be executed. The second one is the updated of some other extensions DAOs in order to follow the new formatting and include any missing metadata. Different from Job Roles and Contracts, these have been updated automatically, using the `build-dao.sh` script.


bin/build-dao.sh
```diff
@@ -8,34 +8,27 @@
...
-EXTSRCDIR="$EXTROOT/CRM/$NAMESPACE"
-EXTDAODIR="$EXTSRCDIR/DAO"
-XMLBUILD="$EXTROOT/build/xml/schema"
-
-function validateOptions() {
-  if [ -z "$CIVIROOT" -o ! -d "$CIVIROOT" ]; then
-    echo "ERROR: invalid civicrm-dir: [$CIVIROOT]"
-    printUsage
+function validatePaths() {
+  if [ -z "$CIVIROOT" ]; then
```
mickadoo: Any reason you decided to remove the check if it's a directory?

hrjobcontract/CRM/Hrjobcontract/DAO/HRJobContract.php
```diff
@@ -151,6 +151,8 @@ static function &fields()
           'export' => false,
           'import' => false,
           'where' => 'civicrm_hrjobcontract.id',
+          'entity' => 'HRJobContract',
+          'bao' => 'CRM_Hrjobcontract_DAO_HRJobContract',
```
mickadoo: Some of the DAOs have an extra `localizable` field, but other don't. What is this based on?

contactaccessrights/CRM/Contactaccessrights/DAO/Rights.php
```diff
@@ -101,107 +71,112 @@ class CRM_Contactaccessrights_DAO_Rights extends CRM_Core_DAO
...
    * @return array
    *   [CRM_Core_Reference_Interface]
    */
-  static function getReferenceColumns()
-  {
-    if (!self::$_links) {
-      self::$_links = static ::createReferenceColumns(__CLASS__);
-      self::$_links[] = new CRM_Core_Reference_Basic(self::getTableName() , 'contact_id', 'civicrm_contact', 'id');
+  static function getReferenceColumns() {
+    if (!isset(Civi::$statics[__CLASS__]['links'])) {
```
mickadoo: This method (and a few others) seems to have been updated in the autogenerated DAOs, should it also be updated in the manually updated ones?

mickadoo:
Seems like this change in core might have been better to include a default instead of requiring the name of the same class repeated again and again.

bin/build-dao.sh
davialexandre: Based on the tests I did, the `cv path -d '[civicrm.root]` command either returns a valid path, when it can find one or empty when it can't. Based on that, it's ok to assume that if it's not empty than it exists and it's a directory. Basically, `cv` is the one doing that validation now

contactaccessrights/CRM/Contactaccessrights/DAO/Rights.php
```diff
@@ -26,56 +26,26 @@
...
    *
    * @var string
    */
   static $_tableName = 'civicrm_contactaccessrights_rights';
   /**
-   * static instance to hold the field values
-   *
-   * @var array
-   */
-  static $_fields = null;
```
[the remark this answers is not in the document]
davialexandre: They were autogenerated

contactaccessrights/CRM/Contactaccessrights/DAO/Rights.php
davialexandre: Ideally they should, but honestly, I don't feel very comfortable applying these changes manually. The DAOs from Job Roles and Contracts have lots of customizations and some of them are not upgrade-safe (for example, the contract's DAOs all extend a Base custom DAO instead of CRM_Core_DAO). It's hard to be sure that nothing will break if we do this. So, since everything is working fine, for now, I prefer to not do this for now. Also, there will be some work to improve things in Job Contracts in the future (hopefully soon!) and that will be a better opportunity to fix this

hrjobcontract/CRM/Hrjobcontract/DAO/HRJobContract.php
davialexandre: It's used only for multi-language sites. It comes from the schema xml files. If it's not present in these files, GenCode will automatically set them to 0. I've manually did this for the Job Roles and Contracts DAOs

PR #2288, opened 2017-11-14
PCHR-2877: Use core markup in overridden navigation file

AkA84:
## Overview
In order to have the CiviHR logo in the main menu, and to have it being a simple link to the homepage rather than a toggle of a sub-menu dropdown, the [`Bao/Navigation.php`](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/BAO/Navigation.php) file was [overridden](https://github.com/civicrm/civihr/blob/staging/hrui/CRM/Core/BAO/Navigation.php) in the `hrui` extension, so that we could use the markup that we needed

This PR reverts back the markup to the core version (the file is still overridden in general though), and does all the markup customization via JS

## Before
core-logo

This is what the core markup looks like
```html
<li class="menumain crm-link-home" tabindex="3">
  <span class="crm-logo-sm"></span>
  <ul id="civicrm-home" style="display: none;">
    <li><a href="/civicrm/tasksassignments/dashboard#/tasks">CiviHR Home</a></li>
    <li><a href="#" class="crm-hidemenu">Hide Menu</a></li>
    <li><a href="/civicrm/logout?reset=1" class="crm-logout-link">Log out</a></li>
  </ul>
</li>
```

## After
logo

This is what the logo menu item looks like after our JS kicks in
```html
<li class="menumain crm-link-home" tabindex="3">
  <a href="/civicrm/tasksassignments/dashboard#/tasks">
    <span class="menumain-icon">
      <span class="chr_logo chr_logo--default-color"></span>
    </span>
    <span class="menumain-label">Home</span>
  </a>
</li>
``` 

## Technical Details
The assets (js, css) were added via the `hook_civicrm_pageRun` and `hook_civicrm_buildForm` hooks, but those hooks might end being called multiple times on the same page, for example when some ajax-base interaction happens.

```php
function hrui_civicrm_pageRun($page) {
  if (CRM_Core_Config::singleton()->debug) {
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/civihr-popup/attrchange.js');
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/civihr-popup/civihr-popup.js');
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/hrui.js');
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/contact.js');
  } else {
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/dist/hrui.min.js');
  }

  CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.hrui', 'css/hrui.css');
}

function hrui_civicrm_buildForm($formName, &$form) {
  CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.hrui', 'css/hrui.css');

  if (CRM_Core_Config::singleton()->debug) {
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/hrui.js');
  } else {
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/dist/hrui.min.js');
  }
}
```

This resulted in the files being added multiple times and, in the case of JS code, the handler on `$(document).ready()` being called multiple times as well.

The solution was to move the assets inclusion in the `hook_civicrm_coreResourceList` hook instead
```php
function hrui_civicrm_coreResourceList(&$items, $region) {
  if ($region == 'html-header') {
    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/dist/hrui.min.js');
    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.hrui', 'css/hrui.css');
  }
}
```

Additionally, sourcemap are generated from the concatenated files and put inline in the minified file, so that we can always serve the min file also when we are debugging 

## Comments
Swapping the icon (and changing its menu item's markup) via JS is not really ideal, but if we want to leave the original file untouched there is no other way around currently


hrui/hrui.php
```diff
@@ -27,6 +27,24 @@
...
+ */
+function hrui_civicrm_coreResourceList(&$items, $region) {
+  if ($region == 'html-header') {
+    if (CRM_Core_Config::singleton()->debug) {
+      CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/civihr-popup/attrchange.js');
+      CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/civihr-popup/civihr-popup.js');
+      CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/hrui.js');
+      CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/src/contact.js');
+    } else {
+      CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrui', 'js/dist/hrui.min.js');
```
deb1990: Didn't we decide to not using debug mode anymore? In that case, we can just include the min files, and let map files do its job.

AkA84: This is not an angular/requireJS app, it's a bunch of simple JS files, so i kept the original logic. There is not even any map file being created, it has nothing to do with the change in approach we made to the angular apps

deb1990: Okay, Do you think we can implement the same here too by updating the gulpfile? Then we would move towards having a single approach for all JS based extensions.

hrui/js/src/hrui.js
```diff
@@ -11,26 +11,36 @@
...
-      toggleActiveClassOnHoverOnAnyMainMenuItem();
       addUserMenuToMainMenu();
+      amendAppLogoMenuItem();
       amendApplicationForm();
       useFontAwesomeArrowsInSubMenuItems();
+      toggleActiveClassOnHoverOnAnyMainMenuItem();
     });
 
   /**
+   * Customizes the app logo menu item, swithing from the CiviCRM logo
```
davialexandre: s/swithing/switching/

hrui/CRM/Core/BAO/Navigation.php
```diff
@@ -567,33 +567,31 @@ public static function createNavigation($contactID) {
...
       self::retrieve($homeParams, $homeNav);
       if ($homeNav) {
-        if (array_key_exists('query', parse_url($homeNav['url']))) {
-          list($path, $q) = explode('?', $homeNav['url']);
-        } else {
-          $q = NULL;
-          $path = $homeNav['url'];
-        }
-
+        list($path, $q) = explode('?', $homeNav['url']);
```
davialexandre: Are you sure we should remove this override? It was added as part of https://github.com/civicrm/civihr/pull/924 to fix a Notice due to the fact that the T&A URLs don't have a query string. Does that still hold true?

AkA84:
> @AkA84 Does this PR also fix the spinning logo problem?

@davialexandre yes it does, cause the spinning effect is applied on the `.crm-logo-sm` which i remove via JS

hrui/CRM/Core/BAO/Navigation.php
AkA84: good catch, will put back

PR #2293, opened 2017-11-14
PCHR-2921: Add "Son/Daughter" Emergency Contact relationship option

mickadoo:
## Overview

This PR creates a new Emergency Contact relationship option, "Son/Daughter"

## Before

The "Son/Daughter" option did not exist.

## After

The "Son/Daughter" option exists.


mickadoo:
@davialexandre you're right, I was even confused myself. I saw the relationship type updgrades in the hrui extension when I was searching for the string names and continued making the upgrader in there, even though they're different things.

PR #2295, opened 2017-11-15
PCHR-2698: Entitlement calculation updates for Leave in Hours

davialexandre:
## Overview
The entitlement calculation for leave in hours or in days is exactly the same:

`(CE x (WDTWIP / WDIP)) + Number of Public Holidays in Period`

Where:
**CE**: Contractual Entitlement
**WDTW**: Working Days to Work in Period
**WDIP**: Working Days in Period

This works for both because the `WDTWIP / WDIP` doesn't really return a number of days, but the fraction of the amount to be worked during a period, either if it's day or hours. Example:

**Days to work in period**: 125
**Days in period**: 250

This will give us 0.5, meaning that the user is entitled to only half of whatever is set in the contractual entitlement. If the contractual entitlement is 200h, then the proposed entitlement will be 100h. The same way, if the contractul entitlement is 25d, then the proposed entitlement will be 12.5d.

For absence types in hours, Public Holidays are guaranteed to not be added because this setting will always be false in both the absence type and in the contract.

This all means that nothing was updated to the calculation logic itself. However, lots of places in the code were specifically refering to days and this could make developers confused as the code should also work for hours. This PR updates this to make the code and comments more generic in order to avoid such confusion.

## Before

Multiple comments, tests and methods around the `EntitlementCalculation` service had mentions to days.

## After

Methods, comments and tests that worked for both days an hours have been updated to use more generic terms.

Mainly, two methods in the `EntitlementCalculation` service have been renamed:

- `getNumberOfDaysTakenOnThePreviousPeriod()` has been renamed `getAmountUsedInPreviousPeriod()`
- `getNumberOfDaysRemainingInThePreviousPeriod()` has been renamed to `getPreviousPeriodBalance()`

For comments and everything else, the word Amount was usually used to replaced things like "Number of Days". In some places, the word Amount wasn't really necessary and it was dropped (for example, instead of "Number of days brought forward", we now simply have "Brought Forward"

## Comments
The Absence Type has some fields that still mention days (e.g. max_number_of_days_to_carry_forward) but this will not be updated as part of this PR, as it's not related to the  Entitlement Calculation. These fields will be updated late by a separate PR.

-------
[x] Tests pass


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
```diff
@@ -335,8 +335,8 @@ private function exportCSV() {
         'employee_name' => $calculation->getContact()['display_name'],
         'leave_type' => $calculation->getAbsenceType()->title,
         'prev_year_entitlement' => $calculation->getPreviousPeriodProposedEntitlement(),
-        'days_taken' => $calculation->getNumberOfDaysTakenOnThePreviousPeriod(),
-        'remaining' => $calculation->getNumberOfDaysRemainingInThePreviousPeriod(),
+        'days_taken' => $calculation->getAmountUsedInPreviousPeriod(),
```
tunbola: I think we should rename this array key to `amount_taken`

tunbola:
In the after section
s/the work Amount/the word Amount

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
davialexandre: As this is more related to the screen itself, than to the Entitlement Calculation service, I decided to not touch it for now and then change it as part of PCHR-2675

PR #2300, opened 2017-11-16
PCHR-2883: Improve Absence Types calculation unit selector

igorpavlov:
## Overview

This PR disables the Absence Type calculation unit selector if it cannot be changed (for example, when such an absence type has already been used in leave requests).

## Technical Details

In PHP form builder the `disabled` attribute is now conditionally set:

```php
$this->addSelect(
  'calculation_unit',
  $this->getCalculationUnitSelectorParams(),
  TRUE
);
// ...
private function getCalculationUnitSelectorParams () {
  $calculationUnitSelectorParams = [
    'options' => $this->getCalculationUnitOptions(),
    // ...
  ];

  if ($this->absenceTypeHasEverBeenUsed()) {
    $calculationUnitSelectorParams['disabled'] = 'disabled';
  }

  return $calculationUnitSelectorParams;
}
```

However, this is not enough since the disabled selector would not provide a value on form submit which makes such an Absence Type updating impossible due to the failing backend validation. 

Removing `disabled` attribute from the `<select>` would not work since the CRM selector watches the original `<select>` and unsets `disabled` attribute for itself as well. Same happens if you then apply `disabled` attribute to the CRM selector - the original `<select>` would set `disabled` attribute to itself.

So another method needs to be performed on the front-end: adding `<input type hidden>` just before the `<select>`, watch `<select>` value and set it to the hidden input.

```js
function supportValueSubmissionOfAllSelectorsIfDisabled () {
  var $selectors = $('select.crm-select2');

  $.each($selectors, function (index, select) {
    supportValueSubmissionOfSelectorIfDisabled($(select));
  });
}

function supportValueSubmissionOfSelectorIfDisabled ($select) {
  var $inputHidden = $('<input type="hidden" name="' +
    $select.attr('name') + '" value="' +
    $select.val() + '" />');

  $select.before($inputHidden);
  $select.removeAttr('name');
  $select.on('change', function () {
    $inputHidden.val($select.val());
  });
}

$(document).ready(function() {
  // ...
  supportValueSubmissionOfAllSelectorsIfDisabled();
});
```

## Comments

Changing a selector to a label instead would not work since flickering will happen.

-----

✅ Manual Tests - passed
✅ PHP Unit Tests - passed
⏹ Jasmine Tests - n/a
⏹ JS distributive files - not needed
⏹ Backstop Tests - omitted

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
@@ -149,11 +149,7 @@ class CRM_HRLeaveAndAbsences_Form_AbsenceType extends CRM_Core_Form {
     }
     $this->addSelect(
       'calculation_unit',
-      [
-        'options' => $this->getCalculationUnitOptions(),
-        'option_url' => NULL,
-        'label' => ts('Calculate Leave in')
-      ],
+      $this->getCalculationUnitSelectorParams(),
       TRUE
     );
     $this->addYesNo(
@@ -272,6 +268,25 @@ class CRM_HRLeaveAndAbsences_Form_AbsenceType extends CRM_Core_Form {
   private function getDAOFieldAttributes($field) {
     $dao = 'CRM_HRLeaveAndAbsences_DAO_AbsenceType';
     return CRM_Core_DAO::getAttribute($dao, $field);
+  }
+
+  /**
+   * Returns parameters needed for the calculation unit selector
+   *
+   * @return array
+   */
+  private function getCalculationUnitSelectorParams () {
+    $calculationUnitSelectorParams = [
+      'options' => $this->getCalculationUnitOptions(),
+      'option_url' => NULL,
+      'label' => ts('Calculate Leave in')
+    ];
+
+    if ($this->absenceTypeHasEverBeenUsed()) {
+      $calculationUnitSelectorParams['disabled'] = 'disabled';
+    }
+
+    return $calculationUnitSelectorParams;
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
@@ -277,7 +277,6 @@ class CRM_HRLeaveAndAbsences_Form_AbsenceType extends CRM_Core_Form {
    */
   private function getCalculationUnitSelectorParams () {
     $calculationUnitSelectorParams = [
-      'options' => $this->getCalculationUnitOptions(),
       'option_url' => NULL,
       'label' => ts('Calculate Leave in')
     ];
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -320,14 +320,28 @@
...
+                }
+
                 $(document).ready(function() {
                   initToilControls();
                   initCarryForwardControls();
                   initColorPicker();
                   initDeleteButton();
-                  initCalculationUnitControls()
+                  initCalculationUnitControls();
+                  supportDisabledSelector($('select[name=calculation_unit]'));
```
deb1990: Something like `supportValueSubmissionOfDisabledSelector()` would be more appropriate.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
```diff
@@ -275,6 +271,25 @@ private function getDAOFieldAttributes($field) {
...
+   * Returns parameters needed for the calculation unit selector
+   *
+   * @return array
+   */
+  private function getCalculationUnitSelectorParams () {
+    $calculationUnitSelectorParams = [
+      'options' => $this->getCalculationUnitOptions(),
+      'option_url' => NULL,
+      'label' => ts('Calculate Leave in')
+    ];
```
tunbola: The function `getCalculationUnitOptions()` is no longer needed, the reason I created the function was to find a workaround to not disable the calculation unit field and make sure that the selected option is the only option.

We can simply change the the `$calculationUnitSelectorParams` array to

> $calculationUnitSelectorParams= [
>       'option_url' => NULL,
 >      'label' => ts('Calculate Leave in')
 >    ];

Civi will automatically build the select options from the calculation unit option group.
Also we should remove the `getCalculationUnitOptions()` function.

igorpavlov:
@deb1990 I amended the implementation slightly. The new implementation supports enabling selectors back (if needed) and fixes all selectors as well for the future.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
igorpavlov: Done, but slightly differently.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -320,14 +320,44 @@
...
+                  });
+                }
+
+                /**
+                 * Adds <input type="hidden"> before disabled <select>
+                 * with the same, watches <select> value and sets to the <input>
+                 *
+                 * @param {jQueryElement} $select - the selector to be amended
+                 */
+                function supportValueSubmissionOfSelectorIfDisabled ($select) {
```
deb1990: All the function definitions should be after the init function(`document.ready` in this case). Please move the document.ready function to the beginning.

PR #2303, opened 2017-11-16
PCHR-2936: Add the timeUnitApplier smarty modifier

davialexandre:
## Overview

This PR adds something very similar to what was added by https://github.com/civicrm/civihr/pull/2213, but now in the form of a smarty modifier, so that it can be used in Smarty templates in pages that are PHP only.

## Before

It was not possible to apply time units to values in pages using Smarty templates

## After

It's possible to apply time units in Smarty templates. Here are some examples of how this can be done:

```php
<p>You have spent {$timeSpent|timeUnitApplier:'hours'} on this task</p>
<p>Your last login was {$lastLogin|timeUnitApplier:'days'} ago</p>
```

## Comments
Given the Smarty modifer only delegates the job to an internal class, that class can also be used in any PHP unit that requires time units but it's not a Smarty template. To do that, all you need is to call the `CRM_HRCore_String_TimeUnitApplier::apply()` method.

In this commit https://github.com/civicrm/civihr/commit/5a4a4242b9f8ef76c502f9ee4412f9ff9307b90e, I did a very quick refactoring by removing some misleading documentation and unnecessary CiviCRM licensing information.

----
[x] Tests pass


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/String/TimeUnitApplier.php
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Builds the hours part of the time unit. For real numbers, it will consider
+   * only the integer part of it (i.e. for 1.3 it will return 1h)
+   *
+   * @param number $value
+   *
+   * @return string
+   */
+  private static function getHoursPart($value) {
+    if ($value >= 1 - self::$roundMinutes) {
```
tunbola: I think this should be `$value > 1 - self::$roundMinutes`. `TimeUnitApplier::apply(0.75, 'hours')` returns 0h 45m.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/String/TimeUnitApplier.php
```diff
@@ -0,0 +1,125 @@
...
+   * Builds the minutes part of the time unit. Here, $value is expected not to
+   * be the whole value we wan't to apply the unit to, but only the decimal part,
+   * that represents the minutes.
+   *
+   * @param number $value
+   *
+   * @return string
+   */
+  private static function getMinutesPart($value) {
+    if($value && $value <= 1 - self::$roundMinutes && $value >= 0) {
```
tunbola: Do we need `$value >= 0` ? since we are passing in the absolute value to get the remainder with fmod.

tunbola:
In the overview:  s/smart modifier/smarty modifier
In the comments:  s/Do do that/ To do that

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/String/TimeUnitApplier.php
davialexandre: Good catch! I've also fixed this for the JS version

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/String/TimeUnitApplier.php
```diff
@@ -111,7 +115,7 @@ private static function getHoursPart($value) {
    * be the whole value we wan't to apply the unit to, but only the decimal part,
    * that represents the minutes.
    *
-   * @param number $value
+   * @param int $value
```
tunbola: Shouldn't this be int|float also? or probably just float.

davialexandre: Not really, as this is actually the remainder we get here https://github.com/civicrm/civihr/blob/ec929a5576cffaeb92e8dfaf4b3b09bbb6481c46/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/String/TimeUnitApplier.php#L91

PR #2306, opened 2017-11-17
PCHR-2697: Entitlement Calculation page updates to support Leave in Hours

davialexandre:
## Overview
This PR uses the `timeUnitApplier` from https://github.com/civicrm/civihr/pull/2303 in order to consistently display values with time units in the Entitlement Calculation page. 

It also includes some minor text changes to avoid mentioning days where the value can actually also be hours.

## Before

Values were displayed without time units, so it was impossible to know which ones were in days and which ones in hours:

captura de tela de 2017-11-17 13-53-49

Besides that, some fixed texts had specific mentions to the word "days":

captura de tela de 2017-11-17 13-55-11

captura de tela de 2017-11-17 13-57-12

## After

Time units are applied to all values:

captura de tela de 2017-11-17 13-59-01

Texts mentioning days have been updated to either mention "amount" or "hour" (when amount wasn't appropriate):

captura de tela de 2017-11-17 14-00-12
(Note that the time unit was also applied to the proposed entitlement in the calculation details)

The "Days Taken" column has been renamed to "Used in previous period"

captura de tela de 2017-11-17 14-02-39

Also note that the "Prev Yr. Entitlement" and "Prev Yr. Accrued TOIL" have been renamed to "Previous Period Entitlement" and "Previous Period Accrued TOIL". This was done because:
- Periods are not related to years (they can span mutiple years or be less than 12 months)
- The other columns were already using "period" instead of "year", so this makes things more consistent

For overridden entitlements, a h/d suffix is displayed:

peek 2017-11-17 15-21

Note that, while overridding proposed entitlements for hours, the number will be exhibited as its original decimal format (i.e., instead of 217h 30m, the field will show 217.5)

---
[x] Tests pass


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/ContractEntitlementCalculationTest.php
```diff
@@ -275,6 +275,26 @@ public function testGetContractEndDateReturnsTheContractPeriodEndDate() {
     $this->assertEquals($this->contract['period_end_date'], $calculation->getContractEndDate());
   }
 
+  public function testIsCalculationUnitInHoursShouldReturnTrueIfAbsenceTypeCalculationUnitIsInHours() {
+    $absenceType = $this->prophesize(AbsenceType::class);
+    $absenceType->isCalculationUnitInHours()
+                ->shouldBeCalled()
+                ->willReturn(true);
```
tunbola: Since this functionality to mock the absence type class is used like four times already both in this test class and in the EntitlementCalculation test class, I'm wondering if we can wrap it up in a function that accepts a boolean for the return value of isCalculationUnitInHours and put the function in an helper class where it can be called.

davialexandre: I've thought of that at first, but then I thought the amount of duplication is not that big (just 2 lines of code, if you ignore the formatting) and moving it to another method would actually decrease readability, as it your force the dev having to check another method to understand what is going on.

PR #2307, opened 2017-11-20
PCHR-2937: Add new (svg) spinner for civihr extensions

bmnepali:
## Overview
This PR add the new spinner (svg) animated with css in  `org.civicrm.bootstrapcivihr` extension so that it can be reused for better consistency across civihr extensions.

## Before
- Previously the gif is shown as spinner.
2937-before-min

## After
- The gif is replaced by the new spinner.
2937-final-min

## Technical Details
1. Add a spinner svg creating a new folder `img/` (`org.civicrm.bootstrapcivihr/img/spinner.svg`).
```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" id="Layer_1" data-name="Layer 1" viewBox="0 0 80 80">
  <path
      id="civiHRSpinner"
      fill="#0071BD"
      d="M40,72C22.4,72,8,57.6,8,40C8,22.4,
       22.4,8,40,8c17.6,0,32,14.4,32,32c0,1.1-0.9,2-2,2
       s-2-0.9-2-2c0-15.4-12.6-28-28-28S12,24.6,12,40s12.6,
       28,28,28c1.1,0,2,0.9,2,2S41.1,72,40,72z">
   </path>
   <title>
      CiviHRSpinner
   </title>
</svg>
```

2. Create a `mixin` for the spinner in `bootstrapcivihr exension` (`org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss`):
```css
...
@mixin spinner($force-size, $speed: 1.5s, $size: 32px) {
  $size: if($force-size == '', $size, $size !important);
  ...
  height: $size;
  width: $size;
  ...
}
```

3. Override the CiviCRM spinner using the new spinner  in (`org.civicrm.bootstrapcivihr/scss/components/overrides/crm/_spinner.scss`):
```css
/**
 * Overriding the default civicrm core spinner
 * defined in file sites/all/modules/civicrm/css/civicrm.css
 */
.crm-container {
  .dataTables_processing,
  .blockUI.blockOverlay {
    @include spinner('true');
    margin: 11% 50% !important;
  }
}
```

4. Update the loader/spinner directive template in `reqangular extension` (`org.civicrm.reqangular/src/common/templates/loading.html`):
```html
<div>
  <div prevent-animations class="spinner" ng-show="!show"></div>
  <div ng-transclude ng-show="show"></div>
</div>
```
## Comments
As part of this PR, the spinner is applied to all extensions except hrjobcontract and T&A which have their own version of spinner directive. And these two extensions will be using the spinner form bootstrapcivihr extension in respective tasks.

## Tests
✅ Jasmine Tests - passed
✅  CSS distributive files -  created
⏹ Backstop Tests - n/a


org.civicrm.bootstrapcivihr/scss/components/_loader.scss
```diff
@@ -0,0 +1,11 @@
+/**
```
AkA84: this is not a component, and even if it was the name of the file would not reflect the name of the component

You are overriding a core screen style, create a `scss/overrides/crm` folder and place it in a aptly-named partial

org.civicrm.bootstrapcivihr/scss/components/_spinner.scss
```diff
@@ -0,0 +1,3 @@
+.spinner {
+  @include spinner(1s);
```
AkA84: if it's always `1s` both here and in the override, then just put it as a default value in the mixin

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,39 @@
...
+ * $height Width of the spinner
+ */
+@mixin spinner($speed: $speed, $height: $default-height, $width: $default-width) {
+  animation: spin $speed linear infinite;
+  background: url(../img/spinner.svg) no-repeat center center !important;
+  display: block;
+  margin: auto;
+
+  @if $height {
+    height: $height !important;
```
AkA84: use a ternary operator (applies for all the other instances)
```scss
$height: if($height, /*value*/, /*other value*/);
```

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,39 @@
...
+  } @else {
```
AkA84: actually i don't get this if/else at all, you are defining already the default values, so what's the point of this?

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,39 @@
+// Spinner related variables
+$default-height: 32px;
+$default-width: 32px;
+$speed: 1s;
```
AkA84: just define these are the default values, no need for these variables

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,39 @@
+// Spinner related variables
+$default-height: 32px;
+$default-width: 32px;
```
AkA84: why would the width be ever different than the height? the spinner should always be a square

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
bmnepali: Right! thanks for catching it! in fact we won't be needing it!

org.civicrm.bootstrapcivihr/scss/civihr.scss
```diff
@@ -6,6 +6,8 @@
 @import "fonts/*";
 
 @import "overrides/style/*";
+@import "mixins/*";
```
AkA84:
```scss
@import "fonts/*";
@import "mixins/*";

@import "overrides/crm/*";
@import "overrides/style/*";

File: org.civicrm.bootstrapcivihr/scss/components/_spinner.scss [hunk 8]
```diff
@@ -0,0 +1,3 @@
+.spinner {
+  @include spinner(1.5s);
```
AkA84: now why did you change it to 1.5s, when before it was 1s?

if *both* implementations of the spinner requires 1.5s, then just define that as the default value and don't pass anything to the function

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Spinner mixin
+ * $speed Speed to rorate the spinner in seconds
+ * $height Height of the spinner
+ * $height Width of the spinner
+ */
+@mixin spinner($speed: 1s, $height: 32px, $width: 32px) {
+  animation: spin $speed linear infinite;
+  background: url(../img/spinner.svg) no-repeat center center !important;
+  display: block;
+  height: $height !important;
```
AkA84: why `!important`?

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
```diff
@@ -0,0 +1,24 @@
+/**
+ * Spinner mixin
+ * $speed Speed to rorate the spinner in seconds
+ * $height Height of the spinner
+ * $height Width of the spinner
+ */
+@mixin spinner($speed: 1s, $height: 32px, $width: 32px) {
```
AkA84: you are still able to define different width and height, which is something we *do not* want to happen.

org.civicrm.bootstrapcivihr/scss/components/_spinner.scss
bmnepali: > now why did you change it to 1.5s, when before it was 1s?
Previously, the speed was `1s` but i realized the speed should  actually be `1.5s` as per @jon's comment. so i changed it to later.

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
bmnepali: Correct! lets pass a param called `$size: 32px`  whose default size will be `32px`.

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
bmnepali: The core CiviCRM spinner already has defined height and width to be `100%`. So, to override it `!important` is used here.

org.civicrm.bootstrapcivihr/scss/civihr.scss
```diff
@@ -4,8 +4,11 @@
 @import "SCSSROOT/org.civicrm.shoreditch/base/scss/vendor/bootstrap/mixins";
 
 @import "fonts/*";
+@import "mixins/*";
 
+@import "overrides/crm/*";
 @import "overrides/style/*";
+
```
AkA84: don't leave this empty line

org.civicrm.bootstrapcivihr/scss/mixins/_spinner.scss
AkA84: why can't you override it by just using a "heavier" selector? It's not good to have "!important" in a generic mixin just to fix a very specific case

bmnepali: > why can't you override it by just using a "heavier" selector?

Since the styles are added inline and there is this only one element and has no nested elements, don't think we can target styles using any selectors leaving behind the only option to use `!important`.
screen shot 2017-11-20 at 8 37 30 pm

However, I have refactored the mixin avoiding the use of `!important`. And overriding the size of core civicrm spinner  in specific overriding styles.

org.civicrm.bootstrapcivihr/scss/overrides/_variables.scss
```diff
@@ -14,3 +14,4 @@ $steps-height: 24px;
 $steps-compact-height: 28px;
 
 $steps-gutter: ($steps-height / 4);
+$spinner: ($steps-height / 4);
```
AkA84: what is this for, you don't seem to be using it anywhere else and it wasn't here in the previous commits

org.civicrm.bootstrapcivihr/scss/overrides/crm/_spinner.scss
```diff
@@ -0,0 +1,13 @@
+/**
+ * Overriding the default civicrm core spinner
+ * defined in file sites/all/modules/civicrm/css/civicrm.css
+ */
+.crm-container {
+  .dataTables_processing,
+  .blockUI.blockOverlay {
+    @include spinner();
+    height: 32px !important;
```
AkA84: that's not what i was talking about, and it's wrong cause what's the point of providing the `$size` property in the mixin if you then redefine width and height outside of it. Use a more specific/heavier selector so that you don't need to use `!important` in the css

org.civicrm.bootstrapcivihr/scss/overrides/_variables.scss
bmnepali: I am not sure why its changed!  i will fix it. I wont be needing it.

PR #2308, opened 2017-11-20
PCHR-2814: Absence Type not taken into account fix

deb1990:
## Problem 
1. Add two absence periods for 2017 and 2018
2. Add entitlements for your user for 2017 and 2018. To make it easier to spot the issue, make sure the balance is different for two periods - approve one request first
3. Now open New leave request modal - see the balance next to leave type in dropdown, opening and closing balance are coming from current absence period
4. If you select dates from next absence period - balance in dropdown, opening and closing balance do not change

## Before
periodsarenotdif

### Understanding the Problem
This is a regression issue and happened while working on Details tab refactor, previously we had this piece of code https://github.com/civicrm/civihr/blob/212b1d0306f8b943ec5b250ffadde04991ebd424/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js#L732

but the `_loadAbsenceTypes` call was removed and now it is https://github.com/civicrm/civihr/blob/5ad8e01de4e7c86b756e13bb8693b00c93283bb8/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js#L386

Also the _loadAbsenceTypes function is now part of parent(request) controller, so it cannot be called from child(details tab)

### How it is Fixed
1. When changing dates, if period is changed, a event `$rootScope.$broadcast('LeaveRequestPopup::loadAbsenceTypes');` is fired, which is caught by `request.controller`, which then
- Call load all absence types with entitlements `setInitialAbsenceTypes()` for current period.
- Call `setInitialAbsenceTypes()` which sets the first Absence Type selected and sets the  `type_id` of the request.
- Fires `$rootScope.$emit('LeaveRequestPopup::updateBalance')` event to calculate balance.

2. `updateAbsencePeriodDatesTypes()` call the `updateBalance()` only if the period is not changed. Otherwise updation of balance is taken care by Point No 1.
3. In `updateBalance()`, balance is calculated only if `selectedAbsenceType` is present, otherwise the Balance block gets hidden.
4. `canSubmit()` now checks if `vm.absenceTypes` has some values, otherwise it does not let the user save the request.
5. `$rootScope.$broadcast('LeaveRequestPopup::ContactSelectionComplete');` is removed from test cases, as it was unnecessary and the event does not exist anymore.
**Important**
6. Previously the details tab was disabled if `$ctrl.absenceType` were not present. Removed this because, if the user selects a date which has no entitlement, still the user should be able to change to a date which has entitlements.
Instead disabling the Single Day/Multi Day buttons, when user has no entitlements.

## Tests
JS Unit Tests are Passing.


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -608,6 +613,21 @@ define([
...
+     * Initialises watchers
+     */
+    function initWatchers () {
+      $rootScope.$watch(function () {
+        return vm.period;
+      }, function () {
+        _loadAbsenceTypes()
+          .then(function () {
+            setInitialAbsenceTypes();
+            $timeout(function () { $rootScope.$emit('LeaveRequestPopup::updateBalance'); }, 0);
```
igorpavlov: I am not sure that using $timeout for code control is that particular case is a good idea. Why it needs $timeout?

deb1990: I found another bug, after opening the popup, if I directly select a date from next period, the balance calculation was not happening, this was because I changed the flow before, the watcher was getting fired, before the calendar data is loaded. And because of that `from_data_type` was blank, and hence the balance change UI got hidden.

So I realised that the loading of absence period has to be done after the calendar is loaded. So I removed the watcher and added a new event `loadAbsenceTypes`. With this, we dont need the `$timeout`anymore. Updated the description, check for complete details.

deb1990:
This branch is not used anymore. As it was redone by @igorpavlov . Deleting the branch, and closing the PR.

PR #2314, opened 2017-11-21
PCHR-2941: Fix stacked modals on close

igorpavlov:
## Overview

UI Bootstrap Modal decorator that was presented in https://github.com/civicrm/civihr/pull/2152. The problem was that when modals were stacked and then closed in the order different from the order they were opened, the document scroll was not unlocked. This PR fixes this issue.

## Before

Document cannot be scrolled after modal closing.

## After

Document can be scrolled after modal closing.

## Technical Details

The issue happened because each time you open the modal the decorator will remember HTML's and BODY's current Style, set state and then revert to it once the modal is closed:

```js
var lockScrollStyle = 'overflow: hidden; height: 100%; width: 100%;';
var elements = [
  {
    node: $document[0].body,
    originalStyle: $document[0].body.getAttribute('style')
  },
  {
    node: $document[0].getElementsByTagName('html')[0],
    originalStyle: $document[0].getElementsByTagName('html')[0].getAttribute('style')
  }
];

elements.forEach(function (element) {
  element.style = element.node.getAttribute('style');
  element.node.setAttribute('style', element.style + ';' + lockScrollStyle);
});

// ...

  elements.forEach(function (element) {
    element.node.setAttribute('style', element.originalStyle);
  });
});
```
That works when you open and close modals in a queue (if opened, it needs to be closed first). However, if you stack modals and then close the one underneath (which is possible only programmatically), the document will be locked. The explanation:

BODY has a style "color:red;". You open Modal1 and now the style is "color:red;overflow:hidden;", the style that was remembered is 'color:red;". Then you open Modal2 and now the style is "color:red;overflow:hidden;overflow:hidden;" the remembered style is "color:red;overflow:hidden;". If you close Modal1 and then Modal2, then the style will be the remembered style from Modal2 - "color:red;overflow:hidden;".

To prevent that, simply add/remove style stack each time a modal is opened/closed:

```js
var modalInstance;
var lockScrollStyle = ';overflow: hidden; height: 100%; width: 100%';
var elements = [
  $document[0].body,
  $document[0].getElementsByTagName('html')[0]
];

elements.forEach(function (element) {
  element.setAttribute('style',
    (element.getAttribute('style') || '') + lockScrollStyle); // add style stack
});

// ...

  elements.forEach(function (element) {
    element.setAttribute('style',
      (element.getAttribute('style') || '').replace(lockScrollStyle, ''));  // remove style stack
  });
});
```

This allows to disregard the order stacked modal windows are closed.

----

✅ Manual Tests - passed
✅ Jasmine Tests - passed
⚠️ Jasmine Tests were not amended since modal closing testing does not currently work (https://github.com/civicrm/civihr/pull/2152)
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - not needed


org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -17,28 +17,23 @@ define(function () {
      */
     function open () {
       var modalInstance;
-      var lockScrollStyle = 'overflow: hidden; height: 100%; width: 100%;';
+      var lockScrollStyle = ';overflow: hidden; height: 100%; width: 100%';
```
bmnepali: Why `;` is placed in front  as in  ";overflow: hidden; height: 100%; width: 100%' ?

igorpavlov: Good question! This func populates style stack to an existing style stack in the `style` attribute of an HTML node (in our case, `body` and `html`). Style attribute can have the following value cases:

1) *no style attribute*
2) style=""
3) style="width: 50%"
4) style="width: 50%;"

Lets say we populate "width: 100%;" style. After style population we will have:

1) style="width:100%;" - good
2) style="width:100%;" - good
3) style="width: 50%width: 100%" - fail
4) style="width: 50%;width: 100%" - good

Now if we populate ";width: 100%" style. After style population we will have:

1) style=";width:100%" - not perfect, but works
2) style=";width:100%" - not perfect, but works
3) style="width: 50%;width: 100%" - good
4) style="width: 50%;;width: 100%" - not perfect, but works

bmnepali: @igor Thanks! i was worried about the case:

> 4) style="width: 50%;;width: 100%"

PR #2317, opened 2017-11-21
PCHR-2839: Sync Scheduled Job State

mickadoo:
## Overview

The scheduled jobs for some sites may not match the [expected state](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/100532387/Scheduled+jobs). This PR adds an upgrader to sync all scheduled jobs to match the expected state.

## Before

Each site could have had a different configuration for scheduled jobs.

## After

Scheduled jobs for each site match the configuration from the Wiki. Any scheduled jobs not listed in the Wiki will be disabled.


mickadoo:
@jenkins-compucorp test this please

PR #2321, opened 2017-11-22
PCHR-2957: Use option group names in Job Contract dropdowns links

davialexandre:
## Overview
The Job Contracts extension adds some items to the Administer -> Customize Data and Screens -> Dropdowns menu, with links to pages where the users can add/update options to the extension's option groups. The URLs of these links contains the option group ID.

In some very unusual setups, the ID of the option group in the database might be different than the menu item link (especially when data is copied from one site to another), making it impossible for users to add or update the options.

## Before

The links to the Job Contract dropdowns followed this structure: `civicrm/admin/options?gid=<ID>` where ID was the option group ID

## After

The links to the Job Contract dropdowns now have the option group name instead of their ID. They now follow this structure: `civicrm/admin/options/<option_group_name>`

## Comments

As part of this PR, we're also deleting the menu item linking to the  "Pension type" option group. The reason is that this option group doesn't exist anymore. It was originally  deleted by https://github.com/civicrm/civihr/pull/1557, but the menu item was left behind.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -512,7 +512,6 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
     $this->upgrade_1030();
     $this->upgrade_1032();
     $this->upgrade_1033();
-    $this->upgrade_1034();
   }
 
   function upgrade_1001() {
@@ -1146,6 +1145,19 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
   }
 
   /**
+   * Removes menu link pointing to an nonexistent option group and update
+   * dropdown links to use option group names instead of IDs.
+   *
+   * @return bool
+   */
+  public function upgrade_1033() {
+    $this->deletePensionTypeDropdownMenu();
+    $this->updateDropdownMenuItemsLinkToUseOptionGroupName();
+
+    return TRUE;
+  }
+
+  /**
    * Removes the "Pension Type" item from the
    *  "Administer -> Customize Data and Screens -> Dropdowns" menu
    *
@@ -1154,16 +1166,14 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
    *
    * @return bool
    */
-  public function upgrade_1033() {
+  private function deletePensionTypeDropdownMenu() {
     civicrm_api3('Navigation', 'get', [
-      'sequential' => 1,
       'name' => 'hrjc_pension_type',
       'url' => ['LIKE' => 'civicrm/admin/options?gid%'],
       'api.Navigation.delete' => ['id' => '$value.id'],
     ]);
-    CRM_Core_BAO_Navigation::resetNavigation();
 
-    return TRUE;
+    CRM_Core_BAO_Navigation::resetNavigation();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mickadoo:
In description `s/users and add/users can add/`

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -511,6 +511,8 @@ public function upgradeBundle() {
     $this->upgrade_1029();
     $this->upgrade_1030();
     $this->upgrade_1032();
+    $this->upgrade_1033();
```
mickadoo: Do you think we could use traits like in other extensions? I know we already have function here, but going forward it might be nicer to use traits for all of them. 

It probably doesn't make sense to have both traits and keep existing functions in the upgrader class, so might be better to make a ticket to move all upgraders to using traits instead of half-doing it in this ticket.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -511,6 +511,8 @@ public function upgradeBundle() {
...
+    $this->upgrade_1034();
```
mickadoo: I'm not sure this should be two separate upgrades. The `upgrade_XXX` function adds no meaning when split into two and no clients will ever have `upgrade_1033` applied without `upgrade_1034`. I think it would be better to keep just `upgrade_1033` and split the logic for removing the navigation item and updating the URLs into functions with descriptive names.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1143,6 +1145,64 @@ public function upgrade_1032() {
...
+   *  "Administer -> Customize Data and Screens -> Dropdowns" menu
+   *
+   * The option group this menu item links to has been removed by PCHR-1820,
+   * but the menu item itself wasn't, so we're deleting it now.
+   *
+   * @return bool
+   */
+  public function upgrade_1033() {
+    civicrm_api3('Navigation', 'get', [
+      'sequential' => 1,
```
mickadoo: Minor, but sequential isn't required.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1143,6 +1145,64 @@ public function upgrade_1032() {
...
+      'hrjc_pay_cycle',
+      'hrjc_benefit_name',
+      'hrjc_benefit_type',
+      'hrjc_deduction_name',
+      'hrjc_deduction_type',
+      'hrjc_revision_change_reason',
+      'hrjc_contract_end_reason',
+    ];
+
+    foreach($dropdownMenuItems as $menuItem) {
```
mickadoo: space in `foreach (`

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1143,6 +1145,64 @@ public function upgrade_1032() {
...
+      civicrm_api3('Navigation', 'get', [
+        'sequential' => 1,
```
mickadoo: Sequential not required

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1143,6 +1145,64 @@ public function upgrade_1032() {
...
+        'name' => $menuItem,
+        'url' => ['LIKE' => 'civicrm/admin/options?gid%'],
+        'api.Navigation.create' => [
+          'id' => '$value.id',
+          'url' => "civicrm/admin/options/{$menuItem}?reset=1"
```
mickadoo: Personal preference, but I like `sprintf` more because of its type safety and that it avoids potentially evaluating something you don't want to.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
davialexandre: It's possible to use them here, but that would require transforming all the existing upgraders to traits. This extension is already a mess and using traits only for new upgraders would only make it messier. 

Given there are plans to refactor the Job Contract extension in the future, it better to wait for that until we start using traits here

PR #2330, opened 2017-11-24
PCHR-2963: Fix date time format in staff/admin/manager leave reports

igorpavlov:
## Overview

Both Staff and Manager reports have issues with date/time formatting:

1. In Manager report the CRM date format is ignored, single dates show 2 dates instead of 1, TOIL in hours shows time while it should not.
2. In Staff report all dates are just shown in the server format.

This PR fixes these issues in Staff and Manager reports ( for both Admin and SSP).

## After
All dates/times are consistent and follow correct logic.

## Technical Details

1. Utilised the already written date filter, however, it lacked "hours" unit support, which was added.

2. This date filter was added in all modules.

3. HTML was amended to utilise the filter and the new logic.

4. Date format (the actual format, ex. 'YYYY/DD/MM') was imported from CRM into all L&A modules.

## Comments

Absence Tab does not have routing, so resolving the promise that set the date format from CRM performed in the Absence Tab parent component (Absence Tab Container).

----

✅ Manual Tests - passed
⏹ Backstop Tests - omitted
⏹ PHP Unit Tests - not needed


igorpavlov:
@reneolivo I am going to split the date-format.js service method into 2. Currently the function is used in the following files:

image

That all will be refactored.

igorpavlov:
It was suggested by @reneolivo to refactor the date-format.js service. However, after an attempt to refactor I discovered failing tests in Appraisals extension and find refactoring too risky at the moment since the release is coming up. However, here are files that were refactored and the code might be used for the future refactoring:

*REGANGULAR/src/common/services/angular-date/date-format.js*
```js
/* eslint-env amd */

define([
  'common/angular',
  'common/modules/angular-date',
  'common/services/hr-settings'
], function (angular, module) {
  'use strict';

  /**
   * This service provides access to date format from CRM settings
   */
  module.factory('DateFormat', ['$q', 'HR_settings', function ($q, HRSettings) {
    return {
      /**
       * Keeps information about date format
       */
      dateFormat: null,
      /**
       * Fetches date format setting
       * @TODO the format setting should be fetched from CRM settings
       *   and not just return a constant
       *
       * @returns {Promise}
       */
      getDateFormat: function () {
        if (this.dateFormat) {
          return $q.when(this.dateFormat);
        } else {
          return $q.when('dd/MM/yyyy')
            .then(function (result) {
              this.dateFormat = result;

              return result;
            }.bind(this));
        }
      },

      /**
       * Stores date format in HR settings
       *
       * @returns {Promise}
       */
      setDateFormatToHRSettings: function () {
        return this.getDateFormat().then(function (dateFormat) {
          HRSettings.DATE_FORMAT = dateFormat;
        });
      }
    }
  }]);
});
```

*Any routing config*
```js
$routeProvider
  .resolveForAll({
    format: ['DateFormat', function (DateFormat) {
      var dateFormat;

      return DateFormat.getDateFormat()
        .then(function (result) {
          dateFormat = result;

          return DateFormat.setDateFormatToHRSettings();
        })
        .then(function () {
          return dateFormat;
        });
    }]
  })
```

PR #2333, opened 2017-11-27
PCHR-2938: Update spinner in jobcontract extension

bmnepali:
## Overview
This PR updates the spinner in the jobcontract extension by reusing the spinner defined in bootstrapcivihr extensioin.

#### Displays old spinner in hrjobcontract extension:
2938-before

#### Display new spinner in hrjobcontract extension:
2938-final

#### Display contracts after the contract is fully loaded with revisions:
final-2938

## Technical Details
1. Refactor and keep only necessary custom spinner styles specific to hrjobcontract extension in file scss/hrjc/partials/_loader.scss.
```css
.#{$prefix}loader {
  left: 50%;
  position: absolute;
  top: 50%;
  z-index: 1000;
}
```
Here the `z-index` is set to `1000` because any value above it will keep the spinner above navigation bar.

2. Reuse the `.spinner` class defined in `bootstrapcivihr` extension in the spinner directive (js/src/job-contract/directives/loader.js).
```js
loader.className = 'hrjc-loader spinner';
```
3. Mark the contract fully loaded after its revisions are also loaded in file  js/src/job-contract/controllers/contract.js.
```js
ContractService
   .fullDetails(contractId)
   .then(function (results) {
       ...
       ContractRevisionList.fetchRevisions(contractId).then(function (result) {
          ...
          $scope.contractLoaded = true;
        });
     })
```
## Tests
✅ Jasmine Tests - passed
✅ CSS distributive files - not needed to run
⏹ Backstop Tests - n/a


hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -1,56 +1,57 @@
...
-                    loaderSet = false,
-                    positionSet = false;
+  directives.directive('hrjcLoader', ['$rootScope', '$log', function ($rootScope, $log) {
+    $log.debug('Directive: hrjcLoader');
 
-                loader.className = 'hrjc-loader';
+    return {
+      link: function ($scope, el, attrs) {
+        var loader = document.createElement('div');
+        var loaderSet = false;
```
igorpavlov: should be `isLoaderSet`, because otherwise it can be read as `a loader set` or `a set of loaders` where "set" is a noun, not a verb. Same for `positionSet`.

hrjobcontract/css/hrjc.css
```diff
@@ -11572,29 +11572,10 @@ fieldset[disabled]
...
-  background: rgba(255, 255, 255, 0.5) url("../img/loading-overlay.gif") no-repeat center center;
-  -moz-border-radius: 0;
-  -webkit-border-radius: 0;
-  border-radius: 0;
   position: absolute;
-  width: 100%;
-  height: 100%;
-  top: 0;
-  left: 0;
+  top: 50%;
```
igorpavlov: Not quite sure this centers the loader perfectly. Could you check please?

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -1,56 +1,57 @@
...
-                }
+        loader.className = 'hrjc-loader spinner';
 
-                function appendLoader() {
-                    if (!isPositioned()) {
-                        el.css('position','relative');
-                        positionSet = true;
-                    }
+        function isPositioned () {
+          var elPosition = window.getComputedStyle(el[0]).position;
```
igorpavlov: `elPosition` sounds like the code is written in another language. I think it is better to use `elementPosition`. Same applies to everywhere else `el` is used.

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -1,56 +1,57 @@
...
+          return elPosition === 'relative' || elPosition === 'absolute' || elPosition === 'fixed';
+        }
 
-                    el.append(loader);
-                    loaderSet = true;
-                }
+        function appendLoader () {
+          if (!isPositioned()) {
+            el.css('position', 'relative');
+            positionSet = true;
```
igorpavlov: Needs a line break. Please check other places as well.

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -1,56 +1,57 @@
...
-                    return elPosition == 'relative' || elPosition == 'absolute' || elPosition == 'fixed'
-                }
+        loader.className = 'hrjc-loader spinner';
 
-                function appendLoader() {
-                    if (!isPositioned()) {
-                        el.css('position','relative');
-                        positionSet = true;
-                    }
+        function isPositioned () {
```
igorpavlov: No func description, please check all funcs for description.

hrjobcontract/scss/hrjc/partials/_loader.scss
```diff
@@ -1 +1,6 @@
...
   top: 10px;

   &.ng-hide {
      display: none!important;
   }

}
\ No newline at end of file
+.#{$prefix}loader {
+  position: absolute;
```
igorpavlov: Please sort in alphabetic order.

org.civicrm.bootstrapcivihr/gulpfile.js
```diff
@@ -22,7 +22,7 @@ gulp.task('sass', ['sass:sync'], function () {
     .pipe(postcss([postcssPrefix({
       prefix: bootstrapNamespace + ' ',
       exclude: [
-        /^html/, /^body/, /page-civi/, /crm-container/, outsideNamespaceRegExp
+        /^html/, /^body/, /page-civi/, /crm-container/, /spinner/, outsideNamespaceRegExp
```
igorpavlov: Is that what we usually do? Last time there was a discussion that this should not be like that )or maybe I am missing some point). cc @AkA84

hrjobcontract/js/src/job-contract/directives/loader.js
bmnepali: Added!

hrjobcontract/css/hrjc.css
bmnepali: Well, tours out `top: 50%; left:50%` is not enough for positioned element to keep in center. So, I have created a fiddle http://jsfiddle.net/6xo11zwv/266/ with `transform: translate(-50%, -50%);` as well. I have checked with different browsers and it works. I have update the the style.

org.civicrm.bootstrapcivihr/gulpfile.js
bmnepali: @igorpavlov 
Actually, previously something like following was used:
```
[ /^html/, /^body/, /page-civi/, ".crm-container .other-class", ".crm-container .one-other-class"]
```
And after our discussion we decided to use following instead considering it as better approach:
```
[/^html/, /^body/, /page-civi/, /crm-container/]
```
So, this is the approach we will be following.

cc @AkA84

hrjobcontract/css/hrjc.css
```diff
@@ -11572,9 +11572,10 @@ fieldset[disabled]
 }
 /* line 1, ../scss/hrjc/partials/_loader.scss */
 #hrjob-contract .hrjc-loader {
+  left: 50%;
   position: absolute;
   top: 50%;
-  left: 50%;
+  transform: translate(-50%, -50%);
```
igorpavlov: Are you sure you tested in IE11? I think IE11 requires `-ms-` prefix. If yes, then I think this deserves a separate mixin (of course if this mixin does not exists yet, please make sure you check that).

igorpavlov: Also, if you have a variable for loader width, height, you might use a safer method:

`margin-left: -$width/2;` or smth like that.

bmnepali: > Are you sure you tested in IE11? I think IE11 requires -ms- prefix. If yes, then I think this deserves a separate mixin (of course if this mixin does not exists yet, please make sure you check that).

Checked with IE11 using browerstack and it works!

>  if you have a variable for loader width, height

The height and width of the spinner is encapsulated in spinner mixin in bootstrapcivihr extension so, we don't have access to them in this `hrjobcontract` extension. Here we are only reusing the `.spinner` class.

bmnepali: > Are you sure you tested in IE11? I think IE11 requires -ms- prefix. If yes, then I think this deserves a separate mixin (of course if this mixin does not exists yet, please make sure you check that).

Checked with IE11 using browerstack and it works!

>  if you have a variable for loader width, height

The height and width of the spinner is encapsulated in spinner mixin in bootstrapcivihr extension so, we don't have access to them in this `hrjobcontract` extension. Here we are only reusing the `.spinner` class.

igorpavlov: Yep, it is supported https://caniuse.com/#search=translate

org.civicrm.bootstrapcivihr/gulpfile.js
AkA84: why is this even necessary?

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -9,48 +9,71 @@ define([
...
-          loaderSet = true;
+          element.append(loader);
+          isLoaderSet = true;
         }
 
-        function removeLoader () {
-          loaderSet && loader.parentNode.removeChild(loader);
-          loaderSet = false;
+        /**
+         * Returns if the element has position value set
```
igorpavlov: `Checks if...` otherwise it sounds like it returns something if only the position value is set and does not return if the position is not set.

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -9,48 +9,71 @@ define([
...
-          if (positionSet) {
-            el.css('position', '');
-          }
+          return elementPosition === 'relative' || elementPosition === 'absolute' || elementPosition === 'fixed';
         }
 
-        if (attrs.hrjcLoaderShow) {
-          appendLoader();
+        /**
+         * Init Listeners
```
igorpavlov: Initialises listeners

hrjobcontract/js/src/job-contract/directives/loader.js
```diff
@@ -9,48 +9,71 @@ define([
...
+         */
+        function initLIsteners () {
+          $scope.$on('hrjc-loader-show', function () {
+            appendLoader();
+          });
+
```
igorpavlov: No line break needed.

bmnepali:
@AkA84 
> Why don't we use the common loader directive that we have in reqangular?

It's not that we can't use the loader defined in reqangular but its different from the loader directive in T&A and hrjobcontract extension and the way they are used. Which will need some more time to refactor it. So, I have updated the new spinner amending the existing loader directives in respective extensions. Which should be done in separate ticket I guess.

> In the "after" gif i see that at some point the loader appears on top of the content, isn't that a bug?

Actually, the spinner in the before section also appears on top but the spinner is in the transparent container. So, its not a bug. But we need to decide if we should keep the transparent background after testing is done.

org.civicrm.bootstrapcivihr/gulpfile.js
bmnepali: @AkA84 It's necessary as the spinner is used outside the `bootstrap-theme` namespace scope in hrjobcontract and if the `spinner` styles are namespaced within the `bootstrap-theme` in bootstrapcivihr the styles will not be available in hrjobcontract.

AkA84: @bmnepali please check what `outsideNamespaceRegExp` does

AkA84:
> Actually, the spinner in the before section also appears on top but the spinner is in the transparent container. So, its not a bug. But we need to decide if we should keep the transparent background after based on the testing results.

honestly that looks wrong, why should i see a spinner floating on an already-loaded content? That should not happen

org.civicrm.bootstrapcivihr/gulpfile.js
bmnepali: @AkA84 thank!

hrjobcontract/js/src/job-contract/directives/loader.js
bmnepali: done!

hrjobcontract/js/src/job-contract/directives/loader.js
bmnepali: done!

bmnepali:
@AkA84  Fixed the spinner overlapping issue on contract. Updated in PR description.

bmnepali:
@igorpavlov merging it into its parent branch as approval from @AkA84.

PR #2343, opened 2017-11-29
PCHR-2998: Add Upgrader to set Emergency Contact Required Fields

mickadoo:
## Overview

Emergency contacts have no required fields. With the new separation of "Emergency Contacts" and "Dependants" it is important that "Is a Dependant" is always set. In other places we require it (but hide the field) as well as name, mobile number and relationship. These should be required everywhere.

## Before

Emergency contact custom fields had no required fields.

## After

Emergency contacts have the following required fields:

- Name
- Mobile Number
- Relationship with Employee
- Is a Dependant?

### DIFF
--- a/org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1003.php
+++ b/org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1003.php
@@ -4,6 +4,8 @@ trait CRM_Hremergency_Upgrader_Steps_1003 {
 
   /**
    * Makes some emergency contact fields required.
+   *
+   * @return TRUE
    */
   public function upgrade_1003() {
     $params = ['custom_group_id' => 'Emergency_Contacts'];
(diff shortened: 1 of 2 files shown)


org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1003.php
```diff
@@ -0,0 +1,31 @@
...
+
+    $shouldBeRequired = [
+      'Name',
+      'Mobile_number',
+      'Relationship_with_Employee',
+      'Dependant_s_',
+    ];
+
+    foreach ($customFields as $customField) {
+      if (!in_array($customField['name'], $shouldBeRequired)) {
```
tunbola: Why not 

> if (in_array($customField['name'], $shouldBeRequired)) {
     $customField['is_required'] = TRUE;
     civicrm_api3('CustomField', 'create', $customField);
}

return TRUE;

Then the code will be a bit more conpact and we would save one or two lines.

org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1003.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+trait CRM_Hremergency_Upgrader_Steps_1003 {
+
+  /**
+   * Makes some emergency contact fields required.
```
tunbola: Please indicate that this function returns a bool.

org.civicrm.hremergency/CRM/Hremergency/Upgrader/Steps/1003.php
mickadoo: Fair enough, usually when I start writing code I like to have early returns because the alternative if a possibility of nesting more and more `if` statements, but in this case it won't get more complicated so having the api call in the if block is just as neat.

PR #2341, opened 2017-11-29
PCHR-2989: Update ContactSummary CSS

reneolivo:
## Overview
This PR updates the contact summary CSS so it includes the new colours added in Shoreditch.

## Before
civihr_staff compucorp co uk hr17 9000 6

## After
civihr_staff compucorp co uk hr17 9000 7

## Technical Details
Since the Leave Widget referenced the colours defined in Shoreditch and they were updated, the Contact Summary needed to be updated as well by running `gulp sass`.

## Comments
Ran backstop-config (the repo) on contact page and the differences are in colour only (as expected).

PR #2347, opened 2017-12-04
Revert "PCHR-2989: Update ContactSummary CSS"

reneolivo:
Reverts civicrm/civihr#2341


reneolivo:
Force push will be used instead.

PR #2348, opened 2017-12-04
PCHR-2906: Create Public Holiday Requests in the past when MTPHL is set to yes

tunbola:
## Overview
If a public holiday is added in the past or future when there is no absence type with MTPHL = Yes, public holiday leave requests will not be created for any of the the public holiday dates. However when MTPHL is now set = Yes for an absence type, only public holiday leave requests in the future get created. This PR modifies this logic when the MTPHL property of an absence type is toggled, in order to allow leave requests for past holiday dates to also be created.

## Before
Public Holiday leave requests for past public holiday dates(that does not have leave requests created for them for all qualified contacts) are not created when an Absence Type MTPHL is set to Yes.

## After
Public Holiday leave requests for past public holiday dates(that does not have leave requests created for them for all qualified contacts) are now created when an Absence Type MTPHL is set to Yes.

## Technical Details
- A new `createForAll` method was added in the `PublicHolidayLeaveRequestCreation` Service that creates leave requests for all public holidays in all absence periods for all contacts. It only creates for leave requests that does not already exist on that date for a contact.
- Added a new `updateAllExceptThoseAlreadyCreatedInThePast` method in the `PublicHoliday` Service that updates/creates public holiday leave requests in past and future but leaves past public holiday leave requests already created untouched.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -566,7 +566,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    */
   private static function enqueuePublicHolidayLeaveRequestUpdateTask() {
     $task = new CRM_Queue_Task(
-      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllFuturePublicHolidayLeaveRequests', 'run'],
+      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllPublicHolidayLeaveRequestsExceptThoseAlreadyCreatedInThePast', 'run'],
       []
     );
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -566,7 +566,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
    */
   private static function enqueuePublicHolidayLeaveRequestUpdateTask() {
     $task = new CRM_Queue_Task(
-      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllPublicHolidayLeaveRequestsExceptThoseAlreadyCreatedInThePast', 'run'],
+      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllPublicHolidayLeaveRequests', 'run'],
       []
     );
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequest.php
@@ -34,7 +34,7 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequest {
    */
   public function updateAllInTheFuture() {
     $this->deletionLogic->deleteAllInTheFuture();
-    $this->creationLogic->createForAllInTheFuture();
+    $this->creationLogic->createAllInTheFuture();
   }
 
   /**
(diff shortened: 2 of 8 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -1246,6 +1246,76 @@ public function testWillNotCreatePublicHolidayLeaveRequestsInTheFutureForWorkPat
...
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact1['id'], $publicHoliday3));
+
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday1));
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday2));
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday3));
+
+    //The public holiday4 des not exist in any absence period, so will not be created for either contact
+    $this->assertNull(LeaveRequest::findPublicHolidayLeaveRequest($contact1['id'], $publicHoliday4));
+    $this->assertNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday4));
+  }
```
davialexandre: We already have a test checking that we don't create a new request if one already exists (`testItDoesNotCreateALeaveRequestIfTheresIsAlreadyALeaveRequestForTheGivenPublicHolidayAndContact`), so let's make this test only about checking that it create requests for public holidays in the past and in the future

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreation.php
```diff
@@ -359,6 +359,40 @@ public function createAllForAbsencePeriod($absencePeriod, array $contactID = [])
...
+   * absence periods for all contacts. It does not re-create a leave request
+   * that already exists.
+   *
+   * The Public Holiday leave request will not be created if a contact has no
+   * entitlement for the absence type with MTPHL in the absence period the
+   * public holiday date falls in or If the contact does not have a contract
+   * overlapping the public holiday date.
+   *
+   */
+  public function createForAll() {
```
davialexandre: I  was checking the methods names in this class we have have two patterns: those starting with `createAllFor` and those with `createForAll`. The former, creates requests for all Public Holidays for a specific entity (Absence Periods, Contacts, Contracts, WP). The latter, creates requests for a single Public Holiday but for all the records of specific entity set (for all Absence Periods, for all contracts, etc).

Given that, I think we should rename this method to `createAll()`. Also, to make this consistent, I think we should rename `createForAllInTheFuture` to `createAllInTheFuture` (Also, note that the respective method in the Deletion logic class is named `deleteAllInTheFuture` instead of `deleteForAllInTheFuture`)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequest.php
```diff
@@ -37,6 +37,28 @@ public function updateAllInTheFuture() {
...
+   * created for all qualified contacts.
+   *
+   * Basically, this uses the Deletion Logic to delete all the leave requests
+   * for public holidays in the future and then uses the Creation Logic to
+   * create leave requests for all public holidays.
+   *
+   * @see CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion::deleteAllInTheFuture()
+   * @see CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestCreation::createForAll()
+   */
+  public function updateAllExceptThoseAlreadyCreatedInThePast() {
```
davialexandre: All the methods will ignore the requests already created (because they all use the `createForContact` method in the end), but none of them have this explicit in their names, so I don't see why we should do this in this case. I think that `updateAll()` is enough

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -566,7 +566,7 @@ private static function mustUpdatePublicHolidayLeaveRequests($params) {
    */
   private static function enqueuePublicHolidayLeaveRequestUpdateTask() {
     $task = new CRM_Queue_Task(
-      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllFuturePublicHolidayLeaveRequests', 'run'],
+      ['CRM_HRLeaveAndAbsences_Queue_Task_UpdateAllPublicHolidayLeaveRequestsExceptThoseAlreadyCreatedInThePast', 'run'],
```
davialexandre: I don't think we should be explicit about the name here either. A simple 'UpdateAllPublicHolidayLeaveRequests' should be enough

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -1246,6 +1246,69 @@ public function testWillNotCreatePublicHolidayLeaveRequestsInTheFutureForWorkPat
     $this->assertEquals(0, LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement2));
   }
 
+  public function testCreateForAllWillCreatePublicHolidayLeaveRequestsForAllPublicHolidaysInAllPeriodsForAllContacts() {
```
davialexandre: Please update the test name to match the new method name

PR #2351, opened 2017-12-05
PCHR-3019: Bump version to 1.7.2

davialexandre:


davialexandre:
We need it urgently, so I'm merging it to staging even though it wasn't tested by Jenkins. Since this is only changing the info.xml files, it should be fine to skip testing this time

PR #2356, opened 2017-12-07
PCHR-3026: Classes to Represent Pingback Statistics

mickadoo:
## Overview

To make it easier to deal with the statistics that we will send, and avoid using large, complicated arrays these classes will represent all data to be send in the statistics request. 

The goal is to produce a JSON string with [this format](https://gist.github.com/mickadoo/3eb09bafa18760c12fcccced7a1afc33).

## Before

There were no classes to use when gathering and sending statistics.

## After

There are classes to represent all the data being sent. Another class is responsible for producing a JSON string to match the [expected specification](https://gist.github.com/mickadoo/3eb09bafa18760c12fcccced7a1afc33).

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/CiviHRStatisticsJSONConvertor.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/CiviHRStatisticsJSONConvertor.php
@@ -1,0 +1,119 @@
+<?php
+
+use CRM_HRCore_Model_CiviHRStatistics as CiviHRStatistics;
+use CRM_HRCore_Model_ReportConfiguration as ReportConfig;
+use CRM_HRCore_Model_ReportConfigurationAgeGroup as AgeGroup;
+
+/**
+ * Responsible for converting a CiviHRStatistics object to JSON
+ */
+class CRM_HRCore_Service_CiviHRStatisticsJSONConvertor {
+
+  /**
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/CiviHRStatisticsJSONConvertor.php
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Takes a statistics class and converts it to a JSON string.
+   *
+   * @param CiviHRStatistics $stats
+   *
+   * @return string
+   */
+  public static function toJson(CiviHRStatistics $stats) {
+    $array = [
```
MiyaNoctem: Could '$array' have a more meaningful name?

MiyaNoctem:
I can see a clear intent to use specific getters and setters for all attributes of the classes. Do you think it might be worth to future proof this so all attributes of the classes have to be specifically declared to be used, using __get and __set magic functions? Something like:

```
public function __get($attribute){
  throw new Exception("Variable '$attribute' has note been declared");
}
```

mickadoo:
@MiyaNoctem you mean to prevent things like

`$x = $stats->someOtherProperty`

I don't think it's so likely to happen, plus I would hope that code that uses undeclared class properties wouldn't pass code review.

PR #2360, opened 2017-12-14
PCHR-3056: Use Closest Absence Period to Current Date For Manage Entitlement Calculation

tunbola:
## Overview
When there is no absence period with dates containing the current date and a user tries to update entitlements after creating/updating a contract, an Error 500 page appears. This PR fixes the issue by redirecting to Manage entitlements calculation page to calculate entitlements for the closest absence period to the current date after creating/updating a contract. 

## Before
- When a user tries to update entitlements after creating/updating a contract, entitlements is calculated for the absence period with dates containing the current date.

## After
- Rather than calculate entitlements for the absence period explicitly with dates containing the current date when a user tries to update entitlements after creating/updating a contract, entitlements is calculated for the closest absence period to the current date.

## Comments
- FE will use the existing `AbsencePeriod.get()` API to display a text on the Leave tab of the Job contract modal incase there is no absence period.
If the count property for the API response is Zero, that means there is no absence period.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -4,6 +4,7 @@ use CRM_HRCore_Date_BasicDatePeriod as BasicDatePeriod;
 use CRM_HRLeaveAndAbsences_Validator_Date as DateValidator;
 use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 use CRM_HRLeaveAndAbsences_BAO_PublicHoliday as PublicHoliday;
+use CRM_HRLeaveAndAbsences_BAO_AbsencePeriod as AbsencePeriod;
 use CRM_HRLeaveAndAbsences_Exception_InvalidAbsencePeriodException as InvalidAbsencePeriodException;
 
 class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DAO_AbsencePeriod {
@@ -617,5 +618,84 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
     }
 
     return $absencePeriods;
+  }
+
+  /**
+   * Returns the closest Absence Period to the current date.
+   * The absence period could be before the current date or
+   * could be after or could even contain the current date.
+   *
+   * @return AbsencePeriod|null
+   *   Returns null if there are no absence periods.
+   */
+  public static function getClosestToCurrentDate() {
+    $firstPeriod = self::getFirst();
+
+    if(!$firstPeriod) {
+      return null;
+    }
+
+    $firstPeriodEndDate = new DateTime($firstPeriod->end_date);
+    $today = new DateTime('today');
+
+    if($firstPeriodEndDate < $today) {
+      $absencePeriods = self::getPeriodsBetweenDates($firstPeriodEndDate, $today);
+      $periodBeforeOrInCurrentDate = end($absencePeriods);
+      $nextPeriod = $periodBeforeOrInCurrentDate->getNextPeriod();
+      return self::getClosetPeriodToCurrentDate($periodBeforeOrInCurrentDate, $nextPeriod);
+    }
+
+
+    return $firstPeriod;
+  }
+
+  /**
+   * Returns the first Absence Period record.
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod|null
+   */
+  private static function getFirst() {
+    $absencePeriod = new self();
+    $absencePeriod->orderBy('weight');
+    $absencePeriod->find(true);
+
+    if($absencePeriod->id) {
+      return $absencePeriod;
+    }
+
+    return null;
+  }
+
+  /**
+   * Returns the closest absence period to the current date i.e
+   * either of firstPeriod or secondPeriod.
+   * The days interval between today and the end date of the first
+   * period is compared with the days interval between today and the
+   * start date of the second period, the period with which the current
+   * date has the least days interval is returned as the closest period
+   * to the current date.
+   * If the secondPeriod is null, the firstPeriod is simply returned as
+   * the closest period.
+   *
+   * @param AbsencePeriod $firstPeriod
+   * @param AbsencePeriod|null $secondPeriod
+   *
+   * @return AbsencePeriod
+   */
+  private static function getClosetPeriodToCurrentDate(AbsencePeriod $firstPeriod, AbsencePeriod $secondPeriod = null) {
+    $currentDate = new DateTime('today');
+
+    if($secondPeriod) {
+      $firstPeriodEndDate = new DateTime($firstPeriod->end_date);
+      $secondPeriodStartDate = new DateTime($secondPeriod->start_date);
+      $daysDiffForFirstPeriod = $currentDate->diff($firstPeriodEndDate)->days;
+      $daysDiffForSecondPeriod = $currentDate->diff($secondPeriodStartDate)->days;
+
+      if($daysDiffForSecondPeriod < $daysDiffForFirstPeriod){
+        return $secondPeriod;
+      }
+    }
+
+    return $firstPeriod;
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
@@ -631,7 +631,7 @@ class CRM_HRLeaveAndAbsences_BAO_AbsencePeriod extends CRM_HRLeaveAndAbsences_DA
    * period in the future is preferred, If that does not exists, then
    * the closest absence period in the past is considered.
    *
-   * @return AbsencePeriod|null
+   * @return \CRM_HRLeaveAndAbsences_BAO_AbsencePeriod|null
    *   Returns null if there are no absence periods.
    */
   public static function getClosestToCurrentDate() {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
```diff
@@ -142,8 +142,12 @@ public function postProcess() {
   private function getAbsencePeriodFromRequest() {
     $periodId = CRM_Utils_Request::retrieve('id', 'Integer');
     if(!$periodId) {
-      return AbsencePeriod::getCurrentPeriod();
+      $today = new DateTime('today');
+      $absencePeriods = AbsencePeriod::getPeriodsBetweenDates($today);
```
mickadoo: From the name I would have expected `getCurrentPeriod` to return the same as this, and `getPeriodsBetweenDates` doesn't make much sense reading it when only one date is passed in. Not really relevant to this PR, but anyway.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsencePeriod.php
```diff
@@ -44,3 +44,27 @@ function civicrm_api3_absence_period_delete($params) {
...
+ * A valid absence period exists if a period exists that
+ * has its end date greater than or equal to today.
+ *
+ * @param array $params
+ *
+ * @return array API result descriptor
+ *
+ * @throws API_Exception
+ */
+function civicrm_api3_absence_period_validPeriodExists() {
```
mickadoo: Was this not possible with the existing AbsencePeriod API? For me locally I can do a request for `AbsencePeriod.get` with `start_date` > today`?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/AbsencePeriodTest.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @group headless
+ */
+class api_v3_AbsencePeriodTest extends BaseHeadlessTest {
+
+  public function testValidPeriodExistsReturnsFalseWhenThereIsNoAbsencePeriodWithEndDateGreaterThanOrEqualToCurrentDate() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('-5 days'),
+      'end_date'   => CRM_Utils_Date::processDate('yesterday'),
+      'weight'    => 1
```
mickadoo: Is `weight` required in all these calls, or what does it do?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/AbsencePeriodTest.php
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public function testValidPeriodExistsReturnsTrueWhenAbsencePeriodWithEndDateSameAsCurrentDateExists() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('-5 days'),
+      'end_date'   => CRM_Utils_Date::processDate('today'),
+      'weight'    => 1
+    ]);
+
+    $expectedResult = [
```
mickadoo: This could probably all be on one line

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
tunbola: I think its fine to leave the method name as is, The method gets periods between two dates and when the second parameter is not there, it simply returns periods between the first date and the future.  I remember I had some conversation with Davi about the name too before settling for this when I added the method initially.

uk.co.compucorp.civicrm.hrleaveandabsences/api/v3/AbsencePeriod.php
tunbola: Thanks for the suggestion, didn't think of that earlier. 
`AbsencePeriod.get` with `end_date` >= today would be more appropriate though, because `start_date > today` won't catch for example a period with dates `2017-01-01 to 2017-12-31`, the aim is to get the closest period to the current date which also includes period containing the current date.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/AbsencePeriodTest.php
tunbola: Will yank off these tests since we can use the API to retrieve the info after all.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
mickadoo: After the latest changes this PR is pretty small and the code all looks fine for me. I'm not sure if I know enough about the purpose of absence periods, because the whole PR is a bit confusing to me. What are absence periods for?

I can see the point of avoiding a 500, but I'm not sure why leave requests should be shoehorned into an absence period that doesn't actually contain them.  The logic @davialexandre mentioned above sounds like it's getting more and more complicated, and complex business logic usually means we'll have confused users in a few months.

mickadoo:
I think the description needs to be updated

> This PR fixes the issue by redirecting to Manage entitlements calculation page to calculate entitlements for the closest absence period to the current date after creating/updating a contract.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
tunbola: @mickadoo, the issue is actually about job contracts and absence periods not leave requests and absence periods.
Entitlements are calculated for absence periods and if when redirecting to the user to the entitlements calculation page after creating/updating a contract, there is no absence period containing the current date, we will redirect user to calculate entitlements for the closest absence period available, There is a previous/next button that allows the user to navigate absence periods and calculate entitlements for which period they want to calculate for.
You can take a look at the original ticket that documents the bug https://compucorp.atlassian.net/browse/PCHR-2952.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -618,4 +619,83 @@ public static function getPeriodsBetweenDates(DateTime $fromDate, DateTime $toDa
...
+  /**
+   * Returns the closest Absence Period to the current date.
+   * The absence period could be before the current date or
+   * could be after or could even contain the current date.
+   *
+   * @return AbsencePeriod|null
+   *   Returns null if there are no absence periods.
+   */
+  public static function getClosestToCurrentDate() {
+    $firstPeriod = self::getFirst();
```
davialexandre: There are a few problems with this method:
- The description doesn't describe the exact behavior of the method. It doesn't mention that the method favors absence periods in the future, even if we have there is a period in the past closer to the current date.
- It uses this `getFirst()` method which is problematic, because it order things based on the weight but then all the comparisons as done by dates, and we can have a absence period which the dates are in the future but the weight is 1. In the future we'll probably be removing this weigh field, but for now we shouldn't use it for ordering like.
- The method is a bit complex and may result in a lot of queries being sent to the database. 

For the last query, I think we can probably do it with a single query. Something like this:

```php
$today = date('Y-m-d');
$query = "SELECT * FROM absence_period ORDER BY ABS( DATEDIFF(start_date, {$today} ) ), start_date DESC LIMIT 1";
```

I based that on: https://stackoverflow.com/questions/6186962/sql-query-to-show-nearest-date. The ABS() function is so that both -1 and 1 are considered the same and the ordering by start_date is to make sure that period starting in the future will have a higher precedence. Do you think this could work?

tunbola: I didn't know the weight field is not reliable as I based most of the logic on it. 

Went with two queries as agreed since the query above will not give preference to absence periods in the future.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php
```diff
@@ -618,4 +619,48 @@ public static function getPeriodsBetweenDates(DateTime $fromDate, DateTime $toDa
...
+   * The absence period could be the one before the current date or
+   * could be after or could even be the one containing the
+   * current date.
+   *
+   * An absence period containing the current date is the most
+   * preferred, If it does not exist, then the next closest absence
+   * period in the future is preferred, If that does not exists, then
+   * the closest absence period in the past is considered.
+   *
+   * @return AbsencePeriod|null
```
davialexandre: We can't use alias in phpdoc

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsencePeriodTest.php
```diff
@@ -923,4 +923,95 @@ public function testGetPeriodsBetweenDatesReturnsEmptyWhenNoOverlappingPeriodsAr
...
+
+    //Period one is closer to the current date
+    $closestPeriod = AbsencePeriod::getClosestToCurrentDate();
+    $this->assertEquals($period1->id, $closestPeriod->id);
+  }
+
+  public function testGetClosestToCurrentDateReturnsNullWhenThereIsNoAbsencePeriod() {
+    $closestPeriod = AbsencePeriod::getClosestToCurrentDate();
+    $this->assertNull($closestPeriod);
+  }
```
davialexandre: Please also add a test for when there's no absence period

davialexandre: The test is right there @davialexandre. Go home, you're drunk

PR #2362, opened 2017-12-19
PCHR-3046: Add Latest Login for Roles

mickadoo:
## Overview

As part of the ping-back feature we send site statistics to a central server. We want to include the latest login for each role in the site statistics.

## Before

Latest login data was not included in the statistics.

## After

A role service will fetch all Drupal roles in the database and find the most recent login time for each of them. These will be included in the statistics.

## Comments

- The planned structure for role data in the JSON request body is [here](https://gist.github.com/mickadoo/3eb09bafa18760c12fcccced7a1afc33#file-pingback_json_spec-json-L5)
- Previously we just had a `DrupalRoleService`, but in order to make this more CMS-agnostic there is now a `cms_role_service` service in the container, which will return an instance of a `RoleServiceInterface`
- To keep consistency with other classes that interact with the CMS, the `DrupalRoleService` was moved to the CMSData directory


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/CMSRoleServiceFactory.php
```diff
@@ -0,0 +1,23 @@
...
+   * @return RoleServiceInterface
+   */
+  public static function create() {
+    $userSystem = CRM_Core_Config::singleton()->userSystem;
+
+    switch (get_class($userSystem)) {
+      case CRM_Utils_System_Drupal::class:
+        return new DrupalRoleService();
+      default:
+        $msg = sprintf('Unrecognized system "%s"', get_class($userSystem));
```
tunbola: Wondering if the user framework `CRM_Core_Config::singleton()->userFramework` won't be better to display here in case there is an unrecognised system.  The return value seems more user friendly and less technical than returning the class name.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/CMSRoleServiceFactory.php
```diff
@@ -0,0 +1,23 @@
...
+
+use CRM_HRCore_CMSData_Role_DrupalRoleService as DrupalRoleService;
+use CRM_HRCore_CMSData_Role_RoleServiceInterface as RoleServiceInterface;
+
+class CRM_HRCore_Service_CMSRoleServiceFactory {
+
+  /**
+   * @return RoleServiceInterface
+   */
+  public static function create() {
```
tunbola: A brief description for the function won't be a bad idea.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/CMSRoleServiceFactory.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+use CRM_HRCore_CMSData_Role_DrupalRoleService as DrupalRoleService;
+use CRM_HRCore_CMSData_Role_RoleServiceInterface as RoleServiceInterface;
+
+class CRM_HRCore_Service_CMSRoleServiceFactory {
+
+  /**
+   * @return RoleServiceInterface
+   */
```
tunbola: Please indicate that this function throws an exception.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
```diff
@@ -21,19 +15,39 @@ public function getRoleNames() {
    * @inheritdoc
    */
   public function getLatestLoginByRole($roleName) {
```
tunbola: `getLatestLoginForRole` seems to be a better name for this function. `getLatestLoginByRole` gives the impression that the function is retrieving latest login per role.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
```diff
@@ -21,19 +15,39 @@ public function getRoleNames() {
    * @inheritdoc
    */
```
tunbola: Please indicate this this function throws an exception.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/CiviHRStatsGatherer.php
```diff
@@ -51,6 +62,11 @@ public function gather() {
       $stats->addReportConfigurationAgeGroup($group);
     }
 
+    foreach ($this->roleService->getRoleNames() as $rid => $roleName) {
```
tunbola: Would it improve performance if we had the `getLatestLoginByRole` function, rather than get latest login for each role passed in, It will internally get the last login for all the roles using a single query. 
We can then loop through the returned results to `setMostRecentLoginForRole` for each role.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/CMSRoleServiceFactory.php
mickadoo: I didn't know about `userFramework`, this is much better including to use in the switch. I've changed this and the other place where I did a similar comparison (`CRM_HRCore_CMSData_SiteInformationFactory`)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
mickadoo: I've asked the other devs about this because I'm not sure. Personally I always turn off this warning in PHPStorm since I don't think it adds any value, but it will be good to hear other opinions

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/CiviHRStatsGatherer.php
mickadoo: Performance is a good point, and I'll change this. I did really like having the simpler function definition of accepting a string and returning `\DateTime | null`, whereas now it returns an array, where the indices are the role names. It does surely improve performance, but in general I think I prefer having easy to understand functions where you don't need to understand the structure of the return type to use it.

PR #2247, opened 2017-10-26
PCHR-2760: Onboarding Wizard Backstop tests

deb1990:
## Overview
As part of this PR, backstop JS tests has been added for the Onboarding Wizard Pages

## Technical Details
1. `onboarding-wizard-data.js` is added, which contains the form data for the required fields.

## Comments
1. In `onboarding-wizard-data.js` file, the ID's used as selectors, are not completely reliable, as they are generated by Drupal. But there is no other way to select a particular field. So went with this approach.

PR #2363, opened 2017-12-19
PCHR-2760: Fix Onboarding Wizard backstop JS Tests

deb1990:
## Overview
This PR fixes the backstop js tests for Onboarding Wizard, which was introduced on https://github.com/civicrm/civihr/pull/2247.

## Problem
1. The Path was wrong for the some of the JS files
2. For the test cases to work, the form needed to be submitted once.
3. Some of the form data used was not necessary as the already submitted form data could have been used.

## After
1. Fixed the paths in https://github.com/civicrm/civihr/pull/2363/files#diff-633654d50389da72e0fb1986cb0b4928L1

2. Added the requirement of form submission in 
https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/117984407/BackstopJS+CiviHR

3. Removed the unnecessary form data, only kept dependents data, as backstop is not able to use the same from saved data.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/backstop_data/data/onboarding-wizard-data.js
+++ b/uk.co.compucorp.civicrm.hrcore/backstop_data/data/onboarding-wizard-data.js
@@ -1,35 +1,5 @@
 module.exports = (function () {
   return {
-    personalDetails: {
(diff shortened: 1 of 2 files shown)


reneolivo:
Ran the tests locally. They work for all the on-boarding scenarios

PR #2365, opened 2017-12-19
PCHR-3028 Caching Stats

mickadoo:
## Overview

The Ping Back feature will gather statistics on a CiviHR site and send them to a central server. You can see what the request body will look like [here](https://gist.github.com/mickadoo/3eb09bafa18760c12fcccced7a1afc33). 

The statistics are sent in a scheduled job. This job  is run every day. Every time the server receives a different set of statistics from a client site it will store them. However, we only want to store a new result set every week. This will be achieved by only generating a result set once a week, and storing it in a cache. 

When the scheduled job runs it will send the current cache, and the server will determine whether it should save it or not.

## Before

There was no caching of the statistics.

## After

The `CRM_HRCore_Service_Stats_StatsCache` class will only generate new statistics when there is no cache, or if the current cache is more than a week old.

## Technical Details

The `StatsCache` uses a `FileCache`. The `FileCache` is a generic cache that stores serialized data to a file.

## Comments

- Since there are a few classes related to the CiviHR statistics now I moved them to a separate directory (and namespace) `HRCore/Service/Stats`
- As part of this I added some classes related to sending the stats, but this will fully covered in PCHR-3029


MiyaNoctem:
I just realized there was a commit that seems to have the wrong ticket number, 'PCHR-3029: Add logging', is that ok?

mickadoo:
Yeah, like I said in the description this included some classes from PCHR-3029, but the bulk of the work will be completed later in that ticket. I mostly added them as a placeholder for the desired behaviour

PR #2366, opened 2017-12-19
(Proof of Concept) Unified gulpfile

AkA84:
This is the poc for a unified gulpfile for all the civihr extensions.

## Problem
Currently each extension has a separate gulpfile, but the tasks are basically always the same (sass, requirejs, task) with some custom logic here and there. This leads to a lot of duplication that is hard to maintain and that will only grow overt ime

## Solution
The idea is to have a single gulpfile under the hrcore extension, that will be able to run any common task for any given extension, and to watch for any changes of any files in any extension and to react accordingly.

This is how it can be used currently
```bash
gulp sass --ext {extension-name}
gulp requirejs --ext {extension-name}
gulp test --ext {extension-name}

gulp sass:watch
gulp requirejs:watch
gulp test:watch
```
The main tasks (sass, requirejs, test) needs the extension name passed, otherwise they will error out
The watchers (xyz:watch) do not need the extension name, they watch files in every extension. Once a file change, they will find the extension they belong to and then call the main task for that extension

### Custom logic
An extension might have some custom logic for a given task that it needs to run. It could either be some logic that needs to run before or after the main task, or it could be a logic that needs to replace the main task altogether.

To support this, the gulpfile, for any main task, will look for a file under a `gulp-tasks/` folder with the same name as the task. If this file exists, it needs to define a function that once invoked return an object with any of these properties: `main`, `pre`, `post`

`main`: it's a function that replaces the main task altogether, see [here](https://github.com/civicrm/civihr/pull/2366/files#diff-5f9221f200b064549a705645668004eeR7) for an example;
`pre`: it's an array of objects, each object representing a task that needs to be run before the main task;
`post`: it's an array of objects, each object representing a task that needs to be run after the main task;

### Notes
* I left `@NOTE`s in the code to give more details on the proposed solution
* The watchers take something like 20 seconds to start, the final gulpfile will have to be a little bit smarter with the selectors
* The `requirejs` task currently does not support the cross-extension dependencies, unlike what you can find for example in the [hrjobcontract gulpfile](https://github.com/civicrm/civihr/blob/staging/hrjobcontract/gulpfile.js#L21-L25). This was skipped for time constraints, but it shouldn't be a technical problem to implement
* This gulpfile does not cover either T&A or Shoreditch, as those are extensions used in CiviHR, *not* CiviHR extensions

### DIFF
--- a/hrjobcontract/gulp-tasks/sass.js
+++ b/hrjobcontract/gulp-tasks/sass.js
@@ -1,0 +1,18 @@
+var exec = require('child_process').exec;
+
+module.exports = function (SubTask) {
+  return {
+    full: full
+  };
+
+  /**
+   * The app style relies on compass's gems, so we need to rely on it
+   * for the time being
+   */
+  function full (cb) {
+    exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
+      console.log(stdout);
+      cb();
+    });
+  }
+};
--- a/hrjobcontract/gulp-tasks/sass.js
+++ b/hrjobcontract/gulp-tasks/sass.js
@@ -1,18 +1,10 @@
 var exec = require('child_process').exec;
 
 module.exports = function (SubTask) {
-  return {
-    full: full
-  };
-
-  /**
-   * The app style relies on compass's gems, so we need to rely on it
-   * for the time being
-   */
-  function full (cb) {
+  return function (cb) {
     exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
       console.log(stdout);
       cb();
     });
-  }
+  };
 };
--- a/hrjobcontract/gulp-tasks/sass.js
+++ b/hrjobcontract/gulp-tasks/sass.js
@@ -1,10 +1,12 @@
 var exec = require('child_process').exec;
 
-module.exports = function (SubTask) {
-  return function (cb) {
-    exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
-      console.log(stdout);
-      cb();
-    });
+module.exports = function () {
+  return {
+    main: function (done) {
+      exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
+        console.log(stdout);
+        done();
+      });
+    }
   };
 };
--- a/hrjobcontract/gulp-tasks/sass.js
+++ b/hrjobcontract/gulp-tasks/sass.js
@@ -2,6 +2,8 @@ var exec = require('child_process').exec;
 
 module.exports = function () {
   return {
+    // @NOTE this will replace the original "sass:main" task, as job contract
+    // needs a completely different logic for compiling sass
     main: function (done) {
       exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
         console.log(stdout);
--- a/org.civicrm.bootstrapcivihr/gulp-tasks/sass.js
+++ b/org.civicrm.bootstrapcivihr/gulp-tasks/sass.js
@@ -6,14 +6,21 @@ var bootstrapNamespace = '#bootstrap-theme';
 var outsideNamespaceRegExp = /^\.___outside-namespace/;
 
 module.exports = function (SubTask) {
-  return SubTask.pipe(postcss, [postcssPrefix({
-    prefix: bootstrapNamespace + ' ',
-    exclude: [
-      /^html/, /^body/, /page-civi/, /crm-container/, outsideNamespaceRegExp
-    ]
-  })])
-  .pipe(transformSelectors, namespaceRootElements, { splitOnCommas: true })
-  .pipe(transformSelectors, removeOutsideNamespaceMarker, { splitOnCommas: true });
+  return {
+    post: post
+  };
+
+  function post () {
+    return SubTask.pipe(postcss, [postcssPrefix({
+      prefix: bootstrapNamespace + ' ',
+      exclude: [
+        /^html/, /^body/, /page-civi/, /crm-container/, outsideNamespaceRegExp
+      ]
+    })])
+    .pipe(transformSelectors, namespaceRootElements, { splitOnCommas: true })
+    .pipe(transformSelectors, removeOutsideNamespaceMarker, { splitOnCommas: true })
+    .run();
+  }
 };
 
 /**
--- a/org.civicrm.bootstrapcivihr/gulp-tasks/sass.js
+++ b/org.civicrm.bootstrapcivihr/gulp-tasks/sass.js
@@ -1,3 +1,4 @@
+var gulp = require('gulp');
 var postcss = require('gulp-postcss');
 var postcssPrefix = require('postcss-prefix-selector');
 var transformSelectors = require('gulp-transform-selectors');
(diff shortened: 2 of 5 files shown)


reneolivo:
Can we run the gulp commands using NPM: `npm run sass`, `npm run requirejs`, `npm run tests`, etc. And have scripts like these in the package.json file:

```json
"scripts": {
  "requirejs": "gulp requirejs",
  "sass": "gulp sass",
  "tests": "gulp tests"
}
```

This avoids installing gulp globally since NPM scripts first try to find the binaries in the local node_modules/.bin folder.

uk.co.compucorp.civicrm.hrleaveandabsences/gulp-tasks/requirejs.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = function () {
+  // @NOTE one thing to keep in mind is that the custom task always needs to use
+  // __dirname for paths, otherwise the paths will be relative to hrcore
+  var distFolder = path.join(__dirname, '..', 'js/angular/dist/');
+
+  return {
+    // @NOTE here L&A needs to perform to tasks after r.js had run, both for
+    // cleaning the dist/ folder. Notice that the order of the tasks is the
+    // order in which they will be called
+    post: [
```
reneolivo: Can we further split the post and pre tasks into different files?

```js
// gulp-tasks/requirejs.js
return {
  post: [
    require('./post-requirejs-rename'),
    require('./post-requirejs-clean')
  ]
}
```

Or, the main gulpfile can have a few extra hooks: requirejs-rename and requirejs-clean.

deb1990:
> To support this, the gulpfile, for any main task, will look for a file under a gulp-tasks/ folder with the same name as the task. If this file exists, it needs to define a function that once invoked return an object with any of these properties: main, pre, post

Instead of separate files, why dont we create a single file called `gulp-tasks.js` in the root of extensions which will contain the custom logic
```
var exec = require('child_process').exec;

module.exports = function () {
  return {
    sass: {
    	main: function (done) {
	      exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
	        console.log(stdout);
	        done();
	      });
    	}
    },
    requirejs: {},
    ...
  };
};
```

This will create less files and code can be re used if needed without having to import other files.

deb1990:
We should have a command which would build and test all extensions, like
`gulp build`/`gulp` -  would build `sass` and `requirejs` for all extensions
`gulp test` - would unit test all extension

now that we have different extensions using each other's files, this is really necessary to ensure that someone is not forgetting to build a specific extension.

deb1990:
Another item to wish list is to add support of alias names for extensions, as now we need to specify extension names everytime we build, and we are not going to get auto complete, it will be a huge headache. So something like
`gulp sass --ext la`
`gulp sass --ext jc`
`gulp sass --ext jr`
...

would make development much easier.

deb1990:
> This gulpfile does not cover either T&A or Shoreditch, as those are extensions used in CiviHR, not CiviHR extensions

We might think about making this as a npm package, just like `civicrm-scssroot`. So the idea is to create a npm package which will support the idea mentioned in the PR description. Also it can support folders outside of the current folder. 
As I mentioned about the support for aliases, that will help us to support any folders using relative paths. so we can have a json file for aliases
```
{
	jc: 'path to jobcontract',
	jr: 'path to jobroles',
	....
	sh: './../path to shoreditch',
	ta: './../path to t&A'
}
```
this will be an universal solution and would work for non civihr or non civicrm projects too. Would be cool if we can achieve this. What say guys?

igorpavlov:
@AkA84 would it be possible to 

```bash
gulp sass --ext {extension-1} && gulp test --ext {extension-1} && gulp requirejs --ext {extension-1} && gulp sass --ext {extension-2} && gulp test --ext {extension-2} && gulp requirejs --ext {extension-2} ... etc
```
?
Basically, I want a command that will exec all tests and then build everything in CiviHR.

igorpavlov:
> the final gulpfile will have to be a little bit smarter with the selectors

Does this mean "the final gulpfile will have to select only those files that are needed"?

hrjobcontract/gulp-tasks/sass.js
```diff
@@ -0,0 +1,14 @@
+var exec = require('child_process').exec;
+
+module.exports = function () {
+  return {
+    // @NOTE this will replace the original "sass:main" task, as job contract
+    // needs a completely different logic for compiling sass
+    main: function (done) {
+      exec('compass compile', { cwd: __dirname + '/../' }, function (_, stdout, stderr) {
```
igorpavlov: I don't think we should use '_', is that an event param? Better to use 'event' param, it will be skipped by semistandard I think.

org.civicrm.bootstrapcivihr/gulp-tasks/sass.js
```diff
@@ -0,0 +1,56 @@
...
+      {
+        name: 'sass:namespace',
+        fn: function () {
+          var cssDir = __dirname + '/../' + 'css/';
+
+          return gulp.src(cssDir +  '*.css')
+            .pipe(postcss([postcssPrefix({
+              prefix: bootstrapNamespace + ' ',
+              exclude: [
+                /^html/, /^body/, /page-civi/, /crm-container/, outsideNamespaceRegExp
```
igorpavlov: Should not it be `\.crm-container` or am I missing something?

igorpavlov: As per `/page-civi/`, would it match `.page-civicrm-leaveandabsences-dashboard`? Do we expect that?

AkA84:
@reneolivo 
> Can we run the gulp commands using NPM: npm run sass, npm run requirejs, npm run tests
> [...]
> This avoids installing gulp globally

I was thinking about that but in the end decided against it. The reason is simple: it deviates from what devs expect when they see a gulpfile. If you know that there is a gulpfile, you expect to run it with `gulp`, not with `npm`. Any custom commands add complexity to the system, and if we can avoid that, then we should.

Think about how much time has been wasted with the SSP theme because instead of running `npm install` one needs to run `npm run setup` to install all the dependencies. Everyone stumbles on that because it deviates from common expectations

So bottom line is: there is a gulpfile, install gulp globally and use it, like you would do in any other project

AkA84:
@deb1990 

> Instead of separate files, why dont we create a single file called gulp-tasks.js in the root of extensions which will contain the custom logic

Honestly i like more the modular approach of having one file per task (i don't see the problem with having multiple files). What if an extension requires a lot of custom tasks for any given main task? The single file would be huge and harder to read and maintain i think. In the end it's why we are talking about splitting the main gulpfile into smaller modules: it will work as a single huge file anyway, but we still split it because that's the more maintainable approach.

I'm not sure what you mean by "the code can be reused", these files (multiple or single, regardlesss) are made specifically for this singule gulpfile architecture, i don't know how they could be reused


> We should have a command which would build and test all extensions, like
> `gulp build/gulp` - would build sass and requirejs for all extensions
> `gulp test` - would unit test all extension
>
> now that we have different extensions using each other's files, this is really necessary to ensure that someone is not forgetting to build a specific extension.

Was discussing with @igorpavlov and the concern over cross-extension dependency is valid, but instead of addressing it by brute-force approach (rebuild *everything* just to make sure), what i'm planning is to have the requirejs task to be smart enough to detect if any extension has the dependency on the extension that it's currently bundling. So for example:

1. Changes are made on some L&A model
2. `gulp requirejs --ext L&A`
3. the .min files for L&A are created
4. Next the task scans if any other extension, in its `build.js` file, has L&A has a dependency (you can tell that because we are using a special syntax for marking external dependencies)
5. For any extension that is found to have a dependency on L&A, then the equivalent of `gulp requirejs --ext <name-of-extension>` is invoked
6. Then gulp checks if *that* extension is a dependency for other extensions and so on, recursively

That's the goal that I hope to achieve


> Another item to wish list is to add support of alias names for extensions
> [...]
> As I mentioned about the support for aliases, that will help us to support any folders using relative paths. so we can have a json file for aliases

You can't statically define paths for extensions, you can't predict where they will be placed in any given time.

I understand the annoyance with the extension names (I too have had to open the info.xml file multiple times to copypaste its name), but a static list of extension names will be yet another thing to maintain that risks getting out of sync pretty quickly, i don't know if it's really worth it.

> We might think about making this as a npm package, just like civicrm-scssroot

(I was actually thinking of making the `test` object a package, cause that is definitely going to be reused in T&A), This is a solution for supporting multiple extensions in one gulpfile, but T&A is a single extension, i'm not sure if it would make sense to use it there?

uk.co.compucorp.civicrm.hrleaveandabsences/gulp-tasks/requirejs.js
AkA84: not sure what you mean about the extra hooks, the whole idea is that the main gulpfile doesn't know and doesn't need to know about whatever custom logic a single extension needs to run

deb1990:
@AkA84 

> Instead of separate files, why dont we create a single file called gulp-tasks.js in the root of extensions which will contain the custom logic

ok I agree.


> We should have a command which would build and test all extensions, like
gulp build/gulp - would build sass and requirejs for all extensions
gulp test - would unit test all extension

Its a good idea to detect dependencies and would be cool if we can achieve that.
But again isn't it good to have a command which rebuilds everything? because that would make sure that by mistake we are not forgetting to build some extension. 

Also is it a good time to think about removing build files(*.min) from git? now that we have continues integration. Though I am not sure about other architectural/build related changes we need to do before doing this. 

>Another item to wish list is to add support of alias names for extensions

why dont we add a json file but keep it empty, like the backstop config file we have. then the developers can set it up as per their need.
```
{
	jc: '<add path to jobcontract>',
	jr: '<add path to job roles>',
	....
}
```
so we will have the functionality ready but also with the ability to specify custom paths. This feature I really want to push for, because I am already getting a feeling of the annoyance we all are going to face.


>(I was actually thinking of making the test object a package, cause that is definitely going to be reused in T&A), This is a solution for supporting multiple extensions in one gulpfile, but T&A is a single extension, i'm not sure if it would make sense to use it there?

ok lets skip it, lets not increase the scope too much for now.

PR #2368, opened 2017-12-20
PCHR-3067: Add the contactactionsmenu extension.

tunbola:
## Overview
This PR creates the uk.co.compucorp.civicrm.hrcontactactionsmenu  extension. This is just the first commit for this extension and it only contains the basic skeletal extension structure.

This extension will provide functionality that allows other extensions add menu items to the contact actions menu on contact summary page.
The extension will provide a hook called addContactMenuActions, This hook can be invoked by other extensions that needs to add menu items to specific panels of the contact actions menu.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcontactactionsmenu/LICENSE.txt
+++ b/uk.co.compucorp.civicrm.hrcontactactionsmenu/LICENSE.txt
@@ -1,0 +1,667 @@
+Package: uk.co.compucorp.civicrm.hrcontactactionsmenu
+Copyright (C) 2017, Compucorp Ltd <info@compucorp.co.uk>
+Licensed under the GNU Affero Public License 3.0 (below).
+
+-------------------------------------------------------------------------------
+
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+our General Public Licenses are intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  Developers that use our General Public Licenses protect your rights
+with two steps: (1) assert copyright on the software, and (2) offer
+you this License which gives you legal permission to copy, distribute
+and/or modify the software.
+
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.contactactionsmenu/info.xml
```diff
@@ -0,0 +1,21 @@
...
+    <email>info@compucorp.co.uk</email>
+  </maintainer>
+  <releaseDate>2017-12-20</releaseDate>
+  <version>1.7.2</version>
+  <develStage>alpha</develStage>
+  <compatibility>
+    <ver>4.7</ver>
+  </compatibility>
+  <civix>
+    <namespace>CRM/Contactactionsmenu</namespace>
```
davialexandre: please change this to `CRM/HRContactActionsMenu`

uk.co.compucorp.civicrm.contactactionsmenu/info.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<extension key="uk.co.compucorp.civicrm.contactactionsmenu" type="module">
```
davialexandre: To keep it consistent with the other extension, I think it would be better to call it `uk.co.compucorp.civicrm.hrcontactactionsmenu`, as it will be a CiviHR extension for now. We can rename it If we want to make it a standalone extension in the future.

uk.co.compucorp.civicrm.contactactionsmenu/contactactionsmenu.civix.php
```diff
@@ -0,0 +1,445 @@
+<?php
+
+// AUTO-GENERATED FILE -- Civix may overwrite any changes made to this file
+
+/**
+ * The ExtensionUtil class provides small stubs for accessing resources of this
+ * extension.
+ */
+class CRM_Contactactionsmenu_ExtensionUtil {
```
davialexandre: based on the namespace update I mentioned on another comment, this class name will need to be updated too

uk.co.compucorp.civicrm.contactactionsmenu/contactactionsmenu.civix.php
```diff
@@ -0,0 +1,445 @@
...
+  const SHORT_NAME = "contactactionsmenu";
+  const LONG_NAME = "uk.co.compucorp.civicrm.contactactionsmenu";
+  const CLASS_PREFIX = "CRM_Contactactionsmenu";
```
davialexandre: this will need to be updated too

tunbola: I wanted to change this but since its an autogenerated file I was wondering if it won't be reverted back?

uk.co.compucorp.civicrm.contactactionsmenu/info.xml
tunbola: I wanted to change it to the camel case style but the CLASS_PREFIX constant in contactactionsmenu.civix.php is different and the class is autogenerated.

uk.co.compucorp.civicrm.contactactionsmenu/contactactionsmenu.civix.php
davialexandre: I don't think it will. This file is autogenerated but it's never updated as far as I concerned, so the changes won't be reverted

uk.co.compucorp.civicrm.contactactionsmenu/info.xml
davialexandre: yes, it's fine to change it. This file is never auto updated after it has been created

PR #2372, opened 2017-12-21
PCHR-3095: Add support for running tests in Jenkins for feature branches

davialexandre:
## Problem

With our current workflow, not all Pull Requests target staging. Most of the time, a PR will target a feature branch, which groups a number of PRs for smaller tasks, until the feature is complete and finally ready to be merged to staging. 

It was not possible for Jenkins to run the tests for such Pull Requests. Before running the tests, a new site is created using `civibuild` and the PR's branch as the value for the `--hr-ver` option. The command then tries to pull the code from this branch in all the repositories used by CiviHR but, in most of the cases, the branch will only exist in one or two of them and the installation will fail, making Jenkins mark the build as failed.

## Solution

After the site is created and before running the tests, Jenkins:
- Checks out the PR branch
- Merge the PR target branch into the PR branch
- Run upgraders

This is enough to guarantee the tests will run on top of the correct branch. For that reason, for tests, the branch used during the site creation is not really important. Based on that, the solution was to update the Jenkinsfile so that the `staging` branch will always be used while creating the site.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,6 @@ pipeline {
   agent any
 
   parameters {
-    string(name: 'CIVIHR_BRANCH', defaultValue: '', description: 'Default build branch of CiviHR to build site using CiviCRM-Buildkit')
     string(name: 'CIVIHR_BUILDNAME', defaultValue: "hr17-dev_$BRANCH_NAME", description: 'CiviHR site name')
     booleanParam(name: 'DESTROY_SITE', defaultValue: true, description: 'Destroy built site after build finish')
   }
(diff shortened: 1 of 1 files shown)


mickadoo:
I'm just wondering, since staging will always be used during creation what if you have a PR that would require release notes for upgrade, but would work without release notes when creating a new site. It would be nice to be able to test these with a clean site built using the corresponding buildkit branch, otherwise I don't see how we can run tests on any features that has release notes.

davialexandre:
@mickadoo I've thought of that, but then I realized that, for unit tests, this shouldn't/doesn't matter. For this type of test, the upgraders should be enough to make sure the database will be in the correct state to run the tests. If a test fails because of a missing release note, then I think this is an indication that we might be doing something wrong.

Running release notes would certainly affect e2e tests though, but we don't have this type of test for now and, when we have, I don't think we'll run them for every PR. They take a lot of time to run, so we'll need a different strategy when we get there.

PR #2374, opened 2017-12-22
PCHR-2919: Fix Admin Tabs Responsiveness

deb1990:
## Overview
In Job roles when the browser is shrunk, at some point some of the text gets hidden from the right hand panel.

## Before
responsivetabs

## After
a

Also Fixed the same for Absence Tab.
aa

## Technical Details
1. Usage of `container` class is not ideal in the right side area, because `container` class should only be used when using the full width of the screen. So replaced `container` with `container-fluid` in `JobRoles`, `JobContract` and  `Tasks` and `Documents`(https://github.com/compucorp/civihr-tasks-assignments/pull/294) .

2. (Fix for reopened ticket) Removed any code which was changing the width or padding of `.container` or `.container-fluid`.


deb1990:
@AkA84 My changes does not override those css rules. I have changed `max-width`(not `width`), so it still follows those rules but does not allow the width to be more than width of the right side area.

Also having `container` class is necessary, because that makes sure that the right side content does not take full width in wide screens.
cursor_and_asd_dqwd___civihrnew2

AkA84:
@deb1990 but that's not how the `.container` element is supposed to be used, so i think my point still stands

deb1990:
> @deb1990 this PR includes minified js files, but no js changes have been made

I am using VS Code where I have installed a plugin to remove any trailing white space, so it did the same for the min js files. Removed them.

PR #2375, opened 2017-12-22
PCHR-3100: Fix hremergency PHP tests

davialexandre:
## Overview

Jenkins cannot run the hremergency tests. Whenever it tries it, it fails with this error:

```
[civihr_PR-2238-QTNEV4RQWI3LHQFNHVVNONTJAJCGQNOPCH4L6BQMFDWYHF7L54LQ] Running shell script
+ cd /opt/buildkit/build/hr17-dev_PR-2238/sites/all/modules/civicrm/tools/extensions/civihr/org.civicrm.hremergency
+ phpunit4 --testsuite=Unit Tests --log-junit /var/lib/jenkins/workspace/civihr_PR-2238-QTNEV4RQWI3LHQFNHVVNONTJAJCGQNOPCH4L6BQMFDWYHF7L54LQ/reports/phpunit/result-phpunit_hremergency.xml

RuntimeException: Suite includes headless tests (CRM_Hremergency_Service_EmergencyContactServiceTest) which require CIVICRM_UF=UnitTests.

Tip: Run the headless tests and end-to-end tests separately, e.g.
  $ phpunit4 --group headless
  $ phpunit4 --group e2e  

 in Civi\Test\CiviTestListener->autoboot() (line 184 of /opt/buildkit/build/hr17-dev_PR-2238/sites/all/modules/civicrm/Civi/Test/CiviTestListener.php).
```

These tests have been added by https://github.com/civicrm/civihr/pull/2273 and have been failing since then.

## Before

Jenkins was not able to run the hremergency tests.

## After

Jenkins can now run the hremergency tests and they all pass.

## Technical Details

The main reason for the error was that the bootstrap.php file was trying to bootstrap a regular civicrm environment instead of a test one, where the `CIVICRM_UF` constant is set to "UnitTests". This has been fixed by changing this line on the bootstrap.php file:

```php
eval(cv('php:boot --level=full', 'phpcode'));
```

to this (note the `-t` that has been added):

```php
eval(cv('php:boot --level=full -t', 'phpcode'));
```

Besides that, a few adjustments have been made to make sure the tests are consistent with the structure we've been trying to use for all extensions:

- A `BaseHeadlessTest` class was added in order to have a single `setUpHeadless()` method that can be used by all of the extension's tests
- The tests use some helpers and fabricators from hrcore. Before, it was including these files directly in the bootstrap.php file. This has been updated and now we install the hrcore ext in the `setUpHeadless()` method instead, so that these classes can be autoloaded. The reason here is so that we don't need to know the exact folder of the hrcore extension
- One of the helpers being imported from hrcore was a trait. Traits are evaluated when a class is included and this happens before the `setUpHeadless()` method gets called to install the dependencies. This means that at the moment the trait is evaluated, it was not available for autoloading yet and the test would fail. To solve this, the trait was turned into a regular class and its methods have been made static.

## Comments

The reason we didn't catch this issue right when https://github.com/civicrm/civihr/pull/2273 was created, is that we still didn't have support to run tests in Jenkins for PRs targeting a feature branch at that time.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Helpers/SessionHelper.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Helpers/SessionHelper.php
@@ -1,18 +1,18 @@
 <?php
 
-trait CRM_HRCore_Test_Helpers_SessionHelpersTrait {
+class CRM_HRCore_Test_Helpers_SessionHelper {
 
-  protected function registerCurrentLoggedInContactInSession($contactID) {
+  public static function registerCurrentLoggedInContactInSession($contactID) {
     $session = CRM_Core_Session::singleton();
     $session->set('userID', $contactID);
   }
 
-  protected function unregisterCurrentLoggedInContactFromSession() {
+  public static function unregisterCurrentLoggedInContactFromSession() {
     $session = CRM_Core_Session::singleton();
     $session->set('userID', NULL);
   }
 
-  protected function setPermissions(array $permissions = []) {
+  public static function setPermissions(array $permissions = []) {
     CRM_Core_Config::singleton()->userPermissionClass->permissions = $permissions;
   }
 
--- a/org.civicrm.hremergency/CRM/Hremergency/Test/BaseHeadlessTest.php
+++ b/org.civicrm.hremergency/CRM/Hremergency/Test/BaseHeadlessTest.php
@@ -8,6 +8,7 @@ abstract class CRM_Hremergency_Test_BaseHeadlessTest extends PHPUnit_Framework_T
 
   public function setUpHeadless() {
     return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
       ->installMe(__DIR__)
       ->apply();
   }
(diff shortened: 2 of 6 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Test/Helpers/SessionHelper.php
```diff
@@ -1,18 +1,18 @@
 <?php
 
-trait CRM_HRCore_Test_Helpers_SessionHelpersTrait {
+class CRM_HRCore_Test_Helpers_SessionHelper {
```
tunbola: Does it make sense for the HREmergency extension to have its own session helpers trait rather than installing an extension just to use the session helper methods ?
If No, then probably extensions like L&A and other extension that use their own session helpers trait should later eventually rather use the one from HRCore instead of having the same traits repeated all around

davialexandre: Besides the SessionHelper, we also use the ContactFabricator from hrcore (and we can have more in the future!). I don't think it makes sense to duplicate these two classes just to avoid installing hrcore.

You're a right about the other extensions. Eventually, they will have to be updated later to use the hrcore helper instead of they own implementation.

PR #2376, opened 2017-12-22
PCHR-3069: Add the addContactMenuActions Hook implementation

tunbola:
## Overview
This HRContactActionsMenu provides functionality that allows other extensions to add menu items to the contact actions menu on contact summary page via the `addContactMenuActions` hook.
This PR defines the hook and the parameter that other extensions calling the hook will expect.

## Before
- The impementation for the  addContactMenuActions hook was not present.

## After
- The implementation for the  addContactMenuActions hook was added.
- The action to invoke the addContactMenuActions hook will be added in the `contactactionsmenu_civicrm_pageRun` hook implementation. This will be added in subsequent PR when the template for the contact actions menu is added.

 ## Technical Details
A sample implementation of the hook e.g by the Leave and Absence extension will be:
```php
function hrleaveandabsences_addContactActions(ActionsMenu $menu) {
//code 
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
+++ b/uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
@@ -14,11 +14,7 @@ class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest extends BaseHeadle
 
   private $groupTitle;
 
-  public function setUp() {
-    CRM_Utils_Hook::singleton()->setHook('addContactMenuActions', array($this, 'hook_civicrm_addContactMenuActions'));
-  }
-
-  public function hook_civicrm_addContactMenuActions(ActionsMenu $menu) {
+  public function hook_addContactMenuActions(ActionsMenu $menu) {
     $this->groupTitle = 'Test Group';
     $testGroup = new ActionsGroup($this->groupTitle);
     $menu->addToMainPanel($testGroup);
--- a/uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
+++ b/uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
@@ -1,0 +1,16 @@
+<?php
+
+use CRM_HRContactActionsMenu_Hook_AddContactMenuActions as AddContactMenuActionsHook;
+use CRM_HRContactActionsMenu_Component_Menu as ActionsMenu;
+
+/**
+ * Class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest
+ *
+ * @group headless
+ */
+class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest extends BaseHeadlessTest {
+
+  public function testInvokeReturnsAnActionsMenuInstance() {
+    $this->assertInstanceOf(ActionsMenu::class, AddContactMenuActionsHook::invoke());
+  }
+}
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest
+ *
+ * @group headless
+ */
+class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest extends BaseHeadlessTest {
+
+  public function testInvokeReturnsAnActionsMenuInstance() {
+    $this->assertInstanceOf(ActionsMenu::class, AddContactMenuActionsHook::invoke());
+  }
```
davialexandre: It would also be interesting to add a test to check that the invoke() method actually triggers the hook. You can use the `HookInterface` to help you with that. Check the example in the interface's docblock

tunbola: I tried that and took a look at some core tests that tested hooks, Most of them call an API or method that is supposed to implement/call the particular hook and then define the hook implementation in the test class. However in this particular case, The hook is not yet implemented or called by other extensions so I couldn't go with that approach.

davialexandre: I'm not sure I understand. How come is the hook is not called? You can just call it by calling the `invoke()`, can't you?

For the test, I was thinking of something like this:

```php
// This will be called when AddContactMenuActions::invoke() invokes the hook
public function hook_civicrm_addContactMenuActions(ActionsMenu $menu) {
  $dummyGroup = new ActionsGroup('FooBar');
  $menu->addToMainPanelGroup($dummyGroup);
}

public function testInvokeInvokesTheAddContactMenuActionsHook() {
  // Invokes the hook, which will call the hook implementation in this class, which will add a 
 // group to the menu
   $menu = AddContactMenuActions::invoke();
   $groups = $menu->getMainPanelItems();
// We check if the expected group was added to the menu
   $this->assertCount(1, $groups);
   $this-assertEquals('FooBar', $groups[0]->getTitle());
}
```

tunbola: Thanks! Seems to work just fine.
I could almost swear I tried this earlier and it didn't work

uk.co.compucorp.civicrm.hrcontactactionsmenu/tests/phpunit/CRM/HRContactActionsMenu/Hook/AddContactMenuActionsTest.php
```diff
@@ -2,15 +2,33 @@
...
  */
-class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest extends BaseHeadlessTest {
+class CRM_HRContactActionsMenu_Hook_AddContactMenuActionsTest extends BaseHeadlessTest implements HookInterface {
 
-  public function testInvokeReturnsAnActionsMenuInstance() {
-    $this->assertInstanceOf(ActionsMenu::class, AddContactMenuActionsHook::invoke());
+  private $groupTitle;
+
+  public function setUp() {
+    CRM_Utils_Hook::singleton()->setHook('addContactMenuActions', array($this, 'hook_civicrm_addContactMenuActions'));
```
davialexandre: Is this really necessary? I think that the HookInterface should be enough for civi to do this automatically

tunbola: I copied this from core tests, Will remove and see if it works.

PR #2379, opened 2017-12-28
PCHR-2844: Refresh Cancel Rejected requests from Calendar

deb1990:
## Overview
'Cancelled'/'Rejected' requests were shown on manager calendar until refreshing page, this PR fixes that.

## Before
b

## After
a

## Technical Details
1. There were console errors. Thats happens because when a request is changed, the system tries to fetch the request from indexed requests for all months,

`var oldLeaveRequest = leaveRequestFromIndexedList(leaveRequest);`

but for other months than the selected months, this list is empty, so `oldLeaveRequest` becomes `undefined` and hence the error.

2. After a request is changed from say `Approved` to `Rejected`, the same needs to be removed from the calendar because we do not show `Rejected` records. But this check was happening only while loading the requests, not after changing the same. So the same check is added.

3. After a request is changed and the current month has no requests to show, 'There are no staff members matching selected filters' should be displayed. But it was not happening, it has been fixed.

## Comments
This PR also includes a fix for PCHR-2552.
We found a fix for a Backstop test `Leave and Absences: Manager Leave Requests: Show staff edit toil request` previously. Added the same in this PR.

## Update
https://github.com/civicrm/civihr/pull/2379 has been added to fix 79 failing test cases. It happened because `absence-period-data.js` did not have a entry for the year 2018, so 79 test cases failed as soon as we hit 2018. Added code to always add the current year to the mock data, so that we dont have to manually update this file every year.


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -522,8 +532,18 @@ define([
     function updateLeaveRequest (leaveRequest) {
       var oldLeaveRequest = leaveRequestFromIndexedList(leaveRequest);
 
-      deleteLeaveRequest(oldLeaveRequest);
-      addLeaveRequest(leaveRequest);
+      if (oldLeaveRequest) {
```
reneolivo:
```js
if (!oldLeaveRequest) {
  return;
}

// everything else .....
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -522,8 +532,18 @@ define([
...
+        deleteLeaveRequest(oldLeaveRequest);
+
+        if (leaveStatusesToBeDisplayed().indexOf(leaveRequest.status_id) !== -1) {
+          addLeaveRequest(leaveRequest);
+        }
+
+        // if there are no leave requests for the contact id, removing the object for the contact
+        if (Object.keys(leaveRequests[leaveRequest.contact_id]).length === 0) {
```
reneolivo: Please remove this after refactoring the changes we discussed in Hipchat.

PR #2381, opened 2017-12-28
PCHR-3097: test

davialexandre:


davialexandre:
This was just a test to check how/if Jenkins could handle PRs from forks. It will not be implemented for now

PR #2382, opened 2017-12-28
PCHR-3097: Jenkins performance improvements

davialexandre:
## Overview

Currently, it takes about 20 min for Jenkins to finish a build. This PR introduces a few changes that take this time down to ~12min. These changes are mainly about only installing what is really necessary to run the tests and then running them in parallel.

## Before

It would take ~20 min for Jenkins to finish a build. This included:

- Creating a site (using the hr17 build type): ~6 min
- Running the PHP tests: ~8 min
- Installing the JS dependencies (with yarn):  ~2 min
- Running the JS tests: ~3min

For reference, here are the numbers from the last 4 build for staging:

captura de tela de 2018-01-02 09-10-47

## After

Now it takes about ~12 min for Jenkins to complete the Build:

- Creating a site (using the drupal-clean build type): ~3min
- Running the tests (PHP and JS are run in parallel now. The JS dependencies are installed when the tests run): ~9min 
  - For the PHP tests, it takes about 9 min for them to run
  - For JS (installing dependencies + running the tests) it takes about 4min

For reference, here are the numbers from the last 4 build for this PR (note that the average shows a bigger number, but that's because it's also considering builds from before the improvements):

captura de tela de 2018-01-02 09-11-49

### Improvements

About 50% of the improvement comes from the change of the build type used to create the test site. The `hr17` type, which was used before, does a lot of things like installing and configuring multiple drupal modules, installing a bunch of extensions, changing civicrm configuration, creating users etc. All of that takes time, but it's not necessary in order to run the tests as none of these changes are applied to the test database. For these unit tests, all we need a clean civicrm installation on top of Drupal, so changing the build type to `drupal-clean` helped us reduce the time to have a working site to run the tests. The PHP tests themselves install all the required extensions right before they run and the JS tests don't even hit the backend, so what is installed there doesn't really make a difference.

The other part of the improvement was achieved by running the tests in parallel. For the PHP tests, it now takes one more minute to run them, in average, but the overall time to run the tests reduced from 13min to 9min. This was done by adding a new stage called "tests" and then adding a parallel statement to it, containing the two set of tests: 

```groovy
stage('Run tests') {
   parallel {
      stage('Test PHP') {
         // run the tests
      },
      stage('Test JS') {
        // run the tests
      }
   }
}
```
As part of this change, there is also a small change in how the JS dependencies are installed. Before, they were all installed in a separate stage. Now they are installed as part of the "Test JS" stage, right before running the tests for a extension.

### Removal of changeCivihrGitRemove()

As part of the site creation updates, a few bits of code have been removed or updated. 

The `changeCivihrGitRemote()` function has been completely removed. It was an attempt to handle PRs coming from forks, but it would not work as expect as the process is a bit more complex than simply changing the remote URL. For forks, we would need something more or less like this:

- Clone the original repository
- Checkout the enviroment branch
- Add the fork as a remote
- Checkout the branch from the fork
- Merge the enviroment branch from origin into the fork branch

Since 1) We don't really expect people to contribute with CiviHR at the moment, 2) The existing code was not working, 3) It's out of the scope of this PR to deal with forks, I decided to simply get rid of that.

### Fix for the enviroment branch checkout

This PR also includes changes to fix an issue introduced by https://github.com/civicrm/civihr/pull/2372. Before that PR, we would have this to decide which branch to use when creating a new site:

```groovy
def buildBranch = params.CIVIHR_BRANCH != '' ? params.CIVIHR_BRANCH : env.CHANGE_TARGET != null ? env.CHANGE_TARGET : env.BRANCH_NAME != null ? env.BRANCH_NAME : 'staging'
```

Which is something like this:

If we had a branch set in the params, use it. Otherwise, if `CHANGE_TARGET` is not empty (i.e. it's a PR), then use the `CHANGE_TARGET`. If that is empty, then use the `BRANCH_NAME` (this will happen when the build was triggered by a merge into either `master` or `staging` and `BRANCH_NAME` will  be the branch where the merge happened). Finally, if it can't everything else fails (I honestly don't know when that would happen) we use `staging`.

This branch would then be passed to the `civibuild` command to download the right version of CiviHR. With https://github.com/civicrm/civihr/pull/2372 however, this has changed and `staging` would always be used, even if the build had been triggered by a new merge to `master`.

This PR fixes this by defining the enviroment branch with this:

```groovy
def envBranch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
```

There's no way to manually set the branch via params (we never needed this in the first place). So now we only care about `CHANGE_TARGET` and `BRANCH_NAME`. I can't think of any circumstance where both would be empty, so there's no reason to handle this case.

The `envBranch` is then passed to the `cloneCiviHRRepositories()` function which will download the right version of code from the repositories.

### Removal of unnecessary class

The `HrJobRolesTestBase.php` class was not being used anywhere, so it was deleted. This is not related to this PR at all, but since it's just an unused file, I thought it ok to remove it here.


Jenkinsfile
```diff
@@ -46,132 +49,107 @@ pipeline {
...
             wget -O attachments.patch https://gist.githubusercontent.com/davialexandre/199b3ebb2c69f43c07dde0f51fb02c8b/raw/0f11edad8049c6edddd7f865c801ecba5fa4c052/attachments-4.7.27.patch
             patch -p1 -i attachments.patch
             rm attachments.patch
           """
 
-          // Change git remote of civihr ext to support dev version of Jenkins pipeline
-          changeCivihrGitRemote()
-
-          // Get repos & branch name
+          // Get target and PR branches name
```
mickadoo: Why was `changeCivihrGitRemote` removed?

Jenkinsfile
```diff
@@ -46,132 +49,107 @@ pipeline {
...
 
-    /* Testing PHP */
-    stage('Test PHP') {
-      steps {
-        script {
-          for (extension in listCivihrExtensions()) {
-            if (extension.hasPHPTests) {
-              testPHPUnit(extension)
+    stage('Run tests') {
+      parallel {
```
mickadoo: This is pretty nice and clear

Jenkinsfile
```diff
@@ -264,19 +242,14 @@ def getReportLink() {
...
-    cd $CIVICRM_EXT_ROOT/civihr
-    git remote set-url origin ${pulledCvhrRepo}
-    git fetch --all
-  """
+def cloneCiviHRRepositories(String envBranch) {
+  for (repo in listCivihrGitRepoPath()) {
+    sh """
+      git clone ${repo.url} ${repo.folder}
+      cd ${repo.folder}
+      git checkout $envBranch || true
```
mickadoo: Is the `|| true` here just because the branch might not exist?

Jenkinsfile
```diff
@@ -285,7 +258,7 @@ def checkoutPrBranchInCiviHRRepos(String branch) {
   for (repo in listCivihrGitRepoPath()) {
     try {
         sh """
-          cd ${repo}
+          cd ${repo.folder}
```
mickadoo: If `listCivihrGitRepoPath` returns CiviHR and T&A repos, but the PR branch only existing in CiviHR will `git checkout ${branch}` cause an error? I guess that's what the try catch block is for, but it's done differently above with `git checkout $envBranch || true`

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
```diff
@@ -1,130 +0,0 @@
-<?php
```
mickadoo: Can you add some note to the PR description for why this test was deleted

Jenkinsfile
```diff
@@ -46,132 +49,107 @@ pipeline {
...
           """
 
-          // Change git remote of civihr ext to support dev version of Jenkins pipeline
-          changeCivihrGitRemote()
-
-          // Get repos & branch name
+          // Get target and PR branches name
           def prBranch = env.CHANGE_BRANCH
-          def envBranch = env.CHANGE_TARGET
+          def envBranch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
```
mickadoo: Is this related to the performance improvements or why was it changed?

Jenkinsfile
davialexandre: yes

Jenkinsfile
davialexandre: Because it would not work as expected. That function existed as an attempt to support pull requests from forks, but it would not really work as it should. In cases like that, the process is a bit more complex (we need to merge code from two different remotes) and, since it's out of the scope of this PR, I decided to remove the code altogether. I'll add note of this to the PR description

Jenkinsfile
davialexandre: It is. This was necessary because now the civihr code is not downloaded by civibuild anymore. So we need a way to tell which base/enviroment branch to use. This also fix an issue introduced by https://github.com/civicrm/civihr/pull/2372 where `staging` would be used as a base branch to run the tests when something was merged to `master`. I'll add this to the PR description too

Jenkinsfile
davialexandre: You're right. The `try/catch` and the `|| true` are doing exactly the same thing. I'll update the file to use `|| true` here as well, as it's a bit more concise

com.civicrm.hrjobroles/tests/phpunit/HrJobRolesTestBase.php
davialexandre: Will do. I didn't do it before because I thought the [commit](https://github.com/civicrm/civihr/pull/2382/commits/66d41e97676cf87eb90b519f5e3fd456b782bbd4) message what descriptive enough

Jenkinsfile
mickadoo: Thanks, the description was helpful

Jenkinsfile
mickadoo: Thanks, it's a lot clearer now

PR #2387, opened 2018-01-02
PCHR-3020: Fix IE console errors for L&A

deb1990:
## Overview
When you try to approve or reject a leave on SSP using IE11 browser, the modal will just load but not save the leave. And there were errors in Console.
This PR fixes that bug.

## Before
b

## After
a

## Technical Details
1. Previously we were using native `find` method, which does not work on IE browsers, replaced the same with `_.find`.


bmnepali:
@debarshi did you check if there are other instances where native `find()` might have been used?

deb1990:
@bmnepali yes, there are no other instances in L&A

PR #2345, opened 2017-12-01
PCHR-3008: Implement AngularJS Style Guide and JS Semi-Standard across Job Contract app

AkA84:
## Overview
This PR makes the Job Contract angular app and its file to follow both the AngularJS Style Guide and the JS Semi-Standard Style

## Technical Details
Please refer to the technical details in #2066, as this PR does basically the same thing

## Comment
I began to migrate `$scope` to `vm`, but then I realized that so many things could go wrong with so many controllers, and the PR was already pretty huge. So I'll leave that to the dev who is going to work on those controllers next time

----

* [x] Tests pass

PR #2390, opened 2018-01-05
PCHR-3109: Fix Job Contract Modal error on save

deb1990:
## Overview
When staff has any contract (past, current or future one), it's not possible to create another one, endless spinner is displayed in Job Contract modal and JS error appears in console.
This PR fixes the bug.

## Before
b

## After
a

## Technical Details
Job Contract was refactored in https://github.com/civicrm/civihr/pull/2345. 
The last good commit was https://github.com/civicrm/civihr/pull/2345/commits/06ea92c568b84cf6d8e5f9d5e7568130fb65458c.
It got broken in https://github.com/civicrm/civihr/commit/628a36abfd88eac1163ff08444d3c1d94ec81ef8#diff-281ec0dce49dd1c53bf1e8333bc4f375L48

In `modal-contract-new.controller.js`
Before the refactoring, this was the flow of code

```javascript
$scope.entity = {};
...
angular.copy(model, $scope.entity);
$scope.entity.contract = {
	is_primary: 0
};
```

but then it got changed to 
```javascript
$scope.entity = {};
...
$scope.entity.contract = {
	is_primary: 0
};
...
angular.copy(model, $scope.entity);
```

So when we called angular.copy, it replaced `contract` key and we got an error on saveContract function(https://github.com/civicrm/civihr/blob/da7f2a8f13d443c5f5ff4c92ba782a4b84ba5854/hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js#L166).

The order has been changed to previous.


hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -316,6 +313,16 @@ define([
         $scope.entity.leave[index].add_public_holidays = utils.absenceTypes[absenceTypeID].add_public_holiday_to_entitlement;
       });
     }
+
+    /*
+     * Set the entity object
```
AkA84: it's not though, is it? First of all it's copy the entity into the model (so it's doing more things that it's saying), and then it's not setting the `entity` object, it's setting the `contract` property of `entity`

AkA84: honestly i don't know if this warrants a separate function at all. just put this two lines directly in the `init()` function

hrjobcontract/js/test/controllers/modal/modal-contract-new.controller.spec.js
```diff
@@ -69,6 +69,10 @@ define([
         Individual: 'Individual'
       };
 
+      it('sets the contract property', function () {
```
AkA84: it's not descriptive enough. sets to what?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
deb1990: Removed the function all together.

hrjobcontract/js/test/controllers/modal/modal-contract-new.controller.spec.js
deb1990: > @deb1990 add the first bad commit in the PR description, and also reference what file we are talking about, cause you're not mentioning it

Both has been Fixed.

deb1990:
>@deb1990 add the first bad commit in the PR description, and also reference what file we are talking about, cause you're not mentioning it

Both has been Fixed.

PR #2391, opened 2018-01-08
PCHR-3137: Fix actions popup

reneolivo:
## Overview
This PR fixes the actions popup for the search contacts results.

## Before
anim

## After
anim
The popup works in the search contacts results, work patterns, and scheduled jobs even after enabling/disabling records.

## Technical Details

The issue was introduced by #2180. The `.crm-entity` was assumed to be present as a parent of the `.btn-slide` button used for the popup, but this was not the case for the contact's search results.

All references to `$closestEntity` were removed since they were not needed any more. `$closestEntity` was introduced by #1292 as a way to map the events from the open popup to the row it belongs to, but #2180 changed the way this is achieved using the `mapCloneClickEventsToOrigin` method. ([here](https://github.com/civicrm/civihr/pull/2180/files#diff-5aa2eb87442af5119577aea46f98fadcR81)).

This PR also refactors the `civihr-popup.js` file so each function does what it says it does:

`createPopupClone` only creates the clone and adds the `.civihr-popup` class needed for styling:

```js
function createPopupClone () {
  $popupClone = $popup.clone(true)
    .appendTo($body)
    .addClass('civihr-popup');
}
```

`openPopupClone` only positions the popup where it's supposed to open, shows it, and adds the `.civihr-popup-open` class to the body:

```js
function openPopupClone () {
  var buttonOffset = $button.offset();

  $popupClone.css({
    left: +buttonOffset.left - ($popupClone.width() - $button.outerWidth()),
    top: +buttonOffset.top + $button.outerHeight()
  });
  $popupClone.show();
  $body.addClass('civihr-popup-open');
}
```

And the init function explains all the steps taken to display the popup and bind the events:

```js
(function init () {
  closePopupPanels();
  createPopupClone();
  openPopupClone();
  mapCloneClickEventsToOrigin();
  listenToMouseOutEvent();
})();
```

### Limiting the CiviHR's popup menu to #bootstrap-theme

The issues solved by the civihr-popup.js file should only target specific screens that have been migrated to Bootstrap. All other popup menus work well and are not hidden by their parent table, which was one of the reasons for this file:

find contacts hr17 9000
Menu inside a page that has not been migrated to Bootstrap. It works well without the help of the civihr-popup.js

## Comments

The `civihr-popup.js` file is not covered by unit or Backstopjs tests so it was tested manually.


hrui/js/src/civihr-popup/civihr-popup.js
```diff
@@ -7,7 +7,7 @@ Prevent Popups to overflow tables.
   'use strict';
   var $body = $('body');
 
-  $body.on('click', '.btn-slide', function () {
+  $body.on('click', '#bootstrap-theme .btn-slide', function () {
```
AkA84: Update the documentation
```js
/*
 Prevent Popups to overflow tables.
 */
```
to be more specific

PR #2392, opened 2018-01-09
PCHR-3130: Fix for Wrong Select Field values sent to backend in Absence Type Page

deb1990:
## Overview
It's not possible to disable carry forward per leave type once enabled. This PR fixes that bug.
The same bug was present for TOIL Expiry select field, that is fixed too.

## Before
**Carry Forward**
b

**TOIL**
toil_expiry

## After
**Carry Forward**
a

**TOIL**
toil_expirya

## Technical Details

This is a regression issue. In https://github.com/civicrm/civihr/pull/2300, we have added hidden fields to store the select values, but when we untick the select fields from code, the change event(https://github.com/civicrm/civihr/pull/2300/files#diff-80451403917186e0141a3b6700e2b299R357) does not get fired. So added the same in all places where we change the value of select from code.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
@@ -183,7 +183,7 @@
 
                 function hideToilExpiration() {
                   document.getElementById('accrual_expiration_duration').value = '';
-                  $('#accrual_expiration_unit').select2('val', '');
+                  $('#accrual_expiration_unit').select2('val', '').trigger("change");
                   $('.toil-expiration').hide();
                 }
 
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -183,7 +183,7 @@
 
                 function hideToilExpiration() {
                   document.getElementById('accrual_expiration_duration').value = '';
-                  $('#accrual_expiration_unit').select2('val', '');
```
davialexandre: Does this mean the problem would also happen if someone disabled TOIL for an Absence Type. If so, please mention that in the ticket description and tell QA to also check that while testing

davialexandre:
@deb1990 also, I think that the PR title needs to be updated. It's not describing what it does, but how. Not to mention that, after the latest updates the solution is not about clearing the field anymore

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
deb1990: True, added to the ticket.

deb1990:
> @deb1990 also, I think that the PR title needs to be updated. It's not describing what it does, but how. Not to mention that, after the latest updates the solution is not about clearing the field anymore

Fixed

PR #2260, opened 2017-10-31
PCHR-2699: Balance Change Calculation For Leave In Hours

tunbola:
## Overview
Since Leave Request Balance changes can now be calculated in hours, there is a need to re-work the logic for calculating balance changes which before only caters for Leave request calculated in days. This PR makes the necessary changes to make this possible.

## Before
The calculate balance logic does not support calculating balance changes for leave request in hours.

## After
The calculate balance change logic now supports calculating balance changes in days and in hours.

## Technical Details
- Two new classes `LeaveDateHoursAmountDeduction` and `LeaveDateDaysAmountDeduction` were added to help in calculating the amount to be deducted for a leave request date in hours and days respectively.
- For the start and end dates of a leave request in hours, two new fields were introduced to the LeaveRequest entity. `from_date_amount` and `to_date_amount` , these fields store the amount selected by the user requesting the leave for the start and end dates, since the user can modify the hours to be taken for these dates.
- `from_date` and `to_date` fields of the LeaveRequest table were modified to type `DATETIME` since these fields will be storing dates as well as the times for the leave request.
- The `from_date_type` and `to_date_type` fields on the Leave Request table are no longer required since these fields are not relevant to requesting leave in hours. Validations already exist to make them required when calculating leave request in days.
- For leave request in days the from_date time is set to `00:00` and the to_date time is set to `23:59`
- Two new classes `LeaveDaysBalanceChangeCalculation` and `LeaveHoursBalanceChangeCalculation` were added to assist the Balance Change service to determine the amount to be saved for each dates of the leave request taking the `from_date_amount` and `to_date_amount` fields into consideration for Leave in Hours requests.
- The `LeaveRequest.calculateBalanceChange` API now has a new `type_id` required field which enables the API to know the calculation unit of the absence type and what calculations to return. 

## Comments
- No new validations were added for the creating Leave Request in Hours except for validations for fields expected to be present/absent depending on the absence type calculation unit.
- The `validateEntitlementAndWorkingDayAndBalanceChange` was modified to take the `from_date_amount` and `to_date_amount` parameters into consideration when validating leave in hours.
- This PR adds basic tests to make sure that the functionality for calculating balance change in hours work and also ensures that previous tests were not broken, It is not an exhaustive test list and there might still be some issues that may arise which I believe will be fixed and tests added as these issues are fixed, Besides the task is huge already and I wanted to do the minimal work possible.
- The current validation for overlapping leave requests is left as is since same day leave requests is not allowed.

PR #2396, opened 2018-01-10
PCHR-3099: Make sure LeaveRequest.get and .getFull API will return requests ending on the last day of a period

davialexandre:
## Overview

Leave Requests ending on the last day of an Absence Period were not being displayed on the Absence Report and on the Leave Calendar.

## Before

before_3099

## After
 
after_3099

## Technical Details

This was a regression introduced by https://github.com/civicrm/civihr/pull/2260. In that PR, two changes related to the `from_date` and `to_date` fields were made in order to support Leave Request in hours:

- Their type [changed from date to datetime](https://github.com/civicrm/civihr/pull/2260/files#diff-0c81d1d212efc488e009f00d2cad8678R338)
- The LeaveRequest.create API was changed ([here](https://github.com/civicrm/civihr/pull/2260/files#diff-7bf537a036f3438a73ec57bd207d87daR37) and [here](https://github.com/civicrm/civihr/pull/2260/files#diff-7bf537a036f3438a73ec57bd207d87daR784)) to automatically set the hour in case it was not present (i.e. it's a Leave Request in days rather than hours). For the `from_date` field, the hour is set to `00:00:00` and for `to_date` it is set to `23:59:59`

The `LeaveRequest.get` and `getFull` APIs, however, were not updated to follow this change. So, imagine a call like this is sent:

```php
civicrm_api3('LeaveRequest', 'get', [
  'to_date' => ['<=' => '2018-12-31'],
]);
```

When that happens, civi will check that this is a datetime field, will accept the value (since it's a valid date, even though it doesn't have the hour) and convert it to a date time, which will result in a value like this: `20181231000000`. This will then result in a query like `WHERE to_date <= '20181231000000'` and, since during the request creation we automatically set the to_date hour to `23:59:00`, it will not find requests ending on that date.

In order to fix that, an API Wrapper was added to handle the API params and make sure they will always include hours. The logic is the same as the one used for when a Leave Request is created. For `from_date`, we set the hour to `00:00:00` and for `to_date` we set it to `23:59:59`. The decision to use an API wrapper for this is that we need to handle these values before they pass through the validation which converts the value to a datetime. The wrapper is the only way to do that, since the wrappers are always called before the validation by Civi.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
@@ -1,0 +1,161 @@
+<?php
+
+class CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDates implements API_Wrapper {
+
+  /**
+   * Adds Hours to the from_date and to_date fields, in case they're present but
+   * don't have that information.
+   *
+   * For the from_date field, the hour will be set to 00:00:00. For the to_date
+   * field, it will be set to 23:59:59
+   *
+   * @param array $apiRequest
+   *
+   * @return array
+   *   modified $apiRequest
+   */
+  public function fromApiInput($apiRequest) {
+    if ($this->canHandleTheRequest($apiRequest)) {
+      $this->prepareDates($apiRequest);
+    }
+
+    return $apiRequest;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function toApiOutput($apiRequest, $result) {
+    return $result;
+  }
+
+  /**
+   * Returns whether this Wrapper can handle the given $apiRequest or not. It
+   * can only handle the `get` and `getFull` actions of the LeaveRequest API.
+   *
+   * @param array $apiRequest
+   *
+   * @return bool
+   */
+  private function canHandleTheRequest($apiRequest) {
+    $targetActions = ['get', 'getfull'];
+    $isTargetAction = in_array($apiRequest['action'], $targetActions);
+
+    return $apiRequest['entity'] === 'LeaveRequest' && $isTargetAction;
+  }
+
+  /**
+   * If the query contains values for the from_date and to_date fields, this
+   * method makes sure they will always contain information about the hour.
+   *
+   * For the from_date field, it will set the hour as 00:00:00 if it's not
+   * present. For the to_date field, it will set it as 23:59:59 if it's not
+   * present.
+   *
+   * These are datetime fields, and the reason for this is that if there's a
+   * query like "to_date <= 2018-01-09" it will include all the leave requests
+   * ending on that date, regardless of the hour. Without this, such a query
+   * would be interpreted as "to_date <= 2018-01-09 00:00:00" by the database,
+   * and most of the requests ending on 2018-01-09 would not be included in the
+   * result.
+   *
+   * @param array $apiRequest
+   */
+  private function prepareDates(&$apiRequest) {
+    $params = &$apiRequest['params'];
+
+    $fromDate = $this->getValueFromParams($params, 'from_date');
+
+    if ($this->dateIsValid($fromDate) && $this->dateIsMissingHours($fromDate)) {
+      $fromDate = (new DateTime($fromDate))->format('Y-m-d 00:00:00');
+      $this->setValueOfParam($params, 'from_date', $fromDate);
+    }
+
+    $toDate = $this->getValueFromParams($params, 'to_date');
+
+    if ($this->dateIsValid($toDate) && $this->dateIsMissingHours($toDate)) {
+      $toDate = (new DateTime($toDate))->format('Y-m-d 23:59:59');
+      $this->setValueOfParam($params, 'to_date', $toDate);
+    }
+  }
+
+  /**
+   * Check whether the given value is a valid date.
+   *
+   * @param string $date
+   *
+   * @return bool
+   */
+  private function dateIsValid($date) {
+    $date = date_parse($date);
+
+    return $date['error_count'] === 0 && $date['warning_count'] === 0;
+  }
+
+  /**
+   * Checks if the given $date string contains information about hours
+   *
+   * @param string $date
+   *  A string in any of the formats supported by strtotime()
+   *
+   * @return bool
+   */
+  private function dateIsMissingHours($date) {
+    $date = date_parse($date);
+
+    return $date['hour'] === FALSE ||
+           $date['minute'] === FALSE ||
+           $date['second'] === FALSE;
+  }
+
+  /**
+   * Gets the value of the given $field in the $params array.
+   *
+   * This method works for both simple params ($field => $value) and params with
+   * operators ($field => [$operator => $value])
+   *
+   * @param array $params
+   * @param string $field
+   *
+   * @return mixed|null
+   */
+  private function getValueFromParams($params, $field) {
+    if (empty($params[$field])) {
+      return NULL;
+    }
+
+    // When using other operators than =
+    // the param will be an array in this format:
+    // [<operator> => <value>]
+    if (is_array($params[$field])) {
+      return reset($params[$field]);
+    }
+
+    return $params[$field];
+  }
+
+  /**
+   * Sets the value of the given $field in $params to the given $value.
+   *
+   * This method works for both simple params ($field => $value) and params with
+   * operators ($field => [$operator => $value]).
+   *
+   * @param array $params
+   * @param string $field
+   * @param mixed $value
+   */
+  private function setValueOfParam(&$params, $field, $value) {
+    if (!array_key_exists($field, $params)) {
+      return;
+    }
+
+    if (is_array($params[$field])) {
+      $key = key($params[$field]);
+      $params[$field][$key] = $value;
+    }
+    else {
+      $params[$field] = $value;
+    }
+  }
+
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
@@ -1,0 +1,108 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDates as LeaveRequestDates;
+
+/**
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDatesTest extends BaseHeadlessTest {
+
+  /**
+   * @var LeaveRequestDates
+   */
+  private $wrapper;
+
+  public function setUp() {
+    $this->wrapper = new LeaveRequestDates();
+  }
+
+  public function testRequestParamsAreNotUpdatedIfTheRequestIsNotForLeaveRequestGetOrGetFull() {
+    $apiRequest = [
+      'entity' => 'LeaveRequest',
+      //some random action name, so that we have different actions for different tests
+      'action' => 'get' . rand(10, 1000),
+      'params' => [
+        'from_date' => '2016-01-01',
+        'to_date' => '2016-01-01'
+      ]
+    ];
+
+    $wrappedRequest = $this->wrapper->fromApiInput($apiRequest);
+
+    $this->assertEquals($apiRequest, $wrappedRequest);
+  }
+
+  /**
+   * @dataProvider supportedActions
+   */
+  public function testToAndFromDateHoursAreSetIfTheGivenDatesDontHaveHour($action) {
+    $apiRequest = [
+      'entity' => 'LeaveRequest',
+      'action' => $action,
+      'params' => [
+        'from_date' => '2016-02-01',
+        'to_date' => '2016-02-03'
+      ]
+    ];
+
+    $wrappedRequest = $this->wrapper->fromApiInput($apiRequest);
+
+    $this->assertEquals('2016-02-01 00:00:00', $wrappedRequest['params']['from_date']);
+    $this->assertEquals('2016-02-03 23:59:59', $wrappedRequest['params']['to_date']);
+  }
+
+  /**
+   * @dataProvider supportedActions
+   */
+  public function testToAndFromDateHoursAreSetIfTheGivenDatesDontHaveHourAndTheParamsHaveOperators($action) {
+    $apiRequest = [
+      'entity' => 'LeaveRequest',
+      'action' => $action,
+      'params' => [
+        'from_date' => ['>' => '2016-02-01'],
+        'to_date' => ['<=' => '2016-02-03']
+      ]
+    ];
+
+    $wrappedRequest = $this->wrapper->fromApiInput($apiRequest);
+
+    $this->assertEquals(['>' => '2016-02-01 00:00:00'], $wrappedRequest['params']['from_date']);
+    $this->assertEquals(['<=' => '2016-02-03 23:59:59'], $wrappedRequest['params']['to_date']);
+  }
+
+  /**
+   * @dataProvider supportedActions
+   */
+  public function testToAndFromDateHoursAreNotTouchedIfTheGivenDatesAlreadyHaveHours($action) {
+    $apiRequest = [
+      'entity' => 'LeaveRequest',
+      'action' => $action,
+      'params' => [
+        'from_date' => '2016-02-01 11:53:00',
+        'to_date' => '2016-02-03 17:15:10'
+      ]
+    ];
+
+    $wrappedRequest = $this->wrapper->fromApiInput($apiRequest);
+
+    $this->assertEquals('2016-02-01 11:53:00', $wrappedRequest['params']['from_date']);
+    $this->assertEquals('2016-02-03 17:15:10', $wrappedRequest['params']['to_date']);
+  }
+
+  public function testTheOutputIsNeverChanged() {
+    $result = [
+      'test' => '1234'
+    ];
+
+    $wrappedResult = $this->wrapper->toApiOutput([], $result);
+
+    $this->assertEquals($result, $wrappedResult);
+  }
+
+  public function supportedActions() {
+    return [
+      ['get'],
+      ['getfull']
+    ];
+  }
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
@@ -3898,6 +3898,147 @@ class api_v3_LeaveRequestTest extends BaseHeadlessTest {
     $this->assertEquals($contact2['id'], $result['values'][1]['contact_id']);
   }
 
+  public function testGetAndGetFullIncludesAllTheRequestsUpToTheLastDayMinuteIfToDateIsPresentAndHourIsEmpty() {
+    $contract1 = HRJobContractFabricator::fabricate(
+      [ 'contact_id' => 1 ],
+      [ 'period_start_date' => '2016-01-01' ]
+    );
+
+    $contract2 = HRJobContractFabricator::fabricate(
+      [ 'contact_id' => 2 ],
+      [ 'period_start_date' => '2016-01-01' ]
+    );
+
+    LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $contract1['contact_id'],
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-02-20'), //Hour will be automatically set to 00:00:00
+      'to_date' =>  CRM_Utils_Date::processDate('2016-02-23'), //Hour will be automatically set to 23:59:59
+    ], true);
+
+    LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $contract2['contact_id'],
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-02-23 10:00:00'),
+      'to_date' =>  CRM_Utils_Date::processDate('2016-02-23 16:00:00'),
+    ], true);
+
+    $result = civicrm_api3('LeaveRequest', 'get', [
+      'to_date' => ['<=' => '2016-02-23'],
+      'sequential' => 1
+    ]);
+    $this->assertEquals(2, $result['count']);
+    $this->assertEquals($contract1['contact_id'], $result['values'][0]['contact_id']);
+    $this->assertEquals($contract2['contact_id'], $result['values'][1]['contact_id']);
+
+    $result = civicrm_api3('LeaveRequest', 'getFull', [
+      'to_date' => ['<=' => '2016-02-23'],
+      'sequential' => 1
+    ]);
+    $this->assertEquals(2, $result['count']);
+    $this->assertEquals($contract1['contact_id'], $result['values'][0]['contact_id']);
+    $this->assertEquals($contract2['contact_id'], $result['values'][1]['contact_id']);
+  }
+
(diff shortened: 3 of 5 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function toApiOutput($apiRequest, $result) {
+    return $result;
+  }
+
+  private function canHandleTheRequest($apiRequest) {
+    return $apiRequest['entity'] === 'LeaveRequest' &&
```
mickadoo: This condition looks a bit strangely indented, plus it's over 80 characters long, how about splitting it into 

```
    $targetActions = ['get', 'getfull'];
    $isTargetAction = in_array($apiRequest['action'], $targetActions);
    
    return $apiRequest['entity'] === 'LeaveRequest' && $isTargetAction;
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   * result.
+   *
+   * @param array $apiRequest
+   */
+  private function prepareDates(&$apiRequest) {
+    $params = &$apiRequest['params'];
+
+    $fromDate = $this->getValueFromParams($params, 'from_date');
+
+    if($fromDate && $this->dateIsValid($fromDate) && $this->dateIsMissingHours($fromDate)) {
```
mickadoo: Line length is over 80. You could also change `dateIsValid` to return `false` if $fromDate is `null`, avoiding the extra condition here.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      $fromDate = new DateTime($fromDate);
```
mickadoo: You could do the `format` on this line since the `\DateTime` isn't needed elsewhere and the following line is a bit long.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      $this->setValueOfParam($params,'from_date', $fromDate->format('Y-m-d 00:00:00'));
+    }
+
+    $toDate = $this->getValueFromParams($params, 'to_date');
+
+    if($toDate && $this->dateIsValid($toDate) && $this->dateIsMissingHours($toDate)) {
```
mickadoo: Same comments as for `$fromDate`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      $toDate = new DateTime($toDate);
+      $this->setValueOfParam($params, 'to_date', $toDate->format('Y-m-d 23:59:59'));
+    }
+  }
+
+  private function dateIsValid($date) {
```
mickadoo: Missing docblock, also for functions returning booleans I think it's common to follow the format `is__`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+    $date = date_parse($date);
```
mickadoo: Didn't know about this, nice

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * This method works for both simple params ($field => $value) and params with
+   * operators ($field => [$operator => $value])
+   *
+   * @param array $params
+   * @param string $field
+   *
+   * @return mixed|null
+   */
+  private function getValueFromParams(&$params, $field) {
```
mickadoo: Why is `$params` passed by reference? It seems the only possible reason is the `reset` call which resets the array and returns the first element, but I don't see why you'd want to reset the real `$params` array here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+    if(empty($params[$field])) {
+      return null;
```
mickadoo: Boolean lowercase

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+    // [<operator> => <value>]
+    if(is_array($params[$field])) {
+      return reset($params[$field]);
+    }
+
+    return $params[$field];
+  }
+
+  /**
+   * Sets the value of the given $field to the given $value in this object's
```
mickadoo: > in this object's internal $params array.

This sounds like `$params` is a property of this class.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   * @param array $params
+   * @param string $field
+   * @param mixed $value
+   */
+  private function setValueOfParam(&$params, $field, $value) {
+    if(!array_key_exists($field, $params)) {
+      return;
+    }
+
+    if(is_array($params[$field])) {
```
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      $key = key($params[$field]);
```
mickadoo: Nice and clean

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   * For the from_date field, the hour will be set to 00:00:00. For the to_date
+   * field, it will be set to 23:59:59
+   *
+   * @param array $apiRequest
+   *
+   * @return array
+   *   modified $apiRequest
+   */
+  public function fromApiInput($apiRequest) {
+    if($this->canHandleTheRequest($apiRequest)) {
```
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @param string $date
+   *  A string in any of the formats supported by strtotime()
+   *
+   * @return bool
+   */
+  private function dateIsMissingHours($date) {
+    $date = date_parse($date);
+
+    return $date['hour'] === false ||
```
mickadoo: Boolean case

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   * This method works for both simple params ($field => $value) and params with
+   * operators ($field => [$operator => $value])
+   *
+   * @param array $params
+   * @param string $field
+   *
+   * @return mixed|null
+   */
+  private function getValueFromParams(&$params, $field) {
+    if(empty($params[$field])) {
```
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      return null;
+    }
+
+    // When using other operators than =
+    // the param will be an array in this format:
+    // [<operator> => <value>]
+    if(is_array($params[$field])) {
```
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * This method works for both simple params ($field => $value) and params with
+   * operators ($field => [$operator => $value]).
+   *
+   * @param array $params
+   * @param string $field
+   * @param mixed $value
+   */
+  private function setValueOfParam(&$params, $field, $value) {
+    if(!array_key_exists($field, $params)) {
```
mickadoo: Missing space after `if`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      return;
+    }
+
+    if(is_array($params[$field])) {
+      $key = key($params[$field]);
+      $params[$field][$key] = $value;
+    } else {
```
mickadoo: New line for `else`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,143 @@
...
+      $params[$field] = $value;
+    }
+  }
```
mickadoo: The closing brace for the class must have an empty line

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDates as LeaveRequestDates;
+
+/**
+ * Class CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDates
```
mickadoo: Is this line intentional?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @group headless
+ */
+class CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDatesTest extends BaseHeadlessTest {
+
+  /**
+   * @var \CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestDates
```
mickadoo: You can use the alias

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
```diff
@@ -0,0 +1,109 @@
...
+   */
+  private $wrapper;
+
+  public function setUp() {
```
mickadoo: Just want to check, when we say description for all methods, I guess tests are excluded

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
```diff
@@ -0,0 +1,109 @@
...
+      //some random action name, so that we have different actions for different tests
+      'action' => 'get' . rand(10, 1000),
+      'params' => [
+        'test' => '123'
+      ]
+    ];
+
+    $wrappedRequest = $this->wrapper->fromApiInput($apiRequest);
+
+    $this->assertEquals($apiRequest, $wrappedRequest);
```
mickadoo: Would the `$apiRequest` be different even if the `canHandleTheRequest` check passed?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -3898,6 +3898,147 @@ public function testGetAndGetFullReturnsAllDataWhenLoggedInUserHasEditAllContact
...
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-02-20'), //Hour will be automatically set to 00:00:00
+      'to_date' =>  CRM_Utils_Date::processDate('2016-02-20'), //Hour will be automatically set to 23:59:59
+    ], true);
+
+    LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $contract2['contact_id'],
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-02-20 23:59:59'),
+      'to_date' =>  CRM_Utils_Date::processDate('2016-02-23 16:00:00'), //Hour will be automatically set to 23:59:59
```
mickadoo: Why will the hour automatically be changed in `to_date`?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -3898,6 +3898,147 @@ public function testGetAndGetFullReturnsAllDataWhenLoggedInUserHasEditAllContact
...
+      'from_date' => CRM_Utils_Date::processDate('2016-03-25 17:00:00'),
+      'to_date' =>  CRM_Utils_Date::processDate('2016-03-25 18:00:00'),
+    ], true);
+
+    LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $contract['contact_id'],
+      'type_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2016-03-25 19:00:00'),
+      'to_date' =>  CRM_Utils_Date::processDate('2016-03-25 20:00:00'),
+    ], true);
```
mickadoo: `civilint` shows a lot of errors in this class, there are some news ones introduced by this PR such as the lowercase booleans. Do you think it's too much to fix them all in this PR?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
davialexandre: I was sure aliases were not supported in docblocks, but I was just checking [here](https://docs.phpdoc.org/references/phpdoc/types.html#valid-class-name) and it seems they are (assuming that an alias is a "local name"):

> Thus this may be either a Fully Qualified Class Name (FQCN) or if present in a namespace a local name.

I'll update

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
davialexandre: Yeah, they are

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
davialexandre: Good point. I guess I can change the params to include the fields handled by the wrapper to make this a little bit better

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: It won't. This is a copy and paste mistake. Good catch

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: Yup. I'm sure civilint will complain about pretty much every file in civihr. I would rather fix this only when we finally add style checks to Jenkins (it's in the plans) to avoid having regressions in case people forget to run it manually. When that happens we'll probably have one or a few PRs just for code style fixes.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
davialexandre: Good catch. These were where originally part of `LeaveRequestSelect` and I forgot to update the comment after I moved them here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
davialexandre: You're right. There's no need for this here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
davialexandre: I think it's just a matter of what reads better. If `$date` was an object, then I think that an `isValid()` method would be better, since we would write something like `if $date->isValid()`. That is, we are asking the date if it is valid. Here, however, we asking someone else to check if the date is valid, so I think this `if $this->dateIsValid($date)` reads better. But I don't mind changing if you think that the `is_` version is better

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
davialexandre: Looks like `date_parse` will set the error count to 1 if it receives `null`, so there's no need to change anything there. I'm dropping the extra condition

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDatesTest.php
davialexandre: It was added automatically by PhpStorm and I didn't bother to remove it. Removing it now

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
mickadoo: Thumbs up but no change? :smile:

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
mickadoo: Nah, it's fine. It's just as readable both ways

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
```diff
@@ -0,0 +1,161 @@
...
+   * @param array $apiRequest
+   */
+  private function prepareDates(&$apiRequest) {
+    $params = &$apiRequest['params'];
+
+    $fromDate = $this->getValueFromParams($params, 'from_date');
+
+    if ($this->dateIsValid($fromDate) && $this->dateIsMissingHours($fromDate)) {
+      $fromDate = (new DateTime($fromDate))->format('Y-m-d 00:00:00');
+      $this->setValueOfParam($params,'from_date', $fromDate);
```
mickadoo: space after `$params`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestDates.php
davialexandre: oops

PR #2399, opened 2018-01-11
PCHR-3073: Add CMS Data related classes

tunbola:
## Overview
This PR adds some generic CMS related classes and interfaces to the HRCore extension that can be used by other extension to display and perform CMS framework related actions.

## Before
CMS User interfaces related to Roles and Mail notification were not present.

## After
The following interfaces were added in this PR:
- CRM_HRCore_CMSData_UserMailNotifierInterface: This interface provides methods for notifying the CMS user mainly for sending password reset email and welcome email functionality and can be extended by relevant CMS framework class
- CRM_HRCore_CMSData_UserRoleInterface: This interface provides information about the roles of a CMS user.

The following classes were added to implement the Interfaces
- The CRM_HRCore_CMSData_UserRole_Drupal and the CRM_HRCore_CMSData_UserMailNotifier_Drupal class was added to implement the corresponding interface.

The following factory classes were added:
- The CRM_HRCore_CMSData_UserMailNotifierFactory and CRM_HRCore_CMSData_UserRoleFactory.

## Comments
- These classes were added primarily to aid the task being done for https://compucorp.atlassian.net/browse/PCHR-3073

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
@@ -1,0 +1,37 @@
+<?php
+
+use CRM_HRCore_CMSData_UserMailNotifierInterface as UserMailNotifierInterface;
+
+/**
+ * Class CRM_HRCore_CMSData_UserMailNotifier_Drupal
+ */
+class CRM_HRCore_CMSData_UserMailNotifier_Drupal implements UserMailNotifierInterface {
+
+  /**
+   * @var stdClass
+   */
+  private $user;
+
+  /**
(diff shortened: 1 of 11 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
```diff
@@ -0,0 +1,37 @@
...
+   */
+  public function __construct($contactData) {
+    $this->user = user_load($contactData['cmsId']);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function sendWelcomeEmail() {
+   return _user_mail_notify('register_admin_created', $this->user);
```
davialexandre: I'm not sure if we should be calling this function. Usually, functions starting with an underscore are considered private. Isn't there any other alternatives? Maybe you could check with a Drupal dev if it's ok to call this function and/or if there are alternatives

PR #2401, opened 2018-01-12
PCHR-3073: Modify UserNotifier Interface and related classes.

tunbola:
## Overview
This PR is a follow up to [#2399](https://github.com/civicrm/civihr/pull/2399)
It modifies the UserNotifier Interface so that a single instance of the class can be used to send notifications to multiple users, Also the UserInterface Factory now gets the CmsFramework parameter from inside the create method.

## Before
- A new instance of the UserNotifier class would have to be created per user to send notification to multiple users.

## After
- A new getUser method was added to the UserMailNotifierInterface class and the constructor method was removed.
- The create method of the UserMailNotifierFactory was modified to get the userFramework within the method, This is to allow the UserMailNotifier class to be easily made into a civi container that can be called from any extension. There were some issues in passing the userFramework into the create method using civi containers as the `Symfony\Component\ExpressionLanguage\Expression` is not bundled with civi core and its a requirement to add expression in the container.xml file.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
@@ -8,25 +8,29 @@ use CRM_HRCore_CMSData_UserMailNotifierInterface as UserMailNotifierInterface;
 class CRM_HRCore_CMSData_UserMailNotifier_Drupal implements UserMailNotifierInterface {
 
   /**
-   *{@inheritdoc}
+   * Gets the Drupal user object
+   *
+   * @param array $contactData
    *
    * @return \stdClass
    */
-  public function getUser($contactData) {
+  private function getUser($contactData) {
     return user_load($contactData['cmsId']);
   }
 
   /**
    * {@inheritdoc}
    */
-  public function sendWelcomeEmail($user) {
-   return _user_mail_notify('register_admin_created', $user);
+  public function sendWelcomeEmail($contactData) {
+    $user = $this->getUser($contactData);
+    return _user_mail_notify('register_admin_created', $user);
   }
 
   /**
    * {@inheritdoc}
    */
-  public function sendPasswordResetEmail($user) {
+  public function sendPasswordResetEmail($contactData) {
+    $user = $this->getUser($contactData);
     return _user_mail_notify('password_reset', $user);
   }
 }
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
@@ -8,30 +8,25 @@ use CRM_HRCore_CMSData_UserMailNotifierInterface as UserMailNotifierInterface;
 class CRM_HRCore_CMSData_UserMailNotifier_Drupal implements UserMailNotifierInterface {
 
   /**
-   * @var stdClass
-   */
-  private $user;
-
-  /**
-   * CRM_HRCore_CMSData_UserMailNotifier_Drupal constructor.
+   *{@inheritdoc}
    *
-   * @param array $contactData
+   * @return \stdClass
    */
-  public function __construct($contactData) {
-    $this->user = user_load($contactData['cmsId']);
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/UserMailNotifierFactoryTest.php
```diff
@@ -11,17 +11,22 @@
 class CRM_HRCore_CMSData_CMSUserMailNotifierFactoryTest extends CRM_HRCore_Test_BaseHeadlessTest {
 
   public function testItReturnsAnInstanceOfTheExpectedClassWhenCMSIsSupported() {
-    $contactData = ['cmsId' => 1];
-    $mailNotifier = UserMailNotifierFactory::create('Drupal', $contactData);
+    //We need to manually set this value because civicrm sets it to UnitTests
+    //when running in testing environment.
+    CRM_Core_Config::singleton()->userFramework = 'Drupal';
```
davialexandre: Don't you need to unset this afterward, to avoid problems with tests that expect this to be "UnitTests"?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/UserMailNotifierFactoryTest.php
```diff
@@ -11,17 +11,22 @@
...
+    $mailNotifier = UserMailNotifierFactory::create();
     $this->assertInstanceOf(UserMailNotifierInterface::class, $mailNotifier);
   }
 
   public function testItThrowsAnExceptionWhenCMSIsNotSupported() {
+    //We need to manually set this value because civicrm sets it to UnitTests
+    //when running in testing environment.
     $cmsFramework = 'UnrecognizedCMS';
+    CRM_Core_Config::singleton()->userFramework = $cmsFramework;
```
davialexandre: Same thing here. Don't you need to reset it?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserMailNotifier/Drupal.php
```diff
@@ -8,30 +8,25 @@
...
-  /**
-   * CRM_HRCore_CMSData_UserMailNotifier_Drupal constructor.
+   *{@inheritdoc}
    *
-   * @param array $contactData
+   * @return \stdClass
    */
-  public function __construct($contactData) {
-    $this->user = user_load($contactData['cmsId']);
+  public function getUser($contactData) {
```
davialexandre: Does this method really need to be public? What I suggested (sorry if I wasn't clear) was something more or less like this:

```php
public function sendWelcomeEmail($contactData) {
  $user = $this->getUser($contactData);
  return _user_mail_notify('register_admin_created', $user);
}
``` 

This way, `getUser()` will only be used internally, so there's no need for it to be public

davialexandre:
@tunbola Another thing: please add the "ready to test" label

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/UserMailNotifierFactoryTest.php
```diff
@@ -11,17 +11,24 @@
...
 
   public function testItReturnsAnInstanceOfTheExpectedClassWhenCMSIsSupported() {
-    $contactData = ['cmsId' => 1];
-    $mailNotifier = UserMailNotifierFactory::create('Drupal', $contactData);
+    //We need to manually set this value because civicrm sets it to UnitTests
+    //when running in testing environment.
+    CRM_Core_Config::singleton()->userFramework = 'Drupal';
+    $mailNotifier = UserMailNotifierFactory::create();
     $this->assertInstanceOf(UserMailNotifierInterface::class, $mailNotifier);
+    CRM_Core_Config::singleton()->userFramework = 'UnitTests';
```
davialexandre: I think a better approach for this would be:

```php

$previousUserFramework = CRM_Core_Config::singleton()->userFramework = 'UnitTests';
CRM_Core_Config::singleton()->userFramework = 'Drupal';
// test steps
CRM_Core_Config::singleton()->userFramework = $previousUserFramework;
```

This way, we avoid problems if in the future they change `UnitTests` to something else

PR #2402, opened 2018-01-15
PCHR-3091: Fix for AM PM not displayed in Calendar

deb1990:
## Overview
When staff/manager/admin creates multiple days request in calendar and selects half-day for end date (AM/PM), calendar cells are displayed properly, but after refreshing page AM/PM is no more displayed for end date in calendar.
This PR fixes the bug.

## Before
b

## After
a

## Technical Details
This is a regression issue. 

### Problem
When we add a new Request, this is how the dates in the request object looks like

```
from_date:"2018-01-24"
to_date:"2018-01-25"
```

but when it is returned by the API, it looks like the following,
```
from_date:"2018-01-24 00:00:00"
to_date:"2018-01-25 23:59:00"
```

It was introduced in this PR https://github.com/civicrm/civihr/pull/2260. Where we changed the data type 
> from_date and to_date fields of the LeaveRequest table were modified to type DATETIME since these fields will be storing dates as well as the times for the leave request.

So because the date format is changed, this line of code https://github.com/civicrm/civihr/blob/staging/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js#L248 does not return true. 

Changing the code to the following works

```
moment(date).isSame(leaveRequest.to_date, 'day')
```

This was not caught by the unit test, because the Mock Data did not have data in DATETIME format, so updated `leave-request-data.js` and fixed unit tests.

Updated all `data` files in L&A to `*.data.js`, also renamed `option-group-mock-data.js` to `option-group.data.js`.


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/mocks/data/leave-request-data.js
```diff
@@ -16,9 +16,9 @@ define([
         'type_id': absenceTypeData.all().values[0]['id'],
         'contact_id': '202',
         'status_id': optionGroupMock.specificValue('hrleaveandabsences_leave_request_status', 'value', '1'),
-        'from_date': '2016-02-02',
```
AkA84: this file should be `.data.js`, please refer to the style guide (and fix any other data file if needed)

deb1990: Fixed

PR #1051, opened 2016-08-05
PCHR-1294: Add the scheduled job to expire leave balance changes

davialexandre:
This scheduled job will run daily, checking for balance changes with an
expiry date in the past and creating the respective expiration record.

Example:
Imagine you have 3 days Brought Forward that will expire on 2016-04-01. This is how the balance change table will look like:

| id | amount | expiry_date | expired_balance_id |
|----|--------|-------------|--------------------|
| 1 | 3 | 2016-04-01 | NULL |

Now, after running the scheduled job, a new record will be created with the amount of days expired as a negative number. On this example, since no leave was taken, all the 3 days will expire:

| id | amount | expiry_date | expired_balance_id |
|----|--------|-------------|--------------------|
| 1 | 3 | 2016-04-01 | NULL |
| 2 | -3 | 2016-04-01 | 1 |

PR #2406, opened 2018-01-15
PCHR-3098: Remove TOIL Accrue from "Used in Previous Period" on the Entitlement Calculation

davialexandre:
## Overview

On the entitlement calculation page, there are two separate columns to display the number of days accrued in the previous period (the "Previous Period Accrued TOIL" column) and the number of days taken/used as leave for the previous period (The "Used in Previous Period" column). However, the value displayed as the number of days used as leave request was also including the number of days accrued.

So, in a scenario where we had 3 days accrued and 1 days used as leave, the "Previous Period Accrued TOIL" column would display 3, as expected, but the "Used In Previous Period" column should display would display 2 (3 - 1 day used). rather than 1.

## Before

The number of accrued days would also be included in the number of days used in the previous period:

pchr_3098_wrong

## After

The number of accrued days is not included in the number of days used in the previous period anymore:

pchr_3098_correct

## Technical Details
Both TOIL and Leave Requests are stored in the Leave Request table. We differentiate them by the `request_type` field. For TOIL Requests, the value of this field is `toil`, whereas for Leave Requests it is `leave`. The [method that builds the query](https://github.com/civicrm/civihr/blob/staging/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php#L255-L334) to get the leave request balance was getting all the Leave Balance Changes linked to all the Leave Requests, regardless of the request type (Although there is an option to filter out Public Holidays) and this was the reason the accrued days were included. To fix that, I added a new `$excludeToilRequests` param to that method. Its default value is `false`, so it will continue working as before, but, in the `LeavePeriodEntitlement.getLeaveRequestBalance()` method, which is used to get the number of days used in the previous period, we now set it to true.

## Other changes
The commit https://github.com/civicrm/civihr/pull/2406/commits/8dbaf47273293b23e169ed337cd10258af6f21f7 was a quick cleanup to remove some unused aliases.

The commit https://github.com/civicrm/civihr/pull/2406/commits/63051c58256cec24d123cc96c3b811ea07af1711 was another cleanup, now to remove two unused params from the `LeaveBalanceChange::getLeaveRequestBalanceForEntitlement()` method. These params were originally introduced by https://github.com/civicrm/civihr/pull/1051, but later, in  https://github.com/civicrm/civihr/pull/1359, when the logic became more complex in order to be able to also expire TOIL Requests, they have become unnecessary and were not being used anywhere else.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -241,24 +237,21 @@ public static function getBreakdownBalanceForEntitlement($entitlementID) {
...
-   * @param \DateTime $dateLimit
-   *   When given, will make the method count only days taken as leave up to this date
-   * @param \DateTime $dateStart
-   *   When given, will make the method count only days taken as leave starting from this date
    * @param bool $excludePublicHolidays
    *   When true, it won't sum the balance changes for Public Holiday Leave Requests
    * @param bool $includePublicHolidaysOnly
    *   When true, it won't sum only the balance changes for Public Holiday Leave Requests
+   * @param bool $excludeToilRequests
+   *   When true, it won't sum sum the balance changed for TOIL Requests
```
tunbola: s/sum sum/sum

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/EntitlementCalculationTest.php
```diff
@@ -673,6 +672,51 @@ public function testAmountUsedInPreviousPeriodShouldBeTheTotalAmountFromAllAppro
...
+      'request_type' => LeaveRequest::REQUEST_TYPE_TOIL
+    ], true);
+
+    $currentPeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => date('YmdHis', strtotime('2016-01-01')),
+      'end_date' => date('YmdHis', strtotime('2016-12-31')),
+    ], true);
+
+    $calculation = new EntitlementCalculation($currentPeriod, $this->contact, $type);
+    // It should only be 1, as it will not include the 2 days accrued
```
tunbola: s/the 2 days accrued/the 3 days accrued

PR #2409, opened 2018-01-16
PCHR-2588: QA Fixes

mickadoo:
## Overview

After going over the QA plan myself locally I noticed some things that could be improved. This PR fixes those.

## Before

- Roles that had no login date were excluded
- Cache expiration was checking if the cache was created earlier than 1 week ago today at 00:00:00
- Report configuration count was not included
- The warning about setting CIVIHR_STATISTICS_ENDPOINT was unclear
- Error responses from the server were not handled well. Even though an error was generated Civi doesn't really do HTTP status codes, and the HTTP client doesn't really care about them so much anyway.

## After

- Most recent login date for all roles is included (except anonymous and authenticated user)
- Cache expiration checks if the cache was created simply early than 1 week ago (from now)
- Report configuration count is included
- The warning about setting CIVIHR_STATISTICS_ENDPOINT includes a reason why
- Error responses handling is improved. Although basic, the response body is checked, and since we don't expect any response from the server, a non-zero response body is evaluated as a request error

## Notes

`CRM_HRCore_Test_Helpers_SessionHelpersTrait` was renamed and made into a class, so the tests needed to be updated

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
@@ -26,8 +26,9 @@ class CRM_HRCore_CMSData_Role_DrupalRoleService implements RoleServiceInterface{
 
     $result = $query->execute()->fetchAllKeyed();
 
-    $returnArray = [];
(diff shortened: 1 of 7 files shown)


mickadoo:
Closing in favor of https://github.com/civicrm/civihr/pull/2410

PR #2408, opened 2018-01-15
PCHR-3138: Add UserAccount Interface and related classes

tunbola:
## Overview
This PR adds some generic CMS related classes and interfaces to the HRCore extension that can be used by other extension to display and perform CMS framework related actions.
The class added in this PR is the UserAccount Interface and its related classes.

## Before
CMS User interfaces and classes related to User Account were not present.

## After
The following interface was added in this PR:

- CRM_HRCore_CMSData_UserAccountInterface: This interface provides methods functionality for taking some actions/providing some information on a user account. Currently it provides method for cancelling, enabling and disabling a user account and also provides information on whether a user account is disabled or not.

The following class was added to implement the Interface
- The CRM_HRCore_CMSData_UserAccount_Drupal class

The following factory class was added:
- The CRM_HRCore_CMSData_UserAccountFactory.

## Comments
- These classes were added primarily to aid the task being done for https://compucorp.atlassian.net/browse/PCHR-3138
- The cancel method implementation in the CRM_HRCore_CMSData_UserAccount_Drupal allows a user account to be cancelled and make its content belong to the Anonymous user. This has been tested locally on my machine and works as expected.

PR #2413, opened 2018-01-16
PCHR-3138: Add Page URL Callbacks for Disabling/Enabling and Deleting User Account

tunbola:
## Overview
This PR adds the page URL's for disabling, enabling and deleting the CMS user of a contact and the Page callbacks that handles these functionalities.

## Before
The functionality for disabling, enabling and deleting the CMS user of a contact does not exist.

## After
- The User account interface and related classes added in #2408, were utilized to implement the needed functionalities, The disable, enable and cancel methods on this interface were used after getting the appropriate CMS object from the factory.

## Comments
- Commits 57074b5 fixes the access arguments for the User notifier related paths added in a previous PR, the arguments needed is 'access CiviCRM' rather than 'edit all contacts'
- Also there's an error about the CRM_Hrcontactactionsmenu_ExtensionUtil not found in the hrcontactactionsmenu.php file, this was also fixed.
- Also tasks for https://compucorp.atlassian.net/browse/PCHR-3139 has been merged with https://compucorp.atlassian.net/browse/PCHR-3138 and the former will be closed as the tasks are better done in a single PR.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
+++ b/uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
@@ -13,13 +13,7 @@ class CRM_HRContactActionsMenu_Page_UserAccount {
    * contact.
    */
   public static function delete() {
-    $contactID = CRM_Utils_Array::value('cid', $_GET);
-    $contactInfo = ContactHelper::getUserInformation($contactID);
-    $userAccount = UserAccountFactory::create();
-    $userAccount->cancel($contactInfo);
-
-    CRM_Core_Session::setStatus(ts('User account has been deleted'), 'Success', 'success');
-    self::redirectToContactSummaryPage($contactID);
+    self::doUserAction('delete', 'User account has been deleted');
   }
 
   /**
@@ -27,13 +21,7 @@ class CRM_HRContactActionsMenu_Page_UserAccount {
    * contact.
    */
   public static function disable() {
-    $contactID = CRM_Utils_Array::value('cid', $_GET);
-    $contactInfo = ContactHelper::getUserInformation($contactID);
-    $userAccount = UserAccountFactory::create();
-    $userAccount->disable($contactInfo);
-
-    CRM_Core_Session::setStatus(ts('User account has been disabled'), 'Success', 'success');
-    self::redirectToContactSummaryPage($contactID);
+    self::doUserAction('disable', 'User account has been disabled');
   }
 
   /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Function to enable the CMS user account of the
+   * contact.
+   */
+  public static function enable() {
+    $contactID = CRM_Utils_Array::value('cid', $_GET);
+    $contactInfo = ContactHelper::getUserInformation($contactID);
+    $userAccount = UserAccountFactory::create();
```
davialexandre: These 3 lines are repeated in each of the methods. How about extracting them to a separate method?

uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
```diff
@@ -0,0 +1,58 @@
...
+    $userAccount->enable($contactInfo);
+
+    CRM_Core_Session::setStatus(ts('User account has been enabled'), 'Success', 'success');
+
+    $url = CRM_Utils_System::url('civicrm/contact/view', "reset=1&cid=$contactID");
+    CRM_Utils_System::redirect($url);
```
davialexandre: I think that these two can also be extracted

uk.co.compucorp.civicrm.hrcontactactionsmenu/hrcontactactionsmenu.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+use CRM_Hrcontactactionsmenu_ExtensionUtil as E;
```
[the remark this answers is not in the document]
tunbola: Still getting errors after moving the require statement up. I had to add back the class alias.
Seems the alias is defined in the included file does not have effect here.
I remember civi added the alias in this file when I created the extension but I manually removed it.

uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
tunbola: Done

uk.co.compucorp.civicrm.hrcontactactionsmenu/CRM/HRContactActionsMenu/Page/UserAccount.php
tunbola: Its not straightforward to extract, reason is the $contactInfo is needed to perform necessary action on the User object and $contactID is needed to get the contact Info and also for the redirection.
Any Ideas?

davialexandre: I see. How about extract the only first two lines then?

tunbola: Issue is I would still need the first line
`$contactID = CRM_Utils_Array::value('cid', $_GET);` so I can pass the contact ID to the self::redirectToContactSummaryPage function.
Except I have to duplicate the `$contactID = CRM_Utils_Array::value('cid', $_GET);` in the self::redirectToContactSummaryPage function function and then in this function we want to extract the first two lines to which doesn't seem to be necessary to me.

davialexandre: How about this:

```php
private static function doUserAction($action, $successMessage) {
    $contactID = CRM_Utils_Array::value('cid', $_GET);
    $contactInfo = ContactHelper::getUserInformation($contactID);
    $userAccount = UserAccountFactory::create();
    $userAccount->$action($contactInfo);
    CRM_Core_Session::setStatus(ts($successMessage), 'Success', 'success');

    $url = CRM_Utils_System::url('civicrm/contact/view', "reset=1&cid=$contactID");
    CRM_Utils_System::redirect($url);
}
```

And then each callback method would be like:

```php
public function enable() {
  self::doUserAction('enable', 'User account has been enabled');
}

// and so on
```

PR #2418, opened 2018-01-17
PCHR-3143: Add service to parse before-hash query string parameters

bmnepali:
## Overview
This PR adds a new service in `reqangular` extension which will be used to parse the available query strings  in the given url into a list of query strings.

## Technical Details
1. A `beforeHashQueryParams` service is added in `reqangular` extension.

This  service will take  similar to following `url` as paramater
```js
// Usage example, with the current URL as
// /civicrm/contact/view/?reset=1&cid=4&selectedChild=civitasks#/
```

And it will be used as following in controllers:
```js
myModule.controller('myController', function (beforeHashQueryParams) {
  beforeHashQueryParams.parse();   // { reset: 1, cid: 4, selectedChild: 'civitasks' } 
}); 
```

## Tests
✅ JS distributive files - n/a
⏹ CSS distributive files - not needed to run
✅ Jasmine Tests - added
⏹ Backstop Tests - n/a


org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-env amd */
+
+define([
+  'common/modules/services'
+], function (module) {
+  'use strict';
+
+  module.factory('$beforeHashQueryParams', ['$log', '$window',
```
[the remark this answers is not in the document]
bmnepali: Ok! will remove!

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,48 @@
...
+        var cleanUrl = urlToParse.substring(0, hashPos !== -1 ? hashPos : urlToParse.length);
+        var query = urlToParse.indexOf('?');
+        var tokens = cleanUrl.substr(query + 1).split('&');
+        var tokensLength = tokens.length;
+
+        if (query === -1 || cleanUrl.indexOf('=') === -1) {
+          return obj;
+        }
+
+        for (var i = 0; i < tokensLength; i++) {
```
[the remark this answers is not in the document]
bmnepali: Will do.

bmnepali:
Fixed merge conflicts by rebase.

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
```diff
@@ -0,0 +1,54 @@
...
+          // url, expected query params
+          ['/civicrm/contact/view', { }],
+          ['/civicrm/contact/view/?reset=1', { reset: '1' }],
+          ['/civicrm/contact/view/?reset=1&cid=4', { reset: '1', cid: '4' }],
+          [
+            '/civicrm/contact/view/?reset=1&cid=4&selectedChild=civitasks#/',
+            { reset: '1', cid: '4', selectedChild: 'civitasks' }
+          ],
+          [
+            '/civicrm/contact/view/?reset=1&cid=4&selectedChild=civitasks#/tasks',
```
AkA84: this test is pointless, please remove it

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
```diff
@@ -0,0 +1,54 @@
...
+            { reset: '1', cid: '4', selectedChild: 'civitasks' }
+          ],
+          [
+            '/civicrm/contact/view/?reset=1&cid=4&selectedChild',
```
AkA84: also please remove this one

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
```diff
@@ -0,0 +1,54 @@
...
+            { reset: '1', cid: '4', selectedChild: null }
+          ],
+          [
+            '/civicrm/contact/view/?reset=1&cid=4&',
+            { reset: '1', cid: '4' }
+          ],
+          [
+            '/civicrm/contact/view/?reset=1&cid=4&foo=',
+            { reset: '1', cid: '4', foo: '' }
```
AkA84: if `/civicrm/contact/view/?reset=1&cid=4&selectedChild` -> `{ selectedChild: null }`
then for consistency `/civicrm/contact/view/?reset=1&cid=4&foo=` -> `{ foo: '' }`

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
```diff
@@ -0,0 +1,54 @@
...
+          ]
+        ];
+      }));
+
+      describe('beforeHashQueryParams.parse()', function () {
+        it('parses the given url correctly', function () {
+          testMap.forEach(function (testCase) {
+            expect(beforeHashQueryParams.parse(testCase[0])).toEqual(testCase[1]);
```
AkA84: i'm sorry but i really don't like this approach, although i grant that it has its utility in specific cases. By shoving all the test cases in an array, we really lose the expressiveness that one gets from a test suite well written. A good test suite should read almost like a collection of english sentences, and if we keep doing this we completely lose that and instead we are forced to take a look at a bunch of values in an array

This should be rewritten as something like this
```js
describe('when there is no query string in the url', function () {
  it('returns an empty object', function () {
    // /civicrm/contact/view
  });
});

describe('when is a query string, but after the hash', function () {
  it('ignores the query string and returns an empty object', function () {
    // /civicrm/contact/view#/?foo=bar
  });
});

describe('when there is a query string before the hash', function () {
  it('parses the query string and returns key/value pairs in an object', function () {
    // /civicrm/contact/view?foo=bar&baz=bax
  });
});

describe('when there is a query string both before and after the hash', function () {
  it('parses only the query string before the hash', function () {
    // /civicrm/contact/view?foo=bar#/?baz=bax
  });
});

describe('when there is a query string that trails with an ampersand', function () {
  it('parses the query string ignoring the ampersand', function () {
    // /civicrm/contact/view?foo=bar&
  });
});

describe('when there is a query string that trails with a key without value', function () {
  it('parses the query string and assigns a null value to the key', function () {
    // /civicrm/contact/view?foo=bar&baz
    // /civicrm/contact/view?foo=bar&baz=
  });
});
```

of course it's much more code to write, but it's undeniable that one can simply just read at the `describe` and `it` titles and understand immediately how this service behaves, without having to look at the tests implementation.

Imho the object of a test suite is to read well first and then to be written concisely, not the other way around. Not to mention that as a the test author, if you write the test cases in a "conversational" tone, it makes it easier to think about the logic you are implementing

(cc @igorpavlov @reneolivo @deb1990 )

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
```diff
@@ -0,0 +1,54 @@
...
+          });
+        });
+      });
+    });
+  });
+})(CRM);
```
AkA84: there is no reason to have this IIFE, you are not using `CRM`

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,79 @@
...
+        return prepareQueryParamsObject(getQueryParams(cleanUrl));
+      }
+
+      /**
+       * Prepare Query params object
+       *
+       * @param {Array} queryTokens
+       * @return {Object}
+       */
+      function prepareQueryParamsObject (queryTokens) {
```
AkA84: you have a function called `getQueryParams`, but here you are talking about `queryTokens`. Please use consistent naming

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,79 @@
...
+        var queryParams = {};
+
+        queryTokens.forEach(function (token, index) {
```
AkA84: you should be able to use [`_.transform`](https://lodash.com/docs/3.10.1#transform)

```js
return _.transform(queryTokens, function (result, token) {

});

File: org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js [hunk 10]
```diff
@@ -0,0 +1,79 @@
...
+          var splittedToken = token.split('=');
```
AkA84: it's `split`

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,79 @@
...
+
+          if (splittedToken[0]) {
```
AkA84: what if the url was `/civicrm/contact/view/?reset&cid` ?

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
```diff
@@ -0,0 +1,79 @@
...
+            queryParams[decodeURI(splittedToken[0])] = splittedToken.hasOwnProperty(1)
```
AkA84: `splittedToken.hasOwnProperty(1)`
what's this?

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
bmnepali: Fixed!

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
bmnepali: Added the coverage for this case. now in this case the query params object will be:
```js
{ reset: null, cid: null }
```

org.civicrm.reqangular/src/common/services/before-hash-query-params.service.js
bmnepali: `splittedToken.hasOwnProperty(1)` checks if the object has a property called `1`

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
bmnepali: done!

org.civicrm.reqangular/test/services/before-hash-query-params.service.spec.js
bmnepali: removed!

PR #2419, opened 2018-01-18
PCHR-3094: Fix validation errors after changing Leave Request's Absence Type

davialexandre:
## Overview

After changing the Leave Request's Absence Type to a type with a different calculation unit, users were not able to update the request status. Staff members would get an error saying that one of the date type or date amount fields had to be empty and Managers would get an error saying that they could not change the request dates (even though they were not changing any date).

## Before

As a staff member, trying to update the status of a Leave Request that had its Absence Type changed to another one with a different calculation unit would result in a validation error:

before_staff

As a manager, trying to update the status of the same Leave Request would result in a validation error saying that the dates could not be changed:

manager_before

## After

Staff members can now update the Leave Request status after it had its Absence Type changed to another type with a different calculation unit:

staff_after

Managers can update the status of the that Leave Request:

manager_after

## Technical Details

There are a few fields in the Leave Request table that are required to be empty according to the calculation unit of the request's Absence Type. For calculation unit in days, the `from_date_amount` and `to_date_amount` should be empty. For calculation unit in hours, the `from_date_type` and `to_date_type` are the ones that should be empty. 

When switching the Absence Type of a Leave Request to another one with a different calculation unit, the fields required to be empty in the database were not reset and this was causing the errors mentioned above. Example:

- First create a Leave Request for an Absence Type in days. 
  - The FE will not send values for the `from_date_amount` and `to_date_amount` fields (because they're not required for requests in days)
  - In the database, the `from_date_type` and `to_date_type` will not be empty, while the `from_date_amount` and `to_date_amount` will.
- Edit the Leave Request and change the Absence Type to one in hours
  - The FE will not send values for the `from_date_type` and `to_date_type` fields (because they are not required for requests in hours)
  - In the database, the `from_date_type` and `to_date_type` will not be touched (because no value was sent) and old value will be kept. Values will be set for `from_date_amount` and `to_date_amount`
- Now, when we update (to change its status, for example) the Leave Request, all the fields will not be empty and since the calculation unit is now in hours, the validation will give an error saying that the `from_date_type` field should be empty.

The fix here was to, right before saving the updates for a Leave Request, to check if the calculation unit has changed and, in that case reset the values of the fields that should be empty.

This, however, would not work for Leave Requests that were created before this fix and already had the problem. For this reason, an upgrader was added to fix them

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -1378,8 +1378,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    *
    * @param array $params
    *
-   * @return mixed
-   * @throws \Exception
+   * @return array
+   *   The $params array with empty values for the non-mandatory fields
    */
   private static function resetNonMandatoryDateFieldsIfCalculationUnitChanged($params) {
     if (self::calculationUnitChanged($params)) {
@@ -1394,6 +1394,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
         $params['to_date_amount']   = '';
       }
     }
+
     return $params;
   }
 
@@ -1412,13 +1413,13 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   private static function calculationUnitChanged($params) {
     if (empty($params['id'])) {
       // New Leave Request being created, so it didn't change
-      return false;
+      return FALSE;
     }
 
     $currentAbsenceType = self::getCurrentAbsenceType($params['id']);
 
     if ($currentAbsenceType->id == $params['type_id']) {
-      return false;
+      return FALSE;
     }
 
     $newAbsenceType = AbsenceType::findById($params['type_id']);
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -49,6 +49,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     unset($params['is_deleted']);
 
     $datesChanged = self::datesChanged($params);
+
+    $params = self::resetNonMandatoryDateFieldsIfCalculationUnitChanged($params);
+
     $instance = new self();
     $instance->copyValues($params);
 
@@ -102,8 +105,6 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     if($validationMode != self::IMPORT_VALIDATION) {
       self::validateEntitlementAndWorkingDayAndBalanceChange($params, $absenceType, $absencePeriod);
     }
-
-
   }
 
   /**
@@ -1363,5 +1364,79 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     }
 
     return self::calculateBalanceChangeFromCreateParams($params);
+  }
+
+  /**
+   * There are some fields that are only required according to the calculation
+   * unit of the Leave Request's Absence Type. This method makes sure that the
+   * non-required fields according to the calculation unit will be reset.
+   *
+   * When the calculation unit is hours, the from_date_type and to_date_type
+   * fields are not required, while the from_date_amount and to_date_amount are.
+   * When the calculation unit is in days, the from_date_amount and to_date_amount
+   * are required and the from_date_type and to_date_type are not.
+   *
+   * @param array $params
+   *
+   * @return mixed
+   * @throws \Exception
+   */
+  private static function resetNonMandatoryDateFieldsIfCalculationUnitChanged($params) {
+    if (self::calculationUnitChanged($params)) {
+      $absenceType = AbsenceType::findById($params['type_id']);
+
+      if ($absenceType->isCalculationUnitInHours()) {
+        $params['from_date_type'] = '';
+        $params['to_date_type']   = '';
+      }
+      else {
+        $params['from_date_amount'] = '';
+        $params['to_date_amount']   = '';
+      }
+    }
+    return $params;
+  }
+
+  /**
+   * Returns whether the Calculation Unit used for this Leave Request is being
+   * changed or not.
+   *
+   * The calculation unit is defined by the Absence Type. If the user changes
+   * the Leave Request's Absence Type to another type with a different
+   * calculation unit, this method will return true.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private static function calculationUnitChanged($params) {
+    if (empty($params['id'])) {
+      // New Leave Request being created, so it didn't change
+      return false;
+    }
+
+    $currentAbsenceType = self::getCurrentAbsenceType($params['id']);
+
+    if ($currentAbsenceType->id == $params['type_id']) {
+      return false;
+    }
+
+    $newAbsenceType = AbsenceType::findById($params['type_id']);
+
+    return $newAbsenceType->calculation_unit != $currentAbsenceType->calculation_unit;
+  }
+
+  /**
+   * Returns the current Absence Type for the Leave Request with the given ID.
+   *
+   * @param int $leaveRequestID
+   *
+   * @return AbsenceType
+   * @throws \Exception
+   */
+  private static function getCurrentAbsenceType($leaveRequestID) {
+    $currentLeaveRequest = LeaveRequest::findById($leaveRequestID);
+
+    return AbsenceType::findById($currentLeaveRequest->type_id);
   }
 }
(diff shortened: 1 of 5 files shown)


mickadoo:
I'm not sure about the meaning of:

> When switching the Absence Type of a Leave Request to another one with a different calculation unit, the respective where not reset in the database and this was causing the errors mentioned above

`s/and to_date_type will not empty/and to_date_type will not be empty/`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1364,4 +1365,78 @@ private static function getLeaveRequestBalance($params) {
...
+   * non-required fields according to the calculation unit will be reset.
+   *
+   * When the calculation unit is hours, the from_date_type and to_date_type
+   * fields are not required, while the from_date_amount and to_date_amount are.
+   * When the calculation unit is in days, the from_date_amount and to_date_amount
+   * are required and the from_date_type and to_date_type are not.
+   *
+   * @param array $params
+   *
+   * @return mixed
```
mickadoo: return type

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1364,4 +1365,78 @@ private static function getLeaveRequestBalance($params) {
...
+   * @throws \Exception
```
mickadoo: I think we should decide on whether to include these or not, I tried to do a poll on it and it was 2 votes against vs 0 votes for

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1364,4 +1365,78 @@ private static function getLeaveRequestBalance($params) {
...
+   * calculation unit, this method will return true.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private static function calculationUnitChanged($params) {
+    if (empty($params['id'])) {
+      // New Leave Request being created, so it didn't change
+      return false;
```
mickadoo: boolean case

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1364,4 +1365,78 @@ private static function getLeaveRequestBalance($params) {
...
+    }
+
+    $currentAbsenceType = self::getCurrentAbsenceType($params['id']);
+
+    if ($currentAbsenceType->id == $params['type_id']) {
+      return false;
```
mickadoo: case

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1364,4 +1365,78 @@ private static function getLeaveRequestBalance($params) {
...
+    }
+
+    $newAbsenceType = AbsenceType::findById($params['type_id']);
+
+    return $newAbsenceType->calculation_unit != $currentAbsenceType->calculation_unit;
```
mickadoo: this is a nice function, very easy to read

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -49,6 +49,9 @@ public static function create($params, $validationMode = self::VALIDATIONS_ON) {
     unset($params['is_deleted']);
 
     $datesChanged = self::datesChanged($params);
+
+    $params = self::resetNonMandatoryDateFieldsIfCalculationUnitChanged($params);
```
mickadoo: Just a question, since this is making some changes to the `$params` should validation be called after this?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1016.php
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @return bool
+   */
+  public function upgrade_1016() {
+    $calculationUnits = array_flip(AbsenceType::buildOptions('calculation_unit', 'validate'));
+
+    $leaveRequestTable = LeaveRequest::getTableName();
+    $absenceTypeTable = AbsenceType::getTableName();
+
+    $query = "
```
mickadoo: Is it not possible to do this using the API?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1016.php
```diff
@@ -0,0 +1,71 @@
...
+      FROM {$leaveRequestTable} lr
+      INNER JOIN {$absenceTypeTable} a ON lr.type_id = a.id
+    WHERE
+      (a.calculation_unit = %1 AND (lr.from_date_amount IS NOT NULL OR lr.to_date_amount IS NOT NULL ))
+      OR
+      (a.calculation_unit = %2 AND (lr.from_date_type IS NOT NULL OR lr.to_date_type IS NOT NULL ))
+    ";
+
+    $params = [
+      1 => [$calculationUnits['days'], 'Integer'],
```
mickadoo: Is this an array of units or unit IDs?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1016.php
```diff
@@ -0,0 +1,71 @@
...
+      2 => [$calculationUnits['hours'], 'Integer'],
+    ];
+
+    $recordToUpdate = CRM_Core_DAO::executeQuery($query, $params);
+
+    while($recordToUpdate->fetch()) {
```
mickadoo: space after `while`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -4361,4 +4362,64 @@ public function testCalculateBalanceChangeForTwoDaysLeaveRequestWhenStartAndEndD
...
+    // Change absence type to one in hours
+    $leaveRequestParams = array_merge($leaveRequestParams, [
+      'id' => $leaveRequest->id,
+      'type_id' => $absenceTypeInHours->id,
+      // These two are required for Leave in Hours
+      'to_date_amount' => 7.5,
+      'from_date_amount' => 0,
+    ]);
+
+    $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation($leaveRequestParams);
```
mickadoo: This is a bit misleading since it looks like it's creating a new `$leaveRequest`. I know it's the same with the `create` method of the API, so not sure what to suggest here. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1016.php
davialexandre: The API doesn't support complex conditions like the one used here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1016.php
davialexandre: It's an [array of option values](https://github.com/civicrm/civihr/pull/2419/files/4c297f5110db5b80b54f9ef842245e7d418637e3#diff-6875a88975ef40ac160d3a80acd125afR24) indexed by the option value name.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
davialexandre: This is not really about validation, but saving the correct value in the database. Depending on the calculation unit, some fields will simply not be present in the `$params` array. For example, when updating a leave request to an Absence Type in hours, the `from_date_type` and `to_date_type` fields will not be sent in the request. For that reason, the validation will pass ([this validation, more specifically](https://github.com/civicrm/civihr/blob/staging/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php#L226-L275)), but the old values will be kept in the database (because they're not included in the `$params` array). This method ensures that the fields will be included in the params array, and that they will be empty, so that they will be set to null in the database.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
davialexandre: I think it's not as simple as not using them or not, but I removed it for now

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
davialexandre: I changed it to `create`. As you said, it's pretty much the same thing, but it was the best I could do.

davialexandre:
Rebased to fix merge conflicts

PR #2421, opened 2018-01-18
PCHR-3088: Fix bug in Drupal User Account isUserDisabled method

tunbola:
## Overview
This PR fixes a bug in the isUserDisabled method added in #2408

## Before
The isUserDisabled method of the CRM_HRCore_CMSData_UserAccount_Drupal class was not returning correctly.

## After
The isUserDisabled method of the CRM_HRCore_CMSData_UserAccount_Drupal class was fixed to returning correctly.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserAccount/Drupal.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserAccount/Drupal.php
@@ -69,6 +69,6 @@ class CRM_HRCore_CMSData_UserAccount_Drupal implements UserAccountInterface {
   public function isUserDisabled($contactData) {
     $user = $this->getUser($contactData);
 
-    return $user->status;
+    return !$user->status ? true : false;
   }
 }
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserAccount/Drupal.php
```diff
@@ -69,6 +69,6 @@ public function enable($contactData) {
   public function isUserDisabled($contactData) {
     $user = $this->getUser($contactData);
 
-    return $user->status;
+    return !$user->status ? true : false;
```
davialexandre: Why didn't the test catch this? Shouldn't the test be updated so that the error doesn't happen again?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/drupal_function_mocks.php
```diff
@@ -35,6 +35,12 @@ function user_load($userID) {
   $user = new stdClass();
   $user->roles = [1 => 'Fake Role'];
   $user->uid = $userID;
+  $user->status = 1;
+
+  if($userID == 0) {
+    $user->status = 0;
+  }
+
```
davialexandre: I'll approve this change, but this type of change is not very maintainable. Eventually, we'll need to come up with a better solution for mocking this functions. I think we can implement some sort of in-memory database that can be used by these user_ functions to keep track of users

davialexandre:
@tunbola please add the "ready to test" label here

PR #2424, opened 2018-01-19
PCHR-3189: Fix count of public holidays in the entitlement calculation for multiple contracts

davialexandre:
## Overview

Depending on what is set in a staff member's contract, the number of public holidays within the contact's dates during the absence period might be added to the entitlement. If the staff member has more than one contract during the same absence period, the calculation must consider all the public holidays within all of these contracts. There was an error in how the number of public holidays was calculated, which led to the wrong number being added to the entitlement.

## Before

In a scenario where the staff member has multiple contracts in the same absence period and there were public holidays in more than one contract, the number of public holidays added to the entitlement would be wrong.

## After

The number of public holidays added to the entitlement now includes all the public holiday within the contracts in the absence period.

## Technical Details

After the calculation is done, we save it to the database using the `LeavePeriodEntitlement.saveFromCalculation($calculation)` method. It takes an `EntitlementCalculation` instance and calls the `getPublicHolidaysInEntitlement()` method to get a list of all the Public Holidays within all the contracts in the calculation's Absence Period.

This can be seen here (this `savePublicHolidaysBalanceChanges()` is called by `saveFromCalculation()`):

```php
private static function savePublicHolidaysBalanceChanges(
    EntitlementCalculation $calculation,
    LeavePeriodEntitlement $periodEntitlement
  ) {
    $balanceChangeTypes = array_flip(LeaveBalanceChange::buildOptions('type_id', 'validate'));

    $publicHolidays = $calculation->getPublicHolidaysInEntitlement();

    if (!empty($publicHolidays)) {
      LeaveBalanceChange::create([
        'type_id'     => $balanceChangeTypes['public_holiday'],
        'source_id'   => $periodEntitlement->id,
        'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT,
        'amount'      => count($publicHolidays)
      ]);
    }
  }
```
What was causing the issue is the way that the `EntitlementCalculation.getPublicHolidaysInEntitlement()` was implemented:

```php
  public function getPublicHolidaysInEntitlement() {
    $publicHolidays = [];

    $calculations = $this->getContractEntitlementCalculations();
    foreach($calculations as $calculation) {
      $publicHolidays += $calculation->getPublicHolidaysInEntitlement();
    }

    return $publicHolidays;
  }
```

As it's possible to see, it's using the `+=` operator to add new public holidays to the array. Checking the [PHP documentation](http://php.net/manual/en/language.operators.array.php), we can see this:

> The + operator returns the right-hand array appended to the left-hand array; **for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.**

So, when we added the items of one array to the existing array, it would ignore any item with keys that already existed in the `$publicHolidays` array and the final array would not contain the correct number of items. 

As we can see [here](https://3v4l.org/Yn6gg), in order to actually end up with a list containing all the public holidays, the `array_merge` function would be the correct choice. However, if we check again the code snippet from `savePublicHolidaysBalanceChanges()`, we can see that we don't really need a list of public holidays. What we actually need is the number of public holidays in the period. So, to fix the issue, I simply replaced the call to `getPublicHolidaysInEntitlement()` with a call to `getNumberOfPublicHolidaysInEntitlement()`. After that, I realized that `getPublicHolidaysInEntitlement()` was not being used anywhere else anymore, so I removed it all the related code in the `ContractEntitlementCalculation` class.

### Comments

On https://github.com/civicrm/civihr/pull/2424/commits/a893ab83690c59c81b10671bf85e59341531600d I have removed some misleading comment from the `savePublicHolidaysBalanceChanges()` method. The [original version of this method](https://github.com/civicrm/civihr/blob/01db3ee2e509bd9f766b38faf6c97dbe83bfe2c2/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php#L293-L307)  was also responsible to save the public holidays that would be deducted from the entitlement automatically (according to the Absence Type settings). This was later removed on https://github.com/civicrm/civihr/commit/05cabcbc27fd3cecd12c0540d4f1e4ac9c5b5524#diff-d8e6ee2cf0913fc50b896934c6ef0a71, because creating this was not really part of the entitlement calculation and it was causing the final numbers to not add up.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -299,14 +299,14 @@ private static function savePublicHolidaysBalanceChanges(
...
+    $publicHolidays = $calculation->getNumberOfPublicHolidaysInEntitlement();
 
     if (!empty($publicHolidays)) {
       LeaveBalanceChange::create([
         'type_id'     => $balanceChangeTypes['public_holiday'],
         'source_id'   => $periodEntitlement->id,
         'source_type' => LeaveBalanceChange::SOURCE_ENTITLEMENT,
-        'amount'      => count($publicHolidays)
+        'amount'      => $publicHolidays
       ]);
```
tunbola: Please format the array parameters passed to the create method properly. With single space between the array `=>` operator and the array key and values.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php
```diff
@@ -299,14 +299,14 @@ private static function savePublicHolidaysBalanceChanges(
   ) {
     $balanceChangeTypes = array_flip(LeaveBalanceChange::buildOptions('type_id', 'validate'));
 
-    $publicHolidays = $calculation->getPublicHolidaysInEntitlement();
+    $publicHolidays = $calculation->getNumberOfPublicHolidaysInEntitlement();
```
tunbola: The name of this variable should reflect that it stores the number of public holidays and not just public holidays.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -430,8 +430,8 @@ public function testCanSaveALeavePeriodEntitlementFromAnEntitlementCalculation()
     $this->assertEquals($type->id, $periodEntitlement->type_id);
     $this->assertEquals($this->contract['contact_id'], $periodEntitlement->contact_id);
 
-    // 10 + 1 + 3 (Pro Rata + Brought Forward + No. Public Holidays)
-    $this->assertEquals(14, $periodEntitlement->getEntitlement());
+    // 10 + 1 (Pro Rata (Including the Public Holidays) + Brought Forward)
+    $this->assertEquals(11, $periodEntitlement->getEntitlement());
```
tunbola: Off the top of my head, I don't know why this value should change from 14. The public holiday array contained 3 public holidays and now it was replaced with the integer 3 since we now need the number of public holidays. except something changed in the logic which I can't seem to see.
Shouldn't it still be 14?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
```diff
@@ -443,7 +443,7 @@ public function testCanSaveALeavePeriodEntitlementFromAnEntitlementCalculation()
       return $balanceChange->type_id == $balanceChangeTypes['leave'];
     });
     $this->assertCount(1, $leaveBalanceChanges);
-    $this->assertEquals($proRata, reset($leaveBalanceChanges)->amount);
+    $this->assertEquals($proRata - $numberOfPublicHolidays, reset($leaveBalanceChanges)->amount);
```
tunbola: Why do we have to subtract the number of public holidays from proRata?

davialexandre: Because the number of public holidays is included in the calculation pro rata, as it can be seen [here](https://github.com/civicrm/civihr/blob/f2120b399208879e80912e70cee612b6db4c71bf/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/ContractEntitlementCalculation.php#L40-L61)

However, we need two separate Leave Balance Changes. One only with the number of public holidays and another one only with the "Leave" amount. So [here](https://github.com/civicrm/civihr/blob/c34632d99947ad7390c135e4a95fea38defcb641/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlement.php#L236), we separate the values when the balance changes are created

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeavePeriodEntitlementTest.php
davialexandre: It changed because the number of public holidays is already included in the Pro Rata (as the comment says).

The previous version of the test was wrong. It uses an entitlement calculation mock that didn't have all the methods used by the `saveFromCalculation()` method. For saving the Leave amount, the `saveFromCalculation()` calls the `saveLeaveBalanceChange()` method, which calculates the amount with this:

```php
'amount' => $calculation->getProRata() - $calculation->getNumberOfPublicHolidaysInEntitlement()
```

The `getNumberOfPublicHolidaysInEntitlement()` method was not mocked, so it was actually running a DB query to get the number. Since this test doesn't save any Public Holidays in the database, it would return 0 rather than 3, and the store amount was wrong. 

After the changes, we now have a mock for the `getNumberOfPublicHolidaysInEntitlement()`, so the test started failing and the fix was to change the value from 14 to 11.

tunbola: Thanks for the explanation.
The number of public holidays in the test still seems confusing though because its difficult to imagine how the public holiday is part of the proRata, The way it is, if we change the `$numberOfPublicHolidays` to any value it will not affect the test outcome.
I would suggest that we change the `getEntitlementCalculationMock` function a little, Instead of the fifth parameter being the `proRata`, It will be `proRataWithoutPublicHoliday`, then in the function itself, we make the following changes

```php
    $calculation->expects($this->once())
                ->method('getProRata')
                ->will($this->returnValue($proRataWithoutPublicHoliday + $numberOfPublicHolidays));
```
This way it will be easy to see how the public holiday is reflected in the calculation of the overall proRata in the test.

PR #2445, opened 2018-01-30
PCHR-3055: Leave Balance Improvements

deb1990:
### Frontend 
This PR adds the filter by staff option to the Leave Balance Tab.

## Before
2018-01-30 at 4 18 pm

## After
2018-01-30 at 4 16 pm

## Technical Details
1. Added `ui.select` to the `leave-balance-tab-filters.html` for the dropdown.
2. In `leave-balance-tab.component.js`
   - Added `vm.lookupContacts` which holds all contacts loaded via `loadAllContacts()` function.

## Comments
1. Fixed semistandard issues.
2. Fixed dependencies for the touched spec files.

### Backend
Backend changes are pending, the same branch will be used for that.


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab.component.js
```diff
@@ -68,10 +69,21 @@ define([
...
+        .then(function (absenceTypes) {
+          vm.absenceTypes = absenceTypes;
+        });
+    }
+
+    /**
+     * Uses the Contact model to populate a list of all contacts
+     * sorted by sort_name in an ascending order.
+     */
+    function loadAllContacts () {
```
reneolivo: Please document the return value for this function.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-balance-tab.component.spec.js
```diff
@@ -180,6 +194,23 @@ define([
...
+          $rootScope.$digest();
+        });
+
+        it('loads the contacts sorted by sort name', function () {
+          expect(Contact.all).toHaveBeenCalledWith({
+            options: { sort: 'sort_name ASC' }
+          });
+        });
+
+        it('stores the absence periods', function () {
```
reneolivo: why is this test under contacts? I think you have not updated this test.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-balance-tab-filters.component.js
```diff
@@ -12,6 +12,7 @@ define([
       absencePeriods: '<',
       absenceTypes: '<',
       loggedInContactId: '<',
+      lookupContacts: '<',
```
reneolivo: You need to initialize the contact_id filter in this file ([check this](https://github.com/civicrm/civihr/pull/2445/files#diff-7c168ab0b0a7a6aeb18c4f1f49395dffR28)) and [check if it's initialized correctly](https://github.com/civicrm/civihr/blob/d550a20ddb2a8437e0def2dfdc3c9881335568d7/uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-balance-tab-filters.component.spec.js#L46).

deb1990: The following discussion I had with @reneolivo which explains why it was done like this.

Debarshi Bhaumik: I initially did that, but there is another problem. That is the contact_id is not a mandatory filter. so initially it stays null, and because of that, "areFiltersReady()" never returns true initially. So the loader is shown. And once I select a contact, then only the results are shown. That is the reason its not added initially

René Olivo: let me check that function, 1 sec

Debarshi Bhaumik: I have thought about another solution, that is to initiate contact_id as undefined, instead of null. That way areFiltersReady() would work initially

Debarshi Bhaumik: yep sure

René Olivo: I see, leave it like that then, the other option would be to set it as undefined, but that's just confusing

Debarshi Bhaumik: you mean, no need to change anything? I also thought that not having it initially makes more sence than setting undefined or changing the areFiltersReady() function

René Olivo: yes, don't worry about that one, I don't have a better suggestion for now

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-balance-tab.component.spec.js
deb1990: Oops, fixed.

PR #2450, opened 2018-02-05
PCHR-3061: Show backstop scenario label while running

deb1990:
## Overview
As part of this PR, 
1. the Backstop tests now show a text specifying the scenario which is currently running. This helps when backstop produces errors, because from the name we can point to the faulty scenario.

2. we are showing the current number of the scenario, which should help us to know how many tests are done and how many are remaining.

3. Showing a notification once backstop has finished.

## How it Looks
2018-02-06 at 4 40 pm

## Technical Details
1. Renamed `casper_scripts/login.js` → `casper_scripts/initialise.js`.
2.The `initialise` script is added to the scenarios always. And if `scenario.performLogin` is `true`, then only the script to perform login, is performed.
3. Added `gulp-notify` to show a notification to the user.

### Why `gulp-notify` instead of `node-notifier`
`node-notifier` depends on `notify-osd` or `libnotify-bin` external dependency on linux.
But the `gulp-notify` uses `notify-send` npm module, so no need to install external dependencies, which is always better.


uk.co.compucorp.civicrm.hrcore/backstop_data/casper_scripts/initialise.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = function (casper, scenario) {
+  var config = require('../site-config');
+  var loginFormSelector = 'form#user-login-form';
+  var credentials = config.credentials[scenario.credential];
+
+  casper.echo('--------------------------------------------', 'COMMENT');
+  casper.echo('RUNNING SCENARIO: ' + scenario.label, 'PARAMETER');
+
+  if (scenario.performLogin) {
```
AkA84: https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/190972313/Javascript#Javascript-Guardclauses

uk.co.compucorp.civicrm.hrcore/backstop_data/casper_scripts/initialise.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+module.exports = function (casper, scenario) {
+  var config = require('../site-config');
```
AkA84: the `require`s  should happen at the top of the file, not in the `exports

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -199,9 +199,10 @@ function scenariosList () {
 
       scenarios.forEach(function (scenario, index) {
         scenario.credential = scenario.credential || DEFAULT_CREDENTIAL;
+        scenario.onBeforeScript = 'initialise';
```
AkA84: "init" (so you also avoid the controversy between initialize and initialise)

AkA84:
@deb1990 i was thinking: is there some kind of "onError" hook that we can use in backstop? It would be great if we could show not what scenario is running, but what scenario is _has failed_

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
deb1990: Haha, nice point

uk.co.compucorp.civicrm.hrcore/backstop_data/casper_scripts/initialise.js
deb1990: Thanks for catching this.

deb1990:
> @deb1990 i was thinking: is there some kind of "onError" hook that we can use in backstop? It would be great if we could show not what scenario is running, but what scenario is has failed

Could not find anything. But added showing the scenario count while running a scenario, also added notification when gulp finishes.

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -139,7 +140,10 @@ function runBackstopJS (command) {
               configPath: BACKSTOP_DIR_PATH + destFile,
               filter: argv.filter
             }).catch(_.noop).then(function () { // equivalent to .finally()
-              gulp.src(BACKSTOP_DIR_PATH + destFile, { read: false }).pipe(clean());
+              gulp
+                .src(BACKSTOP_DIR_PATH + destFile, { read: false })
+                .pipe(clean())
+                .pipe(notify({ message: 'Backstop has finished', sound: true }));
```
AkA84: Most importantly you need to show a notification when something goes wrong (see if there is some kind of status for the notifications to mark if something has failed/succeeded)

uk.co.compucorp.civicrm.hrcore/backstop_data/casper_scripts/init.js
```diff
@@ -1,10 +1,17 @@
 'use strict';
+var config = require('../site-config');
 
 module.exports = function (casper, scenario) {
-  var config = require('../site-config');
   var loginFormSelector = 'form#user-login-form';
   var credentials = config.credentials[scenario.credential];
 
+  casper.echo('--------------------------------------------', 'COMMENT');
+  casper.echo('Running Scenario ' + scenario.count + ' : ' + scenario.label, 'PARAMETER');
```
AkA84: The formatting should be improved, you are putting the counter before the name of the scenario. Change it to

```bash
Running scenario "Contact Summary" (2 of 124)
```

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
deb1990: Done

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -138,8 +139,15 @@ function runBackstopJS (command) {
             var promise = backstopjs(command, {
               configPath: BACKSTOP_DIR_PATH + destFile,
               filter: argv.filter
-            }).catch(_.noop).then(function () { // equivalent to .finally()
-              gulp.src(BACKSTOP_DIR_PATH + destFile, { read: false }).pipe(clean());
+            }).then(function () {
+              notifier.notify({ message: 'Successful', title: 'BackstopJS', sound: 'Beep' });
+            }).catch(function () {
+              notifier.notify({ message: 'Failed', title: 'BackstopJS', sound: 'Beep' });
+              _.noop();
```
AkA84: there is no need for this

PR #2462, opened 2018-02-09
PCHR-3267: Fix leave widget errors

reneolivo:
## Overview
This PR fixes the following issues with the Leave Widget:

* The leave-widget-absence-types-amount-taken component was using $include instead of $inject for dependency injection.
* An API call for Leave Requests was being made without passing the right absence types.
* A race condition for displaying the next leave deduction was throwing an error because the calculation unit name was not ready.

Also, this PR fixes the path for the sourcemaps.

## Before
will be shown individually in the technical steps since they don't happen at the same time.

## After
screenshot 2018-02-09 05 49 38
No errors in the console.

### $include instead of $inject

Replaced `$include` with `$inject` in the [leave-widget-absence-types-amount-taken component](https://github.com/civicrm/civihr/pull/2462/files#diff-d6462c2eafdfc60fab42eb816712b36aR26).

#### How it looked before fixing:
screenshot 2018-02-09 05 53 47

### Leave Request API call fix

The reason why the API call was failing was because the absence types were not ready. To fix this, the leave-widget-absence-types-amount-taken component now waits until there are absence types to actually make the call:

```js
function areBindingsReady () {
  return vm.absenceTypes && vm.absenceTypes.length && vm.contactId &&
    vm.absencePeriod && vm.leaveRequestStatuses &&
    vm.leaveRequestStatuses.length;
}
```

#### How it looked before fixing:
screenshot 2018-02-09 06 00 42

### calculation unit name race condition

The calculation unit name needs to be defined before passing it to the `timeUnitApplier` filter, but the leave-widget-next-leave component was first loading the next leave, which produces a scope digest, and then it was mapping the calculation unit name for it, which produced a different scope digest. Sometimes the scope digest would be slow enough for the calculation unit name to be ready for the `timeUnitApplier`, but to make sure it always is an extra condition was added to the deduction display:

```html
<div ng-if="nextLeave.nextLeaveRequest && nextLeave.nextLeaveRequest['type_id.calculation_unit_name']">
  <h4>Deduction</h4>
  <span class="leave-widget__next-leave__deduction">
    {{nextLeave.nextLeaveRequest.balance_change | timeUnitApplier:nextLeave.nextLeaveRequest['type_id.calculation_unit_name']}}
  </span>
</div>
```

#### How it looked before fixing:
screenshot 2018-02-09 06 05 30

### Sourcemap paths

The path for sourcemap files was being generated like this:

`path/to/extension/extension-name/source.min.js.map`

instead of:

`path/to/extension/extension-name/js/dist/source.min.js.map`

To fix this issue the relative dist folder path was appended to the extension URL:

```js
gulp.src(path.join(distFolder, '*.min.js'))
  .pipe(replace(/sourceMappingURL=(.+)/, function (regExpMatch, fileName) {
    relativeDistFolderPath = this.file.base.replace(extensionPath, '');

    return 'sourceMappingURL=' + path.join(extensionUrl, relativeDistFolderPath, fileName);
  }))
  .pipe(gulp.dest(distFolder));
```

### Fixing spec files dependencies

When running `gulp watch --ext hrleaveandabsences` the tests would fail because the spec files were not loading the right dependencies. For example, some mock apis were missing, or including the leave widget module instead of the controller.


uk.co.compucorp.civicrm.hrcore/gulp/tasks/requirejs.js
```diff
@@ -165,14 +165,19 @@ function requireJsMainTask (cb) {
...
   var extensionPath = utils.getExtensionPath();
   var extensionUrl = utils.getExtensionUrlPath();
 
   // finds /dist folders that are outside /node_modules:
   distFolder = find.dirSync(/^((?!node_modules).)*\/dist$/, extensionPath).pop();
 
   gulp.src(path.join(distFolder, '*.min.js'))
-    .pipe(replace(/sourceMappingURL=(.+)/, 'sourceMappingURL=' + extensionUrl + '/$1'))
+    .pipe(replace(/sourceMappingURL=(.+)/, function (regExpMatch, fileName) {
+      relativeDistFolderPath = this.file.base.replace(extensionPath, '');
```
AkA84: explain with a comment what's going on here

uk.co.compucorp.civicrm.hrcore/gulp/tasks/requirejs.js
```diff
@@ -165,14 +165,19 @@ function requireJsMainTask (cb) {
  */
 function requireJsSourceMapPath (cb) {
   var distFolder;
+  var relativeDistFolderPath;
```
AkA84: this var should be defined in the function where it is used

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
```diff
@@ -60,8 +60,9 @@ define([
      * @return {Boolean}
      */
     function areBindingsReady () {
-      return vm.absenceTypes && vm.contactId && vm.absencePeriod &&
-        vm.leaveRequestStatuses && vm.leaveRequestStatuses.length;
+      return vm.absenceTypes && vm.absenceTypes.length && vm.contactId &&
```
AkA84: `vm.absenceTypes && vm.absenceTypes.length`

why do you need to check for both?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.spec.js
```diff
@@ -5,8 +5,11 @@ define([
   'common/moment',
   'mocks/helpers/controller-on-changes',
   'mocks/data/option-group.data',
+  'mocks/apis/absence-period-api-mock',
+  'mocks/apis/absence-type-api-mock',
   'mocks/apis/leave-request-api-mock',
-  'leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component'
```
AkA84: why are you not using this component anymore?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.js
reneolivo: `absenceTypes` is passed as a binding and might be undefined at first, and then it might be an empty array, and finally it will have a list of absence types, which is needed for displaying the absence types taken.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-widget/leave-widget-absence-types-amount-taken.component.spec.js
reneolivo: the `leave-widget.component` is including it since that's the base module.

PR #2464, opened 2018-02-09
PCHR-3259: Use lint staged for pre-commit linters

AkA84:
## Overview
This PR introduce the use of [lint-staged](https://github.com/okonet/lint-staged) to manage pre-commit linters, instead of relying on a manually-written hook

### Set up explanation
```json
{
  "scripts": {
    "precommit": "lint-staged"
  }
}
```
This is [husky](https://github.com/typicode/husky), being told to run `lint-staged` whenever a user makes a commit
```json
{
  "lint-staged": {
    "*.js": ["semistandard --fix", "git add"]
  }
}
```
This is the actual `lint-staged` configuration. Here we're telling it to run the `semistandard` linter in auto-fix mode, and if the linter automatically fixes issues on any of the staged files, it should automatically run `git add` on it.

(note that we don't need to exclude the minified files from the selector because `semistandard` already does that by default)

### Adding missing fields to `package.json`
The `package.json` was missing some basic fields, like "name", "license", etc


package.json
```diff
@@ -1,5 +1,32 @@
 {
+  "name": "CiviHR",
+  "description": "HR software for non-profit",
```
igorpavlov: Was this description confirmed with the PM?

package.json
```diff
@@ -1,5 +1,32 @@
...
+  },
   "semistandard": {
-    "globals": [ "CRM", "inject" ]
+    "globals": [
+      "CRM",
+      "inject"
+    ]
+  },
+  "devDependencies": {
+    "husky": "^0.14.3",
```
igorpavlov: Should not the versions be fixed?

package.json
AkA84: I took it from civihr.org (doesn't really matter in the end, this package won't be public anyway)

package.json
AkA84: no need for that, since we're using package-lock files

PR #2469, opened 2018-02-14
PCHR-3074: Add the CMS User Permission Interface and related classes

tunbola:
## Overview
This PR adds some generic CMS related classes and interfaces to the HRCore extension that can be used by other extensions.
The class added in this PR is the UserPermission Interface and its related classes

## Before
CMS UserPermission interface and related classes were not present.

## After
The following interface was added in this PR:
- CRM_HRCore_CMSData_UserPermissiontInterface: This interface provides functionality to check if a CMS user has access to certain permissions or not.

The following class was added to implement the Interface
- The CRM_HRCore_CMSData_UserPermission_Drupal class

The following factory class was added:
- The CRM_HRCore_CMSData_UserPermissionFactory.

## Comments
- These classes were added primarily to aid the task being done in https://compucorp.atlassian.net/browse/PCHR-3074

PR #2470, opened 2018-02-14
PCHR-3074: Display Contact Access Rights Info on Contact Actions Menu

tunbola:
## Overview
This PR displays the Contact Access Rights related menu items on the highlighted panel of the Contact Actions menu using the hook provided by the contactactionsmenu extension.

## Before
- Contact Access Rights related menu Items were displayed statically.

## After
- TheContact Access Rights related menu items are now displayed dynamically depending on the contact being viewed on the contact summary page.

Contact with access to Regions, Locations and ACL Groups

civihr_staff compucorp co uk _ staging17 2018-02-14 13-15-56

Contact with `view all contacts` and `edit all contacts` permission.

admin example com _ staging17 2018-02-14 13-17-23

Contact without `access Civicrm` permission

civihr_staff compucorp co uk _ staging17 2018-02-14 13-26-27

When the Manage Regional Access button is clicked

manageregionalbutton

When the tooltip icon is clicked:

tooltip

## Technical Details
- The Contact Access Rights extension implements the addContactMenuActions hook to display Leave related menu items on the Contact Actions menu.
- The `getRightsByType` method of the `CRM_Contactaccessrights_BAO_Rights` was made a private function,  Two new public methods: `getContactRightsByRegions` and `getContactRightsByLocations` were added with both methods still using the getRightsByType method internally in the service.
- The Contact Access Menu group is not displayed on the contact action menu for a contact without `access CiviCRM` permission. A new UserPermission class was added to HRCore extension to support this functionality #2469,
since the The CRM_Core_Permission permission class only allows to check for permissions for the current logged in user and it has no method that allows to check permission for a contact/user ID.


contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/Contact.php
```diff
@@ -0,0 +1,79 @@
...
+      'api.Group.get' => ['id' => "\$value.group_id", 'sequential' => 1],
+    ]);
+
+    $aclGroups = [];
+
+    if ($result['count'] < 0) {
+      return $aclGroups;
+    }
+
+    $accessControlValue = self::getAccessControlOptionValue();
```
davialexandre: just for consistency, wouldn't it make more sense for the variable to also be named `accessControlOptionValue` as the method that is being used to get it?

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessActionGroup.php
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * @var array
+   */
+  private $contactACLGroups;
+
+  /**
+   * @var CMSUserPermission
+   */
+  private $CMSUserPermission;
```
davialexandre: We don't start variable names with uppercase letters. Please change this to `$cmsUserPermission`

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessActionGroup.php
```diff
@@ -0,0 +1,164 @@
...
+    $actionsGroup = new ActionsGroup($this->getGroupTitle());
+    $isAdmin = $this->isAdmin();
+
+    if ($isAdmin) {
+      $allStaffItem =  new GenericTextItem('All Staff');
+      $actionsGroup->addItem($allStaffItem);
+    }
+
+    if (!$isAdmin) {
+      $regions = $this->contactRightsService->getContactRightsByRegions($this->contactUserInfo['contact_id']);
```
davialexandre: This is a bit too long. the line can be shortened if you create a method like `getContactRegions()` that, internally, call the Contact Rights Service method. The same can be done for the locations. To make your life easier when create the list items, these new methods could return only a list of regions/locations labels, since that's what you really need here

contactaccessrights/CRM/Contactaccessrights/Service/ContactRights.php
```diff
@@ -0,0 +1,85 @@
...
+    ON og.name = rights.entity_type AND og.name = %1
+
+    INNER JOIN civicrm_option_value ov
+    ON ov.id = rights.entity_id
+
+    WHERE rights.contact_id = %2";
+
+    $entityType = $rightType->getEntityType();
+    $queryParams = [
+      1 =>[$entityType, 'String'],
```
davialexandre: missing space before [

contactaccessrights/contactaccessrights.php
```diff
@@ -190,3 +196,46 @@ function contactaccessrights_civicrm_pageRun($page) {
...
+
+  $userPermissions = CMSUserPermissionsFactory::create();
+
+  if (!$userPermissions->check($contactUserInfo, ['access CiviCRM'])) {
+    return;
+  }
+
+  $contactRightsService = new ContactRightsService();
+  $aclGroups = ContactHelper::getACLGroups($contactID);
+
```
davialexandre: Please remove one of these two empty lines here

contactaccessrights/contactaccessrights.php
```diff
@@ -190,3 +196,46 @@ function contactaccessrights_civicrm_pageRun($page) {
...
+  $contactUserInfo = ContactHelper::getUserInformation($contactID);
+
+  if (empty($contactUserInfo['cmsId'])) {
+    return;
+  }
+
+  $userPermissions = CMSUserPermissionsFactory::create();
+
+  if (!$userPermissions->check($contactUserInfo, ['access CiviCRM'])) {
```
davialexandre: Tbh, I was expecting this check to be inside the `CRM_Contactaccessrights_Helper_ContactActionsMenu_ContactAccessActionGroup` class, as it already has some other permissions checks. I think it will the easier to understand things if all the permissions checks are in one single place

contactaccessrights/contactaccessrights.php
```diff
@@ -190,3 +196,46 @@ function contactaccessrights_civicrm_pageRun($page) {
...
+
+  $contactRightsService = new ContactRightsService();
+  $aclGroups = ContactHelper::getACLGroups($contactID);
+
+
+  $contactAccessActionGroup = new ContactAccessActionGroupHelper(
+    $contactUserInfo,
+    $contactRightsService,
+    $userPermissions,
+    $aclGroups
```
davialexandre: I'm not sure if it's a good idea to pass the aclGroups to the ActionGroup like that, because, depending on how it's used it can lead to unexpected behaviors. For example, what happens if I pass an arrays of ACL groups for a contact B, while the information on `$contactUserInfo` is for contact A? Ideally, it would be better if we had some sort of ACL Service, which would expose a method to get the ACL groups of a contact. We would then inject that service here

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessGroupTest.php
```diff
@@ -0,0 +1,188 @@
...
+
+    $contactAccessActionGroup = $contactAccessActionGroup->get();
+    $menuItems = $contactAccessActionGroup->getItems();
+    //Four Items are expected: Region List, Locations List, ACL Group
+    //List item and the Manage Regional Access Button
+    $this->assertCount(4, $menuItems);
+    $this->assertInstanceOf(UserRegionsListItem::class, $menuItems[0]);
+    $this->assertInstanceOf(UserLocationsListItem::class, $menuItems[1]);
+    $this->assertInstanceOf(UserACLGroupsListItem::class, $menuItems[2]);
+    $this->assertInstanceOf(ActionsGroupButtonItem::class, $menuItems[3]);
```
davialexandre: mmmm... so this is probably why you've created the empty classes. I still don't think it's a good idea. I know we don't want to test the whole markup, so maybe we can call the render() method and check that the string contains the group, regions or location labels?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactTest.php
```diff
@@ -0,0 +1,82 @@
...
+    //API call mock to assign Contact to a group
+    $this->adhocProvider->addAction('get', 'access CiviCRM', function ($apiRequest) use($group, $contactID) {
+      return $this->getGroupContactAPIReturnValue($group, $contactID);
+    });
+
+
+    $result = ContactHelper::getACLGroups($contactID);
+    $this->assertEquals([], $result);
+  }
+  private function createACLGroup() {
```
davialexandre: Please add one empty line between methods

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactTest.php
```diff
@@ -0,0 +1,82 @@
...
+    $params = [
+      'sequential' => 1,
+      'title' => 'Test',
+      'group_type' => 'Access Control'
+    ];
+
+    if (!$isACLGroup) {
+      $params['group_type'] = 'Mailing List';
+    }
+    $result = civicrm_api3('Group', 'create', $params);
```
davialexandre: If you're using mocks, why do you need to create this group? Couldn't you mock this too?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
```diff
@@ -0,0 +1,137 @@
...
+        'entity_id' => $region2['id'],
+        'label' => $region2['label'],
+        'value' => $region2['value']
+      ]
+    ];
+
+    $this->assertEquals($expectedArray, $rights);
+  }
+
+  private function createOptionValue($optionGroupName, $value) {
```
davialexandre: Please don't mix helper methods with tests. Put the helper methods at the end of the file, after all the test methods

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
```diff
@@ -0,0 +1,137 @@
...
+        'label' => $location2['label'],
+        'value' => $location2['value']
+      ]
+    ];
+
+    $this->assertEquals($expectedArray, $rights);
+  }
+
+  public function testGetContactRightsByRegions() {
+    $contact =['id' => 5];
```
davialexandre: Missing space before [

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
```diff
@@ -0,0 +1,137 @@
...
+      'name' => $value,
+      'label' => $value,
+      'value' => $value
+    ];
+
+    return OptionValueFabricator::fabricate($params);
+  }
+
+  public function testGetContactRightsByRegionsReturnsEmptyWhenContactHasNoRightsToAnyRegion() {
+    $contact =['id' => 5];
```
davialexandre: missing space

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+    $rights = $this->contactRightsService->getContactRightsByRegions($contact['id']);
+    $this->assertEquals([], $rights);
+  }
+
+  public function testGetContactRightsByLocationsReturnsEmptyWhenContactHasNoRightsToAnyLocation() {
+    $contact =['id' => 5];
```
davialexandre: missing space

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
```diff
@@ -0,0 +1,137 @@
...
+        'id' => $regionRights2['id'],
+        'contact_id' => $contact['id'],
+        'entity_type' => $this->regionEntityType,
+        'entity_id' => $region2['id'],
+        'label' => $region2['label'],
+        'value' => $region2['value']
+      ]
+    ];
+
+    $this->assertEquals($expectedArray, $rights);
```
davialexandre: What about Rights by Location?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
```diff
@@ -1,116 +1,100 @@
...
-      'value' => $value
-    ];
-    $result = OptionValueFabricator::fabricate($params);
-    $result['entity_type'] = $optionGroup;
-    
-    return $result;
-  }
-
-}
+<?php
```
davialexandre: I wonder why github is saying that this whole file was updated. Do you have any ideas?

contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php
```diff
@@ -30,57 +24,4 @@ public static function create($params) {
 
     return $instance;
   }
```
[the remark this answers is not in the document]
tunbola: Another reason is because it being a service allows it to be injected in the ContactAcessRightsMenu Group and makes it easy to be mocked in tests tests. Also I believe the method is more suited for a Service than a BAO because it involves getting information that is not related to the Rights DAO alone but joining to some other tables.

contactaccessrights/CRM/Contactaccessrights/Component/ContactActionsMenu/GenericTextItem.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use CRM_HRContactActionsMenu_Component_GroupItem as ActionsGroupItemInterface;
+
+/**
+ * Class GenericTextItem
+ */
+class CRM_Contactaccessrights_Component_ContactActionsMenu_GenericTextItem implements ActionsGroupItemInterface {
```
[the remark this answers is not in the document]
tunbola: Good Idea

contactaccessrights/CRM/Contactaccessrights/Component/ContactActionsMenu/GroupTitleToolTipItem.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class CRM_Contactaccessrights_Component_ContactActionsMenu_GroupTitleToolTipItem extends GenericTooltipItem {
+
+  /**
+   * Returns the Tooltip Text.
+   *
+   * @return string
+   */
+  public function getToolTipText() {
+    $toolTipText = 'You can specify a users access to staff members by
```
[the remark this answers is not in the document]
tunbola: Yeah, it should be user's access, I saw the answer to the question somewhere too.

contactaccessrights/CRM/Contactaccessrights/Component/ContactActionsMenu/UserACLGroupsListItem.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use CRM_Contactaccessrights_Component_ContactActionsMenu_GenericListItem as GenericListItem;
+
+class CRM_Contactaccessrights_Component_ContactActionsMenu_UserACLGroupsListItem extends GenericListItem {
+
+}
```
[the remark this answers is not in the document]
tunbola: We can but I thought of the following advantages:
- It will be easy to check the items in tests
- Also if any of the items need to be displayed differently later, it would be easier.  e.g its possible that the UserLocations item should be linked to their locations page and the User regions items be linked to the region page or the label for one is to be red and the other blue. 
The way it is, it would be easy to just add the render function for the ListItem that needs to be different in the future and nothing needs to be changed in the Action group helper class and tests.

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/Contact.php
```diff
@@ -0,0 +1,79 @@
...
+        'contact_id' => $contactID,
+      ]);
+
+      //When a contact has CMS account, the cmsId parameter is needed for the
+      //HRCore CMSData classes
+      $output['cmsId'] = $result['id'];
+      $output['name'] = $result['name'];
+    } catch(Exception $e) {}
+
+    return $output;
```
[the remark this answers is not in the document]
tunbola: Its in the contact access extension. Probably I can just use the function directly here, is that fine?

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/Contact.php
```diff
@@ -0,0 +1,79 @@
...
+  public static function getACLGroups($contactID) {
+    $result = civicrm_api3('GroupContact', 'get', [
+      'sequential' => 1,
+      'contact_id' => $contactID,
+      'api.Group.get' => ['id' => "\$value.group_id", 'sequential' => 1],
+    ]);
+
+    $aclGroups = [];
+
+    if ($result['count'] < 0) {
```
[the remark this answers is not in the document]
tunbola: (facepalm)
On second thought, Civi is full of surprises.

contactaccessrights/contactaccessrights.php
tunbola: good Idea.

contactaccessrights/contactaccessrights.php
tunbola: Ok, I can add a service for that. Ideally, where would be the best place or extension to put such a service?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessGroupTest.php
tunbola: Not only that, also incase we need to change the markup for a list item, we can do that with minimal change in the future. 
The render method would spit out html and I wonder how we would check that for the labels, hoping it won't be messy.

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
tunbola: Not sure what you mean, there's a test for Rights by Location.

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
tunbola: tbh, I was surprised when i saw such a huge change, i only made the tests class to extend BaseHeadless and I saw a lot of changes in the diff when making the PR

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactTest.php
tunbola: For some reason, it seems I can only setup one `adhocProvider` per test, I took a look at Core tests and I could not find examples where two `adhocProvider` were added for a test.

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Service/ContactRightsTest.php
davialexandre: Oh, sorry. I completely missed the other test

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Utils/ACLTest.php
davialexandre: Maybe when you edited the file the editor also updated the newline character? I can see that, on staging, this file uses CRLF while most of the other files use LF. What is it using after your change?

contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php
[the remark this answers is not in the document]
davialexandre: It only joins with option value/groups tables so that it can filter by region/location. We do the same thing in L&A to filter Requests by status and all that is done in the BAO, not in the Service. In the end, it's just a method to return a list of Rights given certain criteria. The BAO should be responsible to know how to do that (think of BAOs as gateways to the database). Unfortunately, whoever implemented this made a few mistakes like return an array instead of objects and also including the option value information in the response. Ideally, this is what should have been fixed, but you don't need to worry about that for now.

About the tests, you can inject the BAO the same way as you've injected the service. This is just an instance method

contactaccessrights/CRM/Contactaccessrights/Component/ContactActionsMenu/UserACLGroupsListItem.php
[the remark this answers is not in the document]
davialexandre: There's a difference in changing the class structure to make it easier to test them (this usually reduces coupling) and creating an empty class that isn't  used for anything but the test. In the future, if it really happens that we need a different type of list, we can just create a separate class for that.

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/Contact.php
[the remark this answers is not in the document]
davialexandre: Yeah, it's ok. After all, this extension already depends on the contact actions extension

contactaccessrights/contactaccessrights.php
davialexandre: CRM_ContactActionsMenu_Service_ACLService?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessGroupTest.php
davialexandre: You don't care about the markup. You only care if the string contains the label. Something like `assertContains('Region 1', $item->render())`

contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php
```diff
@@ -1,12 +1,10 @@
...
+use CRM_Contactaccessrights_Utils_RightType_Region as RegionRightType;
+use CRM_Contactaccessrights_Utils_RightType_Location as LocationRightType;
+
 class CRM_Contactaccessrights_BAO_Rights extends CRM_Contactaccessrights_DAO_Rights {
-  /**
-   * Rights grouped by type.
-   *
-   * @var array
-   */
-  private $rights = [];
```
davialexandre: Why was this removed?

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessActionGroup.php
```diff
@@ -0,0 +1,198 @@
...
+   * added.
+   *
+   * @return ActionsGroup
+   */
+  public function get() {
+    $actionsGroup = new ActionsGroup($this->getGroupTitle());
+    $hasAccessCivicrm = $this->hasPermission(['access CiviCRM']);
+
+    if (!$hasAccessCivicrm) {
+      return $actionsGroup;
```
davialexandre: At this point, the group already has a title. Wouldn't this make the title to be rendered on the menu?

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessActionGroup.php
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Gets Contact Access Menu Group with menu items already
+   * added.
+   *
+   * @return ActionsGroup
+   */
+  public function get() {
+    $actionsGroup = new ActionsGroup($this->getGroupTitle());
+    $hasAccessCivicrm = $this->hasPermission(['access CiviCRM']);
```
davialexandre: `$this->hasPermission(['access CiviCRM']);` is readable is enough. Do we really need to store this in a variable that is only used once?

contactaccessrights/tests/phpunit/CRM/Contactaccessrights/BAO/RightsTest.php
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Class CRM_Contactaccessrights_BAO_RightsTest
+ *
+ * @group headless
+ */
+class CRM_Contactaccessrights_BAO_RightsTest extends BaseHeadlessTest {
+
+  private $regionEntityType = 'hrjc_region';
+
+  private $locationEntityType = 'hrjc_location';
```
davialexandre: Instead of hardcoding the types here, please use the values from the `CRM_Contactaccessrights_Utils_RightType_Location` and `CRM_Contactaccessrights_Utils_RightType_Region` classes

contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php
tunbola: It was no longer necessary, The `getRightsByType` method used it to store values in an array initially before returning the result, I rewrote that part not to need this variable anymore.
See: https://github.com/civicrm/civihr/blob/c92225222b9f4fb0461191568883e9bee6df4d24/contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php#L84

contactaccessrights/CRM/Contactaccessrights/Helper/ContactActionsMenu/ContactAccessActionGroup.php
tunbola: I added a check here to not allow that: https://github.com/civicrm/civihr/blob/d0e41964f5d75e4312481af08fc8eca943a328d1/contactaccessrights/contactaccessrights.php#L235

contactaccessrights/CRM/Contactaccessrights/BAO/Rights.php
davialexandre: But why did you rewrite it? Was it causing any problem? Besides, after this refactoring, I don't think the array needs to be indexed by `$entityType` anymore

PR #2472, opened 2018-02-14
(proof-of-concept) refactor details tab spec

deb1990:
### Problem
The "leave-request-popup-details-tab.component.spec.js" is too big, so the idea is to divide this file into smaller chunks.

### Solution
1. `leave-request-popup-details-tab.component.spec.js` contains the unit tests for all the common tests i.e. the unit tests for `leave-request-popup-details-tab.component.js` file. 

2. `request-modal-details-leave.controller.spec.js`, `request-modal-details-sickness.controller.spec.js` and `request-modal-details-toil.controller.spec.js` has been created which contains all the unit tests for their respective type of request.

3. `request-modal-helper.js` is created which contains the common helper functions from `leave-request-popup-details-tab.component.spec.js`, `request-modal-details-leave.controller.spec.js`, `request-modal-details-sickness.controller.spec.js` and `request-modal-details-toil.controller.spec.js`.

4. Generic helper functions like `getStatusValueFromName` and `getUTCDate `, has been moved to `helper.js`.

5. Unit Test added in `leave-request-popup-details-tab.component.spec.js` to test if child controller is overriding the parent component controller. Except that, No new test has been added to limit the scope of the ticket.


igorpavlov:
As already mentioned above, there should be a map `src<->spec`. 

1) So at first I think that the component (lets call it parent controller) should become fully standalone (should not depend on children). It may be already like that, but we need to doublecheck.

2) (!) If we want a map of files to do TDD here properly, we need to be sure that changing parent controller *won't break children controllers without failing test on the parent controller*. Example is the extending functions, at least they should be tested that they are called if they exist in `vm`.

3) Now we are ready to have spec files per children controllers. To be honest, I don't know if they should be completely standalone and not depend on the parent controller or they can (cc @AkA84 @reneolivo). But anyway, now it is safe to move tests to child controllers since the parent controller is protected by extra tests.

deb1990:
@AkA84 @reneolivo @igorpavlov I have updated the PR as per your feedback and documentation has been updated too.

deb1990:
Closed as POC is approved.

PR #2484, opened 2018-02-16
PCHR-3260: Automate `npm install` on every CiviHR extensions

AkA84:
## Overview
This PR adds a two npm scripts

### `sub-install`
This is a `postinstall` script in the root `package.json`, which programmatically runs `npm i` in every CiviHR extension folder with a `package.json` file.

The reason to have this script is multiple:
1. Makes it easier for devs to install all the dev dependencies that he might need when working on the project, without having to run `npm install` manually for each extension
2. It simplifies CI scripts (see changes in `Jenkinsfile`)
3. It's another small step toward a single CiviHR extension

### `clean-dependencies`
This is the reverse of `sub-install`, deleting the node_modules/ folder in both the root and in every sub-folder

## Technical Details
1. The appraisals extension is obsolete and is planned to be deleted, so its `package.json` file had been removed ahead of time so that the script can skip it
2. The hremails extension didn't have its `package-lock.json` file pushed in the repo, this PR fixes that
3. Also, still in hremails, the email-templates/ folder also didn't have a lock file, and the `gulp-sass` that was using was too old for Node LTS


package.json
```diff
@@ -10,7 +10,11 @@
     "url": "https://github.com/civicrm/civihr/issues"
   },
   "scripts": {
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "postinstall": "npm run sub-install",
+    "sub-install": "for file in `npm run folders-with-dependencies --silent`; do cd `dirname $file`; npm i; cd $OLDPWD >> /dev/null; done",
```
davialexandre: oh, I didn't know about this $OLDPWD. Nice!

PR #2485, opened 2018-02-19
(POC) Assigning unique urls to tabs of angular apps in contact summary page

AkA84:
## Problem
There is no way to have sub-routes/sub-states in the angular apps of the contact summary page. This has started to affect the ux with the introduction of the Leave tab, which has 4 tabs that do not have a unique url attached to them. That means that a user cannot land directly on a specific tab

before

Technically the problem is that we have multiple angular apps "live" on the same page, all of them setting up their own routes and, most importantly, all of them setting up the default (`.otherwise()`) route
```js
// job contract
angular.module('job-contract.config', ['job-contract.constants']).config(config);

function config () {
  $routeProvider
    // ...
    .otherwise({ redirectTo: '/' });
}
```
```js
// job roles
angular.module('hrjobroles.config', ['hrjobroles.constants']).config(hrJobRolesConfig);

function hrJobRolesConfig () {
  $routeProvider
    // ...
    .otherwise({ redirectTo: '/' });
}
```
```js
// contact summary
angular.module('contactsummary.config', ['contactsummary.constants']).config(config);

function config () {
  $routeProvider
    // ...
    .otherwise({redirectTo: '/'});
}
```

The `otherwise` route is the part of the problem. Consider this scenario:

1. User lands on the tab "Foo", which contains the `foo` angular app.
2. `foo` defines one route, `/`. Any other route should `.otherwise` to `/`
3. User moves to the tab "Bar", which contains the `bar` angular app.
4. `bar` defines the routes `/abc` and `/def`. Any other route should `.otherwise` to `/abc`
5. Because `foo` was visited earlier, while `bar` will try to automatically go to `/abc`, the route will not be recognized as valid, and as such the browser will go back to `/`, making the routes of `bar` inaccessible to the user

Basically the first app to be loaded, by using `.otherwise` will make impossible to go to the routes defined by any subsequent app that will be loaded by the user by clicking in another tab 

By getting rid of the `.otherwise` route in any of the tabs angular apps the problem of "routes exclusivity" is solved, but another one is introduced. Consider this scenario:

1. User lands on the tab "Foo", which contains the `foo` angular app.
2. `foo` defines one route, `/`. There is no default route forced via `.otherwise`
3. Assuming that the url hash fragment is `#/`, `foo` correctly works, simply by coincidence (the current hash fragment matches its only route, `/`)
3. User moves to the tab "Bar", which contains the `bar` angular app.
4. `bar` defines the routes `/abc` and `/def`. There is no default route forced via `.otherwise`
5. Because `bar` expects the `/abc` route to be the default one, but cannot use `.otherwise` to force the browser to go there, the active route stays `/` (because the url hash fragment is still `#/`) and `bar` does not load correctly

So any solution would have to
1. Not use `otherwise`
2. Find an alternative mechanism to go to the default route of each apps

## Solution
The solution I've found is to leverage jQuery UI's [tabs](https://jqueryui.com/tabs/) the plugin used by civi to handle the contact summary tabs, and it works in two steps:

1. If the app of a tab requires a specific route to be the default one, instead of `/`, then the route will be defined in the `hook_civicrm_tabset` by explicitly putting the route in the tab's url
    ```js
    function bar_civicrm_tabset($tabsetName, &$tabs, $context) {
      $tabs[] = [
        'id' => 'bar-tab',
        'url' => CRM_Utils_System::url('civicrm/contact/view/bar', ['cid' => $contactId], FALSE, '/abc'),
      ]
    }
    ```
2. The `hrui.js` script will add custom logic to the contact summary tabs. The logic simply being: if the url of the active or about-to-be-active tab has an hash fragment, then add it to the url, otherwise default to `/`
    ```js
    $('#mainTabContainer')
      .on('tabscreate', function (event, ui) {
        updateHash(ui.tab);
      })
      .on('tabsbeforeactivate', function (e, ui) {
        updateHash(ui.newTab);
      });

    function updateHash (tab) {
      var $newTab = $(tab);
      var hash = $newTab.find('a').prop('hash');

      if (hash) {
        window.location.hash = hash;
      } else {
        window.location.hash = '/';
      }
    }
    ```
    The reason for why the hash has to be set manually to `window.location.hash` if it's already set to the tab url is that the url of the tab is used to load via ajax a page fragment, *not* to redirect the browser. 
    The reason for why the hash fallbacks to `/` instead of `null`, is to avoid the page to jump to the top every time the hash is reset (= if you put `hash = null`, angular's router will automatically put back the `#`, causing the browser to immediately skip to the top of the page)

after

### Pros
* Despite having to do with the contact summary page, not core override is needed, we can do everything we need within the constraints of CiviHR
* The work required is minimal: just remove any `otherwise` and explicitly add whatever route that is not `/` a specific app needs to land on automatically
* As far as i could tell, it works in every scenario: user lands directly on an app with custom routes; user lands on an app with `/` route and then moves over the the app with custom routes; user lands on an app with `/` route but the url has already some other route in the url hash fragment; etc

### Cons
* Not a big deal, but usually you expect an app to define an `otherwise` rule
* The backend will dictated the default route of an angular app, breaking the separation of concerns principle


hrui/js/src/hrui.js
```diff
@@ -18,6 +26,17 @@
       toggleActiveClassOnHoverOnAnyMainMenuItem();
     });
 
+  function updateHash (tab) {
+    var $newTab = $(tab);
+    var hash = $newTab.find('a').prop('hash');
+
+    if (hash) {
```
deb1990: `window.location.hash = hash || '/'`
(no need to fix in POC, just FYI)

deb1990:
url

I see one problem, if I type a wrong url, the content is gone as there is no redirect. This scenario is unlikely but still we somehow need to fix it.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -370,7 +370,7 @@ function hrleaveandabsences_civicrm_tabset($tabsetName, &$tabs, $context) {
     $contactId = $context['contact_id'];
     $tabs[] = [
       'id'        => 'absence',
-      'url'       => CRM_Utils_System::url('civicrm/contact/view/absence', ['cid' => $contactId]),
+      'url'       => CRM_Utils_System::url('civicrm/contact/view/absence', ['cid' => $contactId], FALSE, '/absence-tab/report'),
```
deb1990: Backend should not know which is the default tab, instead that can be handled by frontend. So here the code should be 
`'url'       => CRM_Utils_System::url('civicrm/contact/view/absence', ['cid' => $contactId], FALSE, '/absence-tab'),`

then in the frontend we can use any of solutions proposed in https://stackoverflow.com/questions/29491079/redirect-a-state-to-default-substate-with-ui-router-in-angularjs

Specially I like this one `As of version 1.0 of the ui-router it supports a redirectTo property.`

AkA84:
@deb1990 

> I can think of one idea, right now when we open tab A and then go to tab B, we still keep the tab A and its HTML+JS in the memory, which is causing this particular issue. Also it increases memory usage.
>
> Instead if we remove no active tabs(ng-if ?) then the whole problem with routing will be solved.

I tried different solutions, that included. Unfortunately it doesn't work, if you remove the angular app from the DOM (the only method to "unbootstrap" an angular module), its routing settings are still applied. I guess it's because the services are singleton, and all the apps are using the same router instance

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
AkA84: > Specially I like this one https://stackoverflow.com/a/46439298/2400594

Unfortunately we are not using ui-router v1, we are at v0.2.15

igorpavlov:
@AkA84 the bug reported by @deb1990 is a bit more concerning. If you visit `/civicrm/contact/view?cid=204&selectedChild=absence#/` then it will fail as well. You can reproduce this bug naturally:

1. visit `/civicrm/contact/view?cid=204&selectedChild=absence#/absence-tab/report`
2. navigate to Job Roles tab
3. navigate back to Leave tab - broken page.

PR #1473, opened 2017-02-17
PCHR-1356: Replace Government ID Field for NI / SSN field

MiyaNoctem:
## Problem
There was a pseudo-field being shown on contacts called "Government ID" which was being used to create a record for the contact in the Identity custom group.  The requirement now was to replace the pseudo-field for a proper custom field associated directly to Individuals.

before

### Backend 
Implemented upgrade that ceates custom group "Inline Custom Data" and "NI / SSN" field. Also, added JS functionality that moves fields on this group to Contact Details group. Removed backend logic that added Government ID to contact edit view.

### Frontend
Remove, `colspan="3"` form  customFields `td` in "hrui.js" and override styles for `custonFields' width`. Added BackstopJS scenarios in `org.civicrm.bootstrapcivicrm/scss/civicrm/contact/pages/_new-individual.scss`

after

1. Fixed size of "NI / SSN" field:
screen shot 2017-03-06 at 6 30 31 pm

2. BackstopJS Scenarios Test:
screencapture-file-users-bmn-buildkit-build-hr16-sites-all-modules-civicrm-tools-extensions-civihr-uk-co-compucorp-civicrm-hrcore-backstop_data-html_report-index-html-1488803890531

PR #2487, opened 2018-02-20
PCHR-2983: Fix NI/SSN block placement

AkA84:
## Overview
This PR fixes two problems

### Inconsistent placing of NI/SSN fields set in personal details tab
The NI/SSN fields (grouped in the "Inline Custom Data" custom fields set) were introduced to replace the old "Government ID" field (see #1473 ), but they were placed inconsistently in the "summary" tab of the contact summary page.

Sometimes they were placed right below the first block of informations (which was the intended placement, see the "after" screenshot), while sometimes they were at the bottom of the page, along with the rest of the custom fields sets (incorrect, see "before" screenshot)

The difference was due to a condition in the code (see "Technical Details") that was put in place for the old "Government ID" field, but that was made obsolete by the changes introduced in the aforementioned PR. Unfortunately the PR didn't remove that condition, which was left there unnoticed.

This PR fixes the issue by simply removing the condition

### Wrong placing of NI/SSN fields in contact edit form
The NI/SSN fields are moved, via js, from the bottom of the contact form to near the top. Unfortunately the script wasn't accurate enough and the fields ended up being placed incorrectly.

The PR fixes the script.

## Before
before

before-fields

## After
after

after-fields

### Details about obsolete condition
The condition was whether there was a "is_government" custom field belonging to a "Identify/Identification" custom fields set defined in the site. This was checked via the api, in the backend
```php
$hideGId = civicrm_api3('CustomField', 'getvalue', array('custom_group_id' => 'Identify', 'name' => 'is_government', 'return' => 'id'));
```
the outcome of the api call was then put in the `CRM` global object
```php
CRM_Core_Resources::singleton()->addSetting(array('hideGId' => $hideGId));
```
which was finally used by `hrui.js` to decide whether or not to move the block
```js
function addGovernmentIdField () {
  if (CRM.cid && CRM.hideGId) {
    // move the block
  }
}
```

### Erratic permissions on hrui.min.js
For some reason i wasn't able to figure out, the permissions of the hrui.min.js file kept changing from `644` to `755` and back when the min file was updated (but randomly though, not every time). Here [are](https://github.com/civicrm/civihr/commit/a9fbd52a9bea67bc7e858423786deba95b365d18#diff-65b33d9256a6a24a5aef531afb640320) [a](https://github.com/civicrm/civihr/commit/bdb90922b3938bc44253d5b57b7d9ed1a9d9013c#diff-65b33d9256a6a24a5aef531afb640320) [few](https://github.com/civicrm/civihr/commit/607db21f8a92a7b1a0e69e2f1ee7c6d0c5e3f1b5#diff-65b33d9256a6a24a5aef531afb640320) [examples](https://github.com/civicrm/civihr/commit/c73699179b11abdecc1a393d56ad04fe4d118127#diff-65b33d9256a6a24a5aef531afb640320)

Given that I wasn't able to pinpoint the exact reason of why this is happening, I decided to just force the permissions whenever the min file is generated:
```js
.pipe(gulp.dest(path.join(__dirname, '..', 'js/dist'), { mode: '0644' }));
```


hrui/js/src/hrui.js
```diff
@@ -214,6 +182,37 @@
...
+  /**
+   * Manipulates, at the DOM level, the blocks/fields belonging to the
+   * Inline Custom Data custom fields set
+   */
+  function manipulateDOMOfInlineCustomData () {
+    // In Personal Details Page
+    var $inlineDataBlock = $('.Inline_Custom_Data');
+
+    if ($inlineDataBlock.length) {
+      $inlineDataBlock
```
reneolivo: Is it possible to move this block inside its own function so we can explain what's going on?

For example:

```js
function manipulateDOMOfInlineCustomData () {
  // ...

  if ($inlineDataBlock.length) {
    doSomethingPrettyWithTheInlineCustomData(); // I have no idea what this does....
  }

  if ($('#customFields').length < 1) {
    wrapEachCustomDataLabelInsideATd(); // for example
  }
}
```

reneolivo:
We need to investigate why `gulp build --ext hrui` changes the file permissions of the minified file:

pchr 2983 fix ni ssn block placement by aka84 pull request 2487 civicrm civihr

PR #2492, opened 2018-02-22
PCHR-3047: Contract update entitlements redirection control

reneolivo:
## Overview
This PR changes the way the leave entitlements update confirmation is displayed when editing job contracts.

The confirmation is now only displayed when the contract's start or end date change, or when the contract's entitlements change. In this case, and if the user confirms the prompt, the user would be redirected to the entitlements manager screen, otherwise it would simply display a success message.

## Before
anim
The entitlement prompt is shown even if nothing changed.

## After
anim

## Technical Details

The Contract modal has a `checkIfEntitlementFieldsChanged` function that determines if the fields relevant for entitlement update have changed:

```js
function checkIfEntitlementFieldsChanged () {
  var hasEndDateChanged = hasContractDateChanged('period_end_date');
  var hasStartDateChanged = hasContractDateChanged('period_start_date');
  var haveEntitlementsChanged = !angular.equals($scope.entity.leave,
    entity.leave);

  return hasStartDateChanged || hasEndDateChanged || haveEntitlementsChanged;
}

function hasContractDateChanged (dateName) {
  var currentDate, currentDateIsDifferentFromOriginal, originalDate,
    oneisEmptyAndTheOtherIsNot;

  originalDate = entity.details[dateName];
  currentDate = $scope.entity.details[dateName] !== ''
    ? $scope.entity.details[dateName]
    : null;

  oneisEmptyAndTheOtherIsNot = (currentDate === null || originalDate === null) &&
    currentDate !== originalDate;
  currentDateIsDifferentFromOriginal = moment.isDate(currentDate) &&
    !moment(originalDate).isSame(moment(currentDate), 'day');

  return oneisEmptyAndTheOtherIsNot || currentDateIsDifferentFromOriginal;
}
```

This function is used to pass a `haveEntitlementFieldsChanged` parameter when the modal is closed:

```js
function save () {
  // ...

  thisIsTheSaveSequence.then(function () {
    // ...
    results.haveEntitlementFieldsChanged = checkIfEntitlementFieldsChanged();

    return results;
  }).then(function (results) {
    $scope.$broadcast('hrjc-loader-hide');
    $modalInstance.close(results);
    pubSub.publish('Contract::updated');
  }
}
```

The Contract controller uses this parameter to determine if the prompt should be shown or if a simple success notification is to be displayed instead:

```js
if (results.haveEntitlementFieldsChanged) {
  utilsService.updateEntitlements($scope.contract.contact_id);
} else {
  notificationService.success('CiviHR', 'Contract updated');
}
```

## Comments
* Test pass.
* Lint issues were fixed.
* Added some missing tests for Contract modal and missing documentation in general.
* Refactored the Contract Modal as much as the ticket allowed.


igorpavlov:
As per GIF, the title of the success message is "CiviHR", which is a default one AFAIK. Could we please change it to "Success" to keep it consistent with other messages on this page? Please confirm with PM first.

igorpavlov:
> This function is used to pass a ................... parameter when the modal is closed:

Was it a placeholder that should have been filled in later?

hrjobcontract/js/src/job-contract/controllers/contract.controller.js
```diff
@@ -292,7 +292,12 @@ define([
         }
 
         CRM.refreshParent('#hrjobroles');
-        utilsService.updateEntitlements($scope.contract.contact_id);
+
+        if (results.haveEntitlementFieldsChanged) {
+          utilsService.updateEntitlements($scope.contract.contact_id);
+        } else {
+          notificationService.success('CiviHR', 'Contract updated');
```
igorpavlov: Should we change 'CiviHR' to 'Success' to keep it consistent?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -269,6 +186,30 @@ define([
       return modalConfirmEdit.result;
     }
 
+    /**
+     * Confirms that the contract change is valid and saves it.
+     */
```
igorpavlov: 1. Please add params description.
2. Can we make this func a promise?
3. This func should be renamed: validateContractAndSave()

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -252,6 +148,27 @@ define([
       return modalChangeReason.result;
     }
 
+    /**
+     * Determines if the contract entitlements have beem modified. These include
```
igorpavlov: beem

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -49,36 +50,31 @@ define([
...
         $rootScope.$broadcast('hrjc-loader-hide');
       });
     }());
 
-    $scope.cancel = function () {
+    /**
+     * Hides the job contract modal, but before doing so asks for confirmation
+     * before changes are lost.
+     */
+    function cancel () {
```
igorpavlov: The func name is 'cancel', but it 'Hides the job contract modal, but before doing so asks for confirmation before changes are lost.' Can we rename this func? promptAboutChangesLossAndCloseOnConfirm()

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -269,6 +186,30 @@ define([
...
+        }
+      }, function (reason) {});
+    }
+
+    /**
+     * Saves all the tabs in the contract's form.
+     *
+     * @return {Promise} resolves when all the tabs are saved.
+     */
     function contractEdit () {
```
igorpavlov: Function name sounds wrong. Could it be "saveAllContractSections" ?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
       });
     }
 
-    function contractChange (reasonId, date) {
-      $scope.$broadcast('hrjc-loader-show');
+    /**
+     * Converts a date string into a Date object (if string is not empty)
```
igorpavlov: Please tell what it does if the string is empty:
```
* Converts a date string into a Date object.
* If the string is empty or invalid date, it ....
```

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+     *
+     * @todo This should probably happen inside the service that returns the data #
+     *
+     * @param {string} dateString
```
igorpavlov: String (capitalised)

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
       });
     }
 
-    function contractChange (reasonId, date) {
-      $scope.$broadcast('hrjc-loader-show');
+    /**
+     * Converts a date string into a Date object (if string is not empty)
+     *
+     * @todo This should probably happen inside the service that returns the data #
```
igorpavlov: I wonder what does the hash mean?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+    function filesValidate () {
+      var entityName, fieldName, i, len, uploaderEntity, uploaderEntityField, uploaderEntityFieldQueue;
+      var fileMaxSize = $scope.fileMaxSize;
+      var uploader = $scope.uploader;
+      var isValid = true;
+
+      for (entityName in uploader) {
+        uploaderEntity = uploader[entityName];
+
+        for (fieldName in uploaderEntity) {
```
igorpavlov: Can we use lodash here _.each + indexes?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+            $scope.filesValidate();
+          };
+        });
+      });
+    }
+
+    /**
+     * Determines if the given contract date has changed from its original value.
+     *
+     * @param {String} dateName either "period_start_date" or "period_end_date".
```
igorpavlov: Please add a space before {String}

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+
+      return originalDateIsEmptyAndCurrentIsNot || currentDateIsDifferentFromOriginal;
+    }
+
+    /**
+     * Initializes the form entity fields by having a copy of the form entity
+     * and converting date strings into date objects.
+     */
+    function initFormEntity () {
+      angular.copy(entity, $scope.entity);
```
igorpavlov: In that case I would add `var details = $scope.entity.details;` at the top, but this is up to you completely.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+     * Handles all the confirmations that need to be displayed when updating the contract.
+     */
+    function processContractUpdate () {
+      if (angular.equals(entity, $scope.entity) &&
+        angular.equals(files, $scope.files) &&
+        !$scope.uploader.details.contract_file.queue.length &&
+        !$scope.uploader.pension.evidence_file.queue.length) {
+        $scope.$broadcast('hrjc-loader-hide');
+        $modalInstance.dismiss('cancel');
+        return;
```
igorpavlov: a line break pls

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,32 +306,202 @@ define([
...
+          processContractUpdate();
         } else {
           CRM.alert(result.message, 'Error', 'error');
           $scope.$broadcast('hrjc-loader-hide');
         }
       }, function (reason) {});
+      $scope.$broadcast('hrjc-loader-hide');
     }
 
     function saveContractChange (reasonId, date) {
```
igorpavlov: It is a bit weird... We have functions 'save', 'contractEdit', 'saveContractChange'. Could you please provide a description for this func so at least we know the difference?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
reneolivo: The "entity" object has many properties, if I store each one in a different variable I'd need to change all the references. I would avoid that if I could :(

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
reneolivo: I'll change the name of the function, but I'll leave it stored in the scope as "cancel"  (`$scope.cancel = cancel`) because that's how it's referenced on the view and the view is shared by 2 different controllers.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
reneolivo: I think it means it needs an instance. The model should return an instance for the contract and the instance should have the dates converted to Date objects instead of strings.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
reneolivo: @igorpavlov I refactored the whole function though I could not add tests for this one. I made sure the functionality remained the same though:

anim

And when I say "the same" it means that new files number start at 1 instead of the last number in the queue, and uploaded files can't be seen until you refresh the window (I'll create ticket for those and add tests to this method on that ticket)

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
reneolivo: As a bonus: added new tests for cancel

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -307,12 +311,13 @@ define([
     }
 
     /**
-     * Converts a date string into a Date object (if string is not empty)
+     * Converts a date string into a Date object.
+     * If the date is not valid it returns the same date string provided.
      *
      * @todo This should probably happen inside the service that returns the data #
```
igorpavlov: Please add another `@todo` here, the function should not return `{Date|String}` but since it is like it was before, we lack test coverage and this PR is already big, lets postpone it till the next wave of refactoring.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -362,25 +362,15 @@ define([
...
-          i = 0;
-          len = uploaderEntityFieldQueue.length;
-          uploaderEntityField = uploaderEntity[fieldName];
-          uploaderEntityFieldQueue = uploaderEntityField.queue;
-
-          for (; i < len && isValid; i++) {
-            isValid = uploaderEntityFieldQueue[i].file.size < fileMaxSize;
+      _.each($scope.uploader, function (uploaderEntity) {
+        _.each(uploaderEntity, function (uploaderEntityField) {
+          for (var i = 0; i < uploaderEntityField.queue && uploaderEntityField.queue.length && isValid; i++) {
```
igorpavlov: How about that?
```js
if (uploaderEntityField.queue.length && isValid) {
  for (var i = 0; i < uploaderEntityField.queue; i++) {
}
```

reneolivo:
#### Updates:

1. The date comparison was changed from `moment.utc` to just `moment` since this would cause issues when converting a local date to a UTC date. This change from local to UTC could shift the date depending on the local timezone and the comparison would think the dates were different even if they originally were the same.
2. Fixed an scenario where the end date would be defined, but then removed. In this case the "Update Entitlements?"  modal should be displayed.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -406,13 +406,20 @@ define([
      * @return {Boolean}
      */
     function hasContractDateChanged (dateName) {
-      var originalDate = entity.details[dateName];
-      var currentDate = $scope.entity.details[dateName];
-      var originalDateIsEmptyAndCurrentIsNot = _.isEmpty(originalDate) && !_.isEmpty(currentDate);
-      var currentDateIsDifferentFromOriginal = moment.isDate(currentDate) &&
-        !moment.utc(originalDate).isSame(moment.utc(currentDate), 'day');
+      var currentDate, currentDateIsDifferentFromOriginal, originalDate,
+        oneisEmptyAndTheOtherIsNot;
```
igorpavlov: oneDateIsEmptyAndOtherIsNot

hrjobcontract/js/test/controllers/modal/modal-contract.controller.spec.js
```diff
@@ -227,6 +227,23 @@ define([
       });
     });
 
+    describe('when closing the modal after removing the period end date', function () {
+      beforeEach(function () {
+        MockContract.contractEntity.details.period_end_date = moment().format('YYYY-MM-DD');
+        makeController();
+        $scope.entity.details.period_end_date = '';
```
igorpavlov: Please add a line break before

hrjobcontract/js/test/controllers/modal/modal-contract.controller.spec.js
```diff
@@ -227,6 +227,23 @@ define([
...
+
+        $scope.save();
+        $scope.$digest();
+      });
+
+      it('returns a "have entitlement fields changed" variable set to true', function () {
```
igorpavlov: it either *sets* or *returns*. which one?

or you may make it "closes the modal and tells it that the entitlement fields have changed"

reneolivo:
Ideally a new test should be added to test that the date comparison is not done in UTC, but in local time. The test should go like this:

```js
describe('When the user is in India and closes the modal without any changes', function () {
  beforeEach(function () {
    moment.tz.setDefault('Indian/Cocos');

    $scope.save();
    $scope.$digest();
  });

  fit('returns a "have entitlement fields changed" variable set to false', function () {
    expect($uibModalInstanceMock.close).toHaveBeenCalledWith(jasmine.objectContaining({
      haveEntitlementFieldsChanged: false
    }));
  });
});
```
* note: that `jasmine.clock().mockDate(...)` is not being used since that's for native js dates, for moment we use `moment.tz.setDefault`.

Unfortunately, we don't have `moment-timezone` installed as a dependency and including it only for jasmine tests would require quite some changes:

* We'd need to define a `moment` dependency instead of `common/moment` (they could coexist)
* Karma would need to serve the moment-timezone dependency, but not include it.
* We'd need to pass the `civicrmPath` variable from karma.conf.js to the specific spec file in order to find the path for moment-timezone.

We'll skip this test for now until we can implement moment-timezone in production or for tests.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -406,13 +406,20 @@ define([
...
 
-      return originalDateIsEmptyAndCurrentIsNot || currentDateIsDifferentFromOriginal;
+      originalDate = entity.details[dateName];
+      currentDate = $scope.entity.details[dateName] !== ''
```
igorpavlov: Could it be `$scope.entity.details[dateName] || null` or does it need to be explicitly `''` to be `null`?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -406,13 +406,20 @@ define([
...
+        ? $scope.entity.details[dateName]
+        : null;
+
+      oneisEmptyAndTheOtherIsNot = (currentDate === null || originalDate === null) &&
```
igorpavlov: Do you really need to `currentDate === null` and not simply `currentDate || originalDate`?

reneolivo:
Commits https://github.com/civicrm/civihr/pull/2492/commits/d90cb063de8f1e7cf7ebc46181ef96ddc40d49eb and https://github.com/civicrm/civihr/pull/2492/commits/5c0bdf0ccadf28750e4947004756af6c7e48af84

### Before
anim


### After

### Technical details

hrjobcontract/js/test/controllers/modal/modal-contract.controller.spec.js
```diff
@@ -322,6 +334,53 @@ define([
...
+          ModalDialogController: 'change',
+          ModalChangeReasonController: {
+            reasonId: 1,
+            date: today
+          }
+        });
+        makeController({ action: 'change' });
+      });
+
+      describe('normal behaviour', function () {
```
igorpavlov: could you please change to 'basic tests' since when you read 'normal behaviour' you automatically ask yourself 'so is there is abnormal behaviour somewhere?' and there is nothing abnormal in errors.

hrjobcontract/js/test/controllers/modal/modal-contract.controller.spec.js
```diff
@@ -335,11 +339,13 @@ define([
     });
 
     describe('Saving the change revision', function () {
+      var changeReason = 123;
```
igorpavlov: changeReasonId

PR #2478, opened 2018-02-15
PCHR-3179: Check if balance change has been changed when changing request status via dropdown

igorpavlov:
## Overview

This PR adds a balance change check when performing certain actions, such as request approval, via Leave Request Actions dropdown that is used in Manager Leave Report and Staff Leave Report. If balance change changed, when approving a request it now requires a review of the updated balance breakdown. The balance change check is omitted for cancellation, rejection and deletion (logic not changed for these actions).

## Before

_Clicking on Approve option does not spawn any check._

## Technical Details

Leave Request Instance can now check if the balance has been changed.

```js
function checkIfBalanceChangeNeedsForceRecalculation () {
  var originalBalanceAmount, unitName;

  return AbsenceType.all({ id: this.type_id })
    .then(AbsenceType.loadCalculationUnits)
    .then(function (absenceTypes) {
      unitName = absenceTypes[0].calculation_unit_name;
    })
    .then(function () {
      return this.getBalanceChangeBreakdown();
    }.bind(this))
    .then(function (originalBalance) {
      originalBalanceAmount = originalBalance.amount;

      return this.calculateBalanceChange(unitName);
    }.bind(this))
    .then(function (balanceChange) {
      return +originalBalanceAmount !== +balanceChange.amount;
    });
}
```

Dialog has been augmented. Now dialog allows you to delay the options such as title, button classes, buttons captions and even loading. This is needed for this PR since we don't know if the balance change has changed and we don't know which dialog to show: the one that simply prompts the action confirmation or the one that prompts the balance breakdown review.

```js
function DialogService ($modal, $rootElement, $templateCache) {
    return {
      open: open
    };

    function open (options) {
      var $children;

      if (options && typeof options !== 'object') {
        return;
      }

      $children = $rootElement.children();

      return $modal.open({
        appendTo: $children.length ? $children.eq(0) : $rootElement,
        size: 'sm',
        controller: 'DialogController',
        template: $templateCache.get('dialog.html'),
        resolve: {
          props: function () {
            return options;
          }
        }
      }).result;
    }
  }
```

```js
function DialogController ($q, $scope, $modalInstance, props) {
    $scope.cancel = cancel;
    $scope.confirm = confirm;

    (function init () {
      assignProps(props);
      props.delayedProps && props.delayedProps().then(function (_props_) {
        assignProps(_props_);
      });
    }());

    function assignProps (props) {
      _.assign($scope, _.defaultsDeep(props, {
        title: 'CiviHR',
        msg: '',
        // ...
[...]

PR #2505, opened 2018-02-26
PCHR-3357: Check balance change when changing status via leave request dropdown (LiH)

igorpavlov:
## Overview
This PR introduces the same functionality as https://github.com/civicrm/civihr/pull/2478, but for **Leave in Hours**.

This PR adds a balance change check when performing certain actions, such as request approval, via Leave Request Actions dropdown that is used in Manager Leave Report and Staff Leave Report. If balance change changed, when approving a request it now requires a review of the updated balance breakdown. The balance change check is omitted for cancellation, rejection and deletion (logic not changed for these actions).

This PR also fixes the UI issue when times are outside of the allowed range, for example, if the "from" time was `9:30`, but the work pattern has been changed and the working day now starts at `10:00`.

## Before

1. Balance change is ignored for LiH
2. If set times are outside of the allowed range, the inputs are empty.

## After

1. Balance change is checked for LiH
2. If set times are outside of the allowed range, the inputs are set to either `min` or `max` allowed values for `from` and `to` times respectively.

## Technical Details

Balance change is now checked properly since it amends the first (and last) days depending on the work date. For example, the time that is currently set in the request is `10:00` - `15:00` and the deduction is set to `4`. The work pattern now changes to `12:00` - `14:00` this means that the deduction cannot be `4` anymore since the maximum allowed value is `2` now.

It also checks the `time_from` field, if it is empty, then it means that this day is a non-working day (as per API response).

```js
function recalculateBalanceChangeForHoursUnit (balanceChange) {
  var balanceObject = _.cloneDeep(balanceChange);
  var multipleDays = balanceObject.breakdown.length > 1;
  var firstDay = _.first(_.values(balanceObject.breakdown));
  var lastDay = _.last(_.values(balanceObject.breakdown));

  return $q.all([
    this.getWorkDayForDate(firstDay.date),
    multipleDays && this.getWorkDayForDate(lastDay.date)
  ])
    .then(function (workDays) {
      firstDay.amount = workDays[0].time_from
        ? Math.min(this['from_date_amount'], getMaxPossibleDeduction(workDays[0]))
        : 0;

      if (multipleDays) {
        lastDay.amount = workDays[1].time_from
          ? Math.min(this['to_date_amount'], getMaxPossibleDeduction(workDays[1]))
          : 0;
      }

      balanceObject.amount = _.reduce(balanceObject.breakdown,
        function (balanceObject, day) {
          return balanceObject - day.amount;
        }, 0);

      return balanceObject;
    }.bind(this));
}
```

If we refer to the example described above (`10:00` - `15:00` --> `12:00` - `14:00`), then times UI should be updated as well. If the minimum allowed time is `12:00` it cannot be `10:00` anymore. The function below sets the default one.
 
```js
function initTime (dateType) {
  var time = moment(vm.request[dateType + '_date']).format('HH:mm');
  var timeObject = vm.uiOptions.times[dateType];

  if (moment.duration(time).subtract(moment.duration(timeObject.min)).asHours() <= 0 ||
    moment.duration(time).subtract(moment.duration(timeObject.max)).asHours() >= 0) {
    return timeObject[dateType === 'from' ? 'min' : 'max'];
  }

  vm.uiOptions.times[dateType].time = time;
}
```

## Comments
https://github.com/civicrm/civihr/pull/2505/commits/df2d2bb86f6a47764dabc68cd405d866cd2febe7 - the issue happened because the time in UI was adjusted, but not in the leave request object.

https://github.com/civicrm/civihr/pull/2505/commits/e47ee7fb3c7b991b18aa62c076bf4b02db57cc6b - the issue happened because the adjustment ignored the situation when the min/max time is '00:00' (for non-working days the min/max time is returned as '' empty string, then converted to '00:00')

https://github.com/civicrm/civihr/pull/2505/commits/d6fe41f03b09d08641e618c9047fde47c038fc04 - the issue happened because getBalanceChangeBreakdown() API returns positive daily amount for days unit, but negative for hours unit.

----

✅Manual Tests - passed
✅Jasmine Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -487,9 +477,8 @@ define([
      * @return {String} amount of hours, eg. '7.5'
      */
     function getTimeDifferenceInHours (timeFrom, timeTo) {
-      return (getMomentDateWithGivenTime(timeTo)
-        .diff(getMomentDateWithGivenTime(timeFrom), 'minutes') / 60)
-        .toString();
+      return moment.duration(timeTo)
+        .subtract(moment.duration(timeFrom)).asHours().toString();
```
reneolivo: why do return a string? I think a number would be more convenient.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -574,6 +563,28 @@ define([
...
+     * then it sets the minimum allowed time for "from" time
+     * and the maximum allowed time for "to" time.
+     *
+     * @param {String} dateType from|to
+     */
+    function initTime (dateType) {
+      var time = moment(vm.request[dateType + '_date']).format('HH:mm');
+      var timeObject = vm.uiOptions.times[dateType];
+
+      if (moment.duration(time).subtract(moment.duration(timeObject.min)).asHours() <= 0 ||
```
reneolivo: can we name these conditions?

```js
var isBeforeWorkPatternRange = moment.duration(time).subtract(moment.duration(timeObject.min)).asHours() <= 0;
var isAfterWorkPatternRange = moment.duration(time).subtract(moment.duration(timeObject.max)).asHours() >= 0;

if (isBeforeWorkPatternRange || isAfterWorkPatternRange) {
  // ...
}
```

and, why not use `getTimeDifferenceInHours` ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/leave-request.instance.js
```diff
@@ -86,6 +87,18 @@ define([
...
 
+      /**
+       * Calculates time difference in hours
+       *
+       * @param  {String} timeFrom in HH:mm format
+       * @param  {String} timeTo in HH:mm format
+       * @return {String} amount of hours, eg. '7.5'
+       */
+      function getMaxPossibleDeduction (workDay) {
+        return moment.duration(workDay.time_to)
```
reneolivo: what happened to `getTimeDifferenceInHours` ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
reneolivo: Why isn't this condition like this instead?

```js
function initTime (dateType) {
  var isBeforeWorkPatternRange, isAfterWorkPatternRange;
  var time = moment(vm.request[dateType + '_date']).format('HH:mm');
  var timeObject = vm.uiOptions.times[dateType];
  var isBeforeWorkPatternRange = getTimeDifferenceInHours(timeObject.min) <= 0;
  var isAfterWorkPatternRange = getTimeDifferenceInHours(timeObject.min) >= 0;
  
  if (dateType === 'from' && isBeforeWorkPatternRange) {
    return timeObject.min;
  } else if (dateType === 'to' && isAfterWorkPatternRange) {
    return timeObject.max;
  }
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/leave-request.instance.js
```diff
@@ -102,22 +115,41 @@ define([
...
+
+            if (multipleDays) {
+              lastDay.amount = workDays[1].time_from
+                ? Math.min(this['to_date_amount'], getMaxPossibleDeduction(workDays[1]))
+                : 0;
+            }
+
+            balanceObject.amount = _.reduce(balanceObject.breakdown,
+              function (balanceObject, day) {
+                return balanceObject - day.amount;
```
reneolivo: is `balanceObject` really an object?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/leave-request.instance.js
```diff
@@ -102,22 +115,41 @@ define([
...
-
-        if (balanceChange.breakdown.length > 1) {
-          _.last(_.values(balanceChange.breakdown)).amount = this['to_date_amount'];
-        }
-
-        balanceChange.amount = _.reduce(balanceChange.breakdown,
-          function (updatedChange, day) {
-            return updatedChange - day.amount;
-          }, 0);
+      function recalculateBalanceChangeForHoursUnit (balanceChange) {
```
reneolivo: What happens to day units?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
reneolivo: Was this not going to be turned into a service?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/instances/leave-request.instance.spec.js
```diff
@@ -279,45 +279,101 @@ define([
...
+                .subtract(moment.duration(workDayTo.time_from)).asHours();
+            instance.from_date_amount = 2000;
+            instance.to_date_amount = 1500;
+            calculateBalanceChange.then(function (_balanceChange_) {
+              balanceChange = _balanceChange_;
+            });
+            $rootScope.$digest();
+          });
+
+          it('reduces "from" day amount to the allowed value', function () {
```
reneolivo: You tested what happens when both dates are outside of the new work pattern times, but where are the tests for when the from time is outside the range, but the to time is alright (and vice-versa)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
igorpavlov: I think we agreed that if this is used in 2 places, this not worth turning it into a service.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
```diff
@@ -951,10 +951,12 @@ define([
 
               it('sets from time to minimum allowed time', function () {
                 expect(controller.uiOptions.times.from.time).toBe(minTime);
+                expect(leaveRequest.from_date.slice(11, 16)).toBe(minTime);
```
reneolivo: I have no idea what's 11 and 16. Use this instead:

```js
expect(moment(leaveRequest.from_date).format('HH:mm')).toBe(minTime);
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
```diff
@@ -951,10 +951,12 @@ define([
...
               });
 
               it('sets to time to maximum allowed time', function () {
                 expect(controller.uiOptions.times.to.time).toBe(maxTime);
+                expect(leaveRequest.to_date.slice(11, 16)).toBe(maxTime);
```
reneolivo: same here

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -577,13 +577,12 @@ define([
...
+      var isOutsideWorkPatternRange =
+        getTimeDifferenceInHours(timeObject.min, time) <= 0 ||
+        getTimeDifferenceInHours(timeObject.max, time) >= 0;
 
-      if (dateType === 'from' && isBeforeWorkPatternRange) {
-        time = timeObject.min;
-      } else if (dateType === 'to' && isAfterWorkPatternRange) {
-        time = timeObject.max;
+      if (isOutsideWorkPatternRange) {
+        time = timeObject[dateType === 'from' ? 'min' : 'max'];
```
reneolivo: You keep using that pattern which is interesting, but it's harder to read (even if you save a few characters). This is easier to read:

```js
time = dateType === 'from' ? timeObject.min : timeObject.max
```

Because instead of having `object -> condition -> propertyA or propertyB`  you have ` condition -> object.propertyA or object.propertyB` which follows more naturally the relationship between object and properties.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
```diff
@@ -919,21 +919,54 @@ define([
             });
 
             describe('when set times are outside the allowed range', function () {
-              var minTime = '09:00';
-              var maxTime = '17:00';
+              initSetTimeAccordingToWorkPattern('02:00', '23:00', '09:00', '17:00');
```
reneolivo: We need to avoid functions with 4 parameters as much as possible because it's harder to know what is what. I'm guessing these are the values (before reading the function, correct me if I'm wrong):

- start time for the first day
- end time for the first day
- start time for the last day
- end time for the last day

I'd suggest either of these options:

```js
initSetTimeAccordingToWorkPattern({
  firstDayStartTime: '02:00',
  firstDayEndTime: '23:00',
  lastDayStartTime: '09:00',
  lastDayEndTime: '17:00'
});

// or

initSetTimeForFirstDay('02:00', '23:00');
initSetTimeForLastDay('09:00', '17:00');

```

After reading `initSetTimeAccordingToWorkPattern` I can't understand the name or what it does. This is what I can infer:
* it sets the start and end time for the leave request (nothing to do with Work Pattern then)
* the last two parameters are the response the work pattern will return for the `getWorkDayForDate` method (so not the last day of the work pattern as I suspected).

Why not refactor into two functions then?

```js
setRequestTimes('02:00', '23:00');
setupWorkPatternTimeResponse('09:00', '17:00'); // or a better name
```

`setRequestTimes` would just do that and `setupWorkPatternTimeResponse` does all the `beforeEach`/`afterEach` we need for setting up the tests.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
```diff
@@ -919,21 +919,54 @@ define([
...
+                expect(leaveRequest.to_date.slice(11, 16)).toBe('00:00');
+              });
+            });
+
+            /**
+             * Initialises set time for a given work pattern range
+             *
+             * @param {String} setFromTime HH:mm
+             * @param {String} setToTime HH:mm
+             * @param {String} wpFromTime HH:mm or empty string
```
reneolivo: Nobody will lose an extra second of their time if you use `workPatternFromTime` and `workPatternToTime` instead.

But, if you have a function that is called `setupWorkPatternTimeResponse` (or similar) you can forget about the `workPattern` prefix and just use `fromTime` and `toTime`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
igorpavlov: I don't really think it deserves two function, this is just to avoid massive code duplication. So I changed the params and name, but left it as one function.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
igorpavlov: Fixed as per previous comment

PR #2509, opened 2018-02-27
PCHR-3372: Take Contact Work Pattern into Consideration when validating Max consecutive leave days

tunbola:
## Overview
When validating the maximum consecutive days that can be taken when a leave request is to be created, the validation does not consider actual working days in the leave request date range based on the contact work pattern, hence leave requests that should have been allowed to be created are not.

## Before
- Take for example a contact using the default work pattern having weekends as non working days for the contact. If a contact requests leave from Monday this week to Monday next week, thats 6 days of leave the the contact is requesting.
Assuming the max consecutive leave days allowed for the absence type the contact is requesting leave for is 6 days, these request should have been allowed but its not allowed currently.
The validation assumes that the contact is requesting for 8 days(weekend inclusive) and does not allow the leave request to be created.

## After
- The `validateLeaveDaysAgainstAbsenceTypeMaxConsecutiveLeaveDay` now actually considers the number of working days in the leave request date range and compares that against the max consecutive leave days allowed for the absence type the contact is requesting leave for. So days like weekends, public holidays and non-working days will not be included in the calculation for number of leave days requested.
- The` LeaveRequest::calculateBalanceChange` method was used to get the number of working days for a leave request date range.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -691,7 +691,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   }
 
   /**
-   * This method checks that the number of days for the Leave Request
+   * This method checks that the number of working days for the Leave Request
    * is not greater than the max_consecutive_leave_days for the absence type
    *
    * @param array $params
@@ -701,14 +701,10 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
    */
   private static function validateLeaveDaysAgainstAbsenceTypeMaxConsecutiveLeaveDays($params, $absenceType) {
-    $fromDate = new DateTime($params['from_date']);
-    $toDate = new DateTime($params['to_date']);
-
-    $interval = $toDate->diff($fromDate);
-    $intervalInDays = $interval->format("%a");
+    $leaveDays = self::getNumberOfWorkingDaysForLeaveRequest($params);
     $maxConsecutiveLeaveDays = $absenceType->max_consecutive_leave_days;
 
-    if (!empty($maxConsecutiveLeaveDays) && $intervalInDays > $maxConsecutiveLeaveDays) {
+    if (!empty($maxConsecutiveLeaveDays) && $leaveDays > $maxConsecutiveLeaveDays) {
       throw new InvalidLeaveRequestException(
         'Only a maximum '. round($maxConsecutiveLeaveDays, 1) .' days leave can be taken in one request. Please modify days of this request',
         'leave_request_days_greater_than_max_consecutive_days',
@@ -1439,5 +1435,34 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $currentLeaveRequest = LeaveRequest::findById($leaveRequestID);
 
     return AbsenceType::findById($currentLeaveRequest->type_id);
+  }
+
+  /**
+   * Get the number of working days for a leave request, Days with
+   * balance change of zero or that evaluates to false are either non
+   * working day, weekend or public holidays.
+   * The days for which the balance amounts are not zero are equal
+   * to the number of working days.
+   *
+   * @param array $params
+   *
+   * @return int
+   */
+  private static function getNumberOfWorkingDaysForLeaveRequest($params) {
+    $fromDate = new DateTime($params['from_date']);
+    $toDate = new DateTime($params['to_date']);
+    $balanceDetails = self::calculateBalanceChange(
+      $params['contact_id'],
+      $fromDate,
+      $toDate,
+      $params['type_id'],
+      !empty($params['from_date_type']) ? $params['from_date_type'] : null,
+      !empty($params['to_date_type']) ? $params['to_date_type'] : null
+    );
+
+    $balanceAmounts = array_column($balanceDetails['breakdown'], 'amount');
+    $balanceAmounts = array_filter($balanceAmounts);
+
+    return count($balanceAmounts);
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -989,13 +989,11 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       }
 
       $dayType = $leaveDateDayTypes[$date->format('Y-m-d')];
-      $publicHolidayLeaveRequestExists = $dayType == $leaveRequestDayTypes['public_holiday'];
 
-      if ($publicHolidayLeaveRequestExists) {
+      if ($dayType == $leaveRequestDayTypes['public_holiday']) {
         $amount = 0.0;
       }
-
-      if(!$publicHolidayLeaveRequestExists){
+      else {
         $amount = LeaveBalanceChange::calculateAmountForDate(
           $leaveRequest,
           $date,
@@ -1048,8 +1046,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       if ($publicHolidayLeaveRequestExists) {
         $dayType = $leaveRequestDayTypes['public_holiday'];
       }
-
-      if(!$publicHolidayLeaveRequestExists) {
+      else {
         $type = ContactWorkPattern::getWorkDayType($contactID, $date);
         $dayType = self::getLeaveRequestDayTypeFromWorkDayType($type);
       }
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -875,6 +882,46 @@ public function testNumberOfDaysOfLeaveRequestShouldNotBeGreaterMaxConsecutiveLe
     ]);
   }
 
+  public function testLeaveRequestCanbeCreatedWhenNumberOfLeaveWorkingDaysNotGreaterThanMaxConsecutiveDays() {
```
davialexandre: s/Canbe/CanBe/

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1440,4 +1436,33 @@ private static function getCurrentAbsenceType($leaveRequestID) {
...
+   * to the number of working days.
+   *
+   * @param array $params
+   *
+   * @return int
+   */
+  private static function getNumberOfWorkingDaysForLeaveRequest($params) {
+    $fromDate = new DateTime($params['from_date']);
+    $toDate = new DateTime($params['to_date']);
+    $balanceDetails = self::calculateBalanceChange(
```
davialexandre: I wonder if we can do this without `calculateBalanceChange()`. This calculation is quite expensive and we are already calling it in the `validateEntitlementAndWorkingDayAndBalanceChange()`, this is just making the Leave Request validation slower than it already is. Besides, I'd say there are some readability and single responsibility issues here. The fact that we're using this `calculateBalanceChange()` method for two different types of validations, shows that it's doing different things. This is not good. In fact, we can see that when you do the whole balance calculation but in the end you only care about the number of balance changes where the amount is > 0. We need to improve this. Perhaps split that method into smaller ones

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -875,6 +882,46 @@ public function testNumberOfDaysOfLeaveRequestShouldNotBeGreaterMaxConsecutiveLe
...
+      'type_id' => $absenceType->id,
+      'contact_id' => $contactID,
+      'status_id' => 1,
+      'from_date' => CRM_Utils_Date::processDate('2017-01-05'),
+      'from_date_type' => 1,
+      'to_date' => CRM_Utils_Date::processDate('2017-01-10'),
+      'to_date_type' => 1,
+      'request_type' => LeaveRequest::REQUEST_TYPE_LEAVE
+    ]);
+  }
```
davialexandre: It would be good to have tests with different work patterns, where some of the normal weekdays are non-working days

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Good Idea, Let me see what I can come up with. I wasn't too comfortable with the expensive call too, Will try split the method.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: Sure.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: I tried splitting the method and merging two validation methods that were using the `calculateBalanceChange()` for their validations. It seems to work fine, Only issue is that the validation method i merged it to i.e validateEntitlementAndWorkingDayAndBalanceChange does not get run when leave requests are imported and which subsequently means that validations for max consecutive days will not be ran either for that case.
What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -98,12 +98,11 @@ public static function validateParams($params, $validationMode = self::VALIDATIO
     self::validateLeaveDatesDoesNotOverlapContractsWithLapses($params);
     self::validateAbsenceTypeIsActiveAndValid($params, $absenceType);
     self::validateTOILRequest($params, $absenceType, $absencePeriod);
-    self::validateLeaveDaysAgainstAbsenceTypeMaxConsecutiveLeaveDays($params, $absenceType);
     self::validateAbsenceTypeAllowRequestCancellationForLeaveRequestCancellation($params, $absenceType);
     self::validateAbsencePeriod($params, $absencePeriod);
 
     if($validationMode != self::IMPORT_VALIDATION) {
-      self::validateEntitlementAndWorkingDayAndBalanceChange($params, $absenceType, $absencePeriod);
+      self::validateMaxConsecutiveDaysAndEntitlementAndWorkingDayAndBalanceChange($params, $absenceType, $absencePeriod);
```
davialexandre: This doesn't look good. What if we need a new validation that also depends on balance changes or work days? Will we just keep increasing this method name? The main point of my other comment was: "I wonder if we can do this without calculateBalanceChange()", but we are still calling that method. The only difference is that now we call it only once.

The only validation that really needs the balance change is the entitlement one (where we check if the request balance is not greater than the remaining balance). Everything else has nothing to do with that. For the Working Days and the Consecutive Days validations, all we need is the information from the Work Pattern. So, when I was talking about splitting things, I was talking about splitting those two types of validations.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1440,4 +1439,47 @@ private static function getCurrentAbsenceType($leaveRequestID) {
...
+
+  /**
+   * Returns the balance change breakdown for a leave request
+   * based on the parameters passed to LeaveRequest.create.
+   *
+   * @param array $params
+   *
+   * @return array
+   */
+  private static function getLeaveBalanceBreakdown($params) {
```
davialexandre: This method ended up being confusing because the Breakdown is just part of what the `calculateBalanceChange()` method returns, but `getLeaveBalanceBreakdown()` returns exactly the same thing as `calculateBalanceChange()`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -1033,6 +1028,38 @@ public static function calculateBalanceChange(
...
+    $leaveRequestDayTypes = array_flip(self::buildOptions('from_date_type', 'validate'));
+    $leaveDateDayTypes = [];
+    foreach ($datePeriod as $date) {
+      $publicHolidayLeaveRequestExists = self::publicHolidayLeaveRequestExists($contactID, $date);
+
+      if ($publicHolidayLeaveRequestExists) {
+        $dayType = $leaveRequestDayTypes['public_holiday'];
+      }
+
+      if(!$publicHolidayLeaveRequestExists) {
```
davialexandre: Instead of the two ifs, I think that an if/else makes more sense here

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -984,23 +978,24 @@ public static function calculateBalanceChange(
...
     foreach ($datePeriod as $date) {
       if($excludeStartAndEndDates) {
         if(in_array($date->format('Y-m-d'), [$fromDate->format('Y-m-d'), $toDate->format('Y-m-d')])) {
           continue;
         }
       }
-      $publicHolidayLeaveRequestExists = self::publicHolidayLeaveRequestExists($contactId, $date);
+
+      $dayType = $leaveDateDayTypes[$date->format('Y-m-d')];
+      $publicHolidayLeaveRequestExists = $dayType == $leaveRequestDayTypes['public_holiday'];
```
davialexandre: The `$publicHolidayLeaveRequestExists` name doesn't make sense anymore. What about `$dayIsPublicHoliday`? Or you can simply use `$dayType == $leaveRequestDayTypes['public_holiday']` as the condition for the if

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -984,23 +978,24 @@ public static function calculateBalanceChange(
...
 
       if ($publicHolidayLeaveRequestExists) {
         $amount = 0.0;
-        $dayType = $leaveRequestDayTypes['public_holiday'];
       }
 
       if(!$publicHolidayLeaveRequestExists){
```
davialexandre: I think an if/else makes more sense here too

PR #2512, opened 2018-03-01
PCHR-3315: Improve UX for horizontally scrolled blocks

igorpavlov:
## Overview
The problem is that horizontally scrolled blocks do not provide enough visual information for users that they can actually be scrolled. This PR adds a directive that can be used for any content that has more than 100% width of its parent element. The directive dynamically adds shadows to the content's edges, visually representing that there is a hidden content that can be seen by scrolling horizontally. The shadows respond to scrolling, resizing and/or DOM population inside the content. If you scroll to one of the edges, the shadow on that edge disappears. If the content's is less or equal to 100% of its parent element, all shadows disappear.

**Directive usage**

```html
<div style='width: 100px'>
  <div style='width: 300px' horizontal-scroll-shadows>Wider content</div>
                            <!-- ^ -->
</div>
```

## Before
You have no idea if you can scroll the table horizontally.

## After

The shadow tells you there is something hidden and you should scroll.

### The directive

```js
/* eslint-env amd */
/* globals MutationObserver */

define([...], function (angular, directives) {
  directives.directive('horizontalScrollShadows', [function () {
    return {
      restrict: 'A', // use directive via attribute
      controller: horizontalScrollShadowsController
    };
  }]);

  horizontalScrollShadowsController.$inject = ['$element', '$scope', '$window'];

  function horizontalScrollShadowsController ($element, $scope, $window) {
    var content, wrapper;

    (function init () {
      wrapContents(); // wraps the content that you set the attribute to, to support the shadows
      initToggleShadowsHandlers();
      toggleShadows();
    }());

    /**
     * Makes shadow toggle on the wrapper element scrolling,
     * or DOM changes inside content element, or window resize
     */
    function initToggleShadowsHandlers () {
      wrapper.on('scroll', toggleShadows); // listens to scroll
      angular.element($window).on('resize', toggleShadows); // listens to window resize
      // that part is tricky. It seems there is no way to nicely track the width of an element
      // 1. Listening to `resize` on an element won't work, it only works for $window.
      // 2. Watching `elem.style.width` won't help, width always be empty unless changed via JS or CSS directly
      // 3. Watching `elem.style` is exhaustive and also throws an exception since style is not just a list of properties, it also contains circular connections that break watch of course.
      // 4. Watching `elem.clientWidth` works, but not totally. It fails to watch some essential changes, especially when the content becomes larger than wrapper.
      // So we use MutationObserver that checks if the content DOM was modified, for example, during load
      (new MutationObserver(toggleShadows)).observe(content[0], { subtree: true, childList: true });
      // and in some cases clientWidth is triggered
      $scope.$watch(function () {
        return wrapper[0].clientWidth;
      }, toggleShadows);
      $scope.$watch(function () {
        return content[0].clientWidth;
      }, toggleShadows);
    }

    /**
     * Toggles shadows depending on the scroll position
     */
    function toggleShadows () {
      var scroll = wrapper.scrollLeft();
      var contentWidth = content.width();
      var wrapperWidth = wrapper.width();

      if (contentWidth <= wrapperWidth) {
        wrapper.removeClass('insetShadowLeft insetShadowRight');

        return;
      }

      wrapper[scroll > 0 ? 'addClass' : 'removeClass']('insetShadowLeft');
      wrapper[scroll < contentWidth - wrapperWidth ? 'addClass' : 'removeClass']('insetShadowRight');
    }

    /**
     * Wraps original contents to make an environment for inset shadows
     */
    function wrapContents () {
      $element.wrap(angular.element(
        '<div class="horizontal-scroll-shadows-content"></div>'));
      content = $element.parent();
      content.wrap(angular.element(
        '<div class="horizontal-scroll-shadows-wrapper"></div>'));
      wrapper = content.parent();
      wrapper.wrap(angular.element(
        '<div class="horizontal-scroll-shadows-master"></div>'));
    }
  }
});
```

```css
.horizontal-scroll-shadows-master {
  overflow: hidden; // this ensures the shadows are hidden behind
  position: relative;
}

.horizontal-scroll-shadows-wrapper {
  overflow-x: auto;

  &::before,
  &::after {
    box-shadow: 0 0 10px 1px rgba(48, 40, 40, 0.5);
    content: '';
    display: block;
    height: calc(100% - 10px); // if height is 100% then the shadow will be cut off
    opacity: 0; // by default the shadows are invisible
    position: absolute;
    top: 5px;
    transition: opacity 0.2s ease-out;
    width: 1px; // in order to make the shadow appear
    z-index: 1; // in order to appear on top of the inset content
  }

  &::before {
    left: -1px; // This hides the actual object leaving the shadow still visible
  }

  &::after {
    right: -1px;
  }

  &.insetShadowLeft::before,
  &.insetShadowRight::after {
    opacity: 1;
  }

  .horizontal-scroll-shadows-content {
    display: inline-block; // to appear with :after and :before properly
    min-width: 100%; // otherwise it will shrink the content with less than 100% width
  }
}
```

```html
<div horizontal-scroll-shadows class="chr_leave-report__table-container">
  <section class="panel panel-default">
    ...
```

## Comments
Indentation was fixed in Karma configuration file.
The support table for mutation observer is here https://caniuse.com/#feat=mutationobserver and it matches CiviHR requirements.

----

✅ Manual Tests (desktop browser: Chrome, Safari, Firefox) - passed
✅ Manual Tests (mobile browser: iOS Safari) - passed
✅ Backstop tests - passed (had to manually change the viewport to mobile in SSP)
✅ CSS distributive files - included
⏹ PHP Unit Tests - not needed


org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
```diff
@@ -0,0 +1,40 @@
...
+.horizontal-scroll-shadows-wrapper {
+  overflow-x: auto;
+
+  &::before,
+  &::after {
+    box-shadow: 0 0 10px 1px rgba(48, 40, 40, 0.5);
+    content: '';
+    display: block;
+    height: calc(100% - 10px); // if height is 100% then the shadow will be cut off
+    opacity: 0; // by default the shadows are invisible
```
reneolivo: Wouldn't `opacity: 0` stop the user from clicking the content under the shadow even if it's not visible?

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
```diff
@@ -0,0 +1,40 @@
...
+    position: absolute;
+    top: 5px;
+    transition: opacity 0.2s ease-out;
+    width: 1px; // in order to make the shadow appear
+    z-index: 1; // in order to appear on top of the inset content
+  }
+
+  &::before {
+    left: -1px; // This hides the actual object leaving the shadow still visible
```
reneolivo: "actual object" ? you mean the content?

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
```diff
@@ -0,0 +1,76 @@
...
+     */
+    function initToggleShadowsHandlers () {
+      wrapper.on('scroll', toggleShadows);
+      angular.element($window).on('resize', toggleShadows);
+      (new MutationObserver(toggleShadows)).observe(content[0], { subtree: true, childList: true });
+      $scope.$watch(function () {
+        return wrapper[0].clientWidth;
+      }, toggleShadows);
+      $scope.$watch(function () {
+        return content[0].clientWidth;
```
reneolivo: why are you watching `clientWidth` twice? you meant `clientHeight` ?

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
```diff
@@ -0,0 +1,76 @@
...
+      var contentWidth = content.width();
+      var wrapperWidth = wrapper.width();
+
+      if (contentWidth <= wrapperWidth) {
+        wrapper.removeClass('insetShadowLeft insetShadowRight');
+
+        return;
+      }
+
+      wrapper[scroll > 0 ? 'addClass' : 'removeClass']('insetShadowLeft');
```
reneolivo: what do you like best:

```js
// A:
wrapper[scroll > 0 ? 'addClass' : 'removeClass']('insetShadowLeft');

// B:
(scroll > 0 ? wrapper.addClass : wrapper.removeClass)('insetShadowLeft');

// C:
var classAction = scroll > 0 ? 'addClass' : 'removeClass';

wrapper[classAction]('insetShadowLeft');

// D:
var classAction = scroll > 0 ? wrapper.addClass : wrapper.removeClass;

classAction('insetShadowLeft');
```

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: The actual element is not even in the viewport, it is hidden completely from user. The shadow does not overlay the content in terms of click propagation. I considered this and tested, it does allow to click through.

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: No, the shadow object, let me rephrase

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
```diff
@@ -0,0 +1,40 @@
+.horizontal-scroll-shadows-master {
```
AkA84: there shouldn't be a "directives/" folder, the style shouldn't really care whether the classes come from a directive, a jquery plugin, or whatever

AkA84: you're not using BEM

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
```diff
@@ -0,0 +1,40 @@
...
+  overflow: hidden; // this ensures the shadows are hidden behind
+  position: relative;
+}
+
+.horizontal-scroll-shadows-wrapper {
+  overflow-x: auto;
+
+  &::before,
+  &::after {
+    box-shadow: 0 0 10px 1px rgba(48, 40, 40, 0.5);
```
AkA84: we already have a default box-shadow value coming from shoreditch, we can't use that?

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
```diff
@@ -0,0 +1,40 @@
...
+
+  &::before {
+    left: -1px; // This hides the actual object leaving the shadow still visible
+  }
+
+  &::after {
+    right: -1px;
+  }
+
+  &.insetShadowLeft::before,
```
AkA84: what's going on with naming convention here? In addition of not following BEM, you're using both camelCase and kebab-case

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-env amd */
+/* globals MutationObserver */
+
+define([
+  'common/angular',
+  'common/modules/directives'
+], function (angular, directives) {
+  'use strict';
+
+  directives.directive('horizontalScrollShadows', [function () {
```
AkA84: i think `x-scroll-shadows` would be less of a mouthful

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
```diff
@@ -0,0 +1,76 @@
...
+      wrapper[scroll > 0 ? 'addClass' : 'removeClass']('insetShadowLeft');
+      wrapper[scroll < contentWidth - wrapperWidth ? 'addClass' : 'removeClass']('insetShadowRight');
+    }
+
+    /**
+     * Wraps original contents to make an environment for inset shadows
+     */
+    function wrapContents () {
+      $element.wrap(angular.element(
+        '<div class="horizontal-scroll-shadows-content"></div>'));
```
AkA84: does this work the same?
```js
$element.wrap('<div class="horizontal-scroll-shadows-content" />')
```

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
```diff
@@ -0,0 +1,76 @@
...
+      content = $element.parent();
```
AkA84:
```js
content = $element.wrap('<div class="horizontal-scroll-shadows-content" />').parent();

File: org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js (hunk 3 above)
Comment (AkA84) [reply org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js]: why do you use both the mutation observer and `$watch` ?

File: org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js [hunk 11]
```diff
@@ -0,0 +1,149 @@
...
+
+define([
+  'common/angular',
+  'common/lodash',
+  'common/angularMocks',
+  'common/directives/horizontal-scroll-shadows.directive'
+], function (angular, _) {
+  'use strict';
+
+  describe('horizontalScrollShadows directive', function () {
```
AkA84: you don't put the type of the thing you are testing, just the name

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
```diff
@@ -0,0 +1,149 @@
...
+
+    beforeEach(module('common.directives'));
+    beforeEach(inject(function (_$compile_, _$document_, _$rootScope_, _$window_) {
+      $compile = _$compile_;
+      $document = _$document_;
+      $rootScope = _$rootScope_;
+      $window = _$window_;
+    }));
+
+    describe('when "horizontal-scroll-shadows" attribute is specified', function () {
```
AkA84: there is really no point to this outer, all-wrapping section. If you are testing the directive, of course you're interested in when the directive is applied. Please remove it

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        $document.find('body').append(element);
+
+        content = $document.find('.horizontal-scroll-shadows-content');
+        wrapper = $document.find('.horizontal-scroll-shadows-wrapper');
+
+        // Overflow is needed to be set to "auto", otherwise scroll won't work
+        wrapper.css({ 'overflow-x': 'auto' });
+      });
+
+      describe('when not scrollable', function () {
```
AkA84: more descriptive please. When the content is not scrollable, otherwise it seems that you're talking about the directive

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
```diff
@@ -0,0 +1,149 @@
...
+          wrapper.width(100);
+          content.width(100);
+        });
+
+        it('does not add any shadows', function () {
+          expect(wrapper.hasClass('insetShadowLeft')).toBe(false);
+          expect(wrapper.hasClass('insetShadowRight')).toBe(false);
+        });
+
+        describe('but then window is resized so it becomes scrollable', function () {
```
AkA84: `when the window is resized and the content becomes scrollable`

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
igorpavlov: I think I attempted to describe that in the PR, but let me clarify a bit more.

`MutationObserver` tracks DOM manipulation, for example, when the content is being loaded. When DOM is populated, that will **not** change the `clientWidth`.

`ClientWidth`, on the other hand,  is needed when you dynamically change the style of the content **or** wrapper. DOM is not touched in this case, so `MutationObserver` will **not** be  triggered.

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
igorpavlov: A.

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: @AkA84 where would you like me to put this file then?

AkA84: write it as a BEM component, put it in components/

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: @AkA84 in Shoreditch we have `$crm-form-item-shadow: 0 3px 6px 0 rgba(48, 40, 40, 0.25);`

while I need `0 0 10px 1px rgba(48, 40, 40, 0.5);`

I don't think I can apply then variable and then just change position props.

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: I have to put it in components out of namespace

igorpavlov: Done.

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: Done.

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
igorpavlov: `x-scroll-shadows` won't work

image

igorpavlov: Now it is `h-scroll-shadows`

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
igorpavlov: Yes. thank you

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
igorpavlov: No. .wrap() returns the **wrapped** element, no the **wrapper**

image

igorpavlov: Done.

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
igorpavlov: Done.

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
igorpavlov: Done

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
igorpavlov: Done

org.civicrm.reqangular/test/directives/horizontal-scroll-shadows.directive.spec.js
igorpavlov: Done.

org.civicrm.bootstrapcivihr/scss/directives/horizontal-scroll-shadows.scss
igorpavlov: I had to put it in components/outside-namespace, otherwise, won't work for SSP

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
reneolivo: I'm fine with it, tbh but I prefer option B because option A reads like this to me:
`object -> condition -> methodA or methodB -> parameters`

And option B reads:

`condition -> object.methodA or object.methodB -> parameters`

Even if there's repetition the order of object -> method -> parameters is preserved better.

But not a big deal.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
...
+    // leaving the actual shadow still visible
+    left: -1px;
+  }
+
+  &::after {
+    right: -1px;
+  }
+
+  &.chr_h-scroll-shadows__wrapper--inset-shadow-left::before,
+  &.chr_h-scroll-shadows__wrapper--inset-shadow-right::after {
```
AkA84: you don't need to nest it and use `&`, put it outside in the root level
```scss
.chr_h-scroll-shadows__wrapper {
}

.chr_h-scroll-shadows__wrapper--inset-shadow-left,
.chr_h-scroll-shadows__wrapper--inset-shadow-right {
  &::before,
  &::after {

  }
}
```

org.civicrm.reqangular/src/common/directives/horizontal-scroll-shadows.directive.js
reneolivo: why not just `scroll-shadows` ?

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
AkA84: also, i can't see the reason for having these two distinct modifiers as they are all doing the same thing (setting the `opacity` to 1. I think you can have a single modifier then, something like
```scss
.chr_h-scroll-shadows__shadow--show
```

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
+.chr_h-scroll-shadows__master {
+  overflow: hidden; // this ensures the shadows are hidden behind
+  position: relative;
+}
+
+.chr_h-scroll-shadows__wrapper {
```
AkA84: i'm not sure these are the most descriptive names we can find, what about something like
```
.chr_h-scroll-shadows__wrapper
  .chr_h-scroll-shadows__shadows
    .chr_h-scroll-shadows__content 
```

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
...
+
+  &::before,
+  &::after {
+    box-shadow: 0 0 10px 1px rgba(48, 40, 40, 0.5);
+    content: '';
+    display: block;
+    height: calc(100% - 10px); // if height is 100% then the shadow will be cut off
+    opacity: 0; // by default the shadows are invisible
+    position: absolute;
+    top: 5px;
```
AkA84: why this `5px`?

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
...
+.chr_h-scroll-shadows__wrapper {
+  overflow-x: auto;
+
+  &::before,
+  &::after {
+    box-shadow: 0 0 10px 1px rgba(48, 40, 40, 0.5);
+    content: '';
+    display: block;
+    height: calc(100% - 10px); // if height is 100% then the shadow will be cut off
+    opacity: 0; // by default the shadows are invisible
```
AkA84: i think with better names for the element (see above), this comment is superfluous

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
...
+    right: -1px;
+  }
+
+  &.chr_h-scroll-shadows__wrapper--inset-shadow-left::before,
+  &.chr_h-scroll-shadows__wrapper--inset-shadow-right::after {
+    opacity: 1;
+  }
+
+  .chr_h-scroll-shadows__content {
+    display: inline-block; // to appear with :after and :before properly
```
AkA84: what does "to appear with :after and :before properly" mean? what happens if you don't set `inline-block`?

org.civicrm.reqangular/src/common/directives/h-scroll-shadows.directive.js
```diff
@@ -47,30 +48,28 @@ define([
     function toggleShadows () {
       var scroll = wrapper.scrollLeft();
       var contentWidth = content.width();
+      var wrapperClassPrefix = directiveClassPrefix + '__wrapper';
       var wrapperWidth = wrapper.width();
 
       if (contentWidth <= wrapperWidth) {
-        wrapper.removeClass('insetShadowLeft insetShadowRight');
+        wrapper.removeClass(wrapperClassPrefix + '--inset-shadow-left');
```
reneolivo:
```js
var leftShadowClassName = directiveClassPrefix + '__wrapper--inset-shadow-left';
var rightShadowClassName = directiveClassPrefix + '__wrapper--inset-shadow-right';

if (contentWidth <= wrapperWidth) {
  wrapper.removeClass(leftSadowClassName, rightShadowClassName);

  return;
}

wrapper[scroll > 0 ? 'addClass' : 'removeClass'](leftSadowClassName);
wrapper[scroll < contentWidth - wrapperWidth ? 'addClass' : 'removeClass'](rightShadowClassName);
// I still think `contidion -> object.methodA or object.methodB -> parameter` is better

File: org.civicrm.reqangular/src/common/directives/h-scroll-shadows.directive.js [hunk 21]
```diff
@@ -0,0 +1,75 @@
...
+
+    (function init () {
+      wrapContents();
+      initToggleShadowsHandlers();
+      toggleShadows();
+    }());
+
+    /**
+     * Makes shadow toggle on the wrapper element scrolling,
+     * or DOM changes inside content element, or window resize
```
AkA84: i think the whole difference between using the observer and using `$watch` is worth being documented

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
igorpavlov: I don't always put the whole code, but in this case every line is essential.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
igorpavlov: I was thinking about that.
.chr_h-scroll-shadows__shadows - is not "shadows", it is a wrapper that holds the __content inside so it makes it scrollable.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
igorpavlov: explained in code

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/_h-scroll-shadows.scss
igorpavlov: It will collapse the content.

org.civicrm.reqangular/src/common/directives/h-scroll-shadows.directive.js
igorpavlov: I split the func into 5 funcs. So now it should be descriptive.

org.civicrm.reqangular/src/common/directives/h-scroll-shadows.directive.js
igorpavlov: `(scroll > 0 ? wrapper.addClass : wrapper.removeClass)(leftShadowClassName);`

doesn't work.

org.civicrm.bootstrapcivihr/scss/components/outside-namespace/scroll-shadows.scss
```diff
@@ -0,0 +1,42 @@
+.chr_scroll-shadows__wrapper {
```
AkA84: wait, why is this outside the namespace?

PR #2513, opened 2018-03-01
PCHR-3392: Set some Installation Custom Field Group Reserved Yes

ajesamson:
## Overview
This PR ensure the following custom groups are reserved if they exist in any installations.

 * Identification
 * Job contract dates
 * Job contract summary
 * Job summary

## Before
customer_group_installation_before

## After
customer_group_installation_after

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1163,16 +1163,47 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
    * @return bool
    */
   public function upgrade_1034() {
-    $result = civicrm_api3('CustomGroup', 'get', array(
+    $result = civicrm_api3('CustomGroup', 'get', [
       'sequential' => 1,
       'return' => ['id'],
       'name' => 'Contact_Length_Of_Service',
-    ));
-  
-    civicrm_api3('CustomGroup', 'create', array(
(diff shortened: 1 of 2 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1163,16 +1163,47 @@ public function upgrade_1033() {
...
+   */
+  public function upgrade_1035() {
+    $customGroups = [
+      'HRJobContract_Dates',
+      'HRJob_Summary',
+      'HRJobContract_Summary'
+    ];
+    
+    $result = civicrm_api3('CustomGroup', 'get', [
+      'sequential' => 1,
```
tunbola: No need for sequential here also.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1163,16 +1163,47 @@ public function upgrade_1033() {
...
+      'return' => ['id'],
+      'name' => ['IN' => $customGroups],
+    ]);
+
+    if (count($result['values'])) {
```
tunbola: You can use `$result['count'] > 0` rather than counting how many elements are in `$result['values']`

tunbola:
@ajesamson , Also please add a screenshot that depicts the work done in PR, The description says that these Custom groups were set to reserved but Gif looks like they were deleted. Or don't reserved custom groups show up?

PR #2516, opened 2018-03-05
PCHR-3243: Fix backstopjs task's return code

AkA84:
The `backstopjs` gulp tasks always returned a code `0` even if BackstopJS failed (either because of some errors or because some diffs were found). This made the tasks not suitable for automated process. This PR fixes the issue, so that the task returns a code `1` if any error occured

Additionally, the "contact access rights" scenarios were updated as the "manage access rights" buttons was moved inside the contact actions panel. This was done as part of the "contact actions" epic but unfortunately I forgot to update the scenarios as well. 

button

Also please note that the contact used in the scenario had been changed from "staff" to "admin", as now the button appears only for users that have the "access CiviHR" permission

Finally, this PR also updates the BackstopJS version to the latest currently available

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/contact-summary.js
+++ b/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/contact-summary.js
@@ -14,8 +14,13 @@ module.exports = (function () {
 
       return new Promise(function (resolve) {
         casper.then(function () {
-          casper.click('#manage-roles-and-teams');
+          this.showActions();
+        }.bind(this));
+
+        casper.then(function () {
+          casper.click('[data-contact-access-rights]');
           casper.waitWhileVisible('.spinner');
+
           resolve(this.waitForModal('contact-access-rights'));
         }.bind(this));
       }.bind(this));
--- a/uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
+++ b/uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
@@ -131,30 +131,33 @@ function runBackstopJS (command) {
 
   return getRolesAndIDs()
     .then(function (contactIdsByRoles) {
-      return new Promise(function (resolve) {
+      return new Promise(function (resolve, reject) {
         var isBackstopJSSuccessful;
 
         gulp.src(BACKSTOP_DIR_PATH + FILES.tpl)
           .pipe(file(destFile, tempFileContent(contactIdsByRoles)))
           .pipe(gulp.dest(BACKSTOP_DIR_PATH))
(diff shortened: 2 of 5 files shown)


uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/contact-summary.js
```diff
@@ -14,8 +14,13 @@ module.exports = (function () {
...
       return new Promise(function (resolve) {
         casper.then(function () {
-          casper.click('#manage-roles-and-teams');
+          this.showActions();
+        }.bind(this));
+
+        casper.then(function () {
+          casper.click('[data-contact-access-rights]');
           casper.waitWhileVisible('.spinner');
+
```
igorpavlov: Not sure why an empty line here

AkA84: same reason why we leave an empty line before the `return`, makes the code more readable

AkA84:
apologies @igorpavlov , i forgot to push the json file

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -131,30 +131,33 @@ function runBackstopJS (command) {
...
-                  sound: 'Beep'
-                }));
-            });
-
-            resolve(promise);
+            })
+              .then(function () {
+                isBackstopJSSuccessful = true;
+              })
+              .catch(_.noop).then(function () { // equivalent to .finally()
```
igorpavlov: I think `catch` block should be "hidden". I pay too much attention to it in the current code design.

```js
.then(function () {
}).catch(0)
.then(function () { // < I want to see the then block, not the catch block first

})
```

AkA84: That would be a very unusual syntax that we don't have anywhere else, so I don't think it would be a good idea.

The only two choices that i see are what is there now or 
```js
.then(function () {
})
.catch(_.noop)
.then(function () {
});
```
but given that the `catch` + `then` series is meant to be a substitute of just one command, it made sense to me to put them in one line "as one" rather than one after the other

igorpavlov: Please use the second one and not the `.catch(_.noop).then(function ()` since it is difficult to read.

uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/contact-summary.js
igorpavlov: OK.

PR #2518, opened 2018-03-05
PCHR-3423: Do not allow deleted Public holiday Leave Request for Contact to be re-created

tunbola:
## Overview
Currently when a public holiday leave request is deleted by the Admin for a contact, If a condition that triggers the creation of public holiday leave requests occurs e.g change in Work Pattern, the public holiday leave request is created for the contact again.

## Before
When a public holiday leave request is deleted by the Admin for a contact, it gets created again when a condition that triggers public holiday creation happens.

## After
When a public holiday leave request is deleted by the Admin for a contact, it does not get created again when a condition that triggers public holiday creation happens.

- The Admin deletes public holiday leave requests via the `LeaveRequestService.delete` function, rather than hard delete the requests, the function was modified to now soft delete public holiday leave requests.
```php
if ($leaveRequest->request_type == LeaveRequest::REQUEST_TYPE_PUBLIC_HOLIDAY) {
       $publicHoliday = new PublicHoliday();
       $publicHoliday->date = $leaveRequest->from_date;
       $this->publicHolidayLeaveRequestDeletionService->deleteForContact($leaveRequest->contact_id, $publicHoliday, TRUE);
}
```
Note the TRUE passed as the last parameter to `publicHolidayLeaveRequestDeletionService.deleteForContact` method. This method was modified to accept an optional parameter to specify whether to soft delete or hard delete a public holiday leave request.
It is still necessary to hard delete some public holiday leave requests for example when a Work pattern changes, all future public holiday leave requests are hard deleted and then created again with the new respective balance changes.

- The logic for creating public holiday leave requests was modified to check not only for the existence of a public holiday request before attempting to create a public holiday request. The logic now checks for existence of both soft deleted and non deleted public holiday leave requests, if such public holiday requests exists for the given date for the contact, the request does not get created.

```php
  public function createForContact($contactID, PublicHoliday $publicHoliday) {
    if (!$this->absenceType) {
      return;
    }

    $existingLeaveRequest = LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday, FALSE);
    if($existingLeaveRequest) {
      return;
    }

    $leaveRequest = $this->createLeaveRequest($contactID, $this->absenceType, $publicHoliday);
    $this->createLeaveBalanceChangeRecord($leaveRequest);
    $this->recalculateExpiredBalanceChange($leaveRequest);

    return $leaveRequest;
  }
```
Note the last parameter passed to the `LeaveRequest::findPublicHolidayLeaveRequest` function. The method was modified to either find all  public holiday requests or only non deleted public holiday requests via the optional parameter `nonDeletedOnly`.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -803,25 +803,43 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    *
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
-   * @param boolean $nonDeletedOnly
-   *   When true will only return result if the given public
-   *   holiday request is not yet deleted.
    *
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest|null
    */
-  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday, $nonDeletedOnly = TRUE) {
+  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday) {
     $leaveRequest = new self();
     $leaveRequest->contact_id = (int)$contactID;
     $leaveRequest->from_date = date('Ymd', strtotime($publicHoliday->date));
     $leaveRequest->request_type = self::REQUEST_TYPE_PUBLIC_HOLIDAY;
-    if ($nonDeletedOnly) {
-      $leaveRequest->is_deleted = 0;
-    }
+    $leaveRequest->is_deleted = 0;
 
     $leaveRequest->find(true);
     if($leaveRequest->id) {
-      $leaveRequest->fetch();
+      return $leaveRequest;
+    }
 
+    return null;
+  }
+
+  /**
+   * Returns a LeaveRequest instance representing the Public Holiday Leave Request
+   * for the given $publicHoliday and assigned to the Contact with the given
+   * $contactID, the leave request is returned as long as it exists in the database
+   * either soft deleted or not.
+   *
+   * @param int $contactID
+   * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   *
+   * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest|null
+   */
+  public static function findPublicHolidayLeaveRequestEvenIfDeleted($contactID, PublicHoliday $publicHoliday) {
+    $leaveRequest = new self();
+    $leaveRequest->contact_id = (int)$contactID;
+    $leaveRequest->from_date = date('Ymd', strtotime($publicHoliday->date));
+    $leaveRequest->request_type = self::REQUEST_TYPE_PUBLIC_HOLIDAY;
+
+    $leaveRequest->find(true);
+    if($leaveRequest->id) {
       return $leaveRequest;
     }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -803,15 +803,20 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    *
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   * @param boolean $nonDeletedOnly
+   *   When true will only return result if the given public
+   *   holiday request is not yet deleted.
    *
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest|null
    */
-  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday) {
+  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday, $nonDeletedOnly = TRUE) {
     $leaveRequest = new self();
     $leaveRequest->contact_id = (int)$contactID;
     $leaveRequest->from_date = date('Ymd', strtotime($publicHoliday->date));
     $leaveRequest->request_type = self::REQUEST_TYPE_PUBLIC_HOLIDAY;
-    $leaveRequest->is_deleted = 0;
+    if ($nonDeletedOnly) {
+      $leaveRequest->is_deleted = 0;
+    }
 
     $leaveRequest->find(true);
     if($leaveRequest->id) {
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
@@ -73,8 +73,11 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion {
     if(!$leaveRequest) {
       return;
     }
-    
+
     LeaveRequest::softDelete($leaveRequest->id);
+    foreach($leaveRequest->getDates() as $date) {
+      $this->recalculateDeductionForOverlappingLeaveRequestDate($leaveRequest, new DateTime($date->date));
+    }
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
@@ -42,33 +42,57 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion {
    * Deletes the Public Holiday Leave Request for the contact and Public Holiday.
    *
    * If there are LeaveRequestDates overlapping the public holiday, their
-   * balance change amount will be updated to no be 0 anymore.
+   * balance change amount will be updated to not be 0 anymore.
    *
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
-   * @param boolean $softDelete
-   *   If true, the public holiday request will be soft deleted
    */
-  public function deleteForContact($contactID, PublicHoliday $publicHoliday, $softDelete = FALSE) {
+  public function deleteForContact($contactID, PublicHoliday $publicHoliday) {
     $leaveRequest = LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday);
 
     if(!$leaveRequest) {
       return;
     }
 
+    $this->deleteDatesWithBalanceChanges($leaveRequest);
+    $leaveRequest->delete();
+
+  }
+
+  /**
+   * Soft Deletes the Public Holiday Leave Request for the contact and Public Holiday.
+   *
+   * If there are LeaveRequestDates overlapping the public holiday, their
+   * balance change amount will be updated to not be 0 anymore.
+   *
+   * @param int $contactID
+   * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   */
+  public function softDeleteForContact($contactID, PublicHoliday $publicHoliday) {
+    $leaveRequest = LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday);
+
+    if(!$leaveRequest) {
+      return;
+    }
+
+    $this->deleteDatesWithBalanceChanges($leaveRequest);
+    LeaveRequest::softDelete($leaveRequest->id);
+  }
+
+  /**
+   * Deletes the public holiday leave dates and balance changes.
+   * It also updates the balance change of any leave request overlapping
+   * the public holiday date to not be zero again but the amount calculated
+   * from the contact's work pattern for the date.
+   *
+   * @param \CRM_HRLeaveAndAbsences_BAO_LeaveRequest $leaveRequest
+   */
+  private function deleteDatesWithBalanceChanges(LeaveRequest $leaveRequest) {
     foreach($leaveRequest->getDates() as $date) {
       LeaveBalanceChange::deleteForLeaveRequestDate($date);
       $this->recalculateDeductionForOverlappingLeaveRequestDate($leaveRequest, new DateTime($date->date));
       $date->delete();
     }
-
-    if ($softDelete) {
-      LeaveRequest::softDelete($leaveRequest->id);
-    }
-    else {
-      $leaveRequest->delete();
-    }
-
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
@@ -56,7 +56,6 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion {
 
     $this->deleteDatesWithBalanceChanges($leaveRequest);
     $leaveRequest->delete();
-
   }
 
   /**
@@ -74,8 +73,7 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion {
     if(!$leaveRequest) {
       return;
     }
-
-    $this->deleteDatesWithBalanceChanges($leaveRequest);
+    
     LeaveRequest::softDelete($leaveRequest->id);
   }
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
@@ -46,8 +46,10 @@ class CRM_HRLeaveAndAbsences_Service_PublicHolidayLeaveRequestDeletion {
    *
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   * @param boolean $softDelete
+   *   If true, the public holiday request will be soft deleted
    */
-  public function deleteForContact($contactID, PublicHoliday $publicHoliday) {
+  public function deleteForContact($contactID, PublicHoliday $publicHoliday, $softDelete = FALSE) {
     $leaveRequest = LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday);
 
     if(!$leaveRequest) {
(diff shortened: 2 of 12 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -803,15 +803,20 @@ private static function validateAbsenceTypeIsActiveAndValid($params, $absenceTyp
...
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   * @param boolean $nonDeletedOnly
+   *   When true will only return result if the given public
+   *   holiday request is not yet deleted.
    *
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest|null
    */
-  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday) {
+  public static function findPublicHolidayLeaveRequest($contactID, PublicHoliday $publicHoliday, $nonDeletedOnly = TRUE) {
```
davialexandre: We should avoid adding new params like this as much as possible. The reason is that now the method can do two different things. 
I know that, sometimes, adding a new param is a way to avoid having to duplicate a huge query, but here the method is so simple that it's ok to have a separate method to get only non-deleted public holiday leave requests

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletion.php
```diff
@@ -46,8 +46,10 @@ public function deleteForAllContacts(PublicHoliday $publicHoliday) {
    *
    * @param int $contactID
    * @param \CRM_HRLeaveAndAbsences_BAO_PublicHoliday $publicHoliday
+   * @param boolean $softDelete
+   *   If true, the public holiday request will be soft deleted
    */
-  public function deleteForContact($contactID, PublicHoliday $publicHoliday) {
+  public function deleteForContact($contactID, PublicHoliday $publicHoliday, $softDelete = FALSE) {
     $leaveRequest = LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday);
```
davialexandre: The comment I left about the `findPublicHolidayLeaveRequest` method also applies here.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,56 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
     $this->assertNull(LeaveRequest::findPublicHolidayLeaveRequest(3, $publicHoliday));
   }
 
+  public function testCanFindAPublicHolidayLeaveRequestReturnsResultWhenNonDeletedIsFalseAndPublicHolidayRequestIsDeleted() {
+    AbsencePeriodFabricator::fabricate([
```
davialexandre: You're mixing two name styles here and now, if you try to read the method name, it doesn't make sense. Separating the methods will probably help with the test names

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,56 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
+
+    $publicHolidayRequest = PublicHolidayLeaveRequestFabricator::fabricate($contactID, $publicHoliday);
+    LeaveRequest::softDelete($publicHolidayRequest->id);
+
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday, FALSE));
```
davialexandre: Wouldn't it be better to check for the leave request id here? Just checking that it's not null is a bit too vague

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,56 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+    ]);
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
+
+    PublicHolidayLeaveRequestFabricator::fabricate($contactID, $publicHoliday);
+
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday, FALSE));
```
davialexandre: Maybe check the id here too

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,56 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
+
+    $publicHolidayRequest = PublicHolidayLeaveRequestFabricator::fabricate($contactID, $publicHoliday);
+    LeaveRequest::softDelete($publicHolidayRequest->id);
+
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contactID, $publicHoliday, FALSE));
+  }
+
+  public function testCanFindAPublicHolidayLeaveRequestReturnsResultWhenNonDeletedIsFalseAndPublicHolidayRequestIsNotDeleted() {
```
davialexandre: Same problem with the name here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -235,6 +235,24 @@ public function testDeleteSoftDeletesTheLeaveRequest() {
...
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = CRM_Utils_Date::processDate('2017-10-10');
+
+    $publicHolidayLeaveRequest = PublicHolidayLeaveRequestFabricator::fabricate($this->leaveContact, $publicHoliday);
+    $this->getLeaveRequestServiceWhenCurrentUserIsAdmin()->delete($publicHolidayLeaveRequest->id);
+
+    $publicHolidayLeaveRequestRecord = new LeaveRequest();
+    $publicHolidayLeaveRequestRecord->id = $publicHolidayLeaveRequest->id;
+    $publicHolidayLeaveRequestRecord->find(true);
```
davialexandre: These three lines of code can be replaced with `LeaveRequest::findById($publicHolidayLeaveRequest->id);`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
```diff
@@ -71,6 +71,43 @@ public function testCanDeleteAPublicHolidayLeaveRequestForASingleContact() {
     $this->assertEquals(0, LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement));
   }
 
+  public function testDeleteForContactWithSoftDeleteAsTrue() {
```
davialexandre: Please describe what should happen

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
```diff
@@ -71,6 +71,43 @@ public function testCanDeleteAPublicHolidayLeaveRequestForASingleContact() {
...
+    );
+    $periodEntitlement->contact_id = 2;
+    $periodEntitlement->type_id = $this->absenceType->id;
+
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $periodEntitlement->contact_id],
+      ['period_start_date' => $absencePeriod->start_date]
+    );
+
+    $publicHoliday = $this->instantiatePublicHoliday('2017-01-01');
```
davialexandre: In a few other tests you did the instantiation manually. How about extracting this `instantiatePublicHoliday` method to a trait so that it can be easily reused in these places where you're doing things manually?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
```diff
@@ -71,6 +71,43 @@ public function testCanDeleteAPublicHolidayLeaveRequestForASingleContact() {
...
+
+    $deletionLogic = new PublicHolidayLeaveRequestDeletion(new JobContractService());
+    $deletionLogic->deleteForContact($periodEntitlement->contact_id, $publicHoliday, TRUE);
+
+    $this->assertEquals(0, LeaveBalanceChange::getLeaveRequestBalanceForEntitlement($periodEntitlement));
+    //Check that the public holiday leave request is soft deleted.
+    $publicHolidayLeaveRequestRecord = new LeaveRequest();
+    $publicHolidayLeaveRequestRecord->id = $publicHolidayRequest->id;
+    $publicHolidayLeaveRequestRecord->is_deleted = 1;
+    $publicHolidayLeaveRequestRecord->find(TRUE);
```
davialexandre: Another way of checking this would be to use `LeaveRequest::findById()` and then asserting that is_deleted is true

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: That means we would have two identical methods with only a line being the difference between both methods? The parameter seems to me as a filter that filters off some records and we have many of such methods in L&A.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
tunbola: `LeaveRequest::findById()` is a modified version that only returns results for leave request with is_deleted = 0.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
davialexandre: Yeah, that's exactly what I meant. We do have methods like that in L&A and they're bad as well. Unfortunately, sometimes these methods are so complex that duplicating them would be risky, so adding a new param sounds better. However, in this case here, the code is so simple that the duplication is fine compared with the readability improvement we'll have. 

Just for reference, here are some places where we duplicated code to improved readability:
The [AbsencePeriod::getCurrentPeriod()](https://github.com/civicrm/civihr/blob/99ed22a5709a4d11ed6c8b7bb2739c19b4d08890/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php#L300-L310) method is exactly the same as the [AbsencePeriod::getPeriodOverlappingDate()](https://github.com/civicrm/civihr/blob/99ed22a5709a4d11ed6c8b7bb2739c19b4d08890/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsencePeriod.php#L320-L332)

The methods [AbsenceType::getEnableAbsenceTypes()](https://github.com/civicrm/civihr/blob/99ed22a5709a4d11ed6c8b7bb2739c19b4d08890/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L608-L620) and [AbsenceType::getEnabledSicknessAbsenceTypes()](https://github.com/civicrm/civihr/blob/99ed22a5709a4d11ed6c8b7bb2739c19b4d08890/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L627-L634) are another interesting example. Technically, we don't really need the `getEnabledSickenessAbsenceTypes()` methods. We could have just added a new param to `getEnabledAbsenceTypes()` and that would work the same way, but the readbility is much better with the second method.

I'm sure that if you look at the code there are some more examples of this

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
davialexandre: Oh, I forgot that

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
tunbola: LeaveRequest::findById() is a modified version that only returns results for leave request with is_deleted = 0.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,57 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
+
+    $publicHolidayRequest = PublicHolidayLeaveRequestFabricator::fabricate($contactID, $publicHoliday);
+    LeaveRequest::softDelete($publicHolidayRequest->id);
+
+    $publicHolidayRequest = LeaveRequest::findPublicHolidayLeaveRequestEvenIfDeleted($contactID, $publicHoliday);
+    $this->assertNotNull($publicHolidayRequest->id);
```
davialexandre: Why not assert the the ID is equal to the id of the leave request you deleted?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,57 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
+
+    PublicHolidayLeaveRequestFabricator::fabricate($contactID, $publicHoliday);
+
+    $publicHolidayRequest = LeaveRequest::findPublicHolidayLeaveRequestEvenIfDeleted($contactID, $publicHoliday);
+    $this->assertNotNull($publicHolidayRequest->id);
```
davialexandre: Maybe it's better to check for the id here too

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -91,6 +91,41 @@ public function testItDoesNotCreateALeaveRequestIfTheresIsAlreadyALeaveRequestFo
...
+      'end_date' => CRM_Utils_Date::processDate('last day of this year')
+    ]);
+
+    $periodEntitlement = $this->createLeavePeriodEntitlementMockForBalanceTests();
+    $periodEntitlement->contact_id = 2;
+    $periodEntitlement->type_id = $this->absenceType->id;
+
+    $date = new DateTime('first monday of this year');
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = $date->format('YmdHis');
```
davialexandre: Please use the new trait here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,57 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+  public function testFindPublicHolidayLeaveRequestEvenIfDeletedReturnsNullForANonExistentPublicHolidayRequest() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2017-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2017-12-01')
+    ]);
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
```
davialexandre: Please use the new trait here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,57 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+  public function testFindPublicHolidayLeaveRequestEvenIfDeletedReturnsResultForANonDeletedPublicHolidayRequest() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2017-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2017-12-01')
+    ]);
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
```
davialexandre: Please use the new trait here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -216,6 +216,57 @@ public function testShouldReturnNullIfItCantFindAPublicHolidayLeaveRequestForACo
...
+  public function testFindPublicHolidayLeaveRequestEvenIfDeletedReturnsResultForADeletedPublicHolidayRequest() {
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2017-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2017-12-01')
+    ]);
+
+    $contactID = 2;
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = '2017-01-01';
```
davialexandre: Please use the new trait here

PR #2523, opened 2018-03-07
PCHR-3375: Add Configuration Menu Items

mickadoo:
## Overview

This PR adds some new menu items to the "Administer" submenu.

## Before

The "Administer" submenu did not contain the new items.

image

## After

The following items were added to the "Administer" submenu. Indentation indicates nesting. All items except for "Leave" were added.

Permissions for all new items is `Access CiviCRM` except for the submenu items of "Leave" which have the permission `administer leave and absences`.

```
Localise CiviHR	
Job Contract	
	Contract Types
	Normal Place of Work
	Contract End Reasons
	Standard Full Time Hours
	Pay Scales
	Benefits
	Deductions
	Insurance Plan Types
Job Roles	
	Locations
	Regions
	Departments
	Levels
	Cost Centres
Other Staff Details	
	Prefixes
	Emergency Contact Relationships
	Manager Types
	Career History
	Disability Types
	Qualifications – Skill Categories
	Qualifications – Skill Levels
Leave (exists)
	---------------------------
	Sickness Reasons
	TOIL to be Accrued
	Work Pattern Change Reasons
	Work Pattern Day Equivalents
	---------------------------
	Import Leave/Absence Requests
```

image

## Comments

I refer to "Administer" as the parent menu, but in CiviHR it has been relabeled to "Configure"


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/NavigationMenuHelper.php
```diff
@@ -25,24 +25,30 @@ public static function findMenuItemByPath($menu, $path) {
...
+   *   Attributes of item to insert, e.g. ['label' => 'New Item']
    */
-  public static function insertChild(&$menu, $parentPath, $itemAttributes) {
-    $parent = &self::findMenuItemReferenceByPath($menu, $parentPath);
-
-    if (!$parent) {
-      $err = sprintf('Cannot find parent item "%s"', $parentPath);
-      throw new \Exception($err);
-    }
+  public static function insert(&$menu, $path, $attributes = []) {
```
mickadoo: These changes were to allow insertion of items into the top level (root) menu

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/NavigationMenu/ConfigurationMenuModifier.php
```diff
@@ -0,0 +1,186 @@
...
+
+    foreach ($optionGroupLinks as $itemName => $optionGroup) {
+      $link = 'civicrm/admin/options/' . $optionGroup . '?reset=1';
+      $path = $parent . '/' . $itemName;
+      $this->insert($menu, $path, $permission, $link);
+    }
+
+    // insert dividers
+    $numberOfDividers = 2;
+    $dividerText = '---------------------';
```
davialexandre: Wow, is this really how we add a divider to a CiviCRM menu? Wouldn't it be better to add some sort of `insertDivider()` method to the NavigationMenuHelper?

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -280,16 +281,22 @@ function hrcore_civicrm_pageRun($page) {
...
-  _hrcore_createSelfServicePortalMenu($params);
+function hrcore_civicrm_navigationMenu(&$menu) {
+  _hrcore_renameMenuLabel($menu, 'Contacts', 'Staff');
+  _hrcore_renameMenuLabel($menu, 'Administer', 'Configure');
+
+  _hrcore_coreMenuChanges($menu);
+  _hrcore_createHelpMenu($menu);
+  _hrcore_createDeveloperMenu($menu);
+  _hrcore_setDynamicMenuIcons($menu);
+  _hrcore_createSelfServicePortalMenu($menu);
```
davialexandre: Any plans to also create modifiers for the existing modifications? I really like your solution, but if we don't use it everywhere, people won't know what is the "right" way to do it and there's a chance they will keep using the helper functions

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/NavigationMenu/ConfigurationMenuModifier.php
```diff
@@ -0,0 +1,186 @@
...
+  private function addLeaveMenuItems(&$menu) {
+    $parent = 'Administer/leave_and_absences';
+    $permission = 'administer leave and absences';
+
+    $optionGroupLinks = [
+      'Sickness Reasons' => 'hrleaveandabsences_sickness_reason',
+      'TOIL to be Accrued' => 'hrleaveandabsences_toil_amounts',
+      'Work Pattern Change Reasons' => 'hrleaveandabsences_work_pattern_change_reason',
+      'Work Pattern Day Equivalents' => 'hrleaveandabsences_leave_days_amounts'
+    ];
```
davialexandre: It's not clear to me why are these items being added dynamically. Is this a requirement from the specification? If it is not, then why don't we add them the same way we add the other items of the leave and absences menu?

Besides, I think that it's a bit confusing that some items will be added by hrcore and others by L&A.

Note that I'm talking about L&A here, but this actually applies to the other modifications done by this class

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/NavigationMenuHelper.php
mickadoo: The description is what's wrong there. In the previous version the parent path was provided, like

`NavigationMenuHelper::insert($menu, 'Events', ['name' => 'Dashboard']);`

But now the name is inferred from the item you want to insert.

`NavigationMenuHelper::insert($menu, 'Events/Dashboard');`

Both of the above examples would insert an item called "Dashboard" into the "Events" submenu.

You can still override it using the optional attributes. The reason I changed it is that I found it easier to use this because I can focus only on the item I want to insert, and not the parent. If the parent doesn't exist it'll still throw an exception.

I can update the method description to match what it does now.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/NavigationMenuHelper.php
```diff
@@ -260,6 +263,21 @@ private static function &getSiblingsReference(&$menu, $path) {
     return $siblings;
   }
 
+  /**
+   * Gets the parent path from a given path, e.g. "Dashboard" from
+   * "Events/Dashboard"
+   *
```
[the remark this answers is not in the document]
mickadoo: Again, it was a mistake in the doc-block. I've updated it to show that it would return "Events"

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/NavigationMenu/ConfigurationMenuModifier.php
mickadoo: Well I don't think we use dividers anywhere else. It's a good idea to add a helper method. I think I'll maybe suggest to Mike / Jamie that we split this into a separate ticket. It's not a very elegant solution, it would probably be better to add some CSS class to nav items and apply a border to the bottom of them to create a divider. I'll remove this from the PR.

uk.co.compucorp.civicrm.hrcore/hrcore.php
mickadoo: Absolutely. I would much prefer to move this stuff into a modifier, but was focused on getting the ticket working first. I'd prefer to do this in the final ticket for this sub-task, after everything else is working.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/NavigationMenu/ConfigurationMenuModifier.php
mickadoo: When I started on this ticket I asked in the backend room about the choice between using dynamic inserts and database inserts. 

image

I got the impression that we were favoring dynamic inserts, although now I'm not so sure. Since these are all new submenus, and not alterations of core items, I think we can probably add them to the database, as long as you agree it makes sense too.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1012.php
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Gets the link to edit an option group
+   *
+   * @param string $groupName
+   *
+   * @return string
+   */
+  private function up1020_getOptionGroupLink($groupName) {
```
davialexandre: s/1020/1012/

uk.co.compucorp.civicrm.hrcore/hrcore.civix.php
```diff
@@ -292,7 +292,9 @@ function _hrcore_civix_insert_navigation_menu(&$menu, $path, $item) {
     $first = array_shift($path);
     foreach ($menu as $key => &$entry) {
       if ($entry['attributes']['name'] == $first) {
-        if (!$entry['child']) $entry['child'] = array();
+        if (!isset($entry['child'])) {
+          $entry['child'] = array();
+        }
```
davialexandre: Is this a fix for that annoying notice that started appearing recently? How did you fix this manually? This file is created automatically by civix, so if you did that, there's a good chance the change will be lost when civix updates the file again.

mickadoo: Is this the notice you're talking about?

```
Notice: Undefined index: child in _hrcore_civix_insert_navigation_menu() (line 295 of /repos/civihr/uk.co.compucorp.civicrm.hrcore/hrcore.civix.php).
```

It's caused by the new items added from [this commit](https://github.com/civicrm/civihr/commit/87100db7ddb42cf2c79c1d75ffb3f621cef70cb3) although I wouldn't blame those changes. It seems a wrong way of doing it in the `civix` function, if they want to add the `child` array if it doesn't exist surely it should check `isset`.

I know these files can be overwritten, but I was thinking whoever makes changes to it in the future would probably see the reason for this change and keep it. If you like I can

PR #2524, opened 2018-03-08
PCHR-3420: Hide Options on Localisation Page

ajesamson:
## Overview
A new permission called `access root menu items and configurations` has been added. The implication of this on Localisation Page is that some fields will not be available for editing by users that do not have such permission.

## Before
before

## After
after

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/LocalisationPageFilter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/LocalisationPageFilter.php
@@ -1,6 +1,19 @@
 <?php
 
 class CRM_HRCore_Hook_BuildForm_LocalisationPageFilter {
+
+  private $restrictedSettings = [
+    'customTranslateFunction',
+    'contact_default_language',
+    'fieldSeparator',
+    'inheritLocale',
+    'lcMessages',
+    'legacyEncoding',
+    'monetaryThousandSeparator',
+    'monetaryDecimalPoint',
+    'moneyformat',
+    'moneyvalueformat'
+  ];
   /**
    * Hides options on the localisation page
    *
@@ -23,7 +36,7 @@ class CRM_HRCore_Hook_BuildForm_LocalisationPageFilter {
    * @return bool
    */
   private function shouldHandle($formName) {
-    if ($formName == CRM_Admin_Form_Setting_Localization::class) {
+    if ($formName === CRM_Admin_Form_Setting_Localization::class) {
       return TRUE;
     }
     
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -116,13 +116,43 @@ function hrcore_civicrm_buildForm($formName, &$form) {
...
 
   foreach ($listeners as $currentListener) {
     $currentListener->handle($formName, $form);
   }
 }
 
+/**
+ * Implements hook_civicrm_validateForm
+ *
+ * @param $formName
```
mickadoo: Missing types

mickadoo:
This is a decent working solution, although I have some concerns about it.

- Removing elements from a core form causes warnings because CiviCRM [expects those elements to exist](https://github.com/civicrm/civicrm-core/blob/ece6501cb7994371b48e870294590f23851266e7/CRM/Admin/Form/Setting/Localization.php#L131)
- Adding a CSS file to hide the help icons is like trying to treat the symptoms, but it would be good if we could find the source of the problem

I know there are methods for [freezing](https://pear.php.net/manual/en/package.html.html-quickform.html-quickform-element.freeze.php) form elements. I played around with this a little, but because of how CiviCRM [renders frozen elements](https://github.com/civicrm/civicrm-core/blob/e9bc5dcc5df622777ebda3e109142404152783c8/CRM/Core/Form/Renderer.php#L136) it's difficult to both freeze and hide an element.

I think an ideal solution would be one where the form structure is not altered, but where you cannot see or change certain elements if you don't have the permission.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/LocalisationPageFilter.php
```diff
@@ -0,0 +1,81 @@
...
+  
+  /**
+   * Checks if the hook should be handled.
+   *
+   * @param string $formName
+   *
+   * @return bool
+   */
+  private function shouldHandle($formName) {
+    if ($formName == CRM_Admin_Form_Setting_Localization::class) {
```
mickadoo: We can use strict comparison here

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/LocalisationPageFilter.php
```diff
@@ -0,0 +1,81 @@
...
+  }
+  
+  /**
+   * Hide options on localisation page if user does not have required permission
+   *
+   * @param CRM_Core_Form $form
+   */
+  private function filterLocalisationOptions($form) {
+    $canViewAllFields = CRM_Core_Permission::check('access root menu items and configurations ');
+    $settings = [
```
mickadoo: I noticed that the CSS classes we want to hide are all the same format and match the names of these `$settings`. I haven't checked this totally, but if they are the same it would be nice to combine them, for example using a private property like:

```
private $restrictedSettings = [
  'customTranslateFunction',
  ...
];
```

and then use it in both `filterLocalisationOptions` and `hideOptionLabels`. 

This would prevent duplication.

PR #2525, opened 2018-03-09
PCHR-3382: Modify Jobcontract.getContactsWithContractsInPeriod API And Reserve Work Location Type

tunbola:
## Problem
This PR makes some changes needed by [#324](https://github.com/compucorp/civihr-tasks-assignments/pull/324). 
- To send the daily reminder email, we need to ensure that the recipients have an active contract, the  
`Jobcontract.getContactsWithContractsInPeriod` provides this information but does not accept a list of contact ID's to check for. The `Jobcontract.getContractsWithDetailsInPeriod` accepts a list of contacts to check for but the information returned is far too much than what we need.
- Daily reminder emails are now to be sent to the primary work emails of a recipient with an active contract. The Work location type is not reserved in the database and as such the name can be changed which may affect the sql query in code where we have asked that the email be of the Work type.

location types home work _ staging17 2018-03-07 18-32-05

## Solution
- The `Jobcontract.getContactsWithContractsInPeriod` API was modified to accept a list of contacts to check for.
- The Work location type was reserved in the database via an upgrader in HRCore extension so that the name can not be edited via the UI.

location types home work _ staging17 2018-03-09 11-50-19

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
+++ b/hrjobcontract/CRM/Hrjobcontract/BAO/HRJobContract.php
@@ -641,10 +641,11 @@ class CRM_Hrjobcontract_BAO_HRJobContract extends CRM_Hrjobcontract_DAO_HRJobCon
    *  A date string in a format supported by strtotime
    * @param string|null $endDate
    *  A date string in a format supported by strtotime
+   * @param null|array $contactID
    *
    * @return array
    */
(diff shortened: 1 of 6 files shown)


tunbola:
Closing this because the branch name does not match the one in T&A repo.
Here's the new one: https://github.com/civicrm/civihr/pull/2526

PR #2531, opened 2018-03-14
PCHR-3451: New Option Group Work Pattern Change

ajesamson:
## Overview
'Work Pattern Change Reasons' currently uses a shared option group with contract change reason. Both of them use the option group hrjc_revision_change_reason. 'Work Pattern Change Reasons' should use a separate option group 'hrleaveandabsences_work_pattern_change_reason' which need to be created.

## Before
before

## After
after

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -512,6 +512,8 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
     $this->upgrade_1030();
     $this->upgrade_1032();
     $this->upgrade_1033();
+    $this->upgrade_1034();
+    $this->upgrade_1035();
     $this->upgrade_1036();
     $this->upgrade_1037();
   }
@@ -1254,37 +1256,42 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
   }
 
   /**
-   * Create new option group and values
+   * Create new option group hrleaveandabsences_work_pattern_change_reason
+   * with values As per contract, Change in contractual hours and
+   * Change in contract type.
+   *
+   * @return bool
    */
   public function upgrade_1037() {
+    $result = civicrm_api3('OptionGroup', 'get', [
+      'name' => 'hrleaveandabsences_work_pattern_change_reason'
+    ]);
+
+    if ($result['count'] > 0) {
+      return TRUE;
+    }
+    
+    $groupName = 'hrleaveandabsences_work_pattern_change_reason';
     $groupValues = [
       'As per contract',
       'Change in contractual hours',
       'Change in contract type'
     ];
+    $optionParams = [
+      'name' => $groupName,
+      'title' => 'Leave and Absence Work Pattern Change Reason',
+      'is_active' => 1
+    ];
 
-    $result = civicrm_api3('OptionGroup', 'get', [
-      'name' => 'hrleaveandabsences_work_pattern_change_reason'
-    ]);
+    civicrm_api3('OptionGroup', 'create', $optionParams);
 
-    if ($result['count'] == 0) {
-      $optionParams = [
-        'name' => "hrleaveandabsences_work_pattern_change_reason",
-        'title' => "Leave and Absence Work Pattern Change Reason",
-        'is_active' => 1
+    foreach ($groupValues as $value) {
+      $opValueParams = [
+        'option_group_id' => $groupName,
+        'name' => $value,
+        'label' => $value,
       ];
-
-      civicrm_api3('OptionGroup', 'create', $optionParams);
-
-      foreach ($groupValues as $key => $value) {
-        $opValueParams = [
-          'option_group_id' => 'hrleaveandabsences_work_pattern_change_reason',
-          'name' => $value,
-          'label' => $value,
-          'value' => $key,
-        ];
-        civicrm_api3('OptionValue', 'create', $opValueParams);
-      }
+      civicrm_api3('OptionValue', 'create', $opValueParams);
     }
 
     return TRUE;
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -513,6 +513,7 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
     $this->upgrade_1032();
     $this->upgrade_1033();
     $this->upgrade_1036();
+    $this->upgrade_1037();
   }
 
   function upgrade_1001() {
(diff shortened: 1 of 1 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -513,6 +513,7 @@ public function upgradeBundle() {
     $this->upgrade_1032();
     $this->upgrade_1033();
     $this->upgrade_1036();
```
mickadoo: I noticed calls to 1034 and 1035 are missing from `upgradeBundle`. It's not critical that these are run during installation (they will be run later anyway) but can you add them in this PR just to be consistent?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
     return TRUE;
   }
 
+  /**
```
mickadoo: missing `return`

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
...
+   * Create new option group and values
```
mickadoo: Do you think you can expand on the description a little bit, from a glance it should be clear which option group is being created

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
...
+      'As per contract',
+      'Change in contractual hours',
+      'Change in contract type'
+    ];
+
+    $result = civicrm_api3('OptionGroup', 'get', [
+      'name' => 'hrleaveandabsences_work_pattern_change_reason'
+    ]);
+
+    if ($result['count'] == 0) {
```
mickadoo: How about changing this to use an early return, like

```
if ($result['count'] > 0) {
  return TRUE;
}

// create the group
```

This would avoid an extra level of indentation for most of the function

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
...
+      $optionParams = [
+        'name' => "hrleaveandabsences_work_pattern_change_reason",
```
mickadoo: single quotes

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
...
+        'title' => "Leave and Absence Work Pattern Change Reason",
+        'is_active' => 1
+      ];
+
+      civicrm_api3('OptionGroup', 'create', $optionParams);
+
+      foreach ($groupValues as $key => $value) {
+        $opValueParams = [
+          'option_group_id' => 'hrleaveandabsences_work_pattern_change_reason',
```
mickadoo: maybe it would be cleaner, and shorter to make 'hrleaveandabsences_work_pattern_change_reason' a variable, since it's long and is used twice

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1253,43 @@ public function upgrade_1036() {
...
+          'name' => $value,
+          'label' => $value,
+          'value' => $key,
```
mickadoo: I don't think you need to set the value here, can you check? From what I remember CiviCRM will automatically assign it the next available value.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1254,37 +1263,42 @@ public function upgrade_1036() {
...
-      'name' => 'hrleaveandabsences_work_pattern_change_reason'
-    ]);
+    civicrm_api3('OptionGroup', 'create', $optionParams);
 
-    if ($result['count'] == 0) {
-      $optionParams = [
-        'name' => "hrleaveandabsences_work_pattern_change_reason",
-        'title' => "Leave and Absence Work Pattern Change Reason",
-        'is_active' => 1
+    foreach ($groupValues as $key => $value) {
```
mickadoo: `$key` is not used

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -497,21 +497,30 @@ public function upgradeBundle() {
     $this->upgrade_1006();
     $this->upgrade_1008();
     $this->upgrade_1009();
+    $this->upgrade_1010();
```
mickadoo: I think some of these might be intentionally omitted, can we remove all new calls in this PR except for the one to your new upgrader?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -512,7 +512,10 @@ public function upgradeBundle() {
     $this->upgrade_1030();
     $this->upgrade_1032();
     $this->upgrade_1033();
+    $this->upgrade_1034();
+    $this->upgrade_1035();
```
davialexandre: What are these methods? Why are you adding them here if the only method you added is upgrade_1037?

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1252,6 +1255,48 @@ public function upgrade_1036() {
...
 
+  /**
+   * Create new option group hrleaveandabsences_work_pattern_change_reason
+   * with values As per contract, Change in contractual hours and
+   * Change in contract type.
+   *
+   * @return bool
+   */
+  public function upgrade_1037() {
+    $result = civicrm_api3('OptionGroup', 'get', [
```
davialexandre: This option group will be used by leave and absences, so it should be added there, not here. Besides that, new option groups should be defined using a XML file, like the ones we have [here](https://github.com/civicrm/civihr/tree/staging/uk.co.compucorp.civicrm.hrleaveandabsences/xml/option_groups). This XML will be loaded automatically when creating a new site, but for existing site you'll still need an upgrader. Check if it's possible to make the upgrader use the definition from the XML file.

Finally, we might want to make this Option Group reserved. I don't think it will be a good idea to let the users delete it (I think it's ok if they update or change the option values though).

ajesamson: This will be migrated as instructed. A new ticket has been created to handle defining new option group in the XML file like the ones suggested.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
ajesamson: There were implementations but no call to the upgrade function, causing some inconsistencies with the upgrade numbering. That was the reason i added those two.

PR #2533, opened 2018-03-15
PCHR-3162: Update label

mickadoo:
## Overview

This fixes issues reported in the QA for the configuration menu changes

## Before

One of the labels was incorrect

## After

'Normal Place of Work' is now 'Normal Places of Work'


ajesamson:
Looks good.

PR #2534, opened 2018-03-15
PCHR-3438: Allow set times and custom duration in TOIL request

igorpavlov:
## Overview

At the moment you cannot specify time for TOIL requests and cannot specify custom deduction based on the date/time selection. This PR introduces this possibility.

### 1. Move everything related to Leave type (not TOIL type) to a separate controller.

In order to start extending TOIL Subcontroller we need to get rid of the logic from the Details Component, that is not really related to TOIL. This can be accomplished by moving it to the Leave Subcontroller (note, Sickness Subcontroller inherits the functionality from the Leave Subcontroller).

The idea is to introduce more "extensible" functions in the Details Component and extend them in the Leave Subcontroller, so the functionality remains the same for Leave/Sickness types, but is excluded from TOIL Subcontroller. Here is an example:

```js
// before
// Details Component
.then(function () {
  if (!vm.isMode('create') && isCalculationUnit('hours')) {
    return initTimes() // in TOIL the initiation is completely different
// after
// Details Component
.then(function () {
  if (!vm.isMode('create')) {
    return $q.resolve()
      .then(vm.initTimesExtended) // Leave has its own initiation, TOIL has its own as well
```

### 2. Remove old code

Now TOIL utilises Time Amount Picker directive and does not need code for conversion from hours+minutes to hours since the directive handles this itself.

```js
// TOIL instance
transformAttributes: function (attributes) { // not needed anymore
  var duration = Number(attributes.toil_duration);
  if (duration) {
    attributes.toilDurationHours = Math.floor(duration / 60).toString();
    attributes.toilDurationMinutes = (duration % 60).toString();
  }
```

### 3. New duration calculation

The new default duration is now calculated based on the dates and times you specify. Basically, the new duration is a difference between two dates/times.

```js
function calculateDuration () {
  // ...
  detailsController.uiOptions.max_toil_duration_and_accrual = moment.duration(moment(detailsController.request.to_date).diff(detailsController.request.from_date)).asHours();
```

If the TOIL is in hours, then the default Accrue value is based on the selected Duration.

```js
$rootScope.$watch(
  function () {
   return detailsController.uiOptions.toil_duration_in_hours;
  },
  function (oldValue, newValue) {
    if (detailsController.isCalculationUnit('hours') && oldValue !== newValue) {
      detailsController.request.toil_to_accrue =
      detailsController.uiOptions.toil_duration_in_hours;
    }
    // ...
```

The ranges for time inputs are static (unlike in Leave/Sickness where they depend on the Work Pattern).

```js
function setTimeInputsRanges () {
  if (detailsController.uiOptions.multipleDays) {
    ['from', 'to'].forEach(function (dateType) {
      detailsController.uiOptions.times[dateType].min = '00:00';
      detailsController.uiOptions.times[dateType].max = '23:45';
    });
  } else {
    detailsController.uiOptions.times.from.min = '00:00';
    detailsController.uiOptions.times.from.max = '23:30';
    detailsController.uiOptions.times.to.min = '00:15';
    detailsController.uiOptions.times.to.max = '23:45';
  }
}
```

The templates also have been amended to support Custom Duration.

```html
<div class="row"
  ... Default Duration
        <strong>{{detailsTab.uiOptions.toil_duration_in_hours | timeUnitApplier : 'hours'}}
          of overtime worked</strong>
        <a class="pointer"
          ng-click="detailsTab.uiOptions.toil_duration_expanded = true">
          Customise this amount</a>
  ... Custom Duration
        <span>Customise deduction for </span>
        <strong>{{detailsTab.uiOptions.fromDate | formatDate}}</strong>
        <br class="hidden-xs"/>
        <span class="text-muted">(eg. exclude lunch time)</span>
  ... Time amount picker for Duration value
        <time-amount-picker
          time-amount-picker-value="detailsTab.uiOptions.toil_duration_in_hours"
          time-amount-picker-min-amount="0"
          time-amount-picker-max-amount="detailsTab.uiOptions.max_toil_duration_and_accrual"
          time-amount-picker-interval="detailsTab.uiOptions.time_interval"
        ></time-amount-picker>
  ... Time amount picker for Accrual value
        <time-amount-picker
          ...
        ></time-amount-picker>
```

### 4. Do not amend time if TOIL

Previously, on the backend in case of a request **not** in hours, the time will automatically be adjusted to `00:00` for "from" datetime and to `23:59` for "to" datetime. Since TOIL now accepts time for both *hours* and *days* absence types, this logic should be amended:

```php
if($absenceType->isCalculationUnitInHours() ||
  $params['request_type'] === CRM_HRLeaveAndAbsences_BAO_LeaveRequest::REQUEST_TYPE_TOIL) { // this condition added
return; // skip amending
```
### 5. Sort TOIL amounts for TOIL in Days

```js
function loadToilAmounts () {
  return OptionGroup.valuesOf('hrleaveandabsences_toil_amounts')
    .then(function (amounts) {
      detailsController.toilAmounts = _.indexBy(amounts, 'value'); // before
      detailsController.toilAmounts = _.sortBy(amounts, 'value'); // after
```

### 6. Remove stale code

Disabling time fields is not needed anymore (not because of this particular PR), but because the time should be disabled if only the date is not presented. The timepickers should not also be disabled while they are loading - this is an unnecessary action since they are hidden during load.

## Comments

This PR also fixes a couple of side issues:

1) Fixes the layout for the "to" deduction for leave request in hours.
2) Fixes the JavaScript error when you open a leave request modal and switch to "hours" absence type immediately.

⚠️Please note, Github hides large diffs, so make sure to expand source files with large amount of changes.

-----

✅Manual Tests - passed
✅Jasmine Tests - passed
✅PHP Unit Tests - passed
⏹Backstop tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -841,19 +593,14 @@ define([
      *
      * @param {String} dateType from|to
      */
-    function resetUIDayTypesTimeAndDeductions (dateType) {
+    function resetUIInputs (dateType) {
```
reneolivo: `resetUIDayTypesTimeAndDeductions` was more specific, why rename it?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
     $this->assertEquals($toDate->format('Y-m-d') ." 23:59:00", $leaveRequest->to_date);
   }
 
+  public function testTheTimeForFromAndToDateOfLeaveRequestStaysUntouchedWhenRequestIsTOIL() {
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'allow_accruals_request'   => true,
+      'max_leave_accrual'        => 10,
```
tunbola: Please leave a single space between the array key operand and value i.e` 'max_leave_accrual' => 10,` for this array.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
...
+      // The calculation unit is in days since otherwise a check for
+      // calculation unit in days will take action and this test will be invalid
```
tunbola: Not sure I understand this comment, if we don't set the calculation unit the fabricator will set it to days by default so no need to pass the calculation unit here.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
...
+      'calculation_unit'         => 1,
+      'allow_accrue_in_the_past' => true
+    ]);
+    $expectedFromDate = '2016-01-08 01:23:45';
+    $expectedToDate = '2016-01-10 12:34:56';
+    $fromDate = new DateTime($expectedFromDate);
+    $toDate = new DateTime($expectedToDate);
+    $contactID = 1;
+    $this->registerCurrentLoggedInContactInSession($contactID);
```
tunbola: Don't think we need this, test should still work without it since we are not checking permissions.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
...
+      'max_leave_accrual'        => 10,
+      // The calculation unit is in days since otherwise a check for
+      // calculation unit in days will take action and this test will be invalid
+      'calculation_unit'         => 1,
+      'allow_accrue_in_the_past' => true
+    ]);
+    $expectedFromDate = '2016-01-08 01:23:45';
+    $expectedToDate = '2016-01-10 12:34:56';
+    $fromDate = new DateTime($expectedFromDate);
+    $toDate = new DateTime($expectedToDate);
```
tunbola: Why not just 
```
$fromDate = new DateTime('2016-01-08 01:23:45');
$toDate = new DateTime('2016-01-10 12:34:56');
``` 
You can always do `$fromDate->format('YmdHis')` anywhere you need the datetime.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
...
+
+    $period = AbsencePeriodFabricator::fabricate([
+      'start_date' => $fromDate->format('YmdHis'),
+      'end_date' => $toDate->format('YmdHis'),
+    ]);
+
+    $contract = HRJobContractFabricator::fabricate(
+      ['contact_id' => $contactID],
+      ['period_start_date' => $fromDate->format('Y-m-d')]
+    );
```
tunbola: Thinking about this now, will it be a good idea to have a `instantiateLeaveRequirements` method that creates contract, entitlement and all for tests that only need the requirement with no particular care to date and balance amount and we can put that in a trait and call it anytime we have such tests and then focus on the test proper.
Is that a good idea @davialexandre ?
This is just a comment and not part of the review @igorpavlov

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4366,6 +4367,61 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
...
+      'type_id' => $absenceType->id,
+      'from_date' => $expectedFromDate,
+      'from_date_type' => $this->leaveRequestDayTypes['all_day']['value'],
+      'to_date' => $expectedToDate,
+      'to_date_type' => $this->leaveRequestDayTypes['all_day']['value'],
+      'status_id' => 3,
+      'toil_duration' => 10,
+      'toil_to_accrue' => 2,
+      'request_type' => LeaveRequest::REQUEST_TYPE_TOIL,
+      'sequential' => 1
```
tunbola: no need for sequential.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/toil-request.instance.js
```diff
@@ -21,45 +21,22 @@ define([
...
-        },
-
         /**
          * Override of parent method
          *
          * @param {object} result - The accumulator object
          * @param {string} key - The property name
          */
         toAPIFilter: function (result, __, key) {
-          if (!_.includes(['balance_change', 'dates', 'comments', 'files', 'toilDurationHours', 'toilDurationMinutes'], key)) {
```
reneolivo: why was `toilDurationHours` removed but not `toilDurationMinutes` ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -289,22 +284,12 @@ define([
...
-            return loadTimeRangesFromWorkPattern('from')
-              .then(function () {
-                setDeductionMaximumBoundary('from', true);
-              })
-              .then(setRequestDateTimesAndDateTypes);
-          }
-        })
+        .then(setRequestDateTimesAndDateTypes)
         .then(vm.setDaysSelectionModeExtended)
+        .then(finishLoadingTimesAndDateTypes)
```
reneolivo: isn't this supposed to go in a `.finally` statement?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
```diff
@@ -1039,15 +744,9 @@ define([
...
-          if (absenceTypeUnitChanged && isCalculationUnit('hours')) {
-            return loadTimeRangesFromWorkPattern('from')
-              .then(function () {
-                setDeductionMaximumBoundary('from', true);
-              });
-          }
-        })
+        .then(absenceTypeUnitChanged && vm.onAbsenceTypeUpdateExtended)
         .then(absenceTypeUnitChanged && setRequestDateTimesAndDateTypes)
+        .then(finishLoadingTimesAndDateTypes)
```
reneolivo: `.finally` ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-leave.controller.spec.js
```diff
@@ -170,6 +170,21 @@ define([
             expect(controller.canCalculateChange()).toBe(true);
           });
         });
+
+        describe('when unit changed to hours by changing an absence type', function () {
```
reneolivo: `when the unit changes to hours because the absence type changed`

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: Sure. I just followed this from other places in the same file.

igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: Done.

igorpavlov: image

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: I would like to keep assertions as clean as possible. I would like to compare values to simple strings rather than datetime object formatted.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: @tunbola I think this is an awesome idea, that will reduce the code dup and will increase the clarity of tests suites.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
igorpavlov: It is now a generic function, amended it, there was a mistake actually.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/instances/toil-request.instance.js
igorpavlov: 🤷

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
igorpavlov: @reneolivo I don't think so, this func needs to fire **before** balance calculation.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-request-popup-details-tab.component.js
igorpavlov: Same as above ^ @reneolivo

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-leave.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
tunbola: Won't make it less clean if we keep them as the date objects

PR #2538, opened 2018-03-20
PCHR-3474: Add TOIL times in reports

igorpavlov:
## Overview

This PR adds times for TOIL requests in Staff Leave Report and Manage Leave Report.

## Before

image

image

## After

image

image

## Technical Details

1. Create a shared partial for request dates/times:

```html
<!-- only conditions changed -->
{{request.from_date | formatDate : null : (request.request_type === 'toil' ? 'hours' : unit)}}
<span ng-if="request.dates.length > 1">
  -
  {{request.to_date | formatDate : null : (request.request_type === 'toil' ? 'hours' : unit)}}
</span>
<span ng-if="request.dates.length === 1 && (unit === 'hours' || request.request_type === 'toil')">
  -
  {{request.to_date.slice(11, 16)}}
</span>
```

2. Then replace the dup code with a new partial in reports.

```html
<td ng-init="unit = absenceType.calculation_unit_name"
  ng-include="$root.sharedPathTpl + 'components/partials/leave-request-dates.html'"></td>
```

----

Only manual tests were performed.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/partials/leave-request-dates.html
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/partials/leave-request-dates.html
@@ -1,0 +1,9 @@
+{{request.from_date | formatDate : null : (request.request_type === 'toil' ? 'hours' : unit)}}
+<span ng-if="request.dates.length > 1">
+  -
+  {{request.to_date | formatDate : null : (request.request_type === 'toil' ? 'hours' : unit)}}
+</span>
+<span ng-if="request.dates.length === 1 && (unit === 'hours' || request.request_type === 'toil')">
+  -
+  {{request.to_date.slice(11, 16)}}
+</span>
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/manage-leave-requests.html
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/manage-leave-requests.html
@@ -184,17 +184,8 @@
                       {{request.balance_change > 0 ? "+" : ""}}{{request.balance_change |
                         timeUnitApplier : absenceType.calculation_unit_name}}
                     </td>
-                    <td ng-init="unit = absenceType.calculation_unit_name">
-                      {{request.from_date | formatDate : null : request.request_type !== 'toil' ? unit : null}}
-                      <span ng-if="request.dates.length > 1">
-                        -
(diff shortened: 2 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/partials/leave-request-dates.html
```diff
@@ -0,0 +1,9 @@
+{{request.from_date | formatDate : null : (request.request_type === 'toil' ? 'hours' : unit)}}
```
deb1990: If request type is `toil`, we are selecting unit as hours. What if the date does not have hours(normal date only), what will be displayed then?

deb1990:
@igorpavlov  The way we are displaying dates and times in the L&A is not really readable. 
`16/03/2018 00:30` is really hard to read, instead something like `March 16th, 2018 12:30am`(format can be decided) would be much better. Also changing the UI looks better to me
2018-03-21 at 11 27 am

What do you think? Also could you discuss this with Mike and Jon and see if they agree. Then we can create a new ticket.

igorpavlov:
@deb1990 the date format is defined by CRM settings. I realise it may be difficult to read probably because of a personal preference and/or a place where you live. In different countries there are different formats, I also find some formats difficult to read, but that's because I got used to specific ones. For example, I cannot easily read MM/DD/YYYY, but it is a standard in US.

Please have a look at this article https://en.wikipedia.org/wiki/Date_format_by_country.

The format for time is always HH:mm, no matter what the preference is. I realise for some people 11pm is more familiar than 23:00 but this is how it works now, this was confirmed and passed QA earlier when Leave in Hours was introduced.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/partials/leave-request-dates.html
igorpavlov: In the database we store datetime at the moment. As per PCHR-3427, TOIL will always have times, both "from" and "to". Currently, the datetime is stored as 00:00-23:59 for TOIL. This is why PCHR-3476 exists, it will convert 23:59 to 23:45.

So, answering your question: after PCHR-3427 is finished, it will not be possible to create a TOIL without times specified. Already existing requests will also have times: 00:00-23:45.

deb1990:
@igorpavlov yeah I totally forgot about the CRM settings. Yes it does not make sense to change it as user still has option to change it.

PR #2544, opened 2018-03-22
PCHR-3440: Fix JobContract Pay Tab fields after importing a Contract for a Contact

tunbola:
## Overview
After Importing a Contact and subsequently importing a contract for the contact, the Pay, Pay Scale/Grade, Pay, Pay Cycle fields are not editable and the dropdown fields are not populated with any options. 

## Before
- There was a JS error in the console `TypeError: Cannot read property 'length' of null`. After debugging ,  the error is coming from trying to get the length property of some variables that are not yet set. Mainly the `entityPay.annual_benefits` and `entityPay.annual_deductions`.

This issue is a regression which was introduced by this [PR](https://github.com/civicrm/civihr/pull/2345)
which moved the initialization of the `entityPay.annual_benefits` and `entityPay.annual_deductions` variables from its original position before the `$scope.collapseBenDed = !entityPay.annual_benefits.length && !entityPay.annual_deductions.length;` expression.

screenshot-3 1

## After
 These expression utilizing these variables where moved to the init function just right after these variables were initialized. The Pay, Pay Scale/Grade, Pay, Pay Cycle fields are now editable and the dropdown fields are  populated with relevant options. 

rachaeljones gmail com _ staging17 2018-03-22 17-55-03

## Tests
Manual tests were done. Initially after importing the contract before the change, the fields Pay, Pay Scale/Grade, Pay, Pay Cycle fields could not be edited as the select options were not even populated due to the JS error. After the change, the fields were populated and the information displayed as it should, I could also save and the values selected for these fields were retained when viewing the tab again after the save.


igorpavlov:
Just a question: the radio buttons look off the layout, was it always like that?

igorpavlov:
Please add according labels to this PR. Please add at least Bug and Ready to test.

Was this issue a regression? If it was, the appropriate label should be added and also it should be explained when and why this issue appeared.

igorpavlov:
Is this piece of code covered with tests?

igorpavlov:
Please add a section describing which test were performed to make sure this is a good change. If there were only manual tests, it is essential to describe what exactly was tested.

tunbola:
@igorpavlov , Yeah, the radio buttons look off, Its like that on staging and on local. Not sure If it has always been like that.

hrjobcontract/js/src/job-contract/controllers/form/form-pay.controller.js
```diff
@@ -28,7 +28,6 @@ define(function () {
       Hour: 2080
     };
 
-    $scope.collapseBenDed = !entityPay.annual_benefits.length && !entityPay.annual_deductions.length;
```
igorpavlov: I would not actually move $scope to init(). I would rather suggest to check for the prop to exist:

```js
$scope.collapseBenDed =
  (!entityPay.annual_benefits || !entityPay.annual_benefits.length) &&
  (!entityPay.annual_deductions || !entityPay.annual_deductions.length);
```

hrjobcontract/js/src/job-contract/controllers/form/form-pay.controller.js
```diff
@@ -50,6 +49,7 @@ define(function () {
       entityPay.pay_is_auto_est = '0';
       entityPay.annual_benefits = entityPay.annual_benefits || [];
       entityPay.annual_deductions = entityPay.annual_deductions || [];
+      $scope.collapseBenDed = !entityPay.annual_benefits.length && !entityPay.annual_deductions.length;
```
igorpavlov: See above, this line would not be then needed.

PR #2443, opened 2018-01-29
PCHR-3164: Add End Time for a single day request in hours

igorpavlov:
## Overview

This PR allows to specify End ("to") Time for Single Day Requests in Hours and also shows End Time for such requests in both Staff Leave Report and Manage Leave Report.

#### Mobile view

image

### Preparation for UI customisation

In order to create a custom UI for single day request in hours. 4 templates were created for each (A) calculation unit - hours/days and (B) days selection mode - multiple/single,

```html
<!-- common.html -->
<ng-include src="$root.sharedPathTpl +
  'components/leave-request-popup/details-tab/by-unit-and-days-mode/' +
  +  detailsTab.selectedAbsenceType.calculation_unit_name + '-unit-' +
  +  (detailsTab.uiOptions.multipleDays ? 'multiple-days' : 'single-day') +
  +  '.html'"></ng-include>
```

image

This change was needed since the From Deduction field goes *after* the End Time input in the requirements, and in the current template it is vice versa. Since we don't want to have duplicate code in one file, it was decided to split the view into these 4 templates and then freely customise them.

### Allowing to specify end time for single leave request in hours

Preparing the End Time field:

```js
function loadTimeRangesFromWorkPattern (type) {
  // ...
  var isSingleDayRequest = !vm.uiOptions.multipleDays;

  // ...

  disableAndShowLoadingTimeInput(type);
  isSingleDayRequest && disableAndShowLoadingTimeInput('to');

  return vm.request.getWorkDayForDate(convertDateToServerFormat(date))
    .then(function (response) {
      // ...

      enableAndSetDataToTimeInput(type, response);
      isSingleDayRequest && enableAndSetDataToTimeInput('to', response);
    })
    .catch(handleError)
    .finally(function () {
      timeObject.loading = false;

      isSingleDayRequest && (vm.uiOptions.times['to'].loading = false);
    });
}
```

And finally rendering it in the template:

```html
<div class="row" ng-hide="detailsTab.isLeaveType('toil')">
  <div class="col-xs-12 form-horizontal">
    <div class="form-group">
      <label class="col-sm-4 control-label required-mark">
        End time
      </label>
      <div class="col-sm-4 chr_leave-request-date-picker-container">
        <crm-loading show="!detailsTab.uiOptions.times.to.loading"
          class="chr_leave-request-modal__small_spinner" // IMPORTANT, TIME "TO" DEPENDS ON DATE "FROM"
[...]

PR #2546, opened 2018-03-23
PCHR-3460: Add Tool Icon to Edit Job Role Options

ajesamson:
## Overview
Job role options such as location, region, department and level do not have link to edit their details right from the job role page. This PR caters for that by providing a wrench icon next to each of them which can be clicked for editing purpose.

## Before
before

## After
after

## Technical Details
In order to add tool icons, an anchor tag with wrench icon type child element was added next to each select elements to be modified. 

```html
<a class="pointer">
      <i class="crm-i fa-wrench" ng-click="jobroles.openOptionsEditor('/civicrm/admin/options/hrjc_level_type?reset=1', 'hrjc_level_type')"></i>
</a>
```

Attached to the wrench icon is a controller class to handle loading the page in a pop up window, editing and creating new item. This is achieved by calling another service called `crmAngService`
```js
function openOptionsEditor (url, optionType) {
      crmAngService.loadForm(url)
        .on('crmUnload', function () {
          getOptionGroupTypeValues(optionType);
        });
    }
```
---
✅Manual Tests - passed
✅Jasmine Tests - passed


igorpavlov:
Please enhance this PR by:

1) Adding appropriate labels. It looks like this is an "enhancement". Please also add "Ready to test" label to spawn Jenkins automated testing and make sure to wait until all tests pass before sending the ticket to QA.

2) Please provide technical description like here https://github.com/civicrm/civihr/pull/2443. You should provide some code snippets, for example:
```html
<div class="form-inline"> <!-- needs to be inline to... -->
  <div class="crm_custom-select">
    <ui-select
      ... <!-- ng-model is used to... -->
      ng-model="jobroles.editData['new_role_id']['location']"
      ...
      name="newLocation">
      <ui-select-match prevent-animations allow-clear placeholder="Select a location">{{$select.selected.value.title}}</ui-select-match>
      <ui-select-choices prevent-animations repeat="location.value.value as (key,location) in jobroles.LocationsData | filter: { value: { title: $select.search }}">
        <div ng-bind-html="location.value.title | highlight: $select.search">
  <a href="/civicrm/hours_location"
     ...
     data-option-edit-path="civicrm/hours_location">
    <i class="crm-i fa-wrench">
```

3) Please also provide a section with explaining which tests did you run. Do we have BackstopJS coverage for that page?

4) Please replace JPEGs with GIFs in the Before and After sections - that would allow to instantly understand why did you add these wrenches.

5) In the Overview you say that "gear" icons added, however, they are "wrenches".

com.civicrm.hrjobroles/views/include/add_new_role.html
```diff
@@ -85,63 +85,104 @@
...
-                    ng-model="jobroles.editData['new_role_id']['location']"
-                    theme="civihr-ui-select"
-                    title="Choose a location"
-                    name="newLocation">
-                    <ui-select-match prevent-animations allow-clear placeholder="Select a location">{{$select.selected.value.title}}</ui-select-match>
-                    <ui-select-choices prevent-animations repeat="location.value.value as (key,location) in jobroles.LocationsData | filter: { value: { title: $select.search }}">
-                      <div ng-bind-html="location.value.title | highlight: $select.search"></div>
-                    </ui-select-choices>
-                  </ui-select>
+                  <div class="form-inline">
```
igorpavlov: There are 8 very similar blocks added in this PR. Is there a potential way to reduce the code duplication? Maybe we can create a partial template and simply pass parameters in there?

com.civicrm.hrjobroles/views/include/add_new_role.html
```diff
@@ -85,63 +85,104 @@
...
+                        <ui-select-match prevent-animations allow-clear placeholder="Select a level">{{$select.selected.value.title}}</ui-select-match>
+                        <ui-select-choices prevent-animations repeat="level.value.value as (key,level) in jobroles.LevelsData | filter: { value: { title: $select.search }}">
+                          <div ng-bind-html="level.value.title | highlight: $select.search"></div>
+                        </ui-select-choices>
+                      </ui-select>
+                    </div>
+                    <a href="/civicrm/admin/options?gid=111&reset=1"
+                       class="crm-option-edit-link medium-popup crm-hover-button"
+                       target="_blank" title="Edit Task Types"
+                       data-option-edit-path="civicrm/admin/options?gid=111&reset=1">
```
igorpavlov: Why do we use `gid=111`? Looks like we request data by fixed ID. Are there any other places in the code using the same technique? If yes, please refer to the exact line of existing code. This is absolutely essential to describe this in the Technical Description section.

com.civicrm.hrjobroles/views/include/add_new_role.html
ajesamson: Good idea. I will explore that.

com.civicrm.hrjobroles/views/include/add_new_role.html
ajesamson: I actually worked with a [document](https://docs.google.com/spreadsheets/d/1cXYoRoBv0XV5UXruxw33y_OR4CiyIwLRKjbh4i1yEGo/edit#gid=0) in order to get the url to render. I will check if there is a better way to achieve this.

igorpavlov: @ajesamson Maybe this is an expected approach? Please look at other examples in the existing code and provide a reference to them at first.

com.civicrm.hrjobroles/views/include/add_new_role.html
```diff
@@ -85,63 +85,104 @@
...
+                      </ui-select>
+                    </div>
+                    <a href="/civicrm/admin/options?gid=117&reset=1"
+                       class="crm-option-edit-link medium-popup crm-hover-button"
+                       target="_blank" title="Edit Task Types"
+                       data-option-edit-path="civicrm/admin/options?gid=117&reset=1">
+                      <i class="crm-i fa-wrench"></i>
+                    </a>
+                  </div>
+
```
AkA84: empty line

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -11,12 +11,12 @@ define([
   JobRolesController.$inject = [
     '$filter', '$log', '$q', '$rootElement', '$route', '$routeParams', '$scope',
     '$timeout', '$uibModal', 'DOMEventTrigger', 'settings', 'HR_settings',
-    'dateValidation', 'filtersService', 'jobRoleService', 'pubSub'
+    'dateValidation', 'filtersService', 'jobRoleService', 'pubSub', 'crmAngService'
```
igorpavlov: Please arrange alphabetically:

```js
'$filter', '$log', '$q', '$rootElement', '$route', '$routeParams', '$scope',
'$timeout', '$uibModal', 'crmAngService', 'dateValidation', 'DOMEventTrigger',
'filtersService', 'HR_settings', 'jobRoleService', 'pubSub', 'settings'
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -475,6 +476,19 @@ define([
       });
     }
 
+    /**
+     * Trigger opening of editor window for editing provided option type
```
igorpavlov: Please always describe the function like "what it does":

`Opens editor window for editing a provided option type`

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -475,6 +476,19 @@ define([
...
+     *
+     * @param url
```
igorpavlov: I don't think you need URL to pass considering which URLs I see you pass in HTML templates:

```js
function openOptionsEditor (optionType) {
  var url = '/civicrm/admin/options/' +optionType + '?reset=1';

  crmAngService.loadForm(url)
    .on('crmUnload', function () {
      getOptionGroupTypeValues(optionType);
    });
}
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -524,6 +538,8 @@ define([
                       value: data.values[i]['value'],
                       is_active: data.values[i]['is_active']
                     };
+                    // Store the Department types so we can reuse later
```
igorpavlov: We are always encouraged to refactor old code (which is a duplicate feast at the moment). I would suggest smth like that, but with caution. Please make sure there are test suites for that part of code, if not, maybe this change is not worth it.

```js
// Replacement starting from `var DepartmentList = {};`
var optionsData = {
  'hrjc_department': { storage: 'DepartmentsData', data: {} },
  ...,
  'cost_centres': { storage: 'CostCenters', data: [] }
};

angular.forEach(data['optionGroupData'], function (optionGroupId, optionGroupName) {
  for (var i = 0; i < data.count; i++) {
    if (optionGroupId === data.values[i]['option_group_id']) {
      if (optionGroupName === 'cost_centres') {
        optionsData[optionGroupName].push({
          id: data.values[i]['id'],
          title: data.values[i]['label'],
          weight: data.values[i]['weight'],
          is_active: data.values[i]['is_active']
        });
      } else {
        optionsData[optionGroupName].data[data.values[i]['value']] = {
          id: data.values[i]['id'],
          title: data.values[i]['label'],
          value: data.values[i]['value'],
          is_active: data.values[i]['is_active']
        };
      }

      vm[optionsData[optionGroupName].storage] = getActiveValues(optionsData[optionGroupName].data);
    }
  }
});
```

cc @AkA84

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -33,6 +34,30 @@ define([
...
+        });
+        initController();
+        ctrl.openOptionsEditor(url, optionType);
+      });
+
+      it('calls the crmAngService with the requested url', function () {
+        expect(crmAngService.loadForm).toHaveBeenCalledWith(url);
+      });
+    });
+
```
igorpavlov: The tests that check for `vm.DepartmentsData` and similar storages are missing. Please add them.

com.civicrm.hrjobroles/views/include/add_new_role.html
```diff
@@ -99,11 +99,8 @@
                         </ui-select-choices>
                       </ui-select>
                     </div>
-                    <a href="/civicrm/admin/options/hrjc_location?reset=1"
-                       class="crm-option-edit-link medium-popup crm-hover-button"
-                       target="_blank" title="Edit Locations"
-                       data-option-edit-path="civicrm/admin/options/hrjc_location?reset=1">
-                      <i class="crm-i fa-wrench"></i>
+                    <a class="pointer">
+                      <i class="crm-i fa-wrench" ng-click="jobroles.openOptionsEditor('/civicrm/admin/options/hrjc_location?reset=1', 'hrjc_location')"></i>
```
igorpavlov: If you are OK with removing URL param from the function, please make sure to update here as well.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
igorpavlov: @ajesamson actually, since we are lacking suites that test `vm. DepartmentsData` etc, I think this function should be refactored after tests are added. Writing this because in the comment above I was not sure if it worth it, now I realise it is.

cc @AkA84

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -477,12 +477,13 @@ define([
     }
 
     /**
-     * Trigger opening of editor window for editing provided option type
+     * Open editor for option editing
```
igorpavlov: The descriptions should say what function does, not what you could do with it. So `Opens` instead of `Open`.

NO: I can "Open editor for option editing"
YES: It (function) "Opens editor for option editing"

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -511,98 +512,36 @@ define([
...
+                    value: data.values[i]['value'],
+                    is_active: data.values[i]['is_active']
+                  };
+                }
+
+                vm[optionsData[optionGroupName].storage] = getActiveValues(optionsData[optionGroupName].data);
               }
             }
           });
-
```
igorpavlov: That line break looks actually OK to me. It separates assignments from for-loop.

We have a Styleguide for AngularJS https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS that we try to follow.

Could you please revert it back?

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -50,10 +50,10 @@ define([
           };
         });
         initController();
-        ctrl.openOptionsEditor(url, optionType);
       });
 
       it('calls the crmAngService with the requested url', function () {
+        ctrl.openOptionsEditor(optionType);
```
igorpavlov: Why was it moved? We should keep `it` blocks as clean as possible.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -75,18 +75,6 @@ define([
               'is_reserved': '0',
               'is_active': '1'
             },
-            {
```
igorpavlov: Why was it moved?

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -126,22 +114,82 @@ define([
...
+            'hrjc_region': '107'
           }
         }));
         initController();
         $rootScope.$digest();
       });
 
       it('builds the "CostCentreList" array of objects containing the "weight" property', function () {
-        expect(ctrl.CostCentreList.length).toBe(3);
+        expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
```
igorpavlov: Why was it an array and now it is an object?

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -126,22 +114,82 @@ define([
...
         expect(ctrl.CostCentreList[0].weight).not.toBeNull();
         expect(ctrl.CostCentreList[1].weight).not.toBeNull();
         expect(ctrl.CostCentreList[2].weight).not.toBeNull();
       });
+
+      it('builds the "hrjc_level_type" array of objects containing the "value" property', function () {
+        expect(Object.keys(ctrl.LevelsData).length).toBe(1);
+        expect(ctrl.LevelsData.value).not.toBeNull();
```
igorpavlov: No need for testing if it is not null, the test above will fail if it is null anyway. Same for other test suites.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -126,22 +114,82 @@ define([
...
       it('builds the "CostCentreList" array of objects containing the "weight" property', function () {
-        expect(ctrl.CostCentreList.length).toBe(3);
+        expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
         expect(ctrl.CostCentreList[0].weight).not.toBeNull();
         expect(ctrl.CostCentreList[1].weight).not.toBeNull();
         expect(ctrl.CostCentreList[2].weight).not.toBeNull();
       });
+
+      it('builds the "hrjc_level_type" array of objects containing the "value" property', function () {
+        expect(Object.keys(ctrl.LevelsData).length).toBe(1);
```
igorpavlov: Can we test that the data matches the mocks? The test checks if something is inside, but now *what* is inside.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -126,22 +114,82 @@ define([
...
 
       it('builds the "CostCentreList" array of objects containing the "weight" property', function () {
-        expect(ctrl.CostCentreList.length).toBe(3);
+        expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
         expect(ctrl.CostCentreList[0].weight).not.toBeNull();
         expect(ctrl.CostCentreList[1].weight).not.toBeNull();
         expect(ctrl.CostCentreList[2].weight).not.toBeNull();
       });
+
+      it('builds the "hrjc_level_type" array of objects containing the "value" property', function () {
```
igorpavlov: Array or objects is called a *collection* in JS

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
ajesamson: Alright noted.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
ajesamson: Alright noted.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
ajesamson: There is no specific reason. I just felt it should be there, showing action that took place and what is expected. Will be moving it back.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
ajesamson: I want to make it consistent with other sample data (other than CostCentreList)

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
ajesamson: Will be renaming to collection.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
ajesamson: Its because of the way the data is now stored `{data: []}' instead of just `[]`

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -126,22 +114,82 @@ define([
...
+      it('builds the "CostCentreList" collection containing the "weight" property', function () {
+        expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
         expect(ctrl.CostCentreList[0].weight).not.toBeNull();
         expect(ctrl.CostCentreList[1].weight).not.toBeNull();
         expect(ctrl.CostCentreList[2].weight).not.toBeNull();
       });
+
+      it('builds the "hrjc_level_type" collection containing the "value" property', function () {
+        expect(Object.keys(ctrl.LevelsData).length).toBe(1);
+        expect(ctrl.LevelsData[Object.keys(ctrl.LevelsData)[0]]['value']).toBe('Senior Manager');
```
igorpavlov: Let's do not hardcode 'Senior Manager' in two places, we already have it in the mock, just compare to a value from the mock. Same for other test suites.

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -60,88 +60,41 @@ define([
...
+      it('builds the "CostCentreList" collection containing the "weight" property', function () {
+        expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
         expect(ctrl.CostCentreList[0].weight).not.toBeNull();
         expect(ctrl.CostCentreList[1].weight).not.toBeNull();
         expect(ctrl.CostCentreList[2].weight).not.toBeNull();
       });
+
+      it('builds the "hrjc_level_type" collection containing the "value" property', function () {
+        expect(Object.keys(ctrl.LevelsData).length).toBe(1);
+        var levelValue = Mock.option_values.values[0].value;
```
igorpavlov: The variable name `Mock` is not really descriptive, may we change it to `JobRoleDataMock` or smth like that?

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -60,88 +60,41 @@ define([
...
+        expect(ctrl.LevelsData[Object.keys(ctrl.LevelsData)[0]]['value']).toBe(levelValue);
+      });
+
+      it('builds the "hrjc_department" collection containing the "value" property', function () {
+        expect(Object.keys(ctrl.DepartmentsData).length).toBe(1);
+        var departmentValue = Mock.option_values.values[4].value;
```
igorpavlov: Please do not put vars in `it` block, we want to keep it as clean and readable as possible.

```js
describe('getOptionValues', function () {
  var expectedDepartmentValue, expectedLocationValue ...;

  beforeEach(function () {
    departmentValue = ...;
  });

  it(..., function () {
    expect(...).toBe(expectedDepartmentValue);
  });
});
```

com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -59,89 +59,44 @@ define([
...
-            }
-          ],
-          'optionGroupData': {
-            'cost_centres': '124'
-          }
-        }));
+        spyOn(jobRoleService, 'getOptionValues').and.returnValue($q.resolve(Mock.option_values));
         initController();
         $rootScope.$digest();
+        expectedLevelValue = Mock.option_values.values[0].value;
```
igorpavlov: Please move assignments to the top of the function, then add an empty line before `spyOn(...` to separate assignments block and func calls block.

PR #2550, opened 2018-03-26
PCHR-3451: Default Existing Work Pattern to As per contract

ajesamson:
## Overview
Since new work pattern are being created, the need arose to update existing records to the new default value of `As per contract`. This PR caters for that.

## Before
before

## After
after

## Technical Details
All existing records are fetched and updated to the new default value using chained api call.

```php
civicrm_api3('ContactWorkPattern', 'get', [
      'return' => ['id'],
      'api.ContactWorkPattern.update' => [
        'change_reason' => 1,
        'id' => "\$value.id"
      ],
]);
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
@@ -1,0 +1,22 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1022 {
+  
+  /**
+   * Default existing work pattern to As per contract (new default)
+   *
+   * @return bool
+   */
+  public function upgrade_1022()
+  {
+    civicrm_api3('ContactWorkPattern', 'get', [
+      'return' => ['id'],
+      'api.ContactWorkPattern.update' => [
+        'change_reason' => 1,
+        'id' => "\$value.id"
+      ],
+    ]);
+    
+    return TRUE;
+  }
+}
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
@@ -9,13 +9,15 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1022 {
    */
   public function upgrade_1022()
   {
-    civicrm_api3('ContactWorkPattern', 'get', [
-      'return' => ['id'],
-      'api.ContactWorkPattern.update' => [
-        'change_reason' => 1,
-        'id' => "\$value.id"
-      ],
-    ]);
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * Default existing work pattern to As per contract (new default)
+   *
+   * @return bool
+   */
+  public function upgrade_1022()
+  {
+    civicrm_api3('ContactWorkPattern', 'get', [
+      'return' => ['id'],
+      'api.ContactWorkPattern.update' => [
```
mickadoo: I think `update` is deprecated, in the API explorer it has a ~strikethrough~ effect. Can you test if this works with `create` and change it if it does?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
```diff
@@ -0,0 +1,22 @@
...
+        'change_reason' => 1,
```
mickadoo: Instead of hardcoding the change reason can you fetch the `value` of the OptionValue based on machine name and `option_group_id`. It's best not to depend on `value` as we've had problems before where it has been different depending on the system.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
```diff
@@ -9,13 +9,15 @@
...
-    civicrm_api3('ContactWorkPattern', 'get', [
-      'return' => ['id'],
-      'api.ContactWorkPattern.update' => [
-        'change_reason' => 1,
-        'id' => "\$value.id"
-      ],
-    ]);
+    $query = 'UPDATE civicrm_hrleaveandabsences_contact_work_pattern set change_reason =
+      (
+        SELECT cov.value FROM civicrm_option_group cog LEFT JOIN civicrm_option_value cov
```
mickadoo: It doesn't make much difference to the query result, but it would be clearer to exchange the `FROM` table with the join table. You're selecting FROM `civicrm_option_value` and joining to `civicrm_option_group` to filter the result.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1022.php
```diff
@@ -9,13 +9,15 @@
...
   public function upgrade_1022()
   {
-    civicrm_api3('ContactWorkPattern', 'get', [
-      'return' => ['id'],
-      'api.ContactWorkPattern.update' => [
-        'change_reason' => 1,
-        'id' => "\$value.id"
-      ],
-    ]);
+    $query = 'UPDATE civicrm_hrleaveandabsences_contact_work_pattern set change_reason =
```
mickadoo: Just for consistency `set` should be uppercase too

PR #2551, opened 2018-03-26
PCHR-3481: Bump version to 1.7.5

davialexandre:


hrident/info.xml
```diff
@@ -10,11 +10,10 @@
   </maintainer>
   <releaseDate>2017-08-22</releaseDate>
```
ajesamson: Is this release date correct or should it be `2018-03-26` since the `develStage` has been updated?

hrvisa/info.xml
```diff
@@ -10,11 +10,10 @@
   </maintainer>
   <releaseDate>2017-08-22</releaseDate>
```
ajesamson: Hope this `releaseDate` date is correct as it was not updated?

uk.co.compucorp.civicrm.hrcontactactionsmenu/info.xml
```diff
@@ -11,7 +11,7 @@
   </maintainer>
   <releaseDate>2017-12-20</releaseDate>
```
ajesamson: Do we need to update releaseDate?

org.civicrm.hremergency/info.xml
```diff
@@ -2,19 +2,19 @@
 <extension key="org.civicrm.hremergency" type="module">
   <file>hremergency</file>
   <name>Emergency Contacts (CiviHR)</name>
-  <description></description>
+  <description>Emergency Contacts extensions for CiviHR</description>
```
ajesamson: I think this should be `<description>Emergency Contacts extension for CiviHR</description>`.
`extensions` => `extension`

hrident/info.xml
davialexandre: The date is correct. This extension has not been updated in a while. In fact, it will soon be deleted, as it's not installed anymore on new sites.

hrvisa/info.xml
davialexandre: same thing as hrident

hrident/info.xml
ajesamson: Great.

hrvisa/info.xml
ajesamson: Alright @davialexandre

PR #2536, opened 2018-03-19
PCHR-3426: Split request record button into two buttons

igorpavlov:
## Overview

This PR amends the "Record new absence" button dropdown with 3 options (including Overtime) button into "Record new absence" button dropdown with 2 options (not more Overtime) + "Submit Overtime" button. 

Please also see:
https://github.com/compucorp/org.civicrm.shoreditch/pull/100
https://github.com/compucorp/civihr-employee-portal-theme/pull/328

### Admin

image

### SSP

1 ssp

### Admin

2 admin

### SSP

Desktop:
2 ssp

Mobile:
2 ssp m

### 1. Remove TOIL (Overtime) button from the dropdown

```js
vm.leaveRequestOptions = [
  { type: 'leave', icon: 'briefcase', label: 'Leave' },
  { type: 'sickness', icon: 'stethoscope', label: 'Sickness' },
  { type: 'toil', icon: 'calendar-plus-o', label: 'Overtime' } // got rid of it
];
```

### Create a SASS partial for the leave request record buttons component

```css
leave-request-record-actions {
  display: block; // allows to nicely position the component in different layouts

  .btn-group {
    float: left; // instead on 'inline-block' that leaves unexpected gaps between buttons

  .btn-group:first-child { // this layout considers only 2 buttons. 3 buttons would be almost impossible to position in a row in a mobile layout, so we do not consider more than 2 buttons here
    margin-right: $crm-gap-medium; // an 8px gap

    .dropdown-menu { // align the dropdown to the right of the button
      left: auto;
      right: 0;

@media screen and (max-width: $screen-xs-max) { // mobile
  leave-request-record-actions {
    .btn-group {
      width: calc((100% - #{$crm-gap-medium}) / 2); // make buttons half width

    .btn {
      width: 100%;
      padding-right: 0; // padding should be discarded, otherwise there is no space for text
      padding-left: 0;

    .btn-group:first-child .dropdown-menu {
        left: 0; // align dropword left to the button
        right: auto;

```
### Amend HTML template

```html
<div class="clearfix"> <!-- to float correctly -->
  <div class="btn-group" uib-dropdown>
    <button type="button" class="btn btn-primary btn-lg dropdown-toggle" uib-dropdown-toggle>
      ...
    <ul class="dropdown-menu" role="menu" uib-dropdown-menu>
      ...
  </div>
  <div class="btn-group"> <!-- for consistency -->
    <button
      class="leave-request-record-actions__new-toil
        btn btn-secondary btn-lg"
[...]

PR #2552, opened 2018-03-26
PCHR-3494: Allow to edit TOIL in days options from Leave Request Modal

igorpavlov:
## Overview

This PR adds a wrench icon next to TOIL accrual options to allow to edit the list of these options.

## Before

_There is no possibility to change options from the modal._

## After

_By clicking on the wrench icon you can edit the options from the modal._

### 1. Add a CRM functionality wrapper as an Angular service

CRM is a global variable that contains, for example, the `loadForm` method, which is needed to open the options editor in the CRM modal. In order to not use the global CRM variable directly in the Angular component, the wrapper service was created.

```js
(function (CRM) {
  // ... 
    services.factory('crmAngService', function () {
      return {
        loadForm: loadForm // the list of function may be populated in the future

    // ...
    function loadForm (url, options) {
      return CRM.loadForm(url, options); // it is a simple wrapper to avoid using CRM var directly
    // ...
}(CRM)); // The only place the global CRM variable is used
```
### 2. Allow not to cache option groups

In order to update option groups on the fly in the Leave Request Modal we need to allow not to cache the results.

_Option group model_
```js
valuesOf: function (names, cache) {
  return optionGroupAPI.valuesOf(names, cache);
```

_Option group API service_
```js
valuesOf: function (groupNames, params, cache) {
  // ...
  return this.sendGET('OptionValue', 'get', _.assign({
    // ...
  }, params), cache)
    .then(// ...
```

### 3. Add a wrench icon and call the editor on click

_Modal TOIL subcomponent_
```js
function openToilInDaysAccrualOptionsEditor () {
  crmAngService.loadForm('/civicrm/admin/options/hrleaveandabsences_toil_amounts?reset=1')
    .on('crmFormSuccess', loadToilAmounts);
```

ℹ️Please note, this wrench icon is not shown in SSP and only visible to Admin.

_Modal TOIL template partial_
```html
<div
  ng-if="detailsTab.selectedAbsenceType.calculation_unit_name === 'days'"
  ng-init="showTOILAccrualsOptionEditorIcon = $root.section === 'admin-dashboard' || $root.section === 'absence-tab'">
  <div ng-class="{ 'col-xs-11': showTOILAccrualsOptionEditorIcon, 'col-xs-12': !showTOILAccrualsOptionEditorIcon }">
    <!-- Options Selector here -->
  <div ng-if="showTOILAccrualsOptionEditorIcon" class="col-xs-1 control-label">
    <a class="pointer">
      <i class="crm-i fa-wrench" ng-click="detailsTab.openToilInDaysAccrualOptionsEditor()"></i>
```

## Comments

- Job Role and Contact spec files were renamed to support TDD and their JS style was fixed.

------

✅Manual Tests - passed
✅Jasmine Tests - passed
⚠️Backstop tests - existing passed, but new suite cannot be created due to https://github.com/civicrm/civihr/pull/2536 since buttons have been changed
⏹PHP Unit Tests - not needed


org.civicrm.reqangular/test/services/api/option-group.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        it('passes extra parameters to sendGET', function () {
+          expect(OptionGroupAPI.sendGET.calls.mostRecent().args[2]).toEqual(
+            jasmine.objectContaining(params)
+          );
+        });
+      });
+
+      describe('when no caching is needed', function () {
```
reneolivo: is caching the default behaviour? Can we add a test to check if cache is done by default?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-toil.controller.js
```diff
@@ -412,6 +416,16 @@ define([
...
     }
 
+    /**
+     * Opens the CRM modal that allows to edit TOIL in days amounts options
+     * and reloads these options in the Leave Request Modal
+     * if they are changed via the CRM modal
+     */
+    function openToilInDaysAccrualOptionsEditor () {
+      crmAngService.loadForm('/civicrm/admin/options/hrleaveandabsences_toil_amounts?reset=1')
+        .on('crmFormSuccess', loadToilAmounts);
```
reneolivo: what about passing a cache parameter so we can still benefit from cached toil amounts?

```js
function openToilInDaysAccrualOptionsEditor () {
  crmAngService.loadForm('/civicrm/admin/options/hrleaveandabsences_toil_amounts?reset=1')
    .on('crmFormSuccess', function () {
      // false means no cache:
      loadToilAmounts(false);
    });
}

function loadToilAmounts (cache) {
  // ...
}
```

org.civicrm.reqangular/test/services/api/option-group.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+    ]));
+
+    beforeEach(inject(function (_$q_) {
+      $q = _$q_;
+      methods = Object.keys(OptionGroupAPI);
+    }));
+
+    it('has expected interface', function () {
+      expect(methods).toContain('valuesOf');
```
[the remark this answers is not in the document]
igorpavlov: Not because of TDD, there were no TDD in this case since the spec was just missing. I just copied this block from somewhere. Now I look at it and it does not make sense since this exists:

```js
beforeEach(function () {
  OptionGroupAPI.valuesOf(optionGroupName);
```

which already checks if it exists and it is a function.

org.civicrm.reqangular/test/services/api/option-group.spec.js
igorpavlov: Though it was actually tested already, but it was not obvious. So I think I improved it a bit now.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-toil.controller.js
igorpavlov: Done.

org.civicrm.reqangular/test/services/crm-ang.service.spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+(function (CRM) {
+  define([
+    'common/angular',
+    'common/angularMocks',
+    'common/services/crm-ang.service'
+  ], function () {
+    'use strict';
+
+    describe('crmAng', function () {
```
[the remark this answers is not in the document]
igorpavlov: Done.

igorpavlov:
@reneolivo I also fixed a test that for some reason tested a call to option group API, while actually option group **model** should be tested there.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-toil.controller.js
igorpavlov: Not done! If we do that, then if we change options, then close modal, then reopen modal - it will should previously cached results. We need to somehow **flush** the cache when options are changed via wrench.

cc @AkA84

AkA84: What if, when you close the modal:
* You flush the $http cache (https://stackoverflow.com/questions/31168708/invalidating-all-http-caches-in-angular, first thing i found)
* Reload the options

Of course there is a bit of difference between making it work and making it work and having it implemented in a clean way (for example I wouldn't want a controller to flush the cached data related to an api service), but let's see first if it works

igorpavlov: @AkA84 When you say "Modal" which modal do you refer to: editor or leave request modal? I think that it would be actually better to have a following logic:

1) If you load something with cache enabled - leave it as it is now
2) If you load something with cache disabled - flush the cache (new functionality)

cc @reneolivo

org.civicrm.reqangular/src/common/services/api.js
```diff
@@ -42,6 +42,20 @@ define([
...
     }
 
+    /**
+     * Constructs a GET query string exactly the same way as $http service does
+     *
+     * @param  {Object} json that is passed as "params" to $http
+     * @return {String} query string
+     */
+    function jsonToQueryString (json) {
+      return '?' + Object.keys(json).sort().map(function (key) {
```
reneolivo: Do you think you could use jQuery's param method here?
http://api.jquery.com/jquery.param/

```js
$.param({a: 1, b: 2, c: "100"}); // a=1&b=2&c=100
$.param({a: [1, 2, 3]}); // a%5B%5D=1&a%5B%5D=2&a%5B%5D=3
```

igorpavlov: I was thinking about this, but there are 2 reasons for not doing this:
1) jQuery is a global thing
2) jQuery.param would not sort query params, while $http does, we would need to also use smth like `_(json).toPairs().sortBy(0).fromPairs().value()`.

I would honestly leave this function as it is now.

org.civicrm.reqangular/src/common/models/option-group.js
```diff
@@ -1,28 +1,31 @@
...
-                }
-            });
+        /**
+         * Returns the values of a sigle or multiple option groups
+         *
+         * @param  {String|Array} names
+         *   Bases on the type of the parameter, the method will return
+         *   either an array of values (string) or an object (array)
+         * @param  {Object} params optional parameters for the query
+         * @param  {Boolean} cache optional parameter to cache the query or not
```
AkA84: you can mark optional parameters with `[]` (http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

org.civicrm.reqangular/src/common/services/api.js
AkA84: Doesn't [`$httpParamSerializer`](https://docs.angularjs.org/api/ng/service/$httpParamSerializer) do the same

AkA84: if you really need this function (and again please triple check, i have the feeling we are reinventing the wheel here), please move it toward the bottom, follow the style guide

org.civicrm.reqangular/src/common/services/api/option-group.js
```diff
@@ -52,16 +52,17 @@ define([
...
-       * @param {object} params optional parameters for the query
+       * @param  {Object} params optional parameters for the query
+       * @param  {Boolean} cache optional parameter to cache the query or not
        * @return {Promise}
        *   Resolves to an array with the values (if `groupNames` is a string)
        *   or an object with keys as the group names and values as the
        *   array of their option values (if `groupNames` is an array)
        */
-      valuesOf: function (groupNames, params) {
+      valuesOf: function (groupNames, params, cache) {
```
AkA84: remove the `params` (see above), they shouldn't have been added in the first place (i checked in our codebase and we are not using it anywhere, as we shouldn't)

org.civicrm.reqangular/src/common/services/crm-ang.service.js
```diff
@@ -0,0 +1,29 @@
...
+    services.factory('crmAngService', function () {
+      return {
+        loadForm: loadForm
+      };
+    });
+
+    /**
+     * Opens a CRM form with a provided url and options
+     * This is simply a wrapper to CRM.loadForm
+     * Reference: https://docs.civicrm.org/dev/en/latest/framework/ajax
```
AkA84: http://usejsdoc.org/tags-see.html

org.civicrm.reqangular/src/common/services/crm-ang.service.js
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @param  {String} url
+     * @param  {Object} options
+     * @return {jQuery}
```
AkA84: looks like it's returning the jQuery library

org.civicrm.reqangular/src/common/services/crm-ang.service.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-env amd */
+
+(function (CRM) {
+  define([
+    'common/lodash',
+    'common/modules/services'
+  ], function (_, services) {
+    'use strict';
+
+    services.factory('crmAngService', function () {
```
AkA84: do we need the `Service` suffix?

org.civicrm.reqangular/test/models/option-group.spec.js
```diff
@@ -35,12 +35,56 @@ define([
...
+      describe('when just a single option group name is passed', function () {
+        beforeEach(function () {
+          OptionGroup.valuesOf(optionGroupName);
+        });
+
+        it('calls the correct API method', function () {
+          expect(optionGroupAPI.valuesOf.calls.mostRecent().args[0]).toBe(optionGroupName);
+        });
+
+        it('caches requests by default', function () {
```
AkA84: have a 
```js
describe('caching', function () {
});
```
block where you test both the default caching behaviour and the usage of the `cache` param

org.civicrm.reqangular/test/services/api.spec.js
```diff
@@ -99,26 +103,84 @@ define([
...
+          it('does not flush cache before HTTP call', function () {
+            expect($cacheFactory.get('$http').remove).not.toHaveBeenCalled();
+          });
+
+          it('caches data by default', function () {
+            expect($cacheFactory.get('$http').info().size).toBe(1);
+          });
+        });
+
+        describe('when "returnCachedData" parameter passed as TRUE', function () {
```
AkA84: `true`, when don't capitalize true/false/null like we do in php

AkA84: put `$cacheFactory.get('$http').info().size` in a well-named variable, so the expectation reads better and it's more expressive

org.civicrm.reqangular/test/services/api.spec.js
```diff
@@ -99,26 +103,84 @@ define([
...
+
+        describe('when "returnCachedData" parameter passed as FALSE', function () {
+          beforeEach(function () {
+            expectAndSendGET(sampleResult, sampleParams, false);
+            $rootScope.$digest();
+          });
+
+          it('flushes cache before HTTP call', function () {
+            expect($cacheFactory.get('$http').remove).toHaveBeenCalledWith(
+              '/civicrm/ajax/rest?action=action&entity=entity&json=%7B%22options%22:%7B%22limit%22:0%7D%7D&sequential=1');
```
AkA84: it's not clear at all where you got this string. Either you put in a well-named variable or maybe you can have it returned by `expectAndSendGET()`. Anyway this needs to be improved

org.civicrm.reqangular/test/services/api.spec.js
```diff
@@ -99,26 +103,84 @@ define([
...
       }
     });
 
-    describe('sendPOST', function () {
+    describe('sendPOST()', function () {
       var promise;
 
-      afterEach(function () { $httpBackend.flush(); });
+      afterEach(function () {
+        $httpBackend.flush();
```
AkA84: we just discussed that we shouldn't do this and use a different format?

org.civicrm.reqangular/test/services/api/option-group.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  'common/models/option-group',
+  'common/mocks/services/api/option-group-mock'
+], function () {
+  'use strict';
+
+  describe('api.option-group', function () {
+    var $q, OptionGroupAPI;
+
+    beforeEach(function () {
+      module('common.models', 'common.mocks');
```
AkA84:
```js
beforeEach(module('common.models', 'common.mocks'));
```

org.civicrm.reqangular/test/services/api/option-group.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    beforeEach(inject([
+      'api.optionGroup',
+      function (_optionGroupAPI_) {
+        OptionGroupAPI = _optionGroupAPI_;
+      }
+    ]));
+
+    beforeEach(inject(function (_$q_) {
```
AkA84: you can do this in the same `beforeEach` above, no need to have two of them

org.civicrm.reqangular/test/services/crm-ang.service.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      }));
+
+      describe('loadForm()', function () {
+        var result;
+        var url = '/some-url';
+        var options = { optionKey: 'optionValue' };
+        var originalResult = 'jQuery object';
+
+        beforeEach(function () {
+          CRM.loadForm = function () {};
```
AkA84: Can't you use [`.createSpy()`](https://jasmine.github.io/2.0/introduction.html#section-Spies:_%3Ccode%3EcreateSpy%3C/code%3E)?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-toil.controller.js
```diff
@@ -3,13 +3,16 @@
 define([
   'common/lodash',
   'common/moment',
-  'leave-absences/shared/modules/controllers'
+  'leave-absences/shared/modules/controllers',
+  'common/services/crm-ang.service'
```
AkA84: the actual services/controllers/directives should be listed in the main modules' files

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
```diff
@@ -47,20 +50,31 @@ define([
...
+      OptionGroup = _OptionGroup_;
       TOILRequestInstance = _TOILRequestInstance_;
 
       spyOn($log, 'debug');
       spyOn(AbsenceTypeAPI, 'calculateToilExpiryDate').and.callThrough();
       spyOn(AbsenceType, 'canExpire').and.callThrough();
       spyOn(AbsenceType, 'calculateToilExpiryDate').and.callThrough();
+      spyOn(OptionGroup, 'valuesOf').and.callThrough();
+
+      crmAngService.loadForm = function () {
```
AkA84: I think it's better to put this inside the "'when user opens TOIL accrual options group editor" describe block

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
```diff
@@ -302,6 +320,34 @@ define([
...
+        describe('when TOIL accruals are edited', function () {
+          beforeEach(function () {
+            $rootScope.$digest();
+          });
+
+          it('reloads TOIL accrual options', function () {
+            expect(Object.keys(controller.toilAmounts).length).toBeGreaterThan(0);
+          });
+
+          it('does not cache TOIL accrual options', function () {
```
AkA84: > "does not cache"

This is misleading, because based on your changes the data is always cached, we're just flushing the cache before caching the data again

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
```diff
@@ -302,6 +320,34 @@ define([
...
+            expect(OptionGroup.valuesOf.calls.mostRecent().args[2]).toBe(false);
```
AkA84: `OptionGroup.valuesOf.calls.mostRecent().args[2]`
put this in a named variable, it makes the test read better

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
```diff
@@ -39,24 +39,34 @@
...
-            <option value="">- select -</option>
-          </select>
-          <span class="crm_custom-select__arrow"></span>
+      <div
+        ng-if="detailsTab.selectedAbsenceType.calculation_unit_name === 'days'"
+        ng-init="showTOILAccrualsOptionEditorIcon = $root.section === 'admin-dashboard' || $root.section === 'absence-tab'">
+        <div ng-class="{ 'col-xs-11': showTOILAccrualsOptionEditorIcon, 'col-xs-12': !showTOILAccrualsOptionEditorIcon }">
+          <div
+            class="crm_custom-select crm_custom-select--full"
+            ng-if="detailsTab.selectedAbsenceType.calculation_unit_name === 'days'">
```
AkA84: but we already know that this is true, otherwise the parent `<div>` would not even be displayed

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
```diff
@@ -39,24 +39,34 @@
...
-            ng-disabled="detailsTab.isMode('view')"
-            ng-options="amount.value as amount.label for amount in detailsTab.toilAmounts"
-            ng-model="detailsTab.request.toil_to_accrue"
-            ng-change="detailsTab.performBalanceChangeCalculation()">
-            <option value="">- select -</option>
-          </select>
-          <span class="crm_custom-select__arrow"></span>
+      <div
+        ng-if="detailsTab.selectedAbsenceType.calculation_unit_name === 'days'"
+        ng-init="showTOILAccrualsOptionEditorIcon = $root.section === 'admin-dashboard' || $root.section === 'absence-tab'">
```
AkA84: too much logic in the view, please put this in the controller

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
```diff
@@ -39,24 +39,34 @@
...
+            <span class="crm_custom-select__arrow"></span>
+          </div>
         </div>
+        <div ng-if="showTOILAccrualsOptionEditorIcon" class="col-xs-1 control-label">
+          <a class="pointer">
+            <i class="crm-i fa-wrench" ng-click="detailsTab.openToilInDaysAccrualOptionsEditor()"></i>
+          </a>
+        </div>
+      </div>
+      <div class="col-xs-12" ng-if="detailsTab.selectedAbsenceType.calculation_unit_name === 'hours'">
```
AkA84: seems you have this condition multiple times in the view, I think you should put it in the controller, something like
```html
<div ng-if="isCalculationIn('hours')">
```

org.civicrm.reqangular/src/common/models/option-group.js
```diff
@@ -1,28 +1,31 @@
...
-                    return optionGroupAPI.valuesOf(names);
-                }
-            });
+        /**
+         * Returns the values of a sigle or multiple option groups
+         *
+         * @param  {String|Array} names
+         *   Bases on the type of the parameter, the method will return
+         *   either an array of values (string) or an object (array)
+         * @param  {Object} params optional parameters for the query
```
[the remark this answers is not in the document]
igorpavlov: Done.

org.civicrm.reqangular/src/common/models/option-group.js
igorpavlov: Done.

org.civicrm.reqangular/src/common/services/api/option-group.js
igorpavlov: Done.

org.civicrm.reqangular/src/common/services/crm-ang.service.js
igorpavlov: Done.

org.civicrm.reqangular/src/common/services/crm-ang.service.js
igorpavlov: Done.

org.civicrm.reqangular/src/common/services/crm-ang.service.js
igorpavlov: We did that in https://github.com/civicrm/civihr/blob/staging/org.civicrm.reqangular/src/common/services/notification.service.js for example

org.civicrm.reqangular/test/services/api.spec.js
igorpavlov: Done.

igorpavlov: Done.

org.civicrm.reqangular/test/services/api.spec.js
igorpavlov: Done and added a comment

org.civicrm.reqangular/test/services/api.spec.js
igorpavlov: Done. Moreover, these ones were not working at all (passing when they should not pass):

```js
$httpBackend.expectPOST('/civicrm/ajax/rest', $httpParamSerializer({
              json: { options: { limit: limit } },
              sequential: 1,
              action: action,
              entity: entity
            }));
```

Now they are fixed.

org.civicrm.reqangular/test/services/api/option-group.spec.js
igorpavlov: Done.

org.civicrm.reqangular/test/services/api/option-group.spec.js
igorpavlov: The one above injects by names, I don't think that will work.

org.civicrm.reqangular/test/services/crm-ang.service.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-toil.controller.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
igorpavlov: Good catch. Done.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-request-popup/details-tab/toil.html
igorpavlov: Done.

PR #2559, opened 2018-04-03
PCHR-3509: Use Role Machine Names Instead of Role Names

mickadoo:
## Overview

When sending stats to civihr.org we send most recent login data. This data uses the role name as an index. When role name is changed, as they were in https://github.com/compucorp/civihr-employee-portal/pull/435, the server does not know which roles are the "base" roles (civihr_admin, civihr_staff, civihr_manager & civihr_local_admin). To avoid problems like this when role name is changes we will use role machine name.

## Before

The stats request used role names as an index for most recent login data

```json
{
  "siteUrl": "https://civihr-staging.civihrhosting.co.uk/",
  "siteName": "CiviHR Staging site",
  "generationDate": "2018-03-28T13:05:17+0000",
  "lastLoginByRole": {
    "administrator": "2018-03-28T13:03:44+0000",
    "HR Admin": "2018-03-28T12:55:13+0000",
    "Staff": "2018-03-28T06:46:23+0000",
    "Manager": "2018-02-02T12:32:48+0000",
    "Regional HR Admin": "2018-02-02T12:32:48+0000",
    "Custom Role": "2018-02-02T12:32:48+0000"
  }
}
```

## After

```json
{
  "siteUrl": "https://civihr-staging.civihrhosting.co.uk/",
  "siteName": "CiviHR Staging site",
  "generationDate": "2018-03-28T13:05:17+0000",
  "lastLoginByRole": {
    "administrator": "2018-03-28T13:03:44+0000",
    "civihr_admin": "2018-03-28T12:55:13+0000",
    "civihr_staff": "2018-03-28T06:46:23+0000",
    "civihr_manager": "2018-02-02T12:32:48+0000",
    "civihr_local_admin": "2018-02-02T12:32:48+0000",
    "Custom Role": "2018-02-02T12:32:48+0000"
  }
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/Role/DrupalRoleService.php
@@ -5,7 +5,7 @@ use CRM_HRCore_CMSData_Role_RoleServiceInterface as RoleServiceInterface;
 /**
  * Implementation of RoleServiceInterface to interact with a Drupal 7 system
  */
-class CRM_HRCore_CMSData_Role_DrupalRoleService implements RoleServiceInterface{
+class CRM_HRCore_CMSData_Role_DrupalRoleService implements RoleServiceInterface {
 
   /**
    * @inheritdoc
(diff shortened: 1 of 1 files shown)


tunbola:
@mickadoo , Looks good. Do we need to update tests based on this change?

mickadoo:
@tunbola all tests in HRCore are passing. Right now for code that uses Drupal function we use the stubs in the drupal_function_mocks.php file that Davi added. 

I don't think mocking `db_query` is an option because this gets very complicated. I spoke with Davi to suggest allowing some tests to bootstrap Drupal and use Drupal functions but he didn't like that idea, and decided that it's ok not to have tests for this.

PR #2561, opened 2018-04-05
PCHR-2576: Fix Issue with Admin Not Being able to Cancel TOIL with Past Dates

tunbola:
## Overview
When a TOIL was requested and the TOIL days are now past and the Absence Type does not allow past accrual, neither Admin nor Staff is able to cancel such request. An error message `You may only request TOIL for overtime to be worked in the future` is displayed on trying to cancel such a TOIL request.
We want the only the Admin and Leave Managers to be able to cancel such TOIL requests. This PR makes changes to make this happen.

## Before
- The Administrator and Leave manager can not cancel a TOIL request with past dates if the Absence type does not allow past accrual.
- The staff can not cancel  a TOIL request with past dates when absence type does not allow past accrual.

## After
- The Administrator and Leave manager can now cancel a TOIL request with past dates even if the Absence type does not allow past accrual.
- The staff can still not cancel  a TOIL request with past dates when absence type does not allow past accrual according to the Ticket acceptance criteria.
- Note that the implementation is still subject to the existing validation for the `allow_request_cancelation` property of the Absence type. i.e if the Absence type does not allow request cancellation, the Leave request validation will still not allow such request to be cancelled by Admin or Leave Manager even though they can cancel TOIL with past dates.

A new method was added to the Leave Request rights Service and used in a validation method in the LeaveRequest Service. 
The validation was done at the LeaveRequest Service rather than at the LeaveRequest BAO since the LeaveRequest BAO validations are not aware of current user and user roles.

```php
  public function canCancelToilWithPastDates($contactID, $absenceTypeID) {
    $absenceType = AbsenceType::findById($absenceTypeID);

    if(!$absenceType->allow_accrue_in_the_past) {
      return $this->currentUserIsManagerOrAdmin($contactID);
    }

    return TRUE;
  }
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -354,7 +354,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    */
   public static function isTOILWithPastDates($params) {
     if($params['request_type'] !== LeaveRequest::REQUEST_TYPE_TOIL) {
-      return false;
+      return FALSE;
     }
 
     $fromDate = new DateTime($params['from_date']);
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -336,8 +336,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $toDate = new DateTime($params['to_date']);
     $todayDate = new DateTime('today');
     $leaveDatesHasPastDates = $fromDate < $todayDate || $toDate < $todayDate;
+    $isACancellationRequest = in_array($params['status_id'], LeaveRequest::getCancelledStatuses());
 
-    if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past) {
+    if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past && !$isACancellationRequest) {
       throw new InvalidLeaveRequestException(
         'You may only request TOIL for overtime to be worked in the future. Please modify the date of this request',
         'leave_request_toil_cannot_be_requested_for_past_days',
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -332,10 +332,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @throws \CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException
    */
   private static function validateTOILPastDays($params, $absenceType) {
-    $fromDate = new DateTime($params['from_date']);
-    $toDate = new DateTime($params['to_date']);
-    $todayDate = new DateTime('today');
-    $leaveDatesHasPastDates = $fromDate < $todayDate || $toDate < $todayDate;
+    $leaveDatesHasPastDates = self::isTOILWithPastDates($params);
     $isACancellationRequest = in_array($params['status_id'], LeaveRequest::getCancelledStatuses());
 
     if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past && !$isACancellationRequest) {
@@ -345,6 +342,26 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
         'from_date'
       );
     }
+  }
+
+  /**
+   * Checks whether the leave request is of type TOIL and has dates in the
+   * past.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  public static function isTOILWithPastDates($params) {
+    if($params['request_type'] !== LeaveRequest::REQUEST_TYPE_TOIL) {
+      return false;
+    }
+
+    $fromDate = new DateTime($params['from_date']);
+    $toDate = new DateTime($params['to_date']);
+    $todayDate = new DateTime('today');
+
+    return $fromDate < $todayDate || $toDate < $todayDate;
   }
 
   /**
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
@@ -117,6 +117,14 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequest {
       );
     }
 
+    if ($this->isTOILWithPastDates($params) && $this->isCancelledStatus($params)) {
+      if (!$this->canCanCancelTOILWithPastDates($params)) {
+        throw new RuntimeException(
+          'You may only cancel TOIL with dates in the future.'
+        );
+      }
+    }
+
     return $this->createRequestWithBalanceChanges($params);
   }
 
@@ -374,5 +382,47 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequest {
     $updateBalanceChange = !empty($params['change_balance']);
 
     return ($leaveRequestDatesNotChanged || $toilDatesAndToAccrueNotChanged) && !$updateBalanceChange;
+  }
+
+  /**
+   * Checks If the current user can cancel a TOIL request type that has past dates.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function canCanCancelTOILWithPastDates($params) {
+    return $this->leaveRequestRightsService->canCancelToilWithPastDates($params['contact_id'], $params['type_id']);
+  }
+
+  /**
+   * Checks whether the leave request is of type TOIL and has dates in the
+   * past.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function isTOILWithPastDates($params) {
+    if($params['request_type'] !== LeaveRequest::REQUEST_TYPE_TOIL) {
+      return false;
+    }
+
+    $fromDate = new DateTime($params['from_date']);
+    $toDate = new DateTime($params['to_date']);
+    $todayDate = new DateTime('today');
+
+    return $fromDate < $todayDate || $toDate < $todayDate;
+  }
+
+  /**
+   * Checks whether the Leave is to be updated to cancelled.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function isCancelledStatus($params) {
+    return in_array($params['status_id'], LeaveRequest::getCancelledStatuses());
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
@@ -117,12 +117,14 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequest {
       );
     }
 
-    if ($this->isTOILWithPastDates($params) && $this->isCancelledStatus($params)) {
-      if (!$this->canCanCancelTOILWithPastDates($params)) {
-        throw new RuntimeException(
-          'You may only cancel TOIL with dates in the future.'
-        );
-      }
+    $isTOILWithPastDates = LeaveRequest::isTOILWithPastDates($params);
+    $isCancelledStatus = $this->isCancelledStatus($params);
+    $canCanCancelTOILWithPastDates = $this->canCanCancelTOILWithPastDates($params);
+
+    if ($isTOILWithPastDates && $isCancelledStatus && !$canCanCancelTOILWithPastDates) {
+      throw new RuntimeException(
+        'You may only cancel TOIL with dates in the future.'
+      );
     }
 
     return $this->createRequestWithBalanceChanges($params);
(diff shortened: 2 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -336,8 +336,9 @@ private static function validateTOILPastDays($params, $absenceType) {
     $toDate = new DateTime($params['to_date']);
     $todayDate = new DateTime('today');
     $leaveDatesHasPastDates = $fromDate < $todayDate || $toDate < $todayDate;
+    $isACancellationRequest = in_array($params['status_id'], LeaveRequest::getCancelledStatuses());
```
mickadoo: If you are updating a leave request without changing the status will this mean `status_id` won't be set? If it isn't then it would trigger a warning

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -117,6 +117,14 @@ protected function update($params) {
       );
     }
 
+    if ($this->isTOILWithPastDates($params) && $this->isCancelledStatus($params)) {
+      if (!$this->canCanCancelTOILWithPastDates($params)) {
+        throw new RuntimeException(
```
mickadoo: Just curious why the use of `RuntimeException`. I'd be happy to see us adopt a more structured approach to exceptions, I usually just use `\Exception` now, but would be good to have some agreement on best practice.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -375,4 +383,46 @@ private function skipBalanceChangeUpdate($params) {
...
+
+  /**
+   * Checks whether the leave request is of type TOIL and has dates in the
+   * past.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function isTOILWithPastDates($params) {
```
mickadoo: This looks like duplication of code in the BAO. Can you extract it to another class and use it in both places?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -375,4 +383,46 @@ private function skipBalanceChangeUpdate($params) {
...
   }
+
+  /**
+   * Checks If the current user can cancel a TOIL request type that has past dates.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  private function canCanCancelTOILWithPastDates($params) {
```
mickadoo: Do you think it's worth having this method?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -375,4 +383,46 @@ private function skipBalanceChangeUpdate($params) {
...
+    return $this->leaveRequestRightsService->canCancelToilWithPastDates($params['contact_id'], $params['type_id']);
```
mickadoo: This line is long, can you adjust the indentation like

```
return $this->leaveRequestRightsService->canCancelToilWithPastDates(
  $params['contact_id'], 
  $params['type_id']
);
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2567,6 +2567,50 @@ public function testLeaveRequestCanNotBeCreatedWhenRequestTypeIsToilAndDatesAreI
...
 
+  public function testLeaveRequestCanBeCancelledWhenRequestTypeIsToilAndDatesAreInThePastAndAbsenceTypeDoesNotAllowPastAccrual() {
+    $leaveRequestStatuses = LeaveRequest::getStatuses();
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2017-01-01'),
+      'end_date'   => CRM_Utils_Date::processDate('2017-06-14'),
+    ]);
+
+    WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => 1]);
+    $contactID = 1;
```
mickadoo: You don't need to fabricate a contact?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -2567,6 +2567,50 @@ public function testLeaveRequestCanNotBeCreatedWhenRequestTypeIsToilAndDatesAreI
...
+      'period_id' => $absencePeriod->id
+    ]);
+
+    $params = [
+      'type_id' => $absenceType->id,
+      'contact_id' => $contactID,
+      'status_id' => $leaveRequestStatuses['cancelled'],
+      'from_date' => CRM_Utils_Date::processDate('2017-02-01'),
+      'to_date' => CRM_Utils_Date::processDate('2017-02-05'),
+      'from_date_type' => 1,
```
mickadoo: Are the type IDs always guaranteed to be the same?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestRightsTest.php
```diff
@@ -220,6 +221,66 @@ public function testCanChangeAbsenceTypeForReturnsFalseWhenCurrentUserNotLeaveCo
     );
   }
 
+  public function testCanCancelToilWithPastDatesReturnsTrueWhenCurrentUserIsManagerAndAbsenceTypeDoesNotAllowPastAccrual() {
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'allow_accruals_request' => true,
```
mickadoo: I think most of the booleans in this file are lower case (UNFORGIVABLE) 

Can you run civilint on it?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -1098,6 +1098,132 @@ public function testGetBreakdownIncludeOnlyTheLeaveBalanceChangesOfTheLeaveReque
...
+
+    $params = $this->getDefaultParams($toilParams);
+    $toilRequest = LeaveRequestFabricator::fabricateWithoutValidation($params);
+
+    //Update the toil status to cancelled.
+    $params['status_id'] = $leaveStatuses['cancelled'];
+    $params['id'] = $toilRequest->id;
+
+    $this->setExpectedException('RuntimeException', 'You may only cancel TOIL with dates in the future.');
+    $this->getLeaveRequestService()->create(
```
mickadoo: I think this can all be on the one line

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -336,8 +336,9 @@ private static function validateTOILPastDays($params, $absenceType) {
...
 
-    if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past) {
+    if ($leaveDatesHasPastDates && !$absenceType->allow_accrue_in_the_past && !$isACancellationRequest) {
```
mickadoo: I'm just wondering, would it be cleaner to have all validation at on layer (e.g. the service layer) instead of having some at service layer and some in BAO?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
```diff
@@ -117,6 +117,14 @@ protected function update($params) {
       );
     }
 
+    if ($this->isTOILWithPastDates($params) && $this->isCancelledStatus($params)) {
+      if (!$this->canCanCancelTOILWithPastDates($params)) {
```
mickadoo: Maybe instead of nested `if`s you could use variables to store the conditions? Just seems like there's no point in a nested `if` with no `else`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestRights.php
```diff
@@ -151,4 +152,25 @@ private static function getLeaveRequestStatuses() {
...
+   * past dates or not.
+   *
+   * @param int $contactID
+   *   The contactID of the leave request
+   * @param int $absenceTypeID
+   *   The absence Type ID of the leave request
+   *
+   * @return bool
+   */
+  public function canCancelToilWithPastDates($contactID, $absenceTypeID) {
```
mickadoo: Would it be possible to move all the logic for `canCancel` into a single `canCancel` method like:

```
public function canCancel($leaveRequestId, $contactId)
```

It might be more readable and avoid duplication of code

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: The status_id is always required whether creating or updating a leave request.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
tunbola: I agree. We could discuss that later. I followed the convention for the other exceptions thrown in the file.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
tunbola: good Idea. The best place it can be is in the Leave BAO. It could be a public method used by both classes.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
tunbola: For one, it makes the code cleaner where the function was used. I can swap it with a variable though, not that I think it really matters.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
tunbola: Ok

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
tunbola: Ok

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestRights.php
tunbola: Not sure I understand what you mean here. There is no method named `canCancel` presently. And I didn't want to overcomplicate things. If eventually the need arises for such a method, this method `canCancelToilWithPastDates` can be made private and can be used in the `canCancel` method.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: No need

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
tunbola: Yeah, mostly for tests, its fine

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: Well, the issue is that the LeaveRequest service is mainly concerned with validations about who can do what. The finer grained validations usually reside on the LeaveRequest BAO. 
This help keep things consistent.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
mickadoo: Hmm, not exactly standard behavior for the API, but if that's the case then you can ignore my comment

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequest.php
mickadoo: Whichever you like, just asking

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
tunbola: I agree. The Leave Request API is an exception.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
mickadoo: I guess that makes sense, in that case it's not really validation in two places, in one place it's more like "security" and in the other it's "validation". Does that make sense?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestRights.php
mickadoo: Fair enough, my reasoning was just that it might be more readable to have `canCancel` which would check `canCancelToilWithPastDates` if it's in the past, but it might be going a step too far. I guess if the validation gets more complicated in the future we could refactor it to something like that then.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -346,6 +344,26 @@ private static function validateTOILPastDays($params, $absenceType) {
...
+   * Checks whether the leave request is of type TOIL and has dates in the
+   * past.
+   *
+   * @param array $params
+   *
+   * @return bool
+   */
+  public static function isTOILWithPastDates($params) {
+    if($params['request_type'] !== LeaveRequest::REQUEST_TYPE_TOIL) {
+      return false;
```
mickadoo: boolean case

PR #2562, opened 2018-04-05
PCHR-3444: Lock Option Group

ajesamson:
## Overview
Phone type option group is currently unlocked. This implies new values can be added. This PR changes this by setting the `is_locked` attribute true.

## Before
before

## After
after

## Technical Details
The phone type option group is not locked by default. Locking it automatically sets `is_active` false ([default value when updating](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/BAO/OptionGroup.php#L95)). Due to this, `is_active` was set to `true` while updating.

```php
$data = [
      'id' => '$value.id',
      'is_locked' => 1,
      'is_active' => 1
];
```


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1012.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 trait CRM_HRCore_Upgrader_Steps_1012 {
-
+
```
mickadoo: Just curious, why did you fix whitespace issues in a different file for this PR?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1015.php
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Update is_locked value for phone type option group
+   *
+   * @return bool
+   */
+  public function upgrade_1012() {
+    $data = [
+      'id' => '$value.id',
+      'is_locked' => 1,
+      'is_active' => 1
```
mickadoo: Maybe you could add a comment about the bug that requires `is_active`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1012.php
ajesamson: The upgrader conflicted with that in this PR and i moved the upgrader in this PR to `CRM_HRCore_Upgrader_Steps_1012 `. That is why it appeared as if i fixed the whitespace.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1015.php
ajesamson: That is a good idea. Thanks.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1015.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+trait CRM_HRCore_Upgrader_Steps_1015 {
+  
+  /**
+   * Update is_locked value for phone type option group
+   *
+   * @return bool
+   */
+  public function upgrade_1012() {
```
mickadoo: oops, missed this the first time. You updated the trait class but didn't change the function name.

PR #2563, opened 2018-04-05
PCHR-3469: Add relationship model

reneolivo:
## Overview
This PR adds the relationship model to reqangular. It also fixes a minor PHP warning that is displayed only in debug mode.

This PR is related to https://github.com/compucorp/civihr-tasks-assignments/pull/332 and the relationship model can be seen in action in that PR.

### Relationship Instance

The instances have a `isValid` method that returns true when all of the following requirements are met:

* The `is_active` field is equal to 1.
* The start date is not defined or it's in the past.
* The end date is not defined or it's in the future.

```js
isValid: function () {
  var isActive = parseInt(this.is_active) === 1;
  var startDateIsInThePast = moment().isSameOrAfter(this.start_date);
  var startDateIsNotDefined = !this.start_date;
  var endDateIsInTheFuture = moment().isSameOrBefore(this.end_date);
  var endDateIsNotDefined = !this.end_date;

  return isActive && (startDateIsNotDefined || startDateIsInThePast) &&
    (endDateIsNotDefined || endDateIsInTheFuture);
}
```

### Relationship model

The relationship model mostly follows the same pattern as other models: it calls a relationship api and wraps the results in relationship instances. It also provides an `allValid` method that goes through the `all` method, but makes sure the parent relationship type is active and filters out any invalid relationship returned by the `all` call:

```js
allValid: function (filters, pagination) {
  filters = _.defaults(filters || {}, {
    'relationship_type_id.is_active': 1
  });

  return this.all(filters, pagination)
    .then(function (result) {
      result.list = result.list.filter(function (relationship) {
        return relationship.isValid();
      });

      return result;
    });
}
```

### PHP Issue

A PHP warning that could only be seen when in debug mode was fixed. This was the error:
dr omar cooper hr17 9200

The problem was that the PHP code wanted to check if an associative array had a 'child' index defined, but was not using the `isset` method, which prevents these type of warnings to happen. This is the final solution:

```php
if (!isset($entry['child'])) $entry['child'] = array();
```

The error is only cosmetic and should not be seen by regular users.


org.civicrm.reqangular/src/common/models/relationship.model.js
```diff
@@ -0,0 +1,60 @@
...
+    return Model.extend({
+      /**
+       * Returns a list of relationships, each converted to a model instance
+       *
+       * @param {Object} filters - Values the full list should be filtered by
+       * @param {Object} pagination
+       *   `page` for the current page, `size` for number of items per page
+       * @return {Promise}
+       */
+      all: function (filters, pagination) {
```
deb1990: `sort, additionalParams` parameters are missing.

org.civicrm.reqangular/src/common/models/relationship.model.js
```diff
@@ -0,0 +1,60 @@
...
+        return RelationshipAPI.all(filters, pagination).then(function (response) {
```
deb1990: Can we mention the `.then` in the next line, little bit easier to read.

org.civicrm.reqangular/test/mocks/data/relationship.data.js
```diff
@@ -0,0 +1,27 @@
...
+    hasStarted: { id: '5', is_active: '1', start_date: previousWeek },
+    hasNotFinished: { id: '6', is_active: '1', end_date: nextWeek }
+  };
+
+  return {
+    named: namedRelationships,
+    all: {
+      'is_error': 0,
+      'version': 3,
+      'count': 6,
```
deb1990: `Object.keys(namedRelationships).length`

org.civicrm.reqangular/test/mocks/data/relationship.data.js
```diff
@@ -0,0 +1,27 @@
...
+    isNotActive: { id: '1', is_active: '0' },
+    isInThePast: { id: '2', is_active: '1', end_date: previousWeek },
+    isInTheFuture: { id: '3', is_active: '1', start_date: nextWeek },
+    isActive: { id: '4', is_active: '1' },
+    hasStarted: { id: '5', is_active: '1', start_date: previousWeek },
+    hasNotFinished: { id: '6', is_active: '1', end_date: nextWeek }
+  };
+
+  return {
+    named: namedRelationships,
```
deb1990: I am little confused with this property called `named`. do we get similar response from api?

org.civicrm.reqangular/test/models/instances/relationship.instance.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    }));
+
+    it('inherits from ModelInstance', function () {
+      expect(_.functions(RelationshipInstance))
+        .toEqual(jasmine.arrayContaining(_.functions(ModelInstance)));
+    });
+
+    describe('isValid()', function () {
+      beforeEach(function () {
+        _.forEach(relationshipData.named, function (relationship, index) {
```
deb1990: Please create a new object `instances` inside the forEach. And test using `instances.isActive` or similar. Right now from the `it` blocks, it looks like you are testing the mock data and not the instances.

org.civicrm.reqangular/test/models/relationship.model.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+    beforeEach(inject(function (_$q_, _$rootScope_,
+      _RelationshipAPI_, _RelationshipInstance_, _RelationshipModel_) {
+      $q = _$q_;
+      $rootScope = _$rootScope_;
+      RelationshipAPI = _RelationshipAPI_;
+      RelationshipInstance = _RelationshipInstance_;
+      RelationshipModel = _RelationshipModel_;
+
+      spyOn(RelationshipAPI, 'all').and.returnValue($q.resolve({
```
deb1990: Please create mock api which should do the same.

org.civicrm.reqangular/test/models/relationship.model.spec.js
```diff
@@ -0,0 +1,105 @@
...
+      }));
+    }));
+
+    it('has the expected api', function () {
+      expect(Object.keys(RelationshipModel)).toEqual(['all', 'allValid']);
+    });
+
+    describe('all()', function () {
+      beforeEach(function () {
+        RelationshipModel.all(filters, pagination)
```
deb1990: Please use https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/121371880/AngularJS#AngularJS-Testingasynchronousmethods when testing async methods.
`done` is missing here.

org.civicrm.reqangular/test/models/relationship.model.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+      beforeEach(function () {
+        spyOn(RelationshipModel, 'all').and.callThrough();
+
+        allValidRelationships = relationshipData.all.values
+          .filter(function (relationship) {
+            return RelationshipInstance.init(relationship, true).isValid();
+          });
+
+        RelationshipModel.allValid(filters, pagination)
```
deb1990: Same as above, use `done`

org.civicrm.reqangular/test/mocks/data/relationship.data.js
reneolivo: This is not from the API. This was added for convenience when testing. Do you need a relationship that is not active? get it from `relationshipData.named.isActive`

igorpavlov:
@rene could you please provide a reference link to what Relationship is? I checked the related PR but I don't see any references there. Is this the ref: https://docs.civicrm.org/user/en/latest/organising-your-data/relationships/ ?

reneolivo:
yes @igorpavlov that's the Entity. In the related PR you might have missed it because the controller that is using it sustained quite a lot of changes. Here's the direct usage:

https://github.com/compucorp/civihr-tasks-assignments/pull/332/files#diff-7e24d6add6cfdc8f257b7ddc9f6cfadaR297

org.civicrm.reqangular/src/common/models/relationship.model.js
```diff
@@ -0,0 +1,65 @@
...
+       * Only returns relationships that are valid.
+       *
+       * @param {Object} filters - filter values to pass to the API.
+       * @param {Object} pagination - number of items to return per page and
+       * the current page to fetch.
+       * @param {string} sort - The field and direction to order by
+       * @param {object} additionalParams - Additional params to pass to the api
+       * @return {Promise}
+       */
+      allValid: function (filters, pagination, sort, additionalParams) {
```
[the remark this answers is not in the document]
reneolivo: As of now, there is no way to determine if a relationship is valid through the API. This is something that was asked in the Core chat. After discussing with @davialexandre we agreed that validating the relationships after requesting them through the API is the best solution for now.

org.civicrm.reqangular/test/models/instances/relationship.instance.spec.js
```diff
@@ -0,0 +1,59 @@
...
+      });
+
+      it('returns true when the relationship end date is in the future', function () {
+        expect(instances.hasNotFinished.isValid()).toBe(true);
+      });
+
+      it('returns false when the relationship end date is in the past', function () {
+        expect(instances.isInThePast.isValid()).toBe(false);
+      });
+    });
```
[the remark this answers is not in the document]
reneolivo: I could add more tests, but those are covered by the isActive relationship since it has no start_date or end_date defined. Should I add those?

org.civicrm.reqangular/test/models/relationship.model.spec.js
```diff
@@ -0,0 +1,113 @@
...
+    it('has the expected api', function () {
+      expect(Object.keys(RelationshipModel)).toEqual(['all', 'allValid']);
+    });
+
+    describe('all()', function () {
+      beforeEach(function (done) {
+        RelationshipModel.all(filters, pagination, sort, additionalParams)
+          .then(function (_result_) {
+            result = _result_;
+          }).finally(done);
```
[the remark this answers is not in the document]
reneolivo: This is part of the style-guide. It's better to follow it than to figure out if the async call would need a `done` or not.

org.civicrm.reqangular/src/common/models/instances/relationship.instance.js
```diff
@@ -0,0 +1,34 @@
...
+
+  instances.factory('RelationshipInstance', RelationshipInstance);
+
+  RelationshipInstance.$inject = ['ModelInstance'];
+
+  function RelationshipInstance (ModelInstance) {
+    return ModelInstance.extend({
+      /**
+       * Determines if the relationship is valid by checking if it's active,
+       * the start date is in the past, and the end date is in the future.
```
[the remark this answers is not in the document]
reneolivo: Reworded it a bit to include that condition

PR #2564, opened 2018-04-09
PCHR-3525: Fix TOIL bugs

igorpavlov:
## Overview

This PR fixes issues with TOIL found by QA. It now:
1) Resets "to" time along with "to" date in UI.
2) Refreshes TOIL accrual options for TOIL in days each time the option editor popup is closed.
3) Fixes edge case for Time Amount Picker, when you change the maximum amount, for example, from 2.5 to 3 - it now shows correct minute options.

## Before

_All issues are shown in the order they are described in the Overview_

### Reset TOIL "to" time

```js
// TOIL controller

// Extend the date resetting function with resetting time
detailsController.resetUIInputsExtended = resetTime;

function resetTime (dateType) {
  detailsController.uiOptions.times[dateType].time = '';
```

### Refresh TOIL options every time the option editor is closed

The event handler that was used previously `crmFormSuccess` doesn't fire on Enabling/Disabling and reordering options. Instead, `crmUnload` is now used which is fired every time the popup is closed (ref: https://docs.civicrm.org/dev/en/latest/framework/ajax/#events_1)

```js
function openToilInDaysAccrualOptionsEditor () {
  crmAngService.loadForm('/civicrm/admin/options/hrleaveandabsences_toil_amounts?reset=1')
    //.on('crmFormSuccess', function () {
    .on('crmUnload', function () {
      loadToilAmounts(false);
```

### Fixes edge case for Time Amount Picker

Changing the value didn't rebuild hours/minutes options, now it does.

```js
$scope.$watch('value', function () {
  parseCurrentValue();
  buildOptions(); // added
```

## Comments

This command throws a warning on Linux machines:

```sh
find ./* -type f -mindepth 1 -name package.json | grep -v 'node_modules\\|bower_components'
```

> find: warning: you have specified the -mindepth option after a non-option argument -type, but options are not positional (-mindepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

So it was changed to:

```sh
find ./* -mindepth 1 -name package.json -type f | grep -v 'node_modules\\|bower_components'
```

----

✅Manual Tests - passed
✅Jasmine Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
```diff
@@ -102,9 +102,13 @@ define([
...
-      it('sorts TOIL accrual options by value', function () {
-        expect(_.pluck(controller.toilAmounts, 'value')).toEqual(
-          _.pluck(controller.toilAmounts, 'value').sort());
+      it('sorts TOIL accrual options by weight', function () {
+        expect(_.pluck(controller.toilAmounts, function (amount) {
+          return +amount.weight;
+        })).toEqual(
+          _.pluck(controller.toilAmounts, function (amount) {
+            return +amount.weight;
+          }).sort());
```
reneolivo: can we remove all this code from the it/expect blocks?

```js

describe('when sorting TOIL accrual options', function () {
  var toilAmounts, toilAmountsSortedByWeight;

  beforeEach(function () {
    toilAmountsSortedByWeight = _.pluck(controller.toilAmounts, function (amount) {
      return +amount.weight;
    }).sort();
    toilAmounts = .pluck(controller.toilAmounts, function (amount) {
      return +amount.weight;
    });
  });

  it('sorts them by weight', function () {
    expect(toilAmounts).toEqual(toilAmountsSortedByWeight);
  });
});
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
```diff
@@ -170,6 +174,17 @@ define([
           expect(controller.request.toil_duration).toBe(controller.uiOptions.toil_duration_in_hours * 60);
         });
 
+        describe('when "to" date selected that is greater than "from" date', function () {
+          beforeEach(function () {
+            requestModalHelper.setTestDates(controller, '03/22/2018');
```
reneolivo: `03/22/2018` `03/22` ? since when do we use fairy date formats?

Also is `22/03/2018` the TO date? what is the FROM date?

Do you think it would be possible to do something like

```js
var fromDate = moment().format('format here');
var toDate = moment().add(1, 'week').format('format here');

// pass the date to the setTestDates method
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-toil.controller.spec.js
igorpavlov: Done.

PR #2566, opened 2018-04-10
PCHR-3464: Hide link to edit locked option groups in forms

mickadoo:
## Overview

For locked option groups we don't want to show the gear icon to allow inline editing. This PR hides that link.

## Before

Locked option groups (where `is_locked` is true) would still have a link to edit them inline in forms.

In this example **phone types** is locked

image

## After

Locked option groups (where `is_locked` is true) do not show a link to edit them inline in forms.

In this example **phone types** is locked

image

## Technical Details

The way this works is based on the implementation of the [addOptionsEditLink](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Form/Renderer.php#L350) method. The method will only add the link _if_ the `data-option-edit-path` is set. By removing it we make sure this method does not show the link.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
@@ -1,0 +1,72 @@
+<?php
+
+class CRM_HRCore_Hook_BuildForm_OptionEditPathFilter {
+
+  /**
+   * @var bool[]
+   *   A mapping of option group name to its "is_locked" status
+   */
+  private $optionGroupLockedStatuses = [];
+
+  /**
+   * Handle the form. Any form that can contain elements referencing an
+   * option group should be handled, i.e. all forms
+   *
+   * @param string $formName
+   * @param CRM_Core_Form $form
+   */
+  public function handle($formName, &$form) {
+    $this->filterOptionEditPaths($form);
+  }
+
+  /**
+   * Loop through the form, removing the 'data-option-edit-path' from form
+   * elements that reference locked option groups. If this attribute is not set
+   * then the icon to provide a shortcut to edit this group will not be shown.
+   *
+   * @see \CRM_Core_Form_Renderer::addOptionsEditLink
+   *
+   * @param CRM_Core_Form $form
+   */
+  private function filterOptionEditPaths($form) {
+    /** @var HTML_QuickForm_element $element */
+    foreach ($form->_elements as &$element) {
+      $optionEditPath = $element->getAttribute('data-option-edit-path');
+      if (!$optionEditPath) {
+        continue;
+      }
+
+      $prefix = 'civicrm/admin/options/';
+      if (strpos($optionEditPath, $prefix) !== 0) {
+        // this is not an option group edit link
+        continue;
+      }
+
+      $optionGroupName = str_replace($prefix, '', $optionEditPath);
+
+      if ($this->isLockedOptionGroup($optionGroupName)) {
+        $element->removeAttribute('data-option-edit-path');
+      }
+    }
+  }
+
+  /**
+   * Check whether a given option group is locked
+   *
+   * @param string $optionGroupName
+   *
+   * @return bool
+   */
+  private function isLockedOptionGroup($optionGroupName) {
+    if (!isset($this->optionGroupLockedStatuses[$optionGroupName])) {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
```diff
@@ -0,0 +1,72 @@
...
+      $optionEditPath = $element->getAttribute('data-option-edit-path');
+      if (!$optionEditPath) {
+        continue;
+      }
+
+      $prefix = 'civicrm/admin/options/';
+      if (strpos($optionEditPath, $prefix) !== 0) {
+        // this is not an option group edit link
+        continue;
+      }
```
davialexandre: Is this path check really necessary? As we can see [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Form.php?utf8=%E2%9C%93#L1332) and [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Form.php?utf8=%E2%9C%93#L1349), it's possible to customize this URL using the `option_url` property. Also, if you check [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/PseudoConstant.php#L483-L495) it looks like there's an edge case scenario for when the optionGroup name is not available in the field metadata. In all of these cases, the might not start with `civicrm/admin/options/`. Should we care about that?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
```diff
@@ -0,0 +1,72 @@
...
+
+      $optionGroupName = str_replace($prefix, '', $optionEditPath);
```
davialexandre: Given what I said above, I think there's a chance that the path will not contain the option group name

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @return bool
+   */
+  private function isLockedOptionGroup($optionGroupName) {
+    if (!isset($this->optionGroupLockedStatuses[$optionGroupName])) {
+      $params = ['return' => 'is_locked', 'name' => $optionGroupName];
+      $result = civicrm_api3('OptionGroup', 'get', $params);
+      $result = array_shift($result['values']);
+      $isLocked = CRM_Utils_Array::value('is_locked', $result, FALSE);
+      $this->optionGroupLockedStatuses[$optionGroupName] = (bool) $isLocked;
```
davialexandre: Wouldn't it be better to do a single API call to get all the locked option groups and cache them instead of sending one new API call for each uncached Option Group?

mickadoo: I guess avoiding separate API calls should be prioritized, even if it means looping through the form twice (once to get the target option groups, and again to unset the data-option-edit-path for those that are locked)

mickadoo: Good idea

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
davialexandre: Based on your investigation, it looks like that `option_url` is used when there a need to point to something that is not really an option group. The edge case also seems to be used for similar purposes. In this case, your solution is good enough as is. No need for changes.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/OptionEditPathFilter.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+class CRM_HRCore_Hook_BuildForm_OptionEditPathFilter {
+
+  /**
+   * @var string[]
+   *   A an array of locked option group names, indexed by group ID
```
davialexandre: s/A an/An/

PR #2570, opened 2018-04-12
PCHR-3567: Restrict Message Template Form Access Based on Permission

ajesamson:
## Overview
Two new granular permissions for `edit message template` permission are being added to CiviCRM Core. The new permissions are `edit system workflow message templates` and `edit user-driven message templates`. The `edit user-driven message templates` is assigned to `civihr_admin` role by default. The initial permission for super admin to `edit message template` permission was also retained. This calls for restriction on the form editing page.

## Before
before

before_2

## After
after

after_2

## Technical Details
A form filter hook `CRM_HRCore_Hook_BuildForm_MessageTemplatesFormFilter` was added to ensure users with one of the permission does not have access to edit form belonging to other permission group.
```php
private function filterMessageTemplates($form) {
    // Only system workflow message template have workflow id set
    $workflowId = $form->getVar("_workflow_id");
    if (isset($workflowId)) {
      $canView = CRM_Core_Permission::check('edit system workflow message templates');
    } else {
      $canView = CRM_Core_Permission::check('edit user-driven message templates');
    }
  
    if (! $canView  && ! CRM_Core_Permission::check('edit message templates')) {
      CRM_Core_Session::setStatus(ts('You do not have permission to view requested page.'), ts('Access Denied'));
      $url = CRM_Utils_System::url('civicrm/admin/messageTemplates', "reset=1");
      CRM_Utils_System::redirect($url);
    }
  }
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/MessageTemplatesFormFilter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/MessageTemplatesFormFilter.php
@@ -1,0 +1,54 @@
+<?php
+
+class CRM_HRCore_Hook_BuildForm_MessageTemplatesFormFilter {
+  
+  /**
+   * Restrict access to form page based on permission
+   *
+   * @param $formName
+   * @param $form
+   */
+  public function handle($formName, &$form) {
+    if (!$this->shouldHandle($formName)) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


ajesamson:
This is closed because the implementation have been [replicated in the core](https://github.com/civicrm/civicrm-core/pull/11974). This is because there is another restriction to be made on the message template view page which has been carried in the core template. It will be a good solution to have the implementations in the same place.

PR #2575, opened 2018-04-18
PCHR-3580: Allow multiple requests on same date on backend

igorpavlov:
## Overview

At the moment it is impossible to create Leave Requests and TOIL Accruals for the same day.

This PR allows to create requests for the same day according to the rules:

1) It should be possible to create multiple leave requests in hours within the same day. They can also "touch" together in terms of time, for example, one request ends at 9:00 and another starts at 9:00 of the same day.
2) It should be possible to create one AM and one PM leave request within the same day using leave in days allowance.
3) It should not be possible to create one leave request in days and one leave request in hours on the same day, regardless of the request in hours start and end time. 
4) It should not be possible to create more than one AM leave request within the same day. Same for PM.
5) It should be possible for a TOIL Accrual request to overlap one or more leave (non-TOIL) requests.
6) It should be possible for multiple TOIL requests to be created for one day as long as their start-end time does not overlap.

## Technical Details

The only function that was touched is `findOverlappingLeaveRequests`.

ℹ️`findOverlappingLeaveRequests` function is only used ONCE in the whole codebase, so it is safe to change accepted args:

image

```php
// since we now need to pass 5 more leave request params
// we now pass a whole assoc array $leaveRequestParams as an arg instead of separate args
public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveRequestStatus = [], $excludePublicHolidayLeaveRequests = TRUE) {
  // these vars are needed since they are either used in multiple places
  // or are simply optional and/or should be converted type-wise
  $fromDate = $leaveRequestParams['from_date'];
  $fromDateType = isset($leaveRequestParams['from_date_type']) ? (int)$leaveRequestParams['from_date_type'] : NULL;
  $toDate = $leaveRequestParams['to_date'];
  $toDateType = isset($leaveRequestParams['to_date_type']) ? (int)$leaveRequestParams['to_date_type'] : NULL;

  // ...

  // due to new rules a the following variables need to be considered
  $leaveRequestAbsenceType = AbsenceType::findById($leaveRequestParams['type_id']);
  $isCalculationUnitInHours = $leaveRequestAbsenceType->isCalculationUnitInHours();
  $absenceTypes = AbsenceType::getEnabledAbsenceTypes();
  $absenceTypesIDsWithSameUnit = [];
  $absenceTypesIDsWithDifferentUnit = [];
  $leaveRequestDayTypes = array_flip(self::buildOptions('from_date_type', 'validate'));
  $halfDayAMID = $leaveRequestDayTypes['half_day_am'];
  $halfDayPMID = $leaveRequestDayTypes['half_day_pm'];
  $isTOIL = $leaveRequestParams['request_type'] === LeaveRequest::REQUEST_TYPE_TOIL;

  // we should consider units
  foreach ($absenceTypes as $absenceType) {
    if ($leaveRequestAbsenceType->calculation_unit === $absenceType->calculation_unit) {
      $absenceTypesIDsWithSameUnit[] = $absenceType->id;
    } else {
      $absenceTypesIDsWithDifferentUnit[] = $absenceType->id;
    }
  }

  // this SQL WHERE rule actually considers only those requests that are needed
  $noOverlappingCondition =
    $isCalculationUnitInHours || $isTOIL
    ?
      "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
      (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
    : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
      ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
      ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");

  // as a generic rule, we do not mind to *completely* overlay TOIL and Non-TOIL (Leave/Sick) requests
  $ignoreRequestTypes =
    'lr.request_type ' . ($isTOIL ? '=' : '!=') . "'" . LeaveRequest::REQUEST_TYPE_TOIL . "'";

  // the query now includes the conditions described above
  $query = "
    SELECT DISTINCT lr.* FROM {$leaveRequestTable} lr
    INNER JOIN {$leaveRequestDateTable} lrd
      ON lrd.leave_request_id = lr.id
    INNER JOIN {$leaveBalanceChangeTable} lbc
      ON lbc.source_id = lrd.id AND lbc.source_type = %1
    WHERE lr.contact_id = %2 AND lr.is_deleted = 0 AND ({$noOverlappingCondition})
      AND {$ignoreRequestTypes}";

  // the rest code remains the same
}
```

And finally, the usage:

```php
private static function validateNoOverlappingLeaveRequests($params) {
  // ...

  $overlappingLeaveRequests = self::findOverlappingLeaveRequests(
    $params, // <-- now an assoc array is passed
    $leaveRequestStatusFilter // not touched
  );

  // ...
}
```

### Fabricator improvement

Now Leave Request fabricator considers calculation unit when adding default params:

```php
private static function mergeDefaultParams($params) {
  $absenceType = AbsenceType::findById($params['type_id']);
  $isCalculationUnitInHours = $absenceType->isCalculationUnitInHours();
  $isTOIL = isset($params['request_type']) && $params['request_type'] === LeaveRequest::REQUEST_TYPE_TOIL;

  $defaultParams = [
    'status_id' => self::getStatusId('approved'),
  ];
 
  if (!$isCalculationUnitInHours) {
    $defaultParams['from_date_type'] = self::getDayTypeId('all_day');

    if (!empty($params['to_date']) && empty($params['to_date_type'])) {
      $defaultParams['to_date_type'] = self::getDayTypeId('all_day');
    }
  } else {
    $defaultParams['from_date_amount'] = 1;
    $defaultParams['to_date_amount'] = 1;
  }

  if ($isTOIL) {
    $defaultParams['toil_duration'] = 1;
    $defaultParams['toil_to_accrue'] = 1;
  }
 
  return array_merge($defaultParams, $params);
}
```

### Bug in tests

CreateLeaveBalanceChange accepts Absence Type ID as a first param, not Absence Period.

image

------

✅PHP Unit Tests - augmented and passed

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -663,13 +663,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     ];
 
     $overlappingLeaveRequests = self::findOverlappingLeaveRequests(
-      $params['contact_id'],
-      $params['from_date'],
-      isset($params['from_date_type']) ? $params['from_date_type'] : null,
-      $params['to_date'],
-      isset($params['to_date_type']) ? $params['to_date_type'] : null,
-      $params['type_id'],
-      $params['request_type'],
+      $params,
       $leaveRequestStatusFilter
     );
 
@@ -865,25 +859,26 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
   }
 
   /**
-   * Returns existing leave requests considered "overlapping" for the given contact
-   * within the date/time period, absence type, request type and statuses supplied.
+   * Returns existing leave requests considered "overlapping"
+   * for the given leave request parameters and statuses supplied.
    *
-   * @param int $contactID
-   * @param string $fromDate
-   * @param string $toDate
-   * @param int $typeID
-   * @param string $requestType
+   * @param array $leaveRequestParams
    * @param array $leaveRequestStatus
    * @param boolean $excludePublicHolidayLeaveRequests
    *   Whether to exclude public holiday leave requests from overlapping leave requests or not
    *
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest[]|null
    */
-  public static function findOverlappingLeaveRequests($contactID, $fromDate, $fromDateType = null, $toDate, $toDateType = null, $typeID, $requestType, $leaveRequestStatus = [], $excludePublicHolidayLeaveRequests = true) {
+  public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveRequestStatus = [], $excludePublicHolidayLeaveRequests = TRUE) {
+    $fromDate = $leaveRequestParams['from_date'];
+    $fromDateType = isset($leaveRequestParams['from_date_type']) ? $leaveRequestParams['from_date_type'] : NULL;
+    $toDate = $leaveRequestParams['to_date'];
+    $toDateType = isset($leaveRequestParams['to_date_type']) ? $leaveRequestParams['to_date_type'] : NULL;
+
     $leaveRequestTable = self::getTableName();
     $leaveRequestDateTable = LeaveRequestDate::getTableName();
     $leaveBalanceChangeTable = LeaveBalanceChange::getTableName();
-    $leaveRequestAbsenceType = AbsenceType::findById($typeID);
+    $leaveRequestAbsenceType = AbsenceType::findById($leaveRequestParams['type_id']);
     $isCalculationUnitInHours = $leaveRequestAbsenceType->isCalculationUnitInHours();
     $absenceTypes = AbsenceType::getEnabledAbsenceTypes();
     $absenceTypesIDsWithSameUnit = [];
@@ -891,7 +886,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $leaveRequestDayTypes = array_flip(self::buildOptions('from_date_type', 'validate'));
     $halfDayAMID = $leaveRequestDayTypes['half_day_am'];
     $halfDayPMID = $leaveRequestDayTypes['half_day_pm'];
-    $isTOIL = $requestType === LeaveRequest::REQUEST_TYPE_TOIL;
+    $isTOIL = $leaveRequestParams['request_type'] === LeaveRequest::REQUEST_TYPE_TOIL;
 
     foreach ($absenceTypes as $absenceType) {
       if ($leaveRequestAbsenceType->calculation_unit === $absenceType->calculation_unit) {
@@ -933,7 +928,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
 
     $params = [
       1 => [LeaveBalanceChange::SOURCE_LEAVE_REQUEST_DAY, 'String'],
-      2 => [$contactID, 'Integer'],
+      2 => [$leaveRequestParams['contact_id'], 'Integer'],
       3 => [CRM_Utils_Date::processDate($fromDate, null, false, 'Y-m-d'), 'String'],
       4 => [CRM_Utils_Date::processDate($toDate, null, false, 'Y-m-d'), 'String'],
       5 => [self::REQUEST_TYPE_PUBLIC_HOLIDAY, 'String']
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -863,6 +863,16 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * for the given leave request parameters and statuses supplied.
    *
    * @param array $leaveRequestParams
+   *   List of parameters to use to create the Case.
+   *   [
+   *     'contact_id' => int,
+   *     'type_id' => int,
+   *     'request_type' => int,
+   *     'from_date' => string,
+   *     'from_date_type' => string|int optional,
+   *     'to_date' => string,
+   *     'to_date_type' => string|int optional
+   *   ]
    * @param array $leaveRequestStatus
    * @param boolean $excludePublicHolidayLeaveRequests
    *   Whether to exclude public holiday leave requests from overlapping leave requests or not
@@ -871,9 +881,9 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    */
   public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveRequestStatus = [], $excludePublicHolidayLeaveRequests = TRUE) {
     $fromDate = $leaveRequestParams['from_date'];
-    $fromDateType = isset($leaveRequestParams['from_date_type']) ? $leaveRequestParams['from_date_type'] : NULL;
+    $fromDateType = isset($leaveRequestParams['from_date_type']) ? (int)$leaveRequestParams['from_date_type'] : NULL;
     $toDate = $leaveRequestParams['to_date'];
-    $toDateType = isset($leaveRequestParams['to_date_type']) ? $leaveRequestParams['to_date_type'] : NULL;
+    $toDateType = isset($leaveRequestParams['to_date_type']) ? (int)$leaveRequestParams['to_date_type'] : NULL;
 
     $leaveRequestTable = self::getTableName();
     $leaveRequestDateTable = LeaveRequestDate::getTableName();
@@ -902,8 +912,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
         "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
         (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
       : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
-        ($fromDateType !== null && (int)$fromDateType === (int)$halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
-        ($toDateType !== null && (int)$toDateType === (int)$halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
+        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
+        ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
 
     $ignoreRequestTypes =
       'lr.request_type ' . ($isTOIL ? '=' : '!=') . "'" . LeaveRequest::REQUEST_TYPE_TOIL . "'";
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -880,9 +880,13 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
    * @return \CRM_HRLeaveAndAbsences_BAO_LeaveRequest[]|null
    */
   public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveRequestStatus = [], $excludePublicHolidayLeaveRequests = TRUE) {
-    $fromDate = $leaveRequestParams['from_date'];
+    $fromDateObj = new DateTime($leaveRequestParams['from_date']);
+    $fromDate = $fromDateObj->format('Y-m-d');
+    $fromDateTime = $fromDateObj->format('Y-m-d H:i:s');
     $fromDateType = isset($leaveRequestParams['from_date_type']) ? (int)$leaveRequestParams['from_date_type'] : NULL;
-    $toDate = $leaveRequestParams['to_date'];
+    $toDateObj = new DateTime($leaveRequestParams['to_date']);
+    $toDate = $toDateObj->format('Y-m-d');
+    $toDateTime = $toDateObj->format('Y-m-d H:i:s');
     $toDateType = isset($leaveRequestParams['to_date_type']) ? (int)$leaveRequestParams['to_date_type'] : NULL;
 
     $leaveRequestTable = self::getTableName();
@@ -891,14 +895,20 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $leaveRequestAbsenceType = AbsenceType::findById($leaveRequestParams['type_id']);
     $isCalculationUnitInHours = $leaveRequestAbsenceType->isCalculationUnitInHours();
     $absenceTypes = AbsenceType::getEnabledAbsenceTypes();
+    $absenceTypesIDs = [];
     $absenceTypesIDsWithSameUnit = [];
     $absenceTypesIDsWithDifferentUnit = [];
     $leaveRequestDayTypes = array_flip(self::buildOptions('from_date_type', 'validate'));
     $halfDayAMID = $leaveRequestDayTypes['half_day_am'];
     $halfDayPMID = $leaveRequestDayTypes['half_day_pm'];
-    $isTOIL = $leaveRequestParams['request_type'] === LeaveRequest::REQUEST_TYPE_TOIL;
+    $toilType = LeaveRequest::REQUEST_TYPE_TOIL;
+    $isTOIL = $leaveRequestParams['request_type'] === $toilType;
+    $isCalculationUnitInHoursOrTOIL = $isCalculationUnitInHours || $isTOIL;
+    $absenceTypesSelector = '';
+    $considerRequestsInDifferentUnit = '';
 
     foreach ($absenceTypes as $absenceType) {
+      $absenceTypesIDs[] = $absenceType->id;
       if ($leaveRequestAbsenceType->calculation_unit === $absenceType->calculation_unit) {
         $absenceTypesIDsWithSameUnit[] = $absenceType->id;
       } else {
@@ -906,17 +916,40 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
       }
     }
 
-    $noOverlappingCondition =
-      $isCalculationUnitInHours || $isTOIL
-      ?
-        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
-        (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
-      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
-        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
-        ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
+    $dateTimeSelector = $isCalculationUnitInHoursOrTOIL
+      ? "lr.from_date < '{$toDateTime}' AND lr.to_date > '{$fromDateTime}'"
+      : "lrd.date BETWEEN '{$fromDate}' AND '{$toDate}'";
+
+    if ($isCalculationUnitInHours) {
+      $absenceTypesSelector = 'AND lr.type_id IN (' . implode(',', $absenceTypesIDsWithSameUnit) . ')';
+
+      if (count($absenceTypesIDsWithDifferentUnit)) {
+        $considerRequestsInDifferentUnit =
+          "OR ((lrd.date BETWEEN '{$fromDate}' AND '{$toDate}') AND lr.type_id IN (" .
+          implode(',', $absenceTypesIDsWithDifferentUnit) . "))";
+      }
+    }
+
+    if (!$isCalculationUnitInHours && !$isTOIL) {
+      $absenceTypesSelector = 'AND lr.type_id IN (' . implode(',', $absenceTypesIDs) . ')';
+    }
+
+    $considerHalfDayRequests =
+      ($fromDateType === $halfDayPMID
+        ? "AND (DATE_FORMAT(lr.to_date, '%Y-%m-%d') != '{$fromDate}' " .
+          "OR lr.to_date_type != {$halfDayAMID})"
+        : "") .
+      ($toDateType === $halfDayAMID
+        ? "AND (DATE_FORMAT(lr.from_date, '%Y-%m-%d') != '{$toDate}' " .
+          "OR lr.from_date_type != {$halfDayPMID})"
+        : "");
+
+    $overlapSelector =
+      "($dateTimeSelector $absenceTypesSelector $considerRequestsInDifferentUnit) " .
+      (!$isCalculationUnitInHoursOrTOIL ? $considerHalfDayRequests : '');
 
     $ignoreRequestTypes =
-      'lr.request_type ' . ($isTOIL ? '=' : '!=') . "'" . LeaveRequest::REQUEST_TYPE_TOIL . "'";
+      'lr.request_type ' . ($isTOIL ? '=' : '!=') . "'" . $toilType . "'";
 
     $query = "
       SELECT DISTINCT lr.* FROM {$leaveRequestTable} lr
@@ -924,7 +957,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
         ON lrd.leave_request_id = lr.id
       INNER JOIN {$leaveBalanceChangeTable} lbc
         ON lbc.source_id = lrd.id AND lbc.source_type = %1
-      WHERE lr.contact_id = %2 AND lr.is_deleted = 0 AND ({$noOverlappingCondition})
+      WHERE lr.contact_id = %2 AND lr.is_deleted = 0 AND ({$overlapSelector})
         AND {$ignoreRequestTypes}";
 
     if (is_array($leaveRequestStatus) && !empty($leaveRequestStatus)) {
@@ -933,15 +966,13 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     }
 
     if ($excludePublicHolidayLeaveRequests) {
-      $query .= " AND lr.request_type != %5";
+      $query .= " AND lr.request_type != %3";
     }
 
     $params = [
       1 => [LeaveBalanceChange::SOURCE_LEAVE_REQUEST_DAY, 'String'],
       2 => [$leaveRequestParams['contact_id'], 'Integer'],
-      3 => [CRM_Utils_Date::processDate($fromDate, null, false, 'Y-m-d'), 'String'],
-      4 => [CRM_Utils_Date::processDate($toDate, null, false, 'Y-m-d'), 'String'],
-      5 => [self::REQUEST_TYPE_PUBLIC_HOLIDAY, 'String']
+      3 => [self::REQUEST_TYPE_PUBLIC_HOLIDAY, 'String']
     ];
 
     $leaveRequest = CRM_Core_DAO::executeQuery($query, $params, true, self::class);
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
@@ -904,7 +904,7 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveRequest extends CRM_HRLeaveAndAbsences_DAO
     $noOverlappingCondition =
       $isCalculationUnitInHours || $isTOIL
       ?
-        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}' AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . "))" .
+        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
         (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
       : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
         ($fromDateType !== null && (int)$fromDateType === (int)$halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -861,32 +863,56 @@ public static function findPublicHolidayLeaveRequestEvenIfDeleted($contactID, Pu
...
+        $absenceTypesIDsWithDifferentUnit[] = $absenceType->id;
+      }
+    }
+
+    $noOverlappingCondition =
+      $isCalculationUnitInHours
+      ?
+        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}' AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . "))" .
+        (count($absenceTypesIDsWithDifferentUnit) ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
+      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")";
```
tunbola: Can you use implode instead of join?  The keyword could be confusing as one would think its part of the SQl query.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1112,12 +1112,98 @@ public function testFindOverlappingLeaveRequestsForOneOverlappingLeaveRequest()
...
+    $testFromDate = '2018-04-13 12:00:00';
+    $testToDate = '2018-04-15 15:00:00';
+
+    $absenceTypes = [
+      'days' => AbsenceTypeFabricator::fabricate([ 'calculation_unit' => 1 ]),
+      'hours' => AbsenceTypeFabricator::fabricate([ 'calculation_unit' => 2 ])
+    ];
+
+    // Test suites: date from, date to, unit, should overlap or not
+    $overlappingRequestsTestSuites = [
```
tunbola: This should be moved to a data provider.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1112,12 +1112,98 @@ public function testFindOverlappingLeaveRequestsForOneOverlappingLeaveRequest()
...
     $endDate = '2016-11-03';
 
-    $overlappingRequests = LeaveRequest::findOverlappingLeaveRequests($contactID, $startDate, $endDate);
+    $overlappingRequests = LeaveRequest::findOverlappingLeaveRequests($contactID, $startDate, $endDate, $this->absenceType->id, LeaveRequest::REQUEST_TYPE_LEAVE);
     $this->assertCount(1, $overlappingRequests);
     $this->assertInstanceOf(LeaveRequest::class, $overlappingRequests[0]);
     $this->assertEquals($leaveRequest1->id, $overlappingRequests[0]->id);
   }
 
+  public function testFindOverlappingLeaveRequestsForHoursUnitRequests() {
```
tunbola: I think there should actually be more detailed tests for this. Real case scenarios e.g A leave request exists 1pm to 2pm and 4pm to 5pm on a single day, would the staff be able to request leave for 3pm to 4:30pm which should not be possible.

Also tests for the inverse: would the staff be able to request leave for 2pm to 3pm ? which should be possible

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1112,12 +1112,98 @@ public function testFindOverlappingLeaveRequestsForOneOverlappingLeaveRequest()
...
+      $requestUnit = $overlappingRequestsTestSuite[2];
+      $shouldRequestOverlap = $overlappingRequestsTestSuite[3];
+      $absenceType = $absenceTypes[$requestUnit];
+
+      $leaveRequestToTest = LeaveRequestFabricator::fabricateWithoutValidation([
+        'type_id' => $absenceType->id,
+        'contact_id' => $contactID,
+        'from_date' => $requestFromDate->format('YmdHis'),
+        'to_date' => $requestToDate->format('YmdHis'),
+      ], true);
```
tunbola: No need to pass TRUE to the fabricator, this will create balance changes which are not necessary for the scope of this test.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -861,32 +859,73 @@ public static function findPublicHolidayLeaveRequestEvenIfDeleted($contactID, Pu
...
+      }
+    }
+
+    $noOverlappingCondition =
+      $isCalculationUnitInHours || $isTOIL
+      ?
+        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
+        (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
+      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
+        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
```
tunbola: No need to check if fromDateType of toDateType is NULL since they will always be set when calculation unit is in days.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -861,32 +859,73 @@ public static function findPublicHolidayLeaveRequestEvenIfDeleted($contactID, Pu
...
+        $absenceTypesIDsWithSameUnit[] = $absenceType->id;
+      } else {
+        $absenceTypesIDsWithDifferentUnit[] = $absenceType->id;
+      }
+    }
+
+    $noOverlappingCondition =
+      $isCalculationUnitInHours || $isTOIL
+      ?
+        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
```
tunbola: These whole logic/query is very unreadable, It would be a good idea to separate them into smaller chunks and assign a readable variable name to each of them.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -861,32 +859,73 @@ public static function findPublicHolidayLeaveRequestEvenIfDeleted($contactID, Pu
...
+        (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
+      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
+        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
+        ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
```
tunbola: I think this looks a bit complex and it will be easier to list scenarios where the leave requests would not overlap and then negate those scenarios to get the desired result where they overlap
-  There will not be overlap when the fromDateType is half day PM and there exists another leave with to date == fromDate and to_date_type as half day AM.
`condition =  NOT(lr.to_date = fromDate AND lr.to_date_type != fromDateType)`
- There will not be overlap when the fromDateType is half day AM (**and a single day request**) and there exists another leave request with from_date with same date as fromDate and from_date_type is half PM
`condition =  NOT(lr.from_date = fromDate AND lr.from_date_type != fromDateType)`
-  There will not be overlap when the toDateType is half day AM and there exists another leave with from_date == toDate and from_date_type as half day PM
`condition =  NOT(lr.from_date = toDate AND lr.from_date_type != fromDateType)`
- There will not be overlap when the toDateType is half day PM (**and a single day request**) and there exists another leave request with to_date == toDate and to_date_type is half AM
`condition =  NOT(lr.to_date = toDate AND lr.to_date_type != toDateType)`

Looking at it this way it seems the scenario for single day requests was missed out of the logic.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1112,12 +1112,236 @@ public function testFindOverlappingLeaveRequestsForOneOverlappingLeaveRequest()
...
+
+      $this->assertCount($shouldRequestOverlap ? 1 : 0, $overlappingRequests);
+      if ($shouldRequestOverlap) {
+        $this->assertInstanceOf(LeaveRequest::class, $overlappingRequests[0]);
+        $this->assertEquals($overlappingRequests[0]->id, $leaveRequestToTestID);
+      }
+    }
+  }
+
+  public function testFindOverlappingLeaveRequestsTOILinDaysTreatedAsRequestInHours() {
```
tunbola: This is a generic test name, at first glance for someone not familiar with the code, the name provokes another question, How are request in hours supposed to be treated?  
This generic name allows a lot of things that could have been splitted into different tests to be bundled as one.

tunbola:
@igorpavlov , I discovered that the tests were more focused on the validation part, i.e making sure that leave requests can now overlap based on the new conditions allowed for requests to overlapp while there is no test for making sure that leave requests on same day can actually be created or if a there can be multiple leave in hours for same day if their times don't overlap.
There might be a problem to achieve that actually because the Leave Request date table has the combination of the leave_request_id and the date as unique fields on the table.
Removing the index alone won't solve the problem, because we need a way to get the dates for a particular leave request without getting dates for other leave request on same day, So a solution that won't break existing code needs to be implemented.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: OK, we agreed that it should actually stay in the suite - this provides an instant overview of which cases are tested.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: I think done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: Bug spotted, thanks a lot!

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
-    $noOverlappingCondition =
-      $isCalculationUnitInHours || $isTOIL
-      ?
-        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
-        (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
-      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
-        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
-        ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
+    $dateTimeSelector = $isCalculationUnitInHoursOrTOIL
+      ? "lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'"
```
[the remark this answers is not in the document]
tunbola: Just thinking about the date formats here, When this function is used in the Leave Request BAO validation it will pass in dates already processed by CRM_Utils_Date::processDate and when the function is called by another function it may accept date in another format. For consistency and so that we would be comparing dates in same format. I would suggest we convert the from date and to date to dateTime objects.
That way here, we can use the sql query parameters %3 and %4 for fromDate and toDate respectively, 
```php
      3 => [CRM_Utils_Date::processDate($fromDate, null, false, 'Y-m-d'), 'String'],
       4 => [CRM_Utils_Date::processDate($toDate, null, false, 'Y-m-d'), 'String'],
``` 
remember to change this to something like:
```php
      3 => [$fromDate->format('Y-m-d H:i:s', 'String'],
       4 => [$toDate->format('Y-m-d H:i:s', 'String'],
```
In the considerHalfDayRequests variable you can maintain the condition and instead use 
`!= {$fromDate->format('Y-m-d')})`.
This would ensure the datetime format is kept consistent throughout the function and under control.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -4071,6 +4082,177 @@ public function testLeaveRequestBeCreatedWhenLeaveRequestHasNoWorkingDayAndValid
...
+    $params['request_type'] = LeaveRequest::REQUEST_TYPE_TOIL;
+    $params['type_id'] = $absenceTypeTOIL->id;
+
+    $leaveRequest2 = LeaveRequestFabricator::fabricate($params, true);
+
+    $this->assertNotNull($leaveRequest1->id);
+    $this->assertNotNull($leaveRequest2->id);
+  }
+
+  public function testTwoLeaveRequestsOneAMAndAnotherPMCanBeCreatedForTheSameDay() {
```
tunbola: What about another test to show the inverse of this test, i.e that two leave requests with same from_date_type and to_date type cannot be created on the same day? i.e a test for criteria 4 `It should not be possible to create more than one AM leave request within the same day. Same for PM.`

tunbola: Also I've not seen any test for criteria 3 on the PR description i.e `It should not be possible to create one leave request in days and one leave request in hours on the same day, regardless of the request in hours start and end time.`

tunbola: Also the test for inverse of criteria 6: `It should be possible for multiple TOIL requests to be created for one day as long as their start-end time does not overlap.` i.e a test to show that an exception will be thrown when creating a TOIL request with overlapping datetime.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
     $halfDayAMID = $leaveRequestDayTypes['half_day_am'];
     $halfDayPMID = $leaveRequestDayTypes['half_day_pm'];
     $isTOIL = $leaveRequestParams['request_type'] === LeaveRequest::REQUEST_TYPE_TOIL;
+    $isCalculationUnitInHoursOrTOIL = $isCalculationUnitInHours || $isTOIL;
 
     foreach ($absenceTypes as $absenceType) {
+      $absenceTypesIDs[] = $absenceType->id;
```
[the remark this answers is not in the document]
igorpavlov: Done

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
-        "(lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'" . (!$isTOIL ? " AND lr.type_id IN (" . join(',', $absenceTypesIDsWithSameUnit) . ")" : "") . ")" .
-        (count($absenceTypesIDsWithDifferentUnit) && !$isTOIL ? " OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" . join(',', $absenceTypesIDsWithDifferentUnit) . "))" : "")
-      : "lrd.date BETWEEN %3 AND %4 AND lr.type_id IN (" . join(',', array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit)) . ")" .
-        ($fromDateType !== null && $fromDateType === $halfDayPMID ? " AND (DATE_FORMAT(lr.to_date, '%y-%m-%d') != DATE_FORMAT('{$fromDate}', '%y-%m-%d') OR lr.to_date_type != {$halfDayAMID})" : "") .
-        ($toDateType !== null && $toDateType === $halfDayAMID ? " AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') OR lr.from_date_type != {$halfDayPMID})" : "");
+    $dateTimeSelector = $isCalculationUnitInHoursOrTOIL
+      ? "lr.from_date < '{$toDate}' AND lr.to_date > '{$fromDate}'"
+      : 'lrd.date BETWEEN %3 AND %4';
+
+    $absenceTypesIDs = array_merge($absenceTypesIDsWithSameUnit, $absenceTypesIDsWithDifferentUnit);
```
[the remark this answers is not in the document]
igorpavlov: (facepalm). Done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+    $absenceTypesSelector = $isCalculationUnitInHoursOrTOIL
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+      ? (!$isTOIL
+        ? 'AND lr.type_id IN (' . implode(',', $absenceTypesIDsWithSameUnit) . ')'
```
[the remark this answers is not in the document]
igorpavlov: Nope, their times will intersect.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+        : '')
+      : 'AND lr.type_id IN (' . implode(',', $absenceTypesIDs) . ')';
+
+    $considerRequestsWithDifferentUnit = count($absenceTypesIDsWithDifferentUnit) && !$isTOIL
+      ? "OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" .
```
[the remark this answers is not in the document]
igorpavlov: I moved the implementation to `dateTimeSelector`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -897,23 +897,48 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+        : "") .
+      ($toDateType === $halfDayAMID
+        ? "AND (DATE_FORMAT(lr.from_date, '%y-%m-%d') != DATE_FORMAT('{$toDate}', '%y-%m-%d') " .
+          "OR lr.from_date_type != {$halfDayPMID})"
+        : "");
+
+    $overlapSelector =
+      "($dateTimeSelector $absenceTypesSelector) " .
+      ($isCalculationUnitInHoursOrTOIL
+        ? $considerRequestsWithDifferentUnit
```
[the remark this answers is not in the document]
igorpavlov: Nope, we don't!

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1157,6 +1145,7 @@ public function testFindOverlappingLeaveRequestsForHoursUnitRequests() {
       $absenceType = $absenceTypes[$requestUnit];
 
       $leaveRequestToTest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'type_id' => $this->absenceType->id,
```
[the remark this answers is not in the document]
igorpavlov: Pardon. Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1205,12 +1194,13 @@ public function testFindOverlappingLeaveRequestsBetweenTOILAndNonTOILRequests()
       $testingRequestType = $testSuite[1];
 
       $leaveRequestToTest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'type_id' => $this->absenceType->id,
```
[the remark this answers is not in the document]
igorpavlov: Fixed.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1285,7 +1276,64 @@ public function testFindOverlappingLeaveRequestsForAMPMRequests() {
     }
   }
 
-  public function testFindOverlappingLeaveRequestsTOILinDaysTreatedAsRequestInHours() {
+  public function testFindOverlappingLeaveRequestsForHalfDaySingleDayRequests() {
+    $contactID = 1;
+
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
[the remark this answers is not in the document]
igorpavlov: It will not work for Days, it needs to be "lr.from_date <= '{$toDate}' AND lr.to_date >= '{$fromDate}'". I went a bit another way than you suggested.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -909,39 +912,37 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
-      : 'AND lr.type_id IN (' . implode(',', $absenceTypesIDs) . ')';
+    if ($isCalculationUnitInHours) {
+      $absenceTypesSelector = 'AND lr.type_id IN (' . implode(',', $absenceTypesIDsWithSameUnit) . ')';
 
-    $considerRequestsWithDifferentUnit = count($absenceTypesIDsWithDifferentUnit) && !$isTOIL
-      ? "OR ((lrd.date BETWEEN %3 AND %4) AND lr.type_id IN (" .
-        implode(',', $absenceTypesIDsWithDifferentUnit) . "))"
-      : "";
+      if (count($absenceTypesIDsWithDifferentUnit)) {
+        $absenceTypesSelector .=
```
tunbola: The name for this variable seems wrong somehow, its selects absence type and dates, maybe a better name?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -909,39 +912,37 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+          "OR ((lrd.date BETWEEN %4 AND %5) AND lr.type_id IN (" .
+          implode(',', $absenceTypesIDsWithDifferentUnit) . "))";
+      }
+    } else if (!$isTOIL) {
```
tunbola: Not sure I understand this if/else, the first condition is `$isCalculationUnitInHours`, shouldn't it be `$isCalculationUnitInHoursOrTOIL` because the way it is now, if its not toil, it will still pass the first condition.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
```diff
@@ -909,39 +912,37 @@ public static function findOverlappingLeaveRequests($leaveRequestParams, $leaveR
...
+        ? "AND (DATE_FORMAT(lr.from_date, '%Y-%m-%d') != %5 " .
           "OR lr.from_date_type != {$halfDayPMID})"
         : "");
 
     $overlapSelector =
       "($dateTimeSelector $absenceTypesSelector) " .
-      ($isCalculationUnitInHoursOrTOIL
-        ? $considerRequestsWithDifferentUnit
-        : $considerHalfDayRequests);
+      (!$isCalculationUnitInHoursOrTOIL ? $considerHalfDayRequests : '');
```
tunbola: I see that the `dateTimeSelector` variable is no longer used, and it still exists?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -1220,8 +1212,8 @@ public function testFindOverlappingLeaveRequestsBetweenTOILAndNonTOILRequests()
   public function testFindOverlappingLeaveRequestsForHalfDayMultipleDayRequests() {
     $contactID = 1;
 
-    $testFromDate = '2018-04-13 00:00:00';
-    $testToDate = '2018-04-15 23:59:00';
+    $testFromDate = '2018-04-13';
+    $testToDate = '2018-04-15';
```
tunbola: Why remove the time?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -4253,6 +4223,188 @@ public function testTwoLeaveRequestsOneAMAndAnotherPMCanBeCreatedForTheSameDay()
...
+    );
+
+    $this->createLeaveBalanceChange($periodEntitlement->id, 100);
+
+    $this->setExpectedException(
+      'CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestException',
+      'This leave request overlaps with another request. Please modify dates of this request'
+    );
+
+    for ($i = 1; $i <= 2; $i++) {
```
tunbola: Can you copy this twice rather than use a loop? store the parameters in a variable and use it twice for the fabricators .

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -4253,6 +4223,188 @@ public function testTwoLeaveRequestsOneAMAndAnotherPMCanBeCreatedForTheSameDay()
...
```
tunbola: Same as above, please store the parameters in a variable, and use it in the fabricator to create both requests. The advantage would be that we would know which of the requests threw the exception as the setExpectedException would be just before the second request.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
```diff
@@ -4253,6 +4223,188 @@ public function testTwoLeaveRequestsOneAMAndAnotherPMCanBeCreatedForTheSameDay()
...
+        'contact_id' => $contactId,
+        'from_date' => CRM_Utils_Date::processDate($date),
+        'to_date' => CRM_Utils_Date::processDate($date),
+        'from_date_type' => $halfDayPMID,
+        'to_date_type' => $halfDayPMID
+      ], true);
+    }
+  }
+
+  public function testTwoLeaveRequestsOneAMAndAnotherInHoursCannotBeCreatedForTheSameDay() {
```
tunbola: testTwoLeaveRequestsInDaysndAnotherInHoursCannotBeCreatedForTheSameDay

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: You, that's why I had `$considerRequestsInDifferentUnit`. I put it back, I think now it look neat enough.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: I think there were excessive conditions:

Before: 
```js
if $isCalculationUnitInHoursOrTOIL
  if !$isTOIL "expression"
  else ""
```
is same as 
```
if $isCalculationUnitInHours "expression"
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequest.php
igorpavlov: At line 941 it is actually used. https://github.com/civicrm/civihr/pull/2575/commits/5404622efe233ef82908d5281870e65ca3c970ec#diff-875730b8f60e0ebe0f4cce61f223c6c8R941

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: `...ForHalfDayMultipleDayRequests...` - requests are in days, no time needed.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: for var i = 1; i <= 2; i++; echo 'Done'.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestTest.php
igorpavlov: Done.

igorpavlov:
@tunbola I squashed my PR Feedback commits into one, there is no point to keep all 5.

PR #2576, opened 2018-04-18
FAR-252: Fix error with Scheduled Job trying to expire balances linked to deleted TOIL requests

tunbola:
## Overview
The SQL query responsible for selecting balance changes to expire selects balance changes linked to already deleted TOIL accrual requests. The query is only supposed to select balance changes linked to non deleted TOIL accrual requests.

## Before
- When SQL query selects balances linked to a deleted TOIL request, this causes an exception to be thrown when selecting dates overlapping balance to [expire ](https://github.com/civicrm/civihr/blob/81d97c204a4e6f84ec836f07b055f481a88b6360/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php#L656), particularly when fetching the LeaveRequest linked to the leave date in order to get the [entitlement](https://github.com/civicrm/civihr/blob/81d97c204a4e6f84ec836f07b055f481a88b6360/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php#L441) . 
An Exception: `Unable to find a Leave Request with id XX` is thrown halting the execution of the Scheduled job script..
 

## After
- Since the query to JOIN to the Leave request table is a LEFT JOIN(And we can't change it to an INNER JOIN because of JOINS to the Period Entitlement table), It is possible for the Expression on the Join condition `ON leave_request_date.leave_request_id = leave_request.id AND leave_request.is_deleted = 0` to be ignored and Null rows are returned for the leave request table instead. This is fixed by checking that the Absence Type ID column of the results fetched is NOT NULL

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
@@ -755,7 +755,8 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChange extends CRM_HRLeaveAndAbsenc
       WHERE balance_to_expire.expiry_date IS NOT NULL AND
             balance_to_expire.expiry_date < %3 AND
             balance_to_expire.expired_balance_change_id IS NULL AND
-            expired_balance_change.id IS NULL
+            expired_balance_change.id IS NULL AND
+            coalesce(leave_request.type_id, period_entitlement.type_id) IS NOT NULL
       ORDER BY balance_to_expire.expiry_date ASC, balance_to_expire.id ASC
     ";
 
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
@@ -3642,6 +3642,53 @@ class CRM_HRLeaveAndAbsences_BAO_LeaveBalanceChangeTest extends BaseHeadlessTest
     $this->assertEquals($expectedBroughtForwardLog, $balanceExpiryLogs[$broughtForwardExpiryRecord->id]);
   }
 
+  public function testCreateExpiryRecordsDoesTryToExpireBalanceChangesLinkedToSoftDeletedToilRequest() {
+    $absencePeriod = AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2017-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2017-12-31')
+    ]);
+
+    $periodEntitlement = LeavePeriodEntitlementFabricator::fabricate([
+      'contact_id' => 1,
+      'period_id' => $absencePeriod->id,
+      'type_id' => 1,
+    ]);
+
+    $toilRequest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $periodEntitlement->contact_id,
+      'type_id' => $periodEntitlement->type_id,
+      'from_date' => CRM_Utils_Date::processDate('2017-05-17'),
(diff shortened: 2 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
```diff
@@ -755,7 +755,8 @@ private static function getBalanceChangesToExpire() {
       WHERE balance_to_expire.expiry_date IS NOT NULL AND
             balance_to_expire.expiry_date < %3 AND
             balance_to_expire.expired_balance_change_id IS NULL AND
-            expired_balance_change.id IS NULL
+            expired_balance_change.id IS NULL AND
+            coalesce(leave_request.type_id, period_entitlement.type_id) IS NOT NULL
```
mickadoo: In the description is says

> This is fixed be checking that the Absence Type ID column of the results fetched for the Leave Request corresponding rows is NOT NULL

Why the check for `period_entitlement.id` then?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -3642,6 +3642,53 @@ public function testCreateExpiryRecordsLogsExpiredBalanceChanges() {
     $this->assertEquals($expectedBroughtForwardLog, $balanceExpiryLogs[$broughtForwardExpiryRecord->id]);
   }
 
+  public function testCreateExpiryRecordsDoesTryToExpireBalanceChangesLinkedToSoftDeletedToilRequest() {
```
mickadoo: I'm not sure if "Does Try To" makes a difference here, it doesn't seem to just try to, it actually does it, right?

How about

`testCreatingExpiryRecordsExpiresBalanceChangesLinkedToSoftDeletedToilRequest`?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
```diff
@@ -3642,6 +3642,53 @@ public function testCreateExpiryRecordsLogsExpiredBalanceChanges() {
...
+    $toilRequest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'contact_id' => $periodEntitlement->contact_id,
+      'type_id' => $periodEntitlement->type_id,
+      'from_date' => CRM_Utils_Date::processDate('2017-05-17'),
+      'to_date' => CRM_Utils_Date::processDate('2017-05-17'),
+      'toil_expiry_date' => CRM_Utils_Date::processDate('2017-08-01'),
+      'toil_to_accrue' => 1,
+      'toil_duration' => 100,
+      'request_type' => LeaveRequest::REQUEST_TYPE_TOIL
+    ], true);
```
mickadoo: not a major thing, but the bools are lower case again :-)

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChange.php
tunbola: Well, either the leave request type_id field or the period entitlement type_id field must Not be null for a valid result set. Checking for only the Leave request type_id will cause the results for period entitlement to be lost, hence the reason for checking for both.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php
tunbola: Should actually be `testCreateExpiryRecordsDoesNotExpireBalanceChangesLinkedToSoftDeletedToilRequest` The records linked to a soft deleted TOIl are not supposed to be expired.

PR #2582, opened 2018-04-25
PCHR-3582: Fix UX for public holiday days in leave request modal

AkA84:
## Overview
When a contact has auto-generated "public holiday" request (this happens when one of the absence types has the "Must staff take public holiday as leave?" setting set to "Yes"), if she selects any of the public holiday dates in the request modal, the modal will mark the date as "Public Holiday" and will not show all the controls usually available to the user (like the day type dropdown)

This also is happening when there aren't auto-generated "public holiday" requests (this happens when none of the absence types has the "Must staff take public holiday as leave?" set to "Yes"), but it's incorrect. Public holiday days in this case should be treated, ux-wise, as any other standard working day.

### With auto-generated public holiday requests
before-w-public-holiday-requests

This is correct. The contact already has a public holiday request, so the system informs her that this date, although it can still be selected, doesn't offer any controls in terms of day type of hours length

### Without auto-generated public holiday requests
before-without-public-holiday-requests

This is incorrect. Given that the contact has not public holiday leave requests, she should be allowed to take a leave request on a public holiday date as she should for any other working day

The source of the issue is that the system is using a "Is this date a public holiday?" discriminant to determine whether to lock the date as a public holiday, while instead it should ignore whether a date is a public holiday *on the calendar*, and instead check if that date has/hasn't a public holiday leave request for the contact instead

### With auto-generated public holiday requests
after-w-public-holiday-requests

Still correct, no changes here

### Without auto-generated public holiday requests
after-without-public-holiday-requests
Now the system correctly allow the user to manage the public holiday date as any other date, and there is no mention of it being a public holiday on the UI.

What is happening now is that there isn't a check no whether the date is a public holiday on a calendar, but whether there is already a public holiday leave request for it (for the contact) or not

## Technical Details
In `RequestModalDetailsLeaveController`, in `filterLeaveRequestDayTypes()` the public holiday check had been changes from
```js
function filterLeaveRequestDayTypes (date, dayType) {
  // ...
  return PublicHoliday.isPublicHoliday(date)
    .then(function (result) {
      if (result) {
        // mark it as "public holiday"
      }
```
to
```js
function filterLeaveRequestDayTypes (date, dayType) {
  // ...
  return hasDatePublicHolidayRequest(date)
    .then(function (result) {
      if (result) {
        // mark it as "public holiday"
      }
}
```
where `hasDatePublicHolidayRequest` simply makes an api call to fetch any "public_holiday" leave request for the given date, and cast the results list as a boolean
```js
function hasDatePublicHolidayRequest (date) {
  return LeaveRequest.all({
    public_holiday: 1,
    from_date: { from: date },
    to_date: { to: date }
  })
    .then(function (result) {
      return !!result.list.length;
    });
}
```


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-leave.controller.js
```diff
@@ -174,6 +174,23 @@ define([
         .subtract(moment.duration(timeFrom)).asHours();
     }
 
+    /**
+     * Checks if for the given date exists a "public holiday" request
+     *
+     * @param {String} date
```
igorpavlov: Please align with return

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/controllers/sub-controllers/request-modal-details-leave.controller.js
```diff
@@ -174,6 +174,23 @@ define([
         .subtract(moment.duration(timeFrom)).asHours();
     }
 
+    /**
+     * Checks if for the given date exists a "public holiday" request
```
igorpavlov: Checks if a "public holiday" request exists for the given date

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/components/leave-request-popup-details-tab.component.spec.js
```diff
@@ -1688,5 +1705,26 @@ define([
       return moment.duration(timeTo)
         .subtract(moment.duration(timeFrom)).asHours().toString();
     }
+
+    /**
+     * Toggles whether there is a "public_holiday" leave requests
```
igorpavlov: leave request (singular)

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/sub-controllers/request-modal-details-leave.controller.spec.js
```diff
@@ -276,6 +276,13 @@ define([
           expect(controller.loading['fromDayTypes']).toBe(false);
           expect(controller.loading['toDayTypes']).toBe(false);
         });
+
+        it('checks for public holiday requests only for the currently selected contact', function () {
+          var args = LeaveRequest.all.calls.mostRecent().args[0];
+
+          expect(args.contact_id).toBeDefined();
```
igorpavlov: May be better to check if `controller.request.contact_id` is defined rather than `args.contact_id`? It just adds some readability imho.

AkA84: i'm checking if the `LeaveRequest.all()` method had been called with the contact id as the param, not whether the contact id is present in the controller (i need to assume it is)

PR #2583, opened 2018-04-26
PCHR-3625: Add cache parameter to relationship model and api

reneolivo:
## Overview
This PR allows relationship requests to control their cache behavior.

This PR is part of https://github.com/compucorp/civihr-tasks-assignments/pull/351
For Before and After screens, and more in depth technical details please refer to that PR.

## Technical Details
The `additionalParams` was removed since it was not necessary for neither the relationship model or api. Instead, it was replaced for a `cache` parameter:

```js
// relationship api:
api.extend({
  all: function (filters, pagination, sort, cache) {
    // ...
    return this.getAll('Relationship', filters, pagination, sort, null, null, cache);
  }
});
```

```js
// relationship model:
model.extend({
  all: function (filters, pagination, sort, cache) {
    return RelationshipAPI.all(filters, pagination, sort, cache) // ...
  },
  allValid: function (filters, pagination, sort, cache) {
    return this.all(filters, pagination, sort, cache) // ..
  }
});
```


org.civicrm.reqangular/src/common/services/api/relationship.api.js
```diff
@@ -18,15 +18,16 @@ define([
...
-       * @param {object} pagination
-       * @param {string} sort
+       * @param  {Object}  filters
+       * @param  {Object}  pagination
+       * @param  {String}  sort
+       * @param  {Boolean} cache
        * @return {Promise}
        */
-      all: function (filters, pagination, sort, additionalParams) {
+      all: function (filters, pagination, sort, cache) {
```
igorpavlov: Where is this method currently used? Anywhere `additionalParams` passed?

reneolivo: The relationship model is only being used in the PCHR-3369 epic branch. You can find the only usage here: https://github.com/compucorp/civihr-tasks-assignments/blob/7f8e6aa1630dfb3451577c015bd9de00018aff60/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.controller.js#L363-L373

The `additionalParams` was not being used and was only added to follow the order of the `getAll` method of the API file.

igorpavlov: OK!

PR #2584, opened 2018-04-26
PCHR-3570: Footer Updates

mickadoo:
## Overview

The footer text should be updated to show the licence version and the site URL instead of "Project Website".

We will also restrict the site status to only root users.

## Before

* The licence was "Powered by CiviHR version 1.7.5.  CiviHR is openly available under the GNU AGPL License and can be downloaded from the Project website ."
* Any user with access to CiviCRM could see the site status in the footer

#### As Root User

image

#### As Non-Root

image

## After

* The licence is "Powered by CiviHR 1.7.5. CiviHR is openly available under Version 3 of the GNU AGPL License and can be downloaded from www.civihr.org."
* Only root users can see the site status in the footer

#### As Root User

image

#### As Non-Root

image

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Page/View/Summary.tpl
+++ b/hrjobcontract/CRM/Hrjobcontract/Page/View/Summary.tpl
@@ -49,7 +49,7 @@
 
         {* CRM-12735 - Conditionally include the Actions and Edit buttons if contact is NOT in trash.*}
         {if !$isDeleted}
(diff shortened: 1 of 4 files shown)


mickadoo:
Not sure why Jenkins is showing this as in progress as it already [passed](https://jenkins.compucorp.co.uk/blue/organizations/jenkins/civihr/detail/PR-2584/2/pipeline) the build

PR #2587, opened 2018-04-26
PCHR-3581: Support multiple requests for the same day in leave calendar

igorpavlov:
## Overview

This PR adds a support for multiple requests in Leave Calendar. The possible same day leave requests combinations are described here: https://github.com/civicrm/civihr/pull/2575.

## After

_Note: the same set of requests is used here as seen in Before gif_

### Components Support

Previously, the data about leave requests was passed and generated by Day component like that:

```js
// leave request data
vm.contactData.leaveRequest; // passed from Month component
// data for UI
vm.contactData.isAccruedTOIL; // generated by Day component, as an example
```

As we now can have more than one leave requests in a day, we need to restructure the data:

```
// leave requests data
vm.contactData.leaveRequests // [LeaveRequestInstance, LeaveRequestInstance ...]
// data for UI to show in the tooltip (all requests)
vm.contactData.leaveRequestsAttributes // { leaveRequestId: { labels, times etc... }, ... }
// data for UI to show in the day cell
vm.contactData.leaveRequestsToShowInCell // if there are TOIL and non-TOIL requests, show only TOIL requests
```

### UI

All properties were preserved:

```html
<div class="chr_leave-calendar__day-container"
  ... 
  ng-class="{
    'chr_leave-calendar__day-container--type--weekend': day.contactData.isWeekend,
    'chr_leave-calendar__day-container--non-working': day.contactData.isNonWorkingDay,
    'chr_leave-calendar__day-container--public-holiday': day.contactData.isPublicHoliday
  }">
  <div class="chr_leave-calendar__item"
    ...
    ng-hide="day.contactData.leaveRequests.length >= 3 && leaveRequestIndex > 0"
    ng-class="{
      'chr_leave-calendar__item--half': day.contactData.leaveRequestsToShowInCell.length >= 2,
      'chr_leave-calendar__item--requested': leaveRequestAttributes.isRequested,
      'chr_leave-calendar__item--alt': leaveRequestAttributes.isAccruedTOIL
    }">
```

## Comments

Please note, the tooltip is not functioning at the moment as it should, this will be made in a separate PR.

Tooltips showing logic is now moved to UI, this will be improved in the next PR.

------
✅Manual Tests - passed
✅Jasmine Tests - passed
✅JS distributive files - included
✅Backstop tests - passed
✅CSS distributive files - included
⏹PHP Unit Tests - not needed

PR #2590, opened 2018-04-27
PCHR-3581: Improve tooltips for multiple requests for same day in calendar + fix various bugs

igorpavlov:
## Overview

This PR is a continuation to https://github.com/civicrm/civihr/pull/2587. It styles tooltips in the Leave Calendar, remains them shown until unhovered, makes their inner items clickable and by click they now open a leave request they correspond to.

### Making tooltip stay shown until unhovered

This is an ad-hoc solution, it allows to unhover the day cell and at the same time leave the tooltip shown if it is hovered.

```js
function toggleTooltip (sourceElement, isHovered) {
  $timeout(function () {
    vm.tooltip[sourceElement + '_hovered'] = isHovered;

    vm.tooltip.show =
      vm.tooltip.day_cell_hovered || vm.tooltip.tooltip_hovered;
  }, isHovered ? 0 : 100);
}
```

```html
<div class="chr_leave-calendar__day-container"
   ...
  ng-mouseenter="day.toggleTooltip('day_cell', true)"
  ng-mouseleave="day.toggleTooltip('day_cell', false)"

...
<script type="text/ng-template"
    id="chr_leave-calendar__day-tooltip">
  <div
      ng-mouseenter="day.toggleTooltip('tooltip', true)"
      ng-mouseleave="day.toggleTooltip('tooltip', false)">
```

### Styling tooltip

This is also an ad-hoc solution until it is decided to make all tooltips like that.

```css
.chr_leave-calendar__day-tooltip {
  &.tooltip.in {
    opacity: 1;
  }

  .tooltip-arrow {
    border-top-color: $crm-white !important;
  }

  .tooltip-inner {
    background-color: $crm-white;
    border-radius: 0;
    box-shadow: $chr-leave-calendar-tooltip-shadow;
    color: $gray-darker;
    padding: 0;
  }
}

...

.chr_leave-calendar__day-tooltip_item {
  min-width: $chr-leave-calendar-tooltip-item-min-width; // 154px, this was agreed with the designer
  padding: $crm-gap-medium;

  &:hover {
    background-color: $gray-light;
  }
}
```

### Tooltip problem in SSP

The problem with the new style tooltip is that the container (table) has `overflow-x: auto` which allows users to scroll it horizontally. Unfortunately, this overflow the tooltips if they are attached to the table, so they must be attached to the body instead. However, this creates a problem with styling since CiviHR L&A does not have yet outside namespace SASS folder. A new folder with the tooltips styling was created to support styling tooltips in both Admin and SSP.

## Comments

- This PR also fixes the issue with Leave in Hours, it is a regression due to PCHR-3427 epic.
- This PR also fixes the issue with closing stacked modals

### Failing PHP Tests

This PR also fixes the issue in PHP tests. 

In the failing code it was intended to set a `from_date` of a leave request to the closest Saturday day, and `to_date` to the closest Sunday day. The test was fine, however, on Sundays the following thing happened:

```php
// If it is Sunday today, then:
CRM_Utils_Date::processDate('saturday') // yields NEXT Saturday, which is 6 days from TODAY
CRM_Utils_Date::processDate('sunday') // yields TODAY
```

This made test fail because a leave request cannot be created with a `from_date` greater than `to_date`. The problem was fixed by hardcoding dates for Sunday and Saturday since test should not care what day is today for this particular test.

----
✅Manual Tests - passed
✅Jasmine Tests - passed
✅JS distributive files - to be added once PR is approved
✅CSS distributive files - to be added once PR is approved
✅Backstop tests - passed
✅PHP Unit Tests - fixed and passed


org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
```diff
@@ -20,38 +20,23 @@ define([
     function open () {
       var modalInstance;
       var elements = [
```
reneolivo: why not

```js
var elements = $document.find('body, html');

elements.addClass('chr_scroll-lock');

// ...

  // ...

  elements.removeClass('chr_scroll-lock');
});
```

reneolivo: Are you sure this solution works in IE11? These lines that are being deleted were added to fix a specific IE11 issue:
https://github.com/civicrm/civihr/pull/2400

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -146,8 +146,14 @@ define([
      * @param  {LeaveRequestInstance} leaveRequest
      */
     function deleteLeaveRequest (leaveRequest) {
-      removeLeaveRequestFromIndexedList(leaveRequest);
-      updateLeaveRequestDaysContactData(leaveRequest);
+      var indexedLeaveRequest = findIndexedLeaveRequest(leaveRequest);
```
reneolivo: what's the difference between a `leaveRequest` and an `indexedLeaveRequest` ? Why isn't this handled at the `removeLeaveRequestFromIndexedList` or `updateLeaveRequestDaysContactData` level?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -169,6 +175,24 @@ define([
...
+      var indexedLeaveRequest;
+
+      _.forEach(leaveRequests[leaveRequest.contact_id], function (day) {
+        indexedLeaveRequest = _.find(day, function (leaveRequestObj) {
+          return +leaveRequestObj.id === +leaveRequest.id;
+        }) || indexedLeaveRequest;
+      });
+
+      return indexedLeaveRequest;
+    }
```
reneolivo: In case this is necessary, why not

```js
var indexedLeaveRequest;

// By using `find` here the loop will stop as soon as an indexedLeaveRequest is found:
_.find(leaveRequests[leaveRequest.contact_id], function (day) {
  indexedLeaveRequest = _.find(day, function (leaveRequestObj) {
    return +leaveRequestObj.id === +leaveRequest.id;
  });

  return indexedLeaveRequest;
});

return indexedLeaveRequest;
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -542,6 +577,13 @@ define([
       }
     }
 
+    function sortLeaveRequests (leaveRequests) {
```
reneolivo: `sortLeaveRequestsByDate`

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-calendar-day.html
```diff
@@ -1,31 +1,37 @@
...
       'chr_leave-calendar__item--half': day.contactData.leaveRequestsToShowInCell.length >= 2,
-      'chr_leave-calendar__item--requested': leaveRequestAttributes.isRequested,
-      'chr_leave-calendar__item--alt': leaveRequestAttributes.isAccruedTOIL
+      'chr_leave-calendar__item--requested': day.contactData.leaveRequestsAttributes[leaveRequest.id].isRequested,
+      'chr_leave-calendar__item--alt': day.contactData.leaveRequestsAttributes[leaveRequest.id].isAccruedTOIL
     }">
     <span ng-if="leaveRequest">
       <span class="chr_leave-calendar__item__label">
-        {{leaveRequestAttributes.label}}
+        {{day.contactData.leaveRequestsAttributes[leaveRequest.id].label}}
```
reneolivo: it would be nice if we had some sort of object that contains the `leaveRequest` and the `leaveRequestAttributes` so we could do something like

```html
<span class="chr_leave-calendar__item__label">
  {{leaveRequest.attributes.label}}
</span>
```

But again, this is just a suggestion, if it's too complicated, leave it.

org.civicrm.reqangular/src/common/decorators/ui-bootstrap/uib-modal.decorator.js
igorpavlov: Done, also improved tests.

igorpavlov: This solution is now pure CSS, so no more replaces, no more hardcoded attributes.

igorpavlov: image
left a comment to QA ^ it works for me in IE11

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: Very good question. This is because `updateLeaveRequestDaysContactData`, imagine you update a leave request `5 May - 10 May` to `4 May - 9 May`. It needs to flush days `5 - 10` and then add for `4 - 9`. We delete indexed request and not the one that was passed in order to flush the `10th May` as well. Then we add `4 May - 9 May`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: Thanks a lot. Done.

uk.co.compucorp.civicrm.hrleaveandabsences/views/shared/components/leave-calendar-day.html
igorpavlov: This is not complicated, the reason is a bit more strong - we do not want to contaminate `leaveRequest` object since it is then passed to other components such as Leave Request Modal. I am not personally a fan of passing the whole object, I would rather pass just ID, but this is how it is done and, I guess, for performance (no need to load leave request again). So we don't really want to change the way it works now and do not want to contaminate LR object at the same time. I agree, this is not really beautiful in the code, I tried to use `ng-init` to store the reference, but `ng-init` does not update later if we modify props of the referenced object, thus UI did not update as well.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
igorpavlov: Not really, not only date/time matter, but also if it is AM/PM.

PR #2112, opened 2017-09-04
PCHR-2430: Fix leave requests action labels

igorpavlov:
## Overview

The Leave Request Action Matrix has been updated. This PR makes the leave request actions comply with it.

## Technical Details

The most major change is amending the following function:

```js
function canLeaveRequestBeCancelled (leaveRequestStatus) {
  var allowCancellationValue = vm.absenceTypes[vm.leaveRequest.type_id]
    .allow_request_cancelation;

  // Admin can always cancel
  if (vm.role === 'admin') {
    return true;
  }

  // Others but Admin can cancel if only the leave request status is
  // either "Awaiting for Approval" or "More Information Required"
  if (!_.includes([statusNames.awaitingApproval, statusNames.moreInformationRequired],
    leaveRequestStatus)) {
    return false;
  }

  // Manager can cancel disregarding the allow_request_cancelation value
  if (vm.role === 'manager') {
    return true;
  }

  // If request can only be cancelled in advance of start date
  if (allowCancellationValue === '3') {
    return moment().isBefore(vm.leaveRequest.from_date);
  }

  // If request can always be cancelled
  return allowCancellationValue === '2';
}
```

## Comments

This PR also fixes the issue with the Filter by Assignee in the Admin Leave Requests. The issue was not with the component itself, but in its HTML template, where the wrong value was set on click. You can notice in the Before GIF the "All" is not selected, but in the After GIF it is.

BackstopJS tests are added because the Filter by Assignee is a specific feature for Admin only.

---

- [x] Tests Pass

PR #2595, opened 2018-04-30
PCHR-3634: Allow staff to cancel an approved/rejected leave request 

AkA84:
## Overview
Staff members were not able to cancel an already approved or rejected leave request, even when they were allowed. This is a bug that was introduced as part of https://github.com/civicrm/civihr/pull/2112, which defined an incorrect action matrix

## Before
before

The action matrix, related to the "cancel" action was
**Admin** can cancel in any case
**Manager** can cancel in any case, but only when the leave request is in "Awaiting for Approval" or "More Information Required"
**Staff** can cancel only when the leave request is in "Awaiting for Approval" or "More Information Required" *and* it's either always allowed to cancel a request or if it's allowed to do it before its start date

## After
after

The action matrix, related to the "cancel" action is
**Admin** can cancel in any case
**Manager** can cancel in any case, but only when the leave request is in "Awaiting for Approval" or "More Information Required"
**Staff** can cancel only if it's always allowed to cancel a request or if it's allowed to do it before its start date

## Technical details
This PR also contains a fix that allows both the future and current balance to be updated on-the-fly (= without reloading) whenever an approved leave request got cancelled.

Before the fix only the current balance would get updated, resulting in the future balance to be wrong, stuck with the old value


igorpavlov:
@AkA84 fixed conflicts

PR #2600, opened 2018-05-01
FAR-248: Show balance changes and individual request values in "Expired" section

tunbola:
## Overview
The balance-change-by-absence-types were missing in the "Expired" section's header (= it was always showing 0), and the values of the individual expired TOIL requests were missing as well when the sections was open. This PR fixes both issues

## Before
before

## After
after

## Technical Details (Backend)
The LeaveRequest.get and the LeaveRequest.getFull API both supports the expired parameter. When the parameter is TRUE, It returns all Leave requests linked to expired balance changes. When the parameter is FALSE it is supposed to return Leave requests linked to Non-Expired balance changes but right now, It returns all Leave Requests whether expired or not.

### Before
- When the Expired parameter is FALSE, the LeaveRequest.get and the LeaveRequest.getFull API returns both expired and non expired Leave requests.

### After
- When the Expired parameter is FALSE, the LeaveRequest.get and the LeaveRequest.getFull API returns only  Leave requests linked to non expired balance changes.
- When the Expired parameter is not present, the LeaveRequest.get and the LeaveRequest.getFull API returns all Leave Requests.

### Additional info
- Changes were made to the `CRM_HRLeaveAndAbsences_API_Query_LeaveRequestSelect` and an extra query condition was added for when the Expired parameter is FALSE.
These are the criteria used to determine Leave Requests linked to non-expired balance changes:

1. Leave requests linked to balance changes with null expiry dates (i.e balance changes that don't expire)
2. Leave requests linked to balance changes with expiry date greater than or equal to today.
3. Leave requests linked to balance changes that have an expiry record created for it but the amount expired is Zero. (This means that the leave contact was able to use up the balance before it expired and the expiry record created will have a value of Zero.)

### Changed `LeaveRequest`'s `LeaveRequestAPI`'s `balanceChangeByAbsenceType` signature
The `balanceChangeByAbsenceType()` method, both in `LeaveRequest` and `LeaveRequestAPI`, had a very verbose and strict signature, listing each parameter separately
```js
// LeaveRequest
balanceChangeByAbsenceType: function (contactId, periodId, statuses, isPublicHoliday) {
  return leaveRequestAPI.balanceChangeByAbsenceType(contactId, periodId, statuses, isPublicHoliday);
}

LeaveRequest.balanceChangeByAbsenceType('1', '2', [ 'approved', 'pending' ], true);
```
this made the method very obscure when being called (given that one has to know exactly what each parameter means) and very hard to maintain and expand upon, given that every time a parameter had to be added (like in this PR), the signature in the method of both services had to be updated, and a manual mapping between the new method's parameter and the API endpoint parameter had to be done.

This PR gets rid of that signature and allows the passing of a single object, that represents directly the parameters that will be passed to the API endpoint itself.

```js
// LeaveRequest
balanceChangeByAbsenceType: function (filters) {
  return leaveRequestAPI.balanceChangeByAbsenceType(this.processFilters(filters));
}

LeaveRequest.balanceChangeByAbsenceType({
  contact_id: '1',
  period_id: '2',
  statuses: {
    in: ['approved', 'pending']
  },
  public_holiday: true
});
```
this makes the method much easier to read when called (all the parameters are explained by their keys) and given that there is no manual mapping between the params given to the model and the params passed to the api endpoint, if one needs to add a new parameter he can simply add it to the method call, without having to add that parameter inside the method's implementation as well

### Fetched balance changes for "expired" section
Thanks to the above refactoring, fetching the balance changes for the "expired" section required only a simple additional api request
```js
function loadBalanceChanges () {
  return $q.all([
    // ...
    LeaveRequest.balanceChangeByAbsenceType(_.assign({}, basicParams, {
      expired: true
    })),
    // ...
  ])
    .then(function (results) {
      vm.absenceTypes.forEach(function (absenceType) {
        absenceType.balanceChanges = {
          // ...
          expired: results[1][absenceType.id],
          // ...
        };
      });
    });
}
```

### Refactored normalization of expired TOIL request objects
The reason why the TOIL requests were missing the value inside the "expired" section is that in that section there are two types of objects being rendered: entitlement breakdown and toil requests. Those objects have a different structure that needs to normalized, as the view expects them to have three key properties: `type.label`, `expiry_date`, and `amount`
```html
<!-- (simplified) -->
<tr ng-repeat="entry in report.sections.expired.data track by $index">
  <td>
    {{entry.type.label}}
    <span>
      (expired {{entry.expiry_date | formatDate}})
    </span>
  </td>
  <td ng-repeat="absenceType in report.absenceTypes track">
    {{ +entry.type_id === +absenceType.id ? entry.amount : '' }}
  </td>
</tr>
```

Those properties are already there on the breakdown object, so the TOIL request objects needed to be normalized in order to have them as well. This was already happening, but not fully
```js
function processExpiredTOILS (list) {
  return $q.resolve()
    .then(function () {
      return list.map(function (listEntry) {
        return {
          // missing the `amount` property
          'expiry_date': listEntry.toil_expiry_date,
          'type': {
            'label': 'Accrued TOIL'
          }
        };
      });
    });
}
```
The normalization logic had been refactored so that all the expected data is there
```js
function processExpiredTOILS (toils) {
  return $q.resolve()
    .then(function () {
      return toils.map(function (toil) {
        return _.assign({}, toil, {
          expiry_date: toil.toil_expiry_date,
          amount: toil.toil_to_accrue,
          type: {
            'label': 'Accrued TOIL'
          }
        });
      });
    });
}
```
(notice that the requests objects are simply enhanced now, and not fully replaced as it was done before)

### Refactored the absence types processing logic
The absence types displayed in the report component are not *all* the absence types available in the system, but rather only the absence types that match a criteria based on their entitlement value and settings
```js
// Only absence types that have an entitlement greater than 0
// or that allow overuse or accrual request
!((absenceType.entitlement === 0) &&
  (absenceType.allow_overuse !== '1') &&
  (absenceType.allow_accruals_request !== '1'));
```
Linking the absence types to their entitlements and filtering them was being done as part of the `loadEntitlements()` function, thus overloading the function with too much logic, and in particular with some logic that had nothing to do with the stated goal of the function (= loading the entitlements)
```js
function loadEntitlements () {
  return Entitlement.all({
    contact_id: vm.contactId,
    period_id: vm.selectedPeriod.id
  }, true)
    .then(function (entitlements) {
      vm.entitlements = entitlements;
    })
    .then(function () {
      vm.absenceTypesFiltered = vm.absenceTypes.filter(function (absenceType) {
        // link & filter
      });
    });
```
The "link & filter" logic had been refactored and isolated, so that it still happens after the entitlements have been loaded, but separately.
```js
.then(function () {
  return $q.all([
    loadEntitlements(),
    loadBalanceChanges()
  ]);
})
.then(processAbsenceTypes) // <--
.then(function () {
  vm.loading.content = false;
});

function processAbsenceTypes () {
  attachEntitlementsToAbsenceTypes();
  filterAbsenceTypes();
}
```


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -1185,8 +1185,8 @@ public function testGetAndGetFullReturnAllLeaveRequestsWhenTheExpiredParamIsNotP
       ['period_start_date' => '2016-01-01']
     );
 
-    // This request has 3 days expired, but will be included on
-    // the response anyway, since the "expired" flag is not set
+    // This request has 3 days expired, but will not included on
```
ajesamson: Guess a `be` is missing in the comment. 
`but will not be included on`

Kindly check the  `set/false` on the next line as well.

ajesamson:
A full stop would have been more appropriate for some places with comma. Kindly see below:
`The LeaveRequest.get and the LeaveRequest.getFull API both supports the expired parameter. When the parameter is TRUE, It returns all Leave requests linked to expired balance changes. When the parameter is FALSE it is supposed to return Leave requests linked to Non-Expired balance changes but right now, It returns all Leave Requests whether expired or not.`

In the technical task, i think you wanted to type `these` instead of `there`. Please see below:
`These are the criteria used to determine Leave Requests linked to non-expired balance changes:`

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -92,6 +93,25 @@ define([
...
     }
 
+    /**
+     * Attaches the entitlement information to the absence type it belongs to
+     * If there is no entitlement for a given absence type, a default
+     * entitlement object is assigned instead
+     */
+    function attachEntitlementsToAbsenceTypes () {
+      vm.absenceTypes = vm.absenceTypes.map(function (absenceType) {
+        var entitlement = _.find(vm.entitlements, function (entitlement) {
```
reneolivo: This would run a little faster:

```js
function attachEntitlementsToAbsenceTypes () {
  var entitlements = _.indexBy(vm.entitlements, 'type_id');

  vm.absenceTypes = vm.absenceTypes.map(function (absenceType) {
    var entitlement = entitlements[absenceType.id];

     // ...
  });
}
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -116,6 +136,18 @@ define([
         });
     }
 
+    /**
+     * Filters the absence types, keeping only the ones that either have
+     * an entitlement greater than 0 or that allow overuse or accrual requests
+     */
+    function filterAbsenceTypes () {
```
reneolivo: `removeAbsenceTypesThatCantBeEntitled` is not the greatest name, but `filterAbsenceTypes` is barely explaining what the function is doing.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -417,6 +445,14 @@ define([
         });
     }
 
+    /**
+     * Process the list of absence types objects by augmenting and filter them
+     */
+    function processAbsenceTypes () {
+      attachEntitlementsToAbsenceTypes();
+      filterAbsenceTypes();
```
reneolivo: why isn't this function used earlier? I don't think we need to load and attach entitlements, and load balance change for absence types that can't be entitled.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -534,6 +573,7 @@ define([
         loadEntitlements(),
         loadBalanceChanges()
       ])
+        .then(processAbsenceTypes)
```
reneolivo: even if this is used in two different places, I rather repeat this:

```js
// ...
  .then(attachEntitlementsToAbsenceTypes)
  .then(filterAbsenceTypes)
// ...
```
The reason being that I have no idea what `processAbsenceTypes`, but `attachEntitlementsToAbsenceTypes` and `filterAbsenceTypes` sound more concrete.

what do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -200,32 +232,43 @@ define([
...
+          }
+        })),
+        LeaveRequest.balanceChangeByAbsenceType(_.assign({}, basicParams, {
+          statuses: {
+            in: [
+              valueOfRequestStatus(sharedSettings.statusNames.awaitingApproval),
+              valueOfRequestStatus(sharedSettings.statusNames.moreInformationRequired)
+            ]
+          }
+        }))
```
reneolivo:
```js
var balanceChangeRequestFilters = {
  holidays: { public_holiday: true },
  expired: { expired: true },
  approved: { statuses: { in: [ valueOfRequestStatus(sharedSettings.statusNames.approved) ] } },
  pending: {
    statuses: {
      in: [
        valueOfRequestStatus(sharedSettings.statusNames.awaitingApproval),
        valueOfRequestStatus(sharedSettings.statusNames.moreInformationRequired)
      ]
    }
};

var balanceChangeRequests = _.transform(balanceChangeRequestFilters, function (promises, filters, requestName) {
  promises[requestName] = LeaveRequest.balanceChangeByAbsenceType(_.assign({}, basicParams, filters);
}, {});

$q.all(balanceChangeRequests)
  .then(function (results) {
    vm.absenceTypes.forEach(function (absenceType) {
      absenceType.balanceChanges = {
        holiday: results.holiday[absenceType.id],
        // etc.
      };
      // if we somehow can make it so its like this it would be even better:
      absenceType.balanceChanges = results[absenceType.id];
    });
  });

```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
AkA84: i'm really not sure how to call it...any name would be describing the full implementation and that would be too far

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
AkA84: yes you need to, because part of the filtering criteria is
```js
absenceType.entitlement === 0
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/staff-leave-report.component.js
AkA84: I was thinking about that, but there is also the argument that one can have that it would be better to handle anything related to the processing (enhance + filter, in this case) of absence types in a single wrapping function. Otherwise it could look like that `attachEntitlementsToAbsenceTypes`and  `filterAbsenceTypes` could be called independently from each other, but that's not the case (you can filter only after you attach)

PR #2577, opened 2018-04-19
PCHR-3162: Configurability Changes

mickadoo:
## Overview

This merges most of the features of the PCHR-3162 epic into staging. There are still ongoing changes, but due to time constraints they will not be included in the next release.

The following tickets are included:

|Ticket|PRs|
|------|---|
|PCHR-3353|https://github.com/civicrm/civihr/pull/2553|
|PCHR-3433|https://github.com/civicrm/civihr/pull/2522|
|PCHR-3392|https://github.com/civicrm/civihr/pull/2513<br>https://github.com/civicrm/civihr/pull/2519<br>https://github.com/civicrm/civihr/pull/2556|
|PCHR-3420|https://github.com/civicrm/civihr/pull/2524|
|PCHR-3354|https://github.com/civicrm/civihr/pull/2498|
|PCHR-3374|https://github.com/civicrm/civihr/pull/2520|
|PCHR-3375|https://github.com/civicrm/civihr/pull/2527<br>https://github.com/civicrm/civihr/pull/2523|
|PCHR-3376|https://github.com/civicrm/civihr/pull/2529|
|PCHR-3365|https://github.com/civicrm/civihr/pull/2503|
|PCHR-3460|https://github.com/civicrm/civihr/pull/2546|
|PCHR-3444|https://github.com/civicrm/civihr/pull/2562|
|PCHR-3506|https://github.com/civicrm/civihr/pull/2557|
|PCHR-3451|https://github.com/civicrm/civihr/pull/2532<br>https://github.com/civicrm/civihr/pull/2542<br>https://github.com/civicrm/civihr/pull/2550|
|PCHR-3464|https://github.com/civicrm/civihr/pull/2566|

PR #2602, opened 2018-05-03
PCHR-3649: Create "Activity Types" shortcut in navigation

mickadoo:
## Overview

Following the release of #2577 there has been no link to edit activity types for non root users. This PR re-adds a link to do that.

## Before

There is no link to edit activity types for non root users

## After

There is a link to edit activty types for non-root users


mickadoo:
Closing as this was decided to be unnecessary

PR #2604, opened 2018-05-09
PCHR-3654: Make it Impossible to Uncheck Work pattern As Default 

tunbola:
## Overview
Currently, It is possible to unselect a work pattern as default, leaving a CiviHR site without a default work pattern which breaks many things. This PR fixes the issue and makes sure that the default work pattern can not be unselected to make it non default, rather default work pattern change should happen from edit/create screen of a new work pattern (This is the current behaviour).

## Before
- It is possible to unselect a work pattern as default.

defaultworkpatternbefore

## After
- It is not possible to unselect the default work pattern and make it non-default

defaultworkpattern

## Note
This PR also contains changes from #2605 which is about making it impossible to set a disabled work pattern as default.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/WorkPattern.php
```diff
@@ -135,14 +135,23 @@ public static function validateDefaultWorkPattern($params) {
...
+    if($isToBeDisabled && $isDefaultWorkPattern) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException(
+        'You cannot disable the default Work Pattern'
+      );
+    }
+
+    $isToBeUncheckedAsDefault = isset($params['is_default']) && !$params['is_default'];
+    if($isToBeUncheckedAsDefault && $isDefaultWorkPattern) {
+      throw new CRM_HRLeaveAndAbsences_Exception_InvalidWorkPatternException(
+        'It is not possible to have no default Work Pattern'
```
ajesamson: Does this also implies `Default Work Pattern required.`? Guess it will be more simplified if it means same thing.

tunbola: Yes a default Work pattern is required. A default work pattern is actually shipped with the default CiviHR installation

PR #2614, opened 2018-05-18
PCHR-3516: Allow Contact to get Leave Requests For Other Contacts via API

tunbola:
## Overview
Based on the specification for the New Staff Calendar, A contact should be able to view other contact's leave requests via the API, This was not previously possible because of the Leave Request ACL rules allows a staff to view only his leave requests, a manager to only view leave requests of his managees while an administrator can view leave requests for all contacts.

## Before
- The former Leave Request ACL rules applies when a contact tries to get another contact's request via the API.

## After
- The Leave ACL query in the `addSelectWhereClause` function in the Leave Request BAO was modified to allow a contact get other contact's leave requests via the API irrespective of whether the contact is a staff, leave manager or Admin.

## Comments
- A couple of tests in the `api_v3_LeaveRequestTest` class was broken after the changes, Most of the broken tests were modified to suit the new ACL. Some tests were also merged to avoid redundancy.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
@@ -1726,7 +1726,7 @@ class api_v3_LeaveRequestTest extends BaseHeadlessTest {
     $this->assertNotEmpty($resultGetFull['values'][$leaveRequest1->id]);
   }
 
-  public function testGetAndGetFullShouldReturnEmptyResponseForALoggedInLeaveManagerWhenUnassignedIsTrue() {
+  public function testGetAndGetFullReturnResultsForUnAssignedContactForLoggedInLeaveManagerWhenUnassignedIsTrue() {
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4178,7 +4089,7 @@ public function testGetBreakdownShouldReturnEmptyIfTheGiveLeaveRequestDoesnExist
     $this->assertEmpty($result['values']);
   }
 
-  public function testGetBreakdownShouldReturnEmptyIfAStaffMemberTriesToGetTheBreakdownOfAnotherStaffMember() {
+  public function testGetBreakdownAlsoReturnsTheBreakdownOfAnotherStaffMemberWhenAStaffMemberTriesAccessingIt() {
```
davialexandre: Just confirming: We well hide this with the API Wrapper later, right?

tunbola: Yes, exactly

PR #2615, opened 2018-05-21
PCHR-3650: New Fields for Pingback

mickadoo:
## Overview

Data for the following new fields should be collected for each client site:

- number of job roles
- number of cost centres
- number of funders
- number of custom field groups
- number of custom fields
- number of case types
- number of HR resources
- number of HR resource types

## Before

Data for these fields was not collected or sent

## After

Data for these fields is collected and sent

## Technical Details

I was unaware of some of these definitions before so it might help the reviewer to know:

- Funder is a list of contact IDs that is part of the job roles table. Each job role can have multiple funders, which are separated by `|`
- HR Resource Type is Drupal content of a specific type. It is viewable from "HR Resources" from the SSP
- HR Resources are files attached to any of the HR Resource Types

## Notes

I did not include HR Resource entity count testing in this PR as it is primarily a Drupal entity and would involve a lot of Drupal function calls, making the coupling between CiviHR and Drupal even stronger

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
@@ -85,6 +85,7 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
     $entityCounts += $this->getRecruitmentEntityCounts();
     $entityCounts += $this->getJobRoleEntityCounts();
     $entityCounts += $this->getCustomDataCounts();
+    $entityCounts += $this->getCaseEntityCounts();
 
     return $entityCounts;
   }
@@ -311,6 +312,18 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
     $counts = [];
     $counts['customGroup'] = $this->getEntityCount('CustomGroup');
     $counts['customField'] = $this->getEntityCount('CustomField');
+
+    return $counts;
+  }
+
+  /**
+   * Gets counts for the case entities
+   *
+   * @return array
+   */
+  private function getCaseEntityCounts() {
+    $counts = [];
+    $counts['caseType'] = $this->getEntityCount('CaseType');
 
     return $counts;
   }
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
@@ -86,6 +86,7 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
     $entityCounts += $this->getJobRoleEntityCounts();
     $entityCounts += $this->getCustomDataCounts();
     $entityCounts += $this->getCaseEntityCounts();
+    $entityCounts += $this->getHrResourceCounts();
 
     return $entityCounts;
   }
@@ -324,6 +325,41 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
   private function getCaseEntityCounts() {
     $counts = [];
     $counts['caseType'] = $this->getEntityCount('CaseType');
+
+    return $counts;
+  }
+
+  /**
+   * Gets a count of all HR Resource related entities
+   *
+   * @return array
+   */
+  private function getHrResourceCounts() {
+    $counts['hrResourceType'] = 0;
+    $counts['hrResource'] = 0;
+
+    if (!$this->isDrupal()) {
+      return $counts;
+    }
+
+    $hrResourceTypeIds = db_select('node', 'n')
+      ->fields('n', ['nid'])
+      ->condition('n.type', 'hr_documents')
+      ->condition('n.status', NODE_PUBLISHED)
+      ->execute()
+      ->fetchCol();
+
+    $resourceTypeCount = count($hrResourceTypeIds);
+    $counts['hrResourceType'] = $resourceTypeCount;
+
+    if ($resourceTypeCount > 0) {
+      $hrResourceCount = (int) db_select('file_usage')
+        ->condition('id', $hrResourceTypeIds, 'IN')
+        ->countQuery()
+        ->execute()
+        ->fetchField();
+      $counts['hrResource'] = $hrResourceCount;
+    }
 
     return $counts;
   }
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
@@ -102,7 +102,7 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
    * @return int
    */
   private function getEntityCount($entity, $params = []) {
-    $params += ['is_deleted' => FALSE];
+    $params += ['is_deleted' => FALSE, 'is_active' => TRUE];
 
     return (int) civicrm_api3($entity, 'getcount', $params);
   }
@@ -273,10 +273,10 @@ class CRM_HRCore_Service_Stats_StatsGatherer {
    * @return array
    */
   private function getJobRoleEntityCounts() {
-    $recruitmentKey = 'com.civicrm.hrjobroles';
+    $jobRoleExtKey = 'com.civicrm.hrjobroles';
     $counts = [];
 
-    if (!ExtensionHelper::isExtensionEnabled($recruitmentKey)) {
+    if (!ExtensionHelper::isExtensionEnabled($jobRoleExtKey)) {
       return $counts;
     }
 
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
```diff
@@ -263,4 +267,101 @@ private function isDrupal() {
     return $userSystem instanceof CRM_Utils_System_Drupal;
   }
 
+  /**
+   * Fetches entity counts for the job role entities
+   *
+   * @return array
+   */
+  private function getJobRoleEntityCounts() {
+    $recruitmentKey = 'com.civicrm.hrjobroles';
```
davialexandre: s/recruitment/jobRoles/

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
```diff
@@ -263,4 +267,101 @@ private function isDrupal() {
...
+  /**
+   * Gets a count of all contacts in the system that have been marked as a
+   * funder
+   *
+   * @return int
+   */
+  private function getFunderCount() {
+    $result = civicrm_api3('HrJobRoles', 'get', ['return' => ['funder']]);
+    $result = array_column($result['values'], 'funder');
+    $funderIds = implode('', $result);
```
davialexandre: I had a look on staging and it seems that, at least when there's only one funder, the `funder` value will not end with a delimiter. So, if you have a result that looks like this:

```
$result[
    ['funder' => 1],
    ['funder' => 2],
    ['funder' => 3],
    ['funder' => 4],
    ['funder' => 5],
    ['funder' => 6],
];
```

Imploding the values will result in something like this: `123456` and then the explode will not work

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
```diff
@@ -263,4 +267,101 @@ private function isDrupal() {
...
+  }
+
+  /**
+   * Gets counts for the case entities
+   *
+   * @return array
+   */
+  private function getCaseEntityCounts() {
+    $counts = [];
+    $counts['caseType'] = $this->getEntityCount('CaseType');
```
davialexandre: This will also include disabled Case Types. Is this expected?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -106,6 +109,23 @@ public function testEntityCountsWillMatchExpectedCount() {
     $this->fabricateLeaveRequest($contactID);
     $this->fabricateLeaveRequest($contactID);
 
+    // expect 2 JobRoles
+    $params = ['contact_id' => $contactID];
+    $contract = CRM_Hrjobcontract_Test_Fabricator_HRJobContract::fabricate($params);
+    $params = ['job_contract_id' => $contract['id']];
+    CRM_Hrjobroles_Test_Fabricator_HrJobRoles::fabricate($params);
+    CRM_Hrjobroles_Test_Fabricator_HrJobRoles::fabricate($params);
```
davialexandre: Please use aliases for the HRJobContract and HrJobRoles fabricators

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -106,6 +109,23 @@ public function testEntityCountsWillMatchExpectedCount() {
...
+
+    // expect 1 + existing CostCenter
+    $existingCostCenterCount = civicrm_api3('OptionValue', 'getcount', [
+      'option_group_id' => 'cost_centres'
+    ]);
+    civicrm_api3('OptionValue', 'create', [
```
davialexandre: Please use a fabricator here

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -115,6 +135,10 @@ public function testEntityCountsWillMatchExpectedCount() {
     $this->assertEquals(1, $stats->getEntityCount('assignment'));
     $this->assertEquals(2, $stats->getEntityCount('document'));
     $this->assertEquals(2, $stats->getEntityCount('leaveRequest'));
+
+    $this->assertEquals(2, $stats->getEntityCount('jobRole'));
+    $this->assertEquals($costCenterCount, $stats->getEntityCount('costCenter'));
+    $this->assertEquals(2, $stats->getEntityCount('caseType'));
```
davialexandre: This test keeps growing and growing with each new entity count added to the pingback payload. Would it be too difficult to have separate tests for each entity?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -115,6 +135,10 @@ public function testEntityCountsWillMatchExpectedCount() {
...
   }
```
davialexandre: Please mention on the PR description that the tests don't cover HR Resources and why

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
mickadoo: That's interesting, I can see what you're talking about on staging too, but I tested this with multiple job roles (both with multiple funders and only single funders) and didn't see this.

On staging after adding a funder for an empty one:

`"funder": "|"` => `"funder": "|3|"`

And after editing one without a delimiter:

`"funder": "345"` => `"funder": "|345|"`

So it looks like delimiters are the expected approach. I'm not sure why staging has this kind of data or whether it's valid, any ideas? I can fix it anyway by using `|` as the glue for implode.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
mickadoo: Good point, `getEntityCount` filters by deleted but it would be good to include `is_active` there too.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
mickadoo: Good idea, will do

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Stats/StatsGatherer.php
davialexandre: I have no ideas why staging is like that, but it makes me wonder if there are any clients sites in the same situation. Using `|` as the glue is a good idea

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
davialexandre: Personally, I think that the bigger issue is the fact that we're using the CMS to store data. Besides the problem with the tests, this also introduces some other issues. For example, how can the sample data extension (which is a CiviCRM extension) create sample HR Resources?

Anyway, I don't think this will be fixed anytime soon and the value of adding tests to cover the HR Resources count is low compared to the amount of work that will be necessary. It's ok to not add tests for this.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -92,53 +108,85 @@ public function testEntityCountsWillMatchExpectedCount() {
...
+    $stats = $this->getGatherer()->gather();
+
+    $this->assertEquals(2, $stats->getEntityCount('leaveRequest'));
+  }
+
+  public function testJobRoleCountWillMatchExpectedCount() {
+    $contactID = ContactFabricator::fabricateWithEmail()['id'];
+
     $params = ['contact_id' => $contactID];
     $contract = CRM_Hrjobcontract_Test_Fabricator_HRJobContract::fabricate($params);
```
davialexandre: You forgot to use the alias here

PR #2617, opened 2018-05-23
PCHR-3702: Changes to leave-notification-badge

deb1990:
## Overview
As part of this PR, the `leave-notification-badge` component has been moved inside reqangular and renamed to `notification-badge`, and now it works for any kind of API, instead of only just leave request.

## Before - After
There is no change in UI. Its and internal change.

## Technical Details
1. `leave-notification-badge` component has been moved inside reqangular and renamed to `notification-badge`.

2.  Now this component works for any kind of API, instead of only just leave request. So the structure of filters has has changed, the new format is
```
[{
  apiName: '<APIName, for example 'Tasks', 'LeaveRequest'>',
  params: { <filters> }	
}, { < more apis can be specified here >}]
```


org.civicrm.reqangular/src/common/components/notification-badge.component.js
```diff
@@ -0,0 +1,60 @@
...
+     * Fetch count of records which matches the filter
+     *
+     * @return {Promise}
+     */
+    function fetchCount () {
+      var promise = [];
+
+      _.each(vm.filters, function (filter) {
+        promise.push(api.getAll(filter.apiName, filter.params, null, null, null, 'getFull', false));
+      });
```
reneolivo: Maps sounds for specific for this:

```js
var promises = _.map(vm.filters, function (filter) {
  return api.getAll(filter.apiName, filter.params, null, null, null, 'getFull', false);
});
```

org.civicrm.reqangular/src/common/components/notification-badge.component.js
```diff
@@ -0,0 +1,60 @@
...
+
+define([
+  'common/lodash',
+  'common/modules/components',
+  'common/services/api'
+], function (_, components) {
+  components.component('notificationBadge', {
+    bindings: {
+      filters: '<',
+      refreshCountEventName: '<'
```
reneolivo: just a question about implementation: why isn't this a string instead? will it change?
```html
refresh-count-event-name="some-event"
```

deb1990: Yes, this is a string which will be passed on from the html. That is because in a page the notification badge can be present multiple times. Suppose once for Leave count, also for tasks or documents or something else. And each of them needs to be refreshed only when their respective count has changed(update, create, delete etc.).

PR #2618, opened 2018-05-23
PCHR-3701: Selectively Modify Leave Request API Field return Value based on User Role and Leave ACL Permissions

tunbola:
## Overview
Based on the new requirements for the new Staff Calendar that allows a user to view leave requests for other user which was formerly impossible #2614 , Even though leave requests for other contacts can now be viewed, there are some sensitive/restricted fields that a user should not be able to see for contacts that he does not have access to. By having access to, I am referring to the Leave ACL rules which here means that a staff can only view these sensitive fields for his own leave requests only and for other's requests, a pre-determined configurable value is shown for these fields. For a manager he can only view these sensitive/restricted fields for the leave requests of his manages only while an Admin can view all fields for all contacts leave requests.

## Before
- A staff can view all fields for a leave requests that does not belong to the staff via the API
- A manager can view all fields for all leave requests of contacts that are not his managees via the API

## After
- A staff can only view true values for non restricted fields for a leave requests that does not belong to the staff via the API. For the restricted fields he only views a pre-determined value that is configurable.
- A manager can view can only view true values for non restricted  fields for all leave requests of contacts that are not his managees via the API. For the restricted fields he only views a pre-determined value that is configurable.

## Technical Details
- A new API wrapper class (`CRM_HRLeaveAndAbsences_API_Wrapper_LeaveRequestFieldsVisibility`) was created that filters these restricted field values based on handlers for each API action described in the `apiActions` property:

``` php
 protected $apiActions = [
    'getfull' => [
      'handler' => 'CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions',
      'request_requires_contact_id' => TRUE
    ],
    'get' => [
      'handler' => 'CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions',
      'request_requires_contact_id' => TRUE
    ],
    'getbreakdown' => [
      'handler' => 'CRM_HRLeaveAndAbsences_API_Handler_GetBreakDownFieldPermissions'
    ],
  ];
```
- The ACL permissions for a contact was implemented in restricting the field values.  
User with access to `administer leave and absences` i.e Admin Role was allowed access to all restricted fields. Users without this permission can only see restricted field values for the contacts they have access to. 

## Comments
- For the restricted fields, An empty string value was used for now till something else is decided. Also I restricted fields that I best thought were sensitive fields but since its configurable, more fields can be easily added and the values changed.
- A broken test [here](https://github.com/civicrm/civihr/blob/b77575ef21a8191d2ec2458bbc596f7d54b3bdf2/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveBalanceChangeTest.php#L995) was also fixed. The test only fails on first install of the test database and does not fail subsequently. The reason for the failure is that the source_id for the balance change fabricated is hardcoded to 100 and when one of the leave dates created in the test function has an ID of 100 in the database, the test fails because at that point it seems as if the leave request is linked to the two original balance changes plus the extra one whose source ID is hardcoded to 100. The issue was fixed by picking the next available source_id after the balance change have been created for the leave dates.

## API Handler Usage
The API handlers created in this PR are re-usable and the concept can also be extended to other extensions. 
Before creating a new handler for an API action, please check and make sure that the existing handlers can be re-used for the API action, if it can't be re-used, create a new handler extending the `CRM_HRLeaveAndAbsences_API_Handler_LeaveFieldPermissions` class which is the parent class for the handlers. 
Implement the required abstract methods after extending this class. Of important note are the following methods to be extended:
`getRestrictedFields()` : This method allows the extending class to provide API fields whose values are to be restricted for users depending on some logic that will be implemented in the child class.
A sample implementation of this method is 
```php
  protected function getRestrictedFields() {
    $fields = [
      'amount' => [
        'restricted_value' => '',
        'level' => 1
      ],
    ];

    return $fields;
  }
```
The restricted value is the value that will replace the original field value while the level property indicates the level the field is in the nested array result set.
`getNewFieldValue($field, $value, $dataIdentifier)` : This method allows the extending class to define a logic with which to replace the field value when passed to it. For example the extending class can decide that only admins will have access to the original value while other roles will see the restricted value.
The $dataIdentifier parameter allows the extending class define the field value that should be passed to the function when a field value is about to be replaced. For example the `CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions` class defines its Identifier key and level like below which means anytime a restricted field is to be replaced, the value of the contact_id is passed to the  getNewFieldValue function. This allows the class to make take decisions regarding the data.
```php
  /**
   * {@inheritdoc}
   */
  protected function getDataRowIdentifierKey() {
    return 'contact_id';
  }

  /**
   * {@inheritdoc}
   */
  protected function getDataRowIdentifierLevel() {
    return 1;
  }
```
The process method of the `CRM_HRLeaveAndAbsences_API_Handler_LeaveFieldPermissions` class contains the logic for processing the API result set and replacing restricted field values. This method should suffice for most API's, but can be overridden by extending class if need be.

There are also some cases where for example , some field must be present in the return parameters for the API result set  to be processed properly, For example for the Leave.get and Leave.getFull API's the contact_id parameter need to be included in the result set in order for the logic for replacing the field values to work properly, Please see how this was implemented [here](https://github.com/civicrm/civihr/blob/525cf84f7d77f6d0873b394a9960e89f98e39128/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestFieldsVisibility.php#L47) and [here](https://github.com/civicrm/civihr/blob/525cf84f7d77f6d0873b394a9960e89f98e39128/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestFieldsVisibility.php#L47)

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissions.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissions.php
@@ -6,10 +6,6 @@ class  CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions extends C
 
   use CRM_HRLeaveAndAbsences_ACL_LeaveInformationTrait;
 
-  const ADMIN_ROLE = 'Admin';
-  const MANAGER_ROLE = 'Manager';
-  const STAFF_ROLE = 'Staff';
-
   /**
    * @var array
    *   Stores the contact Id's that the current user has access to.
@@ -52,59 +48,35 @@ class  CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions extends C
    */
   private $restrictedFields = [
     'from_date_amount' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'to_date_amount' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'type_id' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'balance_change' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'toil_duration' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'toil_to_accrue' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'toil_expiry_date' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
     'sickness_reason' => [
-      'restricted_for' => [
-        self::MANAGER_ROLE => ['replace_with' => ''],
-        self::STAFF_ROLE => ['replace_with' => '']
-      ],
+      'restricted_value' => '',
       'level' => 1
     ],
   ];
@@ -117,7 +89,6 @@ class  CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions extends C
    * @param LeaveRequestRightsService $leaveRequestRights
    */
   public function __construct($apiRequest, LeaveRequestRightsService $leaveRequestRights) {
-    $this->setLoggedInUserRole();
     $this->leaveRequestRightsService = $leaveRequestRights;
     $this->apiRequest = $apiRequest;
   }
@@ -136,8 +107,7 @@ class  CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions extends C
    * @return array
    */
   protected function getAccessibleContacts() {
-    $isAdminUser = $this->loggedInUserRole === self::ADMIN_ROLE;
-    if(!$this->accessibleContacts && !$isAdminUser) {
+    if(!$this->accessibleContacts && !$this->currentUserIsAdmin()) {
       $this->accessibleContacts = $this->leaveRequestRightsService->getLeaveContactsCurrentUserHasAccessTo();
     }
 
@@ -148,21 +118,13 @@ class  CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissions extends C
    * {@inheritdoc}
    */
   protected function getNewFieldValue($field, $oldValue, $dataRowIdentifierValue) {
-    $isAdminUser = $this->loggedInUserRole === self::ADMIN_ROLE;
     $canAccessRestrictedData = $this->canAccessRestrictedData($dataRowIdentifierValue);
 
-    if($isAdminUser || $canAccessRestrictedData) {
+    if($this->currentUserIsAdmin() || $canAccessRestrictedData) {
       return $oldValue;
     }
 
-    $rolesRestrictedFor = array_keys($this->getRestrictedFields()[$field]['restricted_for']);
-    $inRestrictedRoles = in_array($this->loggedInUserRole, $rolesRestrictedFor);
-
-    if(!empty($inRestrictedRoles)) {
-      return $this->getRestrictedFields()[$field]['restricted_for'][$this->loggedInUserRole]['replace_with'];
-    }
-
-    return $oldValue;
+    return $this->getRestrictedFields()[$field]['restricted_value'];
   }
 
   /**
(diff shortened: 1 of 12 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestFieldsVisibility.php
```diff
@@ -0,0 +1,135 @@
...
+      'handler' => 'CRM_HRLeaveAndAbsences_API_Handler_GetBreakDownFieldPermissions'
+    ],
+  ];
+
+  /**
+   * @param $apiRequest
+   * @return mixed
+   */
+  public function fromApiInput($apiRequest) {
+    if ($this->canHandleApiInput($apiRequest)) {
```
davialexandre: Please explain on the dockblock what you're doing with the contact_id here and why

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Wrapper/LeaveRequestFieldsVisibility.php
```diff
@@ -0,0 +1,135 @@
...
+    $options = _civicrm_api3_get_options_from_params($apiRequest['params']);
+    $returnParams = array_keys($options['return']);
+
+    if (!empty($returnParams)) {
+      if (FALSE === array_search('contact_id', $returnParams)) {
+        $returnParams[] = 'contact_id';
+      }
+    }
+
+    return $returnParams;
```
davialexandre: We need to be a bit smarter here. If a request didn't include the contact_id in the return fields, it means they don't want that and will be quite confused when they see the contact_id among the returned fields. It's fine to add the fields to the return params, but when processing the api output we need to remove it in case it not specified

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissionsTest.php
```diff
@@ -0,0 +1,320 @@
...
+                    'date' => '2017-02-01',
+                  ],
+              ],
+          ],
+      ],
+  ];
+
+  public function setUp() {
+
+  }
```
davialexandre: Please remove the empty `setUp()` here

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestRightsTest.php
```diff
@@ -281,6 +282,38 @@ public function testCanCancelToilWithPastDatesReturnsTrueWhenAbsenceTypeAllowsPa
     $this->assertTrue($leaveRightsService->canCancelToilWithPastDates($this->leaveContact, $absenceType->id));
   }
 
+  public function testGetLeaveContactsCurrentUserHasAccessToWhenUserIsStaff() {
```
davialexandre: What is a "Leave Contact"? What is the difference between that and a regular Contact?

Also, for the test name, it would be better to use to describe what is the expected behaviour instead of naming the scenario. For example, this test could be named something like "testStaffMembersShouldOnlyHaveAccessToThemselves". The same apply for the other tests

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -921,11 +921,13 @@ public function testGetFullShouldNotIncludeTheBalanceChangeIfTheReturnOptionIsNo
     $expectedValues = [
       [
         'id' => $leaveRequest1->id,
-        'dates' => $this->createLeaveRequestDatesArray($leaveRequest1)
+        'dates' => $this->createLeaveRequestDatesArray($leaveRequest1),
+        'contact_id' => $leaveRequest1->contact_id
```
davialexandre: We cannot change the API output like that. As I said on another comment, we need to remove the contact_id from the response if it was not specified

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4541,6 +4610,50 @@ public function testTheTimeForFromAndToDateOfLeaveRequestIsAddedCorrectlyWhenLea
     $this->assertEquals($toDate->format('Y-m-d') ." 23:59:00", $leaveRequest->to_date);
   }
 
+  public function testGetAndGetFullReturnsDataForContactIdEvenWhenItsExcludedFromReturnParameters() {
```
davialexandre: This test will probably need to be removed

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissionsTest.php
```diff
@@ -0,0 +1,320 @@
...
+    $expectedParams['values'][5]['from_date_amount'] = '';
+    $expectedParams['values'][5]['to_date_amount'] = '';
+    $expectedParams['values'][5]['balance_change'] = '';
+    $expectedParams['values'][5]['type_id'] = '';
+    $genericFieldHandler->process($results);
+
+    $this->assertEquals($expectedParams, $results);
+  }
+
+  public function testProcessWhenUserIsManager() {
```
davialexandre: The only reason we have this test is because we allow different restricted values for managers and staff members. But do we really need that? After reading the code, I'm not so sure. Compare this test with the staff one for example. They are pretty much the same. The only difference is that you need to set the permission first, but the actual logic works based on what `getLeaveContactsCurrentUserHasAccessTo()` returns and the permission is not used there.

Also there's the problem that this is changing the concept of what is a "Leave Approver". Before a "Leave Approver" was anyone with a Leave Approver relationship with someone else. That it is that + someone who has the "manager leave request in SSP" permission. If we don't differentiate the restricted values for managers and staff members, we would be able to keep the old concept and not have to rely on the permission.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissions.php
```diff
@@ -0,0 +1,196 @@
...
+      'restricted_for' => [
+        self::MANAGER_ROLE => ['replace_with' => ''],
+        self::STAFF_ROLE => ['replace_with' => '']
+      ],
+      'level' => 1
+    ],
+    'type_id' => [
+      'restricted_for' => [
+        self::MANAGER_ROLE => ['replace_with' => ''],
+        self::STAFF_ROLE => ['replace_with' => '']
```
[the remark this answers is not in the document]
tunbola: The plan is to do this later when the generic value has been determined.

PR #2572, opened 2018-04-16
FAR-246: Create the LeaveBalanceChangeExpiryLog Entity

tunbola:
## Overview
Upon investigating an Incident related to expired TOIL for a staff on a client site, It was revealed that some balance changes were expired before their expiry date by the balance expiry scheduled job. The LeaveBalanceChangeExpiryLog Entity was created to allow logging of relevant information when a balance change is expired by the Scheduled job so that if such an incident happens again, it will be easier to debug and find out what exactly happened.

## Before
- The LeaveBalanceChangeExpiryLog Entity does not exist.

## After
- The LeaveBalanceChangeExpiryLog Entity with the following fields was created:

1. id
2. balance_change_id
3. amount
4. source_id
5. source_type
6. expiry_date
7. balance_type_id
8. leave_date
9. leave_request_id
10. created_date

The fields were added in such a way that if the expired balance or the leave dates linked to the expired balance was deleted, the leave request ID or the Entitlement ID(since these does not get deleted) can be used to identify the balance change depending on the type of balance it was.

PR #2625, opened 2018-05-25
PCHR-3514: Add New Field on Leave Types Form: "Hide leave type label on public calendar" 

tunbola:
## Overview
Based on the requirements for the New Calendar,  A new field is needed on the Absence type form page "Hide leave type label on public calendar"  that allows the control of the visibility of the Leave Type label on the frontend in the calendar and feeds.

## Before
- The "Hide leave type label on public calendar does not exist on the Absence Type form page.

## After
- To achieve the objective, a new  field `hide_label`(tinyint)  was added to the Absence Type DAO and the new field incorporated to the Absence type form.

hidelabelafter2

## Comments 
I made some minor manual changes to the LeaveBalanceChangeExpiryLog Entity on this [commit](https://github.com/civicrm/civihr/pull/2572/commits/612387a7a3ddd2b47ee79a6f02cb9f064829c3bb) without re-generating the DAO, so while I re-generated the DAO for the Absence Type Entity , the checksum got updated.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
```diff
@@ -132,6 +132,11 @@ private function addBasicDetailsFields() {
       $this->getDAOFieldAttributes('color'),
       TRUE
     );
+    $this->add(
+      'checkbox',
+      'hide_label',
+      ts('Hide leave type label on public calendars and feeds?')
+    );
```
davialexandre: I think it makes more sense to add the field after the existing ones (i.e. after "Enabled"). The reason is to not change the of the existing fields

PR #2626, opened 2018-05-28
PCHR-3515: My Leave changes

reneolivo:
## Overview
This PR introduces the following changes to My Leave:

* It displays a full list of staffs
* It displays the name of all the contacts
* It displays the filters for contacts
* It only displays staff taking leave by default, but it always displays the current logged in contact (unless filtered out)

## Before
my-leave-before
* No filters shown
* Only the current user is shown
* No names shown

## After
my-leave-after

### Leave Calendar Service

Since some functionality was already repeated between the leave calendar sub controller, these were grouped together into a separate Leave Calendar Service. The other alternative of moving these methods to the parent controller was not considered to avoid bloating it.

The leave calendar service needs to be initialized by passing the view-model containing the contact filters and lookup contacts list:

```js
var leaveCalendar = LeaveCalendarService.init(vm);
```

it returns an instance of the service bound to the provided view-model. This instance has the following methods:

**loadAllLookUpContacts**
```js
function loadAllLookUpContacts () {
  return Contact.all().then(function (contacts) {
    return contacts.list;
  });
}
```
**loadContactIdsToReduceTo**
```js
function loadContactIdsToReduceTo () {
  return loadContracts()
    .then(function (contracts) {
      var contractsInAbsencePeriod = contracts.filter(function (contract) {
        var details = contract.info.details;

        return (
          moment(details.period_start_date).isSameOrBefore(vm.selectedPeriod.end_date) &&
          (moment(details.period_end_date).isSameOrAfter(vm.selectedPeriod.start_date) ||
            !details.period_end_date)
        );
      });

      return _.uniq(contractsInAbsencePeriod.map(function (contract) {
        return contract.contact_id;
      }));
    });
}

function loadContracts () {
  return contracts ? $q.resolve(contracts) : Contract.all();
}
```

and **loadFilteredContacts**
```js
function loadFilteredContacts () {
  return Contact.all(prepareContactFilters(), null, 'display_name')
    .then(function (contacts) {
      return contacts.list;
    });
}

function prepareContactFilters () {
  return {
    department: _.get(vm, 'filters.userSettings.department.value', null),
    level_type: _.get(vm, 'filters.userSettings.level_type.value', null),
    location: _.get(vm, 'filters.userSettings.location.value', null),
    region: _.get(vm, 'filters.userSettings.region.value', null),
    id: {
      'IN': vm.filters.userSettings.contact
        ? [vm.filters.userSettings.contact.id]
        : vm.lookupContacts.map(function (contact) {
          return contact.id;
        })
    }
  };
}
```


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/services/leave-calendar.service.js
```diff
@@ -0,0 +1,120 @@
...
+            return contacts.list;
+          });
+      }
+
+      /**
+       * Returns a list of all contracts. The result is cached locally.
+       *
+       * @return {Promise}
+       */
+      function loadContracts () {
```
igorpavlov: Could you please sort functions alphabetically?

reneolivo:
I had to update the tests for the `leave-calendar.component` file because the staff controller changes modified the results of some of the tests.

I also made a mock controller in case filters are set to be hidden in case a new sub controller is added in the future it will retain the functionality of not fetching the options for the filters, thus reducing the requests to a minimum.

PR #2642, opened 2018-05-31
PCHR-3730: Default Location Types

mickadoo:
## Overview

We currently have a few different location types in CiviHR. This PR will reduce these to "Personal" and "Work"

## Before

On a default CiviHR Installation we had "Personal", "Home", "Work", "Correspondence" and "Billing"

## After

"Billing" location type is disabled. "Correspondence" and "Home" are deleted. "Personal" and "Work" are set as reserved.

All usages of other location types is removed or replaced with either of the two reserved types.

## Technical Details

Since CiviHR now requires the "Personal" location type it will be included in installation. Without this some tests will fail as they previously used the "Home" location type, and "Personal" was not included on installation.


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1017.php
```diff
@@ -102,4 +161,33 @@ private function up1017_convertLocationNamesToIds($locationTypeMap) {
     return $locationTypeMap;
   }
 
+  /**
+   * Looks up the location type ID for a given location type name
+   *
+   * @param $locationTypeName
```
tunbola: Missing param type

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1017.php
```diff
@@ -102,4 +161,33 @@ private function up1017_convertLocationNamesToIds($locationTypeMap) {
...
+    return (int) CRM_Utils_Array::value($locationTypeName, $locationIds);
+  }
+
+  /**
+   * Gets a list of location type names mapped to their database ID
+   *
+   * @return array
+   */
+  private function up1017_getLocationTypeIds() {
+    static $locationTypeIds = [];
```
tunbola: I think I like the use of static here. Would help avoid multiple API calls

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1017.php
```diff
@@ -46,6 +50,62 @@ private function up1017_migrateLocationTypes() {
...
+    ]);
+  }
+
+  /**
+   * Sets all given location types as reserved
+   *
+   * @param array $locationTypes
+   */
+  private function up1017_reserveLocationTypes($locationTypes) {
+    foreach ($locationTypes as $locationType) {
```
tunbola: A little bit of optimization, would be good to store the location id's in an array and then  make the API call to set them as reserved once. Not sure it would make much difference here though since we are not updating many records.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -152,6 +153,34 @@ private function deleteLocationTypes() {
...
     ]);
+
+  }
+
+  /**
+   * Creates location types required by CiviHR and makes them reserved
+   */
+  private function createRequiredLocationTypes() {
+    $locationTypesToCreate = [
+      'Personal',
```
tunbola: Since the Personal type is created on install, What about existing sites, is an upgrader doing this somewhere?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1017.php
mickadoo: Yeah, ideally things like this would be a class property, but it gets more difficult with traits as property names can conflict.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1017.php
mickadoo: I don't think the API can handle multiple updates (i.e. `'id' => ['IN' => [1,2]]`) and to be honest I avoid chaining whenever possible. I don't like the syntax, readability and idea that you don't have control over what happens at each stage (e.g. handling a failed get request when chained to create request). If you think it is necessary I can change it, but since we're just passing two location types here I'd prefer to leave it as it is.

tunbola: Its fine as is. Forgot the API doesn't allow IN for updates.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
mickadoo: Already answered this in chat, but just for posterity the "Personal" type is actually created from [buildkit](https://github.com/civicrm/civicrm-buildkit/blob/a4735eb44f89ef28ce3d2153a6084026c7804f77/app/config/hr17/install.sh#L27). Because some sites could have potentially deleted it we agreed that I would add an upgrader function to ensure the "Personal" location type exists and is in the right state.

PR #2643, opened 2018-05-31
PCHR-3725: Fix tooltips issues

igorpavlov:
## Overview

This PR fixes multiple problems with clickable/tappable tooltips:
1) It makes them work on Android
2) It fixes the inconsistency of their behaviour in iOS 

Additionally it fixes:
- Dates in the tooltips on iOS
- Makes legend badges hoverable (change style on hover)
- Expands legend by default

### Legend badges do not look clickable

kek

### iOS Tapping

soso

### iOS Dates

image

### Legend badges look clickable

leg2

### iOS dates

image

### Android issue

We need to listen not only to `touchend` event, but also to `touchcancel` event (see https://stackoverflow.com/questions/19088117/touchend-not-firing-after-touchmove) 

```js
$overlay.on('touchend touchcancel', function () {
  // ...
```

### iOS issue with taps

If you hide the $overlay element too quick, the click event will be propagated through. So adding 50 milliseconds (could be anything > 0, but safer to make 50) works.

```js
$overlay.on('touchend touchcancel', function () {
  toggleTooltip('source', true, 50);
  // ...
```

### iOS issue with dates
Apparently, iOS will not recognise `YYYY-MM-DD HH:mm:ss` format if passed in a Date constructor:

```js
// works in Chrome, but fails in Safari
new Date('2018-01-01 00:00:00')
```

so another method was used:

```js
// works in Chrome and  Safari
moment('2018-01-01 00:00:00').toDate()
```

---- 

✅Manual Tests - passed
⏹Backstop tests - omitted
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-legend.component.js
```diff
@@ -23,7 +23,7 @@ define([
     var vm = this;
 
     vm.absenceTypesToFilterBy = [];
-    vm.legendCollapsed = true;
+    vm.legendCollapsed = false;
```
reneolivo: ok, so the legends won't be collapsed by default.

PR #2647, opened 2018-06-01
PCHR-3754: BackstopJS improvements

AkA84:
## Overview
This PR introduces several different improvements to the CiviHR's BackstopJS suite

### Use of `class` for page objects
Backstop would reliably throw scripts and timeout errors any time parallel capture was attempted. After a lengthy investigation the cause was found to be the fact that the page objects were objects literal.

```js
module.exports = modal.extend({
  async openCompletedTasksModal () {},
  async openCreateNewTaskModal () {},
  async waitForReady () {}
});
```

This caused problems when multiple Chromium instances were running different engine scripts, as the scripts were literally using the same object at the same time.

Switching from an object literal approach to a class approach solved the problem, given that now each class needs to be "instantiated" before the engine script can use them and as a result each script is using different objects, which cannot interfere with each other even when the scripts are ran in parallel

```js
module.exports = class SSPTasks extends Page {
  async openCompletedTasksModal () {}
  async openCreateNewTaskModal () {}
  async waitForReady () {}
};
```

#### Note
A simpler approach was considered before using this class-based one, which consisted on cloning/copying the objects in the `extend` method of the main `page` page object
```js
extend (page) {
  const clone1 = _.clone(this);
  const clone2 = _.clone(page);

  return Object.assign({}, clone1, clone2);
},
```
but unfortunately it didn't really work, and I wasn't able to figure out why

### Common `SSP` page object
A paren `SSP` page had been created, that every ssp page object will inherit from.

```js
const Page = require('./page');

module.exports = class SSP extends Page {
  async waitForReady () {
    try {
      await this.puppet.waitFor('tasks-notification-badge .chr_leave_notification', { visible: true });
      await this.puppet.waitFor('manager-notification-badge .chr_leave_notification', { visible: true });
    } catch (ex) {}
  }
};
```

This allows to share a common wait logic, that makes backstop wait for the tasks and leave badges to appear in the main nav before proceeding

badges

### Raised engine's `waitTimeout` value
Another reason for why parallel capturing was extremely unreliable was the fact that CiviHR doesn't really perform very well when there are multiple instances of it open (ie. 5 CiviHR instances all trying to access the same tasks dashboard page). This would mean occasional very high wait times for the certain pages to load, which would result in Puppeteer timing out and those pages' scenarios to fail.

For this reason the engine's `waitTimeout` value had been raised to 5 minutes, as to allow the pages enough time to fully load even under heavy-load circumstances.

This had resulted in a significant increase of reliability when parallel capturing

### Added extra engine options to improve stability
More engine options had been added to the config file, that are improving the stability of the test suite

```json
"engineOptions": {
  "slowMo": 250,
  "ignoreHTTPSErrors": true,
  "args": [
    "--no-sandbox",
    "--disable-setuid-sandbox",
    "--disable-gpu",
    "--force-device-scale-factor=1",
    "--disable-infobars=true",
    "--hide-scrollbars"
  ]
},
```

### Added wait logic for modals
The basic `Modal` page object didn't have a shared wait logic applied to all modals (regardless of their content). The logic has now been added, and it simply wait for the `.modal-body` element to appear (which is usually hidden while a spinner is being shown)
```js
async waitForReady () {
  await this.puppet.waitFor('.modal-body', { visible: true });
}
```

### Increased arbitrary time for documents dashboard
This one unfortunately is an ad-hoc fix. Puppeteer seems to be unable to correctly wait for the staff member name to appear in the documents dashboard

documents

Several different selectors were used to no avail, like for example

```js
/* #1 */ await this.puppet.waitFor('.ct-table-documents [href^="/civicrm/contact/view"]', { visible: true }); 
/* #2 */ await this.puppet.waitFor('.ct-table-documents [href^="/civicrm/contact/view"]:empty', { hidden: true });
/* #3 */ await this.puppet.waitFor('.ct-table-documents [href^="/civicrm/contact/view"]:not(:empty)', { visible: true });
```
and the arbitrary 500ms wait amount that was in place before wasn't unfortunately reliable enough, so the amount had to be raised to 3000ms to make sure the staff name would appear before backstop takes the screenshot

*Note: this had to be done also for the Tasks dashboard*

### Use Puppeteer instead of Chromy to generate cookies
In the gulp task chromy was still used to generate the cookies
```js
chromy.chain()
  .goto(config.url)
  .goto(loginUrl)
  .getCookies()
  .result(function (cookies) {
    fs.writeFileSync(cookieFilePath, JSON.stringify(cookies));
  })
  .end()
  .then(function () {
    chromy.close();
    resolve();
  });
```
Given that now the test suite uses puppeteer, it makes sense to use puppeteer to generate the cookies as well

```js
let browser = await puppeteer.launch();
let page = await browser.newPage();
await page.goto(loginUrl);
let cookies = await page.cookies();

fs.existsSync(cookieFilePath) && fs.unlinkSync(cookieFilePath);
fs.writeFileSync(cookieFilePath, JSON.stringify(cookies));

await browser.close();
```

### Added support for `--skipCookies` argument
In case ones doesn't need the cookies to be generated (because they haven't expired yet, for example), then he can pass the `--skipCookies` argument to the gulp task to skip the cookies creation
```shell
gulp backstopjs:test --skipCookies
```

### Converted gulp task to ECMAScript 2017
The entirety of the backstopjs gulp task is now written in ECMAScript 2017 (the switch to puppeteer kinda forced that)

### Improved usage of separate scenario files
Before this PR, if a dev wanted to run only the scenarios contained in a single .json file under the `scenarios/` folder, it would run the task as such
```shell
gulp backstopjs:(reference|test) --configFile ta-documents.json
```
this would empty the reference/test folder and place in it only the screenshots related to, in this case, the scenarios of the `ta-documents.json`. Tthe problem with this approach was that, if the dev wanted later to run the full suite of scenarios, he would have to regenerate the reference files from scratch first.

This had been improved by following what https://github.com/compucorp/backstopjs-config does: each scenario file (or "group", from now on), when run individually, would store its screenshots in a separate folder. 

It works as follow
```shell
gulp backstopjs:reference
gulp backstopjs:test
```
When no group is specified, backstop places the screenshots in `screenshots/_all_/reference` and `screenshots/_all_/test`

```shell
gulp backstopjs:reference --group ta-documents
gulp backstopjs:test --group ta-documents
```
When a group is specified (notice that there is no need to specify the .json extension), then backstop places the screenshots in `screenshots/{name-of-group}/reference` and `screenshots/{name-of-group}/test`, so in this case: `screenshots/ta-documents/reference` and `screenshots/ta-documents/test`

This way each individual scenario file is truly treated as a sub-suite, independent from the overall suite


uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/documents.js
```diff
@@ -43,25 +43,21 @@ module.exports = page.extend({
...
    * Waits until the user name in the "Staff" column and the filter dates are visible
    */
   async waitForReady () {
     await this.puppet.waitFor('.ct-filter-date', { visible: true });
-    await this.puppet.waitFor('.ct-table-documents [href^="/civicrm/contact/view"]', { visible: true });
-    // For some reason Puppetteer considers the user name visible even when it
-    // isn't really yet, this slight delay allows the element to be fully visible
-    // before taking the screenshots
-    await this.puppet.waitFor(500);
+    await this.puppet.waitFor(4000);
```
igorpavlov: I know the explanation is mentioned in the description, but still, any reason why it doesn't show it in 500ms? Is there an animation or is it just Backstop instability?

uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/tabs/absence/calendar.js
```diff
@@ -1,7 +1,9 @@
...
+const SSPMyLeaveCalendar = require('../../ssp-leave-absences-my-leave-calendar');
 
-module.exports = sspMyLeaveCalendar.extend(tab).extend({
-  readySelector: '.chr_leave-calendar__month-body',
-  tabTitle: 'Calendar'
-});
+module.exports = class AbsenceCalendarTab extends SSPMyLeaveCalendar {
+  constructor () {
+    super(...arguments);
+    this.readySelector = '.chr_leave-calendar__month-body';
```
igorpavlov: Please add an empty line after `super` for more readability. Same for other constructors please.

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -1,64 +1,62 @@
...
+const clean = require('gulp-clean');
+const execSync = require('child_process').execSync;
+const file = require('gulp-file');
+const fs = require('fs');
+const gulp = require('gulp');
+const notify = require('gulp-notify');
+const path = require('path');
+const puppeteer = require('puppeteer');
+const Promise = require('es6-promise').Promise;
+
```
igorpavlov: Not sure we need this empty line.

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -1,64 +1,62 @@
...
+const utils = require('../utils');
+
+const BACKSTOP_DIR = path.join(__dirname, '..', '..', 'backstop_data');
```
igorpavlov: Please move `__dirname` to a separate variable, smth like `CURRENT_DIRECTIORY_PATH`

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -1,64 +1,62 @@
...
 });
 
 /**
- * Concatenates all the scenarios (if no specific scenario file is specified)
+ * Returns the list of the scenarios from
+ *   a. All the different groups if `group` is == '_all_',
+ *   b. Only the given group
  *
- * @param  {Object} usersIds
+ * @param {Array} usersIds
```
igorpavlov: Please align:

```js
* @param  {Array} usersIds
* @param  {String} group
* @return {Array}
```

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -1,64 +1,62 @@
...
+ * @param {String} group
  * @return {Array}
  */
-function buildScenariosList (usersIds) {
-  var config = siteConfig();
-  var dirPath = path.join(BACKSTOP_DIR, 'scenarios');
+function buildScenariosList (usersIds, group) {
```
igorpavlov: Maybe it would be better to say `groupName`?

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -94,9 +92,7 @@ function cleanUpAndNotify (success) {
 function constructScenarioUrl (siteUrl, scenarioUrl, usersIds) {
   return scenarioUrl
     .replace('{{siteUrl}}', siteUrl)
-    .replace(/\{\{contactId:([^}]+)\}\}/g, function (fullMatch, user) {
-      return usersIds[user].civi;
-    });
+    .replace(/\{\{contactId:([^}]+)\}\}/g, (__, user) => usersIds[user].civi);
```
igorpavlov: Would be even better if we can tell exactly what is expected:

.replace(/\{\{contactId:(\w+)\}\}/g, (__, user) => usersIds[user].civi);

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
```diff
@@ -174,37 +170,31 @@ function runBackstopJS (command) {
...
+      .on('end', async () => {
+        try {
+          (typeof argv.skipCookies === 'undefined') && await writeCookies();
+          await backstopjs(command, { configPath: FILES.temp, filter: argv.filter });
+
+          success = true;
+        } finally {
+          cleanUpAndNotify(success);
+
+          success ? resolve() : reject(new Error('BackstopJS error'));
```
igorpavlov: Could we provide a more specific error message?

uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/documents.js
AkA84: no, civihr's performance unfortunately. and it gets worse as we increase the number of screenshots captured in parallel

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
AkA84: it was already like this, just to visually separate the 3rd-party modules from our own custom one

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
AkA84: `__dirname` is already a constant, i don't see the benefit of doing that (and devs should already know what it means anyway)

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
AkA84: sorry, i don't understand what you mean

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
AkA84: unfortunately not, it needs to be generic as the error could be

1. Task error
2. Script errors
3. Comparison errors

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
igorpavlov: You may want to specify exactly what do you expect:

```js
.replace(/\{\{contactId:(\s*\d+)\}\}
// meaning matching only digits and optional spaces
```

But its ok to leave like it is.

uk.co.compucorp.civicrm.hrcore/gulp/tasks/backstopjs.js
igorpavlov: Please include this info in the error message.

PR #2648, opened 2018-06-01
PCHR-3519: If Absence Type Public Label is set to true, always include Absence Type field in Leave Requests

tunbola:
## Overview
On #2618 , Changes were made to restrict access to Leave request API field values based on the User and Leave ACL rules. This PR introduces an exception to the condition in which the type_id restricted field values is displayed based on the hide_label property on the Absence Type introduced in #2625 . The rule is that even if a user does not have access to other contact's restricted fields, For the type_id field, if the Absence Type label is set to Public, the user is allowed to view the type_id field values.

## Before
- A user can not view the field value for the type_id field of a contact he does not have access to.

## After
- A user can view the field value for the type_id field of a contact he does not have access to if the Absence Type label is set to public.

## Technical Details
- A new method that sets visible absence types was added to the `GenericLeaveFieldPermissions` class

```php
  private function setVisibleAbsenceTypes() {
    $activeAbsenceTypes = AbsenceType::getEnabledAbsenceTypes();
    $visibleAbsenceTypes = [];
    foreach ($activeAbsenceTypes as $absenceType) {
      if (!$absenceType->hide_label) {
        $visibleAbsenceTypes[] = $absenceType->id;
      }
    }

    $this->visibleAbsenceTypes = $visibleAbsenceTypes;
  }
```

A new check was added in the `getNewFieldValue` function to check if user should access the field value 
```php
    if ($field === 'type_id' && in_array($oldValue, $this->visibleAbsenceTypes)) {
      return $oldValue;
    }
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissionsTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissionsTest.php
@@ -241,7 +241,7 @@ class CRM_HRLeaveAndAbsences_API_Handler_GenericLeaveFieldPermissionsTest extend
 
   public function testProcessWillHideAccessibleFieldsWhenRowIdentifierIsAbsentEvenIfUserHasAccess() {
     $absenceType = AbsenceTypeFabricator::fabricate([
-      'hide_label' => 1
+      'hide_label' => 0
     ]);
     //User is Staff with ID of 204 and has full access to only his data
     $this->setPermissions();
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/API/Handler/GenericLeaveFieldPermissionsTest.php
```diff
@@ -232,6 +240,9 @@ public function testProcessWhenUserIsAdmin() {
   }
 
   public function testProcessWillHideAccessibleFieldsWhenRowIdentifierIsAbsentEvenIfUserHasAccess() {
+    $absenceType = AbsenceTypeFabricator::fabricate([
+      'hide_label' => 1
+    ]);
```
davialexandre: Wouldn't it make more sense to use `hide_label => 0` here, given the test is about checking that users will not be able to see something even if they have access to it?

tunbola: Good idea.

PR #2650, opened 2018-06-01
PCHR-3772: Fix calendar issues with tooltips and leave request modal

igorpavlov:
## Overview

This PR fixes 2 issues:
1) When you open a leave request from the Calendar, modify it, then open again - it has old data
2) When you filter by Absence Types, some leave requests tooltips are missing information

### Stale data in the Leave Request Modal

This is simply due to Leave Request API caching.

```js
// return this.sendGET('LeaveRequest', 'getFull', { id: id })
return this.sendGET('LeaveRequest', 'getFull', { id: id }, false)
```

### Disappearing data in tooltips

This is because Calendar Day component watches leave requests changes and if there are changes, it augments leave requests with additional data. However, if you filter requests out, the leave requests that were already there before filtering stay unchanged and the data will not be populated. In order to trigger that watch every time leave requests are reloaded we need to flush them from Days object.

```js
// the watcher
$scope.$watch('day.contactData.leaveRequests', function () {
  if (vm.contactData && vm.contactData.leaveRequests) {
    vm.contactData.leaveRequests.forEach(resolveLeaveRequestsAdditionalUIData);

// the loader
function loadMonthLeaveRequests () {
  // ...

  flushDays(); // <<< here

  return LeaveRequest.all(params, null, null, null, false)
    .then(...)
}
```

----
✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -607,6 +618,11 @@ define([
       (forceReload || !dataLoaded) && loadMonthData();
     }
 
+    /**
+     * Sorts leave requests by either date or AM/PM
+     *
+     * @param {Array} leaveRequests array of leave requests instances
+     */
     function sortLeaveRequests (leaveRequests) {
```
reneolivo: thank you!

PR #2651, opened 2018-06-02
PCHR-3766: Add managees filter to my leave

reneolivo:
## Overview

* It adds the managees filter on the My Leave calendar section.
* Managers and admins can manage other contacts' requests on My Leave, except their own.

### Before
my-leave-filter-before

### After
anim

### Technical Details

The `filtersByAssignee` property was moved from the leave calendar admin sub controller to the main leave calendar component.

To check if the filter can be shown a `canManageRequests` method was added to the leave calendar component:

```js
function canManageRequests () {
  return _.includes(['admin', 'manager'], actualUserRole);
}
```

Actual role, as the name implies, stores the actual role of the logged in user since this role can be overridden by the request:

```js
function setUserRole () {
  return $q.all([
    checkPermissions(sharedSettings.permissions.admin.administer),
    checkPermissions(sharedSettings.permissions.ssp.manage)
  ]).then(function (results) {
    actualUserRole = results[0] ? 'admin' : (results[1] ? 'manager' : 'staff');
    userRole = vm.roleOverride ? vm.roleOverride : actualUserRole;
  });
}
```

The filter is then shown using the `canManageRequests` method:

```html
<div class="col-md-12 col-lg-3 col-lg-offset-3" ng-if="calendar.canManageRequests() && calendar.showFilters">
  <!-- filter code here -->
</div>
```

### Before
anim

### After
anim

### Technical Details

The `isSelfRecord` property is used to determine if the user can manage the request or not. Previously this property was set to true when the current section was My Leave.

The condition was moved to the request controller file and was expanded so it's only true when the section is My Leave and the users is not creating or editing their own request:

```js
(function init () {
  // ...

  return loadLoggedInContactId()
    .then(initIsSelfRecord)
    .then(function () {
      return $q.all([
        initRoles(),
        loadAbsencePeriods(),
        loadStatuses()
      ]);
    })
    // ...
})();

function initIsSelfRecord () {
  var isSectionMyLeave = $rootScope.section === 'my-leave';
  var isMyOwnRequest = +loggedInContactId === +_.get(vm, 'leaveRequest.contact_id');
  var isNewRequest = !_.get(vm, 'leaveRequest.id');

  vm.isSelfRecord = isSectionMyLeave && (isMyOwnRequest || isNewRequest);
}
```

The `initRoles` method was modified so when the `isSelfRecord` property is true it behaves as if the user is a staff:

```js
function initRoles () {
  role = 'staff';

  if (vm.isSelfRecord) {
    return;
  }

  // ...
}
```

The `isSelfRecord` was moved from the leave popup service file to the request controller file because the later has all the information needed to determine the value of this property and also to avoid repeating the rule to set the value as it was done in many other files (`$rootScope.section === 'my-leave'`)

----
✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -27,7 +27,7 @@ define([
   function controller ($controller, $q, $log, $rootScope, sharedSettings, AbsencePeriod, AbsenceType, PublicHoliday, OptionGroup, checkPermissions) {
     $log.debug('Component: leave-calendar');
 
-    var subController, userRole;
+    var actualUserRole, subController, userRole;
```
igorpavlov: `actualUserRole` sounds a bit vague. Should we name it `userPermissionRole`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/shared/controllers/request.controller.spec.js
```diff
@@ -1323,6 +1323,101 @@
...
+          role = 'admin';
+          leaveRequest = LeaveRequestInstance.init({});
+        });
+
+        describe('when the section is My Leave', function () {
+          beforeEach(function () {
+            $rootScope.section = 'my-leave';
+          });
+
+          describe('and I am checking someone else\'s request', function () {
```
igorpavlov: and user is checking...

same for other places where you say "I"

reneolivo:
@igorpavlov thanks for the feedback. No dist files changes happened, in case you were wondering.

igorpavlov:
@rene I merged this branch as per the permission from you.

PR #2652, opened 2018-06-03
PCHR-3767: Fix staff filters

reneolivo:
## Overview
This PR fixes the following issue:

* As a staff, when going to My Leave, the staffs filter does not work.

## Before
anim

## After
anim

## Technical Details

The issue happened because the look up contacts property needs to be populated in order to for the filter to work. To fix this, the leave calendar service now exposes the `loadLookupContacts` method that it was previously using internally only:

```js
function loadLookUpContacts () {
  var filterByAssignee = _.get(vm, 'filters.userSettings.assignedTo.type', 'all');
  var loadLookUpContactsMethod = contactsLookUpStrategies[filterByAssignee];

  return loadLookUpContactsMethod();
}
```

Please refer to [this PR](https://github.com/civicrm/civihr/pull/2651) for more information about the contacts look up strategies.

The leave calendar staff sub controller was refactored so it uses this look up method when the viewing contact is a staff:

```js
loadContacts: function () {
  if (vm.displaySingleContact) {
    return leaveCalendar.loadFilteredContacts(); // only needs the information about the specific contact.
  } else if (vm.canManageRequests()) {
    return leaveCalendar.loadContactsByAssignationType(); // loads the look ups, but also contracts, and staff have no access to them
  } else {
    // loads contacts and look up contacts at the same time:
    return $q.all({
      contacts: leaveCalendar.loadFilteredContacts(),
      lookupContacts: leaveCalendar.loadLookUpContacts()
    })
      .then(function (result) {
        vm.lookupContacts = result.lookupContacts;

        return result.contacts;
      });
  }
}
```

## Comments
:warning: this PR depends on https://github.com/civicrm/civihr/pull/2651 and it's pointing to that PR's branch (PCHR-3766-add-managees-filter-to-my-leave). Once that PR is merged, the base branch for this PR will change to PCHR-3485-all-staff-leave-calendar

----
✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


igorpavlov:
@rene I changed the target branch and merged this branch as per the permission from you.

PR #2656, opened 2018-06-04
PCHR-3753: Fix My Leave calendar for managers

reneolivo:
## Overview
This PR fixes an issue where **managers** could not see the My Leave calendar page.

## Before
my-leave-error
Error message:
```json
{"error_code":"unauthorized","entity":"HRJobContract","action":"get","is_error":1,"error_message":"API permission check failed for HRJobContract/get call; insufficient permission: require administer CiviHR"}
```

### Managers
my-leave-manager

### Admins
my-leave-admin

### Staff
my-leave-staff

### Contact details on admin portal
staff-calendar

## Technical Details
The issue happened because the leave calendar staff was using the `loadContactsByAssignationType` method of the leave calendar service. This method only works for admin since internally it requests contracts to pass to the leave calendar month for further filtering.

To fix it, the `userPermissionRole` property of the leave calendar component was exposed as public and then the leave calendar staff sub controller was refactored as follows:

```js
loadContacts: function () {
  if (vm.displaySingleContact) {
    return leaveCalendar.loadFilteredContacts();
  } else if (vm.userPermissionRole === 'admin') {
    // renamed from loadContactsByAssignationType to avoid confusion:
    return leaveCalendar.loadContactsForAdmin();
  } else {
    return loadContactsForStaffAndManagers();
  }
}

function loadContactsForStaffAndManagers () {
  return $q.all({
    contacts: leaveCalendar.loadFilteredContacts(),
    lookupContacts: leaveCalendar.loadLookUpContacts()
  })
    .then(function (result) {
      vm.lookupContacts = result.lookupContacts;

      return result.contacts;
    });
}
```

----
✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar.component.js
```diff
@@ -41,6 +41,7 @@ define([
     vm.showContactName = false;
     vm.showFilters = false;
     vm.supportData = {};
+    vm.userpermi = 'staff';
```
igorpavlov: Could you please fix this one and add a test.

reneolivo: Ah sorry, I added a test for that:
https://github.com/civicrm/civihr/pull/2656/files#diff-333e6c3beacc230ca218ee76bc1006c3R124

The issue is that this property is overridden as soon as the controller is initialized. This definition is only here as a reference. Is it ok just to fix the name?

PR #2658, opened 2018-06-04
PCHR-3677: Allow The Value Field For TOIL  and Leave Days Amounts Option Values To Be Editable

tunbola:
## Overview
in #1776 , The value field for Option value was made non editable and non visible. For the `hrleaveandabsences_toil_amounts` and `hrleaveandabsences_leave_days_amounts` option groups, we need to make an exception to this logic as it is necessary for users to be able to edit the value fields for these option groups.

## Before
- The value field for all option values of all option groups is hidden and not editable

optionvaluebefore

## After
- The value field for the `hrleaveandabsences_toil_amounts` and `hrleaveandabsences_leave_days_amounts` option groups option values are now editable while that of other option groups remain non editable.
- Validation was added for the value field to ensure that only numbers and decimal up to two d.p are allowed for the field.

optionvalueafter

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/ValidateForm/AdminFormOptionsValidation.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/ValidateForm/AdminFormOptionsValidation.php
@@ -1,0 +1,72 @@
+<?php
+
+/**
+ * Class CRM_HRCore_Hook_ValidateForm_AdminFormOptionsValidation
+ */
+class CRM_HRCore_Hook_ValidateForm_AdminFormOptionsValidation {
+
+  /**
+   * Determines what happens if the hook is handled. Basically the hook
+   * is only handled for option values of the 'toil_amounts' and 'leave_days_amounts'
+   * option group. It basically ensures that the value of the value field of the
+   * option values for these option groups are validated before submission and
+   * if there are errors the form will not be submitted.
+   *
+   * @param string $formName
+   * @param array $fields
+   * @param mixed $files
+   * @param object $form
+   * @param array $errors
+   */
+  public function handle($formName, &$fields, &$files, &$form, &$errors) {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/ValidateForm/AdminFormOptionsValidation.php
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Class CRM_HRCore_Hook_ValidateForm_AdminFormOptionsValidation
+ */
+class CRM_HRCore_Hook_ValidateForm_AdminFormOptionsValidation {
+
+  /**
+   * Determines what happens if the hook is handled. Basically the hook
+   * is only handled for option values of the 'toil_amounts' and 'leave_days_amounts'
+   * option group. It basically ensures that the value of the value field of the
+   * option values for these option groups are validated before submission and
```
mickadoo: > ensures that the value of the value field of the option values for these option groups are validated

This makes sense after re-reading it a few times, but it hurts my head :crying_cat_face: 

Not asking you to change it though, just lamenting the language required here :smile:

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/ValidateForm/AdminFormOptionsValidation.php
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Checks if the hook should be handled.
+   *
+   * @param string $formName
+   * @param object $form
+   *
+   * @return bool
+   */
+  private function shouldHandle($formName, $form) {
+    if($formName === 'CRM_Admin_Form_Options' && $form->elementExists('value')) {
```
mickadoo: space after `if`, also you could probably use the static `class` constant instead of a string classname

PR #2659, opened 2018-06-04
PCHR-3787: Fix postInstall hooks

davialexandre:
## Overview

Right after creating a new CiviHR site (with any of the build methods available), CiviCRM would tell us that there were some extension updates available. We run all of an extension upgraders during the installation, so it's not possible for any of them to have any pending updates.

After some investigation, it turns out that this was caused by the extension schema version not being saved to the database after the installation is complete. Without that, Civi doesn't know that the extension is already in the latest version and tells the user that they should update it.

## Before

Users would be informed about pending extension updates right after a CiviHR installation.

## After

Users will not informed about pending extension updates right after a CiviHR installation.

## Technical Details
When developing a CiviCRM extension, it's common to use the `civix` tool to automatically generated some skeleton code for us. Among this, is the default implementation of some hooks, which is kept in a file named `<extension>.civix.php` and the Base Upgrader class, which is kept in `CRM/<ExtensionNamespace>/Upgrader/Base.php`.

This tool is constantly updated and some of these updates change the contents and format of these files. Unfortunately, the files need to be updated manually once a new version of `civix` is out (by calling the generate command again) and for this reason, it is easy for files to get outdated.

Outdated files are what caused the issue this PR fixes. On https://github.com/totten/civix/pull/79, they changed the way the schema version is saved to the database after an extension is installed. This is how the installation cycle would work before that PR:

1. Civi calls `hook_civicrm_install()`
 1.1 The implementation of that hook in `<extenstion>.php` would call the implementation inside the `<extension>.civix.php` file
 1.2 The implementation in `<extension>.civix.php` would call the `onInstall()` method in the Base Upgrader class
 1.3 The `onInstall()` method would call the `install()` method
 1.4 After the `install()` finishes the installation, `onInstall()` would save the max revision number (the schema version) to the database
2. Civi calls `hook_civicrm_postInstall()`
 2.1 Nothing would happen, as the extension didn't implement the `postInstall()` hook

And this is how it works after that PR:

1. Civi calls `hook_civicrm_install()`
 1.1 The implementation of that hook in `<extenstion>.php` would call the implementation inside the `<extension>.civix.php` file
 1.2 The implementation in `<extension>.civix.php` would call the `onInstall()` method in the Base Upgrader class
 1.3 The `onInstall()` method would call the `install()` method
2. Civi calls `hook_civicrm_postInstall()`
 2.1 The implementation of that hook in `<extenstion>.php` would call the implementation inside the `<extension>.civix.php` file
 2.2 The implementation in `<extension>.civix.php` would call the `onPostInstall()` method in the Base Upgrader class
 2.3 The `onPostInstall()` saves the schema version to the database

Note that now the responsibility of saving the schema version to the database belongs to the `onPostInstall()` method. Also, note that in order for this to work, it is expected that extensions implement the `hook_civicrm_postInstall()` hook. Part of the changes introduced in https://github.com/totten/civix/pull/79 is exactly about making sure that new extensions created with it would implement that by default. Unfortunately, some CiviHR extension were created before that and the automatically generated files have not been updated ever since. This PR updates these files and make sure that they implement `hook_civicrm_postInstall()` and call the `onPostInstall()` method in the Base Upgrader.

### Comments

Because of this problem, running `drush cvapi Extension.upgrade` on a CiviHR site for the first time after an installation, would run all the upgraders again and result in some duplicated data. A separate PR will be created later to remove this duplicated data from existing sites.

Only these 3 extension have been updated because they were in a unique scenario where the Base Upgrader was created after https://github.com/totten/civix/pull/79, but the `<extension>.php` file was created before that. So the Base Upgrader already contained the change where the schema was saved in the `onPostInstall()` method, but there was no hook calling it. The other extensions still have the old Upgrader which saves the schema version in the `onInstall()` method. So, other than having some outdated files, everything is good there. In order to reduce the number of changes in this PR (and the risk of breaking something with these changes) I decided to keep them as is.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.civix.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.civix.php
@@ -3,6 +3,83 @@
 // AUTO-GENERATED FILE -- Civix may overwrite any changes made to this file
 
 /**
+ * The ExtensionUtil class provides small stubs for accessing resources of this
+ * extension.
+ */
+class CRM_HRLeaveAndAbsences_ExtensionUtil {
+  const SHORT_NAME = "hrleaveandabsences";
+  const LONG_NAME = "uk.co.compucorp.civicrm.hrleaveandabsences";
+  const CLASS_PREFIX = "CRM_HRLeaveAndAbsences";
+
+  /**
+   * Translate a string using the extension's domain.
+   *
+   * If the extension doesn't have a specific translation
+   * for the string, fallback to the default translations.
+   *
+   * @param string $text
+   *   Canonical message text (generally en_US).
+   * @param array $params
+   * @return string
+   *   Translated text.
+   * @see ts
+   */
+  public static function ts($text, $params = array()) {
+    if (!array_key_exists('domain', $params)) {
+      $params['domain'] = array(self::LONG_NAME, NULL);
+    }
+    return ts($text, $params);
+  }
+
+  /**
+   * Get the URL of a resource file (in this extension).
+   *
+   * @param string|NULL $file
+   *   Ex: NULL.
+   *   Ex: 'css/foo.css'.
+   * @return string
+   *   Ex: 'http://example.org/sites/default/ext/org.example.foo'.
+   *   Ex: 'http://example.org/sites/default/ext/org.example.foo/css/foo.css'.
+   */
+  public static function url($file = NULL) {
+    if ($file === NULL) {
+      return rtrim(CRM_Core_Resources::singleton()->getUrl(self::LONG_NAME), '/');
+    }
+    return CRM_Core_Resources::singleton()->getUrl(self::LONG_NAME, $file);
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.civix.php
```diff
@@ -342,7 +441,104 @@ function _hrleaveandabsences_civix_civicrm_alterSettingsFolders(&$metaDataFolder
...
+
+/**
+ * (Delegated) Implements hook_civicrm_entityTypes().
+ *
+ * Find any *.entityType.php files, merge their content, and return.
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_entityTypes
+ */
+
+function _hrleaveandabsences_civix_civicrm_entityTypes(&$entityTypes) {
```
mickadoo: Just curious, but did this change affect anything else? Seems like changing it to return an array of entities would have some impact somewhere

PR #2667, opened 2018-06-05
PCHR-3790: Fix cost centres list error

AkA84:
## Overview
This PR fixes a regression issue introduced in #2546, whereby the cost centres list did not display any option anymore

## Before
The dropdown list is empty, and a JS error is thrown if the job role had already a cost centre assigned to it

before

## After
The dropdown list is now correctly populated, and no JS error is thrown

after

## Technical details
Before the aforementioned PR, the `vm.CostCentreList` list was not being processed by the `getActiveValues` function, for a good reason: the list is an array while the function manipulates any collection to be an object. After the PR the list was erroneously fed to the function, which in turns prevented angular to display it in the select dropdown

This PR simply reinstates the exception for this particular list, while retaining the filtering of non-active items
```js
vm[optionsData[optionGroupName].storage] = optionGroupName !== 'cost_centres'
  ? getActiveValues(optionsData[optionGroupName].data)
  : optionsData[optionGroupName].data.filter(function (optionValue) {
    return optionValue.is_active === '1'
  });
```


com.civicrm.hrjobroles/js/test/controllers/job-roles.controller.spec.js
```diff
@@ -66,7 +66,8 @@ define([
         $rootScope.$digest();
       });
 
-      it('builds the "CostCentreList" collection containing the "weight" property', function () {
+      it('builds the "CostCentreList" array containing the "weight" property', function () {
+        expect(ctrl.CostCentreList).toEqual(jasmine.any(Array));
         expect(Object.keys(ctrl.CostCentreList).length).toBe(3);
```
igorpavlov: This is array now, please do `ctrl.CostCentreList.length`

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -539,7 +539,11 @@ define([
                   };
                 }
 
-                vm[optionsData[optionGroupName].storage] = getActiveValues(optionsData[optionGroupName].data);
+                vm[optionsData[optionGroupName].storage] = optionGroupName !== 'cost_centres'
+                  ? getActiveValues(optionsData[optionGroupName].data)
+                  : optionsData[optionGroupName].data.filter(function (optionValue) {
+                    return optionValue.is_active === '1';
```
igorpavlov: `return !!+optionValue.is_active` may be safer?

PR #2668, opened 2018-06-06
PCHR-3791: Use correct default value for location_standard_hours

AkA84:
## Overview
When the record with `id=1` was deleted from the "Hours Location" list (`/civicrm/hours_location`), opening the "Add new contract" modal would throw a JS error and the "Location/Standard hours" dropdown list would incorrectly display an empty option as the default

## Before
before

## After
after

## Technical Details
The problem was that the default value for that dropdown was hardcoded, always being `1`

```js
(function init () {
  entityHour.location_standard_hours = entityHour.location_standard_hours || '1';
});
```
The solution was simply to default instead to the first item of the list of available options

```js
(function init () {
  entityHour.location_standard_hours = entityHour.location_standard_hours || $scope.utils.hoursLocation[0].id;
});

### CODE REVIEW

File: hrjobcontract/js/src/job-contract/controllers/form/form-hour.controller.js [hunk 1]
```diff
@@ -24,7 +24,7 @@ define([
     $scope.hrsAmountDefined = false;
 
     (function init () {
-      entityHour.location_standard_hours = entityHour.location_standard_hours || '1';
+      entityHour.location_standard_hours = entityHour.location_standard_hours || $scope.utils.hoursLocation[0].id;
```
reneolivo: what will happen if the hours location array is empty? Lets use lodash's get method which makes accessing undefined object/array properties safe:

```js
entityHour.location_standard_hours = entityHour.location_standard_hours || _.get(utilsHoursLocation, '[0].id');
```

`utilsHoursLocation` is defined above.

AkA84: @reneolivo sorry, what would that return if `utilsHoursLocation` is empty?

PR #2670, opened 2018-06-06
PCHR-3636: (WIP) Apply patch to CiviCRM in Jenkinsfile

mickadoo:
## Overview

We need to use the patches to CiviCRM core from our Compucorp fork in Jenkins, otherwise some of the tests for newer features which rely on core changes will fail.

## Before

Jenkins used the `civicrm/civicrm-core` branch for CiviCRM, without any changes

## After

Jenkins applies the `compucorp/civicrm-core` to the branch used for CiviCRM

## Comments

Since the patches will be applied, the manual patching for attachments changes was removed.


mickadoo:
Closing in favor of #2680 as the branch name was wrong

PR #2674, opened 2018-06-07
PCHR-3570: Footer Changes

backstageel:
## Overview
With this change only root admins and other users with "view system status on footer" will be able to view the System Status on Footer as requested on PCHR-3570

## Before
after root_admin
All users could see the system Status

## After
after root_admin
Root Admin

after-hr_admin
HR_Admin

### DIFF
--- a/hrui/templates/CRM/common/footer.tpl
+++ b/hrui/templates/CRM/common/footer.tpl
@@ -28,15 +28,16 @@
   {if !empty($contactId)}
     {include file="CRM/common/contactFooter.tpl"}
   {/if}
-
   <div class="crm-footer" id="civicrm-footer">
     {* PCHR-1323 - Display CiviHR version info. *}
     {ts}Powered by CiviHR {/ts} {civihrVersion}.
-
-    {if !empty($footer_status_severity)}
-      <span class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
-      <a target="_blank" href="{crmURL p='civicrm/a/#/status'}">{$footer_status_message}</a>.
-    </span>&nbsp;
+    {if call_user_func(array('CRM_Core_Permission','check'), 'view system status on footer')}
+      {if !empty($footer_status_severity)}
+        <span class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
+          <a target="_blank" href="{crmURL p='civicrm/a/#/status'}">{$footer_status_message}</a>.
+        </span>
+      &nbsp;
+      {/if}
     {/if}
     CiviHR is openly available under Version 3 of the <a target="_blank" href="http://www.gnu.org/licenses/agpl-3.0.html">GNU AGPL License</a> and can be downloaded from <a target="_blank" href="https://civihr.org">www.civihr.org</a>.
     <div class="text-center">
--- a/uk.co.compucorp.civicrm.hrcore/hrcore.php
+++ b/uk.co.compucorp.civicrm.hrcore/hrcore.php
@@ -318,6 +318,7 @@ function hrcore_civicrm_permission(&$permissions) {
   $prefix = ts('CiviHR') . ': ';
   $permissions['access CiviCRM developer menu and tools'] = ts('Access CiviCRM developer menu and tools');
   $permissions['access root menu items and configurations'] = $prefix . ts('Access root menu items and configurations');
+  $permissions['view system status on footer'] = $prefix . ts('View System Status on Footer');
 }
 
 /**


hrui/templates/CRM/common/footer.tpl
```diff
@@ -28,17 +28,29 @@
   {if !empty($contactId)}
     {include file="CRM/common/contactFooter.tpl"}
   {/if}
-
   <div class="crm-footer" id="civicrm-footer">
     {* PCHR-1323 - Display CiviHR version info. *}
     {ts}Powered by CiviHR {/ts} {civihrVersion}.
 
+    {php}
```
mickadoo: Can you change this to use something like [this](https://github.com/civicrm/civihr/blob/cc7514b47b69f3c1f86915d46e73ced3b4dc3b60/hrjobcontract/CRM/Hrjobcontract/Page/View/Summary.tpl#L52). It's better to not call Drupal functions directly in CiviHR since we're still trying to be CMS-agnostic

hrui/templates/CRM/common/footer.tpl
```diff
@@ -28,17 +28,29 @@
...
+      if(user_access('view system status on footer')):
+    {/php}
+
     {if !empty($footer_status_severity)}
-      <span class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
+      <span
```
mickadoo: some of the alignment looks strange here, I know they're long lines, but maybe it makes sense to keep the text on the same line and only separate by dom elements, e.g.

```
<div>
  <span class"foo">long line of text long line of text long line of text long line of text long line of text long line of text long line of text long line of text </span>
</div>

**Comment by mickadoo** [comment]
@backstageel can you fix the PR description

- The format of the title should be `<ticket number>: Description in UCFirst Capitals`
- There is some extra text overflowing from the title in the description, please remove this
- We follow a format, but I notice you removed the "Before" section, it's important to have this to see what has changed. Please check [other PRs](https://github.com/civicrm/civihr/pull/2667) for an example.

I see you use the `view system status on footer` permission, but where is this permission created? If it's a separate PR in a different repo then usually we ask the same reviewer to review all related PRs for a ticket, unless it's a mix of frontend and backend.

**Comment by backstageel** [comment]
Hi, i made the changes and added some aditional commits to add the new permission. please review

File: uk.co.compucorp.civicrm.hrcore/hrcore.php [hunk 3]
```diff
@@ -318,6 +318,8 @@ function hrcore_civicrm_permission(&$permissions) {
   $prefix = ts('CiviHR') . ': ';
   $permissions['access CiviCRM developer menu and tools'] = ts('Access CiviCRM developer menu and tools');
   $permissions['access root menu items and configurations'] = $prefix . ts('Access root menu items and configurations');
+  $permissions['view system status on footer'] = $prefix . ts('View System Status on Footer');
+
```
ajesamson: Can we remove this extra space.

hrui/templates/CRM/common/footer.tpl
```diff
@@ -28,16 +28,21 @@
   {if !empty($contactId)}
     {include file="CRM/common/contactFooter.tpl"}
   {/if}
-
   <div class="crm-footer" id="civicrm-footer">
     {* PCHR-1323 - Display CiviHR version info. *}
     {ts}Powered by CiviHR {/ts} {civihrVersion}.
 
+    {if call_user_func(array('CRM_Core_Permission','check'), 'view system status on footer')}
```
ajesamson: Kindly align the section properly. Something like this
```html
{if call_user_func(array('CRM_Core_Permission','check'), 'view system status on footer')}
  {if !empty($footer_status_severity)}
    <span class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
      <a target="_blank" href="{crmURL p='civicrm/a/#/status'}">{$footer_status_message}</a>.
    </span>
    &nbsp;
  {/if}
{/if}
```

hrui/templates/CRM/common/footer.tpl
```diff
@@ -28,16 +28,21 @@
...
+      <span
+        class="status{if $footer_status_severity gt 3} crm-error{elseif $footer_status_severity gt 2} crm-warning{else} crm-ok{/if}">
       <a target="_blank" href="{crmURL p='civicrm/a/#/status'}">{$footer_status_message}</a>.
-    </span>&nbsp;
+    </span>
+      &nbsp;
+    {/if}
     {/if}
+
+
```
ajesamson: Kindly remove this extra space.

ajesamson:
The description section is still having `…admins and users with view system status on footer permission`. Is it part of overview?
Kindly add label `ready to test` for jenkins test to run.

hrui/templates/CRM/common/footer.tpl
backstageel: changes made

mickadoo:
@backstageel so I can see now the permission is created in this PR, but it will still not be assigned to any roles by default.

When adding a new permission it usually requires 3 steps:

1. Create the permission in civihr
1. Define default roles in civihr-employee-portal
1. Revert the default permissions feature

You've done step 1 here but we need to do step 2 and 3 in civihr-employee-portal. Have a look at these two commits that [add the permission](https://github.com/compucorp/civihr-employee-portal/commit/9d31d2510b4c89780c56d98a93f481ae4b79dd96) and [add it to the feature](https://github.com/compucorp/civihr-employee-portal/commit/e74dfce00307041d2fbcd5b4a3125d16f5d48053). You'll want to do something similar for your new permission.

After that we need to reset the default permissions for step 3. Previously this was done using a drush command as part of the release notes. However release notes can cause complications if someone forgets to run them, so it's better to use an upgrader for this. Have a look at [this upgrader](https://github.com/compucorp/civihr-employee-portal/blob/abf6e8db1d19b6167f6c60e37e0c943d1bc4a2e8/civihr_employee_portal/civihr_employee_portal.install#L423) that reverts the default permissions.

After you create your upgrader you can run `drush updb -y` to test it locally

PR #2675, opened 2018-06-07
PCHR-3722: Fix leave requests being shown in calendar

igorpavlov:
## Overview

This PR fixes the issue when multiple day Leave Requests in hours which have "to" time earlier than "from" time, are not shown for the last day of the Leave Request in the Leave Calendar.

For example, the Leave Request `2018-06-04 16:00` - `2018-06-05 12:00` will not appear in the Leave Calendar for the day **June 5th**.

## Before

image

## After

image

## Technical Details

When we get the **days** the Leave Request belongs to we should *not* care about time.

```js
var pointerDate = moment(leaveRequest.from_date)
  .hour(0).minute(0).second(0).clone(); // flush time
var toDate = moment(leaveRequest.to_date)
  .hour(0).minute(0).second(0);  // flush time

  // and then we can safely run through the dates
  while (pointerDate.isSameOrBefore(toDate)) {
    // ...
    pointerDate.add(1, 'day');
```

---- 

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/shared/components/leave-calendar-month.component.js
```diff
@@ -378,8 +378,10 @@ define([
      */
     function leaveRequestDays (leaveRequest) {
       var days = [];
-      var pointerDate = moment(leaveRequest.from_date).clone();
-      var toDate = moment(leaveRequest.to_date);
+      var pointerDate = moment(leaveRequest.from_date)
+        .hour(0).minute(0).second(0).clone();
```
reneolivo:
```js
var pointerDate = moment(leaveRequest.from_date).startOf('day');
var toDate = moment(leaveRequest.to_date).startOf('day');
```

Side question: Why do you need to clone it if you already modified it?

igorpavlov: No idea, I also think the .clone() is not needed here.

igorpavlov: It actually goes from here https://github.com/civicrm/civihr/pull/1989, but I also don't think `clone` is needed here. We instantiate a `moment` object from a string and then use `format()` as output. So we don't store references.

PR #2677, opened 2018-06-07
Test

davialexandre:


mickadoo:
@davialexandre I removed the label from this just to avoid Jenkins retesting it after every staging change

davialexandre:
@mickadoo thanks. This can be closed. I created it to test that SQL issue on the Jenkins server, but since you've figured out the problem already, this isn't necessary anymore.

PR #2682, opened 2018-06-12
PCHR-3820: Add Leave Request ACL Rules to LeaveRequest.getAttachments API

tunbola:
## Overview
Currently the LeaveRequest.getAttachments API allows a staff to view leave attachments for another staff. This PR incorporates the Leave ACL rules into this API, basically, A staff should only be able to view own attachments, A manager should be able to view attachments for managees only while Admin can view leave attachments for all contacts leave requests.

## Before
- The LeaveRequest.getAttachments API does not return results based on Leave Request ACL rules

## After
- The LeaveRequest.getAttachments API returns results based on Leave Request ACL rules

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
@@ -2,6 +2,7 @@
 
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_Service_LeaveManager as LeaveManagerService;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestRights as LeaveRequestRights;
 
 class CRM_HRLeaveAndAbsences_Service_LeaveRequestAttachment {
 
@@ -71,5 +72,29 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequestAttachment {
     $params = $this->prepareParametersForAttachmentPayload($params);
 
     return civicrm_api3('Attachment', $action, $params);
+  }
+
+  /**
+   * Uses the Attachment API to retrieve attachments associated with a LeaveRequest.
+   * It ensures that the current user can only retrieve Leave attachments for the
+   * Leave requests linked to the contacts the user has access to. The admin can
+   * retrieve all Leave attachments for all contacts.
+   *
+   * @param array $params
+   *
+   * @return array
+   */
+  public function get($params) {
+    $leaveRequestID = isset($params['entity_id']) ? $params['entity_id'] : '';
+    $leaveRequest = LeaveRequest::findById($leaveRequestID);
+    $leaveManagerService = new LeaveManagerService();
+    $leaveRequestRights = new LeaveRequestRights($leaveManagerService);
+    $accessibleContacts = $leaveRequestRights->getLeaveContactsCurrentUserHasAccessTo();
+
+    if ($leaveManagerService->currentUserIsAdmin() || in_array($leaveRequest->contact_id, $accessibleContacts)) {
+      return $this->callAttachmentAPI('get', $params);
+    }
+
+    return [];
   }
 }
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
@@ -2,9 +2,30 @@
 
 use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
 use CRM_HRLeaveAndAbsences_Service_LeaveManager as LeaveManagerService;
-use CRM_HRLeaveAndAbsences_Service_LeaveRequestRights as LeaveRequestRights;
+use CRM_HRLeaveAndAbsences_Service_LeaveRequestRights as LeaveRequestRightsService;
 
 class CRM_HRLeaveAndAbsences_Service_LeaveRequestAttachment {
+
+  /**
+   * @var LeaveManagerService
+   */
+  private $leaveManagerService;
+
+  /**
+   * @var LeaveRequestRightsService
+   */
+  private $leaveRequestRightsService;
+
+  /**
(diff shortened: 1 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
```diff
@@ -4370,24 +4370,40 @@ public function testGetAttachmentsShouldThrowAnExceptionIfLeaveRequestIDIsMissin
   }
 
   public function testGetAttachments() {
```
davialexandre: I wonder if we still need this test if with all new ones you've added. What does it have that the others don't?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
```diff
@@ -72,4 +73,28 @@ private function callAttachmentAPI($action, $params) {
...
+    $leaveManagerService = new LeaveManagerService();
+    $leaveRequestRights = new LeaveRequestRights($leaveManagerService);
+    $accessibleContacts = $leaveRequestRights->getLeaveContactsCurrentUserHasAccessTo();
+
+    if ($leaveManagerService->currentUserIsAdmin() || in_array($leaveRequest->contact_id, $accessibleContacts)) {
+      return $this->callAttachmentAPI('get', $params);
+    }
+
+    return [];
+  }
```
davialexandre: Please also write tests for this method

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
tunbola: tbh, I wanted to remove it but the reason I left i was because the test checks for all the expected fields that the API is expected to return which the others don't. I could remove it though. What do you think?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
tunbola: I was thinking it will be a duplication of the tests for the LeaveRequest.getAttachments API, since this API uses this method directly.

I could modify this method to accept params and also the LeaveRequestRights service so that I can mock that and pass it in the function but I would still have to create leave requests and all. What do you suggest?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestTest.php
davialexandre: Maybe then it would make more sense to rename this method to better clarify that it is more about the fields? Or create some custom assert method that checks all the fields, reuse it in the other tests and delete this one.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachment.php
davialexandre: Unfortunately, because of how the API works, there's no way to avoid this duplication. The tests wouldn't be completely duplicated though, since, as you said, you can mock a few things. In this case it's better to have some duplication than not having tests for this method

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachmentTest.php
```diff
@@ -151,4 +151,94 @@ private function getDefaultLeaveRequestParams($params = []) {
...
+
+    $staff1Attachment = $leaveRequestAttachmentService->get(['entity_id' => $leaveRequest1->id, 'sequential' => 1]);
+    $staff2Attachment = $leaveRequestAttachmentService->get(['entity_id' => $leaveRequest2->id, 'sequential' => 1]);
+
+    //Admin is able to access attachments for all contacts
+    $this->assertCount(1, $staff1Attachment['values']);
+    $this->assertEquals($staff1Attachment['values'][0]['id'], $attachment1['id']);
+    $this->assertEquals($staff1Attachment['values'][0]['name'], $attachment1['name']);
+
+    $this->assertCount(1, $staff1Attachment['values']);
```
davialexandre: I believe this should actually be `$staff2Attachment`, right?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachmentTest.php
```diff
@@ -108,36 +110,34 @@ public function testDeleteShouldDeleteAttachmentWhenLoggedInUserIsTheLeaveApprov
...
-    $adminID = 2;
     $params = $this->getDefaultLeaveRequestParams();
     $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation($params);
 
-    // Register contact in session and set permission to admin
-    $this->registerCurrentLoggedInContactInSession($adminID);
-    CRM_Core_Config::singleton()->userPermissionClass->permissions = ['administer leave and absences'];
+    $leaveManagerService = $this->getLeaveManagerServiceWhenUserIsAdmin();
+    $leaveRequestRights = new LeaveRightsService($leaveManagerService);
+    $leaveRequestAttachmentService = new LeaveRequestAttachmentService($leaveRequestRights, $leaveManagerService);
```
davialexandre: This instantiation code is duplicated in a few places. What do you think of moving it to a `createLeaveRequestAttachmentsServiceWhenUserIsAdmin()` (I know, a bit too long, but quite descriptive and still shorter than 3 lines of code)

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestAttachmentTest.php
tunbola: Yes

PR #2688, opened 2018-06-18
PCHR-3836: Update Existing Case Types with Case Type Category

ajesamson:
## Overview
After adding [category custom field](https://github.com/compucorp/civihr-tasks-assignments/pull/374) to case type, existing records are to be categorized. If case type name is application, they are defaulted to `Vacancy` otherwise they are defaulted to `Workflow`.

## Before
Existing case types are not grouped by categories

## After
Existing case types are assigned to either vacancy or workflow category.

## Technical Details
The upgrader is executed after the case type category have been setup.
```
$customFieldId = CRM_Core_BAO_CustomField::getCustomFieldID('Category');
if ($customFieldId == null) {
  return FALSE;
}
```

### DIFF
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -182,6 +182,42 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
 
     return TRUE;
   }
+  
+  /**
+   * Sets case type category to Vacancy for application type or
+   * to Workflow if otherwise
+   *
+   * @return bool
+   */
+  public function upgrade_1432() {
+    $customFieldId = CRM_Core_BAO_CustomField::getCustomFieldID('Category');
+    if ($customFieldId == null) {
+      return FALSE;
+    }
+    
+    $category = "custom_" . $customFieldId;
+    // set category to Vacancy if name is application
+    civicrm_api3('CaseType', 'get', [
+      'sequential' => 1,
+      'name' => 'application',
+      'api.CaseType.create' => [
+        'id' => '$value.id',
+        $category => 'Vacancy'
+      ],
+    ]);
+    
+    // set category to Workflow if name is not application
+    civicrm_api3('CaseType', 'get', [
+      'sequential' => 1,
+      'name' => ['!=' => 'application'],
+      'api.CaseType.create' => [
+        'id' => '$value.id',
+        $category => 'Workflow'
+      ],
+    ]);
+    
+    return TRUE;
+  }
 
   /**
    * Replaces (Case) keyword and (Open Case) keyword with (Assignment) keyword
--- a/hrcase/CRM/HRCase/Upgrader.php
+++ b/hrcase/CRM/HRCase/Upgrader.php
@@ -190,33 +190,66 @@ class CRM_HRCase_Upgrader extends CRM_HRCase_Upgrader_Base {
    * @return bool
    */
   public function upgrade_1432() {
-    $customFieldId = CRM_Core_BAO_CustomField::getCustomFieldID('Category');
-    if ($customFieldId == null) {
+    $result = civicrm_api3('CustomField', 'get', [
+      'name' => 'Category',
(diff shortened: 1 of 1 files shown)


hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -182,6 +182,42 @@ public function upgrade_1431() {
...
+      ],
+    ]);
+    
+    // set category to Workflow if name is not application
+    civicrm_api3('CaseType', 'get', [
+      'sequential' => 1,
+      'name' => ['!=' => 'application'],
+      'api.CaseType.create' => [
+        'id' => '$value.id',
+        $category => 'Workflow'
```
davialexandre: It would be nice if we could avoid not having these hardcoded values here, but instead get them from the option value I've mentioned in your other PR

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -182,6 +182,64 @@ public function upgrade_1431() {
...
   }
+  
+  /**
+   * Sets case type category to Vacancy for application type or
+   * to Workflow if otherwise
+   *
+   * @return bool
+   */
+  public function upgrade_1432() {
+    $customFieldId = CRM_Core_BAO_CustomField::getCustomFieldID('Category');
```
davialexandre: What happens here if there's more than one custom field with this same label? Wouldn't it be better to use the API and fetch this based on the field name instead?

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -182,6 +182,64 @@ public function upgrade_1431() {
...
+  }
+  
+  /**
+   * Retrieves option group
+   *
+   * @param $option
+   *
+   * @return array
+   */
+  private function getOptionValue($option) {
```
davialexandre: Since this method specifically returns option values from the Case Type Category Option Group, I think it is better to rename it to something like `getCaseTypeCategoryOptionValue()`

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -182,6 +182,64 @@ public function upgrade_1431() {
...
+          $category => $categoryValue['values'][0]['value']
+        ],
+      ]);
+    }
+    
+    return TRUE;
+  }
+  
+  /**
+   * Retrieves option group
```
davialexandre: It returns an option value, not a group

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -182,6 +182,64 @@ public function upgrade_1431() {
...
+   * @return array
+   */
+  private function getOptionValue($option) {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
+      'option_group_id' => 'category_custom_field',
+      'value' => $option
+    ]);
+    
+    return $result;
```
davialexandre: Since this method is expected to return a single option value, I think we should normalize the result instead of returning the full API response. We really only care about the Option Value values here and not about all the count, is_error, etc.

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -190,7 +190,13 @@ public function upgrade_1431() {
...
    */
   public function upgrade_1432() {
-    $result = civicrm_api3('CustomField', 'get', ['name' => 'Category']);
+    $result = civicrm_api3('CustomField', 'get', [
+      'name' => 'Category',
+      'api.OptionGroup.get' => [
+        'id' => '$value.option_group_id',
+        'name' => 'case_type_category'
+      ],
+    ]);
```
davialexandre: @ajesamson sorry for the confusion, on my last comment I meant the custom group id instead of the option group. Using the option group id like you did here will not help getting the right custom field.

So basically, what you need to do here is something like:

```php
civicrm_api3('CustomField', 'get', [
  'name' => 'Category',
  'custom_group_id' => 'case_type_category'
]);
```

PR #2694, opened 2018-06-19
PCHR-3611: Remove Main And Other Location Types

mickadoo:
## Overview

Remove the "Main" and "Other" location types as they are not required.

## Before

The "Main" and "Other" location types existed on some sites

## After

The "Main" and "Other" location types are removed


mickadoo:
@backstageel this feature hasn't been released yet, or even merged to staging so this upgrader will not have been run on any sites except our local ones and possibly testing sites. Testing sites will be reset after each deployment so it won't be an issue. In future if you're working on a feature and had some issue with an upgrader it's fine to modify it, the only problem is modifying upgraders that have already been merged to staging

PR #2698, opened 2018-06-20
PCHR-3546: Make All Currencies Available

backstageel:
## Overview
Made all currencies available for new installations

## Before
Only few currencies was available on new installations
screenshot from 2018-06-20 18 18 00

## After
All Currencies are available on new CiviHR installations by default
screenshot from 2018-06-20 18 12 42

## Technical Details
Couldn't find any CiviCRM API to retrieve currencies, so made direct query.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
@@ -71,6 +71,7 @@ class CRM_HRCore_Upgrader extends CRM_HRCore_Upgrader_Base {
     $this->createRequiredLocationTypes();
     $this->deleteUnneededCustomGroups();
     $this->createDefaultRelationshipTypes();
+    $this->makeAllCurrenciesAvailable();
     $this->runAllUpgraders();
   }
 
@@ -220,6 +221,29 @@ class CRM_HRCore_Upgrader extends CRM_HRCore_Upgrader_Base {
         'is_reserved' => 0,
         'is_active' => 1,
       ]);
+    }
+  }
+
+  /**
+   * Making All Currencies Available for new installations
+   */
+  private function makeAllCurrenciesAvailable() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'return' => ['name'],
+      'option_group_id' => 'currencies_enabled',
+    ]);
+    $enabledCurrencies = array_column($result['values'], 'name');
+
+    $dao = CRM_Core_DAO::executeQuery('SELECT * from civicrm_currency');
+    while ($dao->fetch()) {
+      if (!in_array($dao->name, $enabledCurrencies)) {
+        civicrm_api3('OptionValue', 'create', [
+          'option_group_id' => 'currencies_enabled',
+          'label' => $dao->name . ' (' . $dao->symbol . ')',
+          'value' => $dao->name,
+          'name' => $dao->name . ' (' . $dao->symbol . ')',
+        ]);
+      }
     }
   }
 


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -188,6 +189,30 @@ private function createDefaultRelationshipTypes() {
...
+    while ($dao->fetch()) {
+      $currencyExists = civicrm_api3('OptionValue', 'get', [
+        'value' => $dao->name,
+      ]);
+      if (empty($currencyExists['values'])) {
+        civicrm_api3('OptionValue', 'create', [
+          'option_group_id' => 'currencies_enabled',
+          'label' => $dao->name . ' (' . $dao->symbol . ')',
+          'value' => $dao->name,
+          'name' => $dao->name,
```
tunbola: Took a quick look in my local db, seems the enabled currencies have the same value for both name and label. 
e.g EUR (€).

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -188,6 +189,30 @@ private function createDefaultRelationshipTypes() {
...
+          'is_active' => 1,
```
tunbola: I don't think we need to se is_active, is_reserved, is_optgroup and is_default, The values set for them are the same they have as the default database values.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -188,6 +189,30 @@ private function createDefaultRelationshipTypes() {
...
 
+  /**
+   * Making All Currencies Available for new installations
+   */
+  private function makeAllCurrenciesAvailable() {
+    $dao = CRM_Core_DAO::executeQuery('SELECT * from civicrm_currency');
+    while ($dao->fetch()) {
+      $currencyExists = civicrm_api3('OptionValue', 'get', [
+        'value' => $dao->name,
+      ]);
```
tunbola: You need to add the option_group_id to the parameter to fetch the option value. In Civi it is possible for an option value with same names to exist and they will belong to different option groups. So for example we can have `GBP` belonging as an option value to another option group.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -188,6 +189,30 @@ private function createDefaultRelationshipTypes() {
     }
   }
 
+  /**
+   * Making All Currencies Available for new installations
+   */
+  private function makeAllCurrenciesAvailable() {
+    $dao = CRM_Core_DAO::executeQuery('SELECT * from civicrm_currency');
+    while ($dao->fetch()) {
```
tunbola: Also since the currency records might be up to 200(on my local installation), wondering if we could optimize this by using a single API call to get all enable currencies names.

```php
$result = civicrm_api3('OptionValue', 'get', array(
  'sequential' => 1,
  'return' => array("name"),
  'option_group_id' => "currencies_enabled",
));
```
Then in this loop, we can check if 
```php
if(!in_array($dao->name, $enabledCurrency){
// create the currency
}
```

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -194,20 +194,18 @@ private function createDefaultRelationshipTypes() {
    */
   private function makeAllCurrenciesAvailable() {
     $dao = CRM_Core_DAO::executeQuery('SELECT * from civicrm_currency');
+    $result = civicrm_api3('OptionValue', 'get', array(
```
tunbola: Please use short array syntax and single quote for strings.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -194,20 +194,18 @@ private function createDefaultRelationshipTypes() {
...
+      'return' => array("name"),
+      'option_group_id' => "currencies_enabled",
+    ));
+    $enabledCurrency = array_column($result['values'],'name');
```
tunbola: I think `enabledCurrencies` might be a better name since there will most likely be more than one.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -188,6 +189,28 @@ private function createDefaultRelationshipTypes() {
     }
   }
 
+  /**
+   * Making All Currencies Available for new installations
+   */
+  private function makeAllCurrenciesAvailable() {
+    $dao = CRM_Core_DAO::executeQuery('SELECT * from civicrm_currency');
```
tunbola: Please move this line to just before the while loop.
You can also add a space before that so the code is not cluttered together

PR #2700, opened 2018-06-21
PCHR-3603: Relabel Line Manager Relationship

backstageel:
## Overview
For old and new sites: Relabel "Line manager" relationship 

'Line Manager is' to 'Is Line Managed by'
'Line Manager' to 'Is Line Manager of'

## Before
screenshot from 2018-06-21 09 17 45

## After
screenshot from 2018-06-21 09 19 03

## Comments
Only changed the Label, because the name is used explicitly in some places, such as https://github.com/compucorp/civihr/blob/0a7d1905be8e5148222bece35aebb2a4884ebfc4/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/Manager.php#L16

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1020.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1020.php
@@ -1,0 +1,31 @@
+<?php
+
+trait CRM_HRCore_Upgrader_Steps_1020 {
+
+  /**
+   * Rename Line Manager Relationship
+   */
+  public function upgrade_1020() {
+    $this->up1020_renameLineManagerRelationship();
+
+    return TRUE;
+  }
+
+  /**
+   * Renaming Line Manager Relationship
+   */
+  private function up1020_renameLineManagerRelationship() {
+    $relationshipType = civicrm_api3('RelationshipType', 'get', [
+      'name_b_a' => 'Line Manager',
+    ]);
+    if (empty($relationshipType['values'])) {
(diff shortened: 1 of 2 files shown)


mickadoo:
If you're just changing the label, can you rename the PR title to "relabel" instead of "rename"?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1020.php
```diff
@@ -0,0 +1,31 @@
...
+  public function upgrade_1020() {
+    $this->up1020_renameLineManagerRelationship();
+
+    return TRUE;
+  }
+
+  /**
+   * Renaming Line Manager Relationship
+   */
+  private function up1020_renameLineManagerRelationship() {
```
mickadoo: Same as my suggestion for the PR title, if it's just the label change please rename this function and update the docblock

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1020.php
```diff
@@ -0,0 +1,31 @@
...
+    $relationshipType = civicrm_api3('RelationshipType', 'get', [
+      'name_b_a' => 'Line Manager',
+    ]);
+    if (empty($relationshipType['values'])) {
```
mickadoo: This works fine, but since you're using `$relationshipType['id']` in the next line why not check for this here instead, just makes more sense when you're reading it

PR #2705, opened 2018-06-21
PCHR-3910: Add calendar feeds module

reneolivo:
## Overview
This PR adds the following changes:

* Removes unnecessary models/instances/apis belonging to the calendar feeds feature.
* Creates a new folder for the calendar feeds feature where all the code should be placed.

## Technical Details
The models, instances, and api files and their respective files were removed since they are currently not needed by this feature. This files will be added later when needed.

For the feature module, the [folders by feature](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#style-y152) approach was implemented. For a sample usage of this implementation please refer to https://github.com/compucorp/civihr/pull/2702

To support test files inside the feature folder the paths for tests was changed:
```js
// From:
path.join(extPath, '**', 'test/**/*.spec.js')

// To:
path.join(extPath, '**/*.spec.js')
```
This means spec files won't just be limited to be included in the "test" folder.

To allow templates to be included in the feature folder the shared-settings module now has a new property:
```js
sourcePath: CRM.vars.leaveAndAbsences.baseURL + '/js/angular/src/leave-absences/'
```

To find the template components can do the following:

```js
template: ['shared-settings', function (sharedSettings) {
  return sharedSettings.sourcePath + 'feature-folder/path/to/file.html';
}];
```

----
✅Manual Tests - passed
✅Jasmine Tests - passed
⏹JS distributive files - not needed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrcore/gulp/tasks/test.js
```diff
@@ -31,7 +31,7 @@ module.exports = [
       if (utils.canCurrentExtensionRun('test')) {
         extPath = utils.getExtensionPath();
         watchPatterns = utils.addExtensionCustomWatchPatternsToDefaultList([
-          path.join(extPath, '**', 'test/**/*.spec.js'),
```
igorpavlov: Would that mean that we will not be able to TDD the files belonging to the old structure?

reneolivo: yes it does 
anim

PR #2707, opened 2018-06-22
PCHR-3871: Create the LeaveRequestCalendarFeedConfig Entity

tunbola:
## Overview
The LeaveRequestCalendarFeedConfig Entity will be used to create/update/delete information related to configuring a leave request calendar feed. The configuration for a calendar feed will determine the leave information each feed will contain. It is likely that a few more fields will be added to this entity as regards feed visibility(i.e contacts allowed to view a feed) and what each feed will be composed of once the decision about how these filters will be structured is finalised.

## Before
The LeaveRequestCalendarFeedConfig Entity does not exist.

## After
The LeaveRequestCalendarFeedConfig Entity is created and contains the following fields:
- id
- title (VARCHAR)
- timezone (VARCHAR)
- hash (VARCHAR)
- is_active (bool)
- created_date (DATETIME)

The hash is the unique identifier for each feed and it will be used as part of the properties that will make up the feed URL. It is created for each feed using the function:
```php
md5(uniqid(rand(), true);
//Sample values
8a87dc95a60b352052cd4164d928a351
8abc95ddcf383b1da8fa4b407ad755b6
```


uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
```diff
@@ -0,0 +1,63 @@
...
+    <required>true</required>
+    <comment>The Calendar Feed Config Timezone</comment>
+  </field>
+
+  <field>
+    <name>hash</name>
+    <type>varchar</type>
+    <length>50</length>
+    <required>true</required>
+    <unique>true</unique>
```
davialexandre: You also need an index declaration like this one, otherwise the UNIQUE constraint will not be added to the sql file: https://github.com/compucorp/civihr/blob/8b865a2815528061338a946ff580b3e9511344ea/uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/AbsencePeriod.xml#L57-L62

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="iso-8859-1" ?>
+
+<table>
+  <base>CRM/HRLeaveAndAbsences</base>
+  <class>LeaveRequestCalendarFeedConfig</class>
```
davialexandre: Is this really a config or will each instance of this be a Feed itself? Will a Feed contain any additional information that is not in the Config already?

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
```diff
@@ -0,0 +1,63 @@
...
+    <type>varchar</type>
+    <length>50</length>
+    <required>true</required>
+    <comment>The Calendar Feed Config Timezone</comment>
+  </field>
+
+  <field>
+    <name>hash</name>
+    <type>varchar</type>
+    <length>50</length>
```
davialexandre: The size of an md5 hash is 32 chars. Why is this field size 50?

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
```diff
@@ -0,0 +1,63 @@
...
+    <type>boolean</type>
+    <default>1</default>
+    <comment>Whether the feed is active or not</comment>
+  </field>
+
+  <field>
+    <name>created_date</name>
+    <type>datetime</type>
+    <comment>The date and time this Calendar Feed Config was created</comment>
+    <add>4.4</add>
```
davialexandre: Shouldn't this be required?

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
```diff
@@ -0,0 +1,63 @@
...
+    <required>true</required>
+    <unique>true</unique>
+    <comment>The Calendar Feed Config Hash. Should be unique per config</comment>
+  </field>
+
+  <field>
+    <name>is_active</name>
+    <type>boolean</type>
+    <default>1</default>
+    <comment>Whether the feed is active or not</comment>
```
davialexandre: Maybe this should be required too?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -0,0 +1,115 @@
...
+   *
+   * @throws InvalidLeaveRequestCalendarFeedConfigException
+   */
+  private static function validateTimezone($params) {
+    $timeZone = CRM_Utils_Array::value('timezone', $params);
+
+    if (!$timeZone) {
+      return;
+    }
+    $timeZoneList = DateTimeZone::listIdentifiers(DateTimeZone::ALL);
```
davialexandre: This extension already has a [validator for dates](https://github.com/compucorp/civihr/blob/1e304cfe35788547e75c09fc3678fa3fa4dac5a7/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Validator/Date.php) and I think that validating timezones is not really the responsibility of a Calendar Feed. What do you think of adding a validator for timezones?

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfigTest.php
```diff
@@ -0,0 +1,98 @@
...
+   * @expectedException CRM_HRLeaveAndAbsences_Exception_InvalidLeaveRequestCalendarFeedConfigException
+   * @expectedExceptionMessage A leave request calendar feed configuration with same title already exists!
+   */
+  public function testCreateWillThrowAnExceptionWhenTitleAlreadyExists() {
+    LeaveRequestCalendarFeedConfig::create([
+      'title' => 'Feed 1',
+      'timezone' => 'Europe/Stockholm',
+    ]);
+
+    LeaveRequestCalendarFeedConfig::create([
```
davialexandre: For this test, I think that using `setExpectedException()` and `setExpectedExceptionMessage()` are better options than the annotations. With the methods you can make sure the test will only pass if the second `create()` call is the one that throws the exception

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
tunbola: It has a default value though, thats why i made it not to be required

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
tunbola: I agree. Just left it like that incase we change the logic later and we might need more chars. I could make it exactly 32 anyway.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
tunbola: Its a config actually. The feed is the actual leave requests data generated off the configurations set for it.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/LeaveRequestCalendarFeedConfig.xml
tunbola: Will add that. Thanks.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
tunbola: I agree its a good Idea.

PR #2717, opened 2018-06-26
PCHR-3558: Rename Job Contract End Reason

ajesamson:
## Overview
This PR renames page title from `Job Contract End Reasons` to `Contract End Reasons`.

## Before
before_hrjc_contract_end_reason

## After
after_hrjc_contract_end_reason

## Technical Details
An upgrader was executed to change the page title for job contract end reasons to `Contract End Reasons`
```php
'api.OptionGroup.create' => [
  'id' => '$value.id',
  'title' => 'Contract End Reasons'
],
```

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1277,6 +1277,24 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
     
     return TRUE;
   }
+
+  /**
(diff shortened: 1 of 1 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1277,6 +1277,23 @@ public function upgrade_1037() {
     
     return TRUE;
   }
+
+  /**
+   * Renames page title from  job contract end reason to contract end reasons
```
tunbola: I think this upgrader actually renames the option group title. It just happens that the option group title is used for the page title.

PR #2720, opened 2018-06-26
PCHR-3429: CiviCRM upgrade

davialexandre:
## Overview

This PR updates CiviHR so that it can be upgraded from CiviCRM 4.7.27 to CiviCRM 5.3.0. It includes fixes for things that stopped working after the upgrade, syncing of overridden files and some minor fixes not related to the upgrade. See "Technical Details" for more information.

## Before

CiviHR could not be installed with CiviCRM 5.3.0

## After 

CiviHR can be installed with CiviCRM 5.3.0

### Syncing of overridden files
As usual with every upgrade, the overridden files had to be synced. To make the review process easier, here is a list of the them with a diff comparing the file to its original in core:

- `CRM/Hrjobcontract/Dedupe/Merger.php`: [diff](https://www.diffchecker.com/rm8Vy70U)
- `CRM/Hrjobcontract/Form/Merge.php`: [diff](https://www.diffchecker.com/PpL8a5jn)
- `templates/CRM/Hrjobcontract/Form/Merge.tpl`: [diff](https://www.diffchecker.com/FuTzYFNl)
- `templates/CRM/Hrjobcontract/Form/Merge.hlp`: [diff](https://www.diffchecker.com/xJxmykaO)
- `hrjobcontracts/CRM/Export/BAO/Export`: [diff](https://www.diffchecker.com/HnOHuVKK)
- `templates/hrui/CRM/Contact/Page/View/Summary.tpl`: [diff](https://www.diffchecker.com/c0rENpFo)
- `templates/hrui/CRM/common/footer.tpl`: [diff](https://www.diffchecker.com/5jm1pZ4w)
- `hrui/CRM/Core/BAO/Navigation.php`: [diff](https://www.diffchecker.com/kQra6Xsf)

Besides syncing the `Summary.tpl` file with core, it was also updated to remove a customization that was not anymore necessary. The customization was added in https://github.com/civicrm/civicrm-core/pull/11439 and it was used to allow the Contact Actions Menu extension to replace the original CiviCRM contact actions menu. Tunbola first tried to make core more flexible by adding a region to the menu area, but until the PR that does that (https://github.com/civicrm/civicrm-core/pull/11439) was merged to core, we decided to go with a temporary solution customizing the `Summary.tpl` file. Now that we'll be upgrading to a CiviCRM version which contains Tunbola's solution, we can get rid of the customization.

The `CustomData.php` and `CustomData.tpl` overridden files in hrjobcontracts were delete in https://github.com/compucorp/civihr/pull/2720/commits/dbe0bec4cfa4746e3b35d55be724725eeca7c4b6, as they were not being used anymore. Pleas check the commit message for more details.

### Order of menu items added with hook_civicrm_navigationMenu

In https://github.com/civicrm/civicrm-core/pull/11772, the `hook_civicrm_navigationMenu()` was updated and now `weight` started being taken into account for menu items added with it. In the past, this was not the case and the menu items were displayed on the menu in the order they were added with the hook, so we never really provided any weight to them. After the upgrade, since they had no weight, civi would ended up placing them in different places that we expected.

The affected items were "Help", "Developer" and "Self Service Portal". They are all added using the hook in the hrcore extension. The expected behavior is for these to always be the last items in the menu, in this exact order. So, to fix the issue, a `_hrcore_getMaxMenuWeight()` helper method was added in `hrcore.php` and it's now used to calculate the item weight which will always be "max weight + 1".

The ideal solution for this would be to actually store those items in the database instead of adding them at runtime, as this is how we're doing things now for new menus. That would required some major refactoring however for something that is out of the scope of the ticket. For that reason, I opted to simply adding the weight to the items.

### Removing "Mailings" from the main navigation

Due to a new permission in civi, "Mailings" started being displayed on the navigation menu. It was removed using `hook_civicrm_navigationMenu()`. Please check the commit message [here](https://github.com/compucorp/civihr/pull/2720/commits/fb26b85abec3f592fb43507f737bc16c4439cee1) for more details

### Small refactoring in the hrcore code used to handle the core menus

While fixing the menu issues mentioned above, I noticed some code duplication that could be easily fixed, so I did that and extracted a couple of helper functions [here](https://github.com/compucorp/civihr/pull/2720/commits/c08053d0bf36338b60e77116c8c1654a91fb5e4a)

### Duplicated Option Value values 

Since CiviCRM 4.7.28 (because of https://github.com/civicrm/civicrm-core/pull/11089), Civi doesn't allow two Option Values with the same value in one Option Group. Unfortunately, in hrjobcontract, there was some code that would result in an attempt to create a duplicated option value and this was causing errors during the installation. This was fixed in https://github.com/compucorp/civihr/pull/2720/commits/f8a4529fc038bd53956146714272d401cad58baf

### MySQL 5.7 fixes

Officially, CiviHR doesn't support MySQL 5.7 yet, but on the Jenkins server tests already run on that version and due to some changes in Civi, some errors started appearing while running the tests:

- Missing `created_date` in some custom groups: Some of the XML definition for Custom Groups in XML are quite old and missed the `created_date` field. Until, 4.7.27, this never caused any problem and the groups would be created just fine. After upgrading to 5.3, the creation started failing because for each missing date civi would translate that to a query like `INSERT INTO custom_group(...other_fields, created_date) VALUES(...other_values, 0)` and 0 is not valid date. I couldn't find the exact change in Civi that caused this, but it was fixed in CiviHR in https://github.com/compucorp/civihr/pull/2720/commits/5d5b301082f375eab737e4d17b4f69b45898f504
- Group by error when calling `WorkPattern.findWithNumberOfWeeksAndHours()`. In MySQL the `ONLY_FULL_GROUP_BY` option is enabled by default. It requires that a `GROUP BY` clause must list all the fields in the select clause. This resulted in problem with the mentioned method, as it had a `GROUP BY` clause with only one of the fields of the `SELECT` clause. This was fixed in  https://github.com/compucorp/civihr/pull/2720/commits/10813c63d7c87e6b32d4c8fad4b409e87cbbd53f and there is more information about it in the commit message. Note that disabling the `ONLY_FULL_GROUP_BY` option in the database would not work, as in  https://github.com/civicrm/civicrm-core/pull/10934 core was changed to make sure the option will ALWAYS be enabled for dev/test sites (i.e. sites created with buildkit)

### Sample data extension errors

After starting to run the tests with CiviCRM 5.3, the tests in the `hrsampledata` extension started failing with this error:

```



ncaught PEAR_Exception: DB Error: unknown error in unknown on line unknown

 DB_Error: DB Error: unknown error in unknown on line unknown

#0 [internal function]: CRM_Core_Error::exceptionHandler(Object(DB_Error))
#1 /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/packages/PEAR.php(921): call_user_func(Array, Object(DB_Error))
#2 /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/packages/DB.php(985): PEAR_Error->__construct('DB Error: unkno...', -1, 16, Array, 'CREATE TABLE `c...')
#3 /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/packages/PEAR.php(575): DB_Error->__construct(-1, 16, Array, 'CREATE TABLE `c...')
#4 [internal function]: PEAR->_raiseError(Object(DB_mysqli), NULL, -1, NULL, NULL, 'CREATE TABLE `c...', 'DB_Error', true)
#5 /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/packages/PEAR.php(224): call_user_func_array(Array, Array)
#6 /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/packages/DB/common.php(1905): PEAR->_ in /opt/buildkit/build/civihr-dev_PR-2720/sites/all/modules/civicrm/CRM/Core/Error.php on line 949

script returned exit code 255
```

This "unknow db error" is actually an error trying to create the table for the HRAbsencePeriod entity, in the old `hrabsence` extension (which was replaced by `hrleaveandabsences`).

The proper fix here is to actually update the sample data extension to use data from the new extension instead of the old one. We already started working on it in PCHR-3186 (Check the `PCHR-3186-fix-sample-data-extension` branch), but this was not complete yet. Because there's no point in fixing something that will be removed, I simply temporarily removed the sample data extension from the list of things to be tested by Jenkins.

### Installing Shoreditch and the Styleguide when running the Jenkins tests

During the site creation, we call the `drush-install.sh` script to install all of the civicrm extensions used in CiviHR. Among them, are the Styleguide and Shoreditch extensions which were not avaiable in the sites created by Jenkins. The extensions are installed using the `Extension.install` API through the `drush cvapi` command. The API will return an error when it tries to install an extension that doesn't exist, but until https://github.com/civicrm/civicrm-drupal/pull/503, the `drush cvapi` command would not fail because of that and the installation would continue without problems. Now, the command fails and to fix that we need to actually download the extensions when building a site with Jenkins.

### Other minor fix and improvements

- On https://github.com/compucorp/civihr/pull/2720/commits/172d06316b95711ecc1080a4b232c0d8c9e53d76 we avoid using drupal functions in a civicrm extension
- On https://github.com/compucorp/civihr/pull/2720/commits/bd84ae7cb8cfa00e23c908b39b871cd77bfb1b4e we avoid errors if the Style Guide extension is not installed
- On https://github.com/compucorp/civihr/pull/2720/commits/8760809277eb4bbaea4339c95ff0573380c52ff5 and https://github.com/compucorp/civihr/pull/2720/commits/d710d584c85793007ec6edd81e379e7be9201459 we fix errors checking for the values in the API response
- On https://github.com/compucorp/civihr/pull/2720/commits/07ef903708593ac8f6ff799844866f7eea43b8cd we add a test to the headless group to avoid warning while running the tests
- On https://github.com/compucorp/civihr/pull/2720/commits/6fbb126f11d3ed3ecb3c9fd1f9eb5e9c9f7096bd we fix the class names of some HRCore tests


hrjobcontract/CRM/Hrjobcontract/Dedupe/Merger.php
```diff
@@ -657,16 +712,11 @@ public static function getJoinOnDedupeTable() {
    *
    * @return string
    */
-  protected static function getWhereString($batchLimit, $isSelected) {
+  protected static function getWhereString($isSelected) {
```
mickadoo: It's probably not caused by this PR specifically, but I notice that `$batchLimit` was removed from the arguments but not the doc-block.

hrui/hrui.php
```diff
@@ -411,13 +411,17 @@ function hrui_setViewOptionsSetting($options = array()) {
...
+      'sequential' => 1
+    ]);
+  } catch(Exception $e) {
+    return;
   }
-  $user = user_load($uf['uf_id']);
-  $content['userid'] = $uf['uf_id'];
-  $content['username'] = !empty($user->name) ? $user->name : '';
+
+  $content['userid'] = $result['values'][0]['id'];
```
mickadoo: Is it possible that there is no `User` match here?

uk.co.compucorp.civicrm.hrcontactactionsmenu/hrcontactactionsmenu.php
```diff
@@ -67,8 +67,12 @@ function hrcontactactionsmenu_civicrm_pageRun(&$page) {
     CRM_Core_Resources::singleton()->addStyleFile($extName, 'css/contactactions.css');
     CRM_Core_Resources::singleton()->addScriptFile($extName, 'js/contactactions.js');
 
-    $templatePath = CRM_Core_Resources::singleton()->getPath($extName,'templates/CRM/HRContactActionsMenu/Page/Inline/Actions.tpl');
-    $page->assign('alternativeActionsTemplate', $templatePath);
+    CRM_Core_Region::instance('contact-page-inline-actions')->update('default', array(
```
mickadoo: I read the commit message, but maybe you could expand a bit on what's changed here in the PR description as I'm not so sure what it is.

Also we can use the `[]` array syntax in the above line

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -411,22 +411,43 @@ function _hrcore_coreMenuChanges(&$params) {
...
+  _hrcore_removeItemFromMainNavigationMenuByName('Redaction Rules',$params[$adminNavId]['child'][$civiCaseNavId]['child']);
+}
 
-  $civiCaseNavId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'CiviCase', 'id', 'name');
-  $redactionRulesNavId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'Redaction Rules', 'id', 'name');
-  $supportNavId = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Navigation', 'Support', 'id', 'name');
+/**
+ * Removes the item with the given $name from the root of the given $menu
+ *
+ * @todo Use the NavigationMenuHelper methods instead of these helper functions here
```
mickadoo: I did think of these, although your suggestion in the PR description to move these to storing in the database seems like a better long term goal. In that case we can probably remove the `@todo` unless you think we're going to act on it

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -439,6 +460,7 @@ function _hrcore_createHelpMenu(&$menu) {
   _hrcore_civix_insert_navigation_menu($menu, '', [
     'name' => ts('Help'),
     'permission' => 'access CiviCRM',
+    'weight' => _hrcore_getMaxMenuWeight($menu) + 1
```
mickadoo: Can you add a comma to the end of all these new items, it helps in the future when reviewing diffs

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Hook/BuildForm/ContactAdvancedSearchTest.php
```diff
@@ -2,6 +2,11 @@
 use CRM_HRCore_Test_BaseHeadlessTest as BaseHeadlessTest;
 use CRM_HRCore_Hook_BuildForm_ContactAdvancedSearch as ContactAdvancedSearch;
 
+/**
+ * Class CRM_HRCore_Hook_BuildForm_ContactAdvancedSearchTest
```
mickadoo: I don't think we usually have this `Class` docblock

hrjobcontract/CRM/Hrjobcontract/Dedupe/Merger.php
davialexandre: Good catch. I actually missed that one

hrui/hrui.php
davialexandre: When the User API can't find a match, it will throw an error, which will be catched by the throw-catch block and this line won't be reached.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Hook/BuildForm/ContactAdvancedSearchTest.php
davialexandre: From a quick look, it seems that most of the test classes actually do have that. It's added automatically by PhpStorm and since we don't really have a rule about it I'd prefer to keep this one for now, until we decide if we should have it or not and them apply that in all places

uk.co.compucorp.civicrm.hrcontactactionsmenu/hrcontactactionsmenu.php
davialexandre: Good point. I completely forgot to mention that in the PR description, but this is now fixed

uk.co.compucorp.civicrm.hrcore/hrcore.php
davialexandre: I kept the `@todo`, but not it also mentions the possibility of saving the menu changes to the database

uk.co.compucorp.civicrm.hrcontactactionsmenu/hrcontactactionsmenu.php
mickadoo: Thanks, that clears it up

PR #2722, opened 2018-06-27
PCHR-3559: Delete Cost Centre Other Option

ajesamson:
## Overview
This PR deletes `Other` option value from option group cost centre for new site. In existing setups, the value is deleted if not in use.

## Before
The option value `Other` is created by default on the system.

## After
The option value is either removed based on job roles status

## Technical Details
The system is first checked for existence of job roles data
```php
$jobRoles = civicrm_api3('HrJobRoles', 'get');
```
after which `Other` option is deleted if no record exist.
```php
civicrm_api3('OptionValue', 'get', [
  'option_group_id' => 'cost_centres',
  'api.OptionValue.delete' => [
    'id' => '$value.id',
    'name' => 'Other'
  ],
]);
```
For systems with job roles data, `other` option is deleted when not in use
```php
if (! $inUse) { // disables cost centre other
  $this->deleteCostCentreOther()
 }
```

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -80,6 +80,66 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
 
     return TRUE;
   }
+  
+  /**
+   * Deletes cost centre "other" option value or disable it if not in use
+   *
+   * @return bool
+   */
+  public function upgrade_1006() {
+    $jobRoles = civicrm_api3('HrJobRoles', 'get');
+    if ($jobRoles['count'] == 0) {
+      $this->deleteCostCentreOther();
+    }
+    
+    $otherId = $this->retrieveCostCentreOtherId();
+    $inUse = FALSE;
+    $roles = $jobRoles['values'];
+    $pattern = '/\|' . $otherId . '\|/';
+    foreach ($roles as $role) {
+      if (preg_match($pattern, $role['cost_center'])) {
+        $inUse = TRUE;
+        break;
+      }
+    }
+    
+    if (!$inUse) { // disables cost centre other
+      civicrm_api3('OptionValue', 'get', [
+        'option_group_id' => 'cost_centres',
+        'name' => 'Other',
+        'api.OptionValue.create' => ['id' => '$value.id', 'is_active' => 0],
+      ]);
+    }
+    
+    return TRUE;
+  }
+  
+  /**
+   * Deletes cost centre option value with name "other"
+   */
+  private function deleteCostCentreOther() {
+    civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'cost_centres',
+      'api.OptionValue.delete' => [
+        'id' => '$value.id',
+        'name' => 'Other'
+      ],
+    ]);
+  }
+  
+  /**
+   * Fetches the id of cost center "other" option value
+   *
+   * @return int
+   */
+  private function retrieveCostCentreOtherId() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'cost_centres',
+      'name' => 'Other',
+    ]);
+    
+    return $result['id'];
+  }
 
   /**
    * Creates a navigation menu item using the API
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -139,7 +139,7 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
       'name' => 'Other',
     ]);
     
-    return ($result['id'])? $result['id'] : null;
+    return isset($result['id'])? $result['id'] : null;
   }
 
   /**
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -82,7 +82,7 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
   }
   
   /**
-   * Deletes cost centre "other" option value or disable it if not in use
+   * Deletes cost centre "other" option value if not in use
    *
    * @return bool
    */
@@ -90,6 +90,7 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
     $jobRoles = civicrm_api3('HrJobRoles', 'get');
     if ($jobRoles['count'] == 0) {
       $this->deleteCostCentreOther();
+      return TRUE;
     }
     
     $otherId = $this->retrieveCostCentreOtherId();
@@ -106,12 +107,8 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
       }
     }
     
-    if (!$inUse) { // disables cost centre other
-      civicrm_api3('OptionValue', 'get', [
-        'option_group_id' => 'cost_centres',
-        'name' => 'Other',
-        'api.OptionValue.create' => ['id' => '$value.id', 'is_active' => 0],
-      ]);
+    if (!$inUse) {
+      $this->deleteCostCentreOther();
     }
     
     return TRUE;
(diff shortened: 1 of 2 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1277,24 +1283,6 @@ public function upgrade_1037() {
...
     return TRUE;
   }
-
-  /**
-   * Renames option group title from job contract end reason
-   * to contract end reasons
-   *
-   * @return bool
-   */
-  public function upgrade_1041() {
```
ajesamson: The function was not properly positioned after merging into the parent ticket. That is why it was refactored.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -515,6 +515,12 @@ public function upgradeBundle() {
     $this->upgrade_1034();
     $this->upgrade_1035();
     $this->upgrade_1036();
+    $this->upgrade_1037();
```
ajesamson: The call to the upgrader were missing from  previous PR for upgrade `1037 - 1041`.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1348,6 +1336,84 @@ public function upgrade_1040() {
...
+      'option_group_id' => 'cost_centres',
+      'api.OptionValue.delete' => [
+        'id' => '$value.id',
+        'name' => 'Other'
+      ],
+    ]);
+  }
+  
+  /**
+   * Fetches the id of cost center other option value
```
mickadoo: It's just a small point, but in places where you mention "other" option value, can you put "other" in quotes? It's a bit confusing otherwise

ajesamson: Noted. thanks.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1348,6 +1336,84 @@ public function upgrade_1040() {
...
+    $pattern = '/(\|' . $otherId . '\|)/';
+    foreach ($roles as $role) {
+      if (preg_match($pattern, $role['cost_center'])) {
+        $inUse = TRUE;
+        break;
+      }
+    }
+    
+    if (! $inUse) { // disables cost centre other
+      civicrm_api3('OptionValue', 'get', [
```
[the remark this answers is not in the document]
ajesamson: Yes i could have used `$otherId`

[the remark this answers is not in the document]
ajesamson: After using the `$otherId` and setting is_active, the upgrader failed after test site was restored and the upgrader ran. The error states that a compulsory field `option_group_id` was missing, prompting revert to previous implementation.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
davialexandre: @ajesamson how come some many upgraders were missing? Are we not testing this changes in new sites?

ajesamson: They are new upgraders recently added but was omitted in the previous PRs.

davialexandre: @ajesamson but why were they omitted? The only reason I can think of is that this is not being properly tested by devs locally

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -80,6 +80,66 @@ public function upgrade_1005() {
...
+  
+  /**
+   * Deletes cost centre option value with name "other"
+   */
+  private function deleteCostCentreOther() {
+    civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'cost_centres',
+      'api.OptionValue.delete' => [
+        'id' => '$value.id',
+        'name' => 'Other'
```
davialexandre: Why do you need to get all the option values if you only want to delete one of them? What about doing something like this:

```php
civicrm_api3('OptionValue', 'get', [
   'option_group_id' => 'cost_centres',
   'name' => 'Other',
   'api.OptionValue.delete' => [
      'id' => '$value.id'
   ]
]);
```

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -80,6 +80,66 @@ public function upgrade_1005() {
...
+   *
+   * @return int
+   */
+  private function retrieveCostCentreOtherId() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'cost_centres',
+      'name' => 'Other',
+    ]);
+    
+    return $result['id'];
```
davialexandre: This method can fail for multiple reasons:

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -80,6 +80,66 @@ public function upgrade_1005() {
...
+      }
+    }
+    
+    if (!$inUse) { // disables cost centre other
+      civicrm_api3('OptionValue', 'get', [
+        'option_group_id' => 'cost_centres',
+        'name' => 'Other',
+        'api.OptionValue.create' => ['id' => '$value.id', 'is_active' => 0],
+      ]);
+    }
```
davialexandre: Just out of curiosity, did anyone ever questioned why we have this? If it is not being used, why can't we delete it? If we don't delete them, it means new sites will be different than existing ones, which is not a good idea. I'd like to know if there's a reason for that

ajesamson: It was actually specified that way as part of the requirement. I will have to discuss with @mi-kon

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
ajesamson: Noted

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -138,7 +139,7 @@ private function retrieveCostCentreOtherId() {
       'name' => 'Other',
     ]);
     
-    return $result['id'];
+    return ($result['id'])? $result['id'] : null;
```
davialexandre: This will still throw an error, won't it? Perhaps you meant something like this: `!empty($result['id']) ? $result['id'] : null`

ajesamson: What i meant is `isset`

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
ajesamson: They were mistakenly omitted. Test was properly done locally as upgrades were carried out successfully and log also indicating their upgrade number without having to do call to them from the installer. After reviewing the page implementation again, it was then discovered they should have been triggered by the installer.

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -80,6 +80,64 @@ public function upgrade_1005() {
...
+  public function upgrade_1006() {
+    $jobRoles = civicrm_api3('HrJobRoles', 'get');
+    if ($jobRoles['count'] == 0) {
+      $this->deleteCostCentreOther();
+      return TRUE;
+    }
+    
+    $otherId = $this->retrieveCostCentreOtherId();
+    if ($otherId == null) {
+      return FALSE;
```
davialexandre: If this returns false the upgrade will fail. Is this really what you want?

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -80,6 +80,64 @@ public function upgrade_1005() {
...
+   *
+   * @return int
+   */
+  private function retrieveCostCentreOtherId() {
+    $result = civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'cost_centres',
+      'name' => 'Other',
+    ]);
+    
+    return isset($result['id'])? $result['id'] : null;
```
davialexandre: Missing space before the `?`

PR #2727, opened 2018-06-30
PCHR-3926: Add karma spec reporter to JS unit tests

AkA84:
## Overview
This PR introduces an improvement to the way JS unit tests results are displayed in the console, by implementing the [karma-spec-reporter](https://github.com/mlex/karma-spec-reporter) package

Quoting @reneolivo 
>  This helps me see the big picture for the file I'm trying to test.
> It also helps me organize tests better since I can read the whole spec for a file.

## Before
Output of `gulp test --ext hrjobcontract`
```
Executed 114 of 114 SUCCESS (0.732 secs / 0 secs)
```

## After
Output of `gulp test --ext hrjobcontract`

after

## Technical Details
The standard way to define the `reporters` list is to do that in the `karma.conf.js` file
```js
// karma.conf.js
config.set({
  reporters: ['spec', 'progress', 'junit']
});
```
Defining the "spec" report in the file didn't work initially though, as any reporters list in the config file would get overridden by the `karma.Server()` invocation in _gulp/test.js_'s `runServer()`
```js
function runServer (configFile, cb) {
 var reporters = argv.reporters ? argv.reporters.split(',') : ['progress'];

  new karma.Server({
    configFile: configFile,
    reporters: reporters,
    singleRun: true
  })
```
The function would pass the list of reporters as an explicit and distinct property to `runServer()`, separate from the whatever is defined in the config file. This was done in order to allow Jenkins to define a different set of reporters that the default ones via a `--reporters` argument
```
// Jenkinsfile
def testJS(hrcoreFolder, java.util.LinkedHashMap extension) {
  echo "JS Testing ${extension.name}"

  sh """
    cd $CIVICRM_EXT_ROOT/civihr/${hrcoreFolder}
    gulp test --ext ${extension.folder} --reporters junit,progress || true
  """
}
```

This PR keeps the ability to define reporters via a CLI argument, but moves the logic in the `karma.conf.js` files, to keep things as standard and intuitive as possible

```js
// gulp
function runServer (configFile, cb) {
  new karma.Server({
    configFile: configFile,
    singleRun: true
  });
```
```js
// a karma.conf file
var argv = require('yargs').argv;

config.set({
  reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
});
```
## Comments
There is a lot of repetition due to the fact that every single extension has its separate `karma.conf.js` file. This should be addressed in a future PR where we will have instead one single conf file that gets adapted dynamically by whichever extension is currently being tested

### DIFF
--- a/com.civicrm.hrjobroles/js/karma.conf.js
+++ b/com.civicrm.hrjobroles/js/karma.conf.js
@@ -70,6 +70,9 @@ module.exports = function (config) {
       }
     },
     reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
+    specReporter: {
+      suppressSkipped: true
+    },
     junitReporter: {
       outputDir: extPath + 'test-reports',
       useBrowserName: false,
--- a/contactaccessrights/js/karma.conf.js
+++ b/contactaccessrights/js/karma.conf.js
@@ -72,6 +72,9 @@ module.exports = function (config) {
       }
     },
     reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
+    specReporter: {
+      suppressSkipped: true
+    },
     junitReporter: {
       outputDir: extPath + 'test-reports',
       useBrowserName: false,
--- a/contactsummary/js/karma.conf.js
+++ b/contactsummary/js/karma.conf.js
@@ -73,6 +73,9 @@ module.exports = function (config) {
       }
     },
     reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
+    specReporter: {
+      suppressSkipped: true
+    },
     junitReporter: {
       outputDir: extPath + 'test-reports',
       useBrowserName: false,
--- a/hrjobcontract/js/karma.conf.js
+++ b/hrjobcontract/js/karma.conf.js
@@ -77,6 +77,9 @@ module.exports = function (config) {
       }
     },
     reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
+    specReporter: {
+      suppressSkipped: true
+    },
     junitReporter: {
       outputDir: extPath + 'test-reports',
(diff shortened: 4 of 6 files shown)


deb1990:
In hrcore's package.json we have `karma-junit-reporter`, do we need it?

com.civicrm.hrjobroles/js/karma.conf.js
```diff
@@ -67,6 +69,7 @@ module.exports = function (config) {
         ]
       }
     },
+    reporters: argv.reporters ? argv.reporters.split(',') : ['spec'],
```
deb1990: Why dont we keep
`reporters: ['spec'],` 
here in the `karma.conf.js` and in ` uk.co.compucorp.civicrm.hrcore/gulp/test.js`
```js
var reporters = argv.reporters ? argv.reporters.split(',') : configFile.reporters;

  new karma.Server({
    configFile: configFile,
    reporters: reporters,
    singleRun: true
  }, function () {
    cb && cb();
  }).start();
```
This way the repetitive logic can be minimized or almost eliminated and it would still be standard because the "normal" reporter is defined in karma.conf.js but only part related to overriding via jenkins is moved to the main gulpfile.

AkA84: Please see the comment

> There is a lot of repetition due to the fact that every single extension has its separate karma.conf.js file. This should be addressed in a future PR where we will have instead one single conf file that gets adapted dynamically by whichever extension is currently being tested

I don't expect that `reporters` line to be change any time soon, so the repetition won't be a hindrance in the meantime, until we have a unified karma.conf file

deb1990:
@AkA84 should not we keep the suppressed tests?
incase we skip a whole file by mistake, there would be no way to know?

PR #2730, opened 2018-07-02
PCHR-3556: Fixed Regression Issue with Adding Deduction

ajesamson:
## Overview
This PR fixes issues with Job Contract Deduction page management after the [previous](https://github.com/compucorp/civihr/pull/2716) upgrade.

## Before
before_deductions

## After
after_deductions

## Technical Details
Updating `option_group` [requires](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/BAO/OptionGroup.php#L95) setting the `is_active` attribute as it defaults to `false` if not set.


mickadoo:
@ajesamson can you check if this is the same on master in CiviCRM core? If not it seems like something that should probably be fixed in core too.

PR #2723, opened 2018-06-27
PCHR-3874: Add feeds dropdown button component

igorpavlov:
## Overview

This PR adds a Calendar Feeds Dropdown Button to SSP My Leave.

## Before

Button simply does not exist.

image

## After

When there are feeds:

When there are no feeds:

When on mobile:

### New model

APIs, Models and Instances are moved from https://github.com/compucorp/civihr/pull/2696

### Amendment of Feed Link Modal service.

The service that opens the Feed Link Modal should accept only the **feed hash** and not the whole URL, it should define the **host URI** itself:

```js
function open (hash) { // accept hash only, not the whole URL
  // full path is now defined inside the service
  var url = HOST_URL + 'calendar-feed?hash=' + hash; 

  $uibModal.open({ // ...
```

```js
// The host constant is defined in a module
return angular.module('calendar-feeds.link-modal', [
  // ...
])
  // Global Drupal object is used to the the host URI
  .constant('HOST_URL', Drupal.absoluteUrl('/'))
```

### Dropdown Button module

Dropdown Button module with a component created.

ℹ️The component accepts the `dropdown-position` parameter that defines the attachment position of the dropdown.

#### Usage

```html
<calendar-feeds-dropdown-button
  dropdown-position="right|left">
</calendar-feeds-dropdown-button>
```

#### Implementation

```js
return {
  // ...
  bindings: {
    dropdownPosition: '@'
  },
  // ...
};

function CalendarFeedsDropdownButtonController (CalendarFeed, CalendarFeedsLinkModal) {
  // ...
  vm.feeds = []; // stores feeds
  vm.loading = {
    feeds: false // indicates feeds being loaded
  };

  vm.openLinkModal = openLinkModal; // can open link modals

  (function init () {
    loadFeeds(); // loads feed on init
  }());

  // ...

  function openLinkModal (feedHash) { // opens link modal by feed hash
    CalendarFeedsLinkModal.open(feedHash);
  }
}
```

```html
<!-- Standard dropdown -->
<div class="btn-group" uib-dropdown>
  <button type="button" class="btn btn-default btn-lg dropdown-toggle"
    uib-dropdown-toggle>
    <span class="btn-icon"><i class="fa fa-rss-square"></i></span>
    Calendar Feeds
  </button>
  <!-- Dropdown attachment position is configurable -->
  <ul class="dropdown-menu dropdown-menu-{{dropdownButton.dropdownPosition}}"
[...]

PR #2733, opened 2018-07-02
PCHR-3908: Display link modal on calendar feed list

reneolivo:
## Overview
This PR adds the functionality to display the calendar feeds link modal on the calendar feeds list when either clicking the configuration title or "View feed link".

## Before
calendar feeds hr17 9200

## After
anim

## Technical Details

A new list module was created. This module includes a "display link" directive (more on this in a bit). This module is also exported as a minified file by adding the module name to the `build.js` file, and also creating a `calendar-feeds.list.js` file that serves as the root requirejs file and angular bootstraping:

```js
(function (require) {
  require([
    'common/angular',
    'leave-absences/calendar-feeds/list/calendar-feeds.list.module'
  ],
  function (angular) {
    angular.bootstrap(
      document.querySelector('[data-leave-absences-calendar-feeds-list]'),
      ['calendar-feeds.list']
    );
  });
})(require);
```

The "View link" is a directive that is activated by class since the "View feed link" element only accepts classes changes, not attributes. The strategy was to add the class to the "View feed link" element, the title, and the hash value to the parent TR element so it's accessible to both:

```tpl
<tr id="LeaveRequestCalendarFeedConfig-{$row.id}" class="crm-entity {$row.class}{if NOT $row.is_active} disabled{/if}" data-hash="{$row.hash}"> <!-- hash value is provided here -->
  <td data-field="title">
    <a href="#" class="calendar-feeds-display-link">{$row.title|escape}</a> <!-- adds the class to the title -->
  </td>
  <!-- ... -->
  <td>{$row.action|replace:'xx':$row.id}</td> <!-- the "View feed link" is auto generated here -->
</tr>
```

The PHP generated "View feed link" has the `calendar-feeds-display-link` added in this manner:

```php
CRM_Core_Action::BASIC => [
  'name' => ts('View Feed Link'),
  'class' => 'calendar-feeds-display-link', // here
   'title' => ts('View Feed Link'),
],
```

Finally, a "display-link" directive was added as follows:

```js
function DisplayLinkFn ($scope, $element, $attr, $ctrl) {
  (function () {
    initEventListeners();
  })();

  function initEventListeners () {
    $element.on('click', function (event) {
      var hash = $element.parents('[data-hash]').attr('data-hash');

      linkModal.open(hash);
    });
  }
}
```

As a side note, the list had an extra TH tag. This was removed.

Spec files were also moved to the test folder and reverted the change that made spec files to be available from the source folder.


igorpavlov:
@reneolivo in the After section please make a note that it will show the full URL to the feed once it is rebased with my PR https://github.com/compucorp/civihr/pull/2723

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/LeaveRequestCalendarFeedConfig.php
```diff
@@ -48,7 +48,11 @@ public function browse() {
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
     CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/hrleaveandabsences.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
+    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/angular/dist/calendar-feeds.list.min.js', 1000);
```
igorpavlov: Why do we hardcode the weight here? If we need to set the weight opposing to `CRM_Core_Resources::DEFAULT_WEIGHT`, we may want to do smth like this:

```CRM_Core_Resources::DEFAULT_WEIGHT + 1```
or
```CRM_Core_Resources::DEFAULT_WEIGHT - 1```

depending on what do you need. That will be better for readability and will justify the choice of the weight.

File: uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/build.js [hunk 2]
```diff
@@ -5,6 +5,7 @@
   modules: [
     {name: 'admin-dashboard'},
     {name: 'absence-tab'},
+    {name: 'calendar-feeds.list'},
```
igorpavlov: Could you please add spaces to objects: `{ name: 'calendar-feeds.list' }` + other objects?

The semistandard will still complain on the "expected expression" issue (or smth like that), you may then use `--no-verify` flag to bypass as an exception.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/calendar-feeds/link-modal/calendar-feeds-link-modal.service.js
```diff
@@ -17,6 +17,12 @@ define(function () {
      * @param {String} url - the url to display on the modal.
      */
     function open (url) {
+      var appendToElement = $document.find('#bootstrap-theme');
```
igorpavlov: I thought modals are styled properly outside namespaces, am I wrong?

Anyway, if we absolutely must do that, there should be a comment to the function I think:

```js
/**
 * Opens a modal with the calendar feed link.
 * @NOTE it injects the modal inside the #bootstrap-theme, if exists, otherwise inside <body>
 *
 * @param {String} url - the url to display on the modal.
 */
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/test/calendar-feeds/link-modal/calendar-feeds-link-modal.service.spec.js
```diff
@@ -50,6 +55,35 @@ define([
       it('passes the dismiss function to the link modal component', function () {
         expect(calendarFeedsLinkModalComponent.dismiss).toEqual(jasmine.any(Function));
       });
+
+      describe('when there is a bootstrap theme element', function () {
+        beforeEach(function () {
+          $uibModal.open.calls.reset();
+          $document.find('body').append('<div id="bootstrap-theme"></div>');
+          CalendarFeedsLinkModal.open(url);
```
igorpavlov: You don't pass an URL anymore, you pass a **hash**.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/LeaveRequestCalendarFeedConfig.php
AkA84: i think the important thing is that the weight of this file is greater than the weight of `reqangular.min.js`, rather than the default weight

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/calendar-feeds/link-modal/calendar-feeds-link-modal.service.js
AkA84: does this work in the SSP? There is no `#bootstrap-theme` element in it

reneolivo: @igorpavlov @AkA84 modals are not displayed properly outside `#bootstrap-theme`, they look like a random block of text/content that appears at the top of the page.

They look OK in other admin sections such as Leave Calendar or Tasks because those sections import bootstrap through SCSS or they are including `bootstrap.css`. They also look OK in SSP because Bootstrap is globally available through SSP Theme.

In this case, instead of importing Bootstrap so it's globally available I decided to attach the modal to `#bootstrap-theme` element when available.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/LeaveRequestCalendarFeedConfig.php
reneolivo: 1001 was decided since reqangular defines a 1000 value.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/calendar-feeds/link-modal/calendar-feeds-link-modal.service.js
AkA84: i agree that the documentation should explicitly reference what's going on with `#bootstrap-theme` and more importantly _why_

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/LeaveRequestCalendarFeedConfig.php
```diff
@@ -48,7 +48,11 @@ public function browse() {
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
     CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/hrleaveandabsences.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
+    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/angular/dist/calendar-feeds.list.min.js', 1001);
     CRM_Core_Resources::singleton()->addScriptFile('civicrm', 'js/jquery/jquery.crmEditable.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
+    CRM_Core_Resources::singleton()->addVars('leaveAndAbsences', [
+      'baseURL' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.hrleaveandabsences'),
```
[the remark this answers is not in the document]
igorpavlov: @reneolivo or you can consult to @tunbola since I just merged a PR https://github.com/compucorp/civihr/pull/2743 that removes local permissions because they are now centralised, as per @tunbola suggestion.

reneolivo:
> @reneolivo in the After section please make a note that it will show the full URL to the feed once it is rebased with my PR #2723

@igorpavlov updated the screenshot instead.

uk.co.compucorp.civicrm.hrleaveandabsences/hrleaveandabsences.php
```diff
@@ -483,6 +483,19 @@ function hrleaveandabsences_addContactMenuActions(ActionsMenu $menu){
...
+ */
+function hrleaveandabsences_civicrm_pageRun(&$page) {
+  // Adds the baseURL var to all L&A pages
+  CRM_Core_Resources::singleton()->addVars('leaveAndAbsences', [
+    'attachmentToken' => CRM_Core_Page_AJAX_Attachment::createToken(),
+    'baseURL' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.hrleaveandabsences'),
+    'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer'),
+    'loggedInUserId' => CRM_Core_Session::getLoggedInContactID(),
+  ]);
+}
```
davialexandre: This will include these variables in ALL civicrm pages. Is this really what you want?

reneolivo: This is what I'm not sure of. There is a positive side effect though: we wouldn't need to manually add the `leaveAndAbsences` var to specific pages when they are implementing a module from L&A.

Something else that we could do is have a list of pages that the var is allowed to exist, like this:

```php
$allowedPages = [
 ClassName_of_the_Page1,
 ClassName_of_the_Page2,
];

if (!in_array($page, $allowedPages)) {
  return;
}

// ...
```

davialexandre: If there's no need to use any of these vars outside L&A pages, then I'd prefer to have them available only in the pages they will be used. If the requirements change in the future, then we can also change the solution.

As for the hook implementation, here's a suggestion. We've been trying to move logic out of the hooks into separate classes. So it would be nice if you could do that here. Check this example: https://github.com/compucorp/civihr/blob/staging/uk.co.compucorp.civicrm.hrcore/hrcore.php#L287-L293

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Hook/PageRun/AddLeaveAndAbsencesVarTest.php
```diff
@@ -0,0 +1,28 @@
...
+    $hook->handle($page);
+
+    $actualVars = CRM_Core_Resources::singleton()->getSettings()['vars'];
+
+    $this->assertRegexp('/[A-z0-9_;]+/', $actualVars['leaveAndAbsences']['attachmentToken']);
+
+    unset($actualVars['leaveAndAbsences']['attachmentToken']);
+
+    $this->assertEquals($expectedVars, $actualVars);
+  }
```
davialexandre: Could you please also add a negative test, to show it will not add the vars to pages it should not handle?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Hook/PageRun/AddLeaveAndAbsencesVar.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+class CRM_HRLeaveAndAbsences_Hook_PageRun_AddLeaveAndAbsencesVar {
```
davialexandre: For the class name, it's usually better to go for a noun instead of verbs. I'd suggest to change `AddLeaveAndAbsencesVar` to something like `LeaveAndAbsencesVarsAdder`

PR #2737, opened 2018-07-03
PCHR-3561: Change Options for Sickness Reason

ajesamson:
## Overview
This PR adds new set of sickness reason option value after either deleting (for new systems) or disabling (for old systems) the existing option values. The new set of sickness reasons are listed below:
* Cold, Cough, Flu - Influenza
* Headache/Migraine
* Gastro-intestinal Problems
* Back Problems
* Injury, Fracture
* Dental and Oral Problems
* Pregnancy Related
* Anxiety/Stress/Depression/Other Psychiatric Illnesses
* Chest and Respiratory Problems
* Ear, Nose, Throat (ENT)
* Endocrine/Glandular Problems
* Eye Problems
* Genitourinary and Gynaecological Disorders
* Other Musculoskeletal Problems
* Surgery Related
* Other

## Before
Existing option values are available after setup.

## After
After running upgrade, only the newly required options are available for selection.

## Technical Details
Leave request was checked for existing record before deleting or deactivating existing options.
```php
$result = civicrm_api3('LeaveRequest', 'get', [
  'sickness_reason' => ['IS NOT NULL' => 1],
]);
if ($result['count'] == 0) {
  $this->up1030_deleteOldSicknessReasons($oldSicknessReasons['values']);
}
else {
  $this->up1030_disableOldSicknessReasons($oldSicknessReasons['values']);
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
@@ -16,12 +16,12 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1030 {
       'option_group_id' => 'hrleaveandabsences_sickness_reason',
     ]);
   
-    $result = $this->up1030_getLeaveRequestTotal();
+    $result = $this->up1030_getTotalSicknessReasonRequests();
     if ($result == 0) {
-      $this->up1030_deleteOldSicknessReasons($oldSicknessReasons['values']);
+      $this->up1030_deleteOptionValues($oldSicknessReasons['values']);
     }
     else {
-      $this->up1030_disableOldSicknessReasons($oldSicknessReasons['values']);
+      $this->up1030_disableOptionValues($oldSicknessReasons['values']);
     }
     
     $this->up1030_addNewSicknessReasons();
@@ -34,7 +34,7 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1030 {
    *
    * @return int
    */
-  private function up1030_getLeaveRequestTotal() {
+  private function up1030_getTotalSicknessReasonRequests() {
     $leaveRequest = new LeaveRequest();
     $leaveRequest->whereAdd('sickness_reason IS NOT NULL');
     $leaveRequest->find();
@@ -48,7 +48,7 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1030 {
    *
    * @param array $sicknessReasons
    */
-  private function up1030_disableOldSicknessReasons($sicknessReasons) {
+  private function up1030_disableOptionValues($sicknessReasons) {
     foreach ($sicknessReasons as $id => $sicknessReason) {
       civicrm_api3('OptionValue', 'create', [
         'id' => $id,
@@ -62,7 +62,7 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1030 {
    *
    * @param array $sicknessReasons
    */
-  private function up1030_deleteOldSicknessReasons($sicknessReasons) {
+  private function up1030_deleteOptionValues($sicknessReasons) {
     foreach ($sicknessReasons as $id => $sicknessReason) {
       civicrm_api3('OptionValue', 'delete', ['id' => $id]);
     }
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
```diff
@@ -0,0 +1,105 @@
...
+    
+    return TRUE;
+  }
+  
+  /**
+   * Counts the total number of leave requests related to sickness reasons
+   *
+   * @return int
+   */
+  private function up1030_getLeaveRequestTotal() {
```
mickadoo: I don't think the method name is very descriptive here, sounds like the total number of leave requests, but actually it's the total number that have a sickness reason

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
```diff
@@ -0,0 +1,105 @@
...
+    // N is Number of rows returned from a query
+    return $leaveRequest->N;
+  }
+  
+  /**
+   * Disables existing "sickness reason" option values
+   *
+   * @param array $sicknessReasons
+   */
+  private function up1030_disableOldSicknessReasons($sicknessReasons) {
```
mickadoo: I'm not sure if "Old" in the method here helps. It's already clear from where the method is called that these are `$oldSicknessReasons`, the method doesn't care about this though

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
```diff
@@ -0,0 +1,105 @@
...
+      ]);
+    }
+  }
+  
+  /**
+   * Deletes existing "sickness reason" option values
+   *
+   * @param array $sicknessReasons
+   */
+  private function up1030_deleteOldSicknessReasons($sicknessReasons) {
```
mickadoo: Same as above, I'm not sure "old" means anything. In fact both this method and `up1030_disableOldSicknessReasons` would probably be clearer just as `up1030_disableOptionValues` and `up1030_deleteOptionValues`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_LeaveRequest as LeaveRequest;
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1030 {
```
mickadoo: General question, we try to make all our upgraders idempotent, I think if you run this upgrader again it will repeatedly delete all option values for this group and recreate them. Do you think it would be possible to make sure it only deletes those outside of the desired options and doesn't recreate them if they exist?

ajesamson: Good idea. I will do the check.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
ajesamson: Noted. I will update appropriately.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
ajesamson: Noted. I will update appropriately.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
ajesamson: I will rename. Thanks

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1030.php
mickadoo: I see that you added a check so the option value will not be created if it exists. This is a positive step but I think you should consider what happens if you run the upgrader again

```
// this will be all the option values (the new ones if run twice)
$oldSicknessReasons = civicrm_api3('OptionValue', 'get', [
  'return' => ['name'],
  'option_group_id' => 'hrleaveandabsences_sickness_reason',
]);
// these will then be disabled or deleted
```

I don't think it's worth spending a lot of time on this, because in reality upgraders will only be run once. But we have had bugs where they were run more than once, and I think it's a good strategy to make them idempotent so please keep this in mind when working on them in the future

PR #2738, opened 2018-07-03
PCHR-3599: Deletes Some Extended Demographic Fields

backstageel:
## Overview
Deletes some Extended Demographic Fields, if they are not used. 

Also prevents its creation on new Installations

## Before
screenshot from 2018-07-03 18 33 05

## After
screenshot from 2018-07-03 18 32 26

## Technical Details
`$query = 'SELECT * FROM ' . $customGroup['table_name'] . ' WHERE ' . $customField['column_name'] . ' NOT LIKE "%Not Applicable%" AND ' . $customField['column_name'] . ' IS NOT NULL AND ' . $customField['column_name'] . ' <> ""';`
This query will return all items where the value is is not null, or is different of "Not Applicable" or is not empty. The '%Not Applicable%' part is because in my installation some fields has non-ascii characters arround some 'Not Applicable' string.


mickadoo:
@backstageel in the description there is a typo 

> This query will return all itens 

Also could you explain a bit more about

> in my installation some fields has strange characters in this field.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,48 @@
...
+
+    return TRUE;
+  }
+
+  /**
+   * Deletes Custom Demographic Fields only if they are not used
+   *
+   * @param $fieldsToDelete
+   */
+  private function deleteExtendDemographicFields($fieldsToDelete) {
```
mickadoo: For upgraders we usually start method names with the "up" + version, e.g. `up1025_deleteExtendedDemographicFields`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,48 @@
...
+      'Sexual_Orientation',
+    ]);
+
+    return TRUE;
+  }
+
+  /**
+   * Deletes Custom Demographic Fields only if they are not used
+   *
+   * @param $fieldsToDelete
```
mickadoo: missing param type

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,48 @@
...
+  private function deleteExtendDemographicFields($fieldsToDelete) {
+    $customGroup = civicrm_api3('CustomGroup', 'get', [
+      'name' => ['LIKE' => 'Extended_Demographics'],
+    ]);
+    $customGroup = array_shift($customGroup['values']);
+    $customFields = civicrm_api3('CustomField', 'get', [
+      'name' => ['IN' => $fieldsToDelete],
+    ]);
+    foreach ($customFields['values'] as $customField) {
+      $query = 'SELECT * FROM ' . $customGroup['table_name'] . ' WHERE ' . $customField['column_name'] . ' NOT LIKE "%Not Applicable%" AND ' . $customField['column_name'] . ' IS NOT NULL AND ' . $customField['column_name'] . ' <> ""';
```
mickadoo: This isn't very readable, how about using `sprintf` and also creating some variables instead of accessing by array every time, for example:

```
$tableName = $customGroup['table_name'];
$queryFormat = 'SELECT COUNT(id) FROM "%s"'
  . ' WHERE "%s" NOT LIKE "%Not Applicable%"'
  . ' AND "%s" IS NOT NULL'
  . ' AND "%s" <> ""';

foreach ($customFields['values'] as $customField) {
  $column = $customField['column_name'];
  $query = sprintf($queryFormat, $tableName, $column, $column, $column);
```

Creating variables helps keep the lines shorter and makes the code more readable.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,48 @@
...
+      $dao = CRM_Core_DAO::executeQuery($query);
```
mickadoo: I changed the query in my comment above to use `SELECT COUNT(id)` which is probably more performant, you might need to update the `fetchAll` method too

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,48 @@
...
+      $isCustomFieldUsed = $dao->fetchAll();
```
mickadoo: I don't think this makes sense for this variable name, `fetchAll` is returning a list of entities. I think it would make sense if you put the `empty` check around this on this line instead, i.e.

```
$isCustomFieldUsed = !empty($dao->fetchAll());
```

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Deletes Custom Demographic Fields only if they are not used
+   *
+   * @param array $fieldsToDelete
+   */
+  private function up1025_deleteExtendDemographicFields($fieldsToDelete) {
+    $customGroup = civicrm_api3('CustomGroup', 'get', [
+      'name' => ['LIKE' => 'Extended_Demographics'],
```
mickadoo: Why did you use `LIKE` here? The custom group name should always be the same

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1025.php
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Deletes Custom Demographic Fields only if they are not used
+   *
+   * @param array $fieldsToDelete
+   */
+  private function up1025_deleteExtendDemographicFields($fieldsToDelete) {
+    $customGroup = civicrm_api3('CustomGroup', 'get', [
+      'name' =>'Extended_Demographics',
```
mickadoo: Tiny, tiny thing, but you're missing a space after `=>`

backstageel: Corrected before merge

PR #2739, opened 2018-07-03
PCHR-3887: Add "Add feed" button to feed dropdown

igorpavlov:
## Overview

This PR adds a "Add new feed" option-button to the Calendar Feeds Dropdown Button component.

## Before

No such option in the dropdown.

image

## After

A new option (appears only for those who has a permission).

## Technical Details

Firstly, we need to check if user has a permission to add a feed, we use `checkPermissions` service for that:

```js
function defineIfCanCreateFeeds () {
  return checkPermissions('can administer calendar feeds')
    .then(function (canAdministerCalendarFeeds) {
      vm.canCreateNewFeed = !!canAdministerCalendarFeeds;
```

Then we add HTML to the component:

```html
<li role="separator" class="divider" ng-if="dropdownButton.canCreateNewFeed"></li>
<li ng-if="dropdownButton.canCreateNewFeed">
  <a href="/civicrm/admin/leaveandabsences/calendar-feeds?action=add&reset=1"
    target="_blank">
    <i class="fa fa-plus"></i> Add new feed
```

### Custom CSS style

According to the mockups the new option should be separated from calendar feeds. However, the bootstrap divider has top and bottom margins, so we need to suppress the margins in this specific component and not globally.

```css
calendar-feeds-dropdown-button {
  .dropdown-menu .divider {
    margin: 0;
```

### Fetching only active (enabled) feeds

```js
return this.sendGET(
  'LeaveRequestCalendarFeedConfig', 'get', {
    is_active: true
  }, false)
```

### Do not force "No cache"

Previously we called API with  `...{}, false`, but now we don't pass the `false`. This is intentional as it does not serve any purpose for the current functionality, we do not need to force "no to cache" behaviour. 

------
✅Manual Tests - passed
⏹Backstop tests - omitted
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/calendar-feeds/services/apis/calendar-feed.api.js
```diff
@@ -21,7 +21,8 @@ define([
      * @return {Promise} resolved with an array of objects of feeds
      */
     function all () {
-      return this.sendGET('LeaveRequestCalendarFeedConfig', 'get', {}, false)
+      return this.sendGET('LeaveRequestCalendarFeedConfig', 'get',
+        { is_active: true })
```
reneolivo: the PR description and previous code had `'LeaveRequestCalendarFeedConfig', 'get', {}, false` (false), but now we don't. Is this intentional? (Not event sure what this "false" means.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_calendar-feeds.scss
```diff
@@ -1,3 +1,9 @@
+calendar-feeds-dropdown-button {
+  .dropdown-menu .divider {
+    margin: 0;
```
reneolivo: @igorpavlov I'm a bit worried about removing the margin here. If the mockup has no margin, should it be the same for every other dropdown with a divider? Why is this an exception? I think we need to agree what's the rule here and apply it globally or just remove this and let Bootstrap do its thing.

uk.co.compucorp.civicrm.hrleaveandabsences/js/angular/src/leave-absences/calendar-feeds/services/apis/calendar-feed.api.js
igorpavlov: Yes, intentional, it forces not saving cached api response. I don't think this is necessary, so I removed that. I updated the commend section.

uk.co.compucorp.civicrm.hrleaveandabsences/scss/components/_calendar-feeds.scss
AkA84: can you show me here screenshots of what it looks like now in this PR and what it look like if we remove this exception and let Bootstrap do its thing

igorpavlov: Without extra styling:

image

With extra styling:

image

igorpavlov: Decision: SSP - remove custom styling

igorpavlov: We have custom styling in JobContract:

```css
#hrjob-contract .dropdown-menu .divider {
  height: 1px;
  margin: 8px 0;
  overflow: hidden;
  background-color: #e5e5e5;
}
```

I don't think the SSP styles will clash with ^.

However, the only place where there is a **markup** with a divider is in reqangular `ui-select` directive. So I think we are safe.

igorpavlov: After styles are removed from SSP theme the dropdown looks like this:

image

PR #2741, opened 2018-07-04
PCHR-3897: Add Calendar Feed dropdown button to Admin Dashboard

igorpavlov:
## Overview

This PR adds the Calendar Feed Dropdown Button to Admin Dashboard. 

## Before

The button does not yet exist on the Admin Dashboard page.

image

## After

The button is injected and behaves exactly as seen at https://github.com/compucorp/civihr/pull/2723.

image

## Technical Details

The button module was used and the component was injected:

```html
<div class="admin-dashboard-page">
  <div class="crm_page__topbar">
    <!-- here. note the dropdown position, this time is has to be on the left -->
    <calendar-feeds-dropdown-button dropdown-position="left">
    <div class="pull-right">
      <leave-request-record-actions
        contact-id="$root.settings.contactId">
```

## Comments

Because we now need to position two elements in the top bar (Feed Dropdown button + Leave Request Record Dropdown), it is better to move the **margin** styling from the buttons themselves to the top bar's **padding**:

```css
.crm_page__topbar {
  height: 65px;
  /* in this case we don't need to style each button */
  padding: #{$content-container-vertical-margin} #{$content-container-side-margin};
```

-----

✅Manual Tests - passed
⏹Jasmine Tests - not needed
⏹Backstop tests - omitted
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/Dashboard.php
```diff
@@ -13,6 +13,7 @@ public function run() {
         'administer leave and absences',
         'access leave and absences in ssp',
         'manage leave and absences in ssp',
+        'can administer calendar feeds'
```
reneolivo: Sorry to meddle with this, but, would this block admins from seeing the whole dashboard just because they can't administer calendar feeds?

tunbola: It won't prevent the admin inasmuch as the admin still has the required permission (`access leave and absences`) to view the page

PR #2742, opened 2018-07-04
PCHR-3608: Makes Contract Length Read Only

backstageel:
## Overview
 Makes contract length read only and hides it from Personal Details Screen

## Before
screenshot from 2018-07-04 15 21 22

## After
screenshot from 2018-07-04 15 20 20

## Technical Details
To hide the field just changed the 'is_active' field of CustomGroup to 0

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1268,23 +1268,15 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
    * @return bool
    */
   public function upgrade_1037() {
-    $customField = civicrm_api3('CustomField', 'get', [
+    civicrm_api3('CustomField', 'get', [
       'name' => 'Length_Of_Service',
+      'api.CustomField.create' => ['id' => '$value.id', 'is_view' => 1],
     ]);
-    civicrm_api3('CustomField', 'create', [
-        'id' => $customField['id'],
-        'is_view' => 1,
-      ]
-    );
 
-    $customGroup = civicrm_api3('CustomGroup', 'get', [
+    civicrm_api3('CustomGroup', 'get', [
       'name' => 'Contact_Length_Of_Service',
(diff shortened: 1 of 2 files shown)


hrjobcontract/xml/length_of_service.xml
```diff
@@ -19,23 +19,5 @@
     </CustomGroup>
   </CustomGroups>
   <CustomFields>
-    <CustomField>
```
ajesamson: Are you sure the requirement is remove the `custom_field`? I think this will remove rather than setting the field to view only.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1261,6 +1262,33 @@ public function upgrade_1036() {
     return TRUE;
   }
 
+  /**
+   * Updates the custom field to view only and disables the custom group
+   *
+   * @return bool
+   */
+  public function upgrade_1037() {
+    $customField = civicrm_api3('CustomField', 'get', [
```
ajesamson: Can we make use of chained API call here?
```php
civicrm_api3('CustomField', 'get', [
  'return' => ['id']
  'name' => 'Length_Of_Service',
  'api.CustomField.create' => ['id' => '$value.id', 'is_view' => 1],
]);
```
Much neater than separating the snippet.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1261,6 +1262,33 @@ public function upgrade_1036() {
...
+      'name' => 'Length_Of_Service',
+    ]);
+    civicrm_api3('CustomField', 'create', [
+        'id' => $customField['id'],
+        'is_view' => 1,
+      ]
+    );
+
+    $customGroup = civicrm_api3('CustomGroup', 'get', [
```
ajesamson: Same thing with chained call here please.
```php
civicrm_api3('CustomGroup', 'get', [
  'return' => ['id'],
  'name' => 'Contact_Length_Of_Service',
  'api.CustomGroup.create' => ['id' => '$value.id', 'is_active' => 0]
]);
```

PR #2745, opened 2018-07-05
PCHR-3903: Implement Visibility Logic for LeaveRequestCalendarFeedConfig.get API

tunbola:
## Overview
Currently, when configuring a LeaveRequestCalendarFeedConfig in the create/update form, you can set who the feed should be visible to based on department and location that the contact should belong to. This PR implements this logic in the LeaveRequestCalendarFeedConfig.get API by checking for the departments and locations that the current user belongs and using this information to determine whether the user has access to fetch a feed configuration or not. Admins (users with the (`can administer calendar feeds` permission) can view all feed configurations, while users without these permissions can only view the feed configuration based on the feed visibility settings.

## Before
- The visibility logic is not implemented for the LeaveRequestCalendarFeedConfig.get API, therefore all users can view all feeds.

## After
- The visibility logic is now implemented for the LeaveRequestCalendarFeedConfig.get API, only users with visibility for a feed configuration can view such while the Admin can view all feed configurations.

## Technical Details
The `addSelectWhereClause` method was implemented in the `LeaveRequestCalendarFeedConfig` BAO to check the the departments and locations the current user has access to. The list of feeds with limited visibility settings is fetched and the ones the current user does not have access to is excluded from the final results.

``` php
public function addSelectWhereClause() {
    if (CRM_Core_Permission::check([['can administer calendar feeds']])) {
      return;
    }

    $limitedVisibilityFeedConfigs = $this->getLimitedVisibilityFeedConfigs();

    if (empty($limitedVisibilityFeedConfigs)) {
      return;
    }

    $accessibleDepartments = $this->getAccessibleDepartmentsForLoggedInContact();
    $accessibleLocations = $this->getAccessibleLocationsForLoggedInContact();

    $inaccessibleFeedConfigs = $this->getNotAccessibleFeedConfigs(
      $limitedVisibilityFeedConfigs,
      $accessibleDepartments,
      $accessibleLocations
    );

    if (empty($inaccessibleFeedConfigs)) {
      return;
    }

    $clauses['id'] = 'NOT IN (' . implode(',', $inaccessibleFeedConfigs) . ')';

    CRM_Utils_Hook::selectWhereClause($this, $clauses);

    return $clauses;
  }
```

### DIFF
--- a/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
+++ b/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
@@ -128,13 +128,13 @@ class api_v3_ContactHRJobRoleTest extends CRM_Hrjobroles_Test_BaseHeadlessTest {
     civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => [$operator => [1]]]);
   }
 
-  public function testGeDoesNotThrowExceptionWhenUsingTheINOperatorForTheContactIdParameter() {
+  public function testGetDoesNotThrowExceptionWhenUsingTheINOperatorForTheContactIdParameter() {
     $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => ['IN' => [1]]]);
 
     $this->assertEquals(0, $values['is_error']);
   }
 
-  public function testGeDoesNotThrowExceptionWhenUsingTheEqualsToOperatorForTheContactIdParameter() {
+  public function testGetDoesNotThrowExceptionWhenUsingTheEqualsToOperatorForTheContactIdParameter() {
     $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => 1]);
 
     $this->assertEquals(0, $values['is_error']);
--- a/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
+++ b/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
@@ -118,6 +118,64 @@ class api_v3_ContactHRJobRoleTest extends CRM_Hrjobroles_Test_BaseHeadlessTest {
     $this->assertEquals($expectedJobRole3, $contactJobRoles[$jobRole3['id']]);
   }
 
+  /**
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage The contact_id parameter only supports the IN operator
+   *
+   * @dataProvider invalidGetContactIdOperators
+   */
+  public function testGetThrowsExceptionForOperatorThatIsNotTheINOperatorForContactIDParameter($operator) {
+    civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => [$operator => [1]]]);
+  }
+
+  public function testGeDoesNotThrowExceptionWhenUsingTheINOperatorForTheContactIdParameter() {
+    $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => ['IN' => [1]]]);
+
+    $this->assertEquals(0, $values['is_error']);
+  }
+
+  public function testGeDoesNotThrowExceptionWhenUsingTheEqualsToOperatorForTheContactIdParameter() {
+    $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => 1]);
+
+    $this->assertEquals(0, $values['is_error']);
+  }
+
+  public function testGetReturnsJobRolesForPassedInContacts() {
+    $contact1 = ContactFabricator::fabricate();
+    $contact2 = ContactFabricator::fabricate();
+    $contact3 = ContactFabricator::fabricate();
+    $contract1 = HRJobContractFabricator::fabricate(['contact_id' => $contact1['id']]);
+    $contract2 = HRJobContractFabricator::fabricate(['contact_id' => $contact2['id']]);
+    $contract3 = HRJobContractFabricator::fabricate(['contact_id' => $contact3['id']]);
+    $jobRole1 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id']]);
+    $jobRole2 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract2['id']]);
+    $jobRole3 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract3['id']]);
+
+    //Get Jobroles for contact1 and contact3
+    $contactJobRoles = civicrm_api3($this->entity, $this->action, [
+      'contact_id' => ['IN' => [$contact1['id'], $contact3['id']]]
+    ])['values'];
+
+    $this->assertCount(2, $contactJobRoles);
+
+    // The CiviCRM only returns non-empty fields. Since we only set the job
+    // contract id, only the ID and the contact_id fields will be returned
+    $expectedJobRole1 = ['id' => $jobRole1['id'], 'contact_id' => $contact1['id']];
+    $expectedJobRole3 = ['id' => $jobRole3['id'], 'contact_id' => $contact3['id']];
+
+    $this->assertEquals($expectedJobRole1, $contactJobRoles[$jobRole1['id']]);
+    $this->assertEquals($expectedJobRole3, $contactJobRoles[$jobRole3['id']]);
+
+    //Get Jobroles for contact2 only
+    $contactJobRoles = civicrm_api3($this->entity, $this->action, [
+      'contact_id' => $contact2['id']
+    ])['values'];
+
+    $this->assertCount(1, $contactJobRoles);
+    $expectedJobRole2 = ['id' => $jobRole2['id'], 'contact_id' => $contact2['id']];
+    $this->assertEquals($expectedJobRole2, $contactJobRoles[$jobRole2['id']]);
+  }
+
   private function setExpectedApiPermissionException($permission) {
     $message = "API permission check failed for {$this->entity}/{$this->action} call; insufficient permission: require {$permission}";
     $this->setExpectedException('CiviCRM_API3_Exception', $message);
@@ -126,5 +184,23 @@ class api_v3_ContactHRJobRoleTest extends CRM_Hrjobroles_Test_BaseHeadlessTest {
   private function registerCurrentLoggedInContactInSession($contactID) {
     $session = CRM_Core_Session::singleton();
     $session->set('userID', $contactID);
+  }
+
+  public function invalidGetContactIdOperators() {
+    return [
+      ['>'],
+      ['>='],
+      ['<='],
+      ['<'],
+      ['<>'],
+      ['!='],
+      ['BETWEEN'],
+      ['NOT BETWEEN'],
+      ['LIKE'],
+      ['NOT LIKE'],
+      ['NOT IN'],
+      ['IS NULL'],
+      ['IS NOT NULL'],
+    ];
   }
 }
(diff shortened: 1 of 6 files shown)


com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
```diff
@@ -118,6 +118,64 @@ public function testTheGetActionReturnsMultipleJobRoles() {
...
+   * @expectedException CiviCRM_API3_Exception
+   * @expectedExceptionMessage The contact_id parameter only supports the IN operator
+   *
+   * @dataProvider invalidGetContactIdOperators
+   */
+  public function testGetThrowsExceptionForOperatorThatIsNotTheINOperatorForContactIDParameter($operator) {
+    civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => [$operator => [1]]]);
+  }
+
+  public function testGeDoesNotThrowExceptionWhenUsingTheINOperatorForTheContactIdParameter() {
```
mickadoo: `s/testGe/testGet/`

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
```diff
@@ -118,6 +118,64 @@ public function testTheGetActionReturnsMultipleJobRoles() {
...
+    $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => ['IN' => [1]]]);
+
+    $this->assertEquals(0, $values['is_error']);
+  }
+
+  public function testGeDoesNotThrowExceptionWhenUsingTheEqualsToOperatorForTheContactIdParameter() {
```
mickadoo: `s/testGe/testGet/`

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
```diff
@@ -118,6 +118,64 @@ public function testTheGetActionReturnsMultipleJobRoles() {
...
+    $values = civicrm_api3('ContactHrJobRoles', 'get', ['contact_id' => 1]);
+
+    $this->assertEquals(0, $values['is_error']);
+  }
+
+  public function testGetReturnsJobRolesForPassedInContacts() {
```
mickadoo: nice clear tests here :+1:

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -232,4 +232,185 @@ public static function getValuesArray($id) {
...
+    $limitedVisibilityFeedConfigs = $this->getLimitedVisibilityFeedConfigs();
+
+    if (empty($limitedVisibilityFeedConfigs)) {
+      return;
+    }
+
+    $accessibleDepartments = $this->getAccessibleDepartmentsForLoggedInContact();
+    $accessibleLocations = $this->getAccessibleLocationsForLoggedInContact();
+
+    $inaccessibleFeedConfigs = $this->getNotAccessibleFeedConfigs(
```
mickadoo: since you're using `inaccessible`  here it might make more sense to rename the `getNotAccessibleFeedConfigs` to use `inaccessible` too

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -232,4 +232,185 @@ public static function getValuesArray($id) {
...
+      $accessibleLocations
+    );
+
+    if (empty($inaccessibleFeedConfigs)) {
+      return;
+    }
+
+    $clauses['id'] = 'NOT IN (' . implode(',', $inaccessibleFeedConfigs) . ')';
+
+    CRM_Utils_Hook::selectWhereClause($this, $clauses);
```
mickadoo: Just curious, why are you calling the hook here? Is this done in the parent? If it is then is it an option to do a `parent::addSelectWhereClause` here instead?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -232,4 +232,185 @@ public static function getValuesArray($id) {
...
+
+  /**
+   * Gets the departments the logged in user has access to from the list of
+   * active departments.
+   *
+   * @return array
+   */
+  private function getAccessibleDepartmentsForLoggedInContact() {
+    $departments = $this->getDepartmentsList();
+    return $this->getLoggedInContactJobRoleAccessFor('department', $departments);
```
mickadoo: I think usually we have an empty line before `return`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -232,4 +232,185 @@ public static function getValuesArray($id) {
...
+   * Returns the Feed Configs that the logged in user does not have visibility
+   * access to based on the given parameters.
+   *
+   * @param array $limitedVisibilityFeedConfigs
+   * @param array $accessibleDepartments
+   * @param array $accessibleLocations
+   *
+   * @return array
+   */
+  public function getNotAccessibleFeedConfigs($limitedVisibilityFeedConfigs, $accessibleDepartments, $accessibleLocations) {
```
mickadoo: For long lines like this I think it's better to split the method signature like

```
public function getNotAccessibleFeedConfigs(
  $limitedVisibilityFeedConfigs,
  $accessibleDepartments,
  $accessibleLocations
) {
```

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -232,4 +232,185 @@ public static function getValuesArray($id) {
...
+      return;
+    }
+
+    $limitedVisibilityFeedConfigs = $this->getLimitedVisibilityFeedConfigs();
+
+    if (empty($limitedVisibilityFeedConfigs)) {
+      return;
+    }
+
+    $accessibleDepartments = $this->getAccessibleDepartmentsForLoggedInContact();
```
mickadoo: These methods are all really well done, but I wonder if maybe it would be cleaner to put this logic into a separate class, for example `$leaveRequestFeedConfigService->getInaccessibleConfigsForContact($contactID);` or 

```
$allConfigs = $leaveRequestFeedConfigService->findAll();
$accessibleConfigs = $leaveRequestFeedConfigService->findAllAccessibleToContact($contactId);
// do some diff of the IDs to get inaccessible ones
```

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/api/v3/LeaveRequestCalendarFeedConfigTest.php
```diff
@@ -27,4 +40,288 @@ public function testGetReturnsUnSerializedValuesForFilterFields() {
...
+        'location' => [3]
+      ],
+      'visible_to' => [
+        'department' => [1,2],
+        'location' => [1]
+      ]
+    ];
+
+    $params = array_merge($defaultParameters, $params);
+    return LeaveRequestCalendarFeedConfig::create($params);
```
mickadoo: empty line before `return`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
tunbola: I think it would be a bit too much to create a separate class for this when the methods are only used by the LeaveRequestCalendarFeedConfig. addSelectWhereClause method only now. But If we ever need to get available feed configs for a user anytime I agree it would be good to move to a class then.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
tunbola: I tried adding parent::addSelectWhereClause but the tests were failing even though the parent calls the hook. I could take a closer look later, although we have this same implementation for the LeaveRequest BAO.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
mickadoo: That's fine, I don't think there's a need to change it now. But in the future I would say it makes things easier to test and reduces class size if we encapsulate logic inside a class, _even if_ that class is only used in one place.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -251,14 +251,7 @@ public function addSelectWhereClause() {
...
 
-    $accessibleDepartments = $this->getAccessibleDepartmentsForLoggedInContact();
-    $accessibleLocations = $this->getAccessibleLocationsForLoggedInContact();
-
-    $inaccessibleFeedConfigs = $this->getNotAccessibleFeedConfigs(
-      $limitedVisibilityFeedConfigs,
-      $accessibleDepartments,
-      $accessibleLocations
-    );
+    $inaccessibleFeedConfigs = $this->getInAccessibleFeedConfigsForCurrentUser($limitedVisibilityFeedConfigs);
```
mickadoo: Sorry, one minor thing - `getInAccessible` isn't "Get In Accessible" should be 'getInaccessibleFeedConfigsForCurrentUser`

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
```diff
@@ -271,6 +264,26 @@ public function addSelectWhereClause() {
...
 
+  /**
+   * Gets the ids of the feeds that are not accessible to the current user from the
+   * list of feed configs passed in.
+   *
+   * @param array $limitedVisibilityFeedConfigs
+   *
+   * @return array
+   */
+  private function getInAccessibleFeedConfigsForCurrentUser($limitedVisibilityFeedConfigs) {
```
mickadoo: Two small things here, maybe change the name to `getInAccessibleFeedConfigIDsForCurrentUser` and you can just 

```
return $this->getInAccessibleFeedConfigs(...);
```

Instead of creating a new variable

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/LeaveRequestCalendarFeedConfig.php
tunbola: Good point

PR #2748, opened 2018-07-09
PCHR-3947: Improve Feed Link Modal UI/UX

igorpavlov:
## Overview

This PR introduces multiple UI/UX improvements to the Feed Link Modal.

## Before

1. Copy button does not react to click, you don't know what's happening
2. Value to be copied can be changed, but there is no reason for that
3. When clicking on the input with a value to be copied, it does not auto-select all text
4. "Done" button appears on the right
5. Hint sentence lacks question mark
6. Modal misses feed title

## After

1. Copy button changes to "Copied!" for 2 seconds
2. Value to be copied cannot be changed, only copied
3. When clicking on the input with a value to be copied, it auto-selects all text
4. "Done" button appears on the left
5. Hint sentence has a question mark
6. Modal shows feed title

### Admin Dashboard

image

### SSP

image

### Copy button click reaction

```js
function showJustCopied () {
  vm.justCopied = true;
  
  // If you click Copy again before it reverted to normal state
  justCopiedTimeout && $timeout.cancel(justCopiedTimeout);

  justCopiedTimeout = $timeout(function () {
    vm.justCopied = false;
  }, 2000); // revert to original state in 2 seconds
}
```

### Do not allow to change the link

The `readonly` attribute does the job here:

```html
<input ... readonly>
```

### Select all text on input click

```html
<input ... ng-click="input.selectInputText()">
```

in combination with:

```js
function selectInputText () {
  input.select();
}
```

does the job.

### Pass feed title to the modal

The implementation as see here https://github.com/compucorp/civihr/pull/2733 was simply extended with a new `title` attribute.

-----

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/LeaveRequestCalendarFeedConfig.tpl
```diff
@@ -23,7 +23,10 @@
             <th>{ts}Actions{/ts}</th>
             </thead>
             {foreach from=$rows item=row}
-              <tr id="LeaveRequestCalendarFeedConfig-{$row.id}" class="crm-entity {$row.class}{if NOT $row.is_active} disabled{/if}" data-hash="{$row.hash}">
+              <tr id="LeaveRequestCalendarFeedConfig-{$row.id}"
+                class="crm-entity {$row.class}{if NOT $row.is_active} disabled{/if}"
+                data-title="{$row.title}"
```
reneolivo: Do you think we might need to use `{$row.title|escape}` instead of just `{$row.title}` ? I have the suspicion that this would help in case the user types something like `Calendar Feed "Users taking leave"` (weird, but could happen).

igorpavlov: Woah, it is not escaped by default? Let me test that.

igorpavlov: Bug caught! Done!

igorpavlov: image

PR #2753, opened 2018-07-10
PCHR-3600: Move HRDemog Fields in Contact Profile

ajesamson:
## Overview
This PR moves the Extended Demographics to a new position behind `Inline Custom Data` in `Personal Details` tab without showing the title.

## Before
before_hrdemog

## After
after_hrdemog

## Technical Details
The page `DOM` was manipulated on rendering to reposition the Extended Demographics details
```js
$('.Inline_Custom_Data')
  .....
  .insertAfter('.crm-summary-contactinfo-block')
  .....
```

-------
✅Manual Tests - passed
✅Jasmine Tests - passed


hrui/js/src/hrui.js
```diff
@@ -283,6 +284,19 @@
       .find('.crm-summary-block').show();
   }
 
+  /**
+   * Moves the "Extended Demographics" block after "Inline Custom Data"
+   * in the personal details tab
+   */
+  function repositionExtendedDemographicsInPersonalDetailsTab () {
```
AkA84: this function has a lot of code in common with `repositionInlineCustomDataBlockInPersonalDetailsTab`
```js
$(<selector-to-move>)
  .removeClass('crm-collapsible collapsed')
  .addClass('crm-summary-block')
  .insertAfter(<selector-to-target>)
  .find('.collapsible-title').hide().end()
  .find('.crm-summary-block').show();
```
meaning that we can make a more generic function that we can then use for both cases
```js
function manipulateDOMOfInlineCustomData () {
  // ...
  commonFunction('.Inline_Custom_Data', '.crm-s mmary-contactinfo-block');
  commonFunction('.Extended_Demographics', '.Inline_Custom_Data');
}
```
tricky part is to find a decent name for it, something like `repositionPersonalDetailsBlock`?

hrui/js/src/hrui.js
```diff
@@ -189,6 +189,7 @@
   function manipulateDOMOfInlineCustomData () {
     if ($('.Inline_Custom_Data').length) {
```
AkA84: this check was necessary because otherwise `repositionInlineCustomDataBlockInPersonalDetailsTab` wouldn't have anything to move, but why is it necessary to do that check for the new function? What if `. Inline_Custom_Data` is not present? Shouldn't we move the "Extended Demographic" block anyway?

ajesamson: You are correct. I initially thought `Inline_Custom_Data` was a determinant to the page rendering before reviewing other section of the code.

hrui/js/src/hrui.js
ajesamson: Good idea 👍. I will refactor it.

hrui/js/src/hrui.js
```diff
@@ -188,9 +188,11 @@
    */
   function manipulateDOMOfInlineCustomData () {
     if ($('.Inline_Custom_Data').length) {
-      repositionInlineCustomDataBlockInPersonalDetailsTab();
+      repositionPersonalDetailsBlock('.Inline_Custom_Data', '.crm-summary-contactinfo-block');
     }
 
+    repositionPersonalDetailsBlock('.Extended_Demographics', '.Inline_Custom_Data');
```
AkA84: this line 
```js
if ($('.Inline_Custom_Data').length) {
```
implies that `.Inline_Custom_Data` might not exist. what then, in that case?

hrui/js/src/hrui.js
```diff
@@ -186,11 +186,14 @@
    * Manipulates, at the DOM level, the blocks/fields belonging to the
    * Inline Custom Data custom fields set
    */
-  function manipulateDOMOfInlineCustomData () {
-    if ($('.Inline_Custom_Data').length) {
-      repositionInlineCustomDataBlockInPersonalDetailsTab();
-    }
+  if ($('.Inline_Custom_Data').length) {
+    repositionPersonalDetailsBlock('.Inline_Custom_Data', '.crm-summary-contactinfo-block');
```
AkA84: I noticed that all we are doing is to _append_ these blocks to `.contactTopBar .contactCardLeft`

screen

if that's all we're doing, then we can simplify things

```js
if ($('.Inline_Custom_Data').length) {
  appendPersonalDetailsBlockToMainContactCard('.Inline_Custom_Data');
}

appendPersonalDetailsBlockToMainContactCard('.Extended_Demographics');
```

by using [`.appendTo()`](http://api.jquery.com/appendto/) we don't need to worry about what elements have already being moved, because anyway we're always appending at the end of the container

ajesamson: Better and much simplified without worrying about `Inline_Custom_Data`.

hrui/js/src/hrui.js
```diff
@@ -271,14 +273,15 @@
...
+   * @param selector
    */
-  function repositionInlineCustomDataBlockInPersonalDetailsTab () {
-    $('.Inline_Custom_Data')
+  function repositionPersonalDetailsBlock (selector) {
+    $(selector)
       .removeClass('crm-collapsible collapsed')
       .addClass('crm-summary-block')
-      .insertAfter('.crm-summary-contactinfo-block')
+      .appendTo('.contactCardLeft')
```
AkA84: please check the DOM you are manipulating. There are multiple `.contactCardLeft` in the page, that's why I wrote `.contactTopBar .contactCardLeft` in my comment

ajesamson: Oh i missed that comment. :(

PR #2762, opened 2018-07-16
PCHR-3975: Disable Prefer Not To Say Gender Option

backstageel:
## Overview
This PR add a function to  disable 'Prefer Not to Say' Gender Option on an existing Upgrader.
## Before
The Prefer Not to Say option was active by default

## After
The Prefer Not to Say Option will be disabled by this Upgrader

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1021.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1021.php
@@ -7,6 +7,7 @@ trait CRM_HRCore_Upgrader_Steps_1021 {
    */
   public function upgrade_1021() {
     $this->up1021_setDefaultGenderOptions();
+    $this->up1021_disableDefaultGenderOptions('Prefer not to say');
 
     return TRUE;
   }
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1021.php
```diff
@@ -41,4 +42,12 @@ private function up1021_setDefaultGenderOptions() {
     }
   }
 
+  private function up1021_disableDefaultGenderOptions($genderOption) {
+    civicrm_api3('OptionValue', 'get', [
```
tunbola: Are we sure this option would be present on all installations? For example it is not present in my local installation. In that case, it may be better to split the API calls, one to check if it exists, the other to disable.

tunbola:
Can you put the  Prefer Not to Say Gender in quotes in the PR description like so 'Prefer Not to Say Gender'

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1021.php
backstageel: I thought that this is implicit on Chained API calls, like if the Get returns nothing, nothing will be disabled. No?

tunbola: I think it will throw an exception when making the create call to disable because by then the Id will not exist. I've seen that type of exception somewhere. You can also test on local by removing the option temporarily/

tunbola: Since you confirmed it works fine after testing locally, Approving this then.

PR #2766, opened 2018-07-17
PCHR-3962: Transform L&A CRM JS files to AMD modules

igorpavlov-zz:
## Overview

This PR transforms the JS files in the L&A `/js/crm` folder to AMD modules and moves all these modules to the `js/src/leave-absences` folder for consistency.

## Before

Plain JS files that are injected as-is to the page.

image

```php
CRM_Core_Resources::singleton()->
  addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences',
    'js/crm/hrleaveandabsences.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
```

## After

The AMD modules that are built into a distributive and then injected onto pages:

image

```php
CRM_Core_Resources::singleton()->
  addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences',
    'js/dist/crm-hrleaveandabsences.min.js', 1001);
```

### Basic details

A goal is to provide more consistency across the FE code. The objectives are:
- Make AMD modules from the files inside the `js/crm` folder to allow to inject each other when needed
- Specify dependencies inside AMD and avoid injecting them directly into the browser via the `<script> tag
- Minify distributive files that are injected

The distributive files are created as usual, at first, the `build.js` specifies which distributive files are needed:

```js
/* eslint-disable */

({
  // ...
  modules: [
    // ...
    { name: 'crm-list-absenceperiod' },
    { name: 'crm-hrleaveandabsences' },
    { name: 'crm-form-manage-entitlements' },
    { name: 'crm-form-absenceperiod' },
    { name: 'crm-form-workpattern' },
    { name: 'spectrum' } // needs to be included without any additional code
```

Then original files were wrapped with AMD (eg. the Work Pattern Form file `js/src/leave-absences/crm/form-workpattern.js`):

```js
define([
  'leave-absences/crm/includes/hrleaveandabsences-init', // main app initiator
  'leave-absences/crm/vendor/inputmask/inputmask.numeric.extensions.min' // dependencies
], function (HRLeaveAndAbsencesApp) {
```
This allows to avoid separate dependencies inclusion like this:

```php
CRM_Core_Resources::singleton()->
  addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences',
    'js/dist/crm-form-workpattern.min.js', 1001);
```

Instead of:

```php
...'js/crm/vendor/inputmask.min.js'
...'js/crm/vendor/inputmask.numeric.extensions.min.js');
...'js/crm/hrleaveandabsences.form.workpattern.js');
```

### Main init file

```js
/* eslint-env amd */
define([], function () {
  // Initialises needed properties
  CRM.HRLeaveAndAbsencesApp = CRM.HRLeaveAndAbsencesApp || {};
  CRM.HRLeaveAndAbsencesApp.Form = CRM.HRLeaveAndAbsencesApp.Form || {};
  CRM.HRLeaveAndAbsencesApp.List = CRM.HRLeaveAndAbsencesApp.List || {};

  // Returns the app object to avoid using too much `CRM` variable in the app modules
  return CRM.HRLeaveAndAbsencesApp;
});
```

This also removes duplicating code from the app modules.

## Comments

Some code was refactored apart from the satisfying Semistandard.

### {literal} and <script>

There is no need to escape the `<script>` tag, just the code inside:

```html
<!-- No -->
{literal}<script> /* code */ </script>{/literal}
<!-- Yes -->
<script>{literal} /* code */ {/literal}</script>
```

You may also notice that even Github likes that more and highlights the JS and HTML (same for Atom editor).

### Indentation

Instead of:

```html
<div>
<script>
{literal}
/* code */
{/literal}
</script>
</div>
```
We now have:

```html
<div>
  <script>
    {literal}
      /* code */
    {/literal}
  </script>
</div>
```
This is better for readability of nested elements.

### Structure suggestions accepted

The https://github.com/compucorp/civihr/pull/2768 containing @AkA84 suggestions was merged (rebased) as is, then the commits from this PR were squashed into one.

Then the `.tpl` and `.php` files were amended in order to support new changes.

However, some changes have to be amended again to make them work.

#### Event names

Namespaced event **ready.something** will trigger on the `document.ready` event as well, so it cannot be used in our case.

```js
$(document).on('ready.togo', () => { /* ... */ });
$(document).trigger('ready.togo');
// The code inside will be executed twice:
// on document ready and then by trigger method call
```

So instead of `ready.` we use `hrappready.`

### Scripts loading order
**🛑The initial suggestion was REMOVED due to inability to properly review this change**

#### What has been done as a part of this PR?

Script loading weights are simply numbers, picked up consistently with other places where Reqangular has a weight of **1000** and AMD modules **1001**.

```php
// Example
CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.reqangular', 'dist/reqangular.min.js', 1000);
CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/crm-app-form-manage-entitlements.min.js', 1001);
```

#### What has been suggested initially?

In order to avoid hardcoded values for the order of scripts loading, eg:

```php
// All we need is to ensure this AMD module is loaded after Reqangular which contains RequireJS
  ->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences',
    'js/dist/crm-app-form-absenceperiod.min.js', 1010);
```

... we introduce a class with order constants:

```php
class CRM_HRCore_Helper_ScriptResource {

  const REQANGULAR_LOAD_WEIGHT = 1000;
  const AMD_MODULE_LOAD_WEIGHT = 1001;
}
```
... so it is used like this:

```php
  ->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences',
    'js/dist/crm-app-form-absenceperiod.min.js', 
    CRM_HRCore_Helper_ScriptResource::AMD_MODULE_LOAD_WEIGHT);
```
### Document.ready inside AMD modules

There is no need to check for the DOM to be ready since AMD modules are loaded at the end just before closing the `<body>` tag.

-------

✅Manual Tests - passed
⏹Jasmine Tests - no coverage
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
@@ -88,7 +88,7 @@ class CRM_HRLeaveAndAbsences_Form_ManageEntitlements extends CRM_Core_Form {
     $this->assign('returnUrl', $session->get('ManageEntitlementsReturnUrl'));
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
-    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.reqangular', 'dist/reqangular.min.js', 1000);
+    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.reqangular', 'js/dist/reqangular.min.js', 1000);
     CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/crm-app-form-manage-entitlements.min.js', 1001);
     parent::buildQuickForm();
   }


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
```diff
@@ -55,7 +55,7 @@ public function buildQuickForm() {
     $this->addButtons($this->getAvailableButtons());
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/hrleaveandabsences.form.absenceperiod.js');
+    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/crm-form-absenceperiod.min.js', 1001);
```
tunbola: Any reason why the weight is hardcoded here?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
```diff
@@ -58,7 +58,7 @@ public function buildQuickForm() {
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/spectrum.css');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/vendor/spectrum-min.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
+    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/spectrum.min.js', 1001);
```
tunbola: Why is weight hardcoded here also and why is the region changed?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
```diff
@@ -88,9 +88,8 @@ public function buildQuickForm() {
     $this->assign('returnUrl', $session->get('ManageEntitlementsReturnUrl'));
 
     CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/vendor/inputmask.min.js');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/vendor/inputmask.numeric.extensions.min.js');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/crm/hrleaveandabsences.form.manage_entitlements.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
```
tunbola: The removed scripts are no longer needed?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/PublicHoliday.tpl
```diff
@@ -24,37 +24,36 @@
...
+              e.preventDefault();
+              CRM.confirm({
+                title: ts('Delete Public Holiday'),
+                message: ts('Are you sure you want to delete this Public Holiday?'),
+                options: {
+                  yes: ts('Yes'),
+                  no: ts('No')
+                }
+              })
+              .on('crmConfirm:yes', REDIRECT_AFTER_DELETE_URL);
```
tunbola: No need for `window.location = deleteUrl`? , So the URL can be passed directly?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/WorkPattern/Calendar.tpl
```diff
@@ -100,19 +99,20 @@
       </table>
     </div>
   {/section}
-  {literal}
-    <script type="text/javascript">
-      CRM.$(function($) {
-        var form = new CRM.HRLeaveAndAbsencesApp.Form.WorkPattern({/literal}'{$delete_url}'{literal});
+  <script type="text/javascript">
+    {literal}
+      CRM.$(document).on('CRMWorkPatternFormScriptIsReady', function () {
```
tunbola: Is this `CRMWorkPatternFormScriptIsReady` generated by Civi automatically?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
igorpavlov: Yes, needs to appear after reqangular is injected.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsenceType.php
igorpavlov: Must appear after reqangular.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/ManageEntitlements.php
igorpavlov: Inputmask is now a dependency for `crm-form-manage-entitlements` and is now inside the dist file `js/dist/crm-form-manage-entitlements.min.js`

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/WorkPattern/Calendar.tpl
igorpavlov: No, this is a custom event, the app JS file is now async and needs to be waited before executing what's inside the `CRM.$(document).on`

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/PublicHoliday.tpl
igorpavlov: Not quite sure I understand. We have a constant REDIRECT_AFTER_DELETE_URL because it is better for readability so it does not mix smarty and JS too much.

tunbola: The initial implementation is this: https://github.com/compucorp/civihr/blob/e8fd6ef4adbec0b4a3f190b50da54c52677601e2/uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/PublicHoliday.tpl#L41, It is passed a callback that that redirects to the callback URL. Now the callback URL is passed directly.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
tunbola: Can we add a comment as to why this is so. A bit difficult to get that by merely looking at the code.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/PublicHoliday.tpl
igorpavlov: Oh snap! Fixing.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
igorpavlov: Can we define constants like this `CRM_Core_Resources::DEFAULT_WEIGHT`, but for example `CRM_Core_Resources::REQANGULAR` and `CRM_Core_Resources::AFTER_REQANGULAR`?

tunbola: That would be nice!

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/PublicHoliday.tpl
igorpavlov: Fixed

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Form/AbsencePeriod.php
igorpavlov: Done.

PR #2771, opened 2018-07-18
PCHR-3537: Add correspondence to the list of phone types to migrate

mickadoo:
## Overview

We encountered an issue when trying to migrate clients that used the "Correspondence" phone types. This fixes it.

## Before

Correspondence was missing from the list of phone types to migrate, this caused an error when trying to delete it when records existed that used it.

## After

Correspondence phone type is migrated to Personal before it is deleted.


mickadoo:
Closing this as it seems like it's probably not required, affected clients might be fixed manually

PR #2772, opened 2018-07-18
PCHR-3942: Standardize js folder of org.civirm.reqangular

AkA84:
## Overview
This PR standardizes the location of the `src/`, `dist/` and `test/` folders in the `org.civicrm.reqangular` extension, by moving them under a `js/` folder 

## Before
before

## After
after

### Simplified path resolution logic in gulp tasks
With this PR, all the extensions have now the same `js/` folder structure.

This means that any gulp task that requires finding files like `build.js`, `karma.conf.js`, etc, can simplify how it looks for them, given that there are no more folder exceptions

```js
// before
path.join(extPath, '**', 'src/**/*.js')
find.fileSync(/js(\/[^/]+)?\/build\.js$/)
find.fileSync('karma.conf.js', utils.getExtensionPath())[0]
findUp.sync('karma.conf.js', { cwd: testFile })

// after
path.join(extPath, 'js/src/**/*.js')
find.fileSync(/js\/build\.js$/)
path.join(utils.getExtensionPath(), 'js/karma.conf.js')
path.join(utils.getExtensionPath(), 'js');
```

### Updated `package-lock.json` files
This PR also updates the `package-lock.json` files of several extensions. This is not due to any dependencies that we use being updated, but simply to some of those dependencies now requiring new versions of their own dependencies, and the lock files need to reflect that

### Fixed a faulty job roles spec file
While testing the gulp tasks, i found that _get-active-values.filter.spec.js_ threw a "module not found" error when executed individually. This was due to the loading of the wrong AMD modules
```js
// before
define([
  'job-roles/filters/get-active-values.filter'
]);

// after
define([
  'job-roles/modules/job-roles.filters'
]);
```


org.civicrm.reqangular/src/common/vendor/angular/angular-animate.min.js.map
```diff
@@ -1,8 +0,0 @@
-{
```
igorpavlov: You delete map files, could you please explain why in the PR description?

org.civicrm.reqangular/js/test/test-main.js
```diff
@@ -1,10 +1,10 @@
 /* globals CRM */
 
-var TEST_REGEXP = /(spec|test)\.js$/i;
+var TEST_REGEXP = /spec\.js$/i;
 var allTestFiles = [];
 var extPath = '/base/tools/extensions/civihr/org.civicrm.reqangular';
-var mocksPath = extPath + '/test/mocks';
-var srcPath = extPath + '/src/common';
+var mocksPath = extPath + '/js/test/mocks';
```
igorpavlov: I think it may be safer to use `Path.join()` here

org.civicrm.reqangular/js/test/test-main.js
```diff
@@ -1,10 +1,10 @@
 /* globals CRM */
 
-var TEST_REGEXP = /(spec|test)\.js$/i;
+var TEST_REGEXP = /spec\.js$/i;
```
igorpavlov: May be  `/\.spec\.js$/i` would be safer to avoid matching files like `autospec.js`

igorpavlov:
I went to `civicrm` folder of the repo and searched for every appearance of `reqangular.min.js`.

image

I have a feeling that we will need to amend these paths as well.

org.civicrm.reqangular/src/common/vendor/angular/angular-animate.min.js.map
AkA84: good catch, i didn't notice this at all, i don't know why it happened. will check

AkA84: It happens in the first commit (https://github.com/compucorp/civihr/pull/2772/commits/76e23a38b1a1a5313a66eb0058752abe47064a6c), at which point our [.gitignore](https://github.com/compucorp/civihr/blob/staging/.gitignore) kicks in and removes them
```
# Exclude all JS source map files
*.js.map
```
in any case we are going to (hopefully) switch to use npm soon for vendor files, so i think this can be ignore

org.civicrm.reqangular/js/test/test-main.js
AkA84: this file is loaded in the browser when karma is launched, it's not executed in node, so `require('path')` won't work

PR #2775, opened 2018-07-19
PCHR-3941: Use RequireJS in all CiviHR extensions

deb1990:
## Overview
As part of this PR(and https://github.com/compucorp/civihr-employee-portal/pull/527, https://github.com/compucorp/civihr-tasks-assignments/pull/394), RequireJS has been used in all CiviHR extensions, and the build process has been updated accordingly. 

## Technical Details
For HR Career, HR Qual, HR IM, Contact Actions, HR CORE, HR UI extensions the following changes has been done
- Linting errors fixed
- RequireJS has been used in the JS files
- Folder structure updated according to styleguide
- Build process has been added using RequireJS
- Minified JS files has been added to the php file 

`hrcase/js/hrcase.js` was not being used, so the file has been removed.

**hrui & reqangular**
As `hrui` uses requirejs now, its necessary to load `reqangular.min.js` in every page where hrui is manipulating the dom. To achieve the same, requirejs library must be attached to every civicrm page. 
 - `reqangular_civicrm_coreResourceList` hook is added to achieve the same.
 - `require.min.js` is removed from `reqangular.min.js` as it is only necessary to attach the requirejs library to the civicrm pages and not the full `reqangular.min.js` file.

**hrui**
In `hrui/js/src/hrui/main.js`
Functions inside `crmLoad` event were not getting executed after applying requirejs, because `crmLoad` event fires before code inside requirejs runs. To avoid this problem the following functions has been moved to `document.ready`
- addUploadFileListener("input[type='file']");
- amendVacancyForm();
- amendContactPageAndForm(e);
- applyMiscChanges();
- changeContactSourceFieldHelpText();

`amendContactPageAndForm()` and `manipulateDOMOfInlineCustomData()` has been broken down into 3 separate functions based on the page they operate
- hideFieldsInPersonalDetailsTab()
- amendPersonalDetailsTab()
- amendContactsPage()

- [x] Manual tests pass


AkA84:
@deb1990 seems you missed `hrmed`

hrui/js/src/hrui.js
```diff
@@ -1,332 +1,10 @@
...
-    });
-  }
-}(CRM.$, CRM._));
+/* eslint-env amd */
+
+(function (require) {
+  require([
+    'hrui/main',
+    'hrui/contact',
+    'hrui/civihr-popup/attrchange.js',
```
AkA84: you don't need the `.js` suffix

hrcase/hrcase.php
```diff
@@ -137,7 +137,6 @@ function hrcase_civicrm_buildForm($formName, &$form) {
       ),
     ));
     }
-    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.hrcase', 'js/hrcase.js');
```
AkA84: why though? We are not removing any js files of unused extensions, why is this one singled out?

hrui/js/src/hrui.js
deb1990: oops, fixed.

hrcase/hrcase.php
deb1990: Only the following extensions has been fixed or ignored
- Code is working fine
- Code is not working, but where the code is used is known to us and will be fixed in future tickets
- Extension is Disabled

But for `civicase`, extension is enabled. And upon further analysis it has been found that the code was written as part of https://issues.civicrm.org/jira/browse/HR-213, which is 4 years ago. Also its not found where the code has been used. Also the `case` menu item is not present anymore. 
So it has been removed.

hrui/js/src/hrui/main.js
```diff
@@ -4,19 +4,19 @@
 (function ($, _) {
   define(function () {
     $(document)
-      .on('crmLoad', function (e) {
-        addUploadFileListener("input[type='file']");
-        amendVacancyForm();
-        amendContactPageAndForm(e);
-        applyMiscChanges();
-        changeContactSourceFieldHelpText();
+      .on('crmLoad', function () {
```
AkA84: i thought that this even happened before this AMD module was loaded?

hrui/js/src/hrui/main.js
```diff
@@ -65,17 +65,15 @@
     /**
      * Amends the contact page and the contact form
      */
-    function amendContactPageAndForm (e) {
-      if (CRM.formName === 'contactForm' || CRM.pageName === 'viewSummary') {
```
AkA84: isn't this check important? Why do we want to (try to) do this DOM manipulation in any page now?

hrui/js/src/hrui/main.js
```diff
@@ -65,17 +65,15 @@
...
-        // Rename "Summary" tab to "Personal Details"
-        // Hack to check contact type - This field only appears for individuals
-        if ($('.crm-contact-job_title', '.crm-summary-contactinfo-block').length) {
-          $('.crm-contact-tabs-list #tab_summary a', e.target).text('Personal Details');
```
AkA84: wasn't `e.target` important to limit the scope of this selector?

hrui/js/src/hrui/main.js
```diff
@@ -200,18 +198,15 @@
     /**
      * Misc changes to the page (hiding elements, inserting new ones, etc)
      */
-    function miscContactPageChanges (target) {
+    function miscContactPageChanges () {
       // Hide current employer and job title
       // Contact summary screen:
       $('div.crm-contact-current_employer, div.crm-contact-job_title', '.crm-summary-contactinfo-block').parent('div.crm-summary-row').hide();
       // Inline edit form
-      $('form#ContactInfo input#employer_id, form#ContactInfo input#job_title', target).closest('div.crm-summary-row').hide();
```
AkA84: wasn't `target` important to limit the scope of this selector?

hrui/js/src/hrui/main.js
```diff
@@ -200,18 +198,15 @@
...
+      $('form#ContactInfo input#employer_id, form#ContactInfo input#job_title').closest('div.crm-summary-row').hide();
       // Contact edit screen
       $('input#employer_id, input#job_title', 'form#Contact').parent('td').hide();
 
-      /* Changes on Add Individual pages and Personal details tab for HR-358 */
-      // Move Job summary to top
-      $('.HRJobContract_Summary', target).insertBefore($('.crm-summary-contactinfo-block'));
```
AkA84: why isn't this needed anymore?

org.civicrm.reqangular/reqangular.php
```diff
@@ -121,21 +133,7 @@ function reqangular_civicrm_pageRun($page) {
...
-      'angularBootstrap' => "$url/js/src/common/vendor/angular/ui-bootstrap",
-      'angularFileUpload' => "$url/js/src/common/vendor/angular/angular-file-upload",
-      'angularResource' => "$url/js/src/common/vendor/angular/angular-resource.min",
-      'angularRoute' => "$url/js/src/common/vendor/angular/angular-route.min",
-      'requireLib' => "$url/js/src/common/vendor/require.min",
-      'reqangular' => "$url/js/dist/reqangular.min",
-    ));
-
-    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.reqangular', 'js/dist/reqangular.min.js', 1000);
+    _attach_reqangular();
```
AkA84: "private" functions should have the name of the extension as namespace `_reqangular_attach_reqangular()`

org.civicrm.reqangular/reqangular.php
```diff
@@ -150,3 +148,36 @@ function reqangular_civicrm_pageRun($page) {
...
 }
+
+/**
+ * Checks if the given form requires reqangular.min.js
+ *
+ * @param  [string]  $formName
+ * @return boolean
+ */
+function _reqangular_checkIfFormRequiresReqangular($formName) {
+  $listOfFormName = ['CRM_Contact_Form_Contact', 'CRM_HRRecruitment_Form_HRVacancy', 'CRM_HRRecruitment_Form_Application'];
```
AkA84:
```php
$whitelist = [
  'CRM_Contact_Form_Contact',
  'CRM_HRRecruitment_Form_HRVacancy',
  'CRM_HRRecruitment_Form_Application'
];
```

org.civicrm.reqangular/reqangular.php
AkA84: maybe `_reqangular_inject_reqangular` is better

org.civicrm.reqangular/reqangular.php
```diff
@@ -107,6 +107,18 @@ function reqangular_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
...
 }
 
+/**
+ * Implements hook_civicrm_buildForm().
+ *
+ * @param string $formName
+ * @param CRM_Core_Form $form
+ */
+function reqangular_civicrm_buildForm($formName, &$form) {
+  if(_reqangular_checkIfFormRequiresReqangular($formName)) {
```
AkA84: maybe `_reqangular_doesFormRequiresReqangular` is better

hrui/js/src/hrui/main.js
deb1990: Yes this should not be removed, reverted.

hrui/js/src/hrui/main.js
deb1990: I have mentioned the same in the PR doc
> Parameters has been removed from the following functions as that is not necessary, and code works without that.

hrui/js/src/hrui/main.js
deb1990: I have mentioned the same in the PR doc
> Parameters has been removed from the following functions as that is not necessary, and code works without that.

org.civicrm.reqangular/reqangular.php
deb1990: Its not required anymore.

org.civicrm.reqangular/reqangular.php
deb1990: Not required anymore

PR #2778, opened 2018-07-20
PCHR-3345: Unable to Create User Account for Restored Contact

ajesamson:
## Overview
This PR fixes issues with creating user account for restored contact.

## Before
Creating user account results in an error, indicating duplicate emails address

restoreerror

## After
after_restored

## Technical Details
Creating user account attempts to create a new user record which is not needed for restored account. User account is therefore checked before processing request.
```php
$user = $this->drupalUserService->findUserByEmail($contact['email']);
 if (!$user) {
   $this->createAccount($contact, $roles);
 }
 else {
   $this->restoreAccount($user, $roles, $contact['id']);
 }
```

When user accounts are deleted, their UFID also gets deleted in the process. When restored, the UFID do not get re-created. Only the account was set to active once again. After checking and updating existing account, new UFID was generated.
```php
public function restoreAccount($user, $roles, $contactId) {
  $this->drupalUserService->setRoles($user->mail, $roles);

  $ufMatch['uf_id'] = $user->uid;
  $ufMatch['contact_id'] = $contactId;
  $ufMatch['uf_name'] = $user->mail;
  civicrm_api3('UFMatch', 'create', $ufMatch);
}
```

Existing roles are ignored while creating the account. They are untouched after initial record removal. User may select an entirely new role, thereby assigning multiple role to account.
```
public function setRoles($email, $roles) {
  $user = user_load_by_mail($email);
  $roles = $this->roleService->getRoleIds($roles);

  user_save($user, ['roles' => $roles]);
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
@@ -44,7 +44,13 @@ class CRM_HRCore_Form_CreateUserRecordTaskForm extends AbstractDrupalInteraction
     $contactsToCreate = $this->getValidContactsForCreation();
 
     foreach ($contactsToCreate as $contact) {
-      $this->createAccount($contact, $roles);
+      $user = $this->drupalUserService->findUserByEmail($contact['email']);
+      if (!$user) {
+        $this->createAccount($contact, $roles);
+      }
+      else {
+        $this->restoreAccount($user, $roles, $contact['id']);
+      }
     }
 
     $haveAccount = $this->getContactsWithAccount();
@@ -218,4 +224,21 @@ class CRM_HRCore_Form_CreateUserRecordTaskForm extends AbstractDrupalInteraction
     return empty($errors) ? TRUE : $errors;
   }
 
+  /**
+   * Restores user account and update roles after delete.
+   * Deleting user account does not remove drupal account.
+   * The user account is just unlinked in the UF match civi table
+   *
+   * @param object $user
+   * @param array $roles
+   * @param int $contactId
+   */
+  public function restoreAccount($user, $roles, $contactId) {
+    $this->drupalUserService->addRoles($user->mail, $roles);
+
+    $ufMatch['uf_id'] = $user->uid;
+    $ufMatch['contact_id'] = $contactId;
+    $ufMatch['uf_name'] = $user->mail;
+    civicrm_api3('UFMatch', 'create', $ufMatch);
+  }
 }
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
@@ -225,7 +225,7 @@ class CRM_HRCore_Form_CreateUserRecordTaskForm extends AbstractDrupalInteraction
   }
 
   /**
-   * Restores user account and update roles after delete.
+   * Restores user account and set role(s) after delete.
    * Deleting user account does not remove drupal account.
    * The user account is just unlinked in the UF match civi table
    *
@@ -234,7 +234,7 @@ class CRM_HRCore_Form_CreateUserRecordTaskForm extends AbstractDrupalInteraction
    * @param int $contactId
    */
   public function restoreAccount($user, $roles, $contactId) {
-    $this->drupalUserService->addRoles($user->mail, $roles);
+    $this->drupalUserService->setRoles($user->mail, $roles);
 
     $ufMatch['uf_id'] = $user->uid;
     $ufMatch['contact_id'] = $contactId;
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
@@ -90,6 +90,19 @@ class CRM_HRCore_Service_DrupalUserService {
   }
 
   /**
+   * Defaults user's role to supplied role(s)
+   *
+   * @param string $email
+   * @param array $roles
+   */
+  public function setRoles($email, $roles) {
+    $user = user_load_by_mail($email);
+    $roles = $this->roleService->getRoleIds($roles);
+
+    user_save($user, ['roles' => $roles]);
+  }
+
+  /**
    * @param string $name
    *
    * @return bool
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
@@ -52,6 +52,17 @@ class CRM_HRCore_Service_DrupalUserService {
   }
 
   /**
+   * Checks if user record exist for email
+   *
+   * @param string $email
+   *
+   * @return object | boolean
+   */
+  public function findUserByEmail($email) {
+    return user_load_by_mail($email);
+  }
+
+  /**
    * @param $email
    */
   public function sendActivationMail($email) {
@@ -64,6 +75,8 @@ class CRM_HRCore_Service_DrupalUserService {
   }
 
   /**
+   * Adds role(s) to user account
+   *
    * @param string $email
    * @param array $roles
    */
--- a/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
+++ b/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/DrupalUserServiceTest.php
@@ -53,6 +53,22 @@ class CRM_HRCore_Service_DrupalUserServiceTest extends CRM_HRCore_Test_BaseHeadl
     $this->assertArrayHasKey(8, $user->roles);
   }
 
+  public function testFindUserByEmailReturnsUserObjectWhenUserEmailExist() {
+    $roleService = $this->prophesize(DrupalRoleService::class);
+    $drupalUserService = new DrupalUserService($roleService->reveal());
+    $user = $drupalUserService->findUserByEmail('johndoe@test.com');
+
+    $this->assertInstanceOf(\stdClass::class, $user);
+  }
+
+  public function testFindUserByEmailReturnsFalseWhenUserEmailDoesNotExist() {
+    $roleService = $this->prophesize(DrupalRoleService::class);
+    $drupalUserService = new DrupalUserService($roleService->reveal());
+    $userAccount = $drupalUserService->findUserByEmail('test@email.com');
+
+    $this->assertFalse($userAccount);
+  }
+
   protected function cleanup() {
     $user = user_load_by_mail($this->testEmail);
     if ($user) {
--- a/uk.co.compucorp.civicrm.hrcore/tests/phpunit/drupal_function_mocks.php
+++ b/uk.co.compucorp.civicrm.hrcore/tests/phpunit/drupal_function_mocks.php
@@ -10,7 +10,14 @@ function drupal_anonymous_user() {}
 
 function user_password() {}
 
-function user_load_by_mail($email) {}
+function user_load_by_mail($email) {
+  if ($email === 'johndoe@test.com') {
+    return new stdClass();
+  }
+  else {
+    return FALSE;
+  }
+}
 
 function user_roles() {
   return [1 => 'Fake Role'];


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -44,7 +44,14 @@ public function postProcess() {
...
 
     foreach ($contactsToCreate as $contact) {
-      $this->createAccount($contact, $roles);
+      $user = $this->drupalUserService->findUserByEmail($contact['email']);
+      if (!$user) {
+        $this->createAccount($contact, $roles);
+      }
+      else {
+        $this->drupalUserService->processUserAccount($user, $roles);
+        $this->addUfId($user, $contact['id']);
```
tunbola: I don't think we need to add the `processUserAccount` method to the DrupalUserService class. For restoring accounts, what is basically happening is just restoring the user roles. I would suggest the following:
- Revert changes to the `createNew` method. 
- Change the `findUserByEmail` to something like `userExistsForEmail` that returns a boolean
- Use the Drupal user service `addRoles` method to add the roles the contact email if a user account for that email already exists, if not create a new user account(default behaviour)
- Extract the function to addUserRole and addUfId to a new function `restoreAccount` within the form class for clarity.

This way we would only need just one public method to the DrupalUserService class and not two. 
Then please add a test for the `userExistsForEmail`(or a better name) method, see sample tests that uses drupal mock function in for the Drupal service test classes.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -46,11 +59,21 @@ public function createNew($email, $active = FALSE, $roles = []) {
...
-    $user = user_save(drupal_anonymous_user(), $params);
+    $user = user_save($user, $params);
 
     return $user;
   }
 
+  /**
+   * Finds a user record by email
+   *
+   * @param string $email
```
tunbola: Please add a line before return

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
ajesamson: Thanks. I will refactor the implementation.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -87,16 +77,36 @@ public function sendActivationMail($email) {
...
+   * @return object
    */
   public function addRoles($email, $roles) {
     $user = user_load_by_mail($email);
     $roles = $this->roleService->getRoleIds($roles);
     $roles = array_unique(array_merge($roles, array_keys($user->roles)));
     $roles = array_combine($roles, $roles);
 
-    user_save($user, ['roles' => $roles]);
+    return user_save($user, ['roles' => $roles]);
```
tunbola: Please revert this change, we don't need it actually.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -44,13 +44,12 @@ public function postProcess() {
...
-      $user = $this->drupalUserService->findUserByEmail($contact['email']);
-      if (!$user) {
+      $userExists = $this->drupalUserService->userExistsForEmail($contact['email']);
+      if (!$userExists) {
         $this->createAccount($contact, $roles);
       }
       else {
-        $this->drupalUserService->processUserAccount($user, $roles);
-        $this->addUfId($user, $contact['id']);
+        $this->drupalUserService->restoreAccount($contact['email'], $roles, $contact['id']);
```
tunbola: We actually don't need a restoreAccount method in the DrupalUserService. The Service should not be concerned about setting Ufid for civicrm, its not its responsibility. What I meant earlier was to retain the addUFId method in the `CreateUserRecordTaskForm` class and create a new method called restoreAccount in this class.

```php
private function restoreAccount($contact, roles) {
$this->drupalUserService->addRoles($contact['email], $roles);
$this->addUfId($contact);
}
```
So the only new method we'll be adding to the DrupalUserService will be the `userExistsForEmail` method.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -87,16 +77,36 @@ public function sendActivationMail($email) {
...
+  }
+
+  /**
+   * Restores user account after delete
+   *
+   * @param $email
+   * @param $roles
+   * @param $contactId
+   */
+  public function restoreAccount($email, $roles, $contactId) {
```
tunbola: No need for this new method.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -59,19 +46,22 @@ public function processUserAccount($user, $roles, $params = []) {
...
+   * Checks if user record exist by email
    *
    * @param string $email
-   * @return array|boolean
+   *
+   * @return boolean
    */
-  public function findUserByEmail($email) {
-    return user_load_by_mail($email);
+  public function userExistsForEmail($email) {
```
tunbola: Please see sample tests as a guide to write test for this method in this classes: https://github.com/compucorp/civihr/blob/9313c6246af92409a664840f6574031edd9a169d/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/UserAccount/DrupalTest.php#L2-L1 .

Drupal is not bootstrapped during Civi tests, so what we do is to have a file that mocks drupal functions and the return results. For the test for this function, you can modify this function here: https://github.com/compucorp/civihr/blob/16324ac5eb8033d87beac6bf1dabb58751528501/uk.co.compucorp.civicrm.hrcore/tests/phpunit/drupal_function_mocks.php#L13-L12, to return TRUE for a particular email and false for any other email, that, way you can test that the function returns true or false.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -44,7 +44,13 @@ public function postProcess() {
     $contactsToCreate = $this->getValidContactsForCreation();
 
     foreach ($contactsToCreate as $contact) {
-      $this->createAccount($contact, $roles);
+      $user = $this->drupalUserService->findUserByEmail($contact['email']);
+      if (!$user) {
+        $this->createAccount($contact, $roles);
+      }
+      else {
+        $this->restoreAccount($user, $roles, $contact['id']);
```
ajesamson: `$user` is needed here because `uid` is required for adding `UFID`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -218,4 +224,19 @@ public function validateInput($fields) {
...
+   * @param array $roles
+   * @param int $contactId
+   */
+  public function restoreAccount($user, $roles, $contactId) {
+    $this->drupalUserService->addRoles($user->mail, $roles);
+
+    $ufMatch['uf_id'] = $user->uid;
+    $ufMatch['contact_id'] = $contactId;
+    $ufMatch['uf_name'] = $user->mail;
+    CRM_Core_BAO_UFMatch::create($ufMatch);
```
tunbola: Would be better to use the Civi API here as it would be more reliable rather than a BAO method

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
```diff
@@ -218,4 +224,19 @@ public function validateInput($fields) {
     return empty($errors) ? TRUE : $errors;
   }
 
+  /**
+   * Restores user account after delete
```
tunbola: Perhaps add a bit more flesh to this. A little explanation about what happens when a contact is soft deleted and why we need this method would be helpful.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -51,6 +51,17 @@ public function createNew($email, $active = FALSE, $roles = []) {
     return $user;
   }
 
+  /**
+   * Checks if user record exist for email
+   *
+   * @param string $email
+   *
+   * @return object | boolean
```
tunbola: object or boolean? Should be object I think.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserAccountInterface.php
```diff
@@ -45,4 +45,13 @@ public function enable($contactData);
...
   public function isUserDisabled($contactData);
+
+  /**
+   * Finds user by email
+   *
+   * @param $mail
+   *
+   * @return mixed
+   */
+  public function findUserByEmail($mail);
```
tunbola: Do we need this here? The `CRM_HRCore_CMSData_Role_DrupalRoleService` class does not extend this class.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/CMSData/UserAccount/Drupal.php
```diff
@@ -71,4 +71,11 @@ public function isUserDisabled($contactData) {
 
     return !$user->status ? true : false;
   }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function findUserByEmail($mail) {
```
tunbola: Do we need this method here? since you added it to the `CRM_HRCore_Service_DrupalUserService` class also and its the `CRM_HRCore_Service_DrupalUserService` class that is used as the drupal service class in the `CreateUserRecordTaskForm` class.
I know the classes are similar and a bit confusing, we need to harmonize both eventually.

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/CMSData/UserAccount/DrupalTest.php
```diff
@@ -50,4 +50,17 @@ public function testIsUserDisabledReturnsTrueWhenUserStatusIsNotActive() {
 
     $this->assertTrue($userAccount->isUserDisabled($contactData));
   }
+
+  public function testFindUserByEmailReturnsTrueWhenUserEmailExist() {
+    $userAccount = new DrupalUserAccount();
+    $user = $userAccount->findUserByEmail('true@email.com');
+
+    $this->assertEquals($user, new stdClass());
```
tunbola: Better to just do
```php
$this->assertInstanceOf(\stdClass::class, $user);
```

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
ajesamson: 👍 I will change to API implementation.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
ajesamson: If requested user is not found by email, response is a boolean.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Form/CreateUserRecordTaskForm.php
ajesamson: 👌 Comments will be updated.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -89,6 +89,19 @@ public function addRoles($email, $roles) {
     user_save($user, ['roles' => $roles]);
   }
 
+  /**
+   * Defaults user's role to supplied role(s)
+   *
+   * @param string $email
+   * @param array $roles
+   */
+  public function setRoles($email, $roles) {
```
tunbola: Please add a test for this method.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
```diff
@@ -64,6 +75,8 @@ public function sendActivationMail($email) {
   }
 
   /**
+   * Adds role(s) to user account
```
tunbola: Would be a good idea to update this method to reflect that it adds roles in addition to existing user roles.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Service/DrupalUserService.php
tunbola: After some discussion with @ajesamson , its a bit difficult to mock tests for this so approving this without a test.

PR #2784, opened 2018-07-23
PCHR-3999: Hide all Civicrm System Notifications

ajesamson:
## Overview
This PR hides System Notifications for non root users.

## Before
before_notification

## After
after_notification

## Technical Details
System notification is disabled for all non root users by implementing `hook_civicrm_check` and clearing the messages.

```
$messages = [];
```
This also implies no system notification will be rendered in the `/civicrm/a/#/status` page.

An existing `view system status in footer` has also been [renamed](https://github.com/compucorp/civihr-employee-portal/pull/524) to `view system notifications` since they both serve similar purpose of restricting access to system notification by non root users.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/hrcore.php
+++ b/uk.co.compucorp.civicrm.hrcore/hrcore.php
@@ -344,6 +344,15 @@ function hrcore_civicrm_permission(&$permissions) {
 }
 
 /**
+ * Removes system notifications for all users
+ *
+ * @param $messages
+ */
+function hrcore_civicrm_check(&$messages) {
+  $messages = [];
+}
+
+/**
  * Renames a menu with the given new label
  *
  * @param array $params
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -343,6 +343,15 @@ function hrcore_civicrm_permission(&$permissions) {
   $permissions['view system status on footer'] = $prefix . ts('View System Status on Footer');
 }
 
+/**
+ * Removes system notifications for all users
+ *
+ * @param $messages
```
mickadoo: Missing type

ajesamson: Ooops. Thanks for catching that.

uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -340,7 +340,19 @@ function hrcore_civicrm_permission(&$permissions) {
...
+  $permissions['view system notifications'] = $prefix . ts('View System Notifications');
+}
+
+/**
+ * Removes system notifications for users without permission
+ *
+ * @param array $messages
+ */
+function hrcore_civicrm_check(&$messages) {
+  $viewSystemNotifications = CRM_Core_Permission::check('view system notifications');
```
mickadoo: I think this would make more sense as `canViewSystemNotifications`

PR #2790, opened 2018-07-25
PCHR-3944: Update to AngularJS v1.7

AkA84:
## Overview
This PR upgrades AngularJS from v1.5.9 to v1.7.2. This is an important upgrade because 1.7 is the last major release of AngularJS, and it's now in LTS.

Both bumps from 1.6 and 1.7 meant dealing with stricter behaviour and breaking changes, so it required various fixes in order for our apps (and their tests) to work as before

### Increased strictness in unit tests
Both 1.6 and 1.7 are stricter in 1.5 when running unit tests, meaning that they reported issues that were simply overlooked in the older version:

#### Unexpected requests
Unexpected http requests that were overlooked now failed the tests. Expectations had to be added for the tests to pass
```js
// example
$httpBackend.whenGET(/action=get&entity=HRJobContract/).respond(200);
```

#### `xyz.templates` module not being used
We cache view templates in a `xyz.templates` module when running tests, but some tests forgot to include the module and as such angular tried to fetch the templates via http requests instead of finding them in the cache.

This wasn't flagged in 1.5, but it was in 1.6 and 1.7. Including the module fixed the issue
```diff
// example
- beforeEach(module('contactsummary', 'contactsummary.mocks'));
+ beforeEach(module('contactsummary', 'contactsummary.mocks', 'contact-summary.templates'));
```

### Managing unhandled rejection in tests
Angular >= 1.6 now throws an error if there are unhandled rejections in the code (= no `.catch()` block). This turned out to be a problem in some spec files, where we intentionally create a rejected promise in order to trigger a certain behaviour. For example:

```js
describe('updateEntitlements()', function () {
  describe('when absence periods are available', function () {
    describe('and user does not confirm to update entitlements', function () {
      beforeEach(function () {
        mockModalInstanceAndReturn($q.reject()); // <---
        utilsService.updateEntitlements();
        $rootScope.$digest();
      });

      it('does not redirect to entitlement updation page for the current user', function () {
      });
    });

    function mockModalInstanceAndReturn (modalReturnValue) {
      var mockModalInstance = { result: modalReturnValue };
      spyOn(mockModalInstance.result, 'then').and.callThrough();
      spyOn($uibModal, 'open').and.returnValue(mockModalInstance);
    }
  });
});
```
Here we explicitly want the `mockModalInstance.result` to be a failed promise, so that the `.then()` in this block code will never be called.

```js
confirmUpdateEntitlements()
  .then(function () {
    $window.location.assign(this.getManageEntitlementsPageURL(contactID));
  }.bind(this));
```

If we handled the rejection as angular would want us to do, then the original code flow would be disrupted, and the test would be meaningful. We do need to have that rejection to be unhandled.

In order to fix the problem, the `$qProvider` provides a `errorOnUnhandledRejections()` method, which turn off the error
```js
beforeEach(module('job-contract', 'leave-absences.models', function ($qProvider) {
  $qProvider.errorOnUnhandledRejections(false);
}));
```

### Use of `$onInit` hook in components
Before the upgrade, we used a custom `init()` IIFE to run any initialization logic for components
```js
(function init () {
  return $q.all([
    loadWorkPatterns(),
    loadJobContractRevisionChangeReasons(true)
  ])
    .finally(function () {
      vm.loading.content = false;
    });
})();
```
this function had to be dumped in favour of the official `$onInit` lifecycle hook, see more details [here](https://toddmotto.com/angular-1-6-is-here#component-and-oninit)
```js
var vm = this;
vm.$onInit = $onInit;

function $onInit () {
  return $q.all([
    loadWorkPatterns(),
    loadJobContractRevisionChangeReasons(true)
  ])
    .finally(function () {
      vm.loading.content = false;
    });
}
```
this change required changing the test files as well, given that the hook is not called automatically when calling `$componentController()` (see [here](https://github.com/angular/angular.js/issues/14129))
```diff
function compileComponent () {
- $componentController('absenceTabContainer');
+ var ctrl = $componentController('absenceTabContainer');
+
+ ctrl.$onInit();
  $rootScope.$digest();
}
```
### Switch to textAngular fork
From https://github.com/compucorp/civihr/pull/2790/commits/149dfc1c7c3b04b0c733e59d52cd26b36123dbf4
> textAngular does not work with Angular 1.7, see textAngular/textAngular#1608 and seems the project had been abandoned (at the time of this writing, the last changes are from 2 years ago). The only way to get it to work is to use a fork, https://github.com/antosarho/textAngularJs/, which contains the necessary fixes

The [rangy](https://github.com/timdown/rangy/blob/master/lib/rangy-core.js) module had to be updated as well given that it's a dependency of textAngular

### Update of `angular-ui-router` to latest legacy version
[`angular-ui-router`](https://github.com/angular-ui/ui-router) had been updated to v0.4.3. This is the last version before the bump to 1.0.0, which introduced many breaking changes that would have caused sever issues to the scope of this ticket, and as such was ignored for the time being

Also note that L&A was using its own local copy of the router. The local copy had been dumped in favour of the one in reqangular


hrjobcontract/js/test/services/contract-pay.service.spec.js
```diff
@@ -15,8 +15,8 @@ define([
       $httpBackend = _$httpBackend_;
       $rootScope = _$rootScope_;
 
-      $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond(MockContract.contractPayment);
-      $httpBackend.whenGET(/action=get&entity=HRJobPay/).respond({});
+      $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond(200);
+      $httpBackend.whenGET(/action=get&entity=HRJobPay/).respond(MockContract.contractPayment);
```
reneolivo: why was this changed and how come it was not breaking before?

hrjobcontract/js/test/services/utils.service.spec.js
```diff
@@ -11,18 +11,21 @@ define([
...
 
-    beforeEach(module('job-contract', 'leave-absences.models', function (_$provide_) {
+    beforeEach(module('job-contract', 'leave-absences.models', function (_$provide_, $qProvider) {
       $provide = _$provide_;
       $provide.value('$window', {
         location: {
           assign: jasmine.createSpy('spy')
         }
       });
+      $qProvider.errorOnUnhandledRejections(false);
```
reneolivo: why is `$qProvider.errorOnUnhandledRejections(false);` being used? I think handling all promise rejections is a good practice and may help uncover issues with the code.

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/instances/leave-request.instance.spec.js
```diff
@@ -184,6 +184,10 @@ define([
 
       function commonSetup (methodName) {
         instance = LeaveRequestInstance.init(helper.createRandomLeaveRequest());
+        instance.id = _.random(1, 10);
```
reneolivo:
```js
instance.id = _.uniqueId();
```

hrjobcontract/js/test/services/contract-pay.service.spec.js
AkA84: the only explanation i can give is that 1.6 and 1.7 are stricter with tests, while 1.5 let a lot of things slide, even if wrong.

You can definitely tell that the "before" was wrong (returning a mocked contract payment on `entity=HRJobContract` instead of `entity=HRJobPay`), but still didn't cause a blip on 1.5. As soon as the version was bumped, you correctly got the error.

I didn't go much deeper than that tbh, at that point I'd have to go reverse engineer what angular does, and didn't seem worth it

hrjobcontract/js/test/services/utils.service.spec.js
AkA84: sorry, i forgot to document it

PR #2794, opened 2018-07-26
PCHR-4019: Allow self leave approver to select leave request status on My Leave in the Leave Request Modal and delete a leave request

igorpavlov:
## Overview

This PR allows self leave approver to select leave request status on **My Leave** in the Leave Request Modal and delete a leave request. In other words, they become "admins" on **My Leave** for their own leave requests if they are self leave approvers.

### Loading logged in contact instance instead of just their ID

Now we use the `Contact.getLoggedIn` that resolves with `ContactInstance` instead of calling `Session.get` to save some lines of code. We need `ContactInstance` to call `ContactInstance.checkIfSelfLeaveApprover`

```js
// var loggedInContactId = '';
var loggedInContact;
// ...

loggedInContact.leaveManagees() // no need to do Contact.find(loggedInContactId) anymore
// ...

loggedInContact.checkIfSelfLeaveApprover() // same as above
```

### Init role now checks if self leave approver

```js
function initRole () {
  role = 'staff';

  return (vm.isSelfRecord
    ? initRoleBasedOnSelfLeaveApproverState()
    : initRoleBasedOnPermissions())

// ...

function initRoleBasedOnSelfLeaveApproverState () {
  return loggedInContact.checkIfSelfLeaveApprover()
    .then(function (isSelfLeaveApprover) {
      if (!isSelfLeaveApprover) {
        return;

      role = 'admin';
      vm.isSelfLeaveApprover = true;

// ....

function initCanManage () { // used in the init chain
  vm.canManage = vm.isRole('manager') || vm.isRole('admin');
}
```

#### Why `role = 'admin'`?

From the business logic perspective a self leave approver "administers" their own leave requests.

#### Why `vm.isSelfLeaveApprover = true`?

This is a new public property because we want to show some specific UI elements to the self leave approver: Status Selector and Delete button. We don't want to show other admin UI elements, such as, for example, the name of the contact on the top of the modal.

We also cannot only use `vm.canManage` property alone, because the templates also consider `vm.isSelfRecord` which prevents some UI elements to be shown when creating/editing you own leave request.

The solution is to use a combination in the templates:

```js
&& !$ctrl.isSelfRecord
// becomes
&& (!$ctrl.isSelfRecord || $ctrl.isSelfLeaveApprover)
```

-----

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -607,28 +607,51 @@ define([
...
-      // If the user is creating or editing their own leave, they will be
-      // treated as a staff regardless of their actual role.
-      if (vm.isSelfRecord) {
-        return;
-      }
+      return (vm.isSelfRecord
+        ? initRoleBasedOnSelfLeaveApproverState()
+        : initRoleBasedOnPermissions())
+        .finally(function () {
+          vm.canManage = vm.isRole('manager') || vm.isRole('admin');
```
reneolivo: can we remove this from here, move it to a `initCanManage` function, and call that function from the `init` sequence after `initRole`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -767,10 +790,11 @@ define([
      *
      * @return {Promise}
      */
-    function loadLoggedInContactId () {
-      return Session.get().then(function (value) {
```
reneolivo: since we now don't use Session, can you please remove any reference to it from this file?

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/controllers/request.controller.spec.js
```diff
@@ -1342,33 +1346,57 @@
...
+
+              initTestController({ leaveRequest: leaveRequest });
+            });
+
+            it('sets the "admin" role', function () {
+              expect(controller.isRole('admin')).toBe(true);
+            });
+
+            it('sets the `isSelfLeaveApprover` public property to `true`', function () {
+              expect(controller.isSelfLeaveApprover).toBe(true);
```
reneolivo: you need to check if `isSelfLeaveApprover` is set to false in the `basic tests` group`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/controllers/request.controller.spec.js
```diff
@@ -1342,33 +1346,57 @@
...
+            });
+
+            it('sets the `canManage` public property to `true`', function () {
```
reneolivo: Need to check if it's set to false as well.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
igorpavlov: Done. I also checked if we need `finally`, I don't think so, so I changed to `then`

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
igorpavlov: Hm... I wonder why linter did not notice this.

igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/controllers/request.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/controllers/request.controller.spec.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
reneolivo: The linter doesn't care if you don't use one of the parameters in your function. I guess it does that in case you need to use the third parameter, but not the second one.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/controllers/request.controller.js
```diff
@@ -507,6 +508,13 @@ define([
       ].concat(defaultStatuses);
     }
 
+    /**
+     * Defines if the contact ca manage the leave request
```
reneolivo: *can

PR #2800, opened 2018-07-30
PCHR-4045: Always treat admin as self leave approver

igorpavlov:
## Overview

This PR makes all admins to be treated as self leave approvers. In other words, admins now can approve their own leave requests or create new approved ones on My Leave page.

### Allowing quick actions

Previously, on My Leave admins without a relationship "has leave approved by" were treated as staff, this is because we passed `role = "staff"` in the quick actions component from the Staff Leave Report component:

```js
// staff-leave-report component
vm.role = ($rootScope.section === 'absence-tab' ? 'admin' : 'staff');
```

Now we init the role based on the permissions and then pass it to the Quick Actions component:

```js
function initRole () {
  vm.role = 'staff';
  // if absence tab, no need to check, it will always be "admin"
  if ($rootScope.section === 'absence-tab') {
     vm.role = 'admin';
     return;

  // checking if staff or admin
  // (no need to check if manager, the report is always for staff or admins)
  return checkPermissions(sharedSettings.permissions.admin.administer)
    .then(function (isAdmin) {
      isAdmin && (vm.role = 'admin');
```

### Allowing status change in the Leave Request Modal

Because the logic becomes a bit complicated, it was decided to follow this concept:
1. Initiate the role based on:
  1.1. Permissions
  1.2. "Has leave approved by" relationship (set to admin if has this relationship)
2. Initiate the "isSelfApprover" property for UI based on the role (is TRUE if admin)

```js
// First, init the role
function initRole () {
  role = 'staff';

  return initRoleBasedOnPermissions()
    .then(vm.isSelfRecord && !vm.isRole('admin') && amendRoleBasedOnSelfLeaveApproverState);

// Then set isSelfApprover to true if admin
function initIsSelfLeaveApprover () {
  if (vm.isRole('admin') && vm.isSelfRecord) {
  vm.isSelfLeaveApprover = true;
```
----

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/components/staff-leave-report.component.js
```diff
@@ -175,6 +178,26 @@ define([
       section.dataIndex = _.indexBy(section.data, 'id');
     }
 
+    /**
+     * Initiates the role based on permissions.
+     * @NOTE It skips the permission check if the section is 'absence-tab'
+     * because only admins can access this section.
+     */
```
deb1990: 1. Return is not documented.
2. When `$rootScope.section === 'absence-tab'`, the function is not returning a promise. But the function return type should always be of same type.

PR #2806, opened 2018-07-31
PCHR-4053: Fix issue for manager self approver

igorpavlov:
## Overview

This PR fixes an issue on Manager Leave for self requests. This is not a regression, it is an old code that has not been considered in other PRs related to the epic.

## Before

Manager is not treated as a self approver for their own requests when have a leave approving relationship set to themselves.

image

## After

Manager is treated as a self approver for their own requests when have a leave approving relationship set to themselves.
_(Please note the default "Approved" status as a visual difference)_

image

## Technical Details

The logic was not updated for the epic.

```js
// was. it caused vm.isSelfRecord to be false even for self requests on Leave Manager
vm.isSelfRecord = isSectionMyLeave && (isMyOwnRequest || isNewRequest);
// now
vm.isSelfRecord = (isSectionMyLeave && isNewRequest) || isMyOwnRequest;
```

-----

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/controllers/request.controller.spec.js
```diff
@@ -645,12 +646,19 @@
           });
         });
 
-        describe('basic tests for when manager opens leave request popup in edit mode', function () {
+        describe('when manager opens someone else\'s leave request', function () {
```
deb1990: Using escape character here makes it less readable, instead please use double quotes outside
`"when manager opens someone else's leave request"`

PR #2807, opened 2018-08-01
PCHR-4076: Fix Filters on Admin Leave - Leave Requests and Calendar Tab

ajesamson:
## Overview
When filtering leave by admin on leave requests and calendar tab, records for staff whose job role are out of active date are returned. This PR fixes this by ensuring that date of filtering falls within staff job roles start and end date.

## Before
Filter result includes staff records whose job role start date is more than filter date or end date is less than the filtering date.

## After
Filter result no longer include staff records whose job role start date is more than filter date or end date is less than the filtering date.

## Technical Details
Date filtering was [applied](https://github.com/compucorp/civihr/blob/staging/com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php#L126) to the filter query to ensure job role start and end date are put into consideration.
```
$today = date('Y-m-d H:i:s');
$dateRestriction = 'a.start_date <= "' . $today . '"';
$dateRestriction .= ' AND (a.end_date >= "'. $today . '" OR a.end_date IS NULL)';
```

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
@@ -126,9 +126,14 @@ class CRM_Hrjobroles_API_Query_ContactHrJobRolesSelect {
   private function addWhere(CRM_Utils_SQL_Select $customQuery) {
     if (!empty($this->params['contact_id'])) {
       $conditions[] = 'jc.contact_id IN (' . implode(',' , $this->params['contact_id']) . ')';
-      $customQuery->where($conditions);
       unset($this->params['contact_id']);
     }
+
+    $today = date('Y-m-d H:i:s');
+    $dateRestriction = 'a.start_date <= "' . $today . '"';
+    $dateRestriction .= ' AND (a.end_date >= "' . $today . '" OR a.end_date IS NULL)';
+    $conditions[] = $dateRestriction;
+    $customQuery->where($conditions);
   }
 
   /**
--- a/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
+++ b/com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
@@ -99,9 +99,9 @@ class api_v3_ContactHRJobRoleTest extends CRM_Hrjobroles_Test_BaseHeadlessTest {
     $contact2 = ContactFabricator::fabricate();
     $contract1 = HRJobContractFabricator::fabricate(['contact_id' => $contact1['id']]);
     $contract2 = HRJobContractFabricator::fabricate(['contact_id' => $contact2['id']]);
-    $jobRole1 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id'],]);
-    $jobRole2 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id'],]);
-    $jobRole3 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract2['id'],]);
+    $jobRole1 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id']]);
+    $jobRole2 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id']]);
+    $jobRole3 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract2['id']]);
 
     $contactJobRoles = civicrm_api3($this->entity, $this->action)['values'];
 
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Test/Fabricator/HrJobRoles.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Test/Fabricator/HrJobRoles.php
@@ -16,6 +16,10 @@ class CRM_Hrjobroles_Test_Fabricator_HrJobRoles {
    * @throws \Exception
    */
   public static function fabricate($params) {
+    if (empty($params['start_date'])) {
+      $params['start_date'] = date('Y-m-d H:i:s');
+    }
+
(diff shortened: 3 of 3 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
```diff
@@ -129,6 +129,11 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
       $customQuery->where($conditions);
       unset($this->params['contact_id']);
     }
+
+    $dateRestriction = 'a.start_date <= "' . date('Y-m-d H:i:s') . '"';
```
tunbola: Better to extract this `date('Y-m-d H:i:s')` into a variable since its used twice already.
e.g `$today = date('Y-m-d H:i:s')`.

com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
```diff
@@ -129,6 +129,11 @@ private function addWhere(CRM_Utils_SQL_Select $customQuery) {
...
+    $dateRestriction .= ' AND (a.end_date >= "'. date('Y-m-d H:i:s') . '" OR a.end_date IS NULL)';
+    $conditions[] = $dateRestriction;
+    $customQuery->where($conditions);
```
tunbola: There's another `$customQuery->where($conditions)` above, we can remove that since this will be called always.

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
```diff
@@ -99,9 +99,9 @@ public function testTheGetActionReturnsMultipleJobRoles() {
     $contact2 = ContactFabricator::fabricate();
     $contract1 = HRJobContractFabricator::fabricate(['contact_id' => $contact1['id']]);
     $contract2 = HRJobContractFabricator::fabricate(['contact_id' => $contact2['id']]);
-    $jobRole1 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id'],]);
-    $jobRole2 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id'],]);
-    $jobRole3 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract2['id'],]);
+    $jobRole1 = HRJobRolesFabricator::fabricate(['job_contract_id' => $contract1['id'], 'start_date' => date('Y-m-d H:i:s'),]);
```
tunbola: It will be better to make start_date with a default value of current date here for the fabricator here: https://github.com/compucorp/civihr/blob/43f80260a8aa664afe5b943fee4878003fc2e208/com.civicrm.hrjobroles/CRM/Hrjobroles/Test/Fabricator/HrJobRoles.php#L6, that way, there will be no need to modify any tests that use the fabricator and it won't break tests. If a test needs a different date, it can be set explicitly.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestCalendarFeedDataTest.php
```diff
@@ -82,26 +82,30 @@ public function testGetWillReturnDataForFeedConfigLocationAndDepartmentContacts(
...
+      'start_date' => date('Y-m-d H:i:s')
     ]);
 
     //contact4 assigned to both location2 and department2
     HRJobRolesFabricator::fabricate([
       'job_contract_id' => $contract4['id'],
       'location' => $location2['value'],
-      'department' => $department2['value']
+      'department' => $department2['value'],
+      'start_date' => date('Y-m-d H:i:s')
```
tunbola: No need for these changes once the start date is set as default in the fabricator.

tunbola:
I feel that the technical details need a bit more flesh, such as some explanation about what API is called for the Admin leave filters and then link it to the query class that needs its where clause modified.

Also in the before and after section, the generalisation that the `job role start and end date is more or less than the filtering date.` might not be true because staff records with end dates more than the filtering date will also be included if their start date is less than filtering date, So I think you need to re-word this.

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
ajesamson: Expression will not be allowed as field default value. Can you suggest an alternative implementation 🤷‍♂️  ?

com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
ajesamson: 👍 Good. Will refactor.

com.civicrm.hrjobroles/CRM/Hrjobroles/API/Query/ContactHrJobRolesSelect.php
ajesamson: This is correct as this will process the `$conditions` together 👌 .

com.civicrm.hrjobroles/tests/phpunit/api/v3/ContactHRJobRoleTest.php
tunbola: What about modifying the fabricate method like this?
```php
  public static function fabricate($params) {
    //Setting the default start date here when not provided
    if (empty($params['start_date'])) {
      $params['start_date'] = date('Y-m-d H:i:s');
    }
    $result = civicrm_api3(
      'HrJobRoles',
      'create',
      array_merge(self::$defaultParams, $params)
    );

    return array_shift($result['values']);
  }
```

PR #2809, opened 2018-08-01
SHOR-22: Remove extra PHP code for appending contact image - covered under shoreditch

swastikpareek:
## Overview
HRUI component enforces a feature which actually fetch the image URL and appends it near the contact title. If we see Plain CiviCRM the Image is already appearing in the DOM we just need to move it to the position.

Since this change is required in CiviCRM too we have moved this feature to shoreditch - refer - https://github.com/civicrm/org.civicrm.shoreditch/pull/264

## Before
civihr - before

## After
civihr - after


swastikpareek:
Closing this PR as it is raised as with a different base fork. Refer - https://github.com/compucorp/civihr/pull/2810

PR #2812, opened 2018-08-02
PCHR-4005: Fix icons in HRMed and HRCareer forms

igorpavlov:
## Overview

The help icons disappeared after CiviCRM update due to changes of the selectors. This PR makes these icons appear again in the option group forms in HRMed and HRCareer extensions.

## Before

Help icons are missing:

image

image

## After

Help icons are back:

image

image

### Structure

The icons are injected in two different extensions: HRMed and HRCareer. In order to avoid code duplication we need to have a function somewhere that appends the help icons, that can be injected and used in different extensions. It was decided to have this function in HRCore:

```
js
|-- src
   |-- hrcore
      |-- helpers.js (contains the func for help icon appending)
      |-- analytics.js (moved GA code from app.js)
      |-- app.js
```

### Helpers module

```js
define(function () {
   return { appendHelpIcon, getCiviCRMFormLabel };

  function appendHelpIcon ($element, title, helpId, helpFile) {
    // create an element
    var $helpIcon = $('<a href class="helpicon" title="' + title + '"></a>');
    // append to a target element
    $element.append($helpIcon);
    // listen to clicks
    $helpIcon.on('click', function () {
      // standard CRM help notification
      CRM.help('', { id: helpId, file: helpFile });
      return false;

  function getCiviCRMFormLabel (formClass, accessNameSelector) {
    // get the form
    var $form = $('.' + formClass);
    // get the CiviCRM ID of the field
    var accessName = $form.find('[data-crm-custom="' + accessNameSelector + '"]').attr('name');
    // get the label
    var labelSelector = 'label[for=' + accessName + ']';

    return $form.find(labelSelector);
```

### Usage of a new module

```js
// this will fire every time *ANY* CRM form is loaded
$('body').on('crmFormLoad', function () {
  // get a label to append the icon to
  var $label = helpers.getCiviCRMFormLabel(
    'custom-group-Medical_Disability', 'Medical_Disability:Condition');

  // check if label is found and icon was *not yet injected*
  if (!$label.length || $label.find('.helpicon').length) {
    return;

  // append the icon
  helpers.appendHelpIcon(
    $label,
    'Condition Help',
    'hrmed-med-condition',
    'CRM/HRMed/Page/helptext');
```

-----
✅Manual Tests - passed
⏹Jasmine Tests - N/A
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


hrcareer/js/src/hrcareer/app.js
```diff
@@ -1,18 +1,28 @@
...
 
-define(function () {
+define(function (CRM) {
   'use strict';
 
-  CRM.$(function ($) {
-    $('body').on('crmFormLoad', function (event) {
-      if (event.profileName === 'hrcareer_tab') {
-        var accessName = $('[data-crm-custom="Career:End_Date"]').attr('name');
+  CRM.$('body').on('crmFormLoad', function () {
```
deb1990: Please create functions to make the code more readable.
Something like
```javascript
CRM.$('body').on('crmFormLoad', function () {
  if(isCareerTab()) {
    addHelpIcon();
  }
})
```

Similarly do this for the other tab also.

deb1990:
@igorpavlov 
About sharing the `injectHelpIcon()` function. 

## Option 1
We edit the dom to make JS unnecessary. But you have already confirmed that this is not possible. So we need to go for other option.

## Option 2
Move the code to `HRCore`. And require it from `hrcareer` or `hrmed`.
But if we do this
1. the function should be bundled in `hrcore.min.js`
2. `hrcore.min.js` is injected in the page before all the other scripts (like we do with `reqangular.min.js`)
3. the extension that needs to use that function defines
```
paths: {
  'hrcore': 'empty:'
}
```
 in its build.js file, so that it doesn’t include the function in its own bundle

## Option 3
Move `injectHelpIcon()` function to the `app.js`of hrcore(https://github.com/compucorp/civihr/blob/PCHR-4005-fix-js-for-hrcareer-and-hrmed/uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js), and we can call this using events.

**In HRCore**
```
document.addEventListener('injectHelpIcon', injectHelpIcon)

function injectHelpIcon (...) {...}
```

**In HRCareer**
```
CRM.$('body').on('crmFormLoad', function () {
  if(isCareerTab()) {
     var event = new CustomEvent('injectHelpIcon', { 
        form: form,
        accessNameSelector:  'Career:End_Date'
        ....
     });
     document.dispatchEvent(event);
  }
})
```
cc @AkA84

hrcareer/js/src/hrcareer/app.js
```diff
@@ -1,18 +1,16 @@
 /* eslint-env amd */
 
-define(function () {
-  'use strict';
+(function (CRM) {
```
AkA84: not a major point, but you might consider
```js
(function ($) {

}(CRM.$));
```

hrcareer/js/src/hrcareer/app.js
```diff
@@ -1,18 +1,16 @@
...
+  define(['hrcore/helpers/help-icon-injector'], function (injectHelpIcon) {
```
AkA84:
```js
define([
  'hrcore/helpers/help-icon-injector'
], function (injectHelpIcon) {
});
```

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
```diff
@@ -1,37 +1,8 @@
...
-
-        dataLayer.push({
-          event: 'virtual-pageview',
-          virtualPageviewPath: contactPagePath + '&selectedChild=' + tabName
-        });
-      });
-    }
-  });
-}(window, CRM, (typeof dataLayer !== 'undefined' ? dataLayer : null)));
+define([
```
AkA84:
```js
define([
  'hrcore/page-tracker',
  'hrcore/help-icon-injector'
], function (trackContactTabVirtualPageviews) {
  trackContactTabVirtualPageviews();
});
```
please keep the folder structure flat for now, it's an unnecessary early optimization at this moment

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
```diff
@@ -1,37 +1,8 @@
...
+  'hrcore/analytics/page-tracker',
+  'hrcore/helpers/help-icon-injector'
+], function (trackContactTabVirtualPageviews) {
+  trackContactTabVirtualPageviews();
```
AkA84: sorry to say but now you gotta test if the page view tracking still works

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/analytics/page-tracker.js
```diff
@@ -0,0 +1,37 @@
+/* global dataLayer */
+/* eslint-env amd */
+
+(function (window, CRM, dataLayer) {
+  define(function () {
+    'use strict';
+
+    return trackContactTabVirtualPageviews;
```
AkA84: it's kinda pointless, just return the function directly
```js
return function () {

}
```

AkA84: same for the helper icon function

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
```diff
@@ -0,0 +1,44 @@
...
+(function (CRM) {
+  define(function () {
+    'use strict';
+
+    return injectHelpIcon;
+
+    /**
+     * Injects a help icon to a form field
+     *
+     * @param {jQuery} formClass - a class of the CiviCRM form
```
AkA84:
```js
* @param {jQuery} formClass a class of the CiviCRM form
```
i don't believe we're using the `-` as a separator (both formats are valid jsdoc formats as far as i know)

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-env amd */
+
+(function (CRM) {
+  define(function () {
+    'use strict';
+
+    return injectHelpIcon;
+
+    /**
+     * Injects a help icon to a form field
```
AkA84: it's not precise i think, aren't you appending an help icon to a form field?

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
```diff
@@ -1,37 +1,8 @@
...
-          virtualPageviewPath: contactPagePath + '&selectedChild=' + tabName
-        });
-      });
-    }
-  });
-}(window, CRM, (typeof dataLayer !== 'undefined' ? dataLayer : null)));
+define([
+  'hrcore/analytics/page-tracker',
+  'hrcore/helpers/help-icon-injector'
+], function (trackContactTabVirtualPageviews) {
```
deb1990: We are injecting `'hrcore/analytics/page-tracker'`, but while using it, we are directly calling `trackContactTabVirtualPageviews `. This is a bit confusing, as it is hard to track which is function is part of which file.

To fix this, the dependency files should always return an object, so that we can do 
```
  'hrcore/analytics/page-tracker',
  'hrcore/helpers/help-icon-injector'
], function (pageTracker) {
  pageTracker.trackContactTabVirtualPageviews();
});
```

Note: When using angular, this is not a problem, because we dont have to return anything. But here as we are returning things, we should have a namespace.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param {jQuery} formClass - a class of the CiviCRM form
+     * @param {String} accessNameSelector - a selector to the field to populate to
+     * @param {String} title - the title of the help icon
+     * @param {String} helpId - the ID of the helper
+     * @param {String} helpFile - the path to the helper template
+     */
+    function injectHelpIcon (formClass, accessNameSelector, title, helpId, helpFile) {
```
AkA84: can't this whole function be more generic? At the end of the day, don't we want to simply append an helper icon to _some_ element? Wouldn't it be more reusable (and with a shorter signature) while still be working as expected:
```js
function injectHelpIcon (selector, icon) {
  var $helpIcon;
  var $element = $(selector);

  if (!$element.length || $element.find('.helpicon').length) {
    return;
  }

  $helpIcon = $('<a href class="helpicon" title="' + icon.title + '"></a>')
    .on('click', function () {
      CRM.help('', { id: icon.helpId, file: icon.helpFile });
    });

  // Populates a space before the icon if there are no other markers
  if (!$element.find(' .crm-marker').length) {
    $element.append('&#160;');
  }

  $element(labelSelector).append($helpIcon);
}
```
Usage:
```js
injectHelpIcon('label[for="xyz"]', {
  title: 'End Date Help',
  helpId: 'hrcareer-enddate',
  helpFile: 'CRM/HRCareer/Page/helptext'
});
```

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
deb1990: Had a discussion with @AkA84 , and the following structure seems agreeable
```
define([
  'hrcore/analytics',
  'hrcore/helpers'
], function (analytics, helpers) {
  analytics.startTrackingViews();
});
```

define([
  'hrcore/helpers'
], function (helpers) {
   helpers.appendHelpIcon();
});

igorpavlov: Done

hrcareer/js/src/hrcareer/app.js
igorpavlov: Done.

hrcareer/js/src/hrcareer/app.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/app.js
igorpavlov: I tested if the func is called with a proper params. And will test again after the latest changes.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/analytics/page-tracker.js
igorpavlov: Not anymore. We now return objects.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
igorpavlov: Done

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers/help-icon-injector.js
igorpavlov: I don't think so. We have a real case at the moment when 2 extensions are injecting an icon into a form field label. The usage is very clean at the moment. However, as a compromise I would suggest this:

```
// helpers.js
return {
  appendHelpIconToCiviCRMFormLabel,
  appendHelpIcon
}

function appendHelpIcon (...) {
  // injects as you suggested
}

function appendHelpIconToCiviCRMFormLabel (...) {
  ... 
  appendHelpIcon(...);
}
```

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers.js
```diff
@@ -0,0 +1,47 @@
...
+     */
+    function appendHelpIcon ($element, title, helpId, helpFile) {
+      var $helpIcon = $('<a href class="helpicon" title="' + title + '"></a>');
+
+      $element.append($helpIcon);
+
+      $helpIcon.on('click', function () {
+        CRM.help('', { id: helpId, file: helpFile });
+
+        return false;
```
AkA84: why is this necessary? The `href` property is empty, so the click event shouldn't do anything?

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Gets a CiviCRM form field label
+     *
+     * @param  {String} formClass a class of the CiviCRM form
+     * @param  {String} accessNameSelector a selector to the field to populate to
+     * @return {jQuery}
+     */
+    function getCiviCRMFormLabel (formClass, accessNameSelector) {
```
AkA84: honestly this looks overkill to me. This method does an overly specific job that can be simply replaced by using jQuery directly
```js
$('label[for="field-name"]');
```
I don't see any complex logic here that is worth encapsulating in a method, we're simply selecting a `label`

uk.co.compucorp.civicrm.hrcore/js/src/hrcore/helpers.js
AkA84: i'm not questioning why `href` is there, i'm asking why do you need to return `false`

igorpavlov: Apologies. `return false` is needed because if you specify `href` as an empty attribute it will simply refresh the current URL if you don't return `false`.

PR #2813, opened 2018-08-02
PCHR-4075: Fix Filters on Admin - Advanced Search - Job Role Filters

ajesamson:
## Overview
Filtering contacts in advanced search by `Job Role`, without specifying any date result in contacts whose job role is no longer active. This PR fixes the issue by ensuring the job role start date is less or equal to current date while the end date is greater or equal to current date or NULL.

## Before
Filtering contact by job role returns record whose job role is inactive with reference to its start and end date whenever start and or end date range criteria is not specified.

## After
Filtering contact by job role without specifying any date filter now returns contacts whose job role is active

## Technical Details
When searching contacts record, check was carried out to see if job role filtering was used and if job roles start and end date are put into consideration.
```
if (!$filterJobRoleByDate && strpos($param[0], 'date')) {
  $filterJobRoleByDate = TRUE;
}
```

If no date criteria is specified and job role filtering is required, then job role date filtering is applied.
```
private function applyJobRoleDateFiltering(&$query) {
  $today = date('Y-m-d H:i:s');
  $dateRestriction = 'civicrm_hrjobroles.start_date <= "' . $today . '"';
  $dateRestriction .= ' AND (civicrm_hrjobroles.end_date >= "' . $today . '" OR civicrm_hrjobroles.end_date IS NULL)';
  $query->_where[] = [$dateRestriction];
}
```

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
@@ -70,15 +70,26 @@ class CRM_Hrjobroles_BAO_Query extends CRM_Contact_BAO_Query_Interface {
     }
 
     public function where(&$query) {
-        $grouping = NULL;
-        foreach ($query->_params as $param) {
-            if ($this->isAJobRoleParam($param)) {
-                if ($query->_mode == CRM_Contact_BAO_QUERY::MODE_CONTACTS) {
-                    $query->_useDistinct = TRUE;
-                }
-                $this->whereClauseSingle($param, $query);
-            }
+      $grouping = NULL;
+      $filterJobRoleByDate = $filterJobRole = FALSE;
+      foreach ($query->_params as $param) {
+        if ($this->isAJobRoleParam($param)) {
+          $filterJobRole = TRUE;
+          $paramHasDate = strpos($param[0], 'date');
+          if (!$filterJobRoleByDate && $paramHasDate) {
+            $filterJobRoleByDate = TRUE;
+          }
+
+          if ($query->_mode == CRM_Contact_BAO_QUERY::MODE_CONTACTS) {
+              $query->_useDistinct = TRUE;
+          }
+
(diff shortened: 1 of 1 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
```diff
@@ -70,15 +70,25 @@ public function from($name, $mode, $side) {
...
-                    $query->_useDistinct = TRUE;
-                }
-                $this->whereClauseSingle($param, $query);
-            }
+      $grouping = NULL;
+      $filterJobRoleByDate = $filterJobRole = FALSE;
+      foreach ($query->_params as $param) {
+        if ($this->isAJobRoleParam($param)) {
+          $filterJobRole = TRUE;
+          if (!$filterJobRoleByDate && strpos($param[0], 'date')) {
```
tunbola: Please extract this condition into a readable variable for clarity.
```php
$isJobRoleDateField = strpos($param[0], 'date');
 if (!$filterJobRoleByDate && $isJobRoleDateField)) {
...

```

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
```diff
@@ -90,6 +100,18 @@ private function isAJobRoleParam($param) {
         return false;
     }
 
+  /**
+   * Applies date filtering for active job roles
+   *
+   * @param CRM_Contact_BAO_Query $query
+   */
+    private function applyJobRoleDateFiltering(&$query) {
```
tunbola: Please check if there are existing tests for this class, if so we need to add for this change too.

com.civicrm.hrjobroles/CRM/Hrjobroles/BAO/Query.php
tunbola: Its optional anyway, just felt it would be easy to comprehend at a glance.

PR #2814, opened 2018-08-03
PCHR-4003: Hide Recruitment Menu in Admin Portal

backstageel:
## Overview
This Upgrader disables and uninstalls the HR Recruitment extension.

## Before
The HR Recruitment was active and there was a "Recruitment" menu in Admin portal

## After
The HR Recruitment will be disabled and uninstalled, so there won't be a "Recruitment" menu in Admin Portal

## Comments

- The extensions folder need to be removed after this task is released to the clients
- This upgrader will also remove all data created by the extension


tunbola:
Some comments. Does uninstalling the automatically remove the `HR Recruitment` menu, I think I saw one other PR where you had to remove that from the navigation table. 
Also can you squash the commits, some of the changes in the commits could as well be with other commit changes.
Also I see commit history of `Fixing merge conflicts with staging`, don't think this should be part of the history? The ideal thing should be to sync your Epic branch with staging and then rebase your tech task on top of the Epic branch. That way commits like this should not show up.

backstageel:
Yes, uninstalling also removes the menu

PR #2818, opened 2018-08-06
PCHR-4098: Misc BackstopJS fixes and improvements

AkA84:
This PR contains miscellaneous improvements and fixes to CiviHR's backstopjs set up

## Updating to latest stable BackstopJS version
~~Update BackstopJS to the beta version, currently `v3.5.0`. The rationale behind this decision is the same as the one described [here](https://github.com/compucorp/backstopjs-config/commit/62d3f9fa8c0e9014f77f001398e802c01c726189)~~

Using the beta version caused an issue in which a node package was not properly installed when running `npm i` in the civihr root, thus causing `gulp` commands to not execute

I couldn't find the reason for it, so instead of using the beta version I used the latest stable version (`v3.5.2`). It doesn't contain the improvement that the beta contains, but it's at least stable

## Fixed AngularJS urls 
After #2790 , the format of urls of angular apps have changed, for example:
```diff
- /civicrm/tasksassignments/dashboard#/tasks
+ /civicrm/tasksassignments/dashboard#!/tasks

- /manager-leave#/manager-leave/requests
+ /manager-leave#!/manager-leave/requests
```

which broke BackstopJS scenarios where we used those urls directly
```json
{
  "label": "SSP / My Leave / Report",
  "url": "{{siteUrl}}/my-leave#/my-leave/report"
}
```
This PR fixes the urls

## Fixed selector for edit my details button
The worked done in as part of https://github.com/compucorp/civihr-employee-portal/pull/520 changed the `href` value of the "Edit my details" button,
```diff
- [href="/my_details/nojs/view"]
+ [href="/edit-my-personal-details/js/view"]
```
which resulted in the "SSP / Edit My Details" as its script relied on the value of `href`, and didn't get updated. The PR fixes the issue by updating the selector

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/ssp-my-details.js
+++ b/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/ssp-my-details.js
@@ -6,7 +6,7 @@ module.exports = class SSPMyDetails extends SSP {
    *
    */
   async showEditMyDetailsPopup () {
-    await this.puppet.click('[href="/my_details/nojs/view"]');
+    await this.puppet.click('[href="/edit-my-personal-details/js/view"]');
     await this.puppet.waitFor('.modal-civihr-custom__section', { visible: true });
   }
 };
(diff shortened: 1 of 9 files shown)


igorpavlov:
Please remove the `enhancement` and add a `bug` label, since you say this PR "fixes" something.

igorpavlov:
You update BS to 3.5.0, could you please mention that in the PR desc?

igorpavlov:
You update BS to 3.5.0, could you please mention that in the PR desc?

PR #2819, opened 2018-08-06
PCHR-4049: Fix gulp tasks

igorpavlov:
## Overview

After https://github.com/compucorp/civihr/pull/2804 `gulp backstopjs:reference` stopped working. This PR fixes this issue. 

### Wrapping tasks in functions

The initial problem was that the "extension related" tasks (like "test", "watch" and "requirejs") were initiated even if just BackstopJS task was execution was asked. Thus, they checked for provided extension, which we never provide when running BackstopJS.

So, in order to avoid this issue we need to wrap these tasks inside functions with callbacks.

```js
gulp.task('build', function (cb) {
  var buildTasksNames = spawnTasks(['sass', 'requirejs', 'test']);
  gulp.series(buildTasksNames)(cb);
// and similar for "watch"

// ...

function spawnTasks (taskNames) {
  return taskNames.map(function (taskName) {
    return utils.spawnTaskForExtension(taskName, tasks[taskName]);
  });
}
```

### Getting rid of ".apply"

`gulp.series` accepts an array of tasks, so this is correct:

```js
gulp.series(sequence);
// where sequence is [task1, task2, task3...]
```

### Parallelising watcher

There is no need to execute watching commands in sequence, watchers may spawn in any order and the faster - the better.

```js
gulp.task('watch', function (cb) {
  var watchTasksNames = spawnTasks(['sass:watch', 'requirejs:watch', 'test:watch']);
  gulp.parallel(watchTasksNames)(cb);
```
----
✅Manual Tests - passed
⏹Jasmine Tests - N/A
⏹JS distributive files - N/A
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/gulp/utils.js
+++ b/uk.co.compucorp.civicrm.hrcore/gulp/utils.js
@@ -306,11 +306,15 @@ function setCurrentExtension (extension) {
  * Spawns a task for an extension on the fly, using the name and function provided
  * The extension specified is set as the current extension before executing the task fn
  *
- * @param {String} taskName
- * @param {Function} taskFn
- * @param {String} extension If undefined, the current extension is used
+ * @param  {String} taskName
+ * @param  {Function} taskFn
+ * @param  {String} [extension] if omitted, the current extension is used
+ * @param  {Function} [callback] if ommited, no callback will be called
+ * @return {String} updated task name as per the selected extension
  */
-function spawnTaskForExtension (taskName, taskFn, extension) {
+function spawnTaskForExtension (taskName, taskFn, extension, callback) {
+  var taskExecution;
+
   extension = extension || getCurrentExtension();
   taskName += ' (' + extension + ')';
 
@@ -319,6 +323,10 @@ function spawnTaskForExtension (taskName, taskFn, extension) {
 
     return taskFn(cb);
   });
+
+  taskExecution = gulp.series([taskName]);
+
+  (callback) && taskExecution(callback);
 
   return taskName;
 }
--- a/uk.co.compucorp.civicrm.hrcore/gulp/utils.js
+++ b/uk.co.compucorp.civicrm.hrcore/gulp/utils.js
@@ -309,12 +309,9 @@ function setCurrentExtension (extension) {
  * @param  {String} taskName
  * @param  {Function} taskFn
  * @param  {String} [extension] if omitted, the current extension is used
- * @param  {Function} [callback] if ommited, no callback will be called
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.hrcore/gulp/utils.js
```diff
@@ -320,6 +324,10 @@ function spawnTaskForExtension (taskName, taskFn, extension) {
     return taskFn(cb);
   });
 
+  taskExecution = gulp.series([taskName]);
```
AkA84: you are saying that this "executes" the task, but if that's the case then this can't be right. The whole purpose of this function was not to execute the task, but just to _define_ it.

It is a programmatic way of doing this

```js
gulp.task('my-task', function () {
  // implementation of task
});
```
this simply defines the task and nothing else, it does not execute it

uk.co.compucorp.civicrm.hrcore/gulp/utils.js
```diff
@@ -320,6 +324,10 @@ function spawnTaskForExtension (taskName, taskFn, extension) {
...
+
+  (callback) && taskExecution(callback);
```
AkA84: why do you need the `()`?

uk.co.compucorp.civicrm.hrcore/gulpfile.js
```diff
@@ -5,25 +5,31 @@ var gulp = require('gulp');
...
  */
-function buildTaskPromises (taskNames) {
+function createTasksArray (taskNames) {
   return taskNames.map(function (taskName) {
-    return utils.spawnTaskForExtension(taskName, tasks[taskName]);
+    var fn = function (cb) {
+      utils.spawnTaskForExtension(taskName, tasks[taskName], null, cb);
+    };
+
+    fn.displayName = taskName;
```
AkA84: what would happen if this line wouldn't be here?

uk.co.compucorp.civicrm.hrcore/gulp/utils.js
igorpavlov: I am not sure.

```js
// task definition
gulp.task('my-task', function () {
  // implementation of task
});

// task execution
gulp.series(['my-task']);
// or
gulp.parallel(['my-task']);
```

uk.co.compucorp.civicrm.hrcore/gulp/utils.js
igorpavlov: Just for readability.

PR #2826, opened 2018-08-16
PCHR-4032: Do not show inactive staff in the calendar months

igorpavlov:
## Overview

This PR makes staff not appear in the Leave Calendar in the months they don't have active contracts for. For example, if there is a staff that has a contract for the period `2018-01-01`-`2018-02-01`, then the staff will **not** be shown for months before and including `2017-12` and after and including`2018-03`.

Please note:
- It does **not** matter if there are leave requests for the period not covered by a contract, staff will still not be shown.
- It will not "cut" the leave request on a specific date of the contract edge, it will show the leave request for the whole month if the contract exists for this month.

For both *Before* and *After* sections the same job contract will be used:

**2018-08-01 to 2018-09-20** (does not cover October)

And the leave request:

**2018-09-03 to 2018-10-02**

## Before

Staff is shown (as well as their leave request) in October. They should not.

## After

Neither staff nor their leave requests are shown in October.

### Be able to fetch contacts IDs by job contract dates

```js
// Job Contract API service
getContactsWithContractsInPeriod: function (startDate, endDate) {
  return this.sendGET('HRJobContract', 'getcontactswithcontractsinperiod', {
    start_date: startDate,
    end_date: endDate
  })
    .then(function (response) {
      return response.values;
```

### Apply the new API to allow to use in Contact.all query

```js
function processContactFilters (filters) {
  // ...

  if (filters && filters.with_contract_in_period) {
    promises.push(getContactIdsWithContractsInPeriod(filters.with_contract_in_period));
  }

// ...

function getContactIdsWithContractsInPeriod (period) {
  return jobContractAPI.getContactsWithContractsInPeriod(
    period[0], period[1])
    .then(function (contactsWithContracts) {
      return _.map(contactsWithContracts, 'id');
```

### Apply the new Contact.all to show only needed staff in the calendar

```js
function loadContacts () {
  return Contact.all({
    // using filter here
    with_contract_in_period: [
      vm.month.days[0].date + ' 00:00:00',
      vm.month.days[vm.month.days.length - 1].date + ' 23:59:59'
    ]
  })
    .then(function (result) {
      var contactIds = _.map(_.values(result.list), 'id');
       // reducing contacts
       vm.contacts = vm.contacts.filter(function (contact) {
        return _.includes(contactIds, contact.id);
```

### Allowing everyone to fetch contact ID by Job Contracts in period

Originally, the `HRJobContract.getcontactswithcontractsinperiod` endpoint is not accessible by Managers or Staff because they lack permissions.

The permission altering hook was not working for 4 years (or never worked) because or a trailing `_` that was probably just a mistake:  https://github.com/compucorp/civihr/blame/staging/hrjobcontract/hrjobcontract.php#L370. The code inside this hook was cleaned because it was never actually executed and a new rule was added instead, the trailing underscore was also removed to make the hook work:

```php
function hrjobcontract_civicrm_alterAPIPermissions($entity, $action, &$params, &$permissions) {
  if ($entity !== 'h_r_job_contract') {
    return;
  }

  $permissions[$entity]['getcontactswithcontractsinperiod'] = ['access AJAX API'];
}
```
This now allows Staff and Managers to fetch this API data.

----

✅Manual Tests - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


org.civicrm.reqangular/js/src/common/models/contact.js
```diff
@@ -120,16 +121,43 @@ define([
          * @return {Promise}
          */
         all: function (filters, pagination, sort, additionalParams) {
+          var response;
+          var extraFilters = {};
+
+          if (filters && filters.with_contract_in_period) {
```
AkA84: @igorpavlov the `all` method already has a mechanism to deal with "foreign filter" that seems to me you are not leveraging at all, please check what `processContactFilters` does

org.civicrm.reqangular/js/src/common/services/api/contract.js
```diff
@@ -43,6 +43,15 @@ define([
 
             return contracts;
           });
+      },
+      getContactsWithContractsInPeriod: function (startDate, endDate) {
```
AkA84: you're not giving any documentation to this method

org.civicrm.reqangular/js/src/common/models/contact.js
igorpavlov: Done.

org.civicrm.reqangular/js/src/common/services/api/contract.js
igorpavlov: Done.

org.civicrm.reqangular/js/test/models/contact.spec.js
```diff
@@ -174,6 +179,45 @@ define([
...
+
+                done();
+              });
+            $rootScope.$digest();
+          });
+
+          it('calls the job contract API', function () {
+            expect(jobContractAPI.getContactsWithContractsInPeriod)
+              .toHaveBeenCalledWith(
+                sampleDates.from,
```
AkA84: what's the reason for this formatting? This
```js
expect(jobContractAPI.getContactsWithContractsInPeriod)
  .toHaveBeenCalledWith(sampleDates.from, sampleDates.to)
```
is still readable and takes less room

org.civicrm.reqangular/js/test/services/api/contract.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    describe('getContactsWithContractsInPeriod()', function () {
+      var result;
+      var sampleDates = {
+        start: '2018-01-01',
+        end: '2018-01-31'
+      };
+
+      beforeEach(function (done) {
+        spyOn(JobContractAPI, 'sendGET').and.returnValue(
+          $q.resolve(jobContractData.contactsWithContractsInPeriod));
```
AkA84:
```js
spyOn(JobContractAPI, 'sendGET').and.returnValue(
  $q.resolve(jobContractData.contactsWithContractsInPeriod)
);
```

org.civicrm.reqangular/js/src/common/models/contact.js
```diff
@@ -86,13 +103,17 @@ define([
...
-          promises.push(jobRoleContactids(_.pick(filters, jobRoleFiltersKeys)));
+          promises.push(jobRoleContactIds(_.pick(filters, jobRoleFiltersKeys)));
         }
 
         if (containsForeignFilters(filters, groupFiltersKeys)) {
           promises.push(Group.contactIdsOf(filters.group_id));
         }
 
+        if (containsForeignFilters(filters, jobContractFiltersKeys)) {
+          promises.push(getContactIdsWithContractsInPeriod(filters.with_contract_in_period));
```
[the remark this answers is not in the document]
igorpavlov: Nice catch, fixed.

org.civicrm.reqangular/js/test/models/contact.spec.js
```diff
@@ -174,6 +179,45 @@ define([
...
+        describe('when filter includes a period for contracts', function () {
+          var result, sampleContact;
+          var sampleDates = {
+            from: '2018-01-01 00:00:00',
+            to: '2018-12-31 23:59:59'
+          };
+
+          beforeEach(function (done) {
+            sampleContact = _.find(contactData.all.values, {
+              id: _.first(jobContractData.contactsWithContractsInPeriod.values).id });
```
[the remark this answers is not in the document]
igorpavlov: Done.

org.civicrm.reqangular/js/test/models/contact.spec.js
```diff
@@ -174,6 +179,45 @@ define([
...
+            Contact.all({
+              with_contract_in_period: [
+                sampleDates.from,
+                sampleDates.to
+              ]
+            })
+              .then(function (_result_) {
+                result = _result_;
+
+                done();
```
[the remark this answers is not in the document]
igorpavlov: Done

org.civicrm.reqangular/js/test/services/api/contract.spec.js
```diff
@@ -0,0 +1,58 @@
...
+        JobContractAPI.getContactsWithContractsInPeriod(
+          sampleDates.start, sampleDates.end)
+          .then(function (_result_) {
+            result = _result_;
+          })
+          .finally(function () {
+            done();
```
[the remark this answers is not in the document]
igorpavlov: Done

org.civicrm.reqangular/js/test/models/contact.spec.js
```diff
@@ -174,6 +179,45 @@ define([
...
+          it('calls the job contract API', function () {
+            expect(jobContractAPI.getContactsWithContractsInPeriod)
+              .toHaveBeenCalledWith(
+                sampleDates.from,
+                sampleDates.to
+              );
+          });
+
+          it('returns only the contacts with job contracts', function () {
+            expect(result.list).toEqual([sampleContact]);
```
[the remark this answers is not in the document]
igorpavlov: @AkA84 While I get your point, I think for the sake of consistency and for **readability** we may want to keep `it` blocks as clean as possible. Yes, this executes one extra unneeded iteration, but the `it` block stays clean out of code and I think this is a good trade-off.

org.civicrm.reqangular/js/test/models/contact.spec.js
igorpavlov: I think I had some longer code and then I just auto-replaced it. Done.

org.civicrm.reqangular/js/test/services/api/contract.spec.js
igorpavlov: @AkA84 I think this one is fine actually. We do this all the time. It does not hurt the readability but it saves a line.

org.civicrm.reqangular/js/test/models/contact.spec.js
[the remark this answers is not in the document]
AkA84: i found it way less readable having `sampleContact` decoupled from the function where it is used. It's like saying that this
```js
var myVar = 'foo'

// ... code that doesn't use myVar

function myFunc () {
  var foo = 'bar' + myVar;
}
```
is more readable than this
```js
// ... code that doesn't use myVar

function myFunc () {
  var myVar = 'foo'
  var foo = 'bar' + myVar;
}
```
if you have `sampleContact` defined outside of the `it`, you're implicitly saying that this var will be used somewhere else as well, which is not the case, thus you're breaking the reader expectation

org.civicrm.reqangular/js/src/common/models/contact.js
```diff
@@ -86,13 +103,17 @@ define([
...
         if (containsForeignFilters(filters, jobRoleFiltersKeys)) {
-          promises.push(jobRoleContactids(_.pick(filters, jobRoleFiltersKeys)));
+          promises.push(jobRoleContactIds(_.pick(filters, jobRoleFiltersKeys)));
         }
 
         if (containsForeignFilters(filters, groupFiltersKeys)) {
           promises.push(Group.contactIdsOf(filters.group_id));
         }
 
+        if (filters && filters.with_contract_in_period) {
```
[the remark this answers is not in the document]
igorpavlov: Done.

PR #2827, opened 2018-08-16
PCHR-3797: Improve Apply Core Fork Patch Script to Avoid API Rate Limit Errors

ajesamson:
## Overview
This PR changes the implementation of Apply Core Fork Patch Script from using the api to using `diff` url when creating patch. This helps to reduce tendency for API rate limit errors.

## Before
API rate limit applies to core fork request for git diff in commits when creating patch.

## After
No API rate limit is applicable to git diff request when applying core fork patch to create patch. 

## Technical Details
The api base url was changed to non api url in order to avoid rate limiting in requests.
```
REPO_BASE_URL="https://github.com/compucorp/civicrm-core"
```

### DIFF
--- a/bin/apply-core-fork-patch.sh
+++ b/bin/apply-core-fork-patch.sh
@@ -2,7 +2,8 @@
 
 set -e
 
-BASE_URL="https://github.com/compucorp/civicrm-core"
+API_URL_BASE="https://api.github.com/repos/compucorp/civicrm-core"
+REPO_BASE_URL="https://github.com/compucorp/civicrm-core"
 LAST_COMMIT_PATCHED_FILE="core-fork-last-commit-patched.txt"
 PATCH_FILE="fork-patch.diff"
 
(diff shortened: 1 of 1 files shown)


bin/apply-core-fork-patch.sh
```diff
@@ -37,7 +38,7 @@ applyPatch () {
 #   None
 #######################################
 createPatch () {
-  curl "$API_URL_BASE/compare/$1...$2" -s -H "Accept: application/vnd.github.v3.diff" > "$civiRoot/$PATCH_FILE"
+  curl "$REPO_BASE_URL/compare/$1...$2.diff" -s -H "Accept: application/vnd.github.v3.diff" > "$civiRoot/$PATCH_FILE"
```
davialexandre: That we're not using the API anymore, that `-H` option isn't necessary

ajesamson: It makes sense since any update in API (to version 4 or so) would have been catered for by URL and so will not affect the script. 👍

PR #2833, opened 2018-08-20
PCHR-4022: Leave Request Templates Should be Moved Under System Workflow Message

ajesamson:
## Overview
On new sites, the leave request templates are still displayed as user-driven messages instead of system workflow messages. This PR fixes the issue.

## Before
Leave request templates are displayed under user-driven messages.

## After
Leave request templates are now displayed under system workflow messages

## Technical Details
The leave request templates are created using managed entities and are setup after upgraders have been executed. This implies the previous [implementation](https://compucorp.atlassian.net/browse/PCHR-3312) will not find the templates for upgrade at the point of execution. This makes it difficult using managed entities for such implementation. In place of managed entities, an upgrader was implemented to setup email templates as System workflow message templates.

To achieve this, the `.mgd.php` file was renamed to `.php` in order to avoid being used as managed entity and also to retain the template content for setup as System Workflow Message Template.  The file was then moved to `message_templates` folder under the `uk.co.compucorp.civicrm.hrleaveandabsences` root directory. Existing references to the template was deleted from the `civicrm_managed` table.

```
$query = 'DELETE FROM civicrm_managed WHERE `entity_type` = "MessageTemplate"
AND module = "uk.co.compucorp.civicrm.hrleaveandabsences"';
```
`civicrm_msg_template` table was also checked if the templates are in correct state. 
```
if ($templateResult['count'] === 0) {
  $this->up1033_setupSystemWorkflowTemplate($optionValue);
}

if ($templateResult['count'] === 1) {
  $this->up1033_moveTemplatesToSystemWorkflow(array_shift($templateResult['values']), $optionValue);
}
```
The system workflow template was setup afresh if no template exist by reading the template file and creating record for each in the `civicrm_msg_template` table.
```
private function up1033_setupSystemWorkflowTemplate($optionValue) {
  $emailTemplates = CRM_Core_Resources::singleton()->getPath('uk.co.compucorp.civicrm.hrleaveandabsences')
      . '/managed_entities/EmailTemplate.php';
  $templates = include $emailTemplates;
  $templateIndex = array_search($optionValue['label'], array_column($templates, 'msg_title'));

  if ($templateIndex >= 0) {
    $template = $templates[$templateIndex];
    $template['workflow_id'] = $optionValue['id'];
    civicrm_api3('MessageTemplate', 'create', $template);

    $this->up1033_duplicateTemplateAsReserved($template, $optionValue['id']);
  }
}
```

In setups where the template exists but not as system workflow, they were migrated one after the other into a system workflow. 
```
private function up1033_moveTemplatesToSystemWorkflow($messageTemplate, $optionValue) {
  civicrm_api3('MessageTemplate', 'create', [
    'workflow_id' => $optionValue['id'],
    'is_default' => 1,
    'is_reserved' => 0,
    'id' => $messageTemplate['id'],
  ]);

  $this->up1033_duplicateTemplateAsReserved($messageTemplate, $optionValue['id']);
}
```

In both cases, a backup template was created in line with way system workflow templates are supposed to work.


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1033.php
```diff
@@ -0,0 +1,131 @@
...
+    $this->up1033_duplicateTemplateAsReserved($messageTemplate, $optionValue['id']);
+  }
+
+  /**
+   * Sets up email message template as system workflow template
+   *
+   * @param Array $optionValue
+   */
+  private function up1033_setupSystemWorkflowTemplate($optionValue) {
+    $emailTemplates = CRM_Core_Resources::singleton()->getPath('uk.co.compucorp.civicrm.hrleaveandabsences')
```
ajesamson: @davialexandre I need your advice on which directory the renamed template should be or better way of implementing this.

davialexandre: Move it a `message_templates` folder in the root folder of the L&A extension.

Note that this is far from ideal. The best thing here would be to probably split that file into smaller ones and maybe even use smarty templates for the html there, but given the amount of time around spend on this ticket, let's not do this now.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1033.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1033 {
+
+  /**
+   * Migrates message templates from managed entities to upgraders
```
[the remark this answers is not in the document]
ajesamson: Noted. Changes will be effected 👍

PR #2834, opened 2018-08-21
PCHR-4061: Allow Passing Contact Subtype as a Parameter

backstageel:
## Overview
This PR allows the user to pass contact subtype as a GET parameter when force is set to 1. Then this subtype is used as the default value on the DataSource selection Form

## Before
It was not possible to pass contact subtype as a parameter.

## After
When force is set to 1 and contactType is set to "Individual", the "Individual" checkbox is selected by default. The same happens with all other contact subtypes.
screenshot_20180821_103049


backstageel:
I closed this PR because i made a mistake on the branch name. I already created new corrctly name branch and PR: https://github.com/compucorp/civihr/pull/2835

PR #2838, opened 2018-08-23
PCHR-4131: Hide Import Menus From Staff Menu

backstageel:
## Overview
This Upgrader changes the options to use names instead of Label on navigation API calls. There was a problem in items where the names are different from Labels. This PR is an amendment to [PCHR-4059: Improve Staff Menu](https://github.com/compucorp/civihr/pull/2836)

## Before
Some Import items were not disabled on Staff menu

## After
All Import items will be disabled on Staff Menu

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1032.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1032.php
@@ -22,8 +22,8 @@ trait CRM_HRCore_Upgrader_Steps_1032 {
     ]);
     $this->up1032_disableImportMenus([
       'Import Contacts',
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1032.php
```diff
@@ -22,8 +22,8 @@ public function upgrade_1032() {
     ]);
     $this->up1032_disableImportMenus([
       'Import Contacts',
```
ajesamson: Does this also need to be changed or is this the menu name?

backstageel: This is the menu name

PR #2845, opened 2018-08-28
PCHR-4146: Adjust The Managed Entities Table

backstageel:
## Overview
When disabling/uninstalling an extension, CiviCRM needs to first check if there are managed entities associated with the extension and also disables/removes them. For some reason, for hrrecruitment the 'Application' CaseType has different ids between civicrm_case_types and civicrm_managed. So, before disabling the extension we first need to adjust these IDs.
This PR changes a previous Upgrader to first make sure that the managed case type ID has the correct value, before disabling the extension.

## Before
On test-2 site the civicrm_case_type and civicrm_managed tables had different values for 'Application' CaseType

## After
The civicrm_case_type and civicrm_managed tables has the same values for 'Application' CaseType, so the hrrecruitment extension can be safely removed.

## Technical Details
This Upgrader use CRM_Core_DAO to fetch and update the value in managed table, because there is no API for that table.


backstageel:
Yes @mickadoo seems like this CaseType was removed and then readded here:PCHR-1928, that's seem to be the cause.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1031.php
```diff
@@ -22,7 +22,9 @@ private function up1031_updateManagedEntitiesTable() {
       'return' => ['id'],
       'name' => 'Application',
     ]);
-
+    if($caseType['count']===0){
```
mickadoo: You're missing some spaces here

```
if ($caseType['count'] === 0){
```

I think it would be clearer to have an empty line before and after this if statement too

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Service/Stats/StatsGathererTest.php
```diff
@@ -9,7 +9,6 @@
 use CRM_HRLeaveAndAbsences_Test_Fabricator_LeaveRequest as LeaveRequestFabricator;
 use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
 use CRM_HRCore_Test_Fabricator_UFMatch as UFMatchFabricator;
```
mickadoo: Have you run the tests locally? It looks like you just removed the `use` statement while the test that uses it is still the same

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1031.php
```diff
@@ -8,11 +8,35 @@
...
 
+  /**
+   * Making sure that the entity_id of CaseType and Managed are equal
+   */
+  private function up1031_updateManagedEntitiesTable() {
+    $caseType = civicrm_api3('CaseType', 'get', [
+      'return' => ['id'],
+      'name' => 'Application',
+    ]);
+    if($caseType['count']===0){
```
mickadoo: You're missing some spaces here and it would look clearer with some empty lines around the `if` block

PR #2848, opened 2018-08-30
PCHR-4150: Grant HR Admin access to message template menu

ajesamson:
## Overview
In order to access the `message template menu`, permission to `access root menu items and configurations` is required. This does not give `HR admin` access to the page even though they have access to a granular permission - `edit user-driven message templates`. This PR adds the two new granular permissions to the existing permission.

## Before
before_permission

## After
after_permission

## Technical Details
The permissions for both `Communications` and `Message Templates` were modified to ensure it grants access to users with any of the following permission - 
* access root menu items and configurations
* edit system workflow message templates
* edit user-driven message templates

```
'api.Navigation.create' => [
  'id' => '$value.id',
  'permission' => implode(',', $permissions),
  'permission_operator' => 'OR'
],
```

## Comments
The permissions were created using this PR [#543](https://github.com/compucorp/civihr-employee-portal/pull/543)

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1033.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1033.php
@@ -1,0 +1,55 @@
+<?php
+
+trait CRM_HRCore_Upgrader_Steps_1033 {
+
+  /**
+   * Updates menu permissions
+   *
+   * @return bool
+   */
+  public function upgrade_1033() {
+    $permissions = [
+      'access root menu items and configurations',
+      'edit system workflow message templates',
+      'edit user-driven message templates'
+    ];
+
+    $this->up1033_updateMessageTemplatePermissions($permissions);
+    $this->up1033_updateCommunicationsPermissions($permissions);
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Steps/1033.php
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Updates message template permissions
+   *
+   * @param array $permissions
+   */
+  private function up1033_updateMessageTemplatePermissions($permissions) {
+    civicrm_api3('Navigation', 'get', [
+      'parent_id' => 'Communications',
+      'url' => 'civicrm/admin/messageTemplates?reset=1',
```
mickadoo: Do you think it would be safer to search by navigation name instead of something that might change like url?

I noticed that in `up1033_updateCommunicationsPermissions` you use `name`

ajesamson: Sure. It will be safer. I will update 👌

PR #2851, opened 2018-08-31
PCHR-3945: Implement styleguide app structure

AkA84:
## Overview
The main purpose of this PR is to adopt the [folder-by-feature structure](https://github.com/johnpapa/angular-styleguide/tree/master/a1#folders-by-feature-structure) proposed in the [AngularJS Style Guide](https://github.com/johnpapa/angular-styleguide/tree/master/a1#folders-by-feature-structure) we are following (or trying to) across the CiviHR AngularJS apps

## Before
Taking the `com.civicrm.hrjobroles` extension as an example

### Folder structure
```
com.civicrm.hrjobroles/
├── js/
│   └── src/
│       ├── job-roles/
│       │   ├── controllers/
│       │   │   ├── job-roles.controller.js
│       │   │   └── modal-dialog.controller.js
│       │   ├── filters/
│       │   ├── modules/
│       │   │   ├── job-roles.config.js
│       │   │   ├── job-roles.constants.js
│       │   │   ├── job-roles.controllers.js
│       │   │   ├── job-roles.core.js
│       │   │   ├── job-roles.filters.js
│       │   │   ├── job-roles.module.js
│       │   │   ├── job-roles.run.js
│       │   │   └── job-roles.services.js
│       │   ├── services/
│       │   └── vendor/
│       └── job-roles.js
└── views/
    ├── include/
    │   ├── add_new_role.html
    │   └── job_role_panel.html
    ├── mainTemplate.html
    └── modalDialog.html
```
### Components registration
*(By "components" here I mean the building blocks of an angular app: controllers, directives, filters, etc)*

The components are grouped by type, each type having its own module (`xyz.controllers`, `xyz.directives`, etc) under the *modules/* folder

```js
define([
  'common/angular',
  'job-roles/controllers/job-roles.controller',
  'job-roles/controllers/modal-dialog.controller'
], function (angular, JobRolesController, ModalDialogController) {
  'use strict';

  return angular.module('hrjobroles.controllers', [])
    .controller(JobRolesController.__name, JobRolesController)
    .controller(ModalDialogController.__name, ModalDialogController);
});
```

Additionally, each component has a custom property called `__name`, which is passed to the registration method (`controller()`, `directive()`, etc). This property is not angular-standard, being something that we came up so that the name of the component is not decoupled from the file where the component is defined

```js
ModalDialogController.__name = 'ModalDialogController';
ModalDialogController.$inject = [/* ... */];

function ModalDialogController (/* ... */) {
  // ...
}

return ModalDialogController;
```

### Modules location
All the modules are placed under the *{root}/js/src/{namespace}/modules/* folder

### Templates location
All the templates are placed under the *{root}/views/* folder, completely decoupled from the angular application they belong to

## After
Taking the `com.civicrm.hrjobroles` extension as an example

```
com.civicrm.hrjobroles/
└── js/
    └── src/
        ├── job-roles/
        │   ├── controllers/
        │   │   ├── add-new-role.html
        │   │   ├── job-role-panel.html
        │   │   ├── job-roles.controller.js
        │   │   ├── job-roles.html
        │   │   ├── modal-dialog.controller.js
        │   │   └── modal-dialog.html
        │   ├── filters/
        │   ├── services/
        │   ├── vendor/
        │   ├── job-roles.config.js
        │   ├── job-roles.constants.js
        │   ├── job-roles.core.js
        │   ├── job-roles.module.js
        │   └── job-roles.run.js
        └── job-roles.js
```

### Components registration
The "type" modules (`xyz.controllers`, `xyz.directives`, etc) are gone, and the components are now registered directly on the main module (`xyz`)

```js
define([
  'common/angular',
  // list of components files...
  // list of modules files ...
], function (angular, /* list of components objects */) {
  'use strict';

  angular.module('hrjobroles', [
    'hrjobroles.core',
    'hrjobroles.config',
    'hrjobroles.run',
    'hrjobroles.constants'
  ])
    .controller(JobRolesController)
    .controller(ModalDialogController)
    .filter(getActiveValues)
    .factory(dateValidation)
    .factory(filtersService)
    .factory(jobRoleService);
});
```

Additionally, the `__name` property has been removed in favour of the "object notation" way of registering components (fully supported from AngularJS v1.7)

Each component file now returns an object: the key is the name the component will be registered as, the value is the component definition itself

```js
ModalDialogController.$inject = [/* ... */];

function ModalDialogController (/* ... */) {
  // ...
}

return { ModalDialogController: ModalDialogController };
```

### Modules location
All the modules are now directly under the *{root}/js/src/{namespace}/* folder.

Being the only files placed directly under the namespace folder reinforces their central role in the application.

### Templates location
The templates are now placed together with the component they belong to, making their relationship clearer

### Only gather .spec files on karma startup
In every *test-main.js* file of each extension there was a reference to the old way we named spec files (*xyz.test.js* as opposed to the new way, *xyz.spec.js*)
```js
var TEST_REGEXP = /(spec|test)\.js$/i;
```
Given that we are not using *.test.js* anymore, the regex has been simplified
```js
var TEST_REGEXP = /spec\.js$/i;
```

### Update templates path in karma.conf
Given that the templates location had changed, the *karma.conf.js* file of each extension had to be updated as well

```diff
// angular templates
- extPath + 'views/**/*.html',
+ extPath + 'js/src/{namespace}/**/*.html',
```
```diff
preprocessors: (function (obj) {
- obj[extPath + 'views/**/*.html'] = ['ng-html2js'];
+ obj[extPath + 'js/src/{namespace}/**/*.html'] = ['ng-html2js'];
  return obj;
})({}),
```

### Remove ineffective cache busting params for templates
See https://github.com/compucorp/civihr/pull/2851#discussion_r214884464

Example of using current timestamp: https://github.com/compucorp/civihr/blob/8ee5e4230c92ead4c5fdfc8ddaed049548606304/com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js#L897

Example of using hardcoded value: https://github.com/compucorp/civihr/blob/8ee5e4230c92ead4c5fdfc8ddaed049548606304/hrjobcontract/js/src/job-contract/controllers/contract-list.controller.js#L166

---

Following is the list of changes specific for each extension

#### Removed unused constants
In `hrjobroles.constants` the following values were not used anymore, so they were removed
```diff
angular.module('hrjobroles.constants', [])
  .constant('settings', {
-   pathApp: CRM.vars.hrjobroles.path,
-   pathRest: CRM.url('civicrm/ajax/rest'),
-   pathTpl: 'views/',
-   pathIncludeTpl: 'views/include/'
  });
```
#### Initialized `settings.debug` correctly
The `settings.debug` was always `true`, no matter what the actual setting on the site was
```diff
angular.module('hrjobroles.constants', [])
  .constant('settings', {
-   debug: true,
+   debug: +CRM.debug,
  });
```

#### Removed unused backend-injected property
The `CRM.vars.hrjobroles.path` was never used, so it was removed
```diff
CRM_Core_Resources::singleton()->addVars('hrjobroles', array(
- 'path' => CRM_Core_Resources::singleton()->getUrl('com.civicrm.hrjobroles')
  'baseURL' => CRM_Extension_System::singleton()->getMapper()->keyToUrl('com.civicrm.hrjobroles')
));
```

#### Standardized name of base url settings
The standard name for the `settings` property where you can get the base url of the extension is `baseUrl`, while in job roles it was called `pathBaseUrl`. The name was then changed for consistency
```diff
angular.module('hrjobroles.constants', [])
  .constant('settings', {
-   pathBaseUrl: CRM.vars.hrjobroles.baseURL + '/js/src/job-roles/'
+   baseUrl: CRM.vars.hrjobroles.baseURL + '/js/src/job-roles/'
  });
```

#### Created `access-rights.constants`
The app was missing a `constants` module where to register the `settings` constant, like it's done for all the other AngularJS apps
```js
angular.module('access-rights.constants', [])
  .constant('settings', {
    debug: +CRM.debug,
    baseUrl: CRM.vars.contactAccessRights.baseURL + '/js/src/access-rights/'
});
```

#### Removed unused controller
There was an used controller, `ContactSummaryCtrl`, that was still hanging around. Both the controller and its test site has been deleted

#### Removed unused constants
The following constants have been removed because they weren't used anywhere
```diff
angular.module('contactsummary.constants', []).constant('settings', {
- classNamePrefix: 'contactSummary-',
- pathApp: '',
- pathRest: CRM.url('civicrm/ajax/rest'),
});
```

#### Removed superflous `contact-summary.run` module
The `contact-summary.run` module was in charge of putting a couple of values in the `$rootScope` that were never unused, thus it has been deleted

```js
angular.module('contactsummary.run', ['contactsummary.constants']).run(run);

run.$inject = ['settings', '$rootScope', '$q', '$log'];

function run (settings, $rootScope, $q, $log) {
  $log.debug('app.run');

  $rootScope.pathTpl = settings.pathTpl;
  $rootScope.prefix = settings.classNamePrefix;
}
```

#### Initialized `settings.debug` correctly
Same thing described for job roles

#### Moved L&A dependency in `contact-summary.core`
The `leave-absences.components.leave-widget` was registered directly on the main `contact-summary` module, but according to the [styleguide](https://github.com/johnpapa/angular-styleguide/tree/master/a1#module-dependencies), a better suited place for it its the `contact-summary.core` module, were cross-app modules are registered

```diff
angular.module('contactsummary.core', [
  'ngRoute',
  'ngResource',
  'ui.bootstrap',
- 'common.services'
+ 'common.services',
+ 'leave-absences.components.leave-widget'
]);
```

#### Moved require of *notification.service.js* into *job-contract.core.js*
Being a cross-app dependency, the file for the notification service is best required in the file of `job-contract.core`, given also that the module already has an explicit dependency on the `common.services` module

#### Moved php-injected vars under CRM.vars
The standard place for an AngularJS app for accessing the php-injected vars of its extension is under the `CRM.vars.{extension}` global property (ie `CRM.vars.contactAccessRights`, `CRM.vars.hrjobroles`, etc)

Job Contract though was placing its vars in a different place, `CRM.jobContractTabApp`

```php
  // ...
  ->addSettingsFactory(function () {
    return array(
      // ...
      'jobContractTabApp' => array(
        // ...
      )
    );
  });
```

For standardization's sake, the vars of Job Contracts are now placed under `CRM.vars.hrjobcontract`

```php
CRM_Core_Resources::singleton()->addVars('hrjobcontract', array(
  // ...
);
```


com.civicrm.hrjobroles/js/test/services/date-validation.service.spec.js
```diff
@@ -102,19 +102,19 @@ define([
...
 
     it('Start date cannot be higher', function () {
-            /* Should pass */
+      /* Should pass */
       expect(function () {
         dateValidation.validate('12/11/2003', '10/01/2009');
       }).not.toThrow();
 
-            /*  Should NOT pass. Start date cannot be higher */
+      /*  Should NOT pass. Start date cannot be higher */
```
igorpavlov: I am confused with these comments. What does "should not pass" mean in this case? Does it mean "should throw an exception"? If yes, I suggest to get rid of these comments, it is clear from the code what should happen.

Same for other places

com.civicrm.hrjobroles/js/test/test-main.js
```diff
@@ -1,4 +1,4 @@
-var TEST_REGEXP = /(spec|test)\.js$/i;
+var TEST_REGEXP = /spec\.js$/i;
```
igorpavlov: `/\.spec\.js$/i`

Otherwise smth like `autospec.js` will also be included.

com.civicrm.hrjobroles/js/src/job-roles/job-roles.constants.js
```diff
@@ -10,11 +10,7 @@ define([
     .constant('settings', {
       classNamePrefix: 'hrjobroles-',
       contactId: decodeURIComponent((new RegExp('[?|&]cid=([^&;]+?)(&|#|;|$)').exec(location.search) || [null, ''])[1].replace(/\+/g, '%20')) || null,
-      debug: true,
-      pathApp: CRM.vars.hrjobroles.path,
-      pathRest: CRM.url('civicrm/ajax/rest'),
-      pathBaseUrl: CRM.vars.hrjobroles.baseURL + '/',
-      pathTpl: 'views/',
-      pathIncludeTpl: 'views/include/'
+      debug: +CRM.debug,
```
igorpavlov: CRM is a global variable, we usually wrap the whole thing into a function and then pass the variable. I am not saying I am a huge fan of doing this, but that would be consistent.

```js
/* eslint-env amd */

(function (CRM) {
  define([
    // ...
  ], function (/* ... */) {
    // ...
  });
}(CRM));

```

hrjobcontract/CRM/Hrjobcontract/Page/JobContractTab.php
```diff
@@ -19,6 +19,19 @@ static function registerScripts() {
 
     $loaded = TRUE;
 
+    CRM_Core_Resources::singleton()->addVars('hrjobcontract', array(
+      'baseURL' => CRM_Core_Resources::singleton()->getUrl('org.civicrm.hrjobcontract'),
+      'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer'),
+      'domainId' => CRM_Core_Config::domainID(),
+      'isLogEnabled'    => (bool) $config->logging,
```
igorpavlov: Could you please fix the spacing?

`'isLogEnabled' => (bool) $config->logging,`

hrjobcontract/js/src/job-contract/controllers/contract-list.html
```diff
@@ -51,7 +51,7 @@ <h3 class="header-bar header-bar-danger" ng-show="contractPast.length">Past Cont
         </div>
       </div>
       <div class="row">
-        <div class="col-xs-12"><hr></div>
+        <div class="col-xs-12"><hr /></div>
```
igorpavlov: The closing slash is not needed:
https://www.w3.org/TR/html/syntax.html#void-elements

hrjobcontract/js/test/services/contract-details.service.spec.js
```diff
@@ -2,22 +2,21 @@
...
-    beforeEach(module('job-contract'));
+    beforeEach(module('job-contract', 'job-contract.templates'));
     beforeEach(inject(function (_contractDetailsService_, _$httpBackend_, _$rootScope_) {
       contractDetailsService = _contractDetailsService_;
       $httpBackend = _$httpBackend_;
       $rootScope = _$rootScope_;
 
       $httpBackend.whenGET(/action=get&entity=HRJobContract/).respond(200);
       $httpBackend.whenGET(/action=get&entity=HRJobDetails/).respond(MockContract.contract);
-      $httpBackend.whenGET(/views.*/).respond({});
```
igorpavlov: This deletion is not mentioned in the PR description, why did we need to mock the backend for `views` and now not?

hrjobcontract/js/test/test-main.js
```diff
@@ -1,8 +1,8 @@
 
-var TEST_REGEXP = /(spec|test)\.js$/i;
+var TEST_REGEXP = /spec\.js$/i;
```
igorpavlov: Same as above `/\.spec\.js$/i`

hrjobcontract/js/test/services/contract-details.service.spec.js
AkA84: The reason why the calls to `views/` were mocked was because the `xyz.templates` module was not used, though available. This module is created on the fly by karma exactly to avoid http requests to templates
```js
// Used to transform angular templates in JS strings
preprocessors: (function (obj) {
  obj[extPath + 'js/src/job-contract/**/*.html'] = ['ng-html2js'];
  return obj;
})({}),
```

Now that the `views/` folder is gone, and we are using the `xyz.templates` module in the test files, we don't need to mock the calls anymore

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -894,7 +893,7 @@ define([
       var modalInstance = $modal.open({
         appendTo: $rootElement.find('div').eq(0),
         template: '',
-        templateUrl: settings.pathApp + 'views/modalDialog.html?v=' + (new Date()).getTime(),
```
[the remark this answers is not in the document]
AkA84: I believe that the cache busting strategy was really not much of a strategy. In some places we were using the current timestamp (meaning the requests were never cached), in some places the cache busting param was hardcoded and never changed (meaning the requests were always cached)

I thought of wiping the slate clean and removing all cache busters params to start, and then later come up with a strategy worth of its name.

For example there is really no reason why we are not using the `$templateCache` service in production, so we could avoid unnecessary http requests. Meanwhile we should keep fetching the templates from the wire when developing. Given that now all extensions have a `settings.debug` property, you can see that we could use that to decide how to serve the templates. 

As I said, this looks something worth a separate ticket, but I should have written it in the PR description. If this sounds reasonable to you, I'll add it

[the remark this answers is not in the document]
igorpavlov: @AkA84 yeah, sounds right, thanks. I am having only one concern, but this is not in the scope of you ticket. The concern is new releases. Are we flushing cache with new releases?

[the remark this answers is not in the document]
AkA84: @igorpavlov i'm really not sure tbh, but given that the vast majority of the templates had hardcoded cache busters (= they never really busted the cache) and so far we didn't hear of any issues with the templates, I have to assume that template cache had not been an issue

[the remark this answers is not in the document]
AkA84: update the PR description

PR #2854, opened 2018-09-05
PCHR-3128: Delete the org.civicrm.hrabsence extension

ajesamson:
## Overview
`org.civicrm.hrabsence` extension was  previously disabled and all functionality migrated to Leave and Absences. This PR completely removes the extension.

## Before
before_absences

## After
after_absences

## Technical Details
The directory was deleted and references removed.  The change [here](https://github.com/compucorp/civihr/blob/83579de87a059d7cbebdb27e5996e4a2fa292f18/Jenkinsfile#L474) was also reverted.


hrreport/CRM/HRReport/Upgrader.php
```diff
@@ -255,4 +234,34 @@ public function upgrade_1403() {
     }
     return TRUE;
   }
+
+  /**
+   * Removes unused absence report
```
davialexandre: s/report/reports/

hrreport/CRM/HRReport/Upgrader.php
```diff
@@ -255,4 +234,34 @@ public function upgrade_1403() {
...
+      'return' => ['name'],
+      'option_group_id' => 'report_template',
+      'name' => ['IN' => $templates],
+    ]);
+
+    if ($reports['count'] > 0) {
+      $templateIds = array_column(array_values($reports['values']), 'id');
+      foreach ($templateIds as $templateId) {
+        civicrm_api3('OptionValue', 'delete', ['id' => $templateId]);
+      }
```
davialexandre: If you check the `.mgd.php` files you've deleted, you'll see that a report is composed of a `ReportTemplate` (The option value you're deleting here) and a `ReportInstance`. You need to delete the report instance too.

hrreport/CRM/HRReport/Upgrader.php
```diff
@@ -255,4 +234,34 @@ public function upgrade_1403() {
...
+
+      $sql = "DELETE FROM `civicrm_managed` WHERE `entity_id` IN ('" . implode("','", $templateIds) . "')";
```
davialexandre: The `entity_id` field is not unique. This query will delete more than you want. Also, keep in mind that the `ReportInstance` instances are also managed entities, so you'll need to remove them here as well

hrreport/CRM/HRReport/Upgrader.php
ajesamson: Alright. I can see they come in pairs of `Template` and `Instance`. Will remove the instances as well. 🗑

hrreport/CRM/HRReport/Upgrader.php
```diff
@@ -247,21 +247,45 @@ public function upgrade_1404() {
...
     if ($reports['count'] > 0) {
       $templateIds = array_column(array_values($reports['values']), 'id');
+      $templateLabels = array_column(array_values($reports['values']), 'label');
       foreach ($templateIds as $templateId) {
         civicrm_api3('OptionValue', 'delete', ['id' => $templateId]);
       }
 
-      $sql = "DELETE FROM `civicrm_managed` WHERE `entity_id` IN ('" . implode("','", $templateIds) . "')";
+      $sql = "DELETE FROM `civicrm_managed` WHERE `entity_type` = 'ReportTemplate'";
+      $sql .= " AND `name` IN ('" . implode(" Template','", $templateLabels) . " Template')";
```
davialexandre: why did you change this to delete reports based on the labels instead of IDs? The only change necessary here was to add `entity_type = 'ReportTemplate'` to the query criteria

hrreport/CRM/HRReport/Upgrader.php
```diff
@@ -247,21 +247,45 @@ public function upgrade_1404() {
...
 
+    $this->_up1404RemoveReportInstance();
+
     return TRUE;
   }
+
+  /**
+   * Removes absence report instances
+   */
+  private function _up1404RemoveReportInstance() {
```
davialexandre: Please never prefix private methods with an underscore. Prefixing a method in a upgrader class is only necessary when we are writing upgrader steps in traits, to avoid conflicts with methods in other traits. Since that is not the case here, you can just name the method as `removeReportInstances()`

ajesamson: Noted. Thanks 👍

hrreport/CRM/HRReport/Upgrader.php
ajesamson: That must have been a misinterpretation 🤷‍♂️ .  More so, the report instances was using names.  Will revert back to id with entity_type clause. ♻️

PR #2855, opened 2018-09-06
PCHR-4068: Improve Job Contract and Job Roles Filter Fields

backstageel:
## Overview
This PR changes the multiple select fields from Job Contract and Job Roles to use select2 plugin on Advanced Search(Staff Directory)

## Before
These fields used normal Multiple select box field
screenshot_20180906_103655

## After
These fields are using select2 in multiple select box fields
screenshot_20180906_123118


ajesamson:
@backstageel The code snippet looks good (💯) but i will suggest you pay attention to the UI regression 🤕. The view in the `After` section (☝️) is not well aligned. Also, i hope you checked all the affected pages where the filter is used.

PR #2859, opened 2018-09-13
WIP: PCHR-4182: Menu Improvements

tunbola:
## Overview
Currently the process of adding/modifying a menu item is a bit complex and takes more time than necessary. This task hopes to simplify the process by making it easier to add or modify a menu item.

## Approach
The approach taken is to swap the underlying data source that returns the menu tree array and feed this to Civi so that it can build the Navigation menu for display as usual. 
So rather than fetching the data from the `civicrm_navigation` table, the data would be supplied via a Menu configuration file that returns the menu Items in an array format.

There is also the case of dynamic menu Items whereby the menu items are generated based on some conditions. In a bid to keep things as simple as possible.

A `CRM_HRCore_Menu_Main` class was created. It has a `getItems` function that returns the menu items, the `getItems` function uses the 
(1) `CRM_HRCore_Menu_CustomFields` class, an example of how dynamic menu items can be returned.
(2) `CRM_HRCore_Menu_SystemSettings` class, an example of how the long Menu.php file can be splitted into several files by moving out menu items with particularly large item sets.

## Injecting the Menu Items
The next hurdle was to feed the returned menu items array to Civi so that the Navvigation menu is displayed as usual. In generating the Menu items array, the same structure gotten from the `Navigation.get` API is used but unnecessary fields such as ParentID(since the array have nested items now), NavID, active e.t.c were removed from the array structure. 

The `buildNavigation` method of the `CRM_Core_BAO_Navigation` class (already overriden file in the HRUI extension) was modified to swap in the data source from getting the navigation menu tree  from the db to getting it from `CRM_HRCore_Menu_Main::getItems()`  function. with a few changes to other places in the file due to the unnecessary menu properties that were discarded.

## Alternative approach to Injecting the Menu Items.
In a bit to reduce changes/overrided to core to the barest minimun, an alternative I considered and tried out is to comment out all implementations of the `hook_civicrm_navigationMenu` for all the extensions and implement this hook only in HRCore

```php
function hrcore_civicrm_navigationMenu(&$params) {
  $params = CRM_HRCore_Menu_Main::getItems();
}
```
This approach had its downside because, it means the menu Items will still be fetched from civicrm_navigation table but only overriden in this hook and it would still involve changes to the `buildNavigation` method of the `CRM_Core_BAO_Navigation` because the items came out in a different order on the UI(Because the weight parameter was missing from the array supplied) because it was re-ordered by Civi.

### DIFF
--- a/hrui/CRM/Core/BAO/Navigation.php
+++ b/hrui/CRM/Core/BAO/Navigation.php
@@ -306,25 +306,26 @@ FROM civicrm_navigation WHERE domain_id = $domainID";
    * @return string
    */
   public static function buildNavigation() {
-    $navigations = self::buildNavigationTree();
+//    $navigations = self::buildNavigationTree();
     $navigationString = '';
 
-    // run the Navigation  through a hook so users can modify it
-    CRM_Utils_Hook::navigationMenu($navigations);
-    self::fixNavigationMenu($navigations);
-
-    // Hooks have added menu items in an arbitrary order. We need to order by
-    // weight again. I would put this function directly after
-    // CRM_Utils_Hook::navigationMenu but for some reason, fixNavigationMenu is
-    // moving items added by hooks on the end of the menu. Hence I do it
-    // afterwards
-    self::orderByWeight($navigations);
+//    // run the Navigation  through a hook so users can modify it
+//    CRM_Utils_Hook::navigationMenu($navigations);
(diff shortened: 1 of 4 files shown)


davialexandre:
Closing, as this was only a POC.The actual improvement was merged as part of https://github.com/compucorp/civihr/pull/2869

PR #2860, opened 2018-09-14
PCHR-3124: Delete Appraisals extension

ajesamson:
## Overview
`uk.co.compucorp.civicrm.appraisals` extension was previously disabled. This PR completely removes the extension references from CiviHR.

## Before
Some references to the disabled appraisals extension existed in CiviHR.

## After
References to the Appraisals in CiviHR are now removed.

## Technical Details
A rule was added to the `build.mock.js` file dues to an error message: `Expected an assignment or function call and instead saw an expression.` The error was raised after removing appraisal mocks and committing the code. The code looks fine but one of the linters used is raising exception which is why it was added.


org.civicrm.reqangular/js/src/common/services/api.js
```diff
@@ -140,7 +140,7 @@ define([
       /**
        * Sends a GET request to the backend endpoint
        *
-       * @param  {String} entity - The entity the request is asking for (Contact, Appraisal, etc)
+       * @param  {String} entity - The entity the request is asking for (Contact, etc)
```
igorpavlov: this comments does not make sense anymore, please change to smth like:

`The entity the request is asking for (e.g. Contact)`

also please check other similar places

hrcase/CRM/HRCase/Upgrader.php
```diff
@@ -352,12 +352,10 @@ private function up1402_removedUnusedManagedEntities($caseTypes, $activityTypes)
...
   private function up1402_removeUnusedCaseTypes() {
-    // Remove (Appraisal) & (Probation) case types
-    $caseTypesToRemove = ['Appraisal', 'Probation'];
+    // Remove (Probation) case types
+    $caseType = 'Probation';
 
-    foreach($caseTypesToRemove as $caseType) {
-      $this->removeUnusedCaseType($caseType);
-    }
+    $this->removeUnusedCaseType($caseType);
```
davialexandre: You don't really need the variable here anymore, right? You can just `$this->removeUnusedCaseType('Probation')`

org.civicrm.bootstrapcivihr/scss/components/_appraisal-cycle-summary.scss
```diff
@@ -1,20 +0,0 @@
-.chr_appraisal-cycle-summary__details {
```
davialexandre: Since you've removed this, please recompile the css file for this extension

org.civicrm.reqangular/js/src/common/bundles/apis.js
```diff
@@ -1,8 +1,6 @@
 /* eslint-env amd */
 
 define([
-  'common/services/api/appraisal',
```
davialexandre: Please rebuild the JS distributive files

PR #2865, opened 2018-09-18
PCHR-4188: Add tab fields map to the Leave Type Wizard and render

igorpavlov:
## Overview

This PR prepares a skeleton of the Leave Type Wizard by providing it with a map of fields and generic UI elements.

Please also see https://github.com/compucorp/org.civicrm.shoreditch/pull/129 for a new `respect-line-breaks` CSS class.

Please also see https://github.com/compucorp/org.civicrm.shoreditch/pull/130 for accordion styling.

## Technical Details

The settings tabs and fields map is probably the most important part of this PR since a lot of things happen around it. It described the structure of fields which allows to loop through it and automate the logic of the user flow.

```js
// the map
{
  name: 'customise',
  label: 'Customise',
  nextSectionButtonCustomText: 'Go forward',
  previousSectionButtonCustomText: 'Go back',
  tabs: [
    {
      label: 'Basic',
      fields: [
        {
          name: 'title',
          label: 'Title',
          helpText: [
            'Sentence 1',
            'Sentence 2',
            LINE_BREAK,
            'Line 2'
          ].join(' ')
```

Then we use this map to render tabs, fields and even actions:

```html
<div class="panel-group">
  <uib-accordion>
    <!-- Section -->
    <div uib-accordion-group class="panel-default {{section.tabs.length > 1 && 'panel-w-vertical-tabs'}}"
        ng-repeat="(sectionIndex, section) in form.sections"
        is-open="section.active"
        ng-click="form.openSection(sectionIndex)">
      <uib-accordion-heading>
        {{section.label}}
      <div class="row" ng-click="$event.stopPropagation()">
        <!-- Section Tabs here -->

<!-- Section Tabs -->
<!-- If more then one tab - draw vertical pills -->
<div class="col-md-3" ng-if="section.tabs.length > 1">
  <ul class="nav nav-pills nav-stacked">
    <li ng-repeat="(tabIndex, tab) in section.tabs"
      ng-class="{ active: tab.active }"
      ng-click="form.openActiveSectionTab(tabIndex)">
      <a href>
        {{tab.label}}

<!-- Section Tab fields -->
<!-- If there is just one tab, spread the form to 100% width -->
<div class="form-horizontal"
  ng-class="{ 'col-md-9': section.tabs.length > 1, 'col-md-12': section.tabs.length === 1 }">
  <div class="form-group" ng-repeat="field in form.getFieldsForActiveTab()">
    <label class="control-label"
      ng-class="{ 'col-md-3': !field.labelLayout, 'col-md-12': field.labelLayout === 'horizontal' }">
      {{field.label}}
      <help-text ng-if="field.helpText">
        <p class="respect-line-breaks">{{field.helpText}}</p>
    <div ng-class="{ 'col-md-9': !field.labelLayout, 'col-md-12': field.labelLayout === 'horizontal' }">
      <!-- We do not draw real fields just yet -->
      [template of the field]
  <hr ng-if="section.tabs.length > 1"/>
  <!-- Actions are reused for tabs, but can be customised -->
  <div class="row">
    <div class="col-md-12">
      <button class="btn btn-secondary pull-right"
        ng-if="!form.isOnSectionLastTab"
        ng-click="form.openNextActiveSectionTab()">
        Next
      <button class="btn btn-primary pull-right"
        ng-if="form.isOnSectionLastTab"
        ng-click="form.openNextSection()">
        {{section.nextSectionButtonCustomText || 'Next section'}}
        <i class="fa fa-chevron-right"
          ng-if="section.nextSectionButtonChevron">
      <button class="btn btn-secondary"
        ng-if="form.isOnSectionFirstTab"
        ng-click="form.openPreviousSection()">
        {{section.previousSectionButtonCustomText || 'Previous section'}}
      <button class="btn btn-secondary"
        ng-if="!form.isOnSectionFirstTab"
        ng-click="form.openPreviousActiveSectionTab()">
        Back
```

And a couple of public functions to make it interactive:

```js
// returns fields to render
function getFieldsForActiveTab () {
  var activeSection = getActiveSection();
  var activeTab = _.find(activeSection.tabs, { active: true });

  return activeTab.fields;
}

// When you click "Next" button
function openNextActiveSectionTab () {
  var activeTabIndex = getActiveTabIndex();

  openActiveSectionTab(activeTabIndex + 1);
}

// When you click "Settings >" button
function openNextSection (currentIndex) {
  var activeSectionIndex = getActiveSectionIndex();

  openSection(activeSectionIndex + 1);
}

// When you click "Back" button
function openPreviousActiveSectionTab () {
  var activeTabIndex = getActiveTabIndex();

  openActiveSectionTab(activeTabIndex - 1);
}

/**
 * Opens previous section
 */
function openPreviousSection () {
  var activeSectionIndex = getActiveSectionIndex();

  openSection(activeSectionIndex - 1);
}

// When you click on a section header
function openSection (sectionIndex) {
  vm.sections.forEach(function (section) {
    section.active = false;
  });

  vm.sections[sectionIndex].active = true;

  openActiveSectionTab(0);
}

// When you click on a tab pill
function openActiveSectionTab (tabIndex) {
  var activeSection = getActiveSection();
  var tabs = activeSection.tabs;

  tabs.forEach(function (tab) {
    tab.active = false;
  });

  tabs[tabIndex].active = true;
  vm.isOnSectionLastTab = tabIndex === tabs.length - 1;
  vm.isOnSectionFirstTab = tabIndex === 0;
}
```

### Moving the component to a different place

The component/module is injected in an existing page for the ease of development. It was moved from the Add Absence Type to List Absence Types page because it is not wrapped in a `<form>` element. Anyway, this is temporary for the sake of the Leave Type Categories epic development.

### Using class instead of ng-class

Here
```html
<div uib-accordion-group class="panel-default {{section.tabs.length > 1 && 'panel-w-vertical-tabs'}}"
```
you may notice that we use this odd class condition. This is needed because we need this class, but we cannot use `ng-class` because it is disrespected by the `uib-accordion-group` directive.

---

✅Manual Tests - passed
⏹Backstop tests - suppressed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
```diff
@@ -0,0 +1,52 @@
...
+<div class="row">
+  <div class="col-md-12">
+    <label>What kind of absence type are you looking to create?</label>
+  </div>
+</div>
+<div class="row">
+  <div class="col-md-2"
+    ng-repeat="leaveTypeCategory in form.leaveTypeCategories">
+    <div class="well pointer text-center"
+      ng-click="form.leaveTypeCategory = leaveTypeCategory.value">
```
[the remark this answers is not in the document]
igorpavlov: `<label>` will require additional CSS because it will not render 100% width of the container. I was thinking about this and think it would make sense to optimise in the next story where we have multiple switchers.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
+<div class="row">
+  <div class="col-md-3">
+    <ul class="nav nav-pills nav-stacked">
+      <li ng-repeat="settingsTab in form.settingsTabs"
+        ng-class="{ active: settingsTab.opened }"
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+        ng-click="form.openSettingsTab(settingsTab.name)">
```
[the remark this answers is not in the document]
igorpavlov: Done. I haven't renamed to `openSectionByName` because that may imply we also have other `openSectionBy[Smth]` functions

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,162 @@ define([], function () {
...
+    vm.openSection = openSection;
+    vm.openSettingsTab = openSettingsTab;
+    vm.selectLeaveTypeCategory = selectLeaveTypeCategory;
+
+    function $onInit () {
+      initDefaultView();
+    }
+
+    /**
+     * Gets fields for the currently opened settings tab
```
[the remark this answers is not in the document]
igorpavlov: No, you do not know what fields are until you get them. You just know they are fields, but you don't know anything particular about them. That's why no `the`.

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/leave-type-wizard/form/components/leave-type-wizard-form.component.spec.js
```diff
@@ -29,11 +30,114 @@ define([
       expect($log.debug).toHaveBeenCalledWith('Controller: LeaveTypeWizardFormController');
     });
 
+    describe('on init', function () {
+      var generalSection, settingsSection, basicSettingsTab, leaveRequestsSettingsTab;
+
+      beforeEach(function () {
+        generalSection = _.find(controller.sections, { name: 'general' });
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+      </li>
+    </ul>
+  </div>
+  <div class="col-md-9">
+    <div class="row" ng-repeat="field in form.getSettingsTabFields()">
+      <div class="col-md-3">
+        <label>
+          {{field.label}}
+          <help-text ng-if="field.helpText">
+            <p class="respect-line-breaks">{{field.helpText}}</p>
```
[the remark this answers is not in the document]
igorpavlov: My fault, I forgot to reference the PR. https://github.com/compucorp/org.civicrm.shoreditch/pull/129

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.html
```diff
@@ -1,3 +1,24 @@
...
+
+  <div class="panel-group">
+    <div class="panel panel-default"
+      ng-repeat="section in form.sections">
+      <div class="panel-heading"
+        ng-class="{ pointer: !section.active }"
+        ng-click="form.openSection(section.name)">
+        <h4 class="panel-title">
+          <small>
+            <i class="fa fa-chevron-{{section.active ? 'down' : 'right'}}"></i>
```
AkA84: i know the UI is not finalized, but FYI this is not the standard markup we have for collapsible panels (check the styleguide)

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
```diff
@@ -0,0 +1,52 @@
...
+      </help-text>
+    </label>
+  </div>
+  <div class="col-md-4">
+    <input type="text" class="form-control"/>
+  </div>
+</div>
+<div class="row">
+  <div class="col-md-12">
+    <label>What kind of absence type are you looking to create?</label>
```
AkA84: be aware that this is not how you build horizontal forms in bootstrap (check either our style guide of the bootstrap documentation)

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
+<div class="row">
+  <div class="col-md-3">
+    <ul class="nav nav-pills nav-stacked">
+      <li ng-repeat="settingsTab in form.settingsTabs"
+        ng-class="{ active: settingsTab.active }"
+        ng-click="form.openSettingsTab(settingsTab.name)">
+        <a href="#">
```
AkA84: i think `<a href>` should be sufficient

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+        </label>
+      </div>
+      <div class="col-md-9">
+        [template of the field]
+      </div>
+    </div>
+    <hr/>
+    <div class="row">
+      <div class="col-md-12">
+        <div class="btn btn-secondary pull-right"
```
AkA84: use `<button>` given that they are buttons

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+          ng-if="!form.isOnLastSettingsTab"
+          ng-click="form.openSettingsTab(form.settingsTabs[form.activeSettingsTabIndex + 1].name)">
```
AkA84: why the need to delegate the indexing logic to the view? Something like this
```html
ng-click="form.openNextTab()"
```
would be much cleaner and you would keep the pagination logic in the component controller instead of leaking it into the view

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+          Next
+        </div>
+        <div class="btn btn-primary pull-right"
+          ng-if="form.isOnLastSettingsTab">
+          Save
```
AkA84: What will be the `ng-click` handler for this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+        <div class="btn btn-secondary"
+          ng-if="form.isOnFirstSettingsTab"
+          ng-click="form.openSection('general')">
```
AkA84: Why the need to hardcode the section name? What if it changes in the future? Can't we have
```html
ng-click="form.openPrevSection()">
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
```diff
@@ -0,0 +1,52 @@
...
+          Choose another category
+        </div>
+        <div class="btn btn-secondary"
+          ng-if="!form.isOnFirstSettingsTab"
+          ng-click="form.openSettingsTab(form.settingsTabs[form.activeSettingsTabIndex - 1].name)">
```
AkA84:
```html
ng-click="form.openPrevTab()"
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
     }]
   };
 
-  function LeaveTypeWizardFormController ($log) {
+  function LeaveTypeWizardFormController ($log, sharedSettings) {
     $log.debug('Controller: LeaveTypeWizardFormController');
+
+    var vm = this;
+    var lineBreak = '\n\n';
```
AkA84:
```js
var LINE_BREAK = '\n\n';
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
```diff
@@ -0,0 +1,52 @@
...
+  <div class="col-md-2"
+    ng-repeat="leaveTypeCategory in form.leaveTypeCategories">
+    <div class="well pointer text-center"
+      ng-click="form.leaveTypeCategory = leaveTypeCategory.value">
+      <i class="leave-type-wizard__category-selector_icon fa fa-{{leaveTypeCategory.icon}}"></i>
+      <div class="radio">
+        <label>
+          <input type="radio"
+            ng-model="form.leaveTypeCategory"
+            checked="{{form.leaveTypeCategory === leaveTypeCategory.value}}"
```
AkA84: should AngularJS infer automatically if the checkbox should be checked or not (https://docs.angularjs.org/api/ng/input/input%5Bradio%5D)

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+
+    vm.leaveTypeCategory = '';
+    vm.leaveTypeCategories = [
```
AkA84: is there any plan to fetch these from the backend at some point?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+      {
+        value: 'leave',
+        label: 'Leave',
+        icon: 'plane'
+      }
+    ];
+    vm.leaveTypeTitle = '';
```
AkA84: what is this for? Can't seem to find it anywhere else in the code

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+        label: 'Leave Category'
+      },
+      {
+        name: 'settings',
+        label: 'Leave Category Settings'
+      }
+    ];
+    vm.sectionsTemplatesPath =
+      sharedSettings.sourcePath + 'leave-type-wizard/form/components/form-sections';
+    vm.settingsTabs = [
```
AkA84: I think both `vm.settingsTab` and `vm.sections` should be better placed in a `leave-type-wizard.constants` module, so you can separate the data from the business logic

AkA84: This object 
```js
{
  name: 'settings',
  label: 'Leave Category Settings'
}
```
and `vm.settingsTab` are decoupled, even though they are obviously related: you are listing the fields of that section defined in the first object. Doesn't it make sense then to make the fields a property of the section object?

The issue with decoupling is more apparent when you see that you had to write a `getSettingsTabFields` method, hardcoding the reference to that particular section both in the method name and its implementation. What if the section change name? Or you have more sections? I think the code should be reworked to be more flexible

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+    function $onInit () {
+      initDefaultView();
+    }
+
+    /**
+     * Gets fields for the currently active settings tab
+     *
+     * @return {Array} collection of fields
+     */
+    function getSettingsTabFields () {
```
AkA84: given my previous comment, this would be benefit from being made more generic, something like
```js
function getActiveTabFieldsOf(sectionName) {

}

getActiveTabFieldsOf('settings');
```
in any case the name is doubly misleading:
* it doesn't say anything about looking for the fields of the _active_ tab
* settings is a _section_ not a tab

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+     * Initiates the default view:
+     * - selects Leave category;
+     * - expands the General section and leaves the Settings section collapsed;
+     * - selects Basic Details settings tab.
+     */
+    function initDefaultView () {
+      vm.leaveTypeCategory = 'leave';
+
+      openSection('general');
+      openSettingsTab('basic-details');
```
AkA84: don't hardcode which section or tab, you know you want to open the first of both, so use indexes

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+    }
+
+    /**
+     * Opens a section by its name and collapses all other sections
```
AkA84: but i thought that using Bootstrap's `.panel-group` was exactly for that? Are we reimplementing something that bootstrap already gives us?

https://github.com/civicrm/org.civicrm.shoreditch/blob/master/base/scss/vendor/bootstrap/_panels.scss#L219-L220

Also, remember we have access to angular's ui-bootstrap: http://angular-ui.github.io/bootstrap/#!#accordion

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
```diff
@@ -12,7 +14,165 @@ define([], function () {
...
+
+      _.find(vm.sections, { name: sectionName }).active = true;
+    }
+
+    /**
+     * Opens a settings tab by its name and collapses all other settings tabs
+     *
+     * @param {String} settingsTabName
+     */
+    function openSettingsTab (settingsTabName) {
```
AkA84: could you make the function more generic without referencing the "settings" section at all? It would make the code more flexible. Something like
```js
function openSectionTab(sectionName, tabName) {

}
```
this would also get rid of all the "settings" hardcoded references of `vm.settingsTab` and `vm.activeSettingsTabIndex` etc

AkA84: I just realized, aren't you reinventing the wheel? [`ui.bootstrap`](http://angular-ui.github.io/bootstrap/#!#tabs) already gives you a way to have functioning tabs out of the box, why can't we use it?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: I didn't provide any yet. It will be in other PRs. There will be something like `form.save()`

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: No, I don't think so. At least not in the mid-term. But I was thinking of moving the categories map to a separate file.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: In case of sections I think it is OK to have them hardcoded.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: Done.

igorpavlov: They are decoupled because the data structure inside sections are different. The General section has a hardcoded form. The Settings section has a form that would depend on the selected leave type. So I think `getSettingsTabFields` makes sense.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: Renamed to `getFieldsForActiveSettingsTab`. It searches for fields for the active **tab** in Settings section

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: This is easier for readability, I did consider indexes. If you need to change sections names - simply batch replace them.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: I did consider this, but UI-bootstrap does not at least provide "do not collapse if opened" feature as far as I know.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: I do not think we need that level of flexibility now. If we want to adapt the wizard for every future forms, we will need to compare and analyse all other planned wizards.

igorpavlov: Because I do not want to provide templates, I need to know which tab is opened to render the fields described in the map.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
AkA84: Given how experimental this wizard is and the possibility that the spec could still change, especially once you'll submit the first working version, i strongly suggest to not hardcode anything and make the system flexible. You're in the early stage of development, it will cost you less to do it now then to do it at the end

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
AkA84: ?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: is this something that was decided in agreement with the backend dev?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: I don't think that IDE's batch replacement is a good substitute for better code.

I don't see any problems with readability with either
```js
openSection(vm.sections[0]);
```
or
```js
openSection(_.first(vm.sections));
```

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: But the ui.bootstrap tabs support dynamic content out of the box, what's the technical limitation? Also, in your solution you are basically recreating/destroying the form of each tab any time get selected/deselected, i think that's something we should avoid performance-wise

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: the question is why the "General" section has an hardcoded form to begin with. You came up with a data structure and a business logic around it that supports dynamic rendering of forms, i think it should support the content of that section as well

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
AkA84: but you called it `openNextSettingsTab`, thus tightly coupling to a specific section. If there isn't a technical reason for why your component needs to rely on hardcoded references, then you need to remove those references

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-general.html
igorpavlov: Sorry, must have missed this one

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: I made it work with indexes.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: I attempted to do what you asked.

igorpavlov: UIBootstrap will render all fields at once and then just toggle the tab contents - that's right. On the other hand, imagine the wizard is large and contains a lot of tabs and fields, it is better to render on demand. The rendering in that case will take complexity O(fields_amount_per_tab) while in the case of tabs it will be O(all_fields_from_all_tabs).

As per inventing a wheel I will still need to toggle `active` class as described here in the example https://angular-ui.github.io/bootstrap/. I do not see a need of use UI bootstrap here, the solution is clean enough as per my opinion and it uses bootstrap markup.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: No, this was agreed during Wiki page reviewal.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: It looks too custom to me to render it from the map.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/form.core.js
```diff
@@ -2,9 +2,12 @@
 
 define([
   'common/angular',
+  'common/directives/help-text.directive',
   'leave-absences/leave-type-wizard/leave-type-wizard.core'
 ], function (angular) {
   angular.module('leave-type-wizard.form.core', [
```
AkA84: Nothing in particular with this line, just wanted to discuss your app structure that i noticed only now (apologies)

What is the need for a `leave-type-wizard.form` feature module? Can't we flatten the structure a bit and just have `leave-type-wizard`? I can't see the need for added complexity given that from all we now, the entire module _is_ the wizard form.

If then in the future for some reason this module grows in complexity, then we can create sub-features (see [LIFT](https://github.com/johnpapa/angular-styleguide/tree/master/a1#lift), but doing it now seems premature?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: But then your solution might be flawed, if it can't support "too" custom content. Isn't the idea of having individual templates for each field there exactly to support whatever each field is (a single text input, a group of checkboxes, some interactive element)?

I know that you haven't implemented the field templates yet, but it seems from your reply that you are not consider these radio buttons a candidate for your system

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
AkA84: > UIBootstrap will render all fields at once and then just toggle the tab contents - that's right. On the other hand, imagine the wizard is large and contains a lot of tabs and fields, it is better to render on demand. The rendering in that case will take complexity O(fields_amount_per_tab) while in the case of tabs it will be O(all_fields_from_all_tabs).

On hand we are talking about displaying a bunch of DOM nodes from the start, which is something that browsers can handle pretty well. On the other we are talking about doing some heavy DOM manipulation on each user interaction via angular, which we know from past experience is something we should avoid if we can

> As per inventing a wheel I will still need to toggle active class as described here in the example https://angular-ui.github.io/bootstrap/. I do not see a need of use UI bootstrap here, the solution is clean enough as per my opinion and it uses bootstrap markup.

I'm not questioning the elegance of your solution, but if we are paying the cost of including a library in our codebase, then we should try to use what that library is providing to us instead of writing a solution ourselves.

Unless there are technical reasons for why using the tabs directive provided by ui.bootstrap are limiting or hindering your solution (and there might be, for example with how the directive expects to write the markup), I don't see a reason for us not to use it. Please don't write code if it's not necessary

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: Not anymore.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/form-sections/leave-type-wizard-form-settings.html
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: @AkA84 Unfortunately, no. All fields **rows** in the Settings section have the same structure. The fields themselves are of course different, but the **table row** structure is the same for each field. That's why I created an autonomous template for rendering these fields from the map. But in case of the General section (one with the title field) - it's structure is not the same as per fields for Settings section does not match the content, so why not be a bit flexible here?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: No, I was talking about collapsing and already expanded section, but yeah, `is-disabled` may be the way to go. But the thing it - the amount of code would still be the same, you will still need this `ng-class={ active: whatever }`

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/form.core.js
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form-tabs.html
```diff
@@ -0,0 +1,50 @@
...
+          Next
+        </button>
+        <button class="btn btn-primary pull-right"
+          ng-if="form.isOnSectionLastTab">
+          Save
+        </button>
+        <button class="btn btn-secondary"
+          ng-if="form.isOnSectionFirstTab"
+          ng-click="form.openSection(sectionIndex - 1)">
+          Choose another category
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form-tabs.html
```diff
@@ -0,0 +1,50 @@
...
+          ng-click="form.openNextActiveSectionTab()">
+          Next
+        </button>
+        <button class="btn btn-primary pull-right"
+          ng-if="form.isOnSectionLastTab">
+          Save
+        </button>
+        <button class="btn btn-secondary"
+          ng-if="form.isOnSectionFirstTab"
+          ng-click="form.openSection(sectionIndex - 1)">
```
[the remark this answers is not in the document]
igorpavlov: Done.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form-tabs.html
```diff
@@ -0,0 +1,50 @@
...
+    <hr/>
+    <div class="row">
+      <div class="col-md-12">
+        <button class="btn btn-secondary pull-right"
+          ng-if="!form.isOnSectionLastTab"
+          ng-click="form.openNextActiveSectionTab()">
+          Next
+        </button>
+        <button class="btn btn-primary pull-right"
+          ng-if="form.isOnSectionLastTab">
```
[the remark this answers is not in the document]
igorpavlov: We don't care about actions yet, this would be handled in a different PR.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -37,9 +37,14 @@ public function browse() {
     $returnURL = CRM_Utils_System::url('civicrm/admin/leaveandabsences/types', 'reset=1');
     CRM_Utils_Weight::addOrder($rows, 'CRM_HRLeaveAndAbsences_DAO_AbsenceType', 'id', $returnURL);
 
-    CRM_Core_Resources::singleton()->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css');
-    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/crm-app-list.min.js', 1001);
-    CRM_Core_Resources::singleton()->addScriptFile('civicrm', 'js/jquery/jquery.crmEditable.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header');
+    CRM_Core_Resources::singleton()
```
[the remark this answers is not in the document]
igorpavlov: Will document.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/form/components/leave-type-wizard-form.component.js
igorpavlov: I attempted to do my best.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/leave-type-wizard.module.js
```diff
@@ -0,0 +1,15 @@
...
+  'leave-absences/leave-type-wizard/components/leave-type-wizard.component',
+  'common/directives/help-text.directive',
+  'leave-absences/leave-type-wizard/leave-type-wizard.core'
+], function (angular, LeaveTypeWizardComponent) {
+  angular.module('leave-type-wizard', [
+    'common.directives',
+    'common.services',
+    'leave-type-wizard.core'
+  ])
+    .component(LeaveTypeWizardComponent.__name, LeaveTypeWizardComponent);
```
AkA84: Just noticed this: with AngularJS 1.7 we don't need to use this custom `__name` property anymore. You can register the component by simply doing
```js
.component(LeaveTypeWizardComponent)
```
if `LeaveTypeWizardComponent` is an object where the key is the name of the component and the value is the component itself, please see [here](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge/components/tasks-notification-badge.component.js#L28-L36) and [here](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge/tasks-notification-badge.module.js#L15) for an example

Please note that this applies for anything in angular: services, directives, controllers, etc

PR #2866, opened 2018-09-19
PCHR-4221: Add Parser to translate menu configuration file to Civi format.

tunbola:
## Overview
The menu configuration file will replace the way menu items for the navigation menu is retrieved. Rather than use the items fetched from the `civicrm_navigation` table, the menu items will be supplied via a menu config file. This file will have the menu in a format that is easy to understand but will need to be translated to the format civi understands in order to add the necessary meta data needed to build the navigation. 
To achieve this, a new class `CRM_HRCore_Menu_Item` which defines what a civiHR menu item will be was created. A helper class `CRM_HRCore_Menu_MenuBuilder` was created to read the configuration file and translate that to a Menu Item Instance.
This Menu Item instance is then fed to the `CRM_HRCore_Menu_CiviAdapterr` parser class to translate it to the format that civi expects in order to generate the Navigation menu.

A sample configuration file for the menu in CiviHR will be like so:

```php
  $menuItems =  [
      'Staff' => [
        'icon'=> 'crm-i fa-users',
        'children' => [
          'New Individual' => [
            'permission' => 'administer CiviCRM',
            'url' => 'civicrm/new_individual',
            'separator' => 1
          ],
        ],
      ],
      'New Organization' => [
        'children' => [
          'New Life Insurance Provider' => 'civicrm/provider',
          'Health Provider' => [
            'children' => [
              'Custom Records' => 'civicrm/sample/bla',
              'Test Menu' => ['children' => CRM_HRCore_Menu_SystemSettings::getItems(),]
            ]
          ],
        ],
      ]
   ];
```

## Before
No functionality exists that will allow the navigation menu to be built from a menu configuration file. The menu Items reside in the database.
## After
Although the actual menu configuration file and the mode of injecting the menu items has not yet been implemented(Will be done in subsequent PRs), the functionality to allow the navigation menu to be built from a menu configuration file is now in place

## Comments
- A follow up PR will add the actual menu configuration file.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/CiviHr.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/CiviHr.php
@@ -1,0 +1,66 @@
+<?php
+
+use CRM_HRCore_Menu_Config as MenuConfig;
+use CRM_HRCore_Menu_Item as MenuItem;
+
+
+class CRM_HRCore_Helper_Menu_CiviHr {
+
+  /**
+   * This function reads the items from a menu config class
+   * and translates it to an array of CiviHR MenuItem objects.
+   * This array of objects can be given to any other function
+   * to translate to whatever format is needed in order to
+   * build the navigation menu.
+   *
+   * @param MenuConfig $menuConfig
+   *
+   * @return MenuItem[]
+   */
+  public function getMenuItems(MenuConfig $menuConfig) {
+   $items = $menuConfig->getItems();
+
+   return $this->generateMenuObjects($items);
+  }
+
+  /**
+   * Generates an array of menuItem objects from the
+   * menu config Items.
+   *
+   * @param array $items
+   *
+   * @return MenuItem[]
+   */
+  private function generateMenuObjects($items) {
+    $menuObjects = [];
+
+    foreach ($items as $label => $attributes) {
+      // If the attribute is not an array, it is the URL based on the
+      // Menu config format.
+      if (!is_array($attributes)) {
+        $attributes = ['url' => $attributes];
+      }
+
+      $menu = new MenuItem($label);
+      $menu->setUrl(CRM_Utils_Array::value('url', $attributes))
+           ->setPermission(CRM_Utils_Array::value('permission', $attributes))
+           ->setOperator(CRM_Utils_Array::value('operator', $attributes))
+           ->setIcon(CRM_Utils_Array::value('icon', $attributes));
+
+      if (!empty($attributes['separator'])) {
+        $menu->addSeparator();
+      }
+
+      if (!empty($attributes['children'])) {
+        $children = $this->generateMenuObjects($attributes['children']);
+        foreach($children as $child) {
+          $menu->addChild($child);
+        }
+      }
+
+      $menuObjects[] = $menu;
+    }
+
+    return $menuObjects;
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
@@ -1,0 +1,85 @@
+<?php
+
+use CRM_HRCore_Menu_Item as MenuItem;
+
+class CRM_HRCore_Helper_Menu_Parser {
+
+  /**
+   * Builds the navigation menu tree and returns in a format expected
+   * by Civi in order to generate the navigation menu. It accepts an
+   * array of MenuItem objects
+   *
+   * @param array $menuObjects
+   *   An array of MenuItem Objects
+   *
+   * @return array
+   *  Sample return format:
+   *  [
+   *    'attributes' =>
+   *      [
+   *        'label' => 'Components',
+   *         'url' => 'civicrm/bla',
+   *         'permission' => 'sample permission',
+   *         'operator' => 'AND',
+   *      ],
+   *   ],
+   */
+  public static function getNavigationTree(array $menuObjects) {
+    $navID =1;
+    return self::buildMenuTree($menuObjects, 0, $navID);
+  }
+
+  /**
+   * Accepts an array of MenuItem Objects to build the menu tree, alongside
+   * the parent and navigation Id required by civi to build the navigation
+   * menu
+   *
+   * @param array $items
+   *   An array of MenuItem Objects.
+   * @param int $parentID
+   * @param int $navID
+   *
+   * @return array
+   */
+  private static function buildMenuTree(array $items, $parentID = 0, &$navID) {
+    $weight = 1;
+    $navigationTree = [];
+
+    foreach ($items as $menuObject) {
+      if(!$menuObject instanceof MenuItem) {
+        throw new RuntimeException('Menu Item should be an instance of '.  MenuItem::class);
+      }
+
+      $attributes = [
+        'label' => $menuObject->getLabel(),
+        'name' => $menuObject->getLabel(),
+        'url' => $menuObject->getUrl(),
+        'icon' => $menuObject->getIcon(),
+        'weight' => $weight,
+        'permission' => $menuObject->getPermission(),
+        'operator' => $menuObject->getOperator(),
+        // Civi understands 1 to be the value for separator after menu item.
+        'separator' => $menuObject->hasSeparator() ? 1 : NULL,
+        'parentID' => $parentID,
+        'navID' => $navID,
+        'active' => 1,
+      ];
+
+      $menuItem = ['attributes' => $attributes];
+      $hasChildren = !empty($menuObject->getChildren());
+
+      if ($hasChildren) {
+        $navID++;
+        $menuItem['child'] = self::buildMenuTree($menuObject->getChildren(), $attributes['navID'], $navID);
+      }
+
+      $navigationTree[] = $menuItem;
+      if (!$hasChildren) {
+        $navID++;
+      }
+      $weight++;
+    }
+
+    return $navigationTree;
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
@@ -73,7 +73,7 @@ class CRM_HRCore_Menu_CiviAdapter {
 
       if ($hasChildren) {
         $navID++;
-        $menuItem['child'] = self::buildMenuTree($menuObject->getChildren(), $attributes['navID'], $navID);
+        $menuItem['child'] = $this->buildNavigationTree($menuObject->getChildren(), $attributes['navID'], $navID);
       }
 
       $navigationTree[] = $menuItem;
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
@@ -28,7 +28,7 @@ class CRM_HRCore_Menu_CiviAdapter {
     //The parentId is zero here to indicate that the top level navigation items
     //has no parents. Also the navID is incremented by 1 for each menu item generated
     //in the returning array, it starts at 1 and is passed by reference here.
-    return self::buildMenuTree($menuItem->getChildren(), 0, $navID);
+    return $this->buildNavigationTree($menuItem->getChildren(), 0, $navID);
   }
 
   /**
@@ -48,7 +48,7 @@ class CRM_HRCore_Menu_CiviAdapter {
    *
    * @return array
    */
-  private function buildMenuTree(array $items, $parentID = 0, &$navID) {
+  private function buildNavigationTree(array $items, $parentID = 0, &$navID) {
     $weight = 1;
     $navigationTree = [];
 
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
@@ -2,7 +2,7 @@
 
 use CRM_HRCore_Menu_Item as MenuItem;
 
-class CRM_HRCore_Helper_Menu_Parser {
+class CRM_HRCore_Menu_CiviAdapter {
 
   /**
    * Builds the navigation menu tree and returns in a format expected
@@ -25,7 +25,7 @@ class CRM_HRCore_Helper_Menu_Parser {
    *   ],
    */
   public static function getNavigationTree(array $menuObjects) {
-    $navID =1;
+    $navID = 1;
     return self::buildMenuTree($menuObjects, 0, $navID);
   }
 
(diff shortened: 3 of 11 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Builds the navigation menu tree for the menu items for the given parent ID.
+   *
+   * @param array $items
+   * @param int $parentID
+   *
+   * @return array
+   */
+  private static function buildMenuTree(array $items, $parentID = 0) {
+    STATIC $navID = 1;
```
mickadoo: How about using an extra argument in the `buildMenuTree` to accept current highest ID? 

I'm just saying that because if you (for some reason) had two calls to `getNavigationTree` then the $navID would not be reset. See [SO](https://stackoverflow.com/a/27879260/1196369)

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,75 @@
...
+    $weight = 1;
+    $navigationTree = [];
+
+    foreach ($items as $label => $attributes) {
+      if (!is_array($attributes)) {
```
mickadoo: Maybe a comment explaining this would be helpful

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,75 @@
...
+
+      $attributes['name'] = $label;
+      $attributes['label'] = $label;
+      $attributes['navID'] = $navID;
+      $attributes['parentID'] = $parentID;
+      $attributes['weight'] = $weight;
+      $attributes['active'] = 1;
+      $menuItem = ['attributes' => $attributes];
+
+      $hasChildren = isset($menuItem['attributes']['children']);
```
mickadoo: Would `empty` be clearer here? I'm just thinking if it has `$menuItem['attributes']['children'] = NULL`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,75 @@
...
+   *    'attributes' =>
+   *      [
+   *        'label' => 'Components',
+   *         'url' => 'civicrm/bla',
+   *         'permission' => 'sample permission',
+   *         'operator' => 'AND',
+   *      ],
+   *   ],
+   */
+  public static function getNavigationTree(Navigation $navigationMenu) {
```
mickadoo: How about avoiding static methods and making this a container service? I know it doesn't have external calls, but it might in the future. What do you think?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/Navigation.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+interface CRM_HRCore_Menu_Navigation {
```
mickadoo: Will the interface ever possibly have more than one concrete class?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Helper/Menu/ParserTest.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+use CRM_HRCore_Test_BaseHeadlessTest as BaseHeadlessTest;
+use CRM_HRCore_Menu_Navigation as Navigation;
+use CRM_HRCore_Helper_Menu_Parser as MenuParser;
+
+/**
+ * Class CRM_HRCore_Helper_Menu_ParserTest
```
mickadoo: Don't think we usually use the `Class` docblock

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Helper/Menu/ParserTest.php
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @group headless
+ */
+class CRM_HRCore_Helper_Menu_ParserTest extends BaseHeadlessTest {
+
+  public function testGetNavigationTreeReturnsMenuItemsInExpectedFormat() {
+    $navigation = $this->prophesize(Navigation::class);
```
mickadoo: With tests that deal with large data sets like this I sometimes find it helpful to put the data in separate files, for example a PHP file that returns an array, or a JSON file that you can use `json_decode` to get the PHP array from

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/Navigation.php
tunbola: Don't think so. Its not necessary right?

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Helper/Menu/ParserTest.php
tunbola: Good idea, just that since the data set is not too large and besides there's the advantage of viewing the whole test at once an understanding what is going on without opening other files.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
tunbola: Good idea.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
tunbola: Good idea.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
tunbola: Good point. Will look into that.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/Navigation.php
mickadoo: Well I'm not sure about how implementation will be in the future. Maybe you'll have some instance of it for sub menus or something. I'm fine with having an interface, just was curious

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Helper/Menu/ParserTest.php
mickadoo: Sure, for now it's fine. I'm just thinking if you have a few different test cases in the future and the test file ends up being hundreds of lines long

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/CiviHr.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+use CRM_HRCore_Menu_Config as MenuConfig;
+use CRM_HRCore_Menu_Item as MenuItem;
+
+
+class CRM_HRCore_Helper_Menu_CiviHr {
```
davialexandre: I think that "helper" is a bit too generic here and might end up passing the wrong impression that we can add anything menu-related to this class. The first two ideas that come to my mind are `MenuBuilder` or `MenuFactory`. I slightly prefer the former, as `MenuFactory` could confuse people thinking that this would be an implementation of the Factory pattern.

Just to be clear, I mean that I think that the class name should be something like `CRM_HRCore_Menu_MenuBuilder` or  `CRM_HRCore_Menu_MenuFactory`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+use CRM_HRCore_Menu_Item as MenuItem;
+
+class CRM_HRCore_Helper_Menu_Parser {
```
davialexandre: What do you think of renaming this to something like `CRM_HRCore_Menu_CiviAdapter`, as this is actually adapting the format from CiviHR to CiviCRM?

By the book, the Adapter pattern is used to adapt the interface of a class to another class, which is not exactly what were doing here, since what we're doing here is adapting things from one system to another, I think the concept can still be used.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/Item.php
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * @var CRM_HRCore_Menu_Item[]
+   */
+  protected $children = [];
+
+  /**
+   * @var CRM_HRCore_Menu_Item
+   */
+  protected $parent = NULL;
```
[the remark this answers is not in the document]
tunbola: tbh, i don't think the parents are necessary. I'll remove that. I only thought its proper if you have a menu item instance, it should be able to supply info as regards its parent and its children, but i've not found a use case for that in code.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Helper/Menu/Parser.php
```diff
@@ -0,0 +1,85 @@
...
+   *    'attributes' =>
+   *      [
+   *        'label' => 'Components',
+   *         'url' => 'civicrm/bla',
+   *         'permission' => 'sample permission',
+   *         'operator' => 'AND',
+   *      ],
+   *   ],
+   */
+  public static function getNavigationTree(array $menuObjects) {
```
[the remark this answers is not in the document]
tunbola: I thought this method might just be used in probably one place and its more of an helper method and there won't be need to mock it in tests, thats why I made it static.

The reason why its not a single menu instance to hold the entire menu instance is because, the menu items does not have a single parent, The top level navigation items Home, Staff, Leave etc are represented as menu item objects that contain their children within them since these top level menu items themselves have no parent, that is why they are supplied as an an array of parent (top level navigation menus) menu items.
What do you think?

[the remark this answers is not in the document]
davialexandre: I think that, in general, we should start adopting the practice of always using instance methods by default and only use static one in situations where it is really necessary and/or make sense for the implementation. I want to see us applying more OO techniques in our codebase and using static methods kind goes against that.

The idea of having a class representing the menu goes along with this goal of making the codebase more OO. There are multiple ways of representing that. One way is as KnpMenu does, which is using an "blank" menu item instance which will only be used to keep a list of children (Home, Staff, Leave, etc) and each child can have its own list of children. Another way is to do something we have in this [laravel-menu library](https://github.com/lavary/laravel-menu), where there's a Menu class, which represents the entire menu, and an Item class, which represents the items in these menus.

The bottom line is, we need to represent the Menu in an OO format and we cannot do that with an array, because arrays are not objects and can contain any type of values.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
```diff
@@ -0,0 +1,88 @@
...
+   *         'operator' => 'AND',
+   *      ],
+   *   ],
+   */
+  public function getNavigationTree(MenuItem $menuItem) {
+    $navID = 1;
+    //The parentId is zero here to indicate that the top level navigation items
+    //has no parents. Also the navID is incremented by 1 for each menu item generated
+    //in the returning array, it starts at 1 and is passed by reference here.
+    return self::buildMenuTree($menuItem->getChildren(), 0, $navID);
```
davialexandre: now that this is not a static method anymore, we need to use `$this` instead of `self`

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
```diff
@@ -0,0 +1,88 @@
...
+   *      [
+   *        'label' => 'Components',
+   *         'url' => 'civicrm/bla',
+   *         'permission' => 'sample permission',
+   *         'operator' => 'AND',
+   *      ],
+   *   ],
+   */
+  public function getNavigationTree(MenuItem $menuItem) {
+    $navID = 1;
```
davialexandre: Would there be a difference if it was 0 instead of 1? How about setting this as the default value of `buildMenuTree()`, the same way that the default value for parent id is 0?

Also, I've just realized that you're using two different names for the same thing. Here you call it "Navigation Tree" but on the other method you call it "Menu Tree". Lets use the same name for both

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Menu/CiviAdapterTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+  private function getExpectedMenuItems() {
+    // The weight is expected to be in incremental order (+1) for each navigation level
+    // starting at 1.
+    // Top level navigation menu items are expected to have a parentID of 0 while inner level
+    // navigation items/children will have their parentID to be equal to the navID of their
+    // parent.
+    // The navID increases sequentially for each menu item by 1, starting from the first
+    // top level menu down to its children, and then on to the next top level menu down to its
+    // children and so on.
```
davialexandre: Please move this to the docblock of this method

uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Menu/MenuBuilderTest.php
```diff
@@ -0,0 +1,169 @@
...
+    $url = 'civcrm/staff';
+    $menuConfigItem = [
+      'Staff' => $url
+    ];
+
+    $menuConfig = $this->prophesize(MenuConfig::class);
+    $menuConfig->getItems()->willReturn($menuConfigItem);
+    $menuBuilder = new MenuBuilder();
+
+    $menuObject = $menuBuilder->getMenuItems($menuConfig->reveal());
```
davialexandre: These 4 lines of code (from 34 to 38) are repeated in all the tests here. I suggest we avoid that by doing this:

- Create a property in this class to store the MenuBuilder instance
- Instantiate the MenuBuilder in the setUp method
- Create a helper method to create a MenuConfig mock that returns a given menu configuration

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/CiviAdapter.php
tunbola: If the navID was set to Zero, Top level menu items having nav ID as zero will have their direct children having the parentID as 0, same as their parent (top level item also have parentId as 0) which will be conflicting.
The reason why I did this way is to mimic Civi that sets the parent ID of top navigation items to NULL but each navigation Item has its own unique ID.

PR #2867, opened 2018-09-26
PCHR-4222: Inject Menu Items Via Navigation Menu Hook

tunbola:
## Overview
in #2866, the menu config file blue print was defined, also the Parser class that translates Menu Item objects to the format civi understands was added.
In this PR, the actual navigation menu config file that contains the menu items is added. These menu items are a reflection of menu items as currently seen on staging. 
Also since the way to add menu items will now be different, we no longer need upgrader code that add/modify menu items in the database, so these code was removed.

## Before
- The navigation menu items are supplied via the civicrm_navigation table and modifications are made to these items via the hook hook_civicrm_navigationMenu by various extensions

## After
- The navigation menu items are now supplied via a menu config file `CRM_HRCore_Menu_MainMenuConfig`.

## Technical Details
- The `CRM_HRCore_Menu_CiviAdapter` class after parsing the Menu Item objects returns the navigation in a format that civi will use to build the navigation menu. This items are then used to swap with the menu items civi pulls from the `civicrm_navigation_table` via the HRCore's implementation of the `hook_civicrm_navigationMenu`
```php
function hrcore_civicrm_navigationMenu(&$params) {
  $mainMenuConfig = new CRM_HRCore_Menu_MainMenuConfig();
  $menuBuilder = new CRM_HRCore_Menu_MenuBuilder();
  $civiAdapter = new CRM_HRCore_Menu_CiviAdapter();

  $menuObject = $menuBuilder->getMenuItems($mainMenuConfig);
  $params = $civiAdapter->getNavigationTree($menuObject);
}
```
- All other extension's implementation of the `hook_civicrm_navigationMenu` were removed, so only the HRCore extension implements the hook.
- Dynamic menu Items such as the menu Items for custom fields were supplied via a class that encapsulates the logic for getting the menu items and returns the menu config items.

## Manual Testing
The implementation was tested on an existing site and also on a new site created with civibuild. In both instances the menu items were compared with the one seen on staging for various roles.


uk.co.compucorp.civicrm.hrcore/hrcore.php
```diff
@@ -324,26 +324,28 @@ function hrcore_civicrm_pageRun($page) {
 }
 
 /**
- * Implements hook_civicrm_coreResourceList().
+ * The implementation of this hook is to ensure that the menu navigation
+ * Items supplied by civiHR overrides the one civi pulls from the database.
+ * The menu items from the civiHR menu config file replaces the navigation
+ * menu items civi stores in the db.
+ *
```
davialexandre: s/civi/Civi/g

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/MainMenuConfig.php
```diff
@@ -0,0 +1,179 @@
...
+          'icon' => 'crm-i fa-users',
+          'children' =>
+            [
+              'New Individual' =>
+                [
+                  'url' => 'civicrm/contact/add?reset=1&ct=Individual',
+                  'permission' => 'add contacts',
+                ],
+              'New Organization' =>
+                [
```
davialexandre: I think we can and we should reduce the number of lines in these files, also, maybe, add some empty lines to improve readability. For example, by simply not putting the opening `[` on a new line will reduce the number of lines in these array by A LOT. For example, check this comparison between a small fraction of the file: https://gist.github.com/davialexandre/539444362f785ed3127df08707de4c51

As you can see, the compact version has 9 fewer lines than the original version, even though it includes empty lines between some items to improve the readability.

Could you please ask the other devs what they think about this?

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/MainMenuConfig.php
```diff
@@ -0,0 +1,179 @@
+<?php
+
+class CRM_HRCore_Menu_MainMenuConfig implements CRM_HRCore_Menu_Config {
```
davialexandre: I have to see I didn't see the idea of every menu being a class coming. I have mentioned a configuration file multiple times, so this is what I was expecting. When I saw the `CRM_HRCore_Menu_Config` interface, I thought it would be implemented by some class that would read this configuration file rather than the holding the configuration itself. Could you give me more details about why you decided to implement it this way?

I do see the advantage of having a class representing SOME of the menus, like the Syleguide one, which is dynamic and has some logic in involved, but for most of the cases, a basic php file containing an array should have been enough

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/MainMenuConfig.php
tunbola: I agree we can improve readability by adding empty lines in between and not putting the opening `[` on a new line. I'll do that.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/MainMenuConfig.php
tunbola: tbh, the classes are not necessary for most of the menu config.
Agree basic php file will do for most of the menus and we don't really need the classes for most of them. Will make the non dynamic ones to be a php file returning an array rather than being classes.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Menu/Config/Configure.php
```diff
@@ -0,0 +1,303 @@
+<?php
+
+return [
```
davialexandre: If I was a dev who just started working on the CiviHR project and my first task was to change something in the menu configuration file I would never think of trying to find this file inside the sources folder. The first thing I would probably do is to look for some "config" folder in the root of the application (the hrcore root folder in this case), as this is the common approach in most systems and frameworks. 

Since we have several files defining the menu, I would suggest placing them all in `hrcore/config/menu`. Also, since these are not classes anymore, there's no need to capitalize the file name.

PR #2868, opened 2018-09-27
PCHR-4245: Fix menu Item Link Target

tunbola:
## Overview
The menu config files has the target attribute for a menu item but when converting to menu object, this is not taken in to account , so some of the links that are supposed to open in a new window are opening in the current window. This PR fixes the issue by adding the target property to the Menu Item object and taking this into account when translating the Menu Items to Civi format using the Civi Adapter class

## Before
- The target attribute for a menu item is not taken into account when translating the Menu Items to Civi format using the Civi Adapter class

## After
- The target attribute for a menu item is now taken into account when translating the Menu Items to Civi format using the Civi Adapter class


ajesamson:
A little typo in the overview `After` section.
```
The target attribute for a menu item is now taken into account when translating the Menu Items to Civi format using the Civi Adapter class
```

PR #2870, opened 2018-09-28
PCHR-4213: Add Category Field to Leave Type

ajesamson:
## Overview
As part of the requirement for simplifying creation of new leave types, this PR creates a category column in the `civicrm_hrleaveandabsences_absence_type` table for grouping leave types created. This is also coupled with option value setup to hold the names.

## Before
Leave category column does not exist in the absence type table.

## After
Leave category column now exist in the absence type table.

## Technical Details
Apart from adding a new category column, the insert query for new site setup was also modified. This is because category field cannot be `null` when creating new leave type.
The Data Access Object (DAO) was updated by running `build-dao.sh ` on leave and absence extension.
Option group and value were also setup to hold the available categories, making it flexible to add new category in the future.

---
## Comment
Some modifications were made on DAO files for `HRLeaveandAbsence`. This is as result of updating the entity using `build-dao.sh`


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/DAO/AbsenceType.php
```diff
@@ -479,6 +484,19 @@ static function &fields() {
           'bao' => 'CRM_HRLeaveAndAbsences_DAO_AbsenceType',
           'localizable' => 0,
         ) ,
+        'category' => array(
```
tunbola: Were this changes manually generated or done by the Build DAO script? I was expecting to see a change in the GenCodeChecksum at the top of the file.
The instructions are here: https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/101089347/Creating+and+Updating+entities+BAO+DAO+for+an+extension

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE `civicrm_hrleaveandabsences_absence_type` (
      `is_sick` tinyint   DEFAULT 0 COMMENT 'A flag which is used to determine if this Absence Type can be used for a Sickness Request',
      `calculation_unit` varchar(512) NOT NULL   COMMENT 'One of the values of the Absence type calculation units option group',
      `hide_label` tinyint   DEFAULT 0 COMMENT 'This controls the visibility of the Leave Type label in the calendar and feeds.',
+      `category` varchar(10) NOT NULL COMMENT 'This is used for grouping leave types',
```
tunbola: The indenting here doesn't seem right.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -134,7 +140,8 @@ INSERT INTO `civicrm_hrleaveandabsences_absence_type`(
   1,
   3,
   1,
-  1
+  1,
+  'Sickness'
```
tunbola: I was thinking there would be a new leave type category table that would then be a foreign key in the absence type table. This means that these categories, Leave, Sickness and TOIL will be system defined. Can they be modified by the admin via the UI? 
Also thinking of some properties we have on the absence type table like is_sick, is_toil that is used to identify what type if leave type it is. Wondering how we can merge this and probably do away with the ones we don't use so we don't keep adding columns to the table.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1034.php
```diff
@@ -0,0 +1,26 @@
...
+   * @return bool
+   */
+  public function upgrade_1034() {
+    $absenceTypeTable = AbsenceType::getTableName();
+
+    if (!SchemaHandler::checkIfFieldExists($absenceTypeTable, 'category')) {
+      CRM_Core_DAO::executeQuery("
+        ALTER TABLE {$absenceTypeTable}
+        ADD category varchar(10) NOT NULL 
+        COMMENT 'This is used for grouping leave types.'
```
tunbola: We can probably move this up to the line above, it will still be readable.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -31,7 +31,7 @@ CREATE TABLE `civicrm_hrleaveandabsences_absence_type` (
      `is_sick` tinyint   DEFAULT 0 COMMENT 'A flag which is used to determine if this Absence Type can be used for a Sickness Request',
      `calculation_unit` varchar(512) NOT NULL   COMMENT 'One of the values of the Absence type calculation units option group',
      `hide_label` tinyint   DEFAULT 0 COMMENT 'This controls the visibility of the Leave Type label in the calendar and feeds.',
-      `category` varchar(10) NOT NULL COMMENT 'This is used for grouping leave types',
+     `category` int unsigned NOT NULL   COMMENT 'This is used for grouping leave types.',
```
tunbola: The SQL statement still reads NOT NULL whereas the xml file says the field is not required.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1034.php
```diff
@@ -7,20 +7,54 @@
...
+        ADD category int unsigned NOT NULL COMMENT 'This is used for grouping leave types.'
       ");
     }
+  }
 
-    return TRUE;
+  /**
+   * Ensures absence type category option group and its values exist, creating one if not
+   */
+  private function up1034_createCategoryOptionValues() {
```
tunbola: The option group install file can be used here as well to install for existing sites.
See here: https://github.com/compucorp/civihr/blob/667735780c972d7e2a91d71eb570c91e03567aca/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1009.php#L22

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1034.php
```diff
@@ -7,20 +7,54 @@
...
+   */
+  private function up1034_createCategoryColumn() {
     $absenceTypeTable = AbsenceType::getTableName();
 
     if (!SchemaHandler::checkIfFieldExists($absenceTypeTable, 'category')) {
       CRM_Core_DAO::executeQuery("
         ALTER TABLE {$absenceTypeTable}
-        ADD category varchar(10) NOT NULL 
-        COMMENT 'This is used for grouping leave types.'
+        ADD category int unsigned NOT NULL COMMENT 'This is used for grouping leave types.'
```
tunbola: Since the category field is required, it is possible that tests that were previously creating/fabricating absence types will now throw errors because of this new field. This might require modifying the fabricators to add a default value for this field.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
ajesamson: Kindly check again. The [required](https://github.com/compucorp/civihr/pull/2870/commits/9bd60e35945379187af0532fb5ef76336180507a#diff-8ff8f4f4dbde96aab412eb3d1754ce62L186) reads true.

uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/AbsenceType.xml
```diff
@@ -181,9 +181,13 @@
   <field>
     <name>category</name>
     <title>Category</title>
-    <type>int unsigned</type>
+    <type>varchar</type>
+    <length>10</length>
```
tunbola: Length should be 512 also to match the length for the Value column on the civicrm_option_value table.

uk.co.compucorp.civicrm.hrleaveandabsences/sql/auto_install.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE `civicrm_hrleaveandabsences_absence_type` (
      `is_sick` tinyint   DEFAULT 0 COMMENT 'A flag which is used to determine if this Absence Type can be used for a Sickness Request',
      `calculation_unit` varchar(512) NOT NULL   COMMENT 'One of the values of the Absence type calculation units option group',
      `hide_label` tinyint   DEFAULT 0 COMMENT 'This controls the visibility of the Leave Type label in the calendar and feeds.',
+     `category` varchar(10) NOT NULL   COMMENT 'This is used for grouping leave types.',
```
tunbola: The upgrader SQL needs to be updated also, it still has the int data type.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Test/Fabricator/AbsenceType.php
```diff
@@ -20,6 +20,10 @@ public static function fabricate($params = []) {
       $params['title'] = static::nextSequentialTitle();
     }
 
+    if (empty($params['category'])) {
```
tunbola: When you try to create an Absence Type via the UI, this might also result in an error since the category field is now required. What would be the default for new Absence Types, Custom? 
We should make sure that the UI is not broken and also add a test to verify which category a new absence type will have. Also is there a plan to introduce this field in the Absence type creation UI?

tunbola: Shouldn't the validation to ensure that the category sent is part of the available category option values be added in this ticket? Or is there a separate task for that?

ajesamson: There is a separate [ticket](https://compucorp.atlassian.net/browse/PCHR-4225) to handle all validation per category.

PR #2872, opened 2018-10-01
PCHR-4193: Fix Issue With CiviTask and CiviDocument Components Not Enabled On New Sites

tunbola:
## Overview
Even though in the Task and Assignments install function [here](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php#L20-L24), the CiviTask and CiviDocuments are to be enabled for new sites, however these components are not enabled when a new site is created.
Issue was traced to the HRUI extension [here](https://github.com/compucorp/civihr/blob/ca1854d9030af597f0d29eea9b5fe33f0a8f9c6b/hrui/hrui.php#L228-L232), the extension replaces the value of the enable_component setting completely and because the HRUI extension is installed after T&A in the `civihr/bin/drush-install.sh` file, the settings set by HRUI overrides the one in T&A.

The core purpose of the HRUI extension setting the `CiviReport and CiviCase` core components active is to make sure that unwanted components such as CiviGrant, CiviPledge are not set active as well, however this does not take into account that another extension might have added its own components to the enabled components list.

## Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

## After
The HRUI extension now fetches the enabled core components list that Civi sets by default and uses that list to filter out unwanted core components that are not to be enabled while also not overriding the component list incase other extensions had modified it.

## Comments
I could not get an API to get the default core components list, the snipped below returns an empty array result set.
```php
$result = civicrm_api3('Setting', 'getdefaults', [
  'name' => "enable_components",
]];
```
The `Civi::settings()->getDefault('enable_components')` function was used to retrieve this list.


hrui/hrui.php
```diff
@@ -224,12 +224,23 @@ function hrui_civicrm_install() {
...
+  $getResult = civicrm_api3('setting', 'getsingle', [
     'domain_id' => CRM_Core_Config::domainID(),
-    'enable_components' => array('CiviReport','CiviCase'),
-  );
-  $result = civicrm_api3('setting', 'create', $params);
+    'return' => ['enable_components'],
+  ]);
+
+  $enabledComponents = $getResult['enable_components'];
+  $enableComponents = array_diff($enabledComponents, $unwantedCoreComponents);
```
davialexandre: I don't understand why `hrui` is enabling/disabling components, as this extension is only supposed to have civicrm core UI customizations (e.g, hiding fields, labels, etc).

Another confusing thing is that it seems we have code that handles enabling components in at least three different parts of the system (hrui, [hrcase](https://github.com/compucorp/civihr/blob/513f9b86955132a8ba03e670ae1e61797ce4bcb4/hrcase/CRM/HRCase/Upgrader.php#L18-L20) and [t&a](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php#L20-L23)), which I think is the biggest reason we had all this confusion in the first place.

What I think we should do is:
- HRCore disables all the components
- T&A enables CiviCase, CiviTask and CiviDocument (We need to make sure CiviCase is not used by any other extension that gets installed before T&A)
- Check which extension actually uses CiviReport (`hrreport`?) and enable it there.

So, basically, hrcore will be responsible for only disabling components (which is different from hrui that is currently responsible for enabling and disabling them), and extensions are responsible for only enabling the ones they need (without disabling previously enabled ones).

Another option would be to move all the enabling/disabling logic to hrcore so that we have this in one single place. My only concern about this is that the CiviTask and CiviDocument componets are defined by T&A and I don't know if it would be possible to enabled these components before T&A is installed.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -85,9 +86,72 @@ public function install() {
...
+   */
+  public function enableComponents() {
+    $this->setComponentsStatus(TRUE, FALSE);
+  }
+
+  /**
+   * Disables required components.
+   */
+  public function disableComponents() {
+    $this->setComponentsStatus(FALSE, FALSE);
```
davialexandre: Why did the signature of this method change? Compared to the previous version we had in `hrcase`, this new one is worse in terms of readability. Previously, we would have something like `setComponentStatus(['CiviCase' => true, 'CiviTask' => false])` and I could easily understand that this would enable CiviCase and disable CiviTask. Now, I cannot understand what is going on unless I actually read the method implementation, because these booleans params simply don't convey any information

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -85,9 +86,72 @@ public function install() {
     $this->deleteUnneededCustomGroups();
     $this->createDefaultRelationshipTypes();
     $this->makeAllCurrenciesAvailable();
+    $this->enableComponentsAndFilterDefaultComponents();
```
davialexandre: The end in the method name indicates it is doing two things. A method should have only one single responsibility

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -85,9 +86,72 @@ public function install() {
...
+   * Enables required components.
+   */
+  public function enableComponents() {
+    $this->setComponentsStatus(TRUE, FALSE);
+  }
+
+  /**
+   * Disables required components.
+   */
+  public function disableComponents() {
```
davialexandre: If this method disables only required components, the method name should say that. Otherwise, it means more like it will disable ALL the components.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -85,9 +86,72 @@ public function install() {
...
+   * from the default components list.
+   */
+  public function enableComponentsAndFilterDefaultComponents() {
+    $this->setComponentsStatus();
+  }
+
+  /**
+   * Enables required components.
+   */
+  public function enableComponents() {
```
davialexandre: Same thing here. If the method only enables the required components, the method name should say that, otherwise it means all

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
```diff
@@ -85,9 +86,72 @@ public function install() {
...
+  /**
+   * This function enables/disables the required components.
+   *
+   * @param bool $status
+   * @param bool $filterCoreComponents
+   *   Whether to remove unwanted core Civi components
+   *   or not from the enabled components list.
+   */
+  private function setComponentsStatus($status = TRUE, $filterCoreComponents = TRUE) {
+    $coreComponents = ['CiviReport','CiviCase'];
```
davialexandre: Missing space after comma

PR #2874, opened 2018-10-02
TMP: Showcasing alternate Civi Upgrade method

mickadoo:
## Overview

DO NOT MERGE THIS PR!

This is a temporary PR to show how we could use separate classes for CiviCRM upgrade steps, and allow discussion on the approach

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Step/Step1104.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Step/Step1104.php
@@ -1,0 +1,7 @@
+<?php
+
+class CRM_HRCore_Upgrader_Step_Step1104 {
+  public function apply() {
+    return true;
+  }
+}
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader.php
@@ -81,6 +81,90 @@ class CRM_HRCore_Upgrader extends CRM_HRCore_Upgrader_Base {
   private $xmlDirectories = ['CustomGroups'];
 
   /**
+   * Determine if there are any pending revisions.
+   *
+   * @return bool
+   */
+  public function hasPendingRevisions() {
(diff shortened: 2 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Step/Step1104.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+class CRM_HRCore_Upgrader_Step_Step1104 {
+  public function apply() {
+    return true;
```
mickadoo: @davialexandre it might be possible to just return `true` from the function that calls this, and avoid requiring a return value here. Maybe I'm missing something, but I don't see the point of it, it's like a very old style of PHP which uses return values to indicate errors instead of throwing exceptions.

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Step/Step1104.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+class CRM_HRCore_Upgrader_Step_Step1104 {
```
mickadoo: @davialexandre do you think it would be good to have an `UpgraderStepInterface` just to help the IDE with auto-completion?

davialexandre: yup

uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Upgrader/Step/Step1104.php
davialexandre: I'm not sure I understand. Are you saying that we can simply not return anything and it will just work or that we need to change something so that it will fail only if an exception is thrown?

mickadoo: > that we need to change something so that it will fail only if an exception is thrown?

This one. I don't know about you, but I've never returned false from an upgrader anyway.

PR #2878, opened 2018-10-08
PCHR-4207: Add leave type category label

ajesamson:
## Overview
A new field called category was introduced in a previous PR. This PR adds the the category field to leave/absence type listing page.

## Before
before_category

## After
after_category

## Technical Details
The categories were fetched using BAO for AbsenceType and mapped to each of the category id in absence type table.
```
private function getCategoryLabel($categoryId) {
  if (empty($this->categories)) {
    $this->categories = AbsenceType::getCategories();
  }

  $categoryLabel = array_search($categoryId, $this->categories);
  if ($categoryLabel !== NULL) {
    return $categoryLabel;
  }
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
@@ -213,6 +213,15 @@ class CRM_HRLeaveAndAbsences_BAO_AbsenceType extends CRM_HRLeaveAndAbsences_DAO_
   }
 
   /**
+   * Returns a list of all possible categories for absence type
+   *
+   * @return array|bool
+   */
+  public static function getCategories() {
+    return self::buildOptions('category');
+  }
+
+  /**
    * Unset the is_default flag for every AbsenceType that has it
    */
   private static function unsetDefaultTypes() {
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
@@ -1,10 +1,12 @@
 <?php
 
 use CRM_HRLeaveAndAbsences_Service_AbsenceType as AbsenceTypeService;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 
 class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
 
-  private $links = array();
+  private $links = [];
+  private $categories = [];
 
   /**
    * @var \CRM_HRLeaveAndAbsences_Service_AbsenceType
@@ -32,6 +34,7 @@ class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
           $this->calculateLinksMask($object),
           ['id' => $object->id]
       );
+      $rows[$object->id]['category'] = $this->getCategoryLabel($rows[$object->id]['category']);
     }
 
     $returnURL = CRM_Utils_System::url('civicrm/admin/leaveandabsences/types', 'reset=1');
@@ -182,5 +185,22 @@ class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
   private function canNotDelete($absenceType) {
     return $this->absenceTypeService->absenceTypeHasEverBeenUsed($absenceType->id)
            || $absenceType->is_reserved;
+  }
+
+  /**
+   * Retrieves the category label from option value
+   *
+   * @param int $categoryId
+   *
+   * @return mixed
+   */
+  private function getCategoryLabel($categoryId) {
+    if (empty($this->categories)) {
+      $this->categories = AbsenceType::getCategories();
+    }
+
+    if (array_key_exists($categoryId, $this->categories)) {
+      return $this->categories[$categoryId];
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -212,6 +212,15 @@ public static function getValuesArray($id) {
     return $absenceType;
   }
 
+  /**
+   * Returns a list of all possible categories for absence type
+   *
+   * @return array|bool
+   */
+  public static function getCategories() {
+    return array_flip(self::buildOptions('category'));
```
tunbola: Do we really need to flip this? Without flipping it returns an array of category labels indexed by the ID, which seems a more natural data set we would want this function to return.

tunbola: Also we need to add a test for this function.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -183,4 +186,22 @@ private function canNotDelete($absenceType) {
     return $this->absenceTypeService->absenceTypeHasEverBeenUsed($absenceType->id)
            || $absenceType->is_reserved;
   }
+
+  /**
+   * Retrieves the category label from option value
+   *
+   * @param $categoryId
```
tunbola: Missing param type

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
ajesamson: > Do we really need to flip this? Without flipping it returns an array of category labels indexed by the ID, which seems a more natural data set we would want this function to return.

That works well with label mapping implementation but could as well be re-written as
```
if (array_key_exists($categoryId, $this->categories)) {
  return $this->categories[$categoryId];
}
```
Or what do you think 🤷‍♂️

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/AbsenceTypeTest.php
```diff
@@ -161,4 +161,17 @@ public function testAbsenceTypeHasEverBeenUsedReturnsFalseWhenAbsenceTypeIsNotLi
     $absenceTypeID = 1;
     $this->assertFalse($this->absenceTypeService->absenceTypeHasEverBeenUsed($absenceTypeID));
   }
+
+  public function testAbsenceTypeGetCategories() {
```
tunbola: No need to add AbsenceType in the test name since the class namespace already has the AbsenceType keyword.. `testGetCategories` should be enough.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
```diff
@@ -212,6 +212,15 @@ public static function getValuesArray($id) {
     return $absenceType;
   }
 
+  /**
+   * Returns a list of all possible categories for absence type
+   *
+   * @return array|bool
```
tunbola: Does this return a boolean also, under what conditions?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -1,10 +1,12 @@
 <?php
 
 use CRM_HRLeaveAndAbsences_Service_AbsenceType as AbsenceTypeService;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 
 class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
 
   private $links = array();
+  private $categories = array();
```
tunbola: Please use short array syntax here and in the statement above.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -183,4 +186,21 @@ private function canNotDelete($absenceType) {
     return $this->absenceTypeService->absenceTypeHasEverBeenUsed($absenceType->id)
            || $absenceType->is_reserved;
   }
+
+  /**
+   * Retrieves the category label from option value
+   *
+   * @param int $categoryId
+   *
+   * @return mixed
```
tunbola: Should this not just return a string?

ajesamson: It used mixed because empty string was not explicitly return if key does not exist.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php
ajesamson: Return type of `buildOptions` is `array|bool` which explains the reason for this return type.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/BAO/AbsenceTypeTest.php
```diff
@@ -865,4 +865,17 @@ public function testIsCalculationUnitInHoursReturnsFalseWhenCalculationUnitIsNot
...
+  public function testGetCategories() {
+    $categoryOptions = [
+      '1' => 'Leave',
+      '2' => 'Sickness',
+      '3' => 'TOIL',
+      '4' => 'Custom'
+    ];
+    $categories = AbsenceType::getCategories();
+
+    $this->assertCount(4, $categories);
```
tunbola: On a side note, we don't really need this assertion, checking that both arrays are equal is enough.

PR #2881, opened 2018-10-09
PCHR-4217: Remove is-default option from leave type list

ajesamson:
## Overview
This PR removes `is_default` from leave/absence type listing page.

## Before
before_is_default

## After
after_is_default

## Technical Details
Since the is_default field was no longer in use, an upgrader was added to remove the column for existing sites.
```
public function upgrade_1035() {
  $absenceTypeTable = AbsenceType::getTableName();
  if (SchemaHandler::checkIfFieldExists($absenceTypeTable, 'is_default')) {
    CRM_Core_DAO::executeQuery("ALTER TABLE {$absenceTypeTable} DROP COLUMN is_default");
  }

  return TRUE;
}
```
Reference were also removed from test implementation and the DAO was rebuilt after updating the [AbsenceType XML](https://github.com/compucorp/civihr/blob/master/uk.co.compucorp.civicrm.hrleaveandabsences/xml/schema/CRM/HRLeaveAndAbsences/AbsenceType.xml#L47) configuration file.


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
```diff
@@ -16,7 +16,6 @@
             <thead class="sticky">
               <th>{ts}Title{/ts}</th>
               <th>{ts}Allow Accruals?{/ts}</th>
-              <th>{ts}Is default?{/ts}</th>
```
tunbola: Does this mean that we would no longer be needing this field ever again (please confirm)? 
In that case it may be a good idea to remove the field from the db and also from the DAO and clean up code that still use the field.

ajesamson: Yes we will not. It was said that 

> is_default is problematic as actually different users may have different sets of leave, and may have more than one leave type in a category

I will add clean up update.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -21,7 +21,7 @@ public function run() {
 
   public function browse() {
     $object = new CRM_HRLeaveAndAbsences_BAO_AbsenceType();
-    $object->orderBy('category');
+    $object->orderBy('weight');
```
tunbola: The PR description says that the absence type is ordered by category, but it is ordered by weight here?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1035.php
```diff
@@ -0,0 +1,24 @@
...
+
+use CRM_Core_BAO_SchemaHandler as SchemaHandler;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+
+trait CRM_HRLeaveAndAbsences_Upgrader_Step_1035 {
+
+  /**
+   * Removes the is_default column from absence type table.
+   * The initial use for setting default type is changed due to introduction of
+   * categories. This mean multiple absence type can belong to same category,
```
tunbola: s/mean/means

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
ajesamson: There was a misinterpretation of one of the ordering requirements, calling for this change. I will update the description.

tunbola:
A bit of correction in the PR description, Its the DAO that actually gets rebuilt/regenerated not the BAO.

PR #2883, opened 2018-10-11
PCHR-4291: Load absence types available (suggested) colours

igorpavlov:
## Overview

This PR provides an ability to load available (suggested) colours for absence types. This is needed to suggest colours when creating a new absence type because all colours should be unique per each absence type.

Note: despite of the suggestion on the colours palette, if all colours have already been used, the full list of colours will be returned in order to not to prevent admins from creating more absence types.

Use Absence Type model method: `AbsenceType.getAvailableColours()`

## Usage

```js
AbsenceType.getAvailableColours()
  .then(function (availableColours) {
    console.log(availableColours);
  });

// yields ["#5A6779", "#3D4A5E", ...]
```

### Move colours list to front-end via constants

A new constants module has been created `js/src/leave-absences/shared/modules/constants.js`.

```js
define([
  'common/angular'
], function (angular) {
  'use strict';

  angular.module('leave-absences.constants', [])
    .constant('absence-type-colours', [
      '#5A6779',
      '#3D4A5E',
      ...
```

### Allow to fetch only needed fields for Absence Types

*Augment Absence Type model to allow to pass additional params*
```js
models.factory('AbsenceType', ...
  // ...
  return Model.extend({
    all: function (params, additionalParams) {
      return absenceTypeAPI.all(params, additionalParams)
        // ...
```

*Augment the Absence Type API to allow to pass additional params*
```js
// note, it now utilises `this.getAll` and not `this.all`
apis.factory('AbsenceTypeAPI', ...
  return api.extend({
    all: function (params, additionalParams) {
      var sort;

      params = _.defaults({}, params, { is_active: true });
      sort = _.get(params, 'options.sort') || 'weight ASC';

      return this.getAll('AbsenceType', params, undefined, sort, additionalParams)
        .then(function (data) {
          return data.list;
```

### Allow to fetch available absence types colours

```js
// absenceTypeColours are stored as a constant in constants module
getAvailableColours: function () {
  return absenceTypeAPI.all(null, { return: ['color'] })
    .then(function (absenceTypes) {
      var usedColours = _.map(absenceTypes, 'color');
      var availableColours = _.difference(absenceTypeColours, usedColours);

      return availableColours;
```
### Backend cleanup

All the code and references related to absence type colours were removed from the backend, including according tests.

-----

✅Manual Tests - passed
⏹Backstop tests - not needed
✅PHP Unit Tests - amended and passed


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/models/absence-type.model.js
```diff
@@ -64,6 +66,22 @@ define([
...
             });
         },
+
+        /**
+         * Retrieves yet unused absence type colours
+         *
+         * @return {Promise} resolves with {Array}
+         */
+        getAvailableColours: function () {
+          return absenceTypeAPI.all(null, { return: ['color'] })
```
reneolivo: why implement the aditional parameter when you can pass `{ return: ['color'] }` directly to the filters?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/modules/constants.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-env amd */
+
+define([
+  'common/angular'
+], function (angular) {
+  'use strict';
+
+  angular.module('leave-absences.constants', [])
+    .constant('absence-type-colours', [
```
reneolivo: why name this constant `absence-type-colours` and not `absenceTypeColours`?

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/modules/constants.spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+  describe('constants', function () {
+    var absenceTypeColours;
+
+    beforeEach(module('leave-absences.constants'));
+    beforeEach(inject(['absence-type-colours', function (_absenceTypeColours_) {
+      absenceTypeColours = _absenceTypeColours_;
+    }]));
+
+    describe('absence types colours', function () {
```
reneolivo: I don't think there was a need to test the constants, but ok 🤷‍♂️

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
```diff
@@ -247,10 +247,7 @@
                   preferredFormat: "hex3",
                   allowEmpty:true,
                   showPaletteOnly: true,
-                  showPalette:true,
-                  {/literal}
-                  palette: {$availableColors}
-                  {literal}
```
reneolivo: why did we remove this?

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/models/absence-type.model.js
igorpavlov: This is in order to be a bit more consistent with the signature of the API.getAll method. This was defined in technical steps and was approved by @AkA84 .

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/modules/constants.js
igorpavlov: Good call. Should we uppercase them instead? https://github.com/johnpapa/angular-styleguide/issues/736

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/modules/constants.spec.js
igorpavlov: What if someone adds a constant that does not comply with requirements?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl
igorpavlov: Because we remove everything related to colours palette from PHP. If we leave this it would access not anymore existing variable.

The variable was removed here https://github.com/compucorp/civihr/pull/2883/files#diff-5d6ce48c2bb46daa103a184e08eaf369L56

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/shared/modules/constants.js
reneolivo: Upper case is much better for constants. Please go ahead!

uk.co.compucorp.civicrm.hrleaveandabsences/js/test/shared/modules/constants.spec.js
reneolivo: typing `red` for example? If that's going to break the code, then it's fine. Though I doubt this would happen. But since it's already here, lets leave it.

ajesamson:
@igorpavlov Kindly other [references](https://github.com/compucorp/civihr/blob/master/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L77) such as param check. 💣

igorpavlov:
@ajesamson The reference you provided is *not* related to the colour *palette*, it is about the colour parameter, we still should be able to save colours. We just don't have a palette coming from backend anymore.

image

^ So I still can pick up and save colours, the palette is just gone.

PR #2884, opened 2018-10-11
PCHR-4249: Existing Leave Types to be Working after Upgrade

ajesamson:
## Overview
This PR updates existing sites absence type by linking the newly created category column to option group used for listing available values.

## Before
Absence type category were not initialized

## After
Absence type category are now initialized

## Technical Details
When the category column was created in `civicrm_hrleaveandabsences_absence_type` table, it was done with `NOT NULL` property. This implies the value will be empty for existing sites. This was used as criteria for differentiating between new and existing systems. 

All absence types were first fetched
```
private function up1036_getAbsenceTypes() {
  $absenceTypes = [];
  $absenceType = new AbsenceType();
  $absenceType->find();
  while($absenceType->fetch()) {
    $absenceTypes[] = clone $absenceType;
  }

  return $absenceTypes;
}
```
and then updated only if category was empty.
```
...
  foreach ($absenceTypes as $absenceType) {
    if (empty($absenceType->category)) {
      $this->up1036_updateAbsenceTypeCategory($absenceType, $categoryOptions);
    }
  }
...
}
```

All the categories were updated based on these criteria:
1. is_sick: Sickness
2. allow_accruals_request: TOIL
3. is_sick and allow_accruals_request: Custom
4. Others: Leave 

In addition to updating the categories, leave types that have *_is_sick_* have the following additional update:
* By default should public holiday be added to the default entitlement?" Set this to NO.
* Set field "Allow staff to request to accrue additional leave of this type during the period" to NO
* Set "Must staff take public holiday as leave?" to NO
* Set "allow leave of this type to be carried forward from one period to another" to NO
* Set "Maximum amount of leave that can be carried forward " to 0
* Set "Carry forward expiration duration and unit" to 0

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1036.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1036.php
@@ -64,6 +64,9 @@ trait CRM_HRLeaveAndAbsences_Upgrader_Step_1036 {
       $absenceType->allow_accruals_request = 0;
       $absenceType->must_take_public_holiday_as_leave = 0;
       $absenceType->allow_carry_forward = 0;
+      $absenceType->max_number_of_days_to_carry_forward = 0;
+      $absenceType->carry_forward_expiration_duration = 0;
+      $absenceType->carry_forward_expiration_unit = 0;
     }
 
     AbsenceType::create($absenceType->toArray());
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1036.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1036.php
@@ -1,0 +1,71 @@
+<?php
+
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
+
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Upgrader/Step/1036.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    if (empty($absenceType->category)) {
+      $absenceType->category = $categoryOptions['leave'];
+    }
+    elseif ($absenceType->category === $categoryOptions['sickness']) {
+      $absenceType->add_public_holiday_to_entitlement = 0;
+      $absenceType->allow_accruals_request = 0;
+      $absenceType->must_take_public_holiday_as_leave = 0;
+      $absenceType->allow_carry_forward = 0;
```
davialexandre: @ajesamson when allow_accruals_request or allow_carry_forward is empty, there other related fields that should be emptied too. Check the validation logic in the Absence Type BAO and try and save some absence types using these fields in the old form to see how this works.

Another interesting test to do is to actually set the existing Sickness type to also allow carry forward and then try to migrate it using this code to see if it will work

ajesamson: I was able to replicate the issue by setting up a new system, create a new absence type and setting the allow carry forward details. The absence type was then activated for `is_sick`, leading to upgrade failure after the upgrader was executed.
The affected fields are now added and tested successfully.

PR #2885, opened 2018-10-12
PCHR-4278: Transform leave type webform to page

ajesamson:
## Overview
This PR removes the Absence Type webform. This is required because of the change in how to process the form using javascript(AngularJS).

## Before
There is a form page to manage absence type item creation using `CRM_Core_Page_Basic` implementation.

## After
The previous form page was converted to normal page which renders javascript managed form dynamically.

## Technical Details
The `CRM_HRLeaveAndAbsences_Page_AbsenceType` class was updated to extend from `CRM_Core_Page` instead of `CRM_Core_Page_Basic`. The reason for this is that the entire logic for automatically handling forms is coming from this class and since there won't be a QuickForm anymore, the change was necessitated. 

The page `run` method was updated to check submitted action type and render the page (with required resources) if `update` or `add` action was requested. Otherwise, the page `browse` function was called to load records.
```
public function run() {
  CRM_Utils_System::setTitle(ts('Leave/Absence Types'));
  $this->absenceTypeService = new AbsenceTypeService();
  $action = CRM_Utils_Request::retrieve('action', 'String', $this, FALSE, 'browse');
  $this->assign('action', $action);

  if ($action & CRM_Core_Action::BROWSE) {
    $this->browse();
  } else {
    $this->loadResources();
  }

  parent::run();
}
```
Page resources loading were also refactored into another function
```
private function loadResources() {
    ->addStyleFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'css/leaveandabsence.css')
    ->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/crm-app-list.min.js', 1001)
    ->addScriptFile('civicrm', 'js/jquery/jquery.crmEditable.js', CRM_Core_Resources::DEFAULT_WEIGHT, 'html-header')
    ->addScriptFile('uk.co.compucorp.civicrm.hrleaveandabsences', 'js/dist/leave-type-wizard-form.min.js', 1001)
    ->addVars('leaveAndAbsences', [
        'baseURL' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.hrleaveandabsences')
    ]);
}
```

The page template was also modified to inject the directive for managing the form wizard
```
<div id="bootstrap-theme" class="crm-leave-and-absences-list-block"
     data-leave-absences-leave-type-wizard>
  <leave-type-wizard></leave-type-wizard>
</div>
```
The template is restricted to only adding and updating record. This meant to restrict deleting of item to the listing page only.
```
{if in_array($action, array(1,2))}
```

## Comments
The `Form/AbsenceType.php` and `Form/AbsenceType.tpl` were removed since they are no longer needed. Webform test implementation for AbsenceType was also removed. All these will be replaced by javascript implementations.


uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
```diff
@@ -1,5 +1,8 @@
-{if $action eq 1 or $action eq 2 or $action eq 8}
-  {include file="CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl"}
+{if $action eq 1 or $action eq 2}
+  <div id="bootstrap-theme" class="crm-leave-and-absences-list-block"
+       data-leave-absences-leave-type-wizard>
```
igorpavlov: Let's use this as an indentation:

```html
<div id="bootstrap-theme" class="crm-leave-and-absences-list-block"
  data-leave-absences-leave-type-wizard> <!-- just 2 spaces here, do not match with attr position above -->
  ...
</div>

File: uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl [hunk 2]
```diff
@@ -1,5 +1,8 @@
-{if $action eq 1 or $action eq 2 or $action eq 8}
-  {include file="CRM/HRLeaveAndAbsences/Form/AbsenceType.tpl"}
+{if $action eq 1 or $action eq 2}
```
igorpavlov: Can we use some other notation (like here https://www.smarty.net/forums/viewtopic.php?p=34638) so we check is the action `IN` one of 2 options instead of checking if it equals one and then another?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
```diff
@@ -1,5 +1,8 @@
...
+  <div id="bootstrap-theme" class="crm-leave-and-absences-list-block"
+       data-leave-absences-leave-type-wizard>
+    <leave-type-wizard></leave-type-wizard>
+  </div>
 {else}
```
igorpavlov: I am not sure I get the idea, you check `if`, then render `<leave-type-wizard>`, but if `else`, you still render `<leave-type-wizard>`?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
ajesamson: Yes we can. This is good especially if we have many options 🤷‍♂️

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -16,6 +16,15 @@ class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
   public function run() {
     CRM_Utils_System::setTitle(ts('Leave/Absence Types'));
     $this->absenceTypeService = new AbsenceTypeService();
+    $action = CRM_Utils_Request::retrieve('action', 'String', $this, FALSE, 'browse');
+    $this->assign('action', $action);
+
+    if ($action & CRM_Core_Action::BROWSE) {
+      $this->browse();
+    } else {
+      $this->loadResources();
```
tunbola: The resources are not needed for browsing? I would have thought the browse would use some JS scripts?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -3,7 +3,7 @@
 use CRM_HRLeaveAndAbsences_Service_AbsenceType as AbsenceTypeService;
 use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 
-class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
+class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page {
```
tunbola: The CRM_Core_Page_Basic is a child of the CRM_Core_Page, I would have thought if we need less functionality, we should have stuck with CRM_Core_Page_Basic class?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -16,6 +16,15 @@ class CRM_HRLeaveAndAbsences_Page_AbsenceType extends CRM_Core_Page_Basic {
   public function run() {
     CRM_Utils_System::setTitle(ts('Leave/Absence Types'));
     $this->absenceTypeService = new AbsenceTypeService();
+    $action = CRM_Utils_Request::retrieve('action', 'String', $this, FALSE, 'browse');
+    $this->assign('action', $action);
+
+    if ($action & CRM_Core_Action::BROWSE) {
```
tunbola: Why do we need to manually assign this now? seems it was done automatically before?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
ajesamson: The JS scripts were added [here](https://github.com/compucorp/civihr/pull/2865/files#diff-d77b9ef8ffe49db8d6d55fb08ac88bd2R44). Will refactor since i have moved the implementation away from listing page. ♻️

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
ajesamson: This same page is used for both listing and modifying of records. As stated in the description, this allow us to render and manage the flow of the form. So we need to know what action was requested by user and render appropriate page.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
ajesamson: CRM_Core_Page_Basic extends the functionality of CRM_Core_Page, adding facility for rendering form (using QuickForm). We don't need all the extras since the form is now processed with AngularJS.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
ajesamson: :cut I have removed the implementation since it's no longer needed.

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
tunbola: I understand, but before this change, the template has logic like `{if $action eq 1 or $action eq 2 or $action eq 8}`, which means the action variable was being sent to the template, why do we need to do this manually now?

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
```diff
@@ -1,13 +1,11 @@
-{if $action eq 1 or $action eq 2}
+{if in_array($action, array(1,2))}
```
igorpavlov: Can we use `[1, 2]` notation instead of `array(1, 2)`?

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
```diff
@@ -189,15 +189,22 @@ private function getCategoryLabel($categoryId) {
 
   /**
    * Loads required page resources
+   *
+   * @param bool $addExtraResources
    */
-  private function loadResources() {
+  private function loadResources($addExtraResources) {
```
igorpavlov: `addExtraResources` may not be descriptive enough. May be name it `addResourcesForAddingOrEditing`? The name I proposed may not be the best one, so please feel free to propose yours.

ajesamson: 👍 Thanks @igorpavlov. @tunbola what do you suggest? I also had difficulty on what to name it without using a long name 🤦‍♂️

uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/Page/AbsenceType.php
ajesamson: The existing implementation does not seem work and not so visible because we have a separate  template handling editing and updating of absence types. So we need to call the `assign` method for the variable to be available in the template.

uk.co.compucorp.civicrm.hrleaveandabsences/templates/CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl
ajesamson: I don't think we can. Using `[]` as `{if in_array($action, [1,2])}` led to error
```
User error: Smarty error: [in CRM/HRLeaveAndAbsences/Page/AbsenceType.tpl line 1]: syntax error: unbalanced parenthesis in if statement (Smarty_Compiler.class.php, line 1277) in Smarty->trigger_error() (line 1100 of /vagrant/test2/sites/all/modules/civicrm/packages/Smarty/Smarty.class.php).
```

PR #2890, opened 2018-10-17
PCHR-4305: Filter Unwanted Fields From Import Contact Screen.

tunbola:
## Overview
As part of the changes introduced by the Staff menu improvements epic, this PR removes hides some unwanted fields and their labels from the Import Contacts page.

## Before
The contact import screen was un-modified.

import contacts _ staging54 2018-10-17 13-46-06

## After
The following fields were hidden from the import contacts page. In order to avoid validation issues and also prevent unexpected results, the default expected values of these fields were also set.

1. Choose Data Source data source field 
2. Contact type 
3. Dedupe options
4. Import Field Separator

import contacts _ staging54 2018-10-17 14-37-00

## Technical Details
Hook `hrcore_civicrm_buildForm` was used to hide freeze these fields and also hide them by injecting appropriate CSS into the form page.

## Manual Testing
A sample CSV was uploaded to create contacts and the contacts were created as expected with no errors to make sure that the functionality still works as expected.

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/ContactImportFieldsFilter.php
+++ b/uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/ContactImportFieldsFilter.php
@@ -1,0 +1,74 @@
+<?php
+
+class CRM_HRCore_Hook_BuildForm_ContactImportFieldsFilter {
+
+  /**
+   * @var array
+   *  A an array of restricted fields as keys and
+   *  the default values they are to be set to
+   *  as values.
+   */
+  private $restrictedFields = [
+    'dataSource' => CRM_Import_DataSource_CSV::class,
+    'contactType' => CRM_Import_Parser::CONTACT_INDIVIDUAL,
+    'dedupe' => NULL,
+    'fieldSeparator' => ',',
+  ];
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/ContactImportFieldsFilter.php
```diff
@@ -0,0 +1,74 @@
...
+      if ($form->elementExists($field)) {
+        $hiddenOptionsStyle .= ",  .crm-import-datasource-form-block-$field";
+        $form->freeze($field);
+        if ($defaultValue) {
+          $defaults[$field] = $defaultValue;
+        }
+      }
+    }
+
+    $hiddenOptionsStyle .= ' { display: none; }';
```
mickadoo: I can see why it needs to be done this way, but it would be so much nicer if Civi had some option like `$form->hide($elementName);`

tunbola: Yeah, I agree, the closest to that is `$form->removeElement($elementName)` but that would totally remove the element which is not what we need.

PR #2893, opened 2018-10-18
PCHR-4307: Filter Unwanted Fields From Import Custom Fields Screen

tunbola:
## Overview
As part of the changes introduced by the Staff menu improvements epic, this PR removes hides some unwanted fields and their labels from the Import Custom Fields page.

## Before
The custom fields import screen was un-modified.

mg8lf3lwka6gymd1ycyjd64cyii6mb

## After
The following fields were hidden from the import custom fields page. In order to avoid validation issues and also prevent unexpected results, the default expected values of these fields were also set.
- Contact type
- Import Field Separator

import multi-value custom data _ staging54 2018-10-18 16-13-08

## Technical Details
Hook hrcore_civicrm_buildForm was used to hide freeze these fields and also hide them by injecting appropriate CSS into the form page.


uk.co.compucorp.civicrm.hrcore/CRM/HRCore/Hook/BuildForm/CustomImportFieldsFilter.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+class CRM_HRCore_Hook_BuildForm_CustomImportFieldsFilter {
+  /**
+   * @var array
+   *  A an array of restricted fields as keys and
```
ajesamson: A type here i guess 🤷‍♂️ 
`An array of restricted fields as key and...`

PR #2895, opened 2018-10-19
PCHR-4313: Imitate BE validations on FE for Leave Type Wizard

igorpavlov:
## Overview

This PR imitates BE validations on FE for better UX.

## After

title

### Validating title

1. We load absence types, but fetch only titles:

```js
function loadUsedLeaveTypeTitles () {
  return AbsenceType.all({}, { return: ['title'] })
    .then(function (absenceTypes) {
      usedLeaveTypeTitles = _.map(absenceTypes, 'title');
```
and then simply use the loaded titles in the watcher:

```js
function watchTitleField () {
  var titleField = vm.fieldsIndexed.title;

  $scope.$watch(function () {
    return titleField.value;
  }, function (title) {
    if (title === '') {
      return;

    if (_.includes(usedLeaveTypeTitles, title)) {
      titleField.error = 'This leave type title is already in use';
```


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
```diff
@@ -363,13 +378,33 @@ define([
...
+     * Initiates a watches over "Title" field.
+     * Watches for already used leave types titles.
+     */
+    function watchTitleField () {
+      var titleField = vm.fieldsIndexed.title;
+
+      $scope.$watch(function () {
+        return titleField.value;
+      }, function (title) {
+        if (title === '') {
```
reneolivo: I'd use `_.isEmpty(title)` to be extra safe
```suggestion
        if (_.isEmpty(title)) {
```

PR #2900, opened 2018-10-24
PCHR-4231: Add BEM class for BackstopJS tests

igorpavlov:
## Overview

This PR adds a BEM class to the "next section" button in order to provide an access to it for BackstopJS.

### Core overrides
_If the PR overrides/patches a core file, then please explain here, for each file:_

1. _Which file it is_
2. _The reason why it was overridden/patched_
3. _What the override/patch consists of_

_Also make sure to update this [wiki page](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/100532509/Overridden+CiviCRM+files) with your changes_

## Comments
_Anything else you would like the reviewer to note_


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
```diff
@@ -484,6 +480,35 @@ define([
...
+      });
+    }
+
+    /**
+     * Sets the availability of the sections that follow the active section.
+     * Also sets the availability of the "Next section" button for the active section.
+     *
+     * @param {Boolean} enabled
+     */
+    function setAvailabilityOfFollowingSections (enabled) {
```
reneolivo: can we change the name of this param to `isDisabled` ? Then we can remove the `!enabled` from here and just use `isDisabled`. Obviously you'd need to change the call to this function.

PR #2903, opened 2018-10-26
PCHR-4310: Modify Bulk Actions After Search for Staff Directory Page

tunbola:
## Overview
This PR modifies the bulk actions available on the search results page for the Staff directory page.

## Before
All bulk actions provided by Civi are available for the Search directory page.
staff directory _ staging54 2018-10-26 12-50-01

## After
After searching only the following bulk actions are available:
* Create User Accounts
* Delete Staff (Delete Contacts)
* Delete Staff permanently (Delete Contacts permanently
* Export Staff (Export Contact)
* Print merge documents

staff directory _ staging54 2018-10-26 12-48-45

## Technical Details
The `buildTaskList` method of the `CRM_HRCore_Form_Search_StaffDirectory` class was modified to return the required bulk actions.
There was an issue with Export Staff bulk action yielding an empty CSV which is as a result of the function processing the export functionality [here](https://github.com/compucorp/civihr/blob/49b363e09bfb5d9fd647ba14f7672be70c686c7e/hrjobcontract/CRM/Export/BAO/Export.php#L1123) , 0 is passed for both `offset` and `rowCount` parameters, this was fixed by making provision for that in the `all` function of `CRM_HRCore_Form_Search_StaffDirectory` class.
```php
    if ($offset || $rowCount) {
      $sql .= " LIMIT $offset, $rowCount";
    }
```
Also after this issue was fixed, all the contacts were exported in the generated CSV irrespective of whether the contacts was checked or not, the issue was fixed by putting the `$includeContactIDs` parameter passed to the `all` function in consideration.

The function to retrieve contact_ids from the form values sent via the checkbox was copied(also refactored) from [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Contact/BAO/Query.php#L3631-L3646)
```php
  public function includeContactIDs() {
    $contactIds = [];
    foreach ($this->formValues as $id => $value) {
      if (substr($id, 0, CRM_Core_Form::CB_PREFIX_LEN) == CRM_Core_Form::CB_PREFIX) {
        $contactIds[] = substr($id, CRM_Core_Form::CB_PREFIX_LEN);
      }
    }

    CRM_Utils_Type::validateAll($contactIds, 'Positive');
    if (!empty($contactIds)) {
      $this->where[] = " ( contact_a.id IN (" . implode(',', $contactIds) . " ) ) ";
      $this->whereClause =  ' WHERE ' . implode(' AND ', $this->where);
    }
  }
```

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Form/Search/StaffDirectoryTest.php
+++ b/uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Form/Search/StaffDirectoryTest.php
@@ -421,6 +421,32 @@ class CRM_HRCore_Form_Search_StaffDirectoryTest extends CRM_HRCore_Test_BaseHead
     $this->assertEquals($expectedResults, $results);
   }
 
+  public function testExpectedTaskListIsReturned() {
+    $formValues = [];
+    $searchDirectory =  new SearchDirectory($formValues);
+
+    $expectedTaskList = [
+      'Create User Accounts(s)',
+      'Delete Staff',
+      'Delete Staff Permanently',
+      'Export Staff',
+      'Print/merge document'
+    ];
+
+    $form = new CRM_Core_Form_Search();
+    $form->setVar('_taskList', [
+      'Create User Accounts(s)',
+      'Delete contacts',
+      'Delete permanently',
+      'Export contacts',
+      'Print/merge document',
+      'Sample Task',
+      'Add Tags'
+    ]);
+
+    $this->assertEquals($expectedTaskList, array_values($searchDirectory->buildTaskList($form)));
+  }
+
   private function extractContactIds($sql) {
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/tests/phpunit/CRM/HRCore/Form/Search/StaffDirectoryTest.php
```diff
@@ -379,6 +379,48 @@ public function testOnlyActiveContactManagersAreReturnedForTheManagerColumn() {
...
+        'manager' => NULL,
+        'location' => NULL,
+        'department' => NULL,
+        'job_title' => NULL,
+      ]
+    ];
+
+    $this->assertEquals($expectedResults, $results);
+  }
+
```
davialexandre: How about adding a test to make sure the form will return the tasklist with only the expected items?

PR #2906, opened 2018-10-26
PCHR-4317: Add menu URL for staff directory.

tunbola:
## Overview
This PR adds the URL for the Staff Directory Page. The URL opens the page listing the current staff.
## Before
- A placeholder URL was in place

## After
staffdirectory

### DIFF
--- a/uk.co.compucorp.civicrm.hrcore/config/menu/main.php
+++ b/uk.co.compucorp.civicrm.hrcore/config/menu/main.php
@@ -1,17 +1,5 @@
 <?php
 
-$result = civicrm_api3('OptionValue', 'getsingle', [
-  'option_group_id' => 'custom_search',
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.hrcore/config/menu/main.php
```diff
@@ -1,5 +1,17 @@
...
+  'name' => 'CRM_HRCore_Form_Search_StaffDirectory',
+]);
+
+$searchDirectoryURL = '';
+if (!empty($result['value'])) {
+  $searchDirectoryURL =
+    "civicrm/contact/search/custom?csid={$result['value']}&force=1&reset=1&select_staff=current";
+}
+
+
```
davialexandre: @tunbola let's avoid adding logic to configuration files. If a menu item requires some logic, then it should be a class

PR #2907, opened 2018-10-26
PCHR-4349: Fix miscellaneous issues in Leave Type Wizard

igorpavlov:
## Overview

This PR fixes a number of miscellaneous issues in Leave Type Wizard:
- Closes colour picker and Backup leave approver fields on unfocusing
- Improves help text providing the title of the field in the help header
- Fixes help text for Title field
- Fixes fields names, default values and help texts
- Fixes UI types of fields and their options
- Adds missing hidden fields (it would work without them since they are set to "False", but this is more consistent with requirements)

## Technical Details

The changes in this PR are really trivial, except unfocusing the Backup leave approver field. The problem is with opening sections with accordions headers. We previously attached an `ng-click` event to open a section to the **whole** accordion section (not only to the header) because that's how it works with **uib-accordion**. However, this triggered opening section event if we clicked at **any** place inside the section, including a form, which is obviously a bad thing. We simply stopper the click propagation previously:

```html
<div class="row" ng-click="$event.stopPropagation()">
```
This is a wrapper of the form, which means if we click on the form it will not propagate to the section. However that created another issue - now the autocomplete input would not close on "blur". The problem is that the input selector closes on **document** click (see https://github.com/angular-ui/ui-select/issues/228). Because the click event never propagated to the document, we had to find another solution.

We attached a function that checks from where the click event was fired. If from the header - open section, otherwise - do nothing.

```js
function checkIfAccordionHeaderClicked ($event) {
  var className = 'panel-heading';
  var $sourceElement = angular.element($event.originalEvent.path);
  var isHeaderOrElementInsideHeader = $sourceElement.hasClass(className) ||
    !!$sourceElement.closest('.' + className).length;

  return isHeaderOrElementInsideHeader;
```

and the usage:

```html
<div uib-accordion-group class="panel-default ..."
  ...
  ng-click="form.checkIfAccordionHeaderClicked($event)
    && !section.disabled
    && form.openSection(sectionIndex)">
```

-----

✅Manual Tests - passed
✅Jasmine Tests - amended & passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


reneolivo:
@igorpavlov as discussed I recommend using a directive that takes care of stopping the accordion from completely closing because:

* It removes DOM references from the component
* It will simplify the header logic (no need to check if you can click on the header + check if it's disabled etc.)
* It allows this functionality to be applied elsewhere

A sample implementation would be:

```js
module.directive('alwaysDisplayAtLeastOneAccordionPanel', function ($scope, $element) {
  var $panelTitle = $element.find('.panel-title');

  $panelTitle.on('click', function (event) {
    var $panel = $(this).parents('.panel');

    if ($panel.hasClass('panel-opened')) {
      event.stopPropagation();
    }
  });
});
```

```html
<uib-accordion always-display-at-least-one-accordion-panel>
  ...
</uib-accordion>
```

(super nice name)


Since there are time constraints and this is my only nitpick I'll approve the PR.

igorpavlov:
@reneolivo I definitely considered something like this, however, due to the lack of time and necessity to move forward I would leave it as-is for now. This solution is pretty much harmless because it does **not** modify DOM, it just replaces the `stopPropagation` adhoc that was accepted earlier. I will definitely consider refactoring at the end of the epic if we have some extra time along with other noted improvements I have.

PR #2914, opened 2018-11-01
PCHR-4203: Support TOIL category in Leave Type Wizard

igorpavlov:
## Overview

This PR adds a possibility to create a TOIL leave type via Leave Type Wizard.

## After

toil

### Performance enhancements

We are now lazy-loading Absence Types titles, colours and contacts. That means we show the wizard sooner and estimate other data to be loaded before user starts interacting with wizard. However, if users starts interacting before all data is loaded - we gracefully handle this.

Note: timeouts were used to imitate the slow speed of loading.

#### Absence types used titles

title

#### Colours and contacts

colours

### Enabling TOIL

TOIL category is already fetched via options groups. At the moment we have a "suppressor" function that disables not yet supported categories. All we need to do is to remove TOIL from the suppressor:

```js
function temporarilySuppressNotYetUsedLeaveCategories () {
  vm.leaveTypeCategories = _.filter(vm.leaveTypeCategories, function (category) {
    return !_.includes(['toil', 'custom'], category.name); // <<< remove TOIL
```
Also we add an icon:

```js
.constant('leave-type-categories-icons', {
  ...
  toil: 'clock-o'
```

### Force values to be sent to backend

Using earlier build functionality and just change constants:

```js
.constant('defaults-by-category', {
  ...
  toil: {
    must_take_public_holiday_as_leave: false,
    allow_accruals_request: true,
    is_sick: false
```

### Hiding a field by Category

This PR requires to hide a specific field if a category was selected:

```js
.constant('fields-hidden-by-category', {
  allow_overuse: {
    leave: false,
    sickness: false,
    toil: true

// ...

function toggleFieldsDependingOnLeaveCategory (category) {
  _.each(fieldsHiddenByCategory, function (categories, fieldName) {
    vm.fieldsIndexed[fieldName].hidden = categories[category];
```

### Change tab label on category change

Simply have a constant and loop on category change:

```js
.constant('custom-tab-names-by-category', {
  'leave-requests': {
    leave: LEAVE_REQUESTS_TAB_DEFAULT_LABEL, // default
    sickness: LEAVE_REQUESTS_TAB_DEFAULT_LABEL,
    toil: 'Using TOIL'

// ...

function setCustomTabNamesByCategory (category) {
  _.each(customTabNamesByCategory, function (categories, tabName) {
    tabsIndexed[tabName].label = categories[category];
```

### _.isEmpty

`_.isEmpty` returns `true` when any Boolean passed. This is a bit weird so instead now we check for emptiness differently: `value === '' || value === null`

### Lazy loading contacts and colours

We are now loading Contacts and Leave Type Colours after showing the form. This in order to speed the loading time up.

```js
loadLeaveTypeCategories()
  ...
  .then(function () {
    ...
    vm.fieldsIndexed.title.waitsForSupportiveData = true;
  })
  ...
  .then(function () {
    return $q.all([
      ...
      loadContacts(),
      loadAvailableColours()
    ]);
  });
```

⚠️Please note, it will be important to check if all supported data is loaded before saving while **editing**. This is a not for future PRs.

----

✅Manual Tests - passed
✅Jasmine Tests - amended & passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
```diff
@@ -64,6 +64,9 @@ define([
         .then(function () {
           tabsIndexed = indexPropertyByName(vm.sections, 'tabs');
           vm.fieldsIndexed = indexPropertyByName(tabsIndexed, 'fields');
+          promises.titlesInUse = fetchAbsenceTypesTitlesInUse();
+
+          promises.titlesInUse.then(validateTitleField);
```
AkA84: this logic doesn't seem to flow naturally. Why would you validate the field after the titles have been fetched? One necessarily doesn't follow the other, as the field should be validated only when the its value changed, not when the titles have loaded

For example: if i go to the wizard page and then do nothing for the next 30 seconds, why should the wizard try to validate the title at all? All you are supposed to be doing here is to fetch the data that will be used for the validation _at a later time_

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
```diff
@@ -96,26 +103,27 @@ define([
...
+     *
+     * @return {Promise}
      */
     function checkIfTitleIsUnique () {
       var titleField = vm.fieldsIndexed.title;
-      var title = titleField.value;
 
-      if (_.isEmpty(title)) {
-        return;
+      if (_.isEmpty(titleField.value) || titleField.waitsForSupportiveData) {
```
AkA84: `waitsForSupportiveData` is not representative of what this flag is about. You turn it `true` right before checking if its unique, you turn it `false` right after the check. Its value then has nothing to do on whether the "supportive data" is already there or not

As I suggested originally, please rename the property to something more descriptive like `validating`

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
igorpavlov: Yes, I am not sure why I added this, good catch. Thanks.

uk.co.compucorp.civicrm.hrleaveandabsences/js/src/leave-absences/leave-type-wizard/components/leave-type-wizard.component.js
igorpavlov: Done.

PR #2920, opened 2018-11-06
PCHR-4328: Create Option Groups for Organization Contact in Job Role

ajesamson:
## Overview
For setting up job role funder, organisation contact records were created for such. This PR creates an option group for the purpose of managing funders list.

## Before
There were no option group for managing job role funders.

## After
An option group is created for managing list of funders.

## Technical Details
An XML file was created for setting up funders option groups. This was then installed by an upgrader, creating the required option group in the process.

```
if ($result['count'] === 0) {
  $file = 'xml/option_groups/organisation_contact_install.xml';
  $this->executeCustomDataFile($file);
}
```

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -79,6 +79,25 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
   }
 
   /**
+   * Creates an option group for organisation provider
+   * used for job role funder
+   *
+   * @return bool
+   */
+  public function upgrade_1007() {
+    $result = civicrm_api3('OptionGroup', 'get', [
+      'name' => 'hrjc_funder',
+    ]);
+
+    if ($result['count'] === 0) {
+      $file = 'xml/option_groups/organisation_provider_install.xml';
+      $this->executeCustomDataFile($file);
+    }
+
+    return TRUE;
+  }
+
+  /**
    * Deletes cost centre option value with name "other"
    */
   private function deleteCostCentreOther() {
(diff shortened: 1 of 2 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,25 @@ public function upgrade_1006() {
...
 
+  /**
+   * Creates an option group for organisation contact
+   * used for job role funder
+   *
+   * @return bool
+   */
+  public function upgrade_1007() {
+    $result = civicrm_api3('OptionGroup', 'get', [
+      'name' => 'hrjc_funders',
```
tunbola: I think we should use singular here too to be consistent also with previous option groups for job contracts and job roles like: hrjc_department, hrjc_location e.t.c

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,25 @@ public function upgrade_1006() {
...
+    ]);
+
+    if ($result['count'] === 0) {
+      $file = 'xml/option_groups/organisation_contact_install.xml';
```
tunbola: Not sure if `organisation_contact` is the best name for this option group file, one might be mistaken that it stores contacts or something and it's a bit generic. What about, `organisation_provider` or a better name. What do you think?
Same goes for other PR.

ajesamson: This is a good idea. 🍭

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,25 @@ public function upgrade_1006() {
     return TRUE;
   }
 
+  /**
+   * Creates an option group for organisation contact
```
tunbola:
```suggestion
   * Creates an option group for organisation provider
```

PR #2942, opened 2018-11-15
PCHR-4331: Access Health Insurance Options Without Navigating

ajesamson:
## Overview
Health insurance providers are previously loaded from contacts but https://github.com/compucorp/civihr/pull/2919 creates an option group meant to replace this. This requires pointing all providers select menu to their option values instead of organisation contact. 

## Before
Health insurance select menu was loaded from organisation contacts created for health insurance providers.
before_health

## After
Health insurance select menu load providers from option group value created for the purpose of storing health insurance providers.

health_after

after_health

## Technical Details
A tool icon was added to health insurance provider. This allows user to access health insurance options without navigating to settings page.
```
<a href ng-click="openProvidersEditor('hrjc_health_insurance_provider')">
</a>
```
The `openProvidersEditor` uses `crmAngService` to load the page. Modifying the options automatically updates the list of loaded values.
Loading/ updating of option values was carried out by implementing CRM api call for `getOptionValues`.
```
  'sequential': 1,
  'is_active': 1,
  'option_group_id.name': { 'IN': multiple ? groupNames : [groupNames] },
  'return': [ 'id', 'label', 'weight', 'value', 'is_active', 'option_group_id', 'option_group_id.name' ],
  'options': {
    'limit': 1000,
    'sort': 'id'
  }
})
```
```
 <ui-select-match
  prevent-animations
  placeholder="Start typing a name...">{{$select.selected.label}}
</ui-select-match>
<ui-select-choices  repeat="provider.value as provider in providers.health_insurance_provider | filter: $select.search"
  refresh="loadProviderOptions(['hrjc_health_insurance_provider'])"
  refresh-delay="0"
prevent-animations>
</ui-select-choices>
```

------
✅Manual Tests - passed
✅Jasmine Tests - passed


hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -254,6 +258,19 @@ define([
       }));
     }
 
+    /**
+     * Fetches list of providers
+     *
+     *  @return {Promise}
+     */
+    function fetchProviders () {
+      var providerGroups = [
```
ajesamson: Other provider groups will be added in subsequent tickets

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -102,6 +104,8 @@ define([
 
       $rootScope.$broadcast('hrjc-loader-show');
       fetchInsurancePlanTypes().then(function () {
+        return fetchProviders();
```
AkA84: it doesn't seem that `fetchProviders` depends on `fetchInsurancePlanTypes`, so we should execute them in parallel

```js
$q.all([
  fetchInsurancePlanTypes(),
  fetchProviders()
])
  .then(function () {
    $rootScope.$broadcast('hrjc-loader-hide');
  });

File: hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js [hunk 3]
```diff
@@ -254,6 +258,19 @@ define([
       }));
     }
 
+    /**
+     * Fetches list of providers
+     *
+     *  @return {Promise}
+     */
+    function fetchProviders () {
```
AkA84: please order the functions alphabetically

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -368,6 +385,20 @@ define([
...
 
+    /**
+     * Opens providers options editor for editing
+     *
+     * @param {String} provider
+     */
+    function openProvidersEditor (provider) {
+      var url = '/civicrm/admin/options/' + provider + '?reset=1';
+
+      crmAngService.loadForm(url)
```
AkA84: where is `crmAngService` coming from?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -381,6 +412,35 @@ define([
...
+
+    /**
+     * Loads updated record of providers option group
+     *
+     * @param {Array} providers
+     * @returns {Promise}
+     */
+    function loadProviderOptions (providers) {
+      var providersList = [];
+      var optionGroup = '';
```
AkA84: this var is not used outside the `forEach` callback, so it should not be declared here

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -381,6 +412,35 @@ define([
...
+
+      return utilsService.getOptionValues(providers)
+        .then(function (data) {
+          providers.forEach(function (provider) {
+            optionGroup = provider.substr(5);
```
AkA84: this looks too much ad-hoc. Why `5`? What are we extracting here? could you move this to a utility function so that both the name and the documentation would make it clearer?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -381,6 +412,35 @@ define([
...
+     * @param {Array} providers
+     * @returns {Promise}
+     */
+    function loadProviderOptions (providers) {
+      var providersList = [];
+      var optionGroup = '';
+
+      return utilsService.getOptionValues(providers)
+        .then(function (data) {
+          providers.forEach(function (provider) {
```
AkA84: this function is too long and complex, extract what is going on here and put it in a separate function and document it. In the end it should look like this
```js
return utilsService.getOptionValues(providers)
  .then(processProvidersApiResult)  // name of function is just a suggestion
  .then(function (providers) {
    $scope.providers = providers;
  });

File: hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js [hunk 8]
```diff
@@ -381,6 +412,35 @@ define([
...
+            optionGroup = provider.substr(5);
+            providersList[optionGroup] = data.values.filter(function (item) {
```
AkA84: Use `Array.prototype.reduce`
```js
function processProvidersApiResult (data) {
  return providers.reduce(function (acc, provider) {
    acc[provider.substr(5)] = data.values
      .filter(function (optionValue) {
        return optionValue['option_group_id.name'] === provider;
      })
      .map(function (optionValue) {
        return {
          id: item.id,
          value: item.value,
          label: item.label
        };
      });
        
    return acc;
  }, {});
}
```

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
```diff
@@ -765,6 +796,35 @@ define([
...
         });
     }
+
+    /**
+     * Loads updated record of providers option group
+     *
+     * @param {Array} providers
+     * @returns {Promise}
+     */
+    function loadProviderOptions (providers) {
```
AkA84: you have all duplicated logic here, isn't there a way to avoid it

hrjobcontract/js/src/job-contract/services/utils.service.js
```diff
@@ -72,6 +73,40 @@ define([
         return deffered.promise;
       },
 
+      getOptionValues: function (groupNames) {
+        var deferred = $q.defer();
+        var multiple = _.isArray(groupNames);
+
+        CRM.api3('OptionValue', 'get', {
```
AkA84: from looking at the other methods in this service, they are all using `API.resource`, so should we use that instead of relying on `CRM.api3`. Let's try to have consistent code

hrjobcontract/js/src/job-contract/services/utils.service.js
```diff
@@ -72,6 +73,40 @@ define([
...
+          'sequential': 1,
+          'is_active': 1,
+          'option_group_id.name': { 'IN': multiple ? groupNames : [groupNames] },
+          'return': [ 'id', 'label', 'weight', 'value', 'is_active', 'option_group_id', 'option_group_id.name' ],
+          'options': {
+            'limit': 1000,
```
AkA84: why 1000? If you want all of them, then put 0

hrjobcontract/js/src/job-contract/services/utils.service.js
```diff
@@ -72,6 +73,40 @@ define([
...
+            'sort': 'id'
```
AkA84: is the sort really necessary?

hrjobcontract/js/src/job-contract/services/utils.service.js
```diff
@@ -72,6 +73,40 @@ define([
...
+          }
+        })
+          .done(function (result) {
```
AkA84: why do you need to manipulate the response? What's the format that you get from the api endpoint?

hrjobcontract/js/test/controllers/modal/modal-contract-new.controller.spec.js
```diff
@@ -216,6 +223,31 @@ define([
       });
     });
 
+    describe('when user clicks on the "Provider options" wrench icon', function () {
+      providersPopupLists = [
```
AkA84: this var is only used in this particular `describe`, so then you shouldn't declare it at the top of the test file, declare it here
```js
describe('when user clicks on the "Provider options" wrench icon', function () {
  var providersPopupLists;
```
also, don't assign values outside of `beforeEach`, otherwise you can't be sure that the state will be fully reset on every test run

```js
describe('....', function () {
  var providersPopupLists;

  beforeEach(function () {
    providersPopupLists = { /* .. */ };
  });
});
```

hrjobcontract/js/test/mocks/data/option-value.data.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-env amd */
+
+define(function () {
+  return {
+    'is_error': 0,
+    'version': 3,
+    'count': 0,
+    'values': []
```
AkA84: this is just an empty data, what's the point?

hrjobcontract/js/test/services/utils.service.spec.js
```diff
@@ -13,6 +13,17 @@ define([
   describe('utilsService', function () {
     var $httpBackend, $provide, $rootScope, $q, $uibModal, utilsService, apiService,
       AbsencePeriod, $window;
+    var groupNames = 'hrjc_health_insurance_provider';
+    var optionValueApiParams = {
```
AkA84: both `groupNames` and `optionValueApiParams` are only used in `describe('getOptionValues')`, so declare it there, and assign their value in a `beforeEach`

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
ajesamson: The provider option group is of the format `hrjc_*` (hrjc_health_insurance_provider). i was removing the `hrjc_` part so i can use the remaining part (health_insurance_provider for instance). I wanted to avoid declaring the provider as object initially. I will refactor to make it cleaner. ♻️

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.controller.js
ajesamson: Will it be advisable to make utilService return the required format (the only draw back is this will enforce other implementation using the getOptionValues to use the same format)? I will need your advice on how best to resolve this.

hrjobcontract/js/src/job-contract/services/utils.service.js
ajesamson: Sorting is not necessary. This will be catered for by UI-Select. Replacing the whole implementation with API.resource as suggested.

hrjobcontract/js/src/job-contract/services/utils.service.js
ajesamson: This was done to make it conform with implementation in job roles and affords having the list of option groups returned as an object. The whole implementation will be replace with API.resource. 

A similar [implementation](https://github.com/compucorp/civihr/blob/master/hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js#L379) to API.resource using `OptionGroup.valuesOf()` was used initially but this was throwing errors. That is why i decided to use CRM.api3.

screen shot 2018-11-15 at 10 19 33 am

hrjobcontract/js/test/controllers/modal/modal-contract-new.controller.spec.js
ajesamson: Noted. This will be updated.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
AkA84: at the very least the ones you added

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -22,6 +22,10 @@ define([
     settings, pubSub) {
     $log.debug('Controller: ModalContractNewController');
 
+    var providersKey = {
```
AkA84: treat it as a constant, and maybe add a inline comment explaining what is the key and what is the value

```js
// <explanation-of-key>: <explanation-of-value>
var PROVIDERS = {
  'hrjc_health_insurance_provider': 'health_insurance_provider'
}
```

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
```diff
@@ -254,6 +264,19 @@ define([
       }));
     }
 
+    /**
+     * Fetches list of providers
+     *
+     *  @return {Promise}
+     */
+    function fetchProviders () {
+      var providerGroups = [
```
AkA84: now that you have a "constant", i think you don't need this function (`fetchProviders`) at all anymore, you could directly do
```js
$q.all([
  fetchInsurancePlanTypes(),
  loadProviderOptions(Object.keys(PROVIDERS))
])
```

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract-new.controller.js
ajesamson: Good thought 👍

PR #2943, opened 2018-11-15
PCHR-4332: Access Life Insurance Options Without Navigating

ajesamson:
## Overview
Life insurance providers are previously loaded from contacts but https://github.com/compucorp/civihr/pull/2919 creates an option group meant to replace this. This requires pointing all providers select menu to their option values instead of organisation contact. 

## Before
Life insurance select menu was loaded from organisation contacts created for life insurance providers.

before_life_insurance

## After
Life insurance select menu load providers from option group value created for the purpose of storing life insurance providers.

after_life_insurance

after_life

## Technical Details
A tool icon was added to health insurance provider. This allows user to access life insurance options without navigating to settings page.
```
<a href ng-click="openProvidersEditor('hrjc_life_insurance_provider')">
</a>
```
```
<ui-select-match
  prevent-animations
  placeholder="Start typing a name...">{{$select.selected.label}}</ui-select-match>
  <ui-select-choices  repeat="provider.value as provider in providers.life_insurance_provider | filter: $select.search"
  refresh="loadProviderOptions(['hrjc_life_insurance_provider'])"
  refresh-delay="0"
prevent-animations>
</ui-select-choices>
```

## Comments
The underlining implementation for handling loading of options values by loadProviderOptions is implemented in https://github.com/compucorp/civihr/pull/2942.

------
✅Manual Tests - passed
✅Jasmine Tests - passed


hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.html
```diff
@@ -754,7 +754,7 @@ <h4>Life Insurance</h4>
             <div class="col-xs-12">
               <div class="form-group">
                 <label class="col-sm-4 control-label">Provider</label>
-                <div class="col-sm-6">
+                <div class="col-sm-5">
```
igorpavlov: What is the reason why we shrink it? Does it just look better as per your opinion?

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.html
```diff
@@ -763,13 +763,12 @@ <h4>Life Insurance</h4>
                       ng-model="entity.health.provider_life_insurance">
                       <ui-select-match
                         prevent-animations
-                        placeholder="Start typing a name or email...">{{$select.selected.label}}</ui-select-match>
-                      <ui-select-choices  repeat="contact.id as contact in contacts.Life_Insurance_Provider | filter: $select.search"
-                        refresh="refreshContacts($select.search, 'Life_Insurance_Provider')"
+                        placeholder="Start typing a name...">{{$select.selected.label}}</ui-select-match>
+                      <ui-select-choices  repeat="provider.value as provider in providers.life_insurance_provider | filter: $select.search"
```
igorpavlov: Please move the `repeat` to the next line aligned with `refresh` and other props.

hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.html
ajesamson: Yes, it looks much better aligning it to the plan type.

PR #2945, opened 2018-11-16
PCHR-4333: Access Pension Provider Options Without Navigation

ajesamson:
## Overview
Pension providers are previously loaded from contacts but https://github.com/compucorp/civihr/pull/2919 creates an option group meant to replace this. This requires pointing all providers select menu to their option values instead of organisation contact.

## Before
before_pension

## After
after_pension

after_pension

## Technical Details
A tool icon was added to pension. This allows user to access pension provider options without navigating to settings page.
```
<div class="col-sm-1 control-label">
  <a href ng-click="openProvidersEditor('hrjc_pension_provider')">
  </a>
</div>
```
```
<ui-select-match
  placeholder="Start typing a name...">{{$select.selected.label}}</ui-select-match>
  <ui-select-choices
    repeat="provider.value as provider in providers.pension_provider | filter: $select.search"
    refresh="loadProviderOptions(['hrjc_pension_provider'])"
    refresh-delay="0">
</ui-select-choices>
```

## Comments
The underlining implementation for handling loading of options values by loadProviderOptions is implemented in https://github.com/compucorp/civihr/pull/2942.

------
✅Manual Tests - passed
✅Jasmine Tests - passed


hrjobcontract/js/src/job-contract/controllers/modal/modal-contract.html
```diff
@@ -865,12 +865,12 @@ <h4>Pension</h4>
...
-                      <ui-select-choices  repeat="contact.id as contact in contacts.Pension_Provider | filter: $select.search"
-                        refresh="refreshContacts($select.search, 'Pension_Provider')"
+                        placeholder="Start typing a name...">{{$select.selected.label}}</ui-select-match>
+                      <ui-select-choices
+                        repeat="provider.value as provider in providers.pension_provider | filter: $select.search"
+                        refresh="loadProviderOptions(['hrjc_pension_provider'])"
                         refresh-delay="0">
-                        <div ng-bind="contact.label"></div>
-                        <small ng-bind="contact.description[0]"></small>
+                        <div ng-bind="provider.label"></div>
```
igorpavlov: Just noticed, is there any reason why we don't use `<div>{{provider.label}}</div>`? If there is no reason I would suggest to get rid of all `ng-bind` and replace with `{{ }}` in the whole document.

hrjobcontract/js/test/controllers/modal/modal-contract-new.controller.spec.js
```diff
@@ -251,6 +255,7 @@ define([
       it('calls the crmAngService with the requested url', function () {
         expect(crmAngService.loadForm).toHaveBeenCalledWith(providersPopupLists[0].popupFormUrl);
```
igorpavlov: You already have 3 lists so it may make sense to wrap in a cycle. This would also mean you will have to worry less if you add more lists.

```js
_.each(providersPopupLists, function (popupList) {
  expect(crmAngService.loadForm).toHaveBeenCalledWith(popupList.popupFormUrl);
});
```

hrjobcontract/js/test/controllers/modal/modal-contract.controller.spec.js
```diff
@@ -604,6 +608,7 @@ define([
       it('calls the crmAngService with the requested url', function () {
         expect(crmAngService.loadForm).toHaveBeenCalledWith(providersPopupLists[0].popupFormUrl);
         expect(crmAngService.loadForm).toHaveBeenCalledWith(providersPopupLists[1].popupFormUrl);
+        expect(crmAngService.loadForm).toHaveBeenCalledWith(providersPopupLists[2].popupFormUrl);
```
igorpavlov: Same as above.

PR #2947, opened 2018-11-16
PCHR-4411: Fix logic for self leave approvers

igorpavlov:
## Overview

This PR applies new logic for **non-self-leave-approvers** admins:
1) They should not be able to select the Approved status from My Leave. Instead they should be treated as regular staff here.
2) They should not be able to select themselves from the Manager Leave screen (also from other screens).

### My Leave

image

### Manager Leave

image

### My Leave

**Note: there is no Leave Status selector.**

image

### Manager Leave

**Note: there is no "admin@example.com" user in the list because this is the current user.**

image

### Fixing the Status selector

We just need to refine the function that deals with the `vm.isSelfApprover` property. Initially this property depended on the role, now it clearly depends on the leave relationship between the admin and themselves:

```js
function initIsSelfLeaveApprover () {
  return loggedInContact.checkIfSelfLeaveApprover()
    .then(function (_canApproveOwnRequests_) {
      canApproveOwnRequests = _canApproveOwnRequests_;
      vm.isSelfLeaveApprover = !!(canApproveOwnRequests && vm.isSelfRecord);
``` 

### Excluding admin from the managed contacts list

Simply filter admin out if non-self-approver:

```js
function fetchManageesForAdmin () {
      return Contact.all()
        .then(function (contacts) {
          if (canApproveOwnRequests) {
            return contacts.list;

          // here
          return _.filter(contacts.list, function (contact) {
            return contact.id !== loggedInContact.id;
```

-----

✅Manual Tests - passed
✅Jasmine Tests - passed
✅JS distributive files - passed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not needed


mi-kon:
@igorpavlov keep in mind not only admins, but also managers have access to manager leave page. It's defined by 'manage L&A' permission

PR #2951, opened 2018-11-20
PCHR-4334: Access Funder Options Without Navigating to Settings Page

ajesamson:
## Overview
Funder providers were previously loaded from contacts but https://github.com/compucorp/civihr/pull/2920 creates an option group meant to replace this. This requires pointing all providers select menu to their option values instead of organisation contact.

## Before
before_funder

## After
after_funder

after_funder

## Technical Details
A tool icon was added to funder. This allows user to access funder provider options without navigating to settings page.
```
<div class="col-sm-2">
  <a class="pointer">
    <i class="crm-i fa-wrench" ng-click="jobroles.openOptionsEditor('hrjc_funder')"></i>
  </a>
</div>
```
```
<ui-select
  class="funder-select"
  prevent-animations
  theme="civihr-ui-select"
  ng-model="funder.funder_id">
  <ui-select-match prevent-animations allow-clear placeholder="Select a funder" class="ui-select-match">
    {{$select.selected.value.title}}
  </ui-select-match>
  <ui-select-choices
    prevent-animations
    class="ui-select-choices"
    repeat="funder.value.value as (key,funder) in jobroles.FundersData | filter: { value: { title: $select.search }}">
      <div ng-bind-html="funder.value.title | highlight: $select.search"></div>
    </ui-select-choices>
</ui-select>
```

## Comment
Leveraging on existing option group implementation, the `hr_funder` was loaded just like other option groups. This meant the loaded option group will be an object with the format show below:
```
{
  value: {
    id: option_id,
    title: option_label,
    value: option_value,
    is_active: option_status
  }
}
```
Because of the way ui-select works, when provided with an object, the data is transformed into 
```
{
  key: index,
  value: {
    id: option_id,
    title: option_label,
    value: option_value,
    is_active: option_status
  }
}
```
This makes it difficult to properly render ui-select especially when updating existing record. Due to this, the existing funder list was modified to instead of setting `funder_id` as object, it is represented as string reference to funder's value. 
```
jobRole.funders.push({
  id: jobRole.funders.length + 1,
  amount: amountFunders[i],
  percentage: percentFunders[i],
  type: funderTypes[i],
  funder_id: funderContactIds[i]
});
```
The configuration for ui-select-choice was also modified to return funder value whenever an item is selected.
```
<ui-select-choices
  prevent-animations
  class="ui-select-choices"
  repeat="funder.value.value as (key,funder) in jobroles.FundersData | filter: { value: { title: $select.search }}">
  <div ng-bind-html="funder.value.title | highlight: $select.search"></div>
</ui-select-choices>
```

[updateFundersContactsList](https://github.com/compucorp/civihr/blob/master/com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js#L1070) implementation was removed in this PR because funders are now loaded from option values. Option value implementation prefetch all record, eliminating issue with partial fetching of records described in https://github.com/compucorp/civihr/pull/1279

------
✅Manual Tests - passed
✅Jasmine Tests - passed


com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
```diff
@@ -214,7 +214,18 @@
             <table class="table">
               <thead>
                 <tr>
-                  <th style="width:40%">Funder</th>
+                  <th style="width:40%">
+                    <div class="row">
```
igorpavlov: Bootstrap grid inside a table header looks a bit strange. Why would we need such a markup?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
```diff
@@ -225,23 +236,22 @@
                 <tr ng-repeat="(key, funder) in jobroles.editData['new_role_id']['funders']">
                   <td>
                     <!--show funder in non-edit mode-->
-                    <span ng-show="!tableform.$visible">{{funder.funder_id.sort_name}}</span>
```
igorpavlov: Why were we using `sort_name` and now `title`? What is the difference between two?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
```diff
@@ -225,23 +236,22 @@
...
+                          {{$select.selected.value.title}}
                         </ui-select-match>
                         <ui-select-choices
                           prevent-animations
                           class="ui-select-choices"
-                          repeat="contact in jobroles.contactList | filter: $select.search"
-                          refresh="jobroles.getContactList($select.search)"
-                          refresh-delay="500">
-                          <div ng-bind-html="contact.sort_name | highlight: $select.search"></div>
+                          repeat="funder.value.value as (key,funder) in jobroles.FundersData | filter: { value: { title: $select.search }}">
```
igorpavlov: `funder.value.value` just looks a bit odd. Could you please explain what is `funder.value` and what is `funder.value.value`? I have a feeling that `funder.value` should be `funder.properties` or smth.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-role-panel.html
```diff
@@ -326,7 +326,18 @@
               <table class="table">
                 <thead>
                   <tr>
-                    <th style="width:40%">Funder</th>
+                    <th style="width:40%">
```
igorpavlov: It seems that we have identical markup in 2 files. Do you think it may be a good idea to extract repeating code into a separate `html` partial file?

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -39,6 +39,7 @@ define([
     vm.LevelsData = {}; // Store the level types
     vm.LocationsData = {}; // Store the location types
     vm.RegionsData = {}; // Store the region types
+    vm.FundersData = {}; // Stores the list of funders
```
igorpavlov: Capitalised naming is used for classes like `const userAccount = new UserAccount()`. Please replace with camel case (`vm.fundersData = {}`). If you are feeling comfortable with the code, please rename the ones above as well (`LevelsData`, `LocationsData` etc, you can do a global case sensitive search)

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -1069,10 +995,10 @@ define([
      */
     function updateFundersContactsList (jobRoleFunders) {
       jobRoleFunders.forEach(function (funder) {
-        var funderData = funder.funder_id;
+        var funderId = funder.funder_id;
 
-        if (!_.includes(Object.keys(fundersContacts), funderData.id)) {
-          fundersContacts[funderData.id] = funderData;
+        if (!_.includes(Object.keys(fundersContacts), funderId)) {
```
igorpavlov: Please use `_.keys(fundersContacts)` it is shorter

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
```diff
@@ -872,10 +801,7 @@ define([
           });
         })
         .then(function () {
-          return getFundersContacts();
-        })
-        .then(function (contacts) {
-          fundersContacts = contacts;
+          fundersContacts = vm.FundersData;
```
igorpavlov: Why do we need two storages for the same data?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
ajesamson: I need to position the tool icon close to the `Funder` header. Floating the tool icon seems far away. 🤷‍♂️ 
tool

Can you please suggest a markup?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
ajesamson: The option values label were represented as [title](https://github.com/compucorp/civihr/blob/master/com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js#L535). This formed the reason for change as contact name were previously displayed using the `sort_name` column data from civicrm_contact table.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-roles.controller.js
ajesamson: This was initially implemented in https://github.com/compucorp/civihr/pull/1279 because the pool of contacts pre-fetched by the lookup dropdown is limited to 25. We can do without this because available option values were all prefetched. References will be removed.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-role-panel.html
ajesamson: I will suggest we leave as is because then we will to start taking of additional logic particular to each. For instance when `Add New Funder` is clicked, different parameters are passed across which may also make page partial file more complicated 🤷‍♂️

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
igorpavlov: I see. The problem you have probably faced is that `<a>` inside `<th>` becomes a **block** and gets rendered underneath the "Funders" label. I think your solution may not be the best one because:
1) it takes to much markup
2) it renders a grid inside a table heading
3) It is caption length dependent
4) it renders the wrench on a distance from the caption

I would suggest try smth like this:

```html
<th style="width:40%">
  Funder
  <button class="btn btn-link btn-xs"
    ng-click="jobroles.openOptionsEditor('hrjc_funder')">
    <i class="btn btn-link btn-xs fa fa-wrench"></i>
  </button>
</th>
```

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
igorpavlov: Shouldn't it be then smth like `funder.properties.value`?

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-role-panel.html
igorpavlov: Fair enough.

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
igorpavlov: right. let's leave it like it is now.

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
```diff
@@ -11,7 +11,7 @@
       </div>
     </header>
   </div>
-  <div class="panel panel-default panel-nested">
+  <span class="panel panel-default panel-nested">
```
igorpavlov: why changed to span?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
```diff
@@ -250,7 +250,9 @@
                         <ui-select-choices
                           prevent-animations
                           class="ui-select-choices"
-                          repeat="funder.value.value as (key,funder) in jobroles.FundersData | filter: { value: { title: $select.search }}">
+                          repeat="funder.value.value as (key,funder) in jobroles.fundersData | filter: { value: { title: $select.search }}"
+                          refresh="jobroles.getOptionGroupTypeValues('hrjc_funder')"
+                          refresh-delay="500">
```
igorpavlov: Why has it been added?

igorpavlov:
Which method did you use to replace all the occurences of `FundersData` and others?

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-role-panel.html
```diff
@@ -362,7 +362,9 @@
                           <ui-select-choices
                             prevent-animations
                             class="ui-select-choices"
-                            repeat="funder.value.value as (key,funder) in jobroles.FundersData | filter: { value: { title: $select.search }}">
+                            repeat="funder.value.value as (key,funder) in jobroles.fundersData | filter: { value: { title: $select.search }}"
+                            refresh="jobroles.getOptionGroupTypeValues('hrjc_funder')"
+                            refresh-delay="500">
```
igorpavlov: Same as above.

ajesamson:
> Which method did you use to replace all the occurences of `FundersData` and others?

I did a global search from my editor. Did you notice any missing update?

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
ajesamson: That must have been a mistake. I will revert. Thanks for catching that 👍

com.civicrm.hrjobroles/js/src/job-roles/controllers/add-new-role.html
ajesamson: The previous implementation has provision for refresh which was omitted in my initial implementation. It is needful as it ensure user always have updated list of funders (in case a new funder was added after page load by another user). The funders list also get updated when the tool icon is clicked but that is a longer route.

com.civicrm.hrjobroles/js/src/job-roles/controllers/job-role-panel.html
ajesamson: This was omitted in the initial implementation while refactoring the ui-select.

PR #2959, opened 2018-11-23
PCHR-4329: Migrate existing organisation contact

ajesamson:
## Overview
Existing organisation contacts were used for health, life and pension insurance provider. Having created an option group in https://github.com/compucorp/civihr/pull/2919 for managing list of providers, this PR migrates providers in existing job roles by creating option value replica.

## Before
Providers in organisation contact were not migrated to option value for their respective option group

## After
Providers which are previously in contact are migrated to option value for their respective option group.

## Technical Details
The install file in [hrjobcontract](https://github.com/compucorp/civihr/blob/master/hrjobcontract/sql/install.sql) was manually modified to remove the contact constraint on `civicrm_hrjobcontract_pension` and `civicrm_hrjobcontract_health`. This was due lack of proper integration with CiviCRM GenCode implementation. Their column types were also changed to varchar(255), the data type used for option value's value column. Their DAO were also modified for same reason, change the data type, maxlength, and pseudoconstant reference to their option group. For existing sites, an upgrader caters for modifying the tables
```
$healthTableName = CRM_Hrjobcontract_BAO_HRJobHealth::getTableName();
CRM_Core_DAO::executeQuery("
  ...
");

$pensionTableName = CRM_Hrjobcontract_BAO_HRJobPension::getTableName();
CRM_Core_DAO::executeQuery("
  ...
");
```

An upgrader was written to migrate the existing records to option value. The upgrade copied the details of each contact
```
$result = civicrm_api3('Contact', 'get', [
  'contact_sub_type' => ['IN' => $provider],
]);
```
and created another record with the same details. 
```
$params = [
  'option_group_id' => 'hrjc_' .  strtolower($contactType),
  'value' => $contact['id'],
  'name' => $contact['organization_name'],
  'label' => $contact['sort_name']
];
```
Due to possibilities of having two different contacts with the same name, all organisation (irrespective of name) were migrated separately to option value. The contact ids become the option value. This means there is no need of updating existing provider record. Also, this helps avoid issue with upgrader overriding a previously updated provider record due to its id being equal to the new ones to be updated.
```
 civicrm_api3('OptionValue', 'create', $params);
```
The provider's option value are not reserved. Any duplicate option value can be safely migrated and the duplicate ones deleted by users.

## Comment
Migrated contact records were not deleted in this PR. This was meant to be a fail-safe recovery mechanism. This will be handled in a separate PR.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobHealth.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobHealth.php
@@ -260,9 +260,9 @@ class CRM_Hrjobcontract_DAO_HRJobHealth extends CRM_Hrjobcontract_DAO_Base
                 'where' => 'civicrm_hrjobcontract_health.provider_life_insurance',
                 'headerPattern' => '',
                 'dataPattern' => '',
-                'pseudoconstant' => array(
+                'pseudoconstant' => [
                   'optionGroupName' => 'hrjc_life_insurance_provider'
-                ),
+                ],
                 'headerPattern' => '/^life\s?insurance\s?provider/i',
                 'entity' => 'HRJobHealth',
                 'bao' => 'CRM_Hrjobcontract_DAO_HRJobHealth',
--- a/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobHealth.php
+++ b/hrjobcontract/CRM/Hrjobcontract/DAO/HRJobHealth.php
@@ -198,9 +198,9 @@ class CRM_Hrjobcontract_DAO_HRJobHealth extends CRM_Hrjobcontract_DAO_Base
                 'where' => 'civicrm_hrjobcontract_health.provider',
                 'headerPattern' => '',
                 'dataPattern' => '',
-                'pseudoconstant' => array(
+                'pseudoconstant' => [
                   'optionGroupName' => 'hrjc_health_insurance_provider'
-                ),
+                ],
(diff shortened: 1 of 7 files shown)


ajesamson:
@davialexandre
I noticed that after running couple of new sites using this branch, pension_type column (civicrm_hrjobcontract_pension table) retained the old column type. An upgrader was previously [created](https://github.com/compucorp/civihr/blob/master/hrjobcontract/CRM/Hrjobcontract/Upgrader.php#L1067) which add constraint without updating the install file. My upgrader removed  this, but i can't say what could be the reason for this behavior on new sites (though the constraints could have prevented changes on setup). Existing site upgrade works fine. Kindly have a look and suggest please.

hrjobcontract/CRM/Hrjobcontract/DAO/HRJobHealth.php
```diff
@@ -191,14 +189,18 @@ static function &fields()
...
                 'import' => true,
+                'maxlength' => 512,
+                'size' => CRM_Utils_Type::HUGE,
                 'where' => 'civicrm_hrjobcontract_health.provider',
                 'headerPattern' => '',
                 'dataPattern' => '',
-                'FKClassName' => 'CRM_Contact_DAO_Contact',
+                'pseudoconstant' => array(
+                  'optionGroupName' => 'hrjc_health_insurance_provider'
+                ),
```
davialexandre: Short array syntax please (also applies to the other pseudoconstants)

PR #2962, opened 2018-11-29
PCHR-4329: Separated constraint checks on pension from column modification

ajesamson:
## Overview
Migration of contacts was previously implemented in https://github.com/compucorp/civihr/pull/2959. This PR separates removal of `pension_type_contact_id_fk` and `pension_type` column modification due to API Error.

## Before
Deployment raises db error when `pension_type_contact_id_fk` foreign key does not exist.

## After
Deployment runs smoothly.

## Comments
When running upgrade, dropping foreign key and modifying the column results in an error when foreign key does not exist on a site db. The previous implementation was therefore updated to ensure foreign key was first removed (if present) before going ahead with column modification.

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Upgrader.php
@@ -1147,11 +1147,17 @@ class CRM_Hrjobcontract_Upgrader extends CRM_Hrjobcontract_Upgrader_Base {
    */
   private function updateJobContractPensionSchema() {
     $pensionTableName = CRM_Hrjobcontract_BAO_HRJobPension::getTableName();
-    CRM_Core_DAO::executeQuery("
-      ALTER TABLE $pensionTableName
-      MODIFY pension_type VARCHAR(512),
-      DROP FOREIGN KEY  pension_type_contact_id_fk
-    ");
+    $pensionTypeConstraintExist =  CRM_Core_DAO::checkConstraintExists(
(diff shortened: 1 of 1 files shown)


hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1147,11 +1147,17 @@ private function updateJobContractHealthSchema() {
    */
   private function updateJobContractPensionSchema() {
     $pensionTableName = CRM_Hrjobcontract_BAO_HRJobPension::getTableName();
-    CRM_Core_DAO::executeQuery("
-      ALTER TABLE $pensionTableName
-      MODIFY pension_type VARCHAR(512),
-      DROP FOREIGN KEY  pension_type_contact_id_fk
-    ");
+    $pensionTypeConstraintExist =  CRM_Core_DAO::checkConstraintExists(
```
tunbola: Seems the space after the `=`  needs to be reduced.

hrjobcontract/CRM/Hrjobcontract/Upgrader.php
```diff
@@ -1147,11 +1147,17 @@ private function updateJobContractHealthSchema() {
...
+      $pensionTableName,
+      'pension_type_contact_id_fk'
```
tunbola: You can extract this `pension_type_contact_id_fk` into a variable since you are using the string in more than one place

PR #2964, opened 2018-12-05
PCHR-4431: Remove the validation that allows only one absence type with mstphl

ajesamson:
## Overview
When creating leave request, there is a validation that ensures only one leave type can have `Must Take Public Holiday as Leave` set to true. This PR removes the validation based on requirement to allow creation of multiple leave type having public holiday added.

## Before
Creating multiple leave type with public holiday was not possible.

## After
Creation of multiple leave type with public holiday is now possible.

## Technical Details
The validation [implementations](https://github.com/compucorp/civihr/blob/master/uk.co.compucorp.civicrm.hrleaveandabsences/CRM/HRLeaveAndAbsences/BAO/AbsenceType.php#L235) for `validateMustTakePublicHolidayAsLeave` were removed from HRLeaveandabsences AbsenceType BAO.


tunbola:
At overview and PR title, looks like the right acronym is MTPHL.
In the before and after section it's better to refer to the change as `creating multiple leave type with MTPHL as true` not possible in before section and then as possible in the after section. Reason is that this setting only controls if public holiday leave requests are created by the system for this absence type or not and it is a bit confusing the way you currently describe it.

You might also want to check the unit tests as there would likely be some tests testing this functionality you just removed.

PR #2968, opened 2018-12-10
PCHR-4432:  Add support for creation of Public Holiday Requests for Absence Types with MTPHL Via Background Job

tunbola:
## Overview
Due to recent changes that allows more than one Absence type to have the MTPHL property as true (see: #2964). This PR makes changes to the background job that creates these public holiday leave requests to accommodate creating public holiday leave requests for as many absence types that have MTPHL as true.

## Before
- The scheduled job that is responsible for creating public holiday leave requests will only create public holiday leave requests for only one absence type with MTPHL as true for contacts.

## After
- The scheduled job that is responsible for creating public holiday leave requests will create public holidays for as many absence types with MTPHL as true for contacts.

## Technical Details
- Apart from modifying the `PublicHolidayLeaveRequestCreation` class to now fetch all absence types with MTPHL using the new `AbsenceType::getAllWithMustTakePublicHolidayAsLeaveRequest()` method, some notable business logic was changed.  
One of such is that when a public holiday leave request is created, it will Zero the balance of any existing leave request already existing for same date. Since multiple public holidays can now exist for same date, this was also affecting an existing public holiday leave request by zeroing the balance change of an existing public holiday leave request.
A new `LeaveBalanceChange::getBalanceChangeToModifyForLeaveDate` method was created to address this issue, the method returns the balance change that can be modified when a public leave request is created or deleted in the system.
When a public holiday is deleted in the system, the balance change of an existing leave request is modified back to the state that it was before the leave request balance was zeroed. However if an existing public holiday request still exists for this date, then the leave request balance should still be kept at zero until the last public holiday leave request is deleted.

```php
  public static function getBalanceChangeToModifyForLeaveDate(LeaveRequest $leaveRequest, DateTime $date) {
    $balanceChanges = self::getBalanceChangesForLeaveRequestDate($leaveRequest, $date);

    if (empty($balanceChanges)) {
      return NULL;
    }

    $leaveBalanceChangeTypes = array_flip(self::buildOptions('type_id', 'validate'));

    $publicHolidayLeaveBalances = FALSE;
    $otherLeaveBalances = [];
    foreach ($balanceChanges as $balanceChange) {
      if ($balanceChange->type_id == $leaveBalanceChangeTypes['public_holiday']) {
        $publicHolidayLeaveBalances = TRUE;
        break;
      }

      $otherLeaveBalances[] = $balanceChange;
    }

    if (!$publicHolidayLeaveBalances && count($otherLeaveBalances) == 1) {
      return $otherLeaveBalances[0];
    }

    return NULL;
  }

```
- Also the current behaviour that when a public holiday leave request is created, the balance change of an existing leave request is changed to zero irrespective of the absence type of the public holiday leave request and that of the existing leave request is maintained.

### DIFF
--- a/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
+++ b/uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
@@ -307,6 +307,77 @@ class CRM_HRLeaveAndAbsences_Service_LeaveRequestTest extends BaseHeadlessTest {
     $this->assertEquals(-1, LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest));
   }
 
+  public function testPublicHolidayLeaveRequestIsDeletedAndBalanceNotRecalculatedForOverlappingLeaveRequestDateIfThereIsAnotherPublicHolidayExistingForThatDate() {
+    $adminID = 4;
+    $this->registerCurrentLoggedInContactInSession($adminID);
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
+
+    WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => 1]);
+
+    $tableName = CRM_HRLeaveAndAbsences_BAO_AbsenceType::getTableName();
+    CRM_Core_DAO::executeQuery("DELETE FROM {$tableName}");
+    $absenceType1 = AbsenceTypeFabricator::fabricate(['must_take_public_holiday_as_leave' => TRUE]);
+    $absenceType2 = AbsenceTypeFabricator::fabricate(['must_take_public_holiday_as_leave' => TRUE]);
+
+    HRJobContractFabricator::fabricate(
+      ['contact_id' => $this->leaveContact],
+      ['period_start_date' => '2016-01-01']
+    );
+
+    $publicHoliday = new PublicHoliday();
+    $publicHoliday->date = CRM_Utils_Date::processDate('2016-10-10');
+
+    $leaveRequest = LeaveRequestFabricator::fabricateWithoutValidation([
+      'from_date' => CRM_Utils_Date::processDate($publicHoliday->date),
+      'to_date' => CRM_Utils_Date::processDate($publicHoliday->date),
+      'contact_id' => $this->leaveContact,
+      'type_id' => $absenceType1->id,
+      'status_id' => 1
+    ], TRUE);
+
+    $this->assertEquals(-1, LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest));
+
+    PublicHolidayLeaveRequestFabricator::fabricate($this->leaveContact, $publicHoliday);
+
+    // Check that two public holiday leave requests are created for the date since there are two
+    // absence types with MTPHL as true.
+    $leaveRequests = LeaveRequest::findAllPublicHolidayLeaveRequests($this->leaveContact, $publicHoliday);
+    $this->assertCount(2, $leaveRequests);
+    $publicHolidayLeaveRequest1 = LeaveRequest::findPublicHolidayLeaveRequest(
+      $this->leaveContact,
+      $publicHoliday,
+      $absenceType1
+    );
+    $publicHolidayLeaveRequest2 = LeaveRequest::findPublicHolidayLeaveRequest(
+      $this->leaveContact,
+      $publicHoliday,
+      $absenceType2
+    );
+
+    //Balance change for Leave request will be zero since public holiday leave requests exists for same date
+    $this->assertEquals(0, LeaveBalanceChange::getTotalBalanceChangeForLeaveRequest($leaveRequest));
+
+    //Delete one of the public holiday leave requests.
+    $this->getLeaveRequestServiceWhenCurrentUserIsAdmin()->delete($publicHolidayLeaveRequest1->id);
+
+    //After deletion, one public holiday leave request is still present for the leave request date and
+    //the leave request balance change is untouched even though the deleted public holiday has same absence
+    //type as leave request.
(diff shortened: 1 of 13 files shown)


uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -380,6 +421,10 @@ public function testItCreatesLeaveRequestsForAllContactsWithContractsOverlapping
     $contact1 = ContactFabricator::fabricate();
     $contact2 = ContactFabricator::fabricate();
 
+    AbsenceType::del($this->absenceType->id);
+    $absenceType1 = AbsenceTypeFabricator::fabricate(['must_take_public_holiday_as_leave' => 1]);
+    $absenceType2 = AbsenceTypeFabricator::fabricate(['must_take_public_holiday_as_leave' => 1]);
+
```
davialexandre: This also seems to be repeated in multiple tests. Please check if it's possible to avoid that

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestDeletionTest.php
```diff
@@ -648,16 +710,16 @@ public function testCanDeleteAllPublicHolidayLeaveRequestsForAbsencePeriodForSpe
...
     //Delete all Public holiday leave requests for absence period for first contact only
     $deletionLogic = new PublicHolidayLeaveRequestDeletion(new JobContractService());
     $deletionLogic->deleteAllForAbsencePeriod($period, [$contact['id']]);
 
     //Leave Request for PublicHoliday will be deleted for contact1 only
-    $this->assertNull(LeaveRequest::findPublicHolidayLeaveRequest($contact['id'], $publicHoliday));
+    $this->assertNull(LeaveRequest::findPublicHolidayLeaveRequest($contact['id'], $publicHoliday, $this->absenceType));
 
-    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday));
+    $this->assertNotNull(LeaveRequest::findPublicHolidayLeaveRequest($contact2['id'], $publicHoliday, $this->absenceType));
```
davialexandre: A comment not about this specific test, but about all the tests in general: I can see you've updated quite a lot of tests to make sure they work fine with multiple absence types. I wonder if it wouldn't be simpler to have separate tests for these new scenarios and only do that where multiple MTPHL is really relevant (that is, if there's already another test checking the multiple absence type logic, do we really need to repeat that in other places?)

tunbola: I think we have quite a lot of tests already and I didn't want to add more than necessary, hence the approach to modify some of the tests to use multiple absence types. I only modified some that are necessary as there are still quite a LOT more that use single absence types.

The aim was just to modify some of the tests mainly tests that are responsible for creating public holiday requests via the various entry points like work pattern, absence period and so on.
If we add new tests for all these scenarios it will be almost a repetition of the tests that use singular absence types as the only difference will be that one test use one absence type while the other uses multiple types and the aim of both variant of tests is just to demonstrate that public holiday leave requests will be created for as many absence type that has MTPHL.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
tunbola: Not sure what you mean and how we can avoid this. I will add a comment for why existing absence types are deleted like you mentioned, but for most cases in the tests, the absence types ids need to be used in the actual tests and would not make the test clear if these were extracted to a function.

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/PublicHolidayLeaveRequestCreationTest.php
```diff
@@ -50,27 +50,38 @@ public function setUp() {
...
+  public function testCanCreatePublicHolidayLeaveRequestsForASingleContactWithCorrectBalanceChanges() {
     $absencePeriod = AbsencePeriodFabricator::fabricate([
       'start_date' => CRM_Utils_Date::processDate('first day of this year'),
       'end_date' => CRM_Utils_Date::processDate('last day of this year')
     ]);
 
-    $periodEntitlement = $this->createLeavePeriodEntitlementMockForBalanceTests();
-    $periodEntitlement->contact_id = 2;
-    $periodEntitlement->type_id = $this->absenceType->id;
+    AbsenceType::del($this->absenceType->id);
```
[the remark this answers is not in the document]
tunbola: I'll add a comment

uk.co.compucorp.civicrm.hrleaveandabsences/tests/phpunit/CRM/HRLeaveAndAbsences/Service/LeaveRequestTest.php
```diff
@@ -296,7 +303,78 @@ public function testPublicHolidayLeaveRequestIsDeletedAndBalanceRecalculatedForO
...
+    $this->registerCurrentLoggedInContactInSession($adminID);
+    AbsencePeriodFabricator::fabricate([
+      'start_date' => CRM_Utils_Date::processDate('2016-01-01'),
+      'end_date' => CRM_Utils_Date::processDate('2016-12-31')
+    ]);
+
+    WorkPatternFabricator::fabricateWithA40HourWorkWeek(['is_default' => 1]);
+
+    $tableName = CRM_HRLeaveAndAbsences_BAO_AbsenceType::getTableName();
+    CRM_Core_DAO::executeQuery("DELETE FROM {$tableName}");
```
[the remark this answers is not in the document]
tunbola: Will extract the deletion to a function.

PR #2974, opened 2018-12-14
PCHR-4448: Fix Warning and Notice Messages on Import Job Contract Wizard

tunbola:
## Overview
When trying to import Job contracts on a site running PHP 7.2.X. Some warnings occur on the various stages of the import screen.

## Before
- Some errors on the various stages of the Job contract import screen.

contractbefore

## After
- One of the errors was related to an error regarding using count function on a variable that does not implement the Countable interface. This warning is displayed by PHP starting from PHP 7.2.
The [here](https://github.com/compucorp/civihr/blob/fb73df3b053bec0bcbadc5f9362704cc14ab8b15/hrjobcontract/CRM/Hrjobcontract/Import/Parser/BaseClass.php#L88) returns NULL sometimes and an error is thrown when using count on the variable h[ere](https://github.com/compucorp/civihr/blob/fb73df3b053bec0bcbadc5f9362704cc14ab8b15/hrjobcontract/CRM/Hrjobcontract/Import/Parser.php#L319). This was fixed by making sure an Array will always be passed to the `setActiveFieldLocationTypes` function.
- There was an issue type casting an Object to an int [here](https://github.com/compucorp/civihr/blob/693e69867d7fbb0e6e8c9514e8a26fa598a8af1a/hrjobcontract/CRM/Hrjobcontract/Import/Parser/Api.php#L571) since the `CRM_Hrjobcontract_BAO_HRJobContract::checkContract` function returns either an object or 0. This was simply fixed by checking if the variable returned is false or not. See [here](https://3v4l.org/cKGGb) for demonstration.
- There was some other error reported in the ticket, which has been fixed on #2973 

contractafter

### DIFF
--- a/hrjobcontract/CRM/Hrjobcontract/Import/Parser/BaseClass.php
+++ b/hrjobcontract/CRM/Hrjobcontract/Import/Parser/BaseClass.php
@@ -82,8 +82,10 @@ class CRM_Hrjobcontract_Import_Parser_BaseClass extends CRM_Hrjobcontract_Import
       $field['headerPattern'] = CRM_Utils_Array::value('headerPattern', $field, '//');
       $this->addField($name, CRM_Utils_Array::value('title', $field, $name), $field['type'], $field['headerPattern'], $field['dataPattern']);
     }
+
+    $mapperLocType = !empty($mapperLocType) ? $mapperLocType : [];
     $this->setActiveFields($this->_mapperKeys);
-    $this->setActiveFieldLocationTypes($this->_mapperLocType);
+    $this->setActiveFieldLocationTypes($mapperLocType);
 
     // Fetch select list options from the database and cache them
     $this->_optionsList['HRJobHour-hours_type'] = CRM_Hrjobcontract_SelectValues::buildDbOptions('hrjc_hours_type');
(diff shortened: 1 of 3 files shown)


ajesamson:
Kindly update the title as `PCHR-4448:`

hrjobcontract/CRM/Hrjobcontract/Import/Parser/BaseClass.php
```diff
@@ -82,8 +82,10 @@ function init() {
       $field['headerPattern'] = CRM_Utils_Array::value('headerPattern', $field, '//');
       $this->addField($name, CRM_Utils_Array::value('title', $field, $name), $field['type'], $field['headerPattern'], $field['dataPattern']);
     }
+
+    $mapperLocType = !empty($mapperLocType) ? $mapperLocType : [];
     $this->setActiveFields($this->_mapperKeys);
-    $this->setActiveFieldLocationTypes($this->_mapperLocType);
```
ajesamson: Since this line was removed, is `$this->_mapperLocType` still being used in other classes? If not i guess we can remove the declaration 🤷‍♂️

tunbola: Oops. It wasn't meant to be removed. Added back. Nice catch!

PR #2979, opened 2018-12-20
PCHR-4444: Funders of Type Individual is not saved with the correct Contact ID

ajesamson:
## Overview
When making request via Contact API, response usually contains job role id if the requested contact has existing job role. This was because job role id is [exportable](https://github.com/compucorp/civihr/blob/master/com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php#L273) and its key [overrides](https://github.com/civicrm/civicrm-core/blob/master/CRM/Contact/BAO/Query.php#L487) that of contact id when loading contact details. 
This process also have effect on adding funder to individual contact type. Instead of saving contact id of the funder, funder's job role id was used in place of its contact id if contact happens to have existing job role.

## Before
Contact API request returns funder id instead of contact id when contact have an existing job role.

## After
Contact API request returns contact id even when the contact have an existing job role.

## Technical Details
A call to the contact api attempts to [fetch](https://github.com/civicrm/civicrm-core/blob/master/CRM/Contact/BAO/Query.php#L486) DAO exportable fields provided by hook implementers of CRM_Contact_BAO_Query_Hook. CRM_Hrjobroles_BAO_Query extends the CRM_Contact_BAO_Query_Interface interface, which means its fields will likely be loaded when fetching contacts especially when specific columns are to be returned. 
Key definition for HRJobroles DAO was changed from `id` to `hrjobroles_id` to avoid conflict with that of contact id
```
'hrjobroles_id' => array(
  'name' => 'id',
  'type' => CRM_Utils_Type::T_INT,
  'required' => true,
  'export' => true,
  'title' => 'Job Role ID',
  'where' => 'civicrm_hrjobroles.id',
  'entity' => 'HrJobRoles',
  'bao' => 'CRM_Hrjobroles_DAO_HrJobRoles',
  'localizable' => 0,
) ,
```

An upgrader was written to cater for situations where individual contact type used as funder had their job roles saved instead of contact id. All existing job role with funder were fetch via api
```
$jobRoles = civicrm_api3('HrJobRoles', 'get', [
  'funder' => ['!=' => '|'],
]);
```
The funders for each job roles were iterated over. A check was made to see if funder matches any contact's jobrole_id. 
```
$jobRoleIds = array_column($contactDetails, 'jobrole_id');
$funderIndex = array_search($funder, $jobRoleIds);
```
If the funder was found, its `id` was used to update the affected job role funder.
```
civicrm_api3('HrJobRoles', 'create', [
  'id' => $key,
  'funder' => '|' . implode('|', $funders) . '|',
]);
```

### DIFF
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
@@ -266,7 +266,7 @@ class CRM_Hrjobroles_DAO_HrJobRoles extends CRM_Core_DAO
   {
     if (!(self::$_fields)) {
       self::$_fields = array(
-        'id' => array(
+        'hrjobroles_id' => array(
           'name' => 'id',
           'type' => CRM_Utils_Type::T_INT,
           'required' => true,
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
@@ -266,7 +266,7 @@ class CRM_Hrjobroles_DAO_HrJobRoles extends CRM_Core_DAO
   {
     if (!(self::$_fields)) {
       self::$_fields = array(
-        'hrjobroles_id' => array(
+        'hr_job_roles_id' => array(
           'name' => 'id',
           'type' => CRM_Utils_Type::T_INT,
           'required' => true,
--- a/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
+++ b/com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
@@ -79,6 +79,85 @@ class CRM_Hrjobroles_Upgrader extends CRM_Hrjobroles_Upgrader_Base {
   }
 
   /**
+   * Updates job role having wrong funder id with correct contact id
+   *
+   * @return bool
+   * @throws CiviCRM_API3_Exception
+   */
+  public function upgrade_1007() {
+    $contactDetails = $this->getContactsWithJobRole();
+    $jobRoles = civicrm_api3('HrJobRoles', 'get', [
+      'funder' => ['!=' => '|'],
+    ]);
+
+    foreach($jobRoles['values'] as $key => $value) {
+      $funders = explode('|', $value['funder']);
+      $funders = array_filter($funders);
+
+      list($funders, $allowUpdate) = $this->changeFunderToCorrectContactId($funders, $contactDetails);
+      if ($allowUpdate) {
+        civicrm_api3('HrJobRoles', 'create', [
+          'id' => $key,
+          'funder' => '|' . implode('|', $funders) . '|',
+        ]);
+      }
+    }
+
+    return TRUE;
+  }
+
+  /**
+   * Retrieves the correct contact for a funder and
+   * indicates update is required when found
+   *
+   * @param $funders
+   * @param $contactDetails
+   *
+   * @return array
+   */
+  private function changeFunderToCorrectContactId($funders, $contactDetails) {
+    $allowUpdate = FALSE;
+    foreach ($funders as $index => $funder) {
+      $jobRoleIds = array_column($contactDetails, 'jobrole_id');
+      $funderIndex = array_search($funder, $jobRoleIds);
+      if ($funderIndex) {
+        $funders[$index] = array_values($contactDetails)[$funderIndex]['id'];
+        $allowUpdate = TRUE;
+      }
+    }
+
+    return [$funders, $allowUpdate];
+  }
+
+  /**
+   * Retrieves contacts with their job contract and job role
+   *
+   * @return array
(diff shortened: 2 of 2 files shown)


com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,85 @@ public function upgrade_1006() {
...
+
+    return TRUE;
+  }
+
+  /**
+   * Retrieves the correct contact for a funder and
+   * indicates update is required when found
+   *
+   * @param $funders
+   * @param $contactDetails
```
davialexandre: missing param types

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,85 @@ public function upgrade_1006() {
...
+   *
+   * @return array
+   */
+  private function changeFunderToCorrectContactId($funders, $contactDetails) {
+    $allowUpdate = FALSE;
+    foreach ($funders as $index => $funder) {
+      $jobRoleIds = array_column($contactDetails, 'jobrole_id');
+      $funderIndex = array_search($funder, $jobRoleIds);
+      if ($funderIndex) {
```
davialexandre: Please check the documentation for the array_search function. It's not safe to use it like this.

davialexandre: Since you use PhpStorm, it would be good if you could install the Php Inspections extension. It would have warned you about this

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,85 @@ public function upgrade_1006() {
...
+    $query = "
+      SELECT c.id, jc.id as contract_id, jr.id AS jobrole_id, jr.funder
+      FROM `civicrm_contact` c
+      INNER JOIN civicrm_hrjobcontract jc ON (c.id = jc.contact_id)
+      LEFT JOIN `civicrm_hrjobroles` jr ON (jc.id = jr.job_contract_id)
+      WHERE c.contact_type = 'Individual'
+    ";
+
+    $result = CRM_Core_DAO::executeQuery($query);
+    while ($result->fetch()) {
```
davialexandre: You don't need the loop here. You can get the same array by simply calling `$result->fetchAll()`

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,85 @@ public function upgrade_1006() {
     return TRUE;
   }
 
+  /**
+   * Updates job role having wrong funder id with correct contact id
```
davialexandre: Please explain here why some job roles have a wrong funder and which rules are used to decide what is the correct contact ID

com.civicrm.hrjobroles/CRM/Hrjobroles/Upgrader.php
```diff
@@ -78,6 +78,85 @@ public function upgrade_1006() {
...
+   * @param array $funders
+   * @param array $contactDetails
+   *
+   * @return array
+   */
+  private function changeFunderToCorrectContactId($funders, $contactDetails) {
+    $allowUpdate = FALSE;
+    foreach ($funders as $index => $funder) {
+      $jobRoleIds = array_column($contactDetails, 'jobrole_id');
+      $funderIndex = array_search($funder, $jobRoleIds, TRUE);
```
davialexandre: Could you please give more details about why you're doing a strict search? My concern here is that funder is a string (because it's one item from the array created with an explode), while `$jobRoleIds` is more likely to be an array of integers, which will fail in a strict comparison.

ajesamson: Using strict is not compulsory here actually. This was suggested when i used php inspection tool. Secondly, the `$jobRoleIds` were represented as array of strings when i tested it. I think its better to remove strict check 🤷‍♂️. What would you advice?

davialexandre: Remove it

com.civicrm.hrjobroles/CRM/Hrjobroles/DAO/HrJobRoles.php
```diff
@@ -266,7 +266,7 @@ static function &fields()
   {
     if (!(self::$_fields)) {
       self::$_fields = array(
-        'hrjobroles_id' => array(
+        'hr_job_roles_id' => array(
```
omarabuhussein: DAO files should be generated from the XML files and they should not be edited manually

ajesamson: Yes that is correct, but this extension have not followed that process for a while now. Plans are in place to ensure its properly migrated to follow appropriate standards. Until then, we just carefully ensure implementations are well tested.

PR #2988, opened 2019-08-12
PCHR-4528: Quick search and Menu fixes

reneolivo:
## Overview

* It changes the font for the Quick Search's placeholder from FontAwesome to Open Sans (the font used by Shoreditch).
* It removes an extra icon on the Quick Search's placeholder.
* It stops the main menu from overlapping on top of the content when viewing the app from a small screen.

### Before
Screen Shot 2019-08-12 at 6 14 40 PM

### After
Screen Shot 2019-08-12 at 6 13 07 PM

### Technical details

The placeholder has been added by core in the following commit:
https://github.com/civicrm/civicrm-core/commit/b30809e49e9ed5142092f2157877a24ad1df17ee#diff-5a8541b7a45cbcc9d60a8898921f4dddR145

The purpose was probably done to append a search icon to the quick search, but this font was also being used to display the placeholder text.

To fix this issue we have redefined the placeholder's font:

```scss
#crm-qsearch-input::placeholder {
  font-family: $font-family-base;
}
```

### Before
pr-before

### After
pr-after

### Technical details

The issue happens because in core the icon for the quick search is manually added to the placeholder each time a new field is selected. This change was introduced here:
https://github.com/civicrm/civicrm-core/commit/b30809e49e9ed5142092f2157877a24ad1df17ee#diff-5d6e10b762b26e6116b7b88ab3be98c7R358

In order to fix the issue we remove the icon added by core after the user has selected a new field:

```js
function onFieldSelectedRemoveSearchIconFromQuickSearch () {
  var quickSearchInput = $('#crm-qsearch-input');

  $('.crm-quickSearchField').click(function () {
    var selectedFieldName = $(this).text().trim();

    setTimeout(function () {
      quickSearchInput.attr('placeholder', selectedFieldName);
    });
  });
}
```

The timeout is needed to execute the placeholder replacement after core has applied changes to the placeholder.

### Before
pr-before

### After
pr-after

### Technical details

This issue happens because the body has a padding that should be around the same height as the HR Menu, but this padding is not set properly because the menu is loaded dynamically and there is no way to tell what the height will be using CSS alone.

To fix this we trigger a `civihr-menu::ready` event after the menu has been loaded and appended in `hrui`. This event is picked up by `org.civicrm.bootstrapcivihr/js/main-menu.js` and the right padding is applied:

```js
$(window).on('civihr-menu::ready', updateContentSpacingRelatedToHRMenu);
// also do the same when the window has been resized:
$(window).resize(updateContentSpacingRelatedToHRMenu);

// ...

function updateContentSpacingRelatedToHRMenu () {
  var mainContentArea = $('.crm-menubar-over-cms-menu')[0];
  var hrMenuHeight = $('#civihr-menu').height();

  mainContentArea.style.removeProperty('padding-top');
  mainContentArea.style.setProperty('padding-top', hrMenuHeight + 'px', 'important');
}
```

As explained in the comments, `style.setProperty()` is used instead of `.css()` because the `!important` rule can only be set with the former function, and this rule is needed because it's set as important by core.

## Comments

* Some lint issues were fixed for `org.civicrm.bootstrapcivihr/scss/overrides/crm/_civicrm-menu.scss`.
* manual tests were conduced to make sure the styles applied are the right ones.


igorpavlov:
@reneolivo I think such changes **may** be a bit dangerous.

https://github.com/compucorp/civihr/pull/2988/files#diff-13c5ff1c4a91aa33cc605ef1536e9ee8L20

In this particular case and other changes in this PR I can see, there would probably be not other elements with such attributes and also they are scoped to the search menu. However, in general, I would recommend to leave these selectors as is, if it is Civi.

